Issue #716, opened 2023-03-08
Allow the creation of a partial index


lukehutch:
**Is your feature request related to a problem? Please describe.**

I need to be able to index sparse boolean columns. The most efficient way to do that in Postgres is to create a partial index, e.g.

```SQL
CREATE INDEX ON mytable((1)) WHERE boolcolumn;
```

I don't see a way to express that in the current Serverpod yaml syntax. Can this be added, please?

vlidholt:
This should be added in the future. In the mean time, it's possible to manually add the index to your database using SQL.

lukehutch:
Also I'm not sure if `WHERE boolcolumn` is going to be interpreted the same way as `WHERE boolcolumn = TRUE`, in terms of whether the partial index is used or not, but the latter syntax would be better for Serverpod partial indexes, because the Serverpod query syntax requires testing `where: (t) => boolcolumn.equals(true)`.

Then the yaml syntax for the index would require the user to specify something like 

```yaml
indexes:
  tablename_boolcolumn_idx:
    fields: boolcolumn
    whereValueEquals: true
```

lukehutch:
The benefits of partial indices are quantified here:

https://stackoverflow.com/a/78725490/3950982

t-kietzmann:
Wanna push this again. Would be a great benefit, especially when working with soft-deletes.

PR #2473, opened 2024-07-11
Add basic partial index support for IS NOT NULL

lukehutch:

This adds basic support for the most useful kind of partial index (#716), the index on `IS NOT NULL`.

The syntax in the index definition is:

```yaml
class: Profile
table: profile
fields:
  name: String
indexes:
  profile_name_idx:
    fields: name
    nonNulls: true    # <---
```

This produces SQL of the form:

```sql
create index profile_name_idx on profile (name) where (name is not null);
```

This also works for multi-column indexes.

For columns consisting of mostly `NULL` values, this can dramatically reduce index size (and therefore also row insert time), by causing `NULL` values not to be indexed: https://stackoverflow.com/a/78725490/3950982

## Pre-launch Checklist


lukehutch:
OK, I made a lot of changes, but I think it's ready to go. I got stuck on some old migrations (somehow this couldn't pick up the changes, if I added or removed `nonNulls: true`), but I deleted all the old migrations and recreated the database, and now the migration system works perfectly when I add or remove this parameter. (In other words, please test this yourself -- I never figured out what got corrupted along the way, and probably I just screwed something up as I kept testing changes, which is now fixed.)

This is going to dramatically reduce my database size and row insertion time, since I have 146 indexes on sparse columns! (Most of them are indexes on a single table, so row insertion time there would suffer without this...)

lukehutch:
In fact I recommend making this type of partial index the default for any nullable field of a table model. It's rare that you would need to index every null value in a column. It's far more common that you'll use null to support sparse data within a column.

SandPod:
It's great that you have taken this on; this is a much-needed feature!

I unfortunately can't jump on a review for this right now but I hope to get around to it over the next couple of weeks.

lukehutch:
That's fine, thanks. I moved forward with this for production, given that it seems to work great without issues, but obviously I didn't write unit tests :)

jodinathan:
so why didn't we go forward with this?  
seems like a nice feature

SandPod:
@jodinathan - Since the PR did not include passing tests and the maintainer team didn't have capacity to take this on the PR grew stale. I believe this is open for grabs again.

The feature is a good addition to the framework.

marcelomendoncasoares:
Closing this PR since it is stale for a long time. If needed, a new PR can be opened targeting last `main`.

Issue #2450, opened 2024-07-08
Decouple logger logic and the session


Isakdl:
Decouple the logging logic from the session implementation, today the logger is tightly coupled with the session and introducing new session types requires explicit implementation of the logging logic. 

Instead, we should reverse the dependency to allow for a more generic approach letting us be more flexible with how we manage sessions.

PR #2479, opened 2024-07-12
refactor: Decouple log manager from runtime settings and serverpod.

Isakdl:
# Changes

- New SessionLogManager that is instantiated inside the Session object
- Old LogManager stripped of all internal methods
- Extensions for internal methods to hide them from the interfaces (to retrieve the session log manger)
- Removing logic to retrieve cached session logs with insights (e.g. logs that has yet to be written).

Related issue:  https://github.com/serverpod/serverpod/issues/2450

## Pre-launch Checklist

## Breaking changes

Small removal of the log writer from the LogManager, I think this is okay as you are not intended to instantiate this class yourself.


tests/serverpod_test_server/test_integration/logging/endpoint_call_logging_test.dart
```diff
@@ -313,6 +313,99 @@ void main() async {
       expect(logs.first.logs, hasLength(1));
       expect(logs.first.logs.first.logLevel, LogLevel.fatal);
     });
+
+    group('Overrides -', () {
```
SandPod: Suggestion: same thing here, maybe move the overrides to another file it is an isolated group.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -112,6 +116,17 @@ abstract class Session {
       _db = server.createDatabase(this);
     }
 
+    var logWriter = Features.enablePersistentLogging
```
SandPod, line 119: Discussion: I know that this is just moved code but we should probably support having both enabled. toggling each of them individually.

packages/serverpod/lib/src/server/log_manager/log_manager.dart
```diff
@@ -424,57 +395,40 @@ class LogManager {
     }
     return null;
   }
+}
 
-  /// Returns a list of logs for all open sessions.
-  List<SessionLogInfo> getOpenSessionLogs(
```
SandPod, line 429: Question: This means that we are removing support for this feature now, in insights now right? Should probably be noted in the description.

SandPod (commented):
AFAIK this looks good! I just have two questions.

tests/serverpod_test_server/test_integration/logging/endpoint_call_logging_test.dart
Isakdl: These where added back by mistake, removed them in a later commit.

packages/serverpod/lib/src/server/session.dart
Isakdl: Yeah, but in a future PR I will rewrite the LogWriters, this logic is still there but we can easily change how this works at a single location, it will make sense when you see the next itteration! :)

packages/serverpod/lib/src/server/log_manager/log_manager.dart
Isakdl: Yes indeed, I'm not sure if we use it, but discussed this with victor a week or so back and it seems like a rather redundant feature, hence removing it.

SandPod (approved):
LGTM! 👍

Issue #2338, opened 2024-06-03
RFC: Extending the streaming interface


vlidholt:
There is potential to improve the experience of working with streams and streaming data (currently implemented via web sockets). This issue outlines how a new interface can work.

![streams](https://github.com/serverpod/serverpod/assets/1539812/ee51d0d7-61cf-45d8-ab89-4871c1972597)

Instead of the current interface, make it possible to pass a `Stream` object as a parameter and return type of an endpoint method. Below is an example of what a bi-directional stream could look like.

```dart
Stream<MyModel> methodWithStreams(Session session, Stream<MyModel> stream) async * {
  await for (var message in stream) {
    yield MyModel(text: message.text);
  }
}
```

If the stream does not have a specified type, it will be considered to support any `SerializableModel` or custom serializable class type. The class name is always encoded together with the object (as it is today).

Streams set up this way remain open until they are either closed and the stream ends or if they end with an exception (e.g., if the connection goes down).

Stream messages are piped through the same WebSocket connection as other streaming messages today. When receiving or sending a stream, a WebSocket connection is automatically opened on the client side.

We can implement a `ReconnectingStream` on the client to maintain a stream that automatically reconnects to the server if it goes down. The `ReconnectingStream` would not close, even if the underlying stream disconnects. Instead, it will try to reconnect and start receiving messages once a connection is available.

_Note: This would be super handy in companion with inheritance in models:_
https://github.com/serverpod/serverpod/issues/1873

S-ecki:
+1 on this, this would be super handy! Imo this would be the least amount of friction for someone looking into serverpod coming from Flutter! 👍

vlidholt:
In addition, we may want to add a convenience method to Serverpod's internal messaging system. A common use case will be to listen from messages in the internal messaging and relaying them to the client. Consider this example (on an `Endpoint`):

```dart
Stream<DeckState> deckUpdates(Session session) async* {
  // Set up a stream for getting notified when we get messages from
  // MessageCentral.
  var streamController = StreamController<DeckState>();
  session.messages.addListener(kDeckUpdatesChannel, (update) {
    streamController.add(update as DeckState);
  });

  // Listen for updates and relay them to the client.
  try {
    await for (var update in streamController.stream) {
      yield update;
    }
  } catch (e) {
    // It's possible we never get here.
    streamController.close();
  }
}
```

[...]

PR #2480, opened 2024-07-12
fix: Encode method stream messages with encode for protocol.

SandPod:
Related to #2338 

### Changes:
- Encodes all method stream messages with encoding for the protocol to ensure that server-only fields are not leaked.
- Only encode serializable objects once by utilising wrapWithClassName instead of taking a string for message creation.
- Write any method stream errors to stderr.

## Pre-launch Checklist

## Breaking changes

None: touches unreleased features.

### DIFF
--- a/packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
+++ b/packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
@@ -505,7 +505,8 @@ class _MethodStreamManager {
             endpoint: message.endpoint,
             method: message.method,
             connectionId: message.connectionId,
-            object: server.serializationManager.wrapWithClassName(e),
+            object: e,
+            serializationManager: server.serializationManager,
           ),
         );
       }
@@ -533,7 +534,8 @@ class _MethodStreamManager {
           endpoint: message.endpoint,
           method: message.method,
           connectionId: message.connectionId,
-          object: server.serializationManager.wrapWithClassName(result),
+          object: result,
+          serializationManager: server.serializationManager,
         ),
       );
     }
@@ -565,7 +567,8 @@ class _MethodStreamManager {
             endpoint: message.endpoint,
             method: message.method,
             connectionId: message.connectionId,
-            object: server.serializationManager.wrapWithClassName(value),
+            object: value,
+            serializationManager: server.serializationManager,
           ),
         );
       },
@@ -588,7 +591,8 @@ class _MethodStreamManager {
               endpoint: message.endpoint,
               method: message.method,
               connectionId: message.connectionId,
-              object: server.serializationManager.wrapWithClassName(e),
+              object: e,
+              serializationManager: server.serializationManager,
             ),
           );
         }
--- a/packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
+++ b/packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
@@ -95,8 +95,10 @@ class MethodWebsocketRequestHandler {
         }
       }
     } catch (e, stackTrace) {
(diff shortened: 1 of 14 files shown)


packages/serverpod_serialization/test/websocket_messages_test.dart
```diff
@@ -1,29 +1,106 @@
...
+    return {
+      'data': data,
+    };
+  }
+
+  @override
+  factory _SimpleData.fromJson(Map<String, dynamic> json) {
+    return _SimpleData(json['data']);
+  }
+}
```
SandPod, line 64: Should add a server only field and validate that it is not included in the serialization.

packages/serverpod_serialization/lib/src/websocket_messages.dart
```diff
@@ -337,26 +350,30 @@ class MethodStreamMessage extends WebSocketMessage {
...
   /// Builds a [MethodStreamMessage] message.
+  /// The [object] must be an object processed by the
+  /// [SerializationManager.wrapWithClassName] method.
   static String buildMessage({
     required String endpoint,
     required String method,
     required UuidValue connectionId,
     String? parameter,
-    required String object,
+    required Map<String, dynamic> object,
```
klkucaj, line 374: Since this is a static method and can be called from anywhere, I suggest passing a `SerializationManager` instance as a parameter. This way, we can use `serializationManager.encodeWithType(...)`, ensuring that the `object` parameter always has the desired structure.

packages/serverpod_serialization/lib/src/websocket_messages.dart
```diff
@@ -281,43 +286,51 @@ class MethodStreamSerializableException extends WebSocketMessage {
...
   /// Builds a [MethodStreamSerializableException] message.
+  /// The [exception] must be a serializable exception processed by the
+  /// [SerializationManager.wrapWithClassName] method.
   static String buildMessage({
     required String endpoint,
     required String method,
     required UuidValue connectionId,
     String? parameter,
-    required String object,
+    required Map<String, dynamic> object,
```
klkucaj, line 313: Same goes for this one.

packages/serverpod_serialization/lib/src/websocket_messages.dart
SandPod: Good suggestion! will do that 👍

klkucaj (approved):

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -92,8 +95,10 @@ class MethodWebsocketRequestHandler {
         }
       }
     } catch (e, stackTrace) {
-      var session = await server.serverpod.createSession();
-      await session.close(error: e, stackTrace: stackTrace);
+      stderr
```
vlidholt, line 98: Will this not make the server noisy if a faulty client sends messages to it?

SandPod: That makes sense! I stole this logging approach from the current streaming implementation: https://github.com/serverpod/serverpod/blob/f5983a2bfb1578003274367d504ab87056f1355f/packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart#L136-L140

But I will add a safeguard only logging these issues if we have verbose (or some other suitable log setting) logging enabled.

SandPod: Added a guard to check if logging malformed calls is enabled. Do you think that is a good solution?

vlidholt: That sounds like the preferred way to do it! 😀

PR #2482, opened 2024-07-12
refactor: 

Isakdl:
# Changes

- Moves logging related variables away from the log cache class into the log manager.

Related issue: https://github.com/serverpod/serverpod/issues/2450

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/packages/serverpod/lib/src/server/log_manager/log_manager.dart
+++ b/packages/serverpod/lib/src/server/log_manager/log_manager.dart
@@ -310,10 +310,7 @@ class SessionLogManager {
   }) async {
     var duration = session.duration;
     var cachedEntry = session.sessionLogs;
-    LogSettings? logSettings;
-    if (session is! StreamingSession) {
-      logSettings = _settingsForSession(session);
-    }
+    LogSettings logSettings = _settingsForSession(session);
 
     if (session.serverpod.runMode == ServerpodRunMode.development) {
       if (session is MethodCallSession) {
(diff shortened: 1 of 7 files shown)


packages/serverpod/lib/src/server/log_manager/log_manager.dart
```diff
@@ -264,7 +264,7 @@ class SessionLogManager {
         return;
       }
 
-      assert(session.sessionLogs.currentEndpoint != null);
+      assert(session.endpoint != null);
```
SandPod, line 274: Quesiton: What happens if endpoint is null here? Should we throw or is that fine?

packages/serverpod/lib/src/server/session.dart
```diff
@@ -341,6 +341,9 @@ class StreamingSession extends Session {
   /// Set if there is an open session log.
   int? sessionLogId;
 
+  /// The endpoint that is currently being processed.
+  String? endpoint;
```
SandPod, line 350: Question: Can we have this non-nullable? Or will that mess with the internal session?

packages/serverpod/lib/src/server/log_manager/log_manager.dart
```diff
@@ -20,13 +20,18 @@ class SessionLogManager {
 
   final LogSettings Function(Session) _settingsForSession;
 
+  int _logOrderId;
+
+  int get _nextLogOrderId => ++_logOrderId;
```
SandPod, line 25: Note(Non blocking): Commit message here is wrong right? It is moved into the log_manager and not the session?

SandPod (approved):
Couple of comments but giving the thumbs up so that you can merge them once it is fixed.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -95,7 +95,7 @@ abstract class Session {
   /// enabled but it will be disabled for internal sessions used by Serverpod.
   final bool enableLogging;
 
-  late final SessionLogManager _logManager;
+  late final SessionLogManager? _logManager;
```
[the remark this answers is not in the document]
Isakdl: Cannot remote the late since it is final and I have to do the setup in the body of the constructor, and we need the else because it is late.

packages/serverpod/lib/src/server/log_manager/log_manager.dart
Isakdl: probably, but I'm working on a new refactor making this assert redundant (removing the nullability of this variable).

packages/serverpod/lib/src/server/session.dart
Isakdl: we can, I have done so in a subsequent change

packages/serverpod/lib/src/server/log_manager/log_manager.dart
Isakdl: yes indeed

PR #2523, opened 2024-07-22
Don't sign GET or HEAD requests

### LINKED ISSUE #2517: `verifyDirectFileUpload` always returns `false` even when upload succeeds
I get an upload description and my client uploads a file to GCP Storage. If I look in the dev console, I can see that the file correctly uploaded to the right bucket.

However, when I use the same upload description and call `verifyDirectFileUpload`, the response is always `false`. There is no error message or exception.

I don't want to just ignore the return value -- what could cause this problem?

The code that checks this is:

```dart
  @override
  Future<bool> fileExists({
    required Session session,
    required String path,
  }) async {
    var response = await _s3Client.headObject(path);
    return response.statusCode == 200;
  }
```

so I assume a `false` response means that GCP is returning a status code other than 200. However this is the production server, so I can't just set a breakpoint here to check it.

It is quite urgent that I get this fixed, since I'm trying to push out my 1.0 release -- would appreciate any help with this. Thanks!

lukehutch:
Don't sign GET or HEAD requests to GCP Storage. Fixes #2517.

## Pre-launch Checklist

### DIFF
--- a/integrations/serverpod_cloud_storage_gcp/lib/src/aws_s3_client/client/client.dart
+++ b/integrations/serverpod_cloud_storage_gcp/lib/src/aws_s3_client/client/client.dart
@@ -2,12 +2,12 @@ import 'dart:convert';
 
 import 'package:amazon_cognito_identity_dart_2/sig_v4.dart';
 import 'package:built_value/serializer.dart';
-import 'exceptions.dart';
 import 'package:http/http.dart';
 import 'package:xml2json/xml2json.dart';
 
 import '../model/list_bucket_result.dart';
 import '../model/list_bucket_result_parker.dart';
+import 'exceptions.dart';
 
 class GCPS3Client {
   final String _secretKey;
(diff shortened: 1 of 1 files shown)


vlidholt:
I will check if this really is broken, as we have a client that I know for sure is currently using this function in production on GCP. I tried it out a few months ago and then everything was working correctly.

This PR removes a layer of security and will not work with the private file storage public. The private file storage allows users to upload files to a bucket that is only accessible from Serverpod and not exposed publicly on the web.

lukehutch:
Right, and obviously the signing needs to work or files cannot be deleted.

It's definitely broken on my deployment. What could break in the signing process?

vlidholt:
@lukehutch, are you sure you are using the right keys in your passwords file?

lukehutch:
@vlidholt Yes, as I said in #2517:

> I verified my `HMACAccessKeyId` and `HMACSecretKey` are correct. I even re-generated these, replaced them in passwords.yaml, updated the GitHub secrets, re-deployed my server, and yet this didn't fix the problem.

marcelomendoncasoares:
Closing now that the issue is solved with the refactor of the storage packages.

PR #2527, opened 2024-07-23
fix: Improve Method Stream Websocket Message data structure.

SandPod:
Another stepping stone towards #2338.

### Changes:
- ~~Adds a single digit version to websocket messages.~~
- ~~Add tests for this change.~~

~~This will allow us to update WebSocket messages in the future allowing us to possibly support multiple clients along with deprecating old ones.~~

### Update:

After the review it was concluded that we can probably do a version handshake as part of the WebSocket upgrade request and therefore we can skip versioning all together at this stage.

The PR now contains the grouping of the metadata for the messages.

## Pre-launch Checklist

## Breaking changes

None - these are changes related to a non-released feature.

### DIFF
--- a/packages/serverpod_serialization/lib/src/websocket_messages.dart
+++ b/packages/serverpod_serialization/lib/src/websocket_messages.dart
@@ -4,12 +4,6 @@ import 'package:serverpod_serialization/serverpod_serialization.dart';
 
 /// Base class for messages sent over a WebSocket connection.
 sealed class WebSocketMessage {
-  /// The version of the protocol.
-  static const int version = 1;
-
-  /// The keyword used for the version in the websocket message.
-  static const String messageVersionKeyword = 'version';
-
   /// The keyword used for the message type in the websocket message.
   static const String messageTypeKeyword = 'type';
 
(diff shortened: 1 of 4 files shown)


packages/serverpod_serialization/lib/src/websocket_messages.dart
```diff
@@ -4,6 +4,29 @@ import 'package:serverpod_serialization/serverpod_serialization.dart';
 
 /// Base class for messages sent over a WebSocket connection.
 sealed class WebSocketMessage {
+  /// The version of the protocol.
+  static const int version = 1;
+
+  /// The keyword used for the version in the websocket message.
+  static const String messageVersionKeyword = 'version';
```
vlidholt, line 11: Should this really be on a per-message level? The protocol version is not likely to change during a session. Perhaps it would be better with a first message setting the expectations for the protocol version?

Using this approach, it would be possible to add this feature at a later stage too, by assuming the version is the first version if there is no first versioning message.

SandPod: I removed the version from the messages and will only keep the improved formatting of the message and some test improvements.

vlidholt (approved):
LGTM

PR #2532, opened 2024-07-24
fix: Handle nullable stream parameters for Method Streams.

SandPod:
Another stepping stone towards #2338:

To preserve backwards compatibility for an endpoint, new parameters are added as optional. However, adding a new parameter as a stream parameter would break compatibility with existing clients by transforming the endpoint from a method call to a method stream.

To prevent this unexpected API-breaking change, we will not allow stream parameters to be nullable.

Since this issue is solvable (e.g., by allowing method calls to invoke streams without the optional stream parameter), we will add information to the OpenMethodStreamCommand listing the stream parameters the stream should open. This approach enables future support for nullable streams without breaking the current protocol.

### Change
- Adds the input streams that should be opened to the OpenMethodStreamCommand.
- Rejects open method stream requests that don't include all required input stream parameters.
- Prevents nullable streams from being used in endpoint methods.

## Pre-launch Checklist

## Breaking changes

None - touches a feature that has not yet been released.

### DIFF
--- a/packages/serverpod/lib/src/server/endpoint_dispatch.dart
+++ b/packages/serverpod/lib/src/server/endpoint_dispatch.dart
@@ -219,6 +219,27 @@ abstract class EndpointDispatch {
 
     return deserializedParams;
   }
+
+  /// Parses a list of requested input stream parameter descriptions and returns
+  /// a list of stream parameter descriptions.
+  ///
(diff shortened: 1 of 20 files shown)


vlidholt (approved):
LGTM 🥳

Issue #2534, opened 2024-07-24
Dynamic web server content is being cached by CDN


lukehutch:
When I deploy a new version of my server to GCP, the web page content dynamically served by Relic does not change. I logged into the VM instance, created a shell in the Docker container, and verified that the new content is there, but it is not showing in my browser. I have waited over 12 hours for the changes to show, and they have not. Therefore, I assume that these dynamic pages are being cached by a CDN with infinite (or large) TTL.

I am serving only static content, but I wrapped every static page in a `Route` so that it would be served dynamically, because of [this note](https://docs.serverpod.dev/concepts/webserver#special-widgets-and-routes) in the docs:

> Static files are configured to be cached hard by the web browser and through Cloudfront's content delivery network (if you use the AWS deployment). If you change static files, they will need to be renamed, or users will most likely access old files. To make this easier, you can add a version number when referencing the static files. The version number will be ignored when looking up the actual file. E.g., /static/my_image@v42.png will serve to the /static/my_image.png file. More advanced cache management will be coming to a future version of Serverpod.

This doesn't say what the CDN policy is on GCP, but I would not expect dynamic pages to be cached by a CDN.

lukehutch:
Here is the CDN policy:

![image](https://github.com/user-attachments/assets/eec00977-bdd0-4238-9f8c-d7f96f0b7967)

How does the CDN determine what is static and what is dynamic content?

vlidholt:
This is most likely depending on your headers on the files you are requesting.

lukehutch:
Yes, that's exactly my point -- Relic needs to return no-cache headers for dynamic routes.

lukehutch:
#2539 allows static content whose path matches a regexp to not be cached. However, the same header change needs to be made for all dynamic content.

SandPod:
Dynamic routes should not be cached, when quickly inspecting the code it does not look like we set any headers for caching at all.

We might need to add a default no-cache header for the dynamic routes.

PR #2539, opened 2024-07-26
Allow static content to not be cached, by path regexp

lukehutch:
Allow static paths to be excluded from CDN caching, by regexp match.

Partially addresses #2534, although for that to be fixed, dynamic responses should also include these same cache denial headers. (I didn't make that change in this PR, but it needs to be done...)

## Pre-launch Checklist

### DIFF
--- a/packages/serverpod/lib/src/relic/routes/static_directory.dart
+++ b/packages/serverpod/lib/src/relic/routes/static_directory.dart
@@ -27,8 +27,23 @@ class RouteStaticDirectory extends Route {
   /// The path to the directory to serve static files from.
   final String? basePath;
 
+  /// A regular expression that will be used to determine if a path should not
+  /// be cached.
+  late final RegExp? noCachePathRegexp;
+
   /// Creates a static directory with the [serverDirectory] as its root.
-  RouteStaticDirectory({required this.serverDirectory, this.basePath});
+  /// If [basePath] is provided, the directory will be served from that path.
+  /// If [noCachePathPattern] is provided, paths matching the provided pattern
+  /// will not be cached. (Otherwise, all paths will be cached for a year.)
+  RouteStaticDirectory({
+    required this.serverDirectory,
+    this.basePath,
+    String? noCachePathPattern,
+  }) {
+    // Pre-compile the regexp pattern, if provided
+    final regexpPath = noCachePathPattern;
+    noCachePathRegexp = regexpPath == null ? null : RegExp(regexpPath);
+  }
 
   @override
   Future<bool> handleCall(Session session, HttpRequest request) async {
@@ -69,7 +84,12 @@ class RouteStaticDirectory extends Route {
       }
 
       // Enforce strong cache control.
-      request.response.headers.set('Cache-Control', 'max-age=31536000');
+      if (noCachePathRegexp != null && noCachePathRegexp!.hasMatch(path)) {
+        request.response.headers
+            .set('Cache-Control', 'max-age=0, s-maxage=0, no-cache, no-store');
+      } else {
+        request.response.headers.set('Cache-Control', 'max-age=31536000');
+      }
 
       var filePath = path.startsWith('/') ? path.substring(1) : path;
       filePath = 'web/$filePath';
--- a/packages/serverpod/lib/src/relic/routes/static_directory.dart
+++ b/packages/serverpod/lib/src/relic/routes/static_directory.dart
@@ -20,11 +20,17 @@ final _contentTypeMapping = <String, ContentType>{
   '.pdf': ContentType('application', 'pdf'),
 };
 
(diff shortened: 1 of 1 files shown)


lukehutch:
See also https://github.com/serverpod/terraform-google-serverpod-cloud-engine/pull/11

packages/serverpod/lib/src/relic/routes/static_directory.dart
```diff
@@ -27,8 +27,23 @@ class RouteStaticDirectory extends Route {
...
+
   /// Creates a static directory with the [serverDirectory] as its root.
-  RouteStaticDirectory({required this.serverDirectory, this.basePath});
+  /// If [basePath] is provided, the directory will be served from that path.
+  /// If [noCachePathPattern] is provided, paths matching the provided pattern
+  /// will not be cached. (Otherwise, all paths will be cached for a year.)
+  RouteStaticDirectory({
+    required this.serverDirectory,
+    this.basePath,
+    String? noCachePathPattern,
```
Isakdl, line 41: Suggestion: Modify this to take the RegExp directly, I think this makes the interface more clear what we expect.

It may also be a good idea to take a list rather than a single value as that would make it easier to opt out of caching for different subroutes.

e.g.

```
[
  RegExp('images/*'),
  RegExp('hello.html'),
]

File: packages/serverpod/lib/src/relic/routes/static_directory.dart [hunk 2]
```diff
@@ -27,8 +27,23 @@ class RouteStaticDirectory extends Route {
   /// The path to the directory to serve static files from.
   final String? basePath;
 
+  /// A regular expression that will be used to determine if a path should not
+  /// be cached.
+  late final RegExp? noCachePathRegexp;
```
Isakdl, line 32: Suggestion: I think we can keep this private I see no value in exposing this variable.

Isakdl (changes requested):
Thank you for the PR @lukehutch!

I have some smaller suggestions, but otherwise happy with this change :).

packages/serverpod/lib/src/relic/routes/static_directory.dart
lukehutch: Done

packages/serverpod/lib/src/relic/routes/static_directory.dart
lukehutch: Done

lukehutch:
Actually I just made a change to allow the max-age to be specified on a per-pattern basis. PTAL.

packages/serverpod/lib/src/relic/routes/static_directory.dart
```diff
@@ -19,6 +20,31 @@ final _contentTypeMapping = <String, ContentType>{
   '.pdf': ContentType('application', 'pdf'),
 };
 
+/// A path pattern to match, and the max age that paths that match the pattern
+/// should be cached for, in seconds.
+class PathCacheMaxAge {
+  final Pattern pathPattern;
```
Isakdl, line 26: TIL: Cool solution!

packages/serverpod/lib/src/relic/routes/static_directory.dart
```diff
@@ -19,6 +20,31 @@ final _contentTypeMapping = <String, ContentType>{
...
+  PathCacheMaxAge({
+    required this.pathPattern,
+    required this.maxAge,
+  });
+
+  bool _shouldCache(String path) => pathPattern is String
+      ? path == pathPattern
+      : pathPattern is RegExp
+          ? (pathPattern as RegExp).hasMatch(path)
+          : false;
```
Isakdl, line 45: NIT: Change this to a multiline method instead as that is much easier to read when having nested cases like this. And extracting the variable into a local one lets us avoid typecasting making safer code!

```suggestion
bool _shouldCache(String path) {
  var pattern = pathPattern;

  if (pattern is String) {
    return path == pattern;
  } else if (pattern is RegExp) {
    return pattern.hasMatch(path);
  }

  return false;
}
```

Isakdl (changes requested):
Perfect change, looking good but have another NIT thing I would like you to change.

Also the linter is failing for some missing docs comments.

packages/serverpod/lib/src/relic/routes/static_directory.dart
lukehutch: Please go ahead and apply your suggested changes, I'm swamped right now! Thanks.

lukehutch:
Please go ahead and make whatever changes toy seem necessary... Also the docs need to be updated for this API change and the one you just merged. Thanks!

lukehutch:
Please make whatever changes you deem necessary. Also this needs to be rebased on the change you just merged, there's a conflict.

PR #2546, opened 2024-07-28
Copy fields of ServerException across to DatabaseException

lukehutch:
All of the context of a `PgException` (specifically, the context in its subclass, `ServerException`) is lost when only the message is preserved in a `DatabaseException`. This means that the user cannot be notified, for example, of the character offset of a syntax error, which can make debugging query errors very difficult (you basically have to print out the query on the server console, copy/paste it into a proper Postgres client UI, and see where the error is highlighted).

This change copies all the fields of `ServerException` across to `DatabaseException` in a backwards-compatible way.

I will leave it up to you to add some of this extra info to the error logs, so that Serverpod Insights can report more info on database errors.

## Pre-launch Checklist


SandPod:
@lukehutch - Would it be OK if I take this PR over and get it in?

lukehutch:
@SandPod please do

SandPod:
Sorry, the close here was not intentional. An incorrect reference to this PR accidentally closed the issue.

SandPod:
This feature was implemented in: https://github.com/serverpod/serverpod/pull/3047

PR #2552, opened 2024-07-30
refactor: Move currentMessageId into the base session.

Isakdl:
# Changes

Moves the currentMessageId from StreamingSession to the base session.
Fixes a bug where the messageId was not recorded for logged database queries.

Motivation for this change is that we want a generic way to deal with the message id rather than having to cast types as this functionality needs to be supported in more than one session.

There is an argument to be made that this id could be handled in the logger instead and have a similar logic to the session logs life cycle. Meaning each message would require an "open" and a "close" method. But this would make this much more complex. Reason for this is that we are writing the logs before we write the message log. Message log being the message received on the stream. The reason for this order is that the message log also records the execution time of the callback. 

Related issue: https://github.com/serverpod/serverpod/issues/2450

## Pre-launch Checklist

## Breaking changes

none


SandPod (approved):
LGTM! 👍

PR #2556, opened 2024-07-30
refactor: Log manager simplified to allow support for additional sessions.

Isakdl:
# Changes

Depends on: https://github.com/serverpod/serverpod/pull/2547, https://github.com/serverpod/serverpod/pull/2552, https://github.com/serverpod/serverpod/pull/2553, https://github.com/serverpod/serverpod/pull/2554

Modifies the implementation of the log manager and log writers to remove branching logic and simplify the control flow. The implementation is no longer tied to specific session types and can therefore work with any sessions and in any potential configuration (caching, database, stdout, etc).

This will also allow the new streaming interface to easily be integrated with the logger.

Review notes:
Review commit by commit, but dependent PRs should be merged / rebased first.

Closes: https://github.com/serverpod/serverpod/issues/2450

## Pre-launch Checklist

## Breaking changes
none


Isakdl:
Comment: Remove the messageId from `MethodStreamSession` as we will not log these for now.

packages/serverpod/lib/src/server/log_manager/log_writers.dart
```diff
@@ -0,0 +1,215 @@
...
+
+    var id = result.id;
+
+    if (id == null) throw StateError('Failed to insert log entry');
+    return id;
+  }
+}
+
+@internal
+class StdOutLogWriter extends LogWriter {
```
SandPod, line 175: Question (for my knowledge): These are all `stdout` writes. Where do we log errors?

packages/serverpod/lib/src/server/log_manager/log_manager.dart
```diff
@@ -316,7 +323,7 @@ class SessionLogManager {
     String? exception,
     StackTrace? stackTrace,
   }) async {
-    await _openStreamLogLock.synchronized(() {});
+    if (!_isLoggingOpened) await _openStreamLogLock.synchronized(() {});
```
SandPod: Question: For me this looks like we only wait on the lock if logging is not opened. But if the lock is never acquired we will just move past this line.

What is this guarding against?

packages/serverpod/lib/src/server/log_manager/log_manager.dart
```diff
@@ -271,6 +274,8 @@ class SessionLogManager {
     required Session session,
   }) async {
     await _openStreamLogLock.synchronized(() async {
+      if (_isLoggingOpened) return;
+
```
SandPod: Suggestion: Remove internal check.

Since we guard against this on the callsite on line 251 this should not be required right?

packages/serverpod/lib/src/server/log_manager/log_manager.dart
```diff
@@ -331,13 +331,9 @@ class SessionLogManager {
...
-      if (session is MethodCallSession) {
-        stdout.writeln(
-            'METHOD CALL: ${session.endpointName}.${session.methodName} duration: ${duration.inMilliseconds}ms numQueries: $_numberOfQueries authenticatedUser: $authenticatedUserId');
-      } else if (session is FutureCallSession) {
-        stdout.writeln(
-            'FUTURE CALL: ${session.futureCallName} duration: ${duration.inMilliseconds}ms numQueries: $_numberOfQueries');
-      }
+      stdout.writeln(
+        'CALL: ${session.callName} duration: ${duration.inMilliseconds}ms numQueries: $_numberOfQueries authenticatedUser: $authenticatedUserId',
+      );
```
SandPod, line 292: Question: Isn't this the same as enabling the `stdout` logger?

packages/serverpod/lib/src/server/session.dart
```diff
@@ -146,6 +148,22 @@ abstract class Session {
...
 
+  LogWriter _createLogWriter(Session session, LogSettingsManager settings) {
+    var logSettings = settings.getLogSettingsForSession(session);
+
+    LogWriter logWriter = switch (Features.enablePersistentLogging) {
+      (true) => DatabaseLogWriter(session),
+      (false) => StdOutLogWriter(session),
+    };
+
+    if ((session is StreamingSession || session is MethodStreamSession) &&
```
SandPod, line 152: Question: Is this check duplicated in other places? Meaning certain attributes are given to StreamingSession based on their class names? If so, could we centralize it to a function?

packages/serverpod/lib/src/server/session.dart
```diff
@@ -578,4 +578,9 @@ extension SessionInternalMethods on Session {
 
     return id;
   }
+
+  /// Returns true if the session is expected to be alive for an extended
+  /// period of time.
+  bool get isLongLived =>
+      this is StreamingSession || this is MethodStreamSession;
 }
```
SandPod, line 586: Suggestion: Use this as the check for the previous question I asked.

Here is an example of another attribute given to StreamingSessions, we should probably consolidate all of these so that another streaming session can be added by just modifying one place.

SandPod (commented):
Awesome work, this is shaping up to become some really elegant code with a clear API.

I have a couple of questions for you and possibly a suggestion to two.

packages/serverpod/lib/src/server/log_manager/log_writers.dart
```diff
@@ -0,0 +1,215 @@
...
+    return id;
+  }
+
+  Future<void> _databaseLogUpdate<T extends TableRow>(
+    Session session,
+    T entry,
+  ) async {
+    var tempSession = await session.serverpod.createSession(
+      enableLogging: false,
+    );
```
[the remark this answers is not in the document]
Isakdl: Good point! But I think we can deal with this in another PR or this one will get bloated.

packages/serverpod/lib/src/server/log_manager/log_writers.dart
Isakdl: Ahh true, the objects that we are logging contains the error if there is one, Here I think we can branch inside the log writer and write to stderr if the logged object contains an error.

packages/serverpod/lib/src/server/log_manager/log_manager.dart
Isakdl: The internal check is vital as `n` number of methods can acquire the lock but only the first should execute the init logic. So the other methods have to eject if the variable was set.

packages/serverpod/lib/src/server/log_manager/log_manager.dart
Isakdl: Not really, these logs are more internal server logs, this PR deals with the session logs. I think we should clean up the server logging in the future too but that is out of scope for this PR.

SandPod (approved):
LGTM! 👍

Issue #2580, opened 2024-08-08
Add stream interface to message central.


SandPod:
**Is your feature request related to a problem? Please describe.**
With the new streaming interface a common use case will be to listen from messages in the internal messaging and relaying them to the client.

```dart
Stream<DeckState> deckUpdates(Session session) async* {
  // Set up a stream for getting notified when we get messages from
  // MessageCentral.
  var streamController = StreamController<DeckState>();
  session.messages.addListener(kChannelName, (state) {
    streamController.add(state as DeckState);
  });

  // Listen for updates and relay them to the client.
  try {
    await for (var state in streamController.stream) {
      yield state;
    }
  } catch (e) {
    // It's possible we never get here.
    streamController.close();
  }
}
```
The problem here is two-fold. First, it requires a bit of knowledge to set up the stream. Second, it's hard to exit the method as we won't get notified when the client is no longer connected.

**Describe the solution you'd like**

A possible solution is to add a stream interface to MessageCentral. Instead of adding a listener, it would be possible to request a stream. The stream is automatically closed (similar to how the current listeners are disposed) when the client disconnects. With the new interface, the above code could be written as:

```dart
Stream<DeckState> deckUpdates(Session session) async* {
  // Listen for changes and relay them to the client.
  var stream = session.messages.createStream<DeckState>(kChannelName);
  await for (var state in stream) {
    yield state;
  }
}
```

The `createStream` method create a stream that listens to updates on a specific channel name in the messages central and relays them to the stream.

PR #2584, opened 2024-08-09
feat: Add possibility to create a stream subscription to message central.

SandPod:
Closes: #2580 

Makes it possible to create a stream that listens to a channel in the message central.

Such as:

```dart
Stream<DeckState> deckUpdates(Session session) async* {
  // Listen for changes and relay them to the client.
  var stream = session.messages.createStream<DeckState>(kChannelName);
  await for (var state in stream) {
    yield state;
  }
}
```

## Pre-launch Checklist

## Breaking changes

None: introduces a new way to subscribe to messages in the message central.

### DIFF
--- a/packages/serverpod/lib/src/server/message_central.dart
+++ b/packages/serverpod/lib/src/server/message_central.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:serverpod/serverpod.dart';
 
 // TODO: Support for server clusters.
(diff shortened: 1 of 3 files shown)


packages/serverpod/lib/src/server/message_central.dart
```diff
@@ -168,4 +170,33 @@ class MessageCentral {
...
+    void addToStream(dynamic message) {
+      try {
+        controller.add(message as T);
+      } catch (e) {
+        controller.addError(e);
+      }
+    }
+
+    addListener(session, channelName, addToStream);
+
```
SandPod, line 195: To investigate: Need to validate that the stream is closed if the listener is removed from some other source.

vlidholt (approved):
LGTM 🥳

Issue #2581, opened 2024-08-08
Possibility to register callback that is run before session is closed.


SandPod:
**Is your feature request related to a problem? Please describe.**
With the new streaming interface, additional cleanups might be needed when the client disconnects but before the session is closed.

**Describe the solution you'd like**
Add two methods to the session, `addWillCloseListener` and `removeWillCloseListener,` that can be used to add or remove callbacks that are executed before the session is closed.

```dart
    session.addWillCloseListener((Session session) => // Some calculation....);
    session.removeWillCloseListener((Session session) => // Some calculation....);
```

These methods take the session as a parameter and the order of the registration should be respected when the callbacks are executed.

Here is an example where we add a log if the method is closed before the input stream has been completed.
```dart
  Stream<int> intEchoStream(Session session, Stream<int> stream) async* {
    var closeLog = (Session session) {
      session.log('Stream closed before completion');
    };
    session.addWillCloseListener(closeLog);

    await for (var value in stream) {
      yield value;
    }

    session.removeWillCloseListener(closeLog);
  }
```

PR #2586, opened 2024-08-09
feat: Add support for willClose listeners on Session.

SandPod:
Closes: #2581 

Adds support for registering listeners that are called when the session is about to close.

This could become handy with the new streaming interface as such:

```dart
  Stream<int> intEchoStream(Session session, Stream<int> stream) async* {
    var closeLog = (Session session) {
      session.log('Stream closed before completion');
    };
    session.addWillCloseListener(closeLog);

    await for (var value in stream) {
      yield value;
    }

    session.removeWillCloseListener(closeLog);
  }
```

## Pre-launch Checklist

## Breaking changes

None - new feature on the streaming interface.

### DIFF
--- a/packages/serverpod/lib/src/server/session.dart
+++ b/packages/serverpod/lib/src/server/session.dart
@@ -14,13 +14,12 @@ import '../cache/caches.dart';
 import '../database/database.dart';
 
 /// A listener that will be called when the session is about to close.
-typedef WillCloseListener = FutureOr<void> Function(Session);
+typedef WillCloseListener = FutureOr<void> Function(Session session);
 
 /// When a call is made to the [Server] a [Session] object is created. It
 /// contains all data associated with the current connection and provides
 /// easy access to the database.
 abstract class Session {
-  /// List of listeners that will be called when the session is about to close.
   final LinkedHashSet<WillCloseListener> _willCloseListeners = LinkedHashSet();
 
   /// Adds a listener that will be called when the session is about to close.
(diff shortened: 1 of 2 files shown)


packages/serverpod/lib/src/server/session.dart
```diff
@@ -12,10 +13,30 @@ import 'package:serverpod/src/server/serverpod.dart';
 import '../cache/caches.dart';
 import '../database/database.dart';
 
+/// A listener that will be called when the session is about to close.
+typedef WillCloseListener = FutureOr<void> Function(Session);
```
vlidholt, line 17: I think you can add parameter names to the prototype here, e.g. `Function(Session session)`. This will aid the editor to give a better autocompletions.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -12,10 +13,30 @@ import 'package:serverpod/src/server/serverpod.dart';
...
+
 /// When a call is made to the [Server] a [Session] object is created. It
 /// contains all data associated with the current connection and provides
 /// easy access to the database.
 abstract class Session {
+  /// List of listeners that will be called when the session is about to close.
```
vlidholt, line 23: Nit. Technically, the `///` have no effect for private methods.

vlidholt (approved):
LGTM, but see comments. 🙂

packages/serverpod/lib/src/server/session.dart
SandPod: True, I can remove this! 👍

Issue #2380, opened 2024-06-14
WASM support, dart:html package is deprecated and should be replaced by package:web


Isakdl:
The package `dart:html` does not support wasm compilation target. This package is in use in a limited number of places in the code but currently blocks support for wasm in a flutter app.

To fix this, the `dart:html` package should be replaced with `package:web`

source: https://dart.dev/libraries/dart-html

SandPod:
It's being worked on by @abdelaziz-mahdy.

PR #2596, opened 2024-08-12
fix: Migrate from dart:html to package:web for WASM support.

abdelaziz-mahdy:
fixes https://github.com/orgs/serverpod/projects/4?pane=issue&itemId=70600442
i didnt test the code, i just migrated to the wasm compatible logic and libs 

Closes: #2380

## Pre-launch Checklist

## Breaking changes


SandPod:
Thanks for taking this on @abdelaziz-mahdy!

There are still uses of `dart:html` that need to be fixed in this file: packages/serverpod_client/lib/src/http/http_status_html.dart.

It seems like the HTML status is not exported from the `package:web` package so we might need to define these manually. You can do this by replacing the conditional import of `http_status_html` and `http_status_io` and define our own solution. This should also be used in the tests.

tests/serverpod_test_client/lib/src/http/http_status_io.dart
```diff
@@ -0,0 +1 @@
+export 'dart:io' show HttpStatus;
```
SandPod, line 1: Suggestion: Only use our definition of `HttpStatus`.

Since we now have our own definition of `HttpStatus` we no longer need to import it from `dart:io`. Then we can remove all the conditional import statements and just have a single definition for it.

tests/serverpod_test_client/lib/src/http/http_status_html.dart
```diff
@@ -0,0 +1,89 @@
+/**
```
SandPod, line 1: Suggestion: Remove this definition.

Is it possible to reuse the definition from the client library instead?

packages/serverpod_client/lib/src/http/http_status_html.dart
```diff
@@ -1 +1,89 @@
-export 'dart:html' show HttpStatus;
+/**
+ * HTTP status codes.  Extracted from dart:io since it does not exist on package:web/web.dart .
+ */
```
SandPod, line 3: Suggestion: Add copyright and license notice.

Please add a copyright and license notice if one is required for this code.

SandPod (changes requested):
Almost there!

The reason the update pubspec is failing is because the pubspec templates have not been updated.

Please update these under `templates/pubspecs`.

abdelaziz-mahdy:
all fixed, only remaining to test the module functionally since i didnt see tests for it (chat and auth ones)

SandPod:
Please fix the last formatting issue also, then we will run the validation on this and try to get it in ASAP.

abdelaziz-mahdy:
Done.

SandPod:
Did a manual test of the google auth workflow based on this branch and everything seems to work as expected:


https://github.com/user-attachments/assets/e389f3dd-e738-41e2-99de-ddee4beaa1bd

SandPod (approved):
Awesome work, thanks for fixing this! 🙌

PR #2597, opened 2024-08-13
fix: Only consider positional `Session` parameter when validating Endpoint method.

### LINKED ISSUE #2557: serverpod generate  creates error and, doesnt create Protocol and Endpoints class
**Describe the bug**

I m running serverpod generate command after modifying my endpoinst and I start to get the error below. I tried to figure it out which can cause this problem but could not success. Maybe its not a bug but I m not sure what is the problem and what cause the problem. 

I upgraded to 2.0.2  but no luck.

Any idea will be appriciated. thanks 

here is the error what I get 

```
Yikes! It is possible that this error is caused by an internal issue with the Serverpod tooling. We would appreciate if you
filed an issue over at Github. Please include the stack trace below and describe any steps you did to trigger the error.
https://github.com/serverpod/serverpod/issues

RangeError (start): Invalid value: Only valid value is 0: 1

#0      RangeError.checkValidRange (dart:core/errors.dart:359:7)
#1      List.sublist (dart:core-patch/growable_array.dart:84:38)
#2      EndpointMethodAnalyzer.parse
(package:serverpod_cli/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart:30:39)
#3      EndpointsAnalyzer._parseLibrary (package:serverpod_cli/src/analyzer/dart/endpoints_analyzer.dart:141:47)
#4      EndpointsAnalyzer.analyze (package:serverpod_cli/src/analyzer/dart/endpoints_analyzer.dart:83:27)
<asynchronous suspension>
#5      performGenerate (package:serverpod_cli/src/generator/generator.dart:53:19)
<asynchronous suspension>
#6      StdOutLogger.progress (package:cli_tools/src/logger/loggers/std_out_logger.dart:141:20)
<asynchronous suspension>
#7      GenerateCommand.run (package:serverpod_cli/src/commands/generate.dart:67:17)
<asynchronous suspension>
#8      CommandRunner.runCommand (package:args/command_runner.dart:212:13)
<asynchronous suspension>
#9      BetterCommandRunner.runCommand (package:cli_tools/src/better_command_runner/better_command_runner.dart:140:7)
<asynchronous suspension>
#10     _main (file:///Users/bilalsimsek/.pub-cache/hosted/pub.dev/serverpod_cli-2.0.2/bin/serverpod_cli.dart:59:3)
<asynchronous suspension>
#11     main.<anonymous closure>
(file:///Users/bilalsimsek/.pub-cache/hosted/pub.dev/serverpod_cli-2.0.2/bin/serverpod_cli.dart:33:9)
<asynchronous suspension>
#12     main (file:///Users/bilalsimsek/.pub-cache/hosted/pub.dev/serverpod_cli-2.0.2/bin/serverpod_cli.dart:30:3)
<asynchronous suspension>
```

**To Reproduce**

**Expected behavior**

**Screenshots**

**Serverpod versions**
 -  _CLI Version:_ Run `serverpod version` to determine the version of installed tooling.
 - _Library version:_ Check the pubspec.yaml file for the version of Serverpod you are using.

**Platform information**
On which platforms do you experience the issue. Please provide as much information here as you can. Where are you running the server, which front end device are you using?

**Additional context**

hampuslavin:
## Description
Fixes #2557, where a public method was incorrectly interpreted as an Endpoint method. This causes the following two `sublist` calls to throw a `RangeError`:
```dart
// tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart

if (isStream) {
      return MethodStreamDefinition(
       // ...
        parameters: parameters.required.sublist(1), // Skip session parameter,
      // ...
      );
    }
return MethodCallDefinition(
    // ...
      parameters: parameters.required.sublist(1), // Skip session parameter,
    // ...
);

Additionally, for the case of a nullable `Session` (`Future<void> hello(Session? session) async {}`), there is now a hint message, see below. 
```

## Changes
- Adds a test case according to the filed bug (#2557)
- Adds an additional check for named parameter in the `EndpointMethodAnalyzer._missingSessionParameter` method. If the first parameter is a `Session` but is named, it shouldn't be interpreted as an Endpoint method.
- Adds a hint message for nullable `Session` argument in a valid Endpoint method.
image

## Pre-launch Checklist

### DIFF
--- a/tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
@@ -6,6 +6,7 @@ import 'package:serverpod_cli/src/analyzer/dart/endpoint_analyzers/endpoint_para
 import 'package:serverpod_cli/src/analyzer/dart/definitions.dart';
 import 'package:serverpod_cli/src/analyzer/dart/element_extensions.dart';
 import 'package:serverpod_cli/src/generator/types.dart';
+import 'package:serverpod_cli/src/analyzer/dart/endpoint_analyzers/keywords.dart';
 
 const _excludedMethodNameSet = {
   'streamOpened',
@@ -92,7 +93,7 @@ abstract class EndpointMethodAnalyzer {
     if (parameters.isEmpty) return true;
 
     bool firstParameterIsNotSession =
-        parameters.first.type.element?.displayName != 'Session';
+        parameters.first.type.element?.displayName != Keyword.sessionClassName;
     if (firstParameterIsNotSession) return true;
 
     if (parameters.first.isNamed) return true;
--- a/tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
@@ -90,7 +90,12 @@ abstract class EndpointMethodAnalyzer {
 
   static bool _missingSessionParameter(List<ParameterElement> parameters) {
     if (parameters.isEmpty) return true;
-    return parameters.first.type.element?.displayName != 'Session';
+
+    bool firstParameterIsNotSession =
+        parameters.first.type.element?.displayName != 'Session';
+    if (firstParameterIsNotSession) return true;
+
+    return parameters.first.isNamed;
   }
 
   static SourceSpanSeverityException? _validateReturnType({
--- a/tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
@@ -94,11 +94,10 @@ abstract class EndpointMethodAnalyzer {
 
     bool firstParameterIsNotSession =
         parameters.first.type.element?.displayName != Keyword.sessionClassName;
-    if (firstParameterIsNotSession) return true;
 
-    if (parameters.first.isNamed) return true;
-
-    return parameters.first.isOptional;
+    return firstParameterIsNotSession ||
+        parameters.first.isNamed ||
+        parameters.first.isOptional;
   }
 
   static SourceSpanSeverityException? _validateReturnType({
--- a/tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
@@ -95,7 +95,9 @@ abstract class EndpointMethodAnalyzer {
         parameters.first.type.element?.displayName != 'Session';
     if (firstParameterIsNotSession) return true;
 
-    return parameters.first.isNamed;
+    if (parameters.first.isNamed) return true;
+
+    return parameters.first.isOptional;
   }
 
   static SourceSpanSeverityException? _validateReturnType({
--- a/tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_parameter_analyzer.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_parameter_analyzer.dart
@@ -54,7 +54,7 @@ abstract class EndpointParameterAnalyzer {
           parameters.first.type.nullabilitySuffix != NullabilitySuffix.none;
       if (sessionIsNullable) {
         exceptions.add(SourceSpanSeverityException(
-            'The "Session" argument in an endpoint method does not have to be nullable, consider removing the "?".',
+            'The "Session" argument in an endpoint method does not have to be nullable, consider making it non-nullable.',
             parameters.first.span,
             severity: SourceSpanSeverity.hint));
       }
--- a/tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_parameter_analyzer.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_parameter_analyzer.dart
@@ -4,6 +4,7 @@ import 'package:analyzer/dart/element/type.dart';
 import 'package:serverpod_cli/src/analyzer/code_analysis_collector.dart';
 import 'package:serverpod_cli/src/analyzer/dart/definitions.dart';
 import 'package:serverpod_cli/src/analyzer/dart/element_extensions.dart';
+import 'package:serverpod_cli/src/analyzer/dart/endpoint_analyzers/keywords.dart';
 import 'package:serverpod_cli/src/generator/types.dart';
 
 abstract class EndpointParameterAnalyzer {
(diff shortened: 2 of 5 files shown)


tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart
```diff
@@ -120,17 +120,21 @@ class ExampleEndpoint extends Endpoint {
     });
   });
 
-  group('Given an endpoint method without Session param when analyzed', () {
+  group('Given an endpoint method without a first unnamed `Session` param', () {
```
SandPod, line 123: Suggestion(NIT): Change to `first positional param`.

In dart, the unnamed parameters are reffered to as positional parameters: https://dart.dev/language/functions#parameters

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart
```diff
@@ -120,17 +120,21 @@ class ExampleEndpoint extends Endpoint {
...
 
     late List<EndpointDefinition> endpointDefinitions;
     late EndpointsAnalyzer analyzer;
-    setUpAll(() async {
-      var endpointFile = File(path.join(testDirectory.path, 'endpoint.dart'));
-      endpointFile.createSync(recursive: true);
-      endpointFile.writeAsStringSync('''
+
+    group(
+        'And the other parameters are not a `Session` parameter when analyzed',
```
SandPod, line 132: Suggestion(NIT): Change `And the other` to `and the other` (and `parameters are not` to `parameter is not`).

All groups and tests are combined to build the test description. A capital letter here would produce the scentence: `Given an endpoint method without a first unnamed "Session" param And the other parameters are not a "Session" parameter when analyzed...`

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart
```diff
@@ -139,21 +143,53 @@ class ExampleEndpoint extends Endpoint {
...
+        var methods = endpointDefinitions.firstOrNull?.methods;
+        expect(methods, isEmpty);
+      });
     });
 
-    test('then no endpoint method definition is created.', () {
-      var methods = endpointDefinitions.firstOrNull?.methods;
-      expect(methods, isEmpty);
+    group(
+        'And the first parameter instead contains a named `Session` parameter when analyzed',
```
SandPod, line 164: Suggestion(NIT): Remove capital A and validate that the sentence makes sense when combined with the outer group.

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart
SandPod, line 123: Suggestion: Split to two groups.

Since the setup is different for the tests, we might as well have two groups defined for them. I understand the thinking here, but when the "shared" part is not defined together for the tests in the group, it could be a bit difficult to find where it is defined.

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
```diff
@@ -90,7 +90,12 @@ abstract class EndpointMethodAnalyzer {
 
   static bool _missingSessionParameter(List<ParameterElement> parameters) {
     if (parameters.isEmpty) return true;
-    return parameters.first.type.element?.displayName != 'Session';
+
+    bool firstParameterIsNotSession =
+        parameters.first.type.element?.displayName != 'Session';
+    if (firstParameterIsNotSession) return true;
+
+    return parameters.first.isNamed;
```
SandPod, line 98: Question: What happens if the session parameter is Optional? That is:
```dart
Future<int> myMethod([Session session]);
```

Maybe we should add a test case for this and possibly modify the logic here to handle that case?

SandPod (changes requested):
Nice work with the tests!

I'm nitty in the beginning to make sure we have the same mindset when it comes to testing :)

Please review the question I had and ping me once it is answered.

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart
hampuslavin: Great, thanks! 🙌

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart
hampuslavin: Ah, good catch! 👍

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
hampuslavin: Nice catch!

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart
hampuslavin: Regarding
>and parameters are not to parameter is not

Since it should be true for any number of parameters, I think "and the other parameters" would be the most accurate test goal. The fact that it's a single parameter in the input data is just a subset of that statement, but generally it would work for any number.

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
hampuslavin: Indeed this also cause the bug (Fixed in [b8e60b5](https://github.com/serverpod/serverpod/pull/2597/commits/b8e60b57c254ba84ba88de20fd2912a4de0eea04))💡. As a follow-up question then, the following is currently interpreted as a valid endpoint definition (note the `?`). Technically it will always work, but I wonder if this should result in an `info` or `warning` message from the generator script?
```dart
  Future<void> hello(Session? session) async {
    return null;
  }
```

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart
hampuslavin: [7c32c17](https://github.com/serverpod/serverpod/pull/2597/commits/7c32c179843f7b45b4b4f6b77cb51d76e26568bc)

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart
hampuslavin: Fixed casing: [7c32c17](https://github.com/serverpod/serverpod/pull/2597/commits/7c32c179843f7b45b4b4f6b77cb51d76e26568bc)

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart
hampuslavin: [7c32c17](https://github.com/serverpod/serverpod/pull/2597/commits/7c32c179843f7b45b4b4f6b77cb51d76e26568bc)

hampuslavin: fixed: 
b4d9149

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart
hampuslavin: Added additional param to match test description: 
7c30ba5

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_parameter_analyzer.dart
```diff
@@ -43,54 +44,69 @@ abstract class EndpointParameterAnalyzer {
...
+
+    bool firstParameterIsSession = parameters.isNotEmpty &&
+        parameters.first.type.element?.displayName == Keyword.sessionClassName;
+
+    if (firstParameterIsSession) {
+      bool sessionIsNullable =
+          parameters.first.type.nullabilitySuffix != NullabilitySuffix.none;
+      if (sessionIsNullable) {
+        exceptions.add(SourceSpanSeverityException(
+            'The "Session" argument in an endpoint method does not have to be nullable, consider removing the "?".',
```
SandPod, line 57: Suggestion: Change to your suggested proposal of `consider making it non-nullable`

hampuslavin: [dce77e1](https://github.com/serverpod/serverpod/pull/2597/commits/dce77e12dce032b4967912fada3137685b0d90fe)

PR #2602, opened 2024-08-13
chore: Update analyze command to allow only giving warnings on new major versions.

SandPod:
Updates the hidden helper command to allow us to check if there are any new major versions out for any of the dependencies in the Serverpod mono repo.

Sample output when running the command on the state of the repo now:

```
$ serverpod analyze-pubspecs --check-latest-version --only-major --ignore-serverpod
Dependencies match.
Checking latest pub versions.
web_socket_channel
local: ^2.4.0
pub:   ^3.0.1
found in:
 • serverpod_client
firebase_auth
local: ^4.4.2
pub:   ^5.1.4
found in:
 • serverpod_auth_firebase_flutter
firebase_core
local: ^2.10.0
pub:   ^3.3.0
found in:
 • serverpod_auth_firebase_flutter
material_design_icons_flutter
local: >=6.0.0 <7.0.0
pub:   ^7.0.7296
found in:
 • serverpod_auth_firebase_flutter
 • serverpod_auth_apple_flutter
email_validator
local: ^2.1.17
pub:   ^3.0.0
found in:
 • serverpod_auth_server
 • serverpod_auth_email_flutter
googleapis
local: ^11.0.0
pub:   ^13.2.0
found in:
 • serverpod_auth_server
image_cropper
local: ^5.0.0
pub:   ^8.0.1
found in:
 • serverpod_auth_shared_flutter
```

## Pre-launch Checklist

## Breaking changes

Changes to a hidden developer command that targets the development of Serverpod.

### DIFF
--- a/tools/serverpod_cli/lib/src/internal_tools/analyze_pubspecs.dart
+++ b/tools/serverpod_cli/lib/src/internal_tools/analyze_pubspecs.dart
@@ -8,9 +8,21 @@ import 'package:serverpod_cli/src/util/directory.dart';
 import 'package:serverpod_cli/src/util/pubspec_helpers.dart';
 import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart';
 
+class CheckLatestVersion {
+  final bool onlyMajorUpdate;
+  final bool ignoreServerpodPackages;
+
+  CheckLatestVersion({
+    required this.onlyMajorUpdate,
+    required this.ignoreServerpodPackages,
+  });
+}
+
 /// The internal tool for analyzing the pubspec.yaml files in the Serverpod
 /// repo.
-Future<bool> pubspecDependenciesMatch(bool checkLatestVersion) async {
+Future<bool> pubspecDependenciesMatch({
+  required CheckLatestVersion? checkLatestVersion,
+}) async {
   var directory = Directory.current;
   if (!isServerpodRootDirectory(directory)) {
     log.error('Must be run from the serverpod repository root');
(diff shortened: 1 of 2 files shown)


tools/serverpod_cli/lib/src/internal_tools/analyze_pubspecs.dart
```diff
@@ -62,8 +81,23 @@ Future<void> _checkLatestVersion(
         log.error(e.message);
       }
 
-      if (latestPubVersion != null &&
-          depVersion != '^${latestPubVersion.toString()}') {
+      if (ignoreServerpodPackages && depName.startsWith('serverpod')) {
```
hampuslavin, line 84: Could this be an issue for potential community packages (i.e. `serverpod-somepackage-adapter`)?  If yes, maybe we need to have an array of known serverpod packages to check against instead.

tools/serverpod_cli/lib/src/internal_tools/analyze_pubspecs.dart
```diff
@@ -62,8 +81,23 @@ Future<void> _checkLatestVersion(
...
+        continue;
+      }
+
+      var depVersion = VersionConstraint.parse(deps.first.version);
```
hampuslavin, line 88: Suggestion to move this line below the if, since it's not used before it.

tools/serverpod_cli/lib/src/internal_tools/analyze_pubspecs.dart
SandPod: Yes that is correct, if a package is named something like that we would ignore it.

But we would only analyze it if one our our Serverpod packages would depend on it. 

This command is only an internal tool to manager our mono repo. So we can probably ignore the wider community use.

vlidholt (approved):
LGTM

tools/serverpod_cli/lib/src/internal_tools/analyze_pubspecs.dart
vlidholt: This is a good observation. Perhaps add a todo and a link to an issue, so it's easy to fix if we run into it in the future?

SandPod: Sounds like a plan! 👍

Issue #2339, opened 2024-06-03
You are able to register a CustomClass and a model with the same name.


Isakdl:
**Describe the bug**
Register a custom class with the name `CustomClass`.

Then create a model:

```yaml
class: CustomClass
fields:
  name: String
```

Run `serverpod generate`

Then no errors are reported.

But the stream deserializer will try to parse all model defined classes as the custom class.

**Expected behavior**
Easiest solution that is reasonable is to not allow this naming collision in our LSP.

PR #2604, opened 2024-08-14
fix: Validate model and custom class names collisions

hampuslavin:
## Description
This PR fixes [#2339](https://github.com/serverpod/serverpod/issues/2339).

Adds an additional check for custom classes to the `Restrictions.validateClassName` method and returns an error in case there are collisions. Tried to keep the error message similar to the one for colliding model class names:
```
Model class collision:  "The class name "HampusCustomClass" is already used by another model class."
Custom class collision: "The class name "HampusCustomClass" is already used by a custom class (package:onboarding_todo_app_shared/onboarding_todo_app_shared.dart)."
```
## Changes
- Fixes a tiny typo
- Adds validation for collisions between model class names and custom class names 
image
image

## Pre-launch Checklist

### DIFF
--- a/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
@@ -164,6 +164,21 @@ class Restrictions {
       ];
     }
 
+    var duplicateExtraClass =
+        config.extraClasses.cast<TypeDefinition?>().firstWhere(
+              (extraClass) => extraClass?.className == className,
+              orElse: () => null,
+            );
+
+    if (duplicateExtraClass != null) {
+      return [
+        SourceSpanSeverityException(
+          'The $documentType name "$className" is already used by a custom class (${duplicateExtraClass.url}).',
+          span,
+        )
+      ];
+    }
+
     var classesByName = parsedModels.classNames[className]?.where(
         (model) => model.moduleAlias == documentDefinition?.moduleAlias);
 
--- a/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_conflicts_test.dart
+++ b/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_conflicts_test.dart
@@ -1,3 +1,4 @@
+import 'package:serverpod_cli/analyzer.dart';
 import 'package:serverpod_cli/src/analyzer/models/stateful_analyzer.dart';
 import 'package:serverpod_cli/src/generator/code_generation_collector.dart';
 import 'package:serverpod_cli/src/test_util/builders/generator_config_builder.dart';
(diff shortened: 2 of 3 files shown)


tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -158,7 +159,22 @@ class Restrictions {
...
+          span,
+        )
+      ];
+    }
+
+    var duplicateExtraClass =
+        (config.extraClasses as List<TypeDefinition?>).firstWhere(
+      (extraClass) => extraClass?.className == className,
+      orElse: () => null,
+    );
```
hampuslavin, line 172: Since `collection` is not a [dependency of serverpod_cli](https://github.com/serverpod/serverpod/blob/main/tools/serverpod_cli/pubspec.yaml), this seems to be the way to do it. Not sure if we want to consider adding the `collection` dependency like we do in the main serverpod package, which would contain the `firstWhereOrNull` helper?

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -164,6 +164,21 @@ class Restrictions {
       ];
     }
 
+    var duplicateExtraClass =
+        config.extraClasses.cast<TypeDefinition?>().firstWhere(
+              (extraClass) => extraClass?.className == className,
+              orElse: () => null,
+            );
```
hampuslavin, line 171: Since collection is not a [dependency of serverpod_cli](https://github.com/serverpod/serverpod/blob/main/tools/serverpod_cli/pubspec.yaml), this seems to be the way to do it. Not sure if we want to consider adding the collection dependency like we do in the main serverpod package, which would contain the firstWhereOrNull helper? I would assume there would be more situations in which it would come in handy.

SandPod: It's probably not a bad thing to add the dependency here as well.
But keeping dependencies up to date is always a hassle. So if we are only interested in the `firstWhereOrNull` helper for lists, we might as well implement it as an extention ourselves and skip the dependency.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_conflicts_test.dart
```diff
@@ -44,6 +45,42 @@ void main() {
     );
   });
 
+  test(
+      'Given a model and a custom class with the same class name, then an error is collected that there is a collision in the class names.',
```
SandPod, line 49: Comment(Does not require fix): The description here is missing the `when` part. This is reflected by all tests in these files. So if it is fix is done, it should be reflected in all files.

SandPod (approved):
LGTM! Nice work! 🎉 🌮

PR #2607, opened 2024-08-14
chore: Bump image_cropper to '>=7.0.0 <9.0.0', min Dart to '3.3.0' and min Flutter to '3.19.0'

mkoehne:
Serverpod uses an old version of image_cropper.
This PR updates image_cropper from 5.0.0 to 8.0.0.

## Pre-launch Checklist

- [ X] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [X ] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
- [ X] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).

## Breaking changes
No breaking changes.


mkoehne:

SandPod:
Hey @mkoehne! 

Thank you for contributing to Serverpod 🚀🙌

Our aim for the framework is to always support the latest version of all dependencies we use, and to be as unopinionated as possible around what dependencies are used by developers.

To reach this goal, we always try to allow the widest version range possible for the dependencies used. This leads to us often support multiple major versions of a package if that is possible.

For minor and patch versions, we often don't update these unless there is an important feature or bugfix required by the framework.

For this contribution to be accepted, please remove all patch and minor version updates and only update the ones that have major version updates.

You can find out which these are by activating the CLI from latest main, and then run the hidden command `serverpod analyze-pubspecs --check-latest-version --only-major --ignore-serverpod` (This command was introduced to main yesterday).

mkoehne:
Hey @SandPod :)

Thanks for the explanation. 
This makes total sens and I have updated the PR to just update image_cropper.
That was the package that was making problems with the my current project and was the only packages that was shown from the new command for the packages that I have previously updated.

modules/serverpod_auth/serverpod_auth_shared_flutter/lib/src/image_uploader.dart
```diff
@@ -69,16 +69,17 @@ class ImageUploader {
...
           AndroidUiSettings(
               toolbarTitle: 'Cropper',
               toolbarColor: toolbarColor,
               toolbarWidgetColor: toolbarWidgetColor,
               initAspectRatio: CropAspectRatioPreset.original,
+              cropStyle: CropStyle.circle,
               lockAspectRatio: false),
           IOSUiSettings(
             minimumAspectRatio: 1.0,
+            cropStyle: CropStyle.circle,
```
SandPod, line 82: Suggestion: It seems like this change was introduced in version `7.0.0` of image cropper. Could we add a range instead, starting from that version? That is `'>=7.0.0 <9.0.0'`

This would allow us to keep our lowest dart version and probably fix the issues encountered in in the CI.

SandPod (changes requested):
Almost there!

modules/serverpod_auth/serverpod_auth_shared_flutter/lib/src/image_uploader.dart
mkoehne: Yeah that is better. I have updated the version.

SandPod (approved):
Awesome work and thank you for fixing this! 🙌

Issue #2192, opened 2024-04-29
Generated copyWith() for nested lists doesn't keep its inner child values


daniel-webstep:
**Describe the bug**
When creating models that contains nested lists, the code generation for `copyWith()` doesn't keep track of the inner child. Instead, an empty list is returned. The `toJson()` works as expected.

**To Reproduce**
1. Create a class like
```
class: NestedLists
fields:
  doubleValue: double
  listDouble: List<double>
  listListDouble: List<List<double>>
```
2. run `serverpod generate`
3. Create the following tests
```
import 'dart:convert';

import 'package:example_server/src/generated/protocol.dart';
import 'package:test/test.dart';

void main() {
  NestedLists baseObject() {
    return NestedLists(doubleValue: 0, listDouble: [
      1
    ], listListDouble: [
      [2]
    ]);
  }

  group('NestedLists', () {
    test('toJson() returns a valid json for nested lists.', () {
      var obj = baseObject();
      var expectedJson = jsonDecode(
          '{"doubleValue": 0.0, "listDouble": [1.0], "listListDouble": [[2.0]]}');
      expect(obj.toJson(), equals(jsonDecode(expectedJson)));
    });
    test('copyWith() does return the same values.', () {
      var obj = baseObject();
      var expectedObj = baseObject();
      expect(obj.copyWith(), expectedObj);
    });
  });
}
```
4. Run the tests
5. See error

**Expected behavior**
```
Expected: _NestedListsImpl:<{"doubleValue":0.0,"listDouble":[1.0],"listListDouble":[[2.0]]}>
  Actual: _NestedListsImpl:<{"doubleValue":0.0,"listDouble":[1.0],"listListDouble":[]}>

package:matcher                                          expect
package:example_server/src/test/nested_lists_test.dart 25:7  main.<fn>.<fn>
```

**Serverpod versions**
 -  _CLI Version:_ Serverpod version: 1.2.6
 - _Library version:_ serverpod: 1.2.6

**Platform information**
Ubuntu 22.04.4 LTS through Docker.

Isakdl:
Thank you for the detailed report!

bfemmer:
I'll pick this one up to work on (I don't see a linked PR ... if anyone is already working this one, please let me know).

SandPod:
Go for it @bfemmer

PR #2612, opened 2024-08-15
fix: Fix copyWith for nested Lists and Maps

hampuslavin:
## Description
Currently `Map` and `List` objects use the [`CloneList`](https://github.com/serverpod/serverpod/blob/c86905f/packages/serverpod_serialization/lib/src/copy_with.dart#L19) and [`CloneMap`](https://github.com/serverpod/serverpod/blob/c86905f/packages/serverpod_serialization/lib/src/copy_with.dart#L28) extensions in their `copyWith` implementation (as defined in the generator [here](https://github.com/serverpod/serverpod/blob/c86905f/tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart#L434-L435)). This doesn't work for nested objects due to dart's casting rules:
```dart
List<List<SimpleData>> simpleList = [[SimpleData()]];
var dynamicCopy = simpleList.clone(); // ✅ Works
List<List<SimpleData>> typedCopy = simpleList.clone(); // ❌ Throws "type 'List<dynamic>' is not a subtype of type 'List<SimpleData>'"
```
The current code doesn't throw for `List` (though it does for `Map`) due to the `whereType<T>()` in the [`CloneList`](https://github.com/serverpod/serverpod/blob/c86905f/packages/serverpod_serialization/lib/src/copy_with.dart#L19) extension. This is what causes the nested `List` to be empty instead, as reported in [#2192](https://github.com/serverpod/serverpod/issues/2192).

There is no way to get around this issue when cloning arbitrarily nested objects (the above also applies to `Map`). [See this issue](https://github.com/dart-lang/language/issues/1574) for a more in depth discussion on the topic.

This PR fixes [#2192](https://github.com/serverpod/serverpod/issues/2192) by leveraging code generation to generate all cloning code explicitly (similar to the to/from JSON handling), which removes the need for casting `List<dynamic>`.

## Changes
- Adds code generation logic for cloning arbitrary `List` and `Map` objects 
- Removes the now unused cloning extensions for `List` and `Map`
- Does all cloning logic ahead of runtime. This means the new way should be more performant because there is no type checking needed during runtime to determine how to clone the type.
- (testing) Adds new properties to the test model `object_with_object.spy.yaml` and corresponding tests that all fail for the current implementation
## Pre-launch Checklist


packages/serverpod_serialization/lib/serverpod_serialization.dart
```diff
@@ -3,7 +3,7 @@ library serverpod_serialization;
 export 'package:uuid/uuid.dart';
 
 export 'src/bytedata_base64_ext.dart';
-export 'src/copy_with.dart';
+export 'src/copy_with.dart'; // hide CloneList, CloneMap, CloneByteData;
```
hampuslavin, line 6: Hide also hides it from where its used within Serverpod

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -464,6 +464,99 @@ class SerializableModelLibraryGenerator {
...
+    return _buildDeepCloneTree(
+      field.type,
+      field.name,
+      root: true,
+    );
+  }
+
+  Expression _buildDeepCloneTree(TypeDefinition type, String variableName,
+      {int depth = 0, bool root = false}) {
+    var isLeafNode = type.generics.isEmpty;
```
hampuslavin, line 485: This should be robust, but somehow looks a bit off to my eye. Maybe there is a cleaner check? I think this should be equal:
```dart
var isLeafNode = !(field.type.isListType || field.type.isMapType);
```

packages/serverpod_serialization/lib/src/copy_with.dart
```diff
@@ -59,7 +39,12 @@ const _noneMutableTypes = [
   UuidValue,
 ];
 
-dynamic _guardedCopyWith(dynamic element) {
+/// Keyword helper for the `cloneTrivial` function
+const cloneTrivialFunctionName = 'cloneTrivial';
```
hampuslavin, line 43: Any ideas for a better name? Maybe `cloneNonNestedTypes` is more clear? Not sure what feels most intuitive!

tests/serverpod_test_server/test_e2e/service_protocol_test.dart
```diff
@@ -430,17 +430,19 @@ void main() {
             .map((c) => c.dartType)
             .toList();
 
-        expect(columns, hasLength(7));
         expect(
             columns,
-            containsAll([
+            equals([
```
hampuslavin, line 435: Unless I'm missing something this is equal but shorter.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -464,6 +464,99 @@ class SerializableModelLibraryGenerator {
...
+      return refer(cloneTrivialFunctionName, serverpodUrl(serverCode))
+          .call([refer(variableName)]);
+    }
+
+    var nextCallback = switch (type.className) {
+      ListKeyword.className =>
+        _buildListCloneCallback(type.generics.first, depth),
+      MapKeyword.className =>
+        _buildMapCloneCallback(type.generics[0], type.generics[1], depth),
+      _ => throw "Can't clone type ${type.className}",
```
hampuslavin, line 496: What's the convention when throwing stuff? Is it better to always wrap it in an Exception type?

tools/serverpod_cli/test/generator/dart/server_code_generator/class_copy_with_test.dart
```diff
@@ -343,8 +344,8 @@ void main() {
     test('then the clone method is called on field when copying the object.',
         () {
       var sourceCode = copyWithMethod?.body.toSource();
-      expect(
-          sourceCode, '{return Example(names: names ?? this.names.clone());}');
+      expect(sourceCode,
+          '{return Example(names: names ?? this.names.map((e0) => _i1.$cloneTrivialFunctionName(e0)).toList());}');
```
SandPod, line 348: Suggestion: Align description with the new expect.

We no longer simply call clone on the object. Please align the description with the new expect of the test.

tests/serverpod_test_server/migrations/migration_registry.txt
```diff
@@ -6,3 +6,4 @@
 
 20240125141253049
 20240725082303667
+20240816094748409
```
SandPod, line 9: Suggestion: Combine these changes with the previous migration.

We try to only introduce a single migration with every major update.
There is no need to record minor fixes since we constantly "nuke" the test docker container.

Instead of creating a new migration, please remove the last one and recreate it with your new model definition.

tests/serverpod_test_server/test/generated_methods/copy_with_mutable_test.dart
```diff
@@ -235,5 +235,110 @@ void main() {
...
+      );
+
+      test('then a nested map should be deeply cloned', () {
+        var copy = objectWithNestedObjects.copyWith();
+
+        expect(copy.nestedDataMap?['firstKey']?[333]?.num, equals(333));
+
+        expect(
+          SerializationManager.encode(copy.nestedDataMap?.toJson()),
+          equals('{"firstKey":[{"k":333,"v":{"num":333}}]}'),
```
SandPod, line 272: Suggestion: Expect the fields intead of the to json representation.

This test now depends on the serializations encode method which is not really related to the functionality that we want to assert.

Like the rest of the test in this file it should be enough to validate that the fields are as expected.

tests/serverpod_test_server/test/generated_methods/copy_with_mutable_test.dart
```diff
@@ -235,5 +235,110 @@ void main() {
...
+          [SimpleData(num: 88), SimpleData(num: 99)],
+        ],
+        nestedDataMap: {
+          'firstKey': {
+            333: SimpleData(num: 333),
+          },
+        },
+      );
+
+      test('then a nested map should be deeply cloned', () {
```
SandPod, line 265: Suggestion: Change description to `then nested map IS deeply cloned`.

I think it is more clear to state the expect rather than the intent.

tests/serverpod_test_server/test/generated_methods/copy_with_mutable_test.dart
```diff
@@ -235,5 +235,110 @@ void main() {
...
+        var copy = objectWithNestedObjects.copyWith();
+
+        expect(copy.nestedDataList, hasLength(1));
+        expect(copy.nestedDataList?.first, hasLength(2));
+        expect(copy.nestedDataList?.first.first.num, equals(88));
+
+        var originalDeepList =
+            objectWithNestedObjects.nestedDataList?.first.first;
+        var newDeepList = copy.nestedDataList?.first.first;
+        expect(originalDeepList, isNot(equals(newDeepList)));
```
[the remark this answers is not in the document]
SandPod: Suggestion: I understand your intent but I would not rely on the reference in dart. The other test in this file validate this condition by mutating the original object and then assert that the clone is not modified.

In the spirit of "least expected" lets use the same pattern for these new tests as well.

tests/serverpod_test_server/test/generated_methods/copy_with_mutable_test.dart
```diff
@@ -235,5 +235,110 @@ void main() {
       expect(mapDataCopy.data['b']?.num, 2);
       expect(mapDataCopy.data['b']?.id, null);
     });
+
+    group('Given an object with nested Lists and Maps when calling copyWith',
```
SandPod, line 239: Question: Are we really validating the mutability here? Are we not validating that the values are still accessible? Maybe split mutability and accessibility validation.

tests/serverpod_test_server/test_e2e/service_protocol_test.dart
SandPod: Equals also puts restrictions on the order of the items in the list while containsAll just validates that each entry is found. After this change, this test would fail if the order of the elements changes.

Since we are fetching something from the database I would assume that order is unpredictable and therefore should be excluded.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -432,13 +433,12 @@ class SerializableModelLibraryGenerator {
           noneMutableTypeNames.contains(field.type.className))) {
         assignment = refer('this').property(field.name);
       } else if (clonableTypeNames.contains(field.type.className)) {
-        assignment = _buildMaybeNullMethodCall(field, 'clone');
+        assignment = _buildClonableCopyWithAssignment(field);
```
SandPod, line 436: Suggestion: Branch based on what should be built here instead of inside of `_buildClonableCopyWithAssignment`.

By making this code a bit more shallow it becomes easier to follow.

SandPod: We could call `_buildDeepCloneTree` on this abstraction level for the related classes.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
SandPod: I think this should be fine, but we might encounter issues if we ever allow "records" for example.

tools/serverpod_cli/test/generator/dart/client_code_generator/class_copy_with_test.dart
```diff
@@ -383,7 +384,8 @@ void main() {
     test('then the clone method is called on field when copying the object.',
         () {
       var sourceCode = copyWithMethod?.body.toSource();
-      expect(sourceCode, '{return Example(map: map ?? this.map.clone());}');
+      expect(sourceCode,
+          '{return Example(map: map ?? this.map.map((key0, value0) => MapEntry(_i1.$cloneTrivialFunctionName(key0), _i1.$cloneTrivialFunctionName(value0))));}');
```
SandPod, line 388: Suggestion: Please update the description so that it matches the new test case.

Same comment here as for the other updated tests.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
SandPod: Yes, we usually try to use a defined exception type instead of throwing just a `String`.

There are a couple of Exceptions defiend that can be handy along with Errors if the state is unrecourable.
I would say an error is more in line with the code here since we always want the generator to stop if this occurs.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -464,6 +464,99 @@ class SerializableModelLibraryGenerator {
...
+    }
+        .call([nextCallback]);
+
+    return type.isListType
+        ? expression.property(ListKeyword.toList).call([])
+        : expression;
+  }
+
+  Expression _buildListCloneCallback(TypeDefinition type, int depth) {
+    var variableName = 'e$depth';
```
SandPod, line 516: Comment: Neat addition with the depth recorded in the variable name. Maybe we should add this to the `fromJson` structure as well? I don't think we record the depth there.

SandPod (changes requested):
Awesome work and great research into finding the issue here.

I have a couple of comments and suggestions.

packages/serverpod_serialization/lib/serverpod_serialization.dart
```diff
@@ -3,7 +3,7 @@ library serverpod_serialization;
 export 'package:uuid/uuid.dart';
 
 export 'src/bytedata_base64_ext.dart';
-export 'src/copy_with.dart';
+export 'src/copy_with.dart' hide CloneByteData;
```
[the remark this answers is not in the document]
hampuslavin: Ahh, right! 👍

packages/serverpod_serialization/lib/src/copy_with.dart
```diff
@@ -59,7 +39,12 @@ const _noneMutableTypes = [
...
+
+/// Clones all types except `List` and `Map`, which can't be deep copied in a type safe manner.
+/// See https://github.com/serverpod/serverpod/pull/2612 for more information.
+T cloneTrivial<T>(T element) {
```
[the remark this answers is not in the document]
hampuslavin: `strictShallowClone` is a good suggestion, thanks! 👍

packages/serverpod_serialization/lib/src/copy_with.dart
```diff
@@ -68,13 +53,14 @@ dynamic _guardedCopyWith(dynamic element) {
...
+  if (element is List || element is Map) {
+    throw Exception("Can't clone `List` or `Map` types");
+  }
 
   try {
-    return element.copyWith();
+    return (element as dynamic).copyWith() as T;
   } on NoSuchMethodError {
-    throw 'No copyWith method found on ${element.runtimeType}';
+    throw Exception('No copyWith method found on ${element.runtimeType}');
```
hampuslavin, line 64: @SandPod Would you like me to throw `NoSuchMethodError` here instead of `Exception`?

packages/serverpod_serialization/lib/src/copy_with.dart
hampuslavin: Good suggestion, moved to the private `Keywords` helper 👍

packages/serverpod_serialization/lib/src/copy_with.dart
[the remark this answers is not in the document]
hampuslavin: @SandPod do you think it's okay that Serverpod exports this helper publicly?

tests/serverpod_test_server/migrations/migration_registry.txt
hampuslavin: Makes sense, will fix! 👍

tests/serverpod_test_server/test/generated_methods/copy_with_mutable_test.dart
hampuslavin: Nice spot! I missed the top level `group`. Do you often have a top level group like that or is it older code that doesn't strictly adhere to given/when/then?

tests/serverpod_test_server/test/generated_methods/copy_with_mutable_test.dart
hampuslavin: I wanted the [Jest's inline snapshot helpers](https://jestjs.io/docs/snapshot-testing#inline-snapshots) badly and couldn't let it go, but I agree it doesn't make sense to pull in `SerializationManager` here 😄 In general I do think snapshots can be great in some scenarios for testing objects because you easily see that something you didn't expect to be there is there.

tests/serverpod_test_server/test_e2e/service_protocol_test.dart
hampuslavin: Ah right, makes sense then 👍  Thanks!

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
hampuslavin: Will throw `UnimplementedError` instead

tools/serverpod_cli/test/generator/dart/client_code_generator/class_copy_with_test.dart
hampuslavin: Ahh, good catch! 💯

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
hampuslavin: Happy to make the change but I personally it's not obvious to me that it would make it more readable. It would look like this:
```dart
if ((field.type.isEnumType ||
      noneMutableTypeNames.contains(field.type.className))) {
    assignment = refer('this').property(field.name);
} else if (clonableTypeNames.contains(field.type.className)) {
  var shouldDeepClone = field.type.isListType || field.type.isMapType;
  assignment = switch (shouldDeepClone) {
    true => _buildDeepCloneTree(
        field.type,
        field.name,
        root: true,
      ),
    false => refer(Keyword.strictShallowCloneFunctionName,
            serverpodUrl(serverCode))
        .call([refer(field.name)]),
  };
} else {
  assignment = _buildMaybeNullMethodCall(field, 'copyWith');
}
```
- In general I believe nested code is harder to follow and especially if you want to quickly skim the code for the relevant parts. When skimming I want to quickly find the relevant branch (i.e. enum, clonable type or else. Having too much noise makes this search harder.
- I also believe it helps with readability when different parts of a function body has roughly the same branch depth/complexity. When one part sticks out like this I think it can be a hint that it should be broken out.

I do agree that _too_ many layers of abstractions makes code hard to read as well, especially in class hierarchies. But for class methods which are easy to jump around between I generally tolerate a bit more.

Again, happy to make the change but would be curious to hear your thoughts on my points above, both in general and in this specific case!

packages/serverpod_serialization/lib/src/copy_with.dart
```diff
@@ -68,13 +50,10 @@ dynamic _guardedCopyWith(dynamic element) {
...
-  if (element is Map) return element.clone();
-
-  try {
-    return element.copyWith();
-  } on NoSuchMethodError {
-    throw 'No copyWith method found on ${element.runtimeType}';
+  if (element is ByteData) return element.clone() as T;
+  if (element is List || element is Map) {
+    throw UnimplementedError("Can't clone `List` or `Map` types");
   }
```
SandPod, line 56: Suggestion: Use `UnsupportedError` instead.

Since the type is actually not supported by the method. Could also explain why these can not be cloned by this method. `"List" and "Map" are not shallow types`.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
SandPod: I think it would make more sense to keep the if statement shallow. That is, add the check for isListType and isMapType to the outer if statement.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
SandPod: Makes total sense! Yea, a follow-up PR would be good, then we maintain the same style across the generated code.

SandPod (commented):
Almost there, Let's just finish the discussion around branching in the generator.

packages/serverpod_serialization/lib/src/copy_with.dart
SandPod: The suggested solution in the PR is good now! 👍

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
hampuslavin: Apologies, I missed that it was possible to simply flatten it out to the top level if/else. Will fix, thanks! 👍

packages/serverpod_serialization/lib/src/copy_with.dart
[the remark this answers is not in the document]
SandPod: Did you ever create the issue? Feel free to link it here as well.

SandPod (approved):
LGTM! 🚀
Nice work! 🙌

PR #2614, opened 2024-08-16
test: Added deprecated methods to serverpod_test project.

christerswahn:
Added deprecated methods to serverpod_test project.

## Pre-launch Checklist

## Breaking changes

n/a

### DIFF
--- a/tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
@@ -819,7 +819,9 @@ class LibraryGenerator {
             ..body = refer('endpoints')
                 .index(literalString(endpoint.name))
                 .asA(refer(endpoint.className, _endpointPath(endpoint)))
-                .property(method.name)
+                .property(
+                  '${_getMethodCallComment(method) ?? ''}${method.name}',
+                )
                 .call([
               refer('session'),
               for (var param in [
@@ -835,6 +837,15 @@ class LibraryGenerator {
       });
     }
     return methodConnectors;
+  }
+
+  String? _getMethodCallComment(MethodCallDefinition m) {
+    for (var a in m.annotations) {
+      if (a.methodCallAnalyzerIgnoreRule != null) {
+        return '\n// ignore: ${a.methodCallAnalyzerIgnoreRule}\n';
+      }
+    }
+    return null;
   }
 
   Map<Object, Object> _buildMethodStreamConnectors(
--- a/tools/serverpod_cli/test/generator/dart/server_code_generator/endpoints_test.dart
+++ b/tools/serverpod_cli/test/generator/dart/server_code_generator/endpoints_test.dart
@@ -240,4 +240,50 @@ void main() {
           contains('streamParams[\'streamParam\']!.cast<String>()'));
     });
   });
+
+  group(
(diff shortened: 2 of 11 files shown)


tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -837,6 +839,15 @@ class LibraryGenerator {
     return methodConnectors;
   }
 
+  String? _getMethodCallComment(MethodCallDefinition m) {
+    for (var a in m.annotations) {
+      if (a.methodCallAnalyzerIgnoreRule != null) {
+        return '\n// ignore: ${a.methodCallAnalyzerIgnoreRule}\n';
```
SandPod, line 845: Suggestion(Non blocking and opinionated): Use a list of strings for the ignore rules and change the name to method call comments.

With the current solution we place half of the statement in the analyzer (`deprecated_member_use_from_same_package`) and half of the statement in the generator (`ignore:`). A suggestion is to have the full statement in the analyzer and then just have a list of comments that are inserted at the method call site.

That way, you can easily see the fully constructed comment in a single place.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -837,6 +839,15 @@ class LibraryGenerator {
     return methodConnectors;
   }
 
+  String? _getMethodCallComment(MethodCallDefinition m) {
```
SandPod, line 842: Suggestion: Please add a test for the library generator to validate that the comment is inserted as expected when supplied.

SandPod (changes requested):
Awesome, thanks for the tests for this.

I have a suggestion for a different approach to the method call comment. Not sure if it is the best way though.

Since we added new functionality to the library generator we should add a test that validates the comment is inserted as well.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
christerswahn: That is true, it would be more of the needed logic for this use case is in one place. However it would mean that the analyzer code knows about code generation - e.g. shall an ignore statement be made for a line of for a file. I thought it better for the analyzer to represent the analyzer rule that is relevant, and the generator to decide how to generate ignore code.

tools/serverpod_cli/test/generator/dart/server_code_generator/endpoints_test.dart
```diff
@@ -240,4 +240,50 @@ void main() {
...
+            .withClassName('${endpointName.pascalCase}Endpoint')
+            .withName(endpointName)
+            .withMethods([
+          MethodDefinitionBuilder().withName(methodName).withAnnotations([
+            const AnnotationDefinition(
+              name: 'Deprecated',
+              arguments: ["'This method is deprecated.'"],
+              methodCallAnalyzerIgnoreRule:
+                  'deprecated_member_use_from_same_package',
+            )
```
SandPod, line 261: Suggestion(Non-Blocking): Add a builder pattern for AnnotationDefinition.

We use a builder pattern for all other definitions. The purpose of that is so that we don't have to change all tests that instantiate the class if we change the signature of the class. This way we always have a solid and valid AnnotationDefinition.

SandPod (approved):
LGTM!

One suggestion that will help future developers and make it easy to maintain tests using the AnnotationDefinition class.

PR #2618, opened 2024-08-16
chore: Enable dependabot version update pull requests for pub.

SandPod:
Allows dependabot to create PRs for pub updates that are not compatible with the current version range.

Configures the versioning strategy to `widen` which means dependabot will atempt to widen the allowed dependency range to include the new release.

Since we are a library mono-repo we need to keep the same version for a dependency across all packages. Dependabot doesn't seem to support that sort of grouping. As a workaround we will limit the number of open pull requests to 1, this makes it easy for us to find all instances of that dependency and update it in all packages without getting spammed.

Asked a question in the Dependabot discussions around the possibility to group a dependency update for all packages here: https://github.com/dependabot/dependabot-core/discussions/10454

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,5 @@
 version: 2
 updates:
-  # Pub dependencies
   - package-ecosystem: "pub"
     directories:
       - packages/**/*
@@ -11,8 +10,9 @@ updates:
     schedule:
       interval: "daily"
     versioning-strategy: "widen"
-    # Since we are using a monorepo, we want to update the same package
+    # Since this is a library monorepo, it is important that the same package is updated
     # in all directories at the same time.
-    # Currently dependabot does not support this, so we limit the number of
-    # open pull requests to 1 to avoid conflicts.
+    # Currently dependabot does not support this type of grouping, therefore the number
+    # of open pull requests is limited to 1. This prevents multiple pull requests
+    # from being opened for the same dependency.
     open-pull-requests-limit: 1
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+  # Pub dependencies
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
...
+      - integrations/**/*
+      - templates/serverpod_templates
+      - tools/serverpod_cli
+    schedule:
+      interval: "daily"
+    versioning-strategy: "widen"
+    # Since we are using a monorepo, we want to update the same package
+    # in all directories at the same time.
+    # Currently dependabot does not support this, so we limit the number of
+    # open pull requests to 1 to avoid conflicts.
```
hampuslavin, line 17: Perhaps you could link to the issue you created here as well?

.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
+version: 2
+updates:
+  # Pub dependencies
```
hampuslavin, line 3: nitpick: redundant comment, suggestion to remove it

.github/dependabot.yml
hampuslavin, line 17: Question: Have you ever discussed what tone of voice you'd like to have in in-code documentation/comments? This one is for Serverpod developers only, but since it's open source it's kind of "public documentation" as well. Perhaps it should be that same as in https://docs.serverpod.dev/? In that case the comment would avoid words like "we" and be more formal
```
Since this is a monorepo, all packages in all directories should be updated at the same time.
Etc...
```

hampuslavin (approved):

.github/dependabot.yml
SandPod: There is no specific issue so far, I have opened a discussion question to try and figure out if it is possible to achieve in some way.

hampuslavin (approved):
Nice! 🌟

PR #2632, opened 2024-08-20
tests: Fix flaky tests around custom password hash generator.

SandPod:
This PR fixes an issue where the last two tests in the `email_auth_provider_test.dart` file could fail.

The test failure could be reproduced to 100% if a delay was added to the test.
The running theory is that the Dart garbage collector jumps in early because the 3third to last group does not use the `session` variable. This causes the finalizer to activate before all "true" references to the variable are out of scope.

I'll try to reproduce the issue in an isolated test and file it to the Dart team.

To fix the tests the new tests were moved to an isolated file and run there. This makes sense because we also set a global static when calling `AuthConfig.set(....)`.

## Pre-launch Checklist

## Breaking changes

None, just test restructuring and fixes.


Isakdl (approved):
Nice! We all hate flaky shit 🛑! 

LGTM! 🚀

PR #2634, opened 2024-08-20
fix: Refactored ServerpodClientShared.callServerEndpoint() and subclasses to unify identical code.

christerswahn:
Refactored ServerpodClientShared.callServerEndpoint() and subclasses to unify identical code.

## Pre-launch Checklist

## Breaking changes

n/a


packages/serverpod_client/lib/src/serverpod_client_browser.dart
```diff
@@ -33,34 +29,20 @@ abstract class ServerpodClient extends ServerpodClientShared {
...
   }
 
-  Future<void> _initialize() async {
-    _initialized = true;
-  }
-
   @override
-  Future<T> callServerEndpoint<T>(
-      String endpoint, String method, Map<String, dynamic> args) async {
-    if (!_initialized) await _initialize();
```
christerswahn, line 43: This has no effect besides potentially relinquishing the thread to the task runner before continuing. I assume that is not necessary, in which case this is dead code that should perhaps be removed.

packages/serverpod_client/lib/src/serverpod_client_io.dart
```diff
@@ -37,65 +36,32 @@ abstract class ServerpodClient extends ServerpodClientShared {
...
   }
 
-  Future<void> _initialize() async {
-    _initialized = true;
-  }
-
   @override
-  Future<T> callServerEndpoint<T>(
-      String endpoint, String method, Map<String, dynamic> args) async {
-    if (!_initialized) await _initialize();
```
christerswahn, line 47: This has no effect besides potentially relinquishing the thread to the task runner before continuing. I assume that is not necessary, in which case this is dead code that should perhaps be removed.

packages/serverpod_client/lib/src/serverpod_client_browser.dart
```diff
@@ -70,28 +52,10 @@ abstract class ServerpodClient extends ServerpodClientShared {
...
-
-      if (logFailedCalls) {
-        print('Failed call: $endpoint.$method');
-        print('$e');
-      }
-      rethrow;
+      return data;
+    } on http.ClientException catch (e) {
+      var message = 'Unknown server response code. ($e)';
+      throw (ServerpodClientException(message, -1));
```
christerswahn, line 58: This behavior will be subtly different with this refactoring. Before, the ServerpodClientException would be thrown immediately without performing the latter part of the catch clause. Now, it will be handled by the full catch clause (which is now moved to the parent method) _just like all other exceptions_, which seems sensible.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -427,6 +429,54 @@ abstract class ServerpodClientShared extends EndpointCaller {
...
+        result = parseData<T>(data, T, serializationManager);
+      }
+
+      onSucceededCall?.call(callContext);
+      return result;
+    } catch (e, s) {
+      onFailedCall?.call(callContext, e, s);
+
+      if (logFailedCalls) {
+        // ignore: avoid_print
```
christerswahn, line 471: This code is moved from the subclasses. Should we perhaps replace the print() calls with a logger?

christerswahn:
This refactoring moves up code that is nearly identical between serverpod_client_browser.dart and serverpod_client_io.dart to their common parent in serverpod_client_shared.dart.

The behavior should be identical with two exceptions, see comments in this PR.

(In future perhaps the actual calls in the subclasses could also be unified, superficially the logic looks entirely similar, but that might subtly risk changing behavior in the client/server communication so I avoided that at least in this step.)

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -427,6 +429,54 @@ abstract class ServerpodClientShared extends EndpointCaller {
     await _sendControlCommandToStream('auth', {'key': authKey});
   }
 
+  /// Performs the actual call to the server.
+  /// Subclasses implement bespoke logic for e.g. web / io clients.
+  Future<String> callServerEndpointImpl<T>(
```
SandPod, line 434: Suggestion: Make this method private.
This is only an implementation detail that should be called internally. We don't need to expose it in the interface.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -427,6 +429,54 @@ abstract class ServerpodClientShared extends EndpointCaller {
...
+    Uri url, {
```
SandPod, line 435: Discussion: Since this isn't the full implementation for calling the server endpoint I think the name might be a bit missleading. Could we think of a better name to reflect that this just handles the request part of calling the server?

SandPod (commented):
Nice improvement! Good to harmonize the code path for shared behavior.

Our e2e tests should validate this so we don't need any new tests.

Just have some minor suggestion and a discussion for you.

packages/serverpod_client/lib/src/serverpod_client_browser.dart
christerswahn: Yes and the data var is guaranteed to be null here I believe.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
christerswahn: Agree!

packages/serverpod_client/lib/src/serverpod_client_shared.dart
christerswahn: Refactored according to conversation, replacing inheritance with composition.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -200,6 +227,7 @@ abstract class ServerpodClientShared extends EndpointCaller {
         'host must end with a slash, eg: https://example.com/');
     assert(host.startsWith('http://') || host.startsWith('https://'),
         'host must include protocol, eg: https://example.com/');
+    _requestDelegate = ServerpodClientRequestDelegateImpl(this);
```
SandPod, line 230: Suggestion: Only pass in fields that are used.

When we have god classases such as `ServerpodClientShared` it is nice to limit what is exposed of it when instantiating other classses. This makes it very clear what is actually used. Since a reference to the `ServerpodCientShared` instance is not needed, we might as well simply pass the what we need:
- `securityContext`
- `connectionTimeout`
- `serializationManager`

An added benifit of this change is that we don't need to store `securityContext` in the shared class anymore.

SandPod (commented):
Single last suggestion!

SandPod (approved):
LGTM! Great improvement of the code 🚀

Issue #2384, opened 2024-06-14
Give a warning or fail the generate command if dependencies are missing


Isakdl:
**Describe the bug**
Since the `serverpod generate` command looks at the dependencies to find serverpod modules, these modules will not be found unless the user has first done a `dart pub get` in the server project. This results in all module code being stripped from the generate command if they are not available at disk. 

We fail silently without any warnings or indications that something went wrong.

**Expected behavior**
If the `serverpod generate` is used without first running `dart pub get` (I.E. all dependencies are available), the command should either print a warning or prompt the user to update the dependencies before trying again.

PR #2637, opened 2024-08-20
 fix: warn if dependencies are missing when generating

hampuslavin:
## Description
Warns if there are packages in the `<app>_server` module that are not installed, which fixes [#2384](https://github.com/serverpod/serverpod/issues/2384).

Also captures exceptions at the top level and surfaces sad-paths that were failing silently.

## Changes
- Adds a warning if packages are not installed
- Prints out exceptions

## Pre-launch Checklist

### DIFF
--- a/tools/serverpod_cli/lib/src/config/config.dart
+++ b/tools/serverpod_cli/lib/src/config/config.dart
@@ -153,41 +153,33 @@ class GeneratorConfig {
   /// All the modules including my self and internal modules.
   List<ModuleConfig> get modulesAll => _modules;
 
-  /// Create a new [GeneratorConfig] by loading the configuration in the [dir].
-  static Future<GeneratorConfig> load([String dir = '']) async {
-    var serverPackageDirectoryPathParts = p.split(dir);
+  /// Create a new [GeneratorConfig] by loading the configuration in the [serverRootDir].
+  static Future<GeneratorConfig> load([String serverRootDir = '']) async {
+    var serverPackageDirectoryPathParts = p.split(serverRootDir);
 
     Pubspec? pubspec;
     try {
-      pubspec = parsePubspec(File(p.join(dir, 'pubspec.yaml')));
+      pubspec = parsePubspec(File(p.join(serverRootDir, 'pubspec.yaml')));
     } catch (_) {}
 
     if (pubspec == null) {
-      log.error(
-        'Failed to load pubspec.yaml. Are you running serverpod from your '
-        'projects root directory?',
-      );
-
       throw const ServerpodProjectNotFoundException(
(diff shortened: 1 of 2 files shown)


tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -152,41 +153,33 @@ class GeneratorConfig {
...
-      log.error(
-        'Failed to load pubspec.yaml. Are you running serverpod from your '
-        'projects root directory?',
-      );
-
       throw const ServerpodProjectNotFoundException(
-        'Failed to load valid pubspec.yaml',
+        'Failed to load pubspec.yaml. Are you running serverpod from your '
+        'projects server root directory?',
       );
```
hampuslavin, line 169: Confusing with two error messages, where one is never surfaced. Thought it was clearer to just have a single error message.

tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -283,6 +275,25 @@ class GeneratorConfig {
...
+
+    if (packageConfig == null) {
+      throw const ServerpodProjectNotFoundException(
+        'Failed to read your server\'s package configuration. Have you run '
+        '`dart pub get` in your server directory?',
+      );
+    }
+
+    var allPackagesAreInstalled = pubspec.dependencies.keys
+        .every((dependencyName) => packageConfig[dependencyName] != null);
```
hampuslavin, line 290: Is there a case for checking `dev_dependencies` as well?

Isakdl: As of today, no since we only care about modules here and they must be normal dependencies.

Isakdl (changes requested):
Looking good! Many nice modifications here! :) 

Small little suggestion on code structure.

tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -217,15 +210,14 @@ class GeneratorConfig {
...
+      throw const ServerpodProjectNotFoundException(
         'Failed to load client pubspec.yaml. If you are using a none default '
         'path it has to be specified in the config/generator.yaml file!',
       );
-      throw const ServerpodProjectNotFoundException(
-        'Failed to load client pubspec.yaml',
-      );
     }
 
+    await _warnIfDependenciesAreNotInstalled(serverRootDir, pubspec);
```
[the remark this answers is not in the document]
hampuslavin: Good point! Will fix 👍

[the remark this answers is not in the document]
hampuslavin: 5bc91a7

Isakdl (approved):
LGTM!

PR #2638, opened 2024-08-20
fix: Make it possible to restart Serverpod server.

SandPod:
Fixes two issues that prevented us from restarting a Serverpod server.

### Issue 1
In the `database_connection` we used to create a copy of the database pool and store it locally. This caused issue when we stopped the existing database pool through the `DatabasePoolManager` since the copied reference would then be preserved towards stopped instance.

To fix the issue we now always fetch the pool from the pool manager.

### Issue 2
As part of the server start we set the _httpServer field. This was a `late final` field, but since it was assigned every time we started the server we would get a `LateInitializationError` on the second assignment.

To fix the `final` declaration was removed since the field in essence is not final.

## Pre-launch Checklist

## Breaking changes

None - the changes should be completely backwards compatible.

### DIFF
--- a/tests/serverpod_test_server/test_integration/server_restart_test.dart
+++ b/tests/serverpod_test_server/test_integration/server_restart_test.dart
@@ -17,18 +17,17 @@ void main() {
   });
 
   test(
-      'Given running Serverpod server when it is shutdown and started then it can be successfully started again.',
+      'Given a running Serverpod server when it is shutdown and restarted then it can be successfully started again.',
       () async {
     await serverpod.start();
 
     await serverpod.shutdown(exitProcess: false);
-    var start = serverpod.start();
 
-    await expectLater(start, completes);
+    await expectLater(serverpod.start(), completes);
   });
 
   test(
-      'Given running Serverpod server when it is shutdown and started then database request can be made.',
+      'Given a running Serverpod server when it is shutdown and started then database request can be made.',
       () async {
     await serverpod.start();
 
@@ -40,16 +39,15 @@ void main() {
   });
 
   test(
-      'Given running Serverpod server when it is shutdown and started then no error is written to stderr.',
+      'Given a running Serverpod server when it is shutdown and started then no error is written to stderr.',
       () async {
     var record = MockStdout();
     await IOOverrides.runZoned(() async {
       await serverpod.start();
 
       await serverpod.shutdown(exitProcess: false);
-      var start = serverpod.start();
 
-      await expectLater(start, completes);
+      await expectLater(serverpod.start(), completes);
     }, stderr: () => record);
 
     expect(record.output, isEmpty);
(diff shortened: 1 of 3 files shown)


tests/serverpod_test_server/test_integration/server_restart_test.dart
```diff
@@ -0,0 +1,57 @@
...
+  setUp(() {
+    serverpod = IntegrationTestServer.create();
+  });
+
+  tearDown(() async {
+    await serverpod.shutdown(exitProcess: false);
+  });
+
+  test(
+      'Given running Serverpod server when it is shutdown and started then it can be successfully started again.',
```
hampuslavin, line 20: nit
```suggestion
      'Given a running Serverpod server when it is shutdown and restarted then it can be successfully started again.',
```

tests/serverpod_test_server/test_integration/server_restart_test.dart
```diff
@@ -0,0 +1,57 @@
...
+      () async {
+    await serverpod.start();
+
+    await serverpod.shutdown(exitProcess: false);
+    var start = serverpod.start();
+
+    await expectLater(start, completes);
```
hampuslavin, line 27: nit: could consider inlining for brevity, just like you do for `session.db.testConnection()` below
```suggestion
    await expectLater(serverpod.start(), completes);
```

tests/serverpod_test_server/test_integration/server_restart_test.dart
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  test(
+      'Given running Serverpod server when it is shutdown and started then database request can be made.',
```
hampuslavin, line 31: nit
```suggestion
      'Given a running Serverpod server when it is shutdown and started then a database request can be made.',
```

tests/serverpod_test_server/test_integration/server_restart_test.dart
```diff
@@ -0,0 +1,57 @@
...
+
+    await serverpod.shutdown(exitProcess: false);
+    await serverpod.start();
+
+    var session = await serverpod.createSession();
+    await expectLater(session.db.testConnection(), completion(true));
+  });
+
+  test(
+      'Given running Serverpod server when it is shutdown and started then no error is written to stderr.',
```
hampuslavin, line 43: nit
```suggestion
      'Given a running Serverpod server when it is shutdown and started then no error is written to stderr.',
```

tests/serverpod_test_server/test_integration/server_restart_test.dart
```diff
@@ -0,0 +1,57 @@
...
+      () async {
+    var record = MockStdout();
+    await IOOverrides.runZoned(() async {
+      await serverpod.start();
+
+      await serverpod.shutdown(exitProcess: false);
+      var start = serverpod.start();
+
+      await expectLater(start, completes);
```
hampuslavin, line 52: nit: same as previous comment

hampuslavin (approved):
Nice! 🙌  Just left som nits

Isakdl (approved):
LGTM 🚀

PR #2644, opened 2024-08-20
fix: Allow Method Streams to reconnect on websocket closed exception.

SandPod:
Fixes an issue where the ClientMethodStreamManager would not correctly reset its internal WebSocket connection state if the WebSocket connection was closed from the Server.

Draft PR - Depends on: https://github.com/serverpod/serverpod/pull/2638

## Pre-launch Checklist

## Breaking changes

None - simple bugfix.


packages/serverpod_client/lib/src/client_method_stream_manager.dart
```diff
@@ -495,9 +495,9 @@ final class ClientMethodStreamManager {
       await webSocket.sink.close();
     } finally {
       /// Close any still open streams with an exception.
```
hampuslavin, line 497: Should the comment be moved as well? Or remove it if it's clear enough by the code?

hampuslavin (approved):
🌟

packages/serverpod_client/lib/src/client_method_stream_manager.dart
SandPod: Good suggestion! Moved it 👍

Issue #1140, opened 2023-07-12
Default value support in protocol / generated code / database


Isakdl:
**Feature: default values.**

Allow the programmer to define a default value for a field in the protocol. The default value should propagate to the SQL table and the generated dart class. These defaults are in most cases constant and does not give a dynamic value.

```yaml
class: Example
table: example
fields:
  toggle: boolean, default=true
  score: int, default=0
```

The default "default" value should be null if field is nullable, if not it is required to be set. In case of List and Map I think it makes sense to initialize them as empty List/Map in the field is not nullable.

We should at least support the following types for default values

- boolean
- int
- double
- String
- DateTime
- Uuid
- Duration

__DateTime__, the default value has to be an ISO8001 string OR `now()` where it would map to the same function in postgresql.
__Uuid__, the default value has to follow the uuid syntax such as the nil uuid `00000000-0000-0000-0000-000000000000`.

For other types we should not support default values (such as custom objects). We should allow defaults for any database type we have implemented.

**Keywords**
Configuring the default value could be done using the following keywords:
- **defualt** - Sets the default value for both the Model and the storage.
- **defaultModel**. - Sets the default value for only the Model.
- **defaultPersist** - Sets the default value for only the storage.

MarkOSullivan94:
Agree with all of this and for me this is one feature I'll be very excited to see added in the 1.2 release

lukehutch:
I definitely need this. I have a method that creates a default instance of a large protocol class, setting fields to `''`, `0`, or `false`. Every time I modify the protocol file, this class breaks, and I have to go edit it.

alguintu:
Curious how you classify default values as a "nice to have" feature instead of an essential functionality in a migration/orm package? Most if not all ORMs I have encountered have this feature to ensure data consistency and simplify api calls for new entitities.

Isakdl:
@alguintu it is classified as nice to have for the 1.2 update. This feature is something we want to build but we have to make a cut somewhere to get the release out. I agree that this is an important feature.

lukehutch:
Hopefully when default values are added, this migration issue can be avoided:

> One or more columns are added to table "social_post" which cannot be added in a table migration. The complete table will be deleted and recreated.

Columns should always be able to be added, without deleting the database, but I assume the reason this is not possible right now is because default values are not supported. The error message should actually explain that, even when default value support is added.

lukehutch:
Other issues that would be fixed if we had default values:

[...]

Issue #236, opened 2022-07-14
default value


lastmeta:
in the yaml file can I specify a default value for the field?

```
class: Address
table: address

fields:  
  address: String
  used: bool
    #default: true  #something like this?
```

fischerscode:
@vlidholt do you envision any special way how this should be implemented? Sounds like a nice PR...

ramyak-mehra:
we can handle the default value on the sql side or dart side

vlidholt:
It would be a cool addition, but I think the implementation needs to be thought through. Handling the default value on the Dart side is probably best, but it may only be possible to do primitive types (other serializable objects would be tricky). Also, what should the syntax look like. Needs to feel clean and be backwards compatible.

fischerscode:
I think it would be best to provide the serialized version of the data in the yaml file. This could be used to set the default in the database, and a dart default could be constructed, too, by using the `SerializationManager`.

I'm currently working on a package that helps specify and parse yaml based configuration files. I think it'd be the perfect fit for specifying the class definitions. (Had this as a usecase in mind)

SandPod:
Closing this as duplicate for the more specified issue here https://github.com/serverpod/serverpod/issues/1140.

PS. This is being implemend right now 🙌

PR #2649, opened 2024-08-21
feat: Implement Default Values for Duration field types in Models 

klkucaj:
### Overview

This is a follow-up to #2383.  
This PR introduces support for default values in the `Duration` field type.

### Default Types

There are three types of default values that can be specified:

- **defaultDatabase**: Adds a default value to the database table column.
- **defaultModel**: Adds a default value to the model field.
- **default**: Adds default values to both the model field and the database table column.

### Default Value Options for `Duration` Fields

Default values can be assigned to `Duration` fields as follows:

#### Human-Readable Duration Strings:

- **Dart Side**: A valid `Duration` string can be specified in a human-readable format, such as `1d 2h 10min 30s 100ms`.
- **Database Side**: The corresponding `Duration` value will be stored in the database column as the total number of milliseconds, e.g., `94230100`.

### Related Issues
- #1140
- #236

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
@@ -2,6 +2,7 @@ import 'package:code_builder/code_builder.dart';
 import 'package:recase/recase.dart';
 import 'package:serverpod_cli/analyzer.dart';
 import 'package:serverpod_cli/src/analyzer/models/definitions.dart';
+import 'package:serverpod_cli/src/analyzer/models/utils/duration_utils.dart';
 import 'package:serverpod_cli/src/generator/dart/library_generators/class_generators/repository_classes.dart';
 import 'package:serverpod_cli/src/generator/dart/library_generators/util/class_generators_util.dart';
 import 'package:serverpod_cli/src/generator/keywords.dart';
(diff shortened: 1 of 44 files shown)


tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -1064,6 +1065,15 @@ class SerializableModelLibraryGenerator {
...
             .call([CodeExpression(Code(defaultValue))]).code;
+      case DefaultValueAllowedType.duration:
+        Duration parsedDuration = parseDuration(defaultValue);
+        return refer(field.type.className).call([], {
+          'days': literalNum(parsedDuration.days),
+          'hours': literalNum(parsedDuration.hours),
+          'minutes': literalNum(parsedDuration.minutes),
+          'seconds': literalNum(parsedDuration.seconds),
+          'milliseconds': literalNum(parsedDuration.milliseconds),
+        }).code;
```
SandPod, line 1076: Question: Are there any tests that validate that this produces the correct result when hours has a value higher than 24? If not, I think we should add tests that cover all of these conversions from the code generator perspective.

SandPod: These could probably be added to the code generator tests.

SandPod (commented):
Nice and quick work!

Single comment regarding testing the extension for breaking Duration apart into hours, minutes, seconds, and milliseconds.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
klkucaj: Done.

Issue #2657, opened 2024-08-23
Protocol::getClassNameForObject is missing delegate to Serverpod protocol file.


SandPod:
**Describe the bug**
In the project generated protocols files `getClassNameForObject` method, we never check if the class name can be retrieved from the Serverpod generated protocol file.

This makes it impossible to serialize Serverpod defined models from a project.

**To Reproduce**
1. Create an instance of a Serverpod defined model, for example `CluserServerInfo('Hello world');`.
2. Try to Serialize it using a project-generated protocol file `Protocol().`
3. Observe exception

```dart
import 'package:serverpod/protocol.dart' as serverpod;
import 'package:serverpod_test_server/src/generated/protocol.dart';
import 'package:test/test.dart';

void main() {
  test(
      'Given a Serverpod defined model when encoding it with type then it is encoded',
      () {
    var protocol = Protocol();
    var serverpodDefinedModel =
        serverpod.ClusterServerInfo(serverId: 'Hello World');
    var encoded = protocol.encodeWithType(serverpodDefinedModel);

    expect(encoded, isA<String>());
  });
}

```

**Expected behavior**
Serialization of Serverpod defined models should be possible when encoding with type.

**Screenshots**

**Serverpod versions**
 -  _CLI Version:_ 2.0.2.
 - _Library version:_ 2.0.2.

KristijanZic:
I still have this error on the Serverpod version: 2.1.0-beta.2

PR #2659, opened 2024-08-23
fix: Allow Serverpod defined models to be encoded and decoded with type.

SandPod:
The generated code on the Server did not properly delegate to the Serverpod Protocol file when for calls to `getClassNameForObject` and `deserialzieByClassName`.

This made it impossible to encode and decode Serverpod-defined models.

Fixes: #2657

### Additional information

An important note here is that we only support encoding and decoding Serverpod defined models on the server side.
The client does not have access to the Serverpod generated client Protocol (referred to as serverpod_service_client).

Therefore no delegated call has been added to the client code.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
@@ -187,6 +187,12 @@ class LibraryGenerator {
           for (var classInfo in models)
             Code.scope((a) =>
                 'if(data is ${a(refer(classInfo.className, classInfo.fileRef()))}) {return \'${classInfo.className}\';}'),
+          if (config.name != 'serverpod' && serverCode)
+            Block.of([
+              Code.scope((a) =>
+                  'className = ${a(refer('Protocol', serverpodProtocolUrl(serverCode)))}().getClassNameForObject(data);'),
+              Code('if(className != null){return \'serverpod.\$className\';}'),
+            ]),
           for (var module in config.modules)
             Block.of([
               Code.scope((a) =>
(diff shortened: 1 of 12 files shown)


tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -212,6 +218,14 @@ class LibraryGenerator {
             Code.scope((a) =>
                 'if(data[\'className\'] == \'${classInfo.className}\'){'
                 'return deserialize<${a(refer(classInfo.className, classInfo.fileRef()))}>(data[\'data\']);}'),
+          if (config.name != 'serverpod' && serverCode)
+            Block.of([
+              Code('if(data[\'className\'].startsWith(\'serverpod.\')){'
+                  'data[\'className\'] = data[\'className\'].substring(${'serverpod'.length + 1});'),
+              Code.scope((a) =>
+                  'return ${a(refer('Protocol', serverpodProtocolUrl(serverCode)))}().deserializeByClassName(data);'),
+              const Code('}'),
```
hampuslavin: General question on this code. We are mixing a lot of `Code` blocks together with actual builder patterns (for example `refer('variable').property(myMethod).call([])` vs `Code('variable.myMethod()')`). When are you choosing which? My impression so far is that `Code` blocks easily get a bit messy and make code reuse harder. Mixing them I also feel makes it a bit harder to read because of the inconsistent building paradigm (template vs builder).

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -187,6 +187,12 @@ class LibraryGenerator {
           for (var classInfo in models)
             Code.scope((a) =>
                 'if(data is ${a(refer(classInfo.className, classInfo.fileRef()))}) {return \'${classInfo.className}\';}'),
+          if (config.name != 'serverpod' && serverCode)
```
hampuslavin, line 190: Obviously this is correct, but just for my understanding. What is this fix doing? Why are we checking `config.name != 'serverpod'`?

hampuslavin (approved):
🌟

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
SandPod: We don't really have a clear strategy here honestly.
Often we just grab what is closest in context and then apply it to our specific scenario.

As far as I understand there are a couple of limitations with the builder pattern that forces us to use code blocks. For example it is not possible to write if statements using the builder pattern, so we are forced to use this middleway.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
SandPod: This is to check that we are not adding this to the Serverpod Server.
Since this would introduce an infinite loop with delegation to itself.

PR #2672, opened 2024-08-26
Fix: Move the auth key from the body to the header for the HTTP requests of endpoint methods.

### LINKED ISSUE #663: Move key-token to header instead of body
As discussed here #657 it may be better to pass the key token for signed in users in the header instead of a URL query parameter.

**Additional context**
This is not a large security risk as the URL isn't typically visible or stored in history or cache. It's possible both a query parameter and header can be used simultaneously, making this change backwards compatible.

christerswahn:
Moves the auth key from the body to the header for the HTTP requests of endpoint methods.

Closes #663 

## Pre-launch Checklist

## Breaking changes

n/a


tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
```diff
@@ -0,0 +1,77 @@
...
+        () async {
+      var key = 'username-4711:password-4711';
+      authKeyManager.put(key);
+
+      var reflectedKey = await client.reflection.reflectAuthenticationKey();
+      expect(reflectedKey, key);
+
+      var receivedToken = await tokenInspectionCompleter.future;
+      expect(receivedToken, key);
+    });
```
SandPod: Suggestion: Break out to two separate tests.

Why always try to validate a single case in each test. For scenarios in e2e cases this is not always possible, but where it is possible we always aim to keep a test as narrow and focused as possible.

In this test, we both validate what the endpoint received (from the session) and what the authentication handler received. This is not correctly reflected in the description. Please break this out into two different tests.

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
```diff
@@ -0,0 +1,77 @@
...
+      expect(isWrappedAuthValue(reflectedHeader.first), true);
+
+      var scheme = reflectedHeader.first.split(' ')[0];
+      expect(scheme, 'Basic');
+
+      var unwrappedKey = unwrapAuthValue(reflectedHeader.first);
+      expect(unwrappedKey, key);
+
+      var receivedToken = await tokenInspectionCompleter.future;
+      expect(receivedToken, key);
```
SandPod, line 74: Suggestion: Break this out into a separate tests.

As far as I understand, the previous expects all make sure that the header follows the Basic scheme.
The highlighted expects instead validate that unwrapping the header successfully reproduces the key.
This seems like an unrelated test given the test description.

tests/serverpod_test_server/test_e2e/websocket_auth_test.dart
```diff
@@ -0,0 +1,65 @@
...
+        serverUrl,
+        authenticationKeyManager: authKeyManager,
+      );
+    });
+
+    tearDown(() async {
+      client.close();
+    });
+
+    test('Access restricted endpoint without authentication', () async {
```
SandPod, line 24: Suggestion: Use `Given, When, Then` pattern for test descriptions.

tests/serverpod_test_server/test_e2e/websocket_auth_test.dart
```diff
@@ -0,0 +1,65 @@
...
+      await client.openStreamingConnection(
+        disconnectOnLostInternetConnection: false,
+      );
+
+      // This should be ignored by the server as user isn't authenticated.
+      await client.signInRequired.sendStreamMessage(SimpleData(num: 666));
+      var streamingStream = client.signInRequired.stream.timeout(
+        const Duration(seconds: 2),
+      );
+      var receivedDataF = streamingStream.single;
```
SandPod, line 34: Suggestion: Rename `receivedDataF` to `receivedData`.

I'm guessing the F is a typo? Or does it mean something? If so, please type it out.

tests/serverpod_test_server/test_e2e/websocket_auth_test.dart
```diff
@@ -0,0 +1,65 @@
...
+
+    test('Access restricted endpoint without authentication', () async {
+      await client.openStreamingConnection(
+        disconnectOnLostInternetConnection: false,
+      );
+
+      // This should be ignored by the server as user isn't authenticated.
+      await client.signInRequired.sendStreamMessage(SimpleData(num: 666));
+      var streamingStream = client.signInRequired.stream.timeout(
+        const Duration(seconds: 2),
```
SandPod, line 32: Discussion(Knowledge sharing): We always try to stay away from making tests depend on timeouts. The reason for this is that these make tests brittle (dependent on execution time) and also might add an unnecessary delay to a test.

In this case, would it be enough to send another message to a different endpoint and validate that this response was received but the one that requires authentication was not?

SandPod: There are cases where we have to wait on a timeout. But in general we try to stay away from them.

SandPod: Same discussion for the other test.

tests/serverpod_test_server/test_e2e/websocket_auth_test.dart
SandPod: Same suggestion for the other test.

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
```diff
@@ -73,5 +73,61 @@ void main() async {
...
+      tokenInspectionCompleter = Completer();
+      var reflectedHeader =
+          await client.reflection.caller.callServerEndpoint<List<String>?>(
+        'reflection',
+        'reflectHttpHeader',
+        {
+          'headerName': 'authorization',
+          'auth': key,
+        },
+      );
```
SandPod, line 96: Suggestion: Use a pure HTTP client to make this request.

Given that changes are made to `callServerEndpoint`, the format of this call might be mutated in such a way that it no longer reflects the "old" approach.
A safer approach would be to bypass our client generated code and simply pass in the data structure we expect to be correctly parsed by the server.

packages/serverpod_client/lib/src/serverpod_client_browser.dart
```diff
@@ -32,14 +32,16 @@ class ServerpodClientRequestDelegateImpl
...
-          .post(
-            url,
-            body: body,
-          )
-          .timeout(connectionTimeout);
+      var response = await _httpClient.post(
+        url,
+        body: body,
+        headers: {
+          if (authenticationValue != null) 'authorization': authenticationValue,
```
hampuslavin, line 42: Question: While HTTP headers are not case sensitive, I think it's most common to use capitalized values (i.e. `Authorization`). Could be worth considering to be as similar to the standard case as possible?

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
```diff
@@ -73,5 +73,61 @@ void main() async {
...
     });
+
+    test(
+        'then old clients passing keys in body should still work with new server code',
+        () async {
+      var key = 'username-4713:password-4713';
+
+      // Intentionally clearing the authKeyManager so that auth is not passed in header.
+      await authKeyManager.remove();
+
```
SandPod, line 84: Suggestion: Add this as a `tearDown` instead.

Having a `tearDown` in the test implementation is a strong signal that other tests are leaky. Please ensure that this is not required or place it as a common tearDown.

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
```diff
@@ -73,5 +73,61 @@ void main() async {
       var receivedToken = await tokenInspectionCompleter.future;
       expect(receivedToken, key);
     });
+
+    test(
+        'then old clients passing keys in body should still work with new server code',
```
SandPod, line 78: Suggestion: Break apart the different test cases in this test.

We are validating multiple things in this test. 
Please break these scenarios apart so that it is clear what is failing.

packages/serverpod_serialization/lib/src/auth_encoding.dart
```diff
@@ -0,0 +1,62 @@
...
+/// It consists of:
+/// - \t (tab character)
+/// - \x20-\x7E (printable ASCII characters, including space)
+/// - \x80-\xFF (extended ASCII characters)
+const _httpHeaderValueRegExpStr = r'[\t\x20-\x7E\x80-\xFF]*';
+
+/// The regular expression for the format of the HTTP "authorization" header value,
+/// including starting with an authentication scheme name.
+final _authValueRegExp =
+    RegExp('^$_authSchemeRegExpStr $_httpHeaderValueRegExpStr\$');
```
hampuslavin, line 27: Question: Just to double check, why is the dollar sign at the end escaped? I would have expected it to to not be escaped to signify end of string match, but perhaps I'm misunderstanding this?

packages/serverpod_serialization/lib/src/auth_encoding.dart
```diff
@@ -0,0 +1,62 @@
...
+
+/// Returns true if the provided value is a valid HTTP "authorization" header value
+/// (which includes starting with an authentication scheme name).
+bool isValidAuthHeaderValue(String value) => _authValueRegExp.hasMatch(value);
+
+/// The regular expression for the format of Serverpod-wrapped auth keys.
+final _wrappedAuthValueRegExp =
+    RegExp('^$serverpodDefaultAuthSchemeName $_base64RegExpStr\$');
```
hampuslavin, line 35: Same as previous

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
```diff
@@ -73,5 +73,61 @@ void main() async {
...
+          'reflection',
+          'reflectHttpHeader',
+          {
+            'headerName': 'authorization',
+          },
+        );
+      } catch (e) {
+        if (e is ServerpodClientException) {
+          statusCode = e.statusCode;
+        }
```
SandPod, line 114: Suggestion: Use the `isA` matcher with the `having` property instead of this if statement.

This ensures the test reports what type was returned if our expected value was not received.

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
```diff
@@ -73,5 +73,61 @@ void main() async {
...
+      }
+      expect(statusCode, equals(401));
```
SandPod, line 116: Suggestion: Utalize`await expectLater` here to await the future with the correct value directly.

This can be used to catch any throws as well using the `throwsA` matcher.

tests/serverpod_test_server/test_e2e/auth_core_test.dart
```diff
@@ -0,0 +1,61 @@
+import 'dart:convert';
+
+import 'package:http/http.dart' as http;
+import 'package:serverpod_test_client/serverpod_test_client.dart';
+import 'package:serverpod_test_server/test_util/config.dart';
+import 'package:serverpod_test_server/test_util/test_key_manager.dart';
+import 'package:test/expect.dart';
+import 'package:test/scaffolding.dart';
+
+void main() {
```
SandPod, line 10: Question: How are these tests differentiated between the tests in: `tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart`?

packages/serverpod_serialization/lib/src/auth_encoding.dart
```diff
@@ -0,0 +1,62 @@
...
+
+/// Returns true if the provided value is a Serverpod-wrapped auth key.
+bool isWrappedAuthValue(String value) =>
+    _wrappedAuthValueRegExp.hasMatch(value);
+
+/// Returns a value that is compliant with the HTTP auth header format
+/// by encoding and wrapping the provided auth key as a Basic auth value.
+String wrapAuthValue(String key) {
```
hampuslavin, line 43: Question: Since this is a header compliant format, I'm wondering if it would make sense to name this something very explicit like `createHttpAuthHeader`. Right now it's a bit more opaque, but perhaps there's some benefit to that?

The same reasoning applies to `isWrappedAuthValue`, which is (as far as I can tell?) equivalent to `isHttpAuthHeader` (or maybe `isDefaultHttpAuthHeader` specifically)?

tests/serverpod_test_server/lib/src/endpoints/reflection.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:serverpod/serverpod.dart';
+
+/// This class is meant for reflecting the received headers, auth keys in endpoint invocations.
+class ReflectionEndpoint extends Endpoint {
```
hampuslavin, line 4: Nit: Since reflection is a common concept in programming (although not supported by Dart), maybe we could name this something more explicit like `ReflectHttpHeadersEndpoint` to avoid any confusion. Maybe a bit of stretch, but I think these small things add up and make the code base easier to navigate.

tests/serverpod_test_server/test_e2e/websocket_auth_test.dart
hampuslavin, line 34: Nit: is this a typo or am I missing something?
```suggestion
      var receivedData = streamingStream.single;
```

tests/serverpod_test_server/test_e2e/websocket_auth_test.dart
hampuslavin, line 24: Suggestion: follow the given/when/then paradigm (applies to all files)

hampuslavin (commented):
Nice work! 👏 Just went through because I was curious and dropped some superficial comments along the way.

packages/serverpod_serialization/test/auth_encoding_test.dart
```diff
@@ -0,0 +1,156 @@
...
+      'key11': [1, 2],
+      'key12': null,
+    }),
+  ];
+
+  group('isValidAuthHeaderValue()', () {
+    for (var key in [
+      ...standardBasicAuthKeys,
+      ...standardNonBasicAuthKeys,
+    ]) {
```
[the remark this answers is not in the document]
hampuslavin: I also think this is more or less supported in different language stacks. The Dart test package sadly doesn't have any built-in support for parameterized tests. We could add our own helper along the way if we had a strong use case. As you mention, you can then add a human readable description for each parameter (not saying we should do it here.
```dart
group("formatDay should format dates correctly:", () {
  var inputsToExpected = {
    DateTime(2018, 11, 01): "Thu 1",
    ...
    DateTime(2018, 11, 07): "Wed 7",
    DateTime(2018, 11, 30): "Fri 30",
  };
  inputsToExpected.forEach((input, expected) {
    test("$input -> $expected", () {
      expect(formatDay(input), expected);
    });
  });
});
```

packages/serverpod_client/lib/src/serverpod_client_browser.dart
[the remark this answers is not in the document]
christerswahn: Not in the browser case, unfortunately. Same gap as with the HTTP status codes.

christerswahn: Good point!

vlidholt (commented):
Great work on this! 🤩 Left a few comments.

packages/serverpod_serialization/lib/src/auth_encoding.dart
christerswahn: It's a dart string escape to prevent the dart string interpolation; yes it represents end-of-string match.

packages/serverpod_serialization/lib/src/auth_encoding.dart
```diff
@@ -0,0 +1,62 @@
+/// Serverpod's default authentication key encoding and decoding.
+library serverpod_serialization.auth_encoding;
```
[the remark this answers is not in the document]
christerswahn: It is needed in order to make a doc comment on the file (aka dart library). I looked it up, it is only needed in this case...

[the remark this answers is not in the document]
christerswahn: Regarding class: Class "normally" denotes a class of objects, while this is a (small) function library. The library as such provides a namespace so in my book that is sufficient?

packages/serverpod_serialization/lib/src/auth_encoding.dart
```diff
@@ -0,0 +1,62 @@
...
+  var encodedKey = base64.encode(utf8.encode(key));
+  return '$serverpodDefaultAuthSchemeName $encodedKey';
+}
+
+/// Returns the auth key from an auth value that has potentially been wrapped.
+/// This operation is the inverse of [wrapAuthValue].
+/// If null is provided, null is returned.
+String? unwrapAuthValue(String? authValue) {
+  if (authValue == null) return null;
+  if (isWrappedAuthValue(authValue)) {
```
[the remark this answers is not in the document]
christerswahn: Re Basic: Hypothetically Serverpod could choose to have a different (but still valid) default auth scheme in the future, e.g. Bearer. That's why it is a constant - it should not be hardcoded on more than one place.

We could of course make it internal with an underscore, maybe that is better?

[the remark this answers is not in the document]
christerswahn: Re returning as-is: This is crucial in order for old clients / stored keys to continue to work.

packages/serverpod_serialization/lib/src/auth_encoding.dart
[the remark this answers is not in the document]
christerswahn: Hypothetically Serverpod could choose to have a different (but still valid) default auth scheme in the future, e.g. Bearer. That's why it is a constant - it should not be hardcoded in more than one place.

We could of course make it internal with an underscore, maybe that is better? But this way users could actually look it up and test on it, if they want to be able to inspect it programmatically at some point.

packages/serverpod_serialization/lib/src/auth_encoding.dart
```diff
@@ -0,0 +1,62 @@
...
+/// Returns true if the provided value is a valid HTTP "authorization" header value
+/// (which includes starting with an authentication scheme name).
+bool isValidAuthHeaderValue(String value) => _authValueRegExp.hasMatch(value);
+
+/// The regular expression for the format of Serverpod-wrapped auth keys.
+final _wrappedAuthValueRegExp =
+    RegExp('^$serverpodDefaultAuthSchemeName $_base64RegExpStr\$');
+
+/// Returns true if the provided value is a Serverpod-wrapped auth key.
+bool isWrappedAuthValue(String value) =>
```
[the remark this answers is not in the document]
christerswahn: This is needed in order for old clients / stored keys to continue to work, as well as new users implementing other schemes e.g. Bearer.

packages/serverpod_serialization/lib/src/auth_encoding.dart
hampuslavin: Ohh alright, that's strange given that nothing follows the dollar sign? 🤔 Anyway, then I see, thanks!

packages/serverpod_serialization/lib/src/auth_encoding.dart
christerswahn: This code is designed to also support the case where users implement their own authorization approach, potentially using some other HTTP auth scheme, e.g. Bearer. It is only in the Basic case we actually perform a modification (wrapping).

packages/serverpod_serialization/test/auth_encoding_test.dart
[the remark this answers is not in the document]
christerswahn: I had 2 discussions with Isak on this :)
Yes there is a point with unrolling, but for this specific kind of unit test my view would be that it is much easier to get an overview of the test, maintain the test, and get a reasonable understanding of the completeness of the test cases to list them like this. Also, otherwise one would wonder if and if so what is different in the test code of dozens of tests.

tests/serverpod_test_server/lib/src/endpoints/reflection.dart
christerswahn: I propose to use Alex' suggestion RequestReflectionEndpoint. The intent is not to have only HTTP specific things but generally for endpoint methods.

tests/serverpod_test_server/test_e2e/auth_core_test.dart
christerswahn: They are logically the same! However this one is on the HTTP request level (bypassing all client code) and the other actually uses the client code but bypasses the top layer in order to force set the desired request shape. 

The other one was the first one I wrote and I kept it for now - should we remove it? As was mentioned in a discussion, that test could be imaged to break upon some future client code change.

Sorry should have explained this in the PR.

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
christerswahn: Yes! See auth_core_test.dart

packages/serverpod/lib/src/server/session.dart
```diff
@@ -413,7 +413,9 @@ class StreamingSession extends Session {
     this.queryParameters = queryParameters;
 
     // Get the the authentication key, if any
-    _authenticationKey = queryParameters['auth'];
+    // TODO: This is insecure, the auth key should be passed in a message after stream connection instead.
```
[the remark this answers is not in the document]
christerswahn: Agree!

packages/serverpod_serialization/test/auth_encoding_test.dart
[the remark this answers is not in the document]
hampuslavin: Looking closer at this code I agree with what @christerswahn said above. I think simple validation functions like this are one of the prime use cases for parameterized tests. In my opinion, unrolling these into 6 different test cases that does the same thing will make it harder to overview what is tested, whereas now it's super quick to see what cases are covered. As always there are some downsides but in this case I think the benefits outweigh them.

tests/serverpod_test_server/lib/src/endpoints/reflection.dart
hampuslavin: I still think the use of `Reflection` like this is ambiguous and I would avoid it completely. May I suggest `ReflectRequestEndpoint` or perhaps `EchoRequestEndpoint` , which is more unambiguous? (cc @SandPod since you also commented on this)

tests/serverpod_test_server/lib/src/endpoints/reflection.dart
```diff
@@ -0,0 +1,25 @@
...
```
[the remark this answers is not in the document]
hampuslavin: Related discussion: https://github.com/serverpod/serverpod/pull/2672#discussion_r1735634361

packages/serverpod_serialization/lib/src/auth_encoding.dart
hampuslavin: Ah, I see! But I'm still not quite following why we have to be opaque/generic with the naming and not state that we're working with HTTP headers? Maybe I'm missing the bigger picture here 😊

tests/serverpod_test_server/lib/src/endpoints/reflection.dart
```diff
@@ -0,0 +1,25 @@
...
+  @override
+  bool get requireLogin => true;
+
+  /// Reflects the authentication key of the session.
+  /// Returns null if the key is not set.
+  Future<String?> reflectAuthenticationKey(Session session) async {
+    var ms = session as MethodCallSession;
+    ms.httpRequest.headers.forEach((key, value) {
+      print('Header: $key: $value');
```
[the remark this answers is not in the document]
christerswahn: Yes, forgot to remove that!

tests/serverpod_test_server/test_e2e/auth_core_test.dart
christerswahn: I've removed the other test.

packages/serverpod_serialization/lib/src/auth_encoding.dart
christerswahn: Ok I've made it a bit more explicit with wrapAuthHeaderValue!

tests/serverpod_test_server/test_e2e/websocket_auth_test.dart
christerswahn: I agree with you, timeouts are un unwanted externality here. This was the most direct way of testing this that I found. If I merely call another endpoint, I'm not directly verifying that *this* endpoint is reacting as I expect.

But thinking about it I can send two messages, one before auth and one after. If the first element I receive back is the first I sent unauthenticated, then it is an error.

packages/serverpod_serialization/lib/src/websocket_messages.dart
```diff
@@ -180,7 +180,8 @@ class OpenMethodStreamCommand extends WebSocketMessage {
         args = data[WebSocketMessageDataKey.args],
         connectionId = UuidValueJsonExtension.fromJson(
             data[WebSocketMessageDataKey.connectionId]),
-        authentication = data[WebSocketMessageDataKey.authentication],
+        authentication =
+            unwrapAuthValue(data[WebSocketMessageDataKey.authentication]),
```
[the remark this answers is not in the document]
christerswahn: The buildMessage method gets its authentication token via connectionDetails.authenticationProvider, which in turn gets it from authenticationKeyManager.getHeaderValue().

The root problem seems to me to be that the code for creating / serializing / deserializing is somewhat spread out, AND that we're using three different interfaces for this handling defined in different places (AuthenticationKeyManager, AuthenticationKeyProvider, AuthenticationHandler) AND don't have specific types for authentication keys / tokens to clarify what we're handling. 

In order to not make the current code more complicated before the bigger auth work I'd propose to have it like this, but add a code comment, and then rework this together with the authentication overhaul. WDYT?

tests/serverpod_test_server/lib/src/endpoints/reflection.dart
SandPod: `Echo` seems like a good suggestion in that case? Maybe even `EchoRequestAttributeEndpoint`?
Good point about mixing this up with general reflections in programming.

packages/serverpod_serialization/lib/src/auth_encoding.dart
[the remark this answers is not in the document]
vlidholt: When we add a new scheme, we can add a new method. I think this would be better. But let's discuss today.

packages/serverpod_serialization/lib/src/auth_encoding.dart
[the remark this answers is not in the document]
vlidholt: I think the correct behaviour would be to check if we are using `Basic` (potentially just checking the beginning of the string). If we are using `Basic` the encoding is incorrect, we should throw a 400 invalid request.

We can discuss how to handle future authentication types. But `Basic` should never fall back to returning the full string.

I may be holding things wrong here, but let's discuss.

tests/serverpod_test_server/test_e2e/auth_core_test.dart
```diff
@@ -0,0 +1,61 @@
...
+      assert(response.success, 'Authentication setup failed');
+      authKey = '${response.keyId}:${response.key}';
+    });
+
+    tearDownAll(() async {
+      client.close();
+    });
+
+    test(
+        'then a call to an authorizaed endpoint method with old style auth key should succeed',
```
SandPod, line 33: Suggestion(NIT): This test is missing the `when` statement.

What is the "action" here and for the next test?

tests/serverpod_test_server/test_e2e/websocket_auth_test.dart
```diff
@@ -0,0 +1,65 @@
...
+        serverUrl,
+        authenticationKeyManager: authKeyManager,
+      );
+    });
+
+    tearDown(() async {
+      client.close();
+    });
+
+    test('then access restricted endpoint without authentication', () async {
```
SandPod, line 24: Suggestion(NIT): This test is missing it's `when` statement.

Same for the next test.

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
```diff
@@ -0,0 +1,161 @@
...
+  late Completer<String> tokenInspectionCompleter;
+  Future<AuthenticationInfo?> authenticationHandler(
+    Session session,
+    String token,
+  ) {
+    tokenInspectionCompleter.complete(token);
+    return Future.value(AuthenticationInfo(1, {}));
+  }
+
+  group('When passing auth key in HTTP header format', () {
```
SandPod, line 19: Suggestion: These tests are missing their `Given` statement.

For example `Given auth key in HTTP header format When calling endpoint then endpoint method should receive plain auth key (i.e. in original format)`.

SandPod (changes requested):
Almost there, still a couple of Suggestions and questions.

packages/serverpod_serialization/test/auth_encoding_test.dart
```diff
@@ -0,0 +1,205 @@
...
+        },
+        'key10': [],
+        'key11': [1, 2],
+        'key12': null,
+      }),
+      'complex json-encoded values are valid',
+    ),
+  ];
+
+  group('Auth header validity checking', () {
```
SandPod, line 100: Suggestion(NIT): Add a `-` or `:` symbol at the end of the grouping to clarify the prefix from the test name in the printed description.

tests/serverpod_test_server/test_e2e/auth_core_test.dart
```diff
@@ -0,0 +1,61 @@
...
+      );
+      var response =
+          await client.authentication.authenticate('test@foo.bar', 'password');
+      assert(response.success, 'Authentication setup failed');
+      authKey = '${response.keyId}:${response.key}';
+    });
+
+    tearDownAll(() async {
+      client.close();
+    });
```
SandPod, line 30: Suggestion: Add removal of the authenticated user to the teardown.

We should try and clean up any artifact created from the test. You can view such a cleanup here: https://github.com/serverpod/serverpod/blob/a3f8ab507b858146e30f0eac207249c12eb3eb59/tests/serverpod_test_server/test_e2e/method_streaming/authenticated_streaming_method_test.dart#L49-L57


I'm not sure that the assert at the end is required.

tests/serverpod_test_server/test_e2e/websocket_auth_test.dart
```diff
@@ -0,0 +1,69 @@
...
+      var receivedData = streamingStream.single;
+
+      expect(receivedData, throwsA(isA<TimeoutException>()));
+    });
+
+    test(
+        'when sending a message to a restricted endpoint with authentication '
+        'then the message should succeed', () async {
+      var response =
+          await client.authentication.authenticate('test@foo.bar', 'password');
```
SandPod, line 45: Suggestion: Please add clean up for this as well.

I gave a suggestion for how this can be done in a previous comment.

SandPod (commented):
Great job, thanks for fixing all the feedback.

Last comment regarding some test fixes then I'm happy to approve this! 🚀

SandPod (approved):
LGTM! 🎉

Issue #2454, opened 2024-07-09
Allow model classes with no fields (for use as events, sent via streaming message)


lukehutch:
Currently it's impossible to create a model `.spy.yaml` with no fields. (`serverpod generate` and the `Serverpod` VS Code plugin both disallow this.)

I need to be able to send events from the server to my app, and some of these events don't have any associated data -- they just cause the app to take some action (such as calling back to the server to refreshing a given database query, when the server knows that those query results will have changed).

Currently I have to create a `placeholder: String?` field in an otherwise empty model `.spy.yaml` just to keep Serverpod happy. 

I know that normally there would be no need to create empty model classes, but being able to send any `SerializableModel` object over the websocket makes it make sense to be able to have empty classes.

lukehutch:
Yes, I could use an enum, but I would have to wrap it in a class to use it as a message anyway. And some of my events have extra fields, while a few just need to trigger a specific view to be refreshed in the app.

This is a pretty low-priority request, since it's easy to work around, but to me it makes sense, if serializable objects can be seen as events when sent via websocket.

Isakdl:
> Yes, I could use an enum, but I would have to wrap it in a class to use it as a message anyway.

Why? At least from the Serverpod perspective you should be able to send the enum as a standalone entity. 

 > And some of my events have extra fields, while a few just need to trigger a specific view to be refreshed in the app.

Fair, but there shouldn't be anything stopping you from sending both the enum and the classes.

lukehutch:
> Why? At least from the Serverpod perspective you should be able to send the enum as a standalone entity.

Ah, I didn't realize individual enum values could be serialized as a `SerializableModel`, but that makes sense. Thanks! I'll close this.

PR #2673, opened 2024-08-26
feat: Support models without fields.

SandPod:
It adds support for models without any defined fields.

These models could be used for streaming connections to signal different events.

Fixes this already closed issue: #2454

## Pre-launch Checklist

## Breaking changes

None - this is a new feature.


tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -152,10 +141,21 @@ class ModelParser {
...
+    if (fieldsNode is! YamlMap) return fields;
+
+    fields.addAll(fieldsNode.nodes.entries.expand((fieldNode) {
+      return _parseModelFieldDefinition(
+        fieldNode,
+        docsExtractor,
+        extraClasses,
+        serverOnlyClass,
+      );
+    }).toList());
```
hampuslavin, line 157: Style question: Do you generally prefer the full function syntax in this project over the shorthand arrow?

tests/serverpod_test_server/test_integration/database_operations/crud/create_test.dart
```diff
@@ -82,4 +82,15 @@ void main() async {
     expect(inserted.id, isNotNull);
     expect(inserted.num, 1);
   });
+
+  test('Given a model without fields when inserting then the model is created.',
```
hampuslavin, line 86: nit 
```suggestion
  test('Given a model without fields when inserting it then the model is created.',
```

hampuslavin (approved):
🚀 🌟

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
SandPod: Not really, but we use function syntax if the body is multiple lines long.
This usually helps with code readability.

PR #2676, opened 2024-08-27
refactor: Break out endpoint's business logic

hampuslavin:
## Description
- This PR attempts to break out the business logic from the three different entry points into an `Endpoint`:
  - standard method
  - new streaming method
  - old `handleStreamMessage` method (to-be deprecated at some point)
- Business logic is defined here as: all logic that is _not_ related to the transport layer 

## Changes
- Breaks out the following business logic into a single place:
  - input parameter validation
  - retrieving the endpoint
  - retrieving the endpoint method
  - checking if the user is authorized
  
## Checklist


packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -52,6 +52,41 @@ abstract class EndpointDispatch {
     return connector;
   }
 
+  Future<(MethodConnector, Map<String, dynamic>)> tryGetEndpointMethod({
+    required Session session,
+    required String endpointPath,
+    required String methodName,
+    required String body,
```
SandPod, line 71: Question: `body` here is really the serialized parameters, right? This layer doesn't really need to know the "body" concept since that more feels like a keyword in the HTTP transport layer.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -52,6 +52,41 @@ abstract class EndpointDispatch {
...
+    }
+
+    var method = connector.methodConnectors[methodName];
+    if (method is! MethodConnector) {
+      throw MethodNotFoundException(
+          'Method "$methodName" not found in endpoint: $endpointPath');
+    }
+
+    var paramMap = parseParameters(
+        body.isEmpty ? null : body, method.params, serializationManager,
```
SandPod, line 157: Question: Would it make sense to have this `empty` check on the inside of the parseParameters method instead? And then strictly take in a typed `String`.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -434,6 +457,30 @@ class ResultInvalidParams extends Result {
   }
 }
 
+abstract class TryGetMethodException implements Exception {
```
SandPod, line 551: Suggestion: Rename to `TryGetEndpointMethodException`.

SandPod: Then it is self documenting in all contexts.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -72,12 +200,6 @@ abstract class EndpointDispatch {
...
     }
 
-    // Find correct connector
-    var connector = getConnectorByName(path);
-    if (connector == null) {
-      return ResultInvalidParams('Endpoint $path does not exist');
-    }
-
     // Read query parameters
     var queryParameters = <String, dynamic>{};
```
SandPod, line 204: Suggestion: Move the `handleUriCall` method out from endpoint_dispatch.

This is now strictly transport layer code right?

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -228,7 +251,7 @@ abstract class EndpointDispatch {
     Map<String, dynamic> additionalParameters = const {},
   }) {
     if (descriptions.isEmpty) return {};
-
+    print('Parsing parameters: $paramString');
```
Isakdl: Suggestion: remove debug print

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -52,6 +52,41 @@ abstract class EndpointDispatch {
     return connector;
   }
 
+  Future<(MethodConnector, Map<String, dynamic>)> tryGetEndpointMethod({
```
Isakdl: Suggestion: Can we find a different name for this method or move the auth logic? This method is not simply retrieving the endpoint as it is also executing the auth.

packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
```diff
@@ -92,7 +92,7 @@ abstract class EndpointWebsocketRequestHandler {
                   .deserializeByClassName(serialization);
 
               if (message == null) throw Exception('Streamed message was null');
-
+              // OLD WAY
```
Isakdl: Suggestion: I don't think this comments add any value, I would remove it.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -110,11 +110,7 @@ abstract class Session {
 
   bool _closed = false;
 
-  /// True if logging is enabled for this session. Normally, logging should be
-  /// enabled but it will be disabled for internal sessions used by Serverpod.
-  final bool enableLogging;
```
Isakdl, line 115: Suggestion: This cannot be removed as this is part of the public interface.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -146,24 +142,28 @@ abstract class Session {
...
-        serverId: server.serverId,
-      );
-    } else {
-      _logManager = null;
+      startLogging();
     }
   }
 
+  /// Turns on logging for the session if it was not already turned on.
+  void startLogging() {
```
Isakdl, line 150: Suggestion: Make private we should not expose this to the users.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -211,24 +328,19 @@ abstract class EndpointDispatch {
...
-      dynamic input, Type type, SerializationManager serializationManager) {
-    return serializationManager.deserialize(input, type);
-  }
-
   /// Parses query parameters from a string into a map of parameters formatted
   /// according to the provided [ParameterDescription]s.
   ///
   /// Throws an exception if required parameters are missing or if the
   /// paramString can't be jsonDecoded.
+  @visibleForTesting
```
SandPod, line 336: Suggestion: If this method needs to be tested individually I would suggest moving to another file and use it from there.

I'm not a big fan of `visibleForTesting` tags. If we promise an external interface we might as well expose it (but not necessarily export it from the Serverpod package).

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -139,6 +139,8 @@ abstract class EndpointDispatch {
       throw EndpointNotFoundException('Endpoint $endpointPath not found');
     }
 
+    if (connector.endpoint.logSessions) session.startLogging();
```
Isakdl, line 170: Question: Why is this added here? It should be unnecessary as the session will track if logging is enabled/needed (internally).

packages/serverpod/lib/src/server/session.dart
Isakdl: maybe move back into constructor? question pending further up though.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -323,7 +323,7 @@ class MethodCallSession extends Session {
     required String method,
     required this.queryParameters,
     required super.authenticationKey,
-    super.enableLogging = true,
+    super.enableLogging,
```
Isakdl, line 326: Question: This changes the default behavior since the variable is set to false now in the base session. Does this really work? The bool here is intended as an override for all other logging settings so that you are able to force turnoff all logging regardless of other settings.

This is mostly used with and important for the internal session.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -52,6 +53,133 @@ abstract class EndpointDispatch {
...
+      methodName: methodName,
+      body: body,
+      serializationManager: serializationManager,
+      additionalParameters: additionalParameters,
+    );
+
+    if (method is! MethodConnector) {
+      throw MethodNotFoundException(
+          'Method "$methodName" not found in endpoint: $endpointPath');
+    }
```
SandPod, line 134: Suggestion: Throw `InvalidEndpointMethodTypeException` from here as well.

No reason have a different behavior for the different types right?

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -52,6 +53,133 @@ abstract class EndpointDispatch {
...
+      endpointPath: endpointPath,
+      methodName: methodName,
+      body: body,
+      serializationManager: serializationManager,
+      additionalParameters: additionalParameters,
+    );
+
+    if (method is! MethodStreamConnector) {
+      throw InvalidEndpointMethodTypeException(
+          'Method "$methodName" not found in endpoint: $endpointPath');
```
SandPod, line 87: Suggestion: This message could be pre-formatted and we just pass in the methodName and the endpointPath.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -121,25 +121,8 @@ class MethodWebsocketRequestHandler {
...
-        endpoint: message.endpoint,
-        method: message.method,
-        responseType: OpenMethodStreamResponseType.endpointNotFound,
-      );
-    }
-
-    // Create session
     var session = MethodStreamSession(
       server: server,
-      enableLogging: endpointConnector.endpoint.logSessions,
```
Isakdl: Comment: Can we inject this variable into the method instead? I still believe centralizing this logic inside the session is better than spreading the logging init logic around the code base.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -52,6 +53,133 @@ abstract class EndpointDispatch {
...
+  }
+
+  Future<EndpointConnector> _tryGetEndpoint(
+      String endpointPath, Session session) async {
+    var connector = getConnectorByName(endpointPath);
+    if (connector == null) {
+      throw EndpointNotFoundException('Endpoint $endpointPath not found');
+    }
+
+    if (connector.endpoint.logSessions) session.startLogging();
```
SandPod, line 170: Suggestion: Maybe rename the method to `enableLogging()`?

packages/serverpod/lib/src/server/session.dart
```diff
@@ -146,24 +142,28 @@ abstract class Session {
     }
 
     if (enableLogging) {
```
SandPod, line 144: Suggestion: `enableLogging` should probably be called `loggingEnabled` right?

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -332,6 +333,7 @@ abstract class EndpointDispatch {
   ///
   /// Throws an exception if required parameters are missing or if the
   /// paramString can't be jsonDecoded.
+  @visibleForTesting
```
Isakdl, line 336: Comment: I'm generally not a fan of this type of thing, I think you should always test the public interfaces only and as far up the call stack as possible within the unit. 

To me this indicates that we either have too coupled code (which I guess is very likely in this case), or that we are testing on the wrong level.

Is it possible to refactor to make this testable in a reasonable way?

Isakdl (commented):
Always nice to break things out! :) Very nice improvements for the code base. 
I have a few thoughts and comments.

packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
```diff
@@ -65,55 +65,50 @@ abstract class EndpointWebsocketRequestHandler {
           var endpointName = data['endpoint'] as String;
           var serialization = data['object'] as Map<String, dynamic>;
 
-          var endpointConnector =
-              server.endpoints.getConnectorByName(endpointName);
-          if (endpointConnector == null) {
+          EndpointConnector endpointConnector;
```
SandPod, line 68: FYI: There is also authentication happening in the `_callStreamOpened` method. Do we need to think about that?

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: Woops 💅 I thought the analyzer caught those so that's why I didn't search for it 🧐

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -160,69 +172,22 @@ class MethodWebsocketRequestHandler {
...
-    try {
-      requestedInputStreams = EndpointDispatch.parseRequestedInputStreams(
-        descriptions: endpointMethodConnector.streamParams,
-        requestedInputStreams: message.inputStreams,
-      );
-    } catch (e) {
+    } on InvalidParametersException catch (e) {
       server.serverpod.logVerbose(
-        'Failed to parse input streams for open stream request: $message',
+        'Failed to parse parameters or input streams for open stream request: $message (${e.message})',
```
SandPod, line 177: Question: How is this formatted now? Does the `InvalidParametersException` contain more details around what could not be parsed?

packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
hampuslavin: Outdated code, I think it was in an earlier commit.

packages/serverpod/lib/src/server/session.dart
hampuslavin: Ah of course 👍 Thanks!

SandPod (commented):
A couple of comments and questions.
Overall, it looks good!

Leaving the session logging thoughts to Isak since he has been reworking that a bit.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: Definitely 👍

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: Good suggestion!

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: Should be! Will try to move it the `server`!

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: Agree! Placing it in a helper file should be the best solution then 👍

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: You are right, forgot about this one. Thanks!

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: Good idea! 👍

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: The problem is that then we shadow the constructor argument.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
hampuslavin: Yes, exactly!

hampuslavin: I assumed we still wanted the `$message` info, so I added the actual exception message inside parentheses.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: Very true, lets find a better name.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: I do agree! I think this is one of those functions which you do really want to have a detailed unit test for. As Alex pointed out here, we could move this to a helper file that is private to the package but still public for testing: https://github.com/serverpod/serverpod/pull/2676#discussion_r1732996759.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
christerswahn: Payload or data? Most important to call it the same everywhere!

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: I was thinking exactly the same but didn't end up at a future proof name. I was considering `tryAuthorizeAndGetEndpointMethod`, which would be more or less true for the current case. The potential problem I thought of was that since this will be the main code path for everything not related to the transport layer, other things might be added here in the future which doesn't fit the description. But maybe that's a future problem and it's very easy to rename so. WDYT?

packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
hampuslavin: Oh, good catch! Will refactor to cover that path too 👍

hampuslavin: [0493f37](https://github.com/serverpod/serverpod/pull/2676/commits/0493f37f73cd42d3f3666bfa9630ff82ad3750af)

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: [f1ca25c](https://github.com/serverpod/serverpod/pull/2676/commits/f1ca25cf356181e0382a1b16027752640672b975)

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: Refactored into different helpers.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
hampuslavin: Refactored into a callback instead: [cec5fad](https://github.com/serverpod/serverpod/pull/2676/commits/cec5fad3d1dececc42a205ca6662da6c4342f464)

packages/serverpod/lib/src/server/session.dart
hampuslavin: Reverted this change: [cec5fad](https://github.com/serverpod/serverpod/pull/2676/commits/cec5fad3d1dececc42a205ca6662da6c4342f464)

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: Reverted here: [cec5fad](https://github.com/serverpod/serverpod/pull/2676/commits/cec5fad3d1dececc42a205ca6662da6c4342f464)

packages/serverpod/lib/src/server/session.dart
hampuslavin: Reverted here: [cec5fad](https://github.com/serverpod/serverpod/pull/2676/commits/cec5fad3d1dececc42a205ca6662da6c4342f464)

packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
hampuslavin: Reverted here: [cec5fad](https://github.com/serverpod/serverpod/pull/2676/commits/cec5fad3d1dececc42a205ca6662da6c4342f464)

packages/serverpod/lib/src/server/session.dart
hampuslavin: Reverted here: [cec5fad](https://github.com/serverpod/serverpod/pull/2676/commits/cec5fad3d1dececc42a205ca6662da6c4342f464)

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: Renamed to `getAuthorizedEndpointMethodStreamConnector`

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: [ee95aed](https://github.com/serverpod/serverpod/pull/2676/commits/ee95aed84f097c90c7d982ac5ed5739a01fec666)

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: [1961a38](https://github.com/serverpod/serverpod/pull/2676/commits/1961a38ff0ce7b4dfb098d2ba1fd80b5dbf20c99)

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: [bba28a1](https://github.com/serverpod/serverpod/pull/2676/commits/bba28a1871abfb72ad2f311ec5642fef584d497c)

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: [bba28a1](https://github.com/serverpod/serverpod/pull/2676/commits/bba28a1871abfb72ad2f311ec5642fef584d497c)

packages/serverpod/lib/src/server/server.dart
```diff
@@ -407,9 +407,110 @@ class Server {
...
+            authenticationKey: authenticationKey,
+            enableLogging: connector.endpoint.logSessions,
+          );
+          return session;
+        },
+        endpointPath: endpointName,
+        methodName: methodName,
+        parameters: queryParameters,
+        serializationManager: serializationManager,
+      );
```
hampuslavin, line 478: @SandPod @Isakdl Not enthralled by the `createSessionCallback`, but at least it avoids any changes to the Session. The obvious drawback is the `late` modifier to the session which can throw if future developers change the behavior. I was considering this instead, which is more code but should be perfectly robust, WDYT?
```dart
MethodCallSession? maybeSession;

    Endpoint endpoint;
    MethodConnector method;
    Map<String, dynamic> paramMap;
    try {
      (endpoint, method, paramMap) =
          await endpoints.getAuthorizedEndpointMethodConnector(
        createSessionCallback: (connector) {
          maybeSession = MethodCallSession(
            server: this,
            uri: uri,
            body: body,
            path: path,
            httpRequest: request,
            method: methodName,
            endpoint: endpointName,
            queryParameters: queryParameters,
            authenticationKey: authenticationKey,
            enableLogging: connector.endpoint.logSessions,
          );
          return maybeSession!;
        },
        endpointPath: endpointName,
        methodName: methodName,
        parameters: queryParameters,
        serializationManager: serializationManager,
      );
    } on MethodNotFoundException catch (e) {
      return ResultInvalidParams(e.message);
    } on InvalidEndpointMethodTypeException catch (e) {
      return ResultInvalidParams(e.message);
    } on EndpointNotFoundException catch (e) {
      return ResultInvalidParams(e.message);
    } on NotAuthorizedException catch (e) {
      return e.authenticationFailedResult;
    } on InvalidParametersException catch (e, stackTrace) {
      var sessionLogId =
          await maybeSession?.close(error: e, stackTrace: stackTrace);
      return ResultInternalServerError(
          e.toString(), stackTrace, sessionLogId ?? 0);
    }

    MethodCallSession? session = maybeSession;
    if (session == null) {
      return ResultInternalServerError(
          'Session not created', StackTrace.current, 0);
    }
```

SandPod: Suggestion: Couldn't you make session nullable? Then we can close it if it has been created?

Another note here is that we will need to call `close` on the session in all cases if an exception occurs.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -122,36 +123,52 @@ class MethodWebsocketRequestHandler {
     WebSocket webSocket,
     OpenMethodStreamCommand message,
   ) async {
-    // Validate targeted endpoint method
-    var endpointConnector =
-        server.endpoints.getConnectorByName(message.endpoint);
-    if (endpointConnector == null) {
+    late MethodStreamSession session;
```
SandPod, line 126: Suggestion: This session must be closed if we fail to call the streaming endpoint.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: > It is not really reflected in the name of the method we are calling

I did update the name to clarify that an authorization takes place, but at some point I think a line has to be drawn. It seems unreasonable to cover every single thing that is happening in this method in its name. The goal I feel should be to cover the clear intent of the method and not do anything unexpected. As part of getting the method, a validation is made that it's a legit call. This seems fully expected to me, but this is somewhat subjective naturally! If you still think so, please give me a code example so I can see the full idea you had.

>  parameters should really be a concern when fetching the endpoint method?

This logic that we may want to update and should be true for all code paths. To me it seems most logical (and expected to future developers) to do it at the place of retrieval, but we can of course move it if we think that would make it even easier to understand and follow these code paths. But again, the goal here is to have a single entry point into all of this business logic.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
SandPod: Great!

packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
```diff
@@ -139,37 +126,41 @@ abstract class EndpointWebsocketRequestHandler {
...
+      try {
+        session.endpoint = endpointConnector.name;
+        connector = await endpointDispatch.getAuthorizedEndpoint(
+            session: session, endpointPath: endpointConnector.name);
+        await connector.endpoint.streamClosed(session);
+      } on NotAuthorizedException {
+        continue;
+      } catch (e) {
+        return;
+      }
```
SandPod, line 163: Suggestion: Move loop out to the outside again.

From my perspective, having the loop on the outside more clearly communicated what was happening.
There is also a slight logical change with this implementation where we would call `return` for all endpoints in the loop if another exception happens as opposed to only skipping a single entry.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -307,8 +276,20 @@ abstract class EndpointMethodConnector {
   /// List of parameters used by the method.
   final Map<String, ParameterDescription> params;
 
+  /// List of [Scope]s that are required to access this [EndpointMethodConnector].
+  Set<Scope> requiredScopes;
+
+  /// States if the [EndpointMethodConnector] only should accept users that are authenticated.
+  /// Default value is false
+  bool requireLogin;
```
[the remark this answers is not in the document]
SandPod: Yes, but you better make it final so that is never mutated after the instance has been created.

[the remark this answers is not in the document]
SandPod: Same for required scopes.

Could even be an idea that this is just straight up delegated to the EndpointConnector directly instead of storing a cached value.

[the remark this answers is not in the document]
SandPod: Shouldn't this change be reflected in the generator as well? So that the constructor of the EndpointMethodConnector is correctly initiated?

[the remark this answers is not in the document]
SandPod: In my mind, all our authentication tests for streaming methods should fail with this change?

SandPod (commented):
A couple of additional comments.

Most importantly we need to remember to close the session in all cases if a call fails.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
hampuslavin: That's what I reckoned as well, but this is actually not the current behavior, and _if_ I do that then a bunch of tests starts failing. Since this is the current behavior I just kept it the same, but maybe there's something here that needs to change. Lets discuss together! 👍

packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
hampuslavin: > we would call return for all endpoints in the loop if another exception happens as opposed to only skipping a single entry.

Maybe I'm missing something, but I don't think this is true? It seems to be the same behavior in both cases.

That said, I do agree that it's more explicit to have the loop outside. The downside was that the function signature needed more arguments in that case and looked a bit ugly, so I weighing the two options I thought this one was best. Lets discuss which one you prefer!

packages/serverpod/lib/src/server/server.dart
hampuslavin: Not sure I follow; the code snippet above makes it nullable? I take it that's preferred then? Regarding closing it, see this comment: https://github.com/serverpod/serverpod/pull/2676#discussion_r1735597748

packages/serverpod/lib/src/server/endpoint_dispatch.dart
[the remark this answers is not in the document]
hampuslavin: Your comments here make me think I completely misunderstood what you suggested the other day then 😄 Lets discuss!

packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
hampuslavin: [8105f19](https://github.com/serverpod/serverpod/pull/2676/commits/8105f196c352927da6fbdd3c8c1318a60cb2a981)

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
hampuslavin: [640023c](https://github.com/serverpod/serverpod/pull/2676/commits/640023c5e4997b5caa1f3ec9a420324d63e493db)

packages/serverpod/lib/src/server/server.dart
hampuslavin: [640023c](https://github.com/serverpod/serverpod/pull/2676/commits/640023c5e4997b5caa1f3ec9a420324d63e493db)

packages/serverpod/lib/src/server/endpoint_dispatch.dart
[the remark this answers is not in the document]
hampuslavin: Refactored to return a call context that exposes `endpoint` and thereby these values: [8573727](https://github.com/serverpod/serverpod/pull/2676/commits/8573727c0e0f74a5ceda85e4d518ae9a005b737b)

packages/serverpod/lib/src/server/endpoint_dispatch.dart
hampuslavin: [resolved offline]: Agree with @SandPod to return a call context to make it clearer ([8573727](https://github.com/serverpod/serverpod/pull/2676/commits/8573727c0e0f74a5ceda85e4d518ae9a005b737b)), and also rename this and remove "Authorize" to make it clearer that a lot of things is happening and not only a single thing.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -122,36 +123,50 @@ class MethodWebsocketRequestHandler {
...
+
+    MethodStreamCallContext methodStreamCallContext;
+    try {
+      methodStreamCallContext =
+          await server.endpoints.getMethodStreamCallContext(
+        createSessionCallback: (connector) {
+          session = MethodStreamSession(
```
SandPod, line 133: Suggestion: Add a close of this session in a `finally` statement to the try-catch.

This session is left open if we for example get a MethodNotFoundException right now.
We should probably also use a nullable session, like we did in the method call case, to get around the `late` initializion.

packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
```diff
@@ -141,34 +141,38 @@ abstract class EndpointWebsocketRequestHandler {
...
+        endpointPath: endpointName,
       );
-      if (authFailed == null) await endpoint.streamClosed(session);
-    } catch (e) {
+      await connector.endpoint.streamClosed(session);
+    } on NotAuthorizedException {
+      return;
+    } catch (_) {
       return;
     }
```
SandPod, line 177: Comment: just noticed that it is extremely important that we catch any exception here, otherwise we won't close the session since this happens after all calls to `_callStreamClosed`.

SandPod (commented):
One last critical thing around the session.

packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
hampuslavin: Yes, exactly!

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
hampuslavin: Good catch 🙏 🌟 [c7c0a55](https://github.com/serverpod/serverpod/pull/2676/commits/c7c0a55585868c1e95d317e8896b8d42059076f1)

SandPod (approved):
LGTM!

Left two minor non blocking NIT suggestions you are free to discard if you don't agree with them.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -122,24 +123,46 @@ class MethodWebsocketRequestHandler {
...
-    var endpointConnector =
-        server.endpoints.getConnectorByName(message.endpoint);
-    if (endpointConnector == null) {
+    Map<String, dynamic> arguments;
+    try {
+      arguments = jsonDecode(message.encodedArgs);
+    } catch (e) {
       server.serverpod.logVerbose(
-        'Endpoint not found for open stream request: $message',
+        'Failed to parse parameters for open stream request: $message ($e)',
```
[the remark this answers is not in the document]
hampuslavin: Good catch!

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -239,10 +222,23 @@ class MethodWebsocketRequestHandler {
             responseType: OpenMethodStreamResponseType.authenticationFailed,
           ),
       };
+    } catch (e) {
+      server.serverpod.logVerbose(
+        'Unexpected error when opening stream: $e',
+      );
+      throw StateError('Unexpected error when opening stream: $e');
+    } finally {
+      if (getMethodStreamCallContextFailed) await maybeSession?.close();
```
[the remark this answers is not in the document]
hampuslavin: Good point :)

PR #2678, opened 2024-08-27
feat: Implement Default Values on Exception model fields

### LINKED ISSUE #236: default value

### LINKED ISSUE #1140: Default value support in protocol / generated code / database

### LINKED ISSUE #2627: Support default values in exceptions
# Feature

With default values (#236) for classes, it makes sense to also support this in exceptions (except the defaultPresist). E.g. add the default keyword and allow the same datatypes and default value support as for normal classes.

klkucaj:
### Overview
This is a follow-up to #2383.  
This PR introduces support for default values in exception fields.

### Default Types
In exception models, there are two types of default values that can be specified:

- **default**: Adds a default value to the model field (Dart code).
- **defaultModel**: Adds a default value specifically to the model field on the Dart side.

### Important Note
Since exceptions are never persisted to the database, we do not support `defaultPersist` for exception fields. This means that only `default` and `defaultModel` are relevant in this context.

However, it’s important to note that `defaultModel` will override `default` if both are specified. Since exceptions only exist in the model (Dart code) and are not persisted, using `defaultModel` in combination with `default` is redundant. The `defaultModel` value will always take precedence over `default`, making it unnecessary to use both in conjunction.

### Default Value Options for Exception Fields
In exception models, default values can be assigned to any field type that is supported in class models. This provides flexibility in specifying initial values for various field types, ensuring consistency across your exception handling.

### Example Exception YAML
```yaml
exception: DefaultException
fields:
  intDefault: int, default=10
  boolDefault: bool, default=true
  stringDefault: String, default='Default message'
  dateTimeDefault: DateTime, default=now
  uuidDefault: UuidValue, default=random
  durationDefault: Duration, default=1d 2h 30min
  defaultModelField: String, defaultModel='Model specific message'
```

### Related Issues
- Closes #1140
- Closes #236
- Closes #2627

## Pre-launch Checklist


tests/serverpod_test_server/lib/src/models/defaults/exception/default_exception.spy.yaml
```diff
@@ -0,0 +1,5 @@
+exception: DefaultException
+fields:
+    defaultMessage: String, default='Default error message'
```
SandPod, line 3: Suggestion: Add all supported data types to the exception.

This asserts we always support everything even if it is removed from models.

SandPod: But i would say it is enought with just the `default` keyword for all the different types.

tests/serverpod_test_server/migrations/migration_registry.txt
```diff
@@ -5,4 +5,4 @@
 ### the conflict by removing and recreating the conflicting migration.
 
 20240125141253049
-20240827080004999
+20240827093851370
```
SandPod, line 8: Suggestion: Remove the recreated migration.

There is no difference in the migrations for exceptions (since they are not stored in the database.)

tests/serverpod_test_server/test/defaults/exception/enum_default_test.dart
```diff
@@ -0,0 +1,57 @@
...
+import 'package:test/test.dart';
+
+void main() async {
+  group(
+    "Given an exception with 'default' fields",
+    () {
+      test(
+        'when an exception is created, then the "defaultMessage" field should match the default value',
+        () {
+          var exception = DefaultException();
```
SandPod, line 11: Suggestion: Validated all data types.

When you expand the model, add check that this works for all the data types as well.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/default/exception/field_exception_default_test.dart
```diff
@@ -0,0 +1,123 @@
...
+import 'package:serverpod_cli/src/test_util/builders/generator_config_builder.dart';
+import 'package:serverpod_cli/src/test_util/builders/model_source_builder.dart';
+import 'package:test/test.dart';
+
+void main() {
+  var config = GeneratorConfigBuilder().build();
+
+  group('Given an exception with fields with a "default" keyword', () {
+    test(
+      'when the field is of type String and the default is set to "Default error message", then the field should have a "default model" value',
```
SandPod, line 14: Question/Discussion: If we would remove support for default values on Models along with all the tests there, we would have no validation that it would still work for exceptions right?

In theory, we could remove support for exceptions then as well. I therefore think it might be good to duplicate the tests to ensure we always have full support.

SandPod: This also makes it clear that the shared logic is used also for serializable exceptions.

SandPod (commented):
Nice work! change is pretty small.
Still would be nice to see more tests around this. Since, from the user perspective, this is separate feature.

SandPod (approved):
LGTM! 🚀

PR #2679, opened 2024-08-27
fix: Respect authentication changes in method streams.

SandPod:
### Background:
For the new streaming methods, authentication for each streaming method call is done on the open stream request.
After that no further validation is required.

However, since authentication for a user can be revoked at any time, there needs to exist a way to close any stream that are utalizing a revoked authentication 

Connected to: #2338

### Fix:
This PR introduced a new channel (or one for each authenticated user) used for communicating revoked authentications. The Serverpod framework will then listen to updates on the authenticated information and close streams if authentication has been changed in a way that would make the streaming method unavailable for the user.

The channel is prefixed with the user ID, and three types of events can be broadcasted on the channel.
- `RevokedAuthenticationUser` -  is used to communicate that all the user's authentication is
- `RevokedAuthenticationAuthId` - is used to communicate that a specific authentication id has been revoked for a user.
- `RevokedAuthenticationScope` is used to communicate that a specific scope (or scopes) have been revoked for the user.

### Additional changes:
***AuthenticatedInfo***
The AuthenticatedInfo object has also been extended with an optional `authId` field that is a `String`. The purpose of this string is to uniquely identify an authentiction key. This opens up the possibility to revoke access for single authentication keys and not a full user or a scope.

***Specific exception on connection closed error***
On the client, if a stream is closed with an exception, a specific stream closed error is thrown instead of the general client error. The purpose of this is to make it possible to build reconnect logic.

### Guidance for review
Best reviewed commit by commit. Changes have been grouped together with the tests to make it clear what has been validated.

## Pre-launch Checklist

## Breaking changes

None - Since this is only used in the new streaming interface these can be considered non breaking changes.
However, documentation around this will need to be added along with instructions on how to post information to these channels for users that have built their own authentication solutions.

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_server/lib/src/business/users.dart
+++ b/modules/serverpod_auth/serverpod_auth_server/lib/src/business/users.dart
@@ -1,3 +1,4 @@
+import 'package:serverpod/protocol.dart';
 import 'package:serverpod/serverpod.dart';
 import 'package:serverpod_auth_server/src/business/config.dart';
 import 'package:serverpod_auth_server/src/business/user_authentication.dart';
@@ -109,10 +110,8 @@ class Users {
     var userInfo = await findUserByUserId(session, userId, useCache: false);
     if (userInfo == null) return null;
 
-    var scopeStrs = <String>[];
-    for (var scope in newScopes) {
-      if (scope.name != null) scopeStrs.add(scope.name!);
-    }
+    var removedScopes = userInfo.scopes.difference(newScopes);
+    var scopeStrs = newScopes.map((s) => s.name).whereType<String>().toList();
     userInfo.scopeNames = scopeStrs;
     await UserInfo.db.updateRow(session, userInfo);
 
@@ -123,6 +122,18 @@ class Users {
 
     if (AuthConfig.current.onUserUpdated != null) {
       await AuthConfig.current.onUserUpdated!(session, userInfo);
+    }
+
+    // Notify if any scopes are revoked for user.
+    if (removedScopes.isNotEmpty) {
+      var removedScopesList =
+          removedScopes.map((s) => s.name).whereType<String>().toList();
+      await session.messages.revokedAuthentication(
+        userId,
+        RevokedAuthenticationScope(
+          scopes: removedScopesList,
+        ),
+      );
     }
 
     await invalidateCacheForUser(session, userId);
--- a/modules/serverpod_auth/serverpod_auth_server/lib/src/business/users.dart
+++ b/modules/serverpod_auth/serverpod_auth_server/lib/src/business/users.dart
@@ -124,8 +124,8 @@ class Users {
       await AuthConfig.current.onUserUpdated!(session, userInfo);
     }
 
-    // Notify if any scopes are revoked for user.
-    if (removedScopes.isNotEmpty) {
+    var scopesHaveBeenRevoked = removedScopes.isNotEmpty;
+    if (scopesHaveBeenRevoked) {
       var removedScopesList =
           removedScopes.map((s) => s.name).whereType<String>().toList();
       await session.messages.authenticationRevoked(
--- a/modules/serverpod_auth/serverpod_auth_server/lib/src/business/users.dart
+++ b/modules/serverpod_auth/serverpod_auth_server/lib/src/business/users.dart
@@ -128,7 +128,7 @@ class Users {
     if (removedScopes.isNotEmpty) {
       var removedScopesList =
           removedScopes.map((s) => s.name).whereType<String>().toList();
-      await session.messages.revokedAuthentication(
+      await session.messages.authenticationRevoked(
         userId,
         RevokedAuthenticationScope(
           scopes: removedScopesList,
--- a/packages/serverpod/lib/src/models/authentication/revoked_authentication_user.spy.yaml
+++ b/packages/serverpod/lib/src/models/authentication/revoked_authentication_user.spy.yaml
@@ -1,0 +1,3 @@
+### Message sent when all authentication for a user is revoked.
+class: RevokedAuthenticationUser
+serverOnly: true
--- a/packages/serverpod/lib/src/server/session.dart
+++ b/packages/serverpod/lib/src/server/session.dart
@@ -631,7 +631,7 @@ class MessageCentralAccess {
   ///
   /// [RevokedAuthenticationScope] is used to communicate that a specific
   /// scope or scopes have been revoked for the user.
-  Future<bool> revokedAuthentication(
+  Future<bool> authenticationRevoked(
     int userId,
     SerializableModel message,
   ) async {
--- a/packages/serverpod/lib/src/server/session.dart
+++ b/packages/serverpod/lib/src/server/session.dart
@@ -5,6 +5,7 @@ import 'dart:typed_data';
 
 import 'package:meta/meta.dart';
 import 'package:serverpod/serverpod.dart';
+import 'package:serverpod/src/generated/protocol.dart';
 import 'package:serverpod/src/server/features.dart';
 import 'package:serverpod/src/server/log_manager/log_manager.dart';
 import 'package:serverpod/src/server/log_manager/log_settings.dart';
@@ -612,6 +613,42 @@ class MessageCentralAccess {
   /// emit an error.
   Stream<T> createStream<T>(String channelName) =>
       _session.server.messageCentral.createStream<T>(_session, channelName);
+
+  /// Broadcasts revoked authentication events to the Serverpod framework.
+  /// This message ensures authenticated connections to the user are closed.
+  ///
+  /// The [userId] should be the [AuthenticationInfo.userId] for the concerned
+  /// user.
+  ///
+  /// The [message] must be of type [RevokedAuthenticationUser],
+  /// [RevokedAuthenticationAuthId], or [RevokedAuthenticationScope].
+  ///
+  /// [RevokedAuthenticationUser] is used to communicate that all the user's
+  /// authentication is revoked.
+  ///
+  /// [RevokedAuthenticationAuthId] is used to communicate that a specific
+  /// authentication id has been revoked for a user.
+  ///
+  /// [RevokedAuthenticationScope] is used to communicate that a specific
+  /// scope or scopes have been revoked for the user.
+  Future<bool> revokedAuthentication(
+    int userId,
+    SerializableModel message,
+  ) async {
+    if (message is! RevokedAuthenticationUser &&
+        message is! RevokedAuthenticationAuthId &&
+        message is! RevokedAuthenticationScope) {
+      throw ArgumentError(
+        'Message must be of type RevokedAuthenticationUser, '
+        'RevokedAuthenticationAuthId, or RevokedAuthenticationScope',
+      );
+    }
+    return _session.server.messageCentral.postMessage(
+      MessageCentralServerpodChannels.revokedAuthentication(userId),
+      message,
+      global: true,
+    );
+  }
 }
 
 /// Internal methods for [Session].
--- a/packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
+++ b/packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
@@ -415,7 +415,7 @@ class _MethodStreamManager {
     SerializationManager serializationManager,
     _RequiredAuthentication? requiredAuthentication,
   ) {
-    late void Function(SerializableModel) revokedAuthenticationCallback;
+    void Function(SerializableModel)? revokedAuthenticationCallback;
 
     if (requiredAuthentication != null) {
       revokedAuthenticationCallback = (event) async {
@@ -457,7 +457,8 @@ class _MethodStreamManager {
       /// or a request from the client.
       if (isCanceled) return;
       isCanceled = true;
-      if (requiredAuthentication != null) {
+      if (requiredAuthentication != null &&
+          revokedAuthenticationCallback != null) {
         session.messages.removeListener(
           MessageCentralServerpodChannels.revokedAuthentication(
             requiredAuthentication.authInfo.userId,
--- a/packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
+++ b/packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
@@ -241,10 +241,10 @@ class MethodWebsocketRequestHandler {
       };
     }
 
-    var authenticationRequired = endpointConnector.endpoint.requireLogin ||
+    var authenticationIsRequired = endpointConnector.endpoint.requireLogin ||
         endpointConnector.endpoint.requiredScopes.isNotEmpty;
     _RequiredAuthentication? requiredAuthentication;
-    if (authenticationRequired) {
+    if (authenticationIsRequired) {
       var authenticationInfo = await session.authenticated;
       if (authenticationInfo == null) {
         throw StateError(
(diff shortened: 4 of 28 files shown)


packages/serverpod/lib/src/models/authentication/revoked_authentication_user.spy.yaml
```diff
@@ -0,0 +1,3 @@
+### Message sent when all authentication for a user is revoked.
+class: RevokedAuthenticationUser
+serverOnly: true
```
Isakdl, line 3: Question: I get that the channel it self has the userId but does it not make sense to include the userId here as well for brevity?

packages/serverpod/lib/src/server/session.dart
```diff
@@ -612,6 +613,42 @@ class MessageCentralAccess {
...
+  ///
+  /// [RevokedAuthenticationScope] is used to communicate that a specific
+  /// scope or scopes have been revoked for the user.
+  Future<bool> revokedAuthentication(
+    int userId,
+    SerializableModel message,
+  ) async {
+    if (message is! RevokedAuthenticationUser &&
+        message is! RevokedAuthenticationAuthId &&
+        message is! RevokedAuthenticationScope) {
```
Isakdl, line 640: comment: Would be nice with inheritance here, or union types 😓

packages/serverpod/lib/src/server/session.dart
```diff
@@ -612,6 +613,42 @@ class MessageCentralAccess {
...
+  ///
+  /// [RevokedAuthenticationUser] is used to communicate that all the user's
+  /// authentication is revoked.
+  ///
+  /// [RevokedAuthenticationAuthId] is used to communicate that a specific
+  /// authentication id has been revoked for a user.
+  ///
+  /// [RevokedAuthenticationScope] is used to communicate that a specific
+  /// scope or scopes have been revoked for the user.
+  Future<bool> revokedAuthentication(
```
hampuslavin, line 634: Suggestion: since an action is taken below (`postMessage`), perhaps this should be named `revokeAuthentication` (imperative command), instead of past tense?

Isakdl, line 634: Suggestion: Use the name `authenticationRevoked` to make it sound more like an event that happened rather than some you are doing. I get that the difference here is super small.

modules/serverpod_auth/serverpod_auth_server/lib/src/business/users.dart
```diff
@@ -125,6 +124,18 @@ class Users {
       await AuthConfig.current.onUserUpdated!(session, userInfo);
     }
 
+    // Notify if any scopes are revoked for user.
+    if (removedScopes.isNotEmpty) {
```
hampuslavin, line 128: Suggestion to replace comment with descriptive variable name, e.g:
```suggestion
    var scopesHaveBeenRevoked = removedScopes.isNotEmpty;
    if (scopesHaveBeenRevoked) {
```
Then it' pretty clear why the rest is happening.

modules/serverpod_auth/serverpod_auth_server/lib/src/business/users.dart
```diff
@@ -125,6 +124,18 @@ class Users {
...
+      var removedScopesList =
+          removedScopes.map((s) => s.name).whereType<String>().toList();
```
hampuslavin, line 130: Just for my understanding, are there different type of scopes? String and something else?

tests/serverpod_test_server/test_integration/websockets/method_websockets/method_streams/revoked_authentication_stream_test.dart
```diff
@@ -0,0 +1,255 @@
...
+
+          inStream.add(1);
+          // Validate that the stream works
+          await valueReceivedCompleter.future;
+          assert(valueReceivedCompleter.isCompleted);
+        }
+      });
+
+      test(
+          'when the authenticated user is revoked then the stream is closed with error.',
```
Isakdl, line 84: Nit: ... stream is closed with **an** error.

Apply to all tests.

tests/serverpod_test_server/test_integration/websockets/method_websockets/method_streams/revoked_authentication_stream_test.dart
```diff
@@ -0,0 +1,255 @@
...
+          session.messages.revokedAuthentication(
+            authenticatedUserId,
+            RevokedAuthenticationUser(),
+          ),
+          completion(true),
+        );
+
+        await expectLater(streamClosedCompleter.future, completes);
+        var exception = await streamClosedCompleter.future;
+        expect(exception, isA<ServerpodClientException>());
```
Isakdl, line 96: Question: what data are we sending with this exception? Would perhaps be nice to validate that too, just to ensure we are not leaking anything.

tests/serverpod_test_server/test_integration/websockets/method_websockets/method_streams/revoked_authentication_stream_test.dart
```diff
@@ -0,0 +1,255 @@
...
+        valueReceivedCompleter = Completer<int>();
+        inStream.add(2);
+        await expectLater(
+          valueReceivedCompleter.future,
+          completion(2),
+        );
+      });
+
+      test(
+          'when the required scope for endpoint is revoked the stream is closed with error.',
```
Isakdl, line 135: Nit:  required scope for **an** endpoint is revoked **then** the stream is closed with **an** error.

Isakdl: Validate the other test descriptions too.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -240,6 +241,23 @@ class MethodWebsocketRequestHandler {
       };
     }
 
+    var authenticationRequired = endpointConnector.endpoint.requireLogin ||
+        endpointConnector.endpoint.requiredScopes.isNotEmpty;
+    _RequiredAuthentication? requiredAuthentication;
+    if (authenticationRequired) {
```
hampuslavin, line 247: Maybe it's just me, but these names are so similar (just the inverse of eachother) that I got a bit confused. Maybe adding an `is` to accentuate that one is a boolean and the other an object could make it a bit easier on the eye 👁️ 

```suggestion
    _RequiredAuthentication? requiredAuthentication;
    
    var authenticationIsRequired = endpointConnector.endpoint.requireLogin ||
        endpointConnector.endpoint.requiredScopes.isNotEmpty;
    if (authenticationIsRequired) {
```

tests/serverpod_test_server/lib/src/endpoints/authentication.dart
```diff
@@ -40,7 +40,7 @@ class AuthenticationEndpoint extends Endpoint {
           email: email,
           userName: 'Test',
           created: DateTime.now(),
-          scopeNames: [],
+          scopeNames: scopes ?? [],
```
Isakdl, line 43: comment: nice catch

Isakdl (changes requested):
Great addition should 🤞 make the streams complete.

I had some nit comments and minor thoughts!

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -391,7 +413,44 @@ class _MethodStreamManager {
...
+            reason: CloseReason.error,
+          );
+        }
+      };
+
+      session.messages.addListener(
+        MessageCentralServerpodChannels.revokedAuthentication(
+          requiredAuthentication.authInfo.userId,
+        ),
+        authenticationRevokedCallback,
```
hampuslavin, line 450: What happens if the `if` was not entered, will `authenticationRevokedCallback` be uninitialized then? If so, wouldn't that risk throwing later if some code tried to check for it?

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -391,7 +413,44 @@ class _MethodStreamManager {
     WebSocket webSocket,
     Session session,
     SerializationManager serializationManager,
+    _RequiredAuthentication? requiredAuthentication,
   ) {
+    late void Function(SerializableModel) authenticationRevokedCallback;
```
hampuslavin, line 418: Suggestion: Name with similar convention to the event that will invoke this callback. The event is `MessageCentralServerpodChannels.revokedAuthentication`, so I would suggest naming this `revokedAuthenticationCallback`.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -391,7 +413,44 @@ class _MethodStreamManager {
...
+
+    if (requiredAuthentication != null) {
+      // If authentication was required, a listener is registered that closes
+      // the stream if the authentication token is revoked.
+      authenticationRevokedCallback = (event) async {
+        var authenticationRevoked = switch (event) {
```
hampuslavin, line 424: Suggestion:  As mentioned elsewhere, make it explicit that it's a boolean to not confuse with other objects
```suggestion
        var authenticationWasRevoked = switch (event) {
```

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -391,7 +413,44 @@ class _MethodStreamManager {
     WebSocket webSocket,
     Session session,
     SerializationManager serializationManager,
+    _RequiredAuthentication? requiredAuthentication,
   ) {
+    late void Function(SerializableModel) authenticationRevokedCallback;
+
+    if (requiredAuthentication != null) {
+      // If authentication was required, a listener is registered that closes
+      // the stream if the authentication token is revoked.
```
hampuslavin, line 422: Question: Not sure what "comment policy" we have in this project. But generally for code such as this (which is very clean, short and clear) I wouldn't consider it necessary.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -720,3 +787,10 @@ class _OutputStreamContext implements _StreamContext {
 abstract interface class _StreamContext {
   StreamController get controller;
 }
+
+class _RequiredAuthentication {
```
hampuslavin, line 791: Question: Not sure exactly why, but somehow my brain doesn't expect `authInfo` here, because that is not "required" as such? The scopes are _required_, but the `authInfo` is just the current state. An example above in the code:
```dart
revokedAuthId.authId == requiredAuthentication.authInfo.authId,
```
In my head it would maybe make more sense with something like:
```dart
revokedAuthId.authId == currentAuthentication.authInfo.authId,
// or maybe even
revokedAuthId.authId == currentUser.authInfo.authId,
// or
revokedAuthId.authId == currentSessionAuth.authInfo.authId,
```

Maybe it's just me that has got the mental model wrong, let me know what your thoughts are!

hampuslavin (commented):
Nice work! 👍  Left a bunch of thoughts, but I'm very new to this code so a heads up that I might have misunderstood some things and concepts 😄

packages/serverpod/lib/src/server/session.dart
hampuslavin: Similar to my comment here: https://github.com/serverpod/serverpod/pull/2679#discussion_r1732927313

> Suggestion: since an action is taken below (postMessage), perhaps this should be named revokeAuthentication (imperative command), instead of past tense?

packages/serverpod/lib/src/models/authentication/revoked_authentication_user.spy.yaml
SandPod: We discussed it a bit, but what would it mean if the user id then is different than the one that the channel has? Should we validate that? It is basically redundant information so I opted to leave it out.

packages/serverpod/lib/src/server/session.dart
SandPod: Sooon thanks to the great work @BenAuerDev is doing! 🙌

packages/serverpod/lib/src/server/session.dart
SandPod: Good suggestion! will fix!

tests/serverpod_test_server/test_integration/websockets/method_websockets/method_streams/revoked_authentication_stream_test.dart
SandPod: I think we should probably close using another exception. Will change it to tthat instead. This is not a regulair client exception.

modules/serverpod_auth/serverpod_auth_server/lib/src/business/users.dart
SandPod: Scopes is an internal class, but we don't support serializing it. Therefore we need to do this dance of converting scopes to strings and strings to scopes in order to serialize it both to the database and to any communication layer.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
SandPod: That is true, in this case we have the whole world so we know we are safe. But it is brittle. I can switch it up a bit to let the function in itself determine if we should remove it or not.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
SandPod: Agree, comments have a tendency to become outdates so we should limit them if possible.
My fear here is this class is a bit of a brainchild of mine and I want to leave as many clues as possible for the next engineer to figure out what is happening. But this should only be limited to required comments so will remove this one.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
SandPod: I understand the thinking, I think we could probably work on the naming and the class name.

What i want to assert is that we always evaluate the authentiction info used to establish the connection. Since the session lives for a longer time the we don't want this evaluation to change. If users want to use a different authentication info they will need to establish a new connection.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
SandPod: Check out the new version now.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
hampuslavin: Thanks for the context, makes sense!

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
hampuslavin: Looks good! 👍

modules/serverpod_auth/serverpod_auth_server/lib/src/business/users.dart
hampuslavin: Ah I see, thanks!

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
hampuslavin: [resolved offline] agreed on `AuthenticationContext` as new name

Isakdl (approved):
LGTM 🚀

Issue #2691, opened 2024-08-30
Add support for environment variables for the secrets needed for S3 and GCP buckets


Isakdl:
**Improvement**
There are two secrets needed for bucket access in Serverpod, e.g. `HMACAccessKeyId` and `HMACSecretKey` but these can only be retrieved from the passwords.yaml.

Add support for them as environment variables as well!

PR #2692, opened 2024-08-30
feat: Ability to register custom environment variables to be loaded as a password.

Isakdl:
# Adds:

- Ability to register custom environment variable to be loaded into the passwords map managed by serverpod.
- Add custom envs for gcp buckets
- Add custom envs for s3 buckets

Motivation, we need to be able to load our bucket secrets from env and we want to keep these variables outside of the serverpod core. With this solution we can still centralize the logic in Serverpod without adding an implicit dependency to other packages. This also makes it possible for our users to add their own custom envs for passwords. 

Closes: #2691 

Review notes: 
Review commit by commit.
Naming on the public method `loadCustomPasswords` is important to discuss. 

## Pre-launch Checklist

## Breaking changes
none

### DIFF
--- a/packages/serverpod/lib/src/server/serverpod.dart
+++ b/packages/serverpod/lib/src/server/serverpod.dart
@@ -744,7 +744,7 @@ class Serverpod {
   void loadCustomPasswords(
     List<({String envName, String alias})> envPasswords,
   ) {
-    _passwords = _passwordManager.mergeCustomPasswords(
+    _passwords = _passwordManager.mergePasswords(
       envPasswords,
       _passwords,
       environment: Platform.environment,
--- a/packages/serverpod/lib/src/server/serverpod.dart
+++ b/packages/serverpod/lib/src/server/serverpod.dart
@@ -739,6 +739,8 @@ class Serverpod {
   /// contains the password. The alias is the key used to access the
   /// the password with the [getPassword] method.
   /// The alias also maps to the name in the config/passwords.yaml file.
+  /// This method may throw a [ArgumentError] if any Serverpod reserved passwords
+  /// are used as aliases or environment variables.
   void loadCustomPasswords(
     List<({String envName, String alias})> envPasswords,
   ) {
--- a/packages/serverpod/lib/src/server/serverpod.dart
+++ b/packages/serverpod/lib/src/server/serverpod.dart
@@ -65,7 +65,10 @@ class Serverpod {
 
   /// The server configuration, as read from the config/ directory.
   late ServerpodConfig config;
+
   Map<String, String> _passwords = <String, String>{};
+
+  late PasswordManager _passwordManager;
 
   /// Custom [AuthenticationHandler] used to authenticate users.
   final AuthenticationHandler? authenticationHandler;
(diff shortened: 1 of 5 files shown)


vlidholt (approved):
LGTM 🥳

packages/serverpod_shared/lib/src/password_manager.dart
```diff
@@ -71,4 +71,36 @@ class PasswordManager {
       return null;
     }
   }
+
+  /// Merge custom passwords with the existing password collection.
+  /// Custom passwords are loaded from the environment variables.
+  /// Throws an [ArgumentError] if the custom password configuration contains reserved keywords
```
SandPod, line 77: Question: What are the reserved keywords? Should we link to them here?

SandPod: Discussion(NIT): Maybe we could remove the `Custom` word? Isn't it enough with just `mergePasswords` or maybe `integratePasswords`?

packages/serverpod_shared/test/password_manager_test.dart
```diff
@@ -176,4 +176,120 @@ development:
...
+      () {
+    var passwordManager = PasswordManager(runMode: 'development');
+
+    var reservedEnvVariables = [
+      'SERVERPOD_DATABASE_PASSWORD',
+      'SERVERPOD_SERVICE_SECRET',
+      'SERVERPOD_REDIS_PASSWORD',
+    ];
+
+    for (var envName in reservedEnvVariables) {
```
SandPod, line 191: Suggestion: Remove the `for` loop and just validate a single case.

I don't think the looping helps the usecase in any way. We might as well remove it so that we get a single reproducable test.

SandPod: Same for other tests.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -729,6 +733,22 @@ class Serverpod {
     return _passwords[key];
   }
 
+  /// Registers passwords to be loaded from the env variables.
+  /// The password can be accessed with the [getPassword] method.
+  /// The envName is the name of the environment variable that
+  /// contains the password. The alias is the key used to access the
+  /// the password with the [getPassword] method.
+  /// The alias also maps to the name in the config/passwords.yaml file.
+  void loadCustomPasswords(
```
SandPod, line 742: Suggestion: Add information about the possible throw here as well.

Since this is the interface the user will actually be using it is important to have that information here.

SandPod: Isn't this actually the `merge` or `integrate` for the user?

SandPod (commented):
Nice clean work, I really appreciate the commit flow!

Have a couple of questions and suggestions.

packages/serverpod_shared/test/password_manager_test.dart
Isakdl: yeah you are probably right

packages/serverpod/lib/src/server/serverpod.dart
Isakdl: This is the public method indeed, my thinking here was that you are loading something from the env. (In the internal one we merge different maps basically, hence the difference). 

I don't think integrate is a good word here though. But +1 on the throw info.

PR #2693, opened 2024-08-30
fix: Fixed top error handling on server's request handler to ensure proper error boundary.

christerswahn:
Fixed top error handling on server's request handler to ensure proper error boundary and returning 500 error code.

## Pre-launch Checklist

## Breaking changes

n/a

### DIFF
--- a/packages/serverpod/lib/src/server/server.dart
+++ b/packages/serverpod/lib/src/server/server.dart
@@ -164,15 +164,7 @@ class Server {
           'received request: ${request.method} ${request.uri.path}',
         );
 
-        try {
-          _handleRequest(request);
-        } catch (e, stackTrace) {
-          stderr.writeln(
-            '${DateTime.now().toUtc()} Internal server error. _handleRequest failed.',
-          );
-          stderr.writeln('$e');
-          stderr.writeln('$stackTrace');
-        }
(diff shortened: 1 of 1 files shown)


packages/serverpod/lib/src/server/server.dart
```diff
@@ -184,8 +176,24 @@ class Server {
     stdout.writeln('$name stopped');
   }
 
+  void _handleRequestWithErrorBoundary(HttpRequest request) async {
+    // [Future.sync] ensures no synchronous error is accidentally thrown from the handler
+    // https://dart.dev/libraries/async/futures-error-handling#solution-using-future-sync-to-wrap-your-code
+    await Future.sync(() {
+      return _handleRequest(request);
```
SandPod, line 183: Question: Is no await required here?

SandPod (commented):
Good catch on this!

The change looks good to me.

packages/serverpod/lib/src/server/server.dart
christerswahn: The callback returns a FutureOr which is passed on to the caller if it is a Future, so the outer await is sufficient.

SandPod (approved):
LGTM! 🦖 
Good catch! 👍

PR #2695, opened 2024-08-30
feat: Add ability to modify maxFileSize and expiration time for gcp and aws buckets.

Isakdl:
# Changes

Use the expiration time input in the bucket and use it when creating the upload URL for s3 and gcp buckets.
Expose the maxFileSize variable for end users, this was previously hard coded at 10mb with no ability to modify this value.

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/integrations/serverpod_cloud_storage_gcp/lib/src/aws_s3_upload/aws_s3_upload.dart
+++ b/integrations/serverpod_cloud_storage_gcp/lib/src/aws_s3_upload/aws_s3_upload.dart
@@ -153,13 +153,21 @@ class GCPS3Uploader {
 
     /// The filename to upload as. If null, defaults to the given file's current filename.
     required String uploadDst,
+    Duration expires = const Duration(minutes: 10),
+    int maxFileSize = 10 * 1024 * 1024,
     bool public = true,
   }) async {
     final endpoint = 'https://storage.googleapis.com/$bucket';
 
     final policy = Policy.fromS3PresignedPost(
-        uploadDst, bucket, accessKey, 15, 10 * 1024 * 1024,
-        region: region, public: public);
+      uploadDst,
+      bucket,
+      accessKey,
+      expires.inMinutes,
+      maxFileSize,
+      region: region,
+      public: public,
+    );
     final key =
         SigV4.calculateSigningKey(secretKey, policy.datetime, region, 's3');
     final signature = SigV4.calculateSignature(key, policy.encode());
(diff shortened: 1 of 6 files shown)


integrations/serverpod_cloud_storage_s3/lib/src/aws_s3_upload/aws_s3_upload.dart
```diff
@@ -156,36 +156,25 @@ class AwsS3Uploader {
...
-    // final req = http.MultipartRequest("POST", uri);
-    // final multipartFile = http.MultipartFile('file', stream, length, filename: path.basename(uploadDst));
 
     final policy = Policy.fromS3PresignedPost(
-        uploadDst, bucket, accessKey, 15, 10 * 1024 * 1024,
-        region: region, public: public);
+      uploadDst,
+      bucket,
+      accessKey,
+      expires.inMinutes,
```
SandPod, line 169: Question: If I understand this correctly, the input value here was 15 before, but now we use a default value of 10 minutes. Is this an intentional change?

integrations/serverpod_cloud_storage_gcp/lib/src/aws_s3_upload/aws_s3_upload.dart
```diff
@@ -153,13 +153,21 @@ class GCPS3Uploader {
...
   }) async {
     final endpoint = 'https://storage.googleapis.com/$bucket';
 
     final policy = Policy.fromS3PresignedPost(
-        uploadDst, bucket, accessKey, 15, 10 * 1024 * 1024,
-        region: region, public: public);
+      uploadDst,
+      bucket,
+      accessKey,
+      expires.inMinutes,
```
SandPod, line 166: Question: Same comment here, the value was 15 before.

SandPod (commented):
Good change, thanks for clearing out all the dead comments.

Had one question.

integrations/serverpod_cloud_storage_s3/lib/src/aws_s3_upload/aws_s3_upload.dart
Isakdl: It is intentional as this is the default in the public interface. As a user I would expect that the default is 10 minutes and changing that to 15 will modify the previous default value.

SandPod: Aah, thanks for the explanation! 👍 
Very strange that we just ignored the value before.

SandPod (approved):
LGTM! 👍

PR #2699, opened 2024-09-02
fix: Add log for when uploads to buckets fails.

Isakdl:
# Changes

Prints a log when the status code is 4xx when uploading files to a bucket.

## Pre-launch Checklist

## Breaking changes
none

### DIFF
--- a/integrations/serverpod_cloud_storage_gcp/lib/src/aws_s3_upload/aws_s3_upload.dart
+++ b/integrations/serverpod_cloud_storage_gcp/lib/src/aws_s3_upload/aws_s3_upload.dart
@@ -61,9 +61,13 @@ class GCPS3Uploader {
     try {
       final res = await req.send();
 
+      if (res.statusCode >= 400 && res.statusCode < 500) {
+        print('Failed to upload to GCP, with reason: ${res.reasonPhrase}');
+      }
+
       if (res.statusCode == 204) return '$endpoint/$uploadDst';
     } catch (e) {
-      print('Failed to upload to AWS, with exception:');
+      print('Failed to upload to GCP, with exception:');
       print(e);
       return null;
     }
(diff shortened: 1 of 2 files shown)


integrations/serverpod_cloud_storage_gcp/lib/src/aws_s3_upload/aws_s3_upload.dart
```diff
@@ -61,9 +61,13 @@ class GCPS3Uploader {
     try {
       final res = await req.send();
 
+      if (res.statusCode >= 400 && res.statusCode < 500) {
+        print('Failed to upload to GCP, with reason: ${res.reasonPhrase}');
```
SandPod, line 65: Question: Shouldn't these be `stdout`/`stderr`?

integrations/serverpod_cloud_storage_gcp/lib/src/aws_s3_upload/aws_s3_upload.dart
```diff
@@ -61,9 +61,13 @@ class GCPS3Uploader {
...
+      }
+
       if (res.statusCode == 204) return '$endpoint/$uploadDst';
     } catch (e) {
-      print('Failed to upload to AWS, with exception:');
+      print('Failed to upload to GCP, with exception:');
```
SandPod, line 70: Suggestion: Could we rename these files to match what they are.

This is quite confusing right now. Or are we using some aws interface for GCP?

SandPod (approved):
LGTM! 🦖 

Had one question and one suggestion.

integrations/serverpod_cloud_storage_gcp/lib/src/aws_s3_upload/aws_s3_upload.dart
Isakdl: We are using the aws api that Google has support for. E.g. they have implemented the same interface to make transitions from s3 easier.

integrations/serverpod_cloud_storage_gcp/lib/src/aws_s3_upload/aws_s3_upload.dart
Isakdl: Maybe a good idea? I followed the existing logging in the file that is using print. Should I change them all?

SandPod: Might as well, unfortunate that we don't have a proper server logging system.

integrations/serverpod_cloud_storage_gcp/lib/src/aws_s3_upload/aws_s3_upload.dart
SandPod: Aah ok! 👍

Issue #1873, opened 2024-02-06
RFC: Add support for inheritance in model files


vlidholt:
A common feature request for Serverpod is to add support for inheritance in model files. It's easy to see how this can be a handy feature. The current approach is to support similar functionality through composition. But there are obvious pros of using inheritance instead.

- By using inheritance over composition, it would be possible to guarantee that fields are present instead of having nullable fields to compose different types of similar objects.
- We could support sealed classes. This would be great to ensure that developers handle all possible cases for a message with different compositions.

However, adding classes is a challenging task.

- To seal the classes, we will need to either place them in the same file or use the `part of` directive.
- Deserialization of inherited classes is tricky because we do not necessarily know the type of the class we are deserializing.
- How do we handle database tables?

## Implementation
Several aspects need to be considered to implement inheritance in model files.

- Add an (optional) `extends` and `sealed` keyword to the model files. The keyword. A base class can set `sealed` to `true`, in which case the class also becomes abstract. Any model in the package that extends the sealed class will be part of the seal.
- All generated files need to be connected through `part of` to support the seal feature.
- When serializing classes that use the `extends` keyword, the serializer will also save the class name (in the same way this happens in streams or when caching objects).
- In the first version, we can make tables optional for extended models, but it should be possible to add a `table` name to the extended model and just add all columns in the object-relation mapping.

## Examples
Setting up a sealed model:

```yaml
class: MyModel
sealed: true
fields:
  name: String
```

Extending a model:

```yaml
class: MyExtendedModel
extends: MyModel
fields:
  anotherName: String
```

(Maybe) not in the first version, add a table to an extended model:

```yaml
class: MyExtendedModelWithTable
extends: MyModel
table: my_extended_model_with_table
fields:
  yetAnotherName: String
```

lukehutch:
> In the first version, we can make tables optional for extended models, but it should be possible to add a `table` name to the extended model and just add all columns in the object-relation mapping.

+1 on this -- I can envision cases where a subclassed model should be able to be used to write to a parent model's table (where any fields added in the subclass are dropped, i.e. they become client/server-only fields, but are never persisted to the database), but also cases where a subclassed model should be able to be associated with its own table (whether or not it adds its own fields).

dkbast:
[...]

Issue #1618, opened 2023-12-01
Protocol format needs support for inheritance


lukehutch:
**Is your feature request related to a problem? Please describe.**

I need a way to create a database table whose columns are a superset of the columns of another database table. In other words, I need to be able to extend a protocol file, and have the generated subclass extend the generated superclass.

I have several needs for this, but principally, for law enforcement purposes, I need to preserve records for something like 3 months after a user tries to delete their account, with these requirements:
  * I need to be able to move user data from the primary table for active users over to an archive table.
  * I need the archive table to have at least the same columns as the primary table.
  * However, I also want to be able to add some more columns to the archive table, indicating the date that the record was archived, and the action that caused the record to be archived.
  * I don't want to just store the JSON serialization of rows of the primary table in a single column of the backup table, because I need the archived data to be queryable too.

Furthermore, I need to be able to store copies of certain record types, for statistical purposes. For example, I manually implemented the boilerplate needed to get templated queries to work (#1392), and now my `Profile` protocol class can serve as a database row, or a set of query criteria to find matching profiles. I want to be able to store the user's latest `Profile` query template in a separate table from the main profile table, so that I can run statistics over that table to see which profile options users most commonly search for.

**Describe the solution you'd like**

Basically this would look like something like

```yaml
class: Profile
table: profile
fields:
  userId: int
  name: String
  age: int
```

```yaml
class: ProfileArchive
extends: Profile
table: profile_archive
fields:
  archivedDate: DateTime
  archivedReason: String
```

and then `ProfileArchive` would have the following fields (and its table `profile_archive` would have the same database columns): `userId`, `name`, `age`, `archivedDate`, and `archivedReason`.

**Describe alternatives you've considered**

I don't currently see a way to set up the database declaratively to allow a protocol file to be stored in two tables (much less where one table has additional columns).

Isakdl:
This sounds like you need view tables, issue tracked here: https://github.com/serverpod/serverpod/issues/228 related discussion on syntax here: https://github.com/serverpod/serverpod/discussions/932

lukehutch:
I don't think so: `CREATE VIEW` creates a view, not a table, and views are not materialized. I need an actual physical copy of the row in a second table, as an archive or backup, because I want to be able to delete the original after a copy of it is archived.

lukehutch:
Correct, no automatic sync is needed.

For my use case, I need to be able to simply take a snapshot of records, and also add extra metadata, such as the timestamp of the snapshot.

[...]

PR #2710, opened 2024-09-04
Feature/extend data classes

BenAuerDev:
This PR adds the functionality to allow using the `extends` keyword to create a class based on another. 

```yaml
class: BaseClass
fields: 
  fieldOne: String
```

```yaml
class: SubClass
extends: BaseClass
fields: 
  fieldTwo: String
```

This generates a SubClass that can be used like

```dart
SubClass(fieldOne: '...', fieldTwo: '...');
```

It is not yet possible to add a DB-Table to a base- or sub-class, as a DB-Class extends TableRow and we can't extend two classes. To achieve this like requested in #1618 we would need to make TableRow an interface which can be implemented by the class so the extends property is free to be used for inheritance. New Issue: #2711

connected to: #1873 

## Review advice
I do recommend going through the changes commit by commit. I tried to make them (somewhat) bite sized.

## Review changes
As proposed in the first review, the functionality was also added to not directly expose the functionality to the user when running `serverpod generate`. To access the functionality the user must run `serverpod generate --experimental-features=inheritance` 

Screenshot from 2024-09-17 08-48-50

Also `ClassDefinition` and `EnumDefintion` now have a `type` field.

## Pre-launch Checklist

## Breaking changes

No breaking changes


SandPod (changes requested):
Awesome work! 🎉 

Thanks for some great readable code that has tests included.

I think we should resolve the extended class connection in a different way. I we take inspiration from how we resolve relations we can work with references instead of making lookups in multiple places.

We should also think about how we want to relase this. It would be great if we can merge your work to main but without exposing the new functionality to our users. That would allow us to get more fixes in before we ship it.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -211,6 +211,69 @@ class Restrictions {
...
+    SourceSpan? span,
+  ) {
+    if (extendsClass is! String) {
+      return [
+        SourceSpanSeverityException(
+          'The "extends" key must be a String.',
+          span,
+        )
+      ];
+    }
```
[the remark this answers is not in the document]
klkucaj: @BenAuerDev check how i did for defaults https://github.com/serverpod/serverpod/blob/main/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions/default.dart

i know it may not be the same structure and logic but use it as reference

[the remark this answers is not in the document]
BenAuerDev: @klkucaj thanks :rocket: 

@SandPod `documentType` represents: 

> The [documentType] represents the parent key of the [documentContents],
in the initial processing this is expected to be the top level model type
we are checking. E.g. 'class', 'enum', 'exception', etc.

Using `documentType` would results in:

error-from-doc-type

compared to before:

error-from-property-name


Which **kind of** makes sense as the value of  `extendsClass`  should be a class but I'm not 100% sure. Mostly because it might be a little harder to understand when the error references `class` type but points to `extends`. 

 What do you think? 

One thing that I just noticed, is that I used "property" where I should have used "type" in the error message

vlidholt:
Wow! Great work @BenAuerDev, really looking forward to getting this into Serverpod! 🚀

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -67,18 +67,21 @@ class SerializableModelLibraryGenerator {
...
           // https://stackoverflow.com/questions/68009392/dart-custom-copywith-method-with-nullable-properties
           if (_shouldCreateUndefinedClass(fields)) _buildUndefinedClass(),
-          _buildModelImplClass(
-            className,
-            classDefinition,
-            tableName,
-            fields,
-          ),
-          if (buildRepository.hasImplicitClassOperations(fields))
+          if (!isBaseClass)
```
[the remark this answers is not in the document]
SandPod: This is not a valid comment 😅

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -211,6 +211,69 @@ class Restrictions {
...
+          'The "extends" property cannot be used on a class with a table definition.',
+          span,
+        )
+      ];
+    }
+
+    return [];
+  }
+
+  List<SourceSpanSeverityException> validateExtendingClassName(
```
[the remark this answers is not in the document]
BenAuerDev: I think here we will also need to keep this

```dart
if (baseClass is ClassDefinition && baseClass.tableName != null) {
      return [
        SourceSpanSeverityException(
          'A base class cannot have a table definition. Please remove the "table" property from the class "$baseClassName".',
          span,
        )
      ];
    }
```

as the mutually exclusive case only covers `extendsClass` and `table`

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -22,17 +22,25 @@ class SerializableModelLibraryGenerator {
   });
 
   /// Generate the file for a model.
-  Library generateModelLibrary(SerializableModelDefinition modelDefinition) {
+  Library generateModelLibrary(
+    SerializableModelDefinition modelDefinition,
+    bool isBaseClass,
+    ClassDefinition? baseClass,
```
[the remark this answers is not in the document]
BenAuerDev: yes that becomes obsolete with the new model resolving workflow

BenAuerDev:
## Comment:

I was not able to add a test covering the check for external models to `inheritance_test.dart`. 

as we create the models for the test like

```dart
var modelSources = [
        ModelSourceBuilder().withYaml(
          '''
          class: Example
          fields:
            name: String
          ''',
        ).build(),
        ModelSourceBuilder().withFileName('example2').withYaml(
          '''
          class: ExampleSubClass
          extends: Example
          table: example_sub_class
          fields:
            age: int
          ''',
        ).build(),
      ];
```

and when used like:

```dart
var modelSources = [
      ModelSourceBuilder().withFileName('example2').withYaml(
        '''
          class: ExampleSubClass
          extends: ServerHealthMetric
          fields:
            age: int
          ''',
      ).build(),
    ];
```

the previous restriction

```dart
  if (baseClass == null) {
      return [
        SourceSpanSeverityException(
          'The class "$baseClassName" was not found in any model.',
          span,
        )
      ];
    }
```

is executed as every class not declared within `var modelSources = [...];` is null. 

But I tested it manually in the command line and the restriction is executed 

Screenshot from 2024-09-11 15-21-38

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -211,6 +211,69 @@ class Restrictions {
...
 
+  List<SourceSpanSeverityException> validateExtendsClassKey(
+    String parentNodeName,
+    dynamic extendsClass,
+    SourceSpan? span,
+  ) {
+    if (extendsClass is! String) {
+      return [
+        SourceSpanSeverityException(
+          'The "extends" key must be a String.',
```
SandPod, line 222: Suggestion(NIT): Refer to the `Keywords.extendsClass` keyword.

If we decide to change the keyword this change will then propagate to all the error messages :)
Same for the other messages.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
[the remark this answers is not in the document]
SandPod: I was incorrect here, your implementation is correct! :)

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -253,6 +261,35 @@ class ProtocolEnumValueDefinition {
...
+}
+
+class BaseClassInheritanceDefinition extends InheritanceDefinition {
+  final List<ClassDefinition?> subClasses;
+
+  BaseClassInheritanceDefinition(
+    super.className,
+    this.subClasses,
+  );
+}
```
SandPod, line 281: Suggestion/Discussion: Rename to `ParentClass`/`ChildClass` or `SuperClass`/`SubClass` - `InheritanceDefinition`.

Since "Base" actually has a meaning in a class [context in Dart](https://dart.dev/language/class-modifiers#base), this naming can be a bit confusing.
For me, it is also a bit unclear what the relationship is to the class from the perspective of this class.

By using the term `Parent`/`Child` or `Super`/`Sub` class I think this becomes a bit more clear.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -253,6 +261,35 @@ class ProtocolEnumValueDefinition {
...
+
+class ExtendsClassInheritanceDefinition extends InheritanceDefinition {
+  final List<SerializableModelFieldDefinition> baseClassFields;
+  final String baseClassPath;
+  ExtendsClassInheritanceDefinition(
+    super.className,
+    this.baseClassFields,
+    this.baseClassPath,
+  );
+}
```
SandPod, line 291: Suggestion/Discussion: With this definition of the class, we need to know up front what we want to retrieve for any use of it. How about we simply store the `ClassDefinition` and let the library generator pick out the fields that are required?

SandPod: Same for the `BaseClassInheritanceDefinition`.

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
```diff
@@ -30,6 +33,46 @@ class ModelDependencyResolver {
...
 
+  static void _resolveInheritance(
+    ClassDefinition classDefinition,
+    List<SerializableModelDefinition> modelDefinitions,
+  ) {
+    if (classDefinition.extendsClass != null) {
+      try {
+        var baseClass = modelDefinitions
+            .whereType<ClassDefinition>()
+            .firstWhere((element) =>
```
SandPod, line 44: Question: The `try/catch` statement here is only for the `firstWhere` statement right?
I would recommend either isolating that specific filtering or finding a non throwing way to find out if an elelement exists. For example the use of `firstWhereOrNull`.

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
```diff
@@ -30,6 +33,46 @@ class ModelDependencyResolver {
...
+        classDefinition.extendsClass = ExtendsClassInheritanceDefinition(
+          baseClass.className,
+          baseClass.fields,
+          p.joinAll([
+            ...baseClass.subDirParts,
+            '${baseClass.fileName}.dart',
+          ]),
+        );
+      } catch (e) {
+        classDefinition.extendsClass = null;
```
SandPod, line 56: Suggestion: If this fails, keep the UnresolvedInheritanceDefinition.

Then we can catch this in the validation and give an error on it.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -253,6 +261,35 @@ class ProtocolEnumValueDefinition {
   ProtocolEnumValueDefinition(this.name, [this.documentation]);
 }
 
+abstract class InheritanceDefinition {
```
SandPod, line 264: Suggestion: If we simply store a reference to the class, do we really need BaseClass and ExtendsClass? Isn't everything we need available through the class definition? Then we could simple have an Unresolved and Resolved Inheritance definition.

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
```diff
@@ -30,6 +33,46 @@ class ModelDependencyResolver {
...
+            .where((element) =>
+                element.extendsClass?.className == classDefinition.className)
+            .toList();
+
+        classDefinition.subClasses = BaseClassInheritanceDefinition(
+          classDefinition.className,
+          subClasses,
+        );
+      } catch (e) {
+        classDefinition.subClasses = null;
```
SandPod, line 71: Suggestion: Same as above, preserve the Unresolved if we are not able to resolve the reference.

tools/serverpod_cli/lib/src/test_util/builders/class_definition_builder.dart
```diff
@@ -22,6 +22,8 @@ class ClassDefinitionBuilder {
   List<_FieldBuilder> _fields;
   List<SerializableModelIndexDefinition> _indexes;
   List<String>? _documentation;
+  InheritanceDefinition? _subClass;
+  InheritanceDefinition? _baseClass;
```
SandPod, line 26: Suggestion: _subClass should be a list, and remember to rename these if you rename `base` to `super`

SandPod (changes requested):
Great improvement!
The resolver step will make this implementation a bit more flexible, still have a couple of comments regarding how the linking is built.

I'll also take a look at enabling testing with "serverpod" files. I believe this should be possible since we use the same interface as the code.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -357,11 +409,24 @@ class SerializableModelLibraryGenerator {
     ClassDefinition classDefinition,
     List<SerializableModelFieldDefinition> fields,
   ) {
+    var parentClass =
+        classDefinition.extendsClass is ResolvedInheritanceDefinition
+            ? (classDefinition.extendsClass as ResolvedInheritanceDefinition)
+                .classDefinition
+            : null;
```
SandPod, line 416: Question: This seems like code duplication of the logic that is part of the entrypoint for the library generated. Should this be placed inside of the classDefinition instead so that it can be reused?

SandPod: Then we could simply ask the classDefiniton for it's parent class.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -985,8 +1057,11 @@ class SerializableModelLibraryGenerator {
     List<SerializableModelFieldDefinition> fields,
     String? tableName, {
     required bool setAsToThis,
+    List<SerializableModelFieldDefinition>? parentClassFields,
   }) {
-    return fields
+    var inheritedFields = parentClassFields ?? [];
```
SandPod, line 1062: Suggestion: maybe this should also be part of the classDefinition? It seems like we are repeating this logic in a lot of places in the library generator as well?

Maybe these should simply be part of the "fields" for the class? Or is it valuable to differentiate between the parent fields and the class fields?

tools/serverpod_cli/lib/src/test_util/builders/class_definition_builder.dart
```diff
@@ -320,4 +325,48 @@ class ClassDefinitionBuilder {
     _isException = isException;
     return this;
   }
+
+  ClassDefinitionBuilder withChildClasses(
```
SandPod, line 329: Suggestion: The builder method indicates it should be possible to define multiple child classes.

Please take in a list of classDefinitions instead and wrap these in ResolvedInheritanceDefinition.

tools/serverpod_cli/lib/src/test_util/builders/class_definition_builder.dart
```diff
@@ -320,4 +325,48 @@ class ClassDefinitionBuilder {
...
+              fieldType,
+              nullable: fieldNullable,
+            )
+            .build(),
+      )
+    ];
+    return this;
+  }
+
+  ClassDefinitionBuilder withExtendsClass(
```
SandPod, line 352: Suggestion: Same for this interface, please take in a single ClassDefinition and wrap it inside of this method.

tools/serverpod_cli/lib/src/test_util/builders/generator_config_builder.dart
```diff
@@ -16,6 +17,7 @@ class GeneratorConfigBuilder {
   List<ModuleConfig> _modules;
   List<TypeDefinition> _extraClasses;
   List<ServerpodFeature> _enabledFeatures;
+  final List<ExperimentalFeature> _enabledExperimentalFeatures;
```
SandPod, line 20: Question: Shouldn't all of the lists here be final? If this is not used for `_enabledFeatures` i would suggest removing the final.

tools/serverpod_cli/test/generator/dart/client_code_generator/inheritance_class_test.dart
```diff
@@ -0,0 +1,234 @@
...
+import 'package:path/path.dart' as path;
+import 'package:test/test.dart';
+
+import 'package:serverpod_cli/src/generator/dart/client_code_generator.dart';
+import 'package:serverpod_cli/src/test_util/builders/class_definition_builder.dart';
+import 'package:serverpod_cli/src/test_util/builders/generator_config_builder.dart';
+import 'package:serverpod_cli/src/test_util/compilation_unit_helpers.dart';
+
+const projectName = 'example_project';
+final config = GeneratorConfigBuilder().withName(projectName).build();
```
SandPod, line 11: Question: How can these tests pass when we are not enabling the "extends" experiment?

Same for the `server_code_generator`.

SandPod (dismissed):
Another round of improvements but still a couple of things to fix.


Please take a look at the comments and let me know if something is unclear.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
BenAuerDev: Yeah this is something we can surely improve by moving this to `ClassDefinition`. :muscle: 

We kind of need to differentiate between the fields of the class and the `parentFields` as class-fields are passed to `this` and parentFields are passed to `super` in the constructor

tools/serverpod_cli/lib/src/commands/generate.dart
```diff
@@ -25,6 +27,13 @@ class GenerateCommand extends ServerpodCommand {
...
       help: 'Watch for changes and continuously generate code.',
     );
+    argParser.addMultiOption(
+      'experimental-features',
+      help:
+          'Enable experimental features. Experimental features might be removed at any time.',
+      allowed: ExperimentalFeature.values.map((e) => e.name),
+      defaultsTo: [],
+    );
   }
```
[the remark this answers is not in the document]
BenAuerDev: sure, I added them on top under 

## Review changes

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -211,6 +213,79 @@ class Restrictions {
...
+      ];
+    }
+
+    // TODO: Remove when inheritance is enabled by default.
+    if (!config.isExperimentalFeatureEnabled(ExperimentalFeature.inheritance)) {
+      return [
+        SourceSpanSeverityException(
+          'The "${Keyword.extendsClass}" key can only be used when the (experimental) inheritance feature is enabled.',
+          span,
+        )
```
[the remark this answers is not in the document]
BenAuerDev: I copied the error message over, as we will remove it soon :)

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -42,26 +48,47 @@ class SerializableModelLibraryGenerator {
...
-          if (buildRepository.hasImplicitClassOperations(fields))
+          if (!isParentClass)
+            _buildModelImplClass(
+              className,
+              classDefinition,
+              tableName,
+              [...extendedClassFields, ...fields],
+            ),
+          if (buildRepository.hasImplicitClassOperations(fields) &&
+              !isParentClass)
```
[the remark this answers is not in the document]
BenAuerDev: thats true, I removed it :)

tools/serverpod_cli/lib/src/test_util/builders/generator_config_builder.dart
BenAuerDev: I investigated it :) 

the linter wanted final only with _enabledExperimentalFeatures because the other fields are reassigned like:

```dart
GeneratorConfigBuilder withPackageType(PackageType type) {
    _type = type;
    return this;
  }
```

I on the other hand initially just added the passed feature to the list 

```dart
  GeneratorConfigBuilder withEnabledExperimentalFeatures(
      List<ExperimentalFeature> features) {
    _enabledExperimentalFeatures.addAll(features);
    return this;
  }
```

I solved it by also assigning a list to `_enabledExperimentalFeatures`

tools/serverpod_cli/test/generator/dart/client_code_generator/inheritance_class_test.dart
BenAuerDev: Uppon investigating it I would say, because we're guarding in `model_parser.dart` and here we create the models with `ClassDefinitionBuilder` and pass them directly to 

```dart
 var codeMap = generator.generateSerializableModelsCode(
      models: models,
      config: config,
  );
```

it works as intended in `inheritance_test.dart` because there we get the models from calling the `StatefulAnalyzer` 

```dart
var models = StatefulAnalyzer(config, modelSources, onErrorsCollector(collector)).validateAll();
```

Is this okay or should we add a guard also in the generator step?

tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -58,6 +59,7 @@ class GeneratorConfig {
     required List<ModuleConfig> modules,
     required this.extraClasses,
     required this.enabledFeatures,
+    required this.experimentalFeatures,
```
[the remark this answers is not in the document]
BenAuerDev: empty string threw an error, I made it an empty list :)

tools/serverpod_cli/test/generator/dart/client_code_generator/inheritance_class_test.dart
Isakdl: I think it is enough to guard against this in the parser, it is the only way a user can use the feature anyway. The code generator is a downstream consumer and shouldn't need to care. This looks good to me.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -85,6 +85,14 @@ class ClassDefinition extends SerializableModelDefinition {
         .cast()
         .firstWhere((element) => element.name == name, orElse: () => null);
   }
+
+  ClassDefinition? get parentClass =>
+      extendsClass is ResolvedInheritanceDefinition
+          ? (extendsClass as ResolvedInheritanceDefinition).classDefinition
+          : null;
```
Isakdl, line 92: Nit: Do not use arrow functions for multiline statements, opt for a normal method in these cases.

Suggestion: Do a local type check of the variable here instead so you can avoid doing a cast. The code works but when casting you are opening you self up to cast errors if someone modifies the code in the wrong way.

Instead this will solve the problem in a cleaner way:

```dart
ClassDefinition? get parentClass {
   var extendedClass = extendsClass;
   if (extendedClass is! ResolvedInheritanceDefinition) return null;

   return extendedClass.classDefinition;
}
```

tools/serverpod_cli/lib/src/config/experimental_feature.dart
```diff
@@ -1,3 +1,5 @@
+// TODO: Remove when inheritance is enabled by default.
+// Tracked by issue: https://github.com/serverpod/serverpod/issues/2711
```
Isakdl, line 2: NIT: put the comment above the inheritance keyword.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -230,7 +230,7 @@ class Restrictions {
     if (!config.isExperimentalFeatureEnabled(ExperimentalFeature.inheritance)) {
       return [
         SourceSpanSeverityException(
-          'The "${Keyword.extendsClass}" key can only be used when the (experimental) inheritance feature is enabled.',
+          'The "${Keyword.extendsClass}" property is not allowed for class type. Valid keys are {class, extends, table, managedMigration, serverOnly, fields, indexes}.',
```
Isakdl: Suggestion: I think we could create a generic solution for this case instead that should be able to print this list dynamically. 

If we add a new optional field to the `ValidateNode` class called `isDisabled` as a flag to enable a keyword (set default to false).

Then we can add a filter for this in the method `_collectInvalidKeyErrors`, like so:

```dart
var validKeys =
      documentStructure.where((e) => !e.isDisabled).map((e) => e.key).toSet();
```
This is the global method to check for valid keys and will work recursively for any level in the node tree.

Then in the class_yaml_definition you just need to set the isEnabled variable from the call to `config.isExperimentalFeatureEnabled(ExperimentalFeature.inheritance)`.

E.g.

```dart
ValidateNode(
    Keyword.extendsClass,
    isDisabled: !restrictions.config.isExperimentalFeatureEnabled(ExperimentalFeature.inheritance)
    ...
),
```

If you leave the test as is it should pass with the above modification.

Isakdl (changes requested):
Good work! I have looked at the latest commits and Alex comments. I have made a few smaller comments for additional changes in the new code.

Ill try to do a pass on the entire PR as well but ill post this for now.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
Isakdl: On second thought perhaps `isHidden` is a better word to use.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -211,6 +213,78 @@ class Restrictions {
...
+    SourceSpan? span,
+  ) {
+    if (extendsClass is! String) {
+      return [
+        SourceSpanSeverityException(
+          'The "${Keyword.extendsClass}" key must be a String.',
+          span,
+        )
+      ];
+    }
```
Isakdl, line 228: Suggestion: With the previously mentioned change of removing the isExperimentalFeatureEnabled from this method, I think the entire method can be deleted as we don't need any key specific validation anymore.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -35,6 +36,13 @@ class ModelParser {
     var className = classNode.value;
     if (className is! String) return null;
 
+    // TODO: Remove this check for `enabledInheritance == true` when we can also implement it for tables
+    // Tracked by issue: https://github.com/serverpod/serverpod/issues/2711
+    // After that, we can remove the check if `enableInheritance == true` and always parse the extends class
+    // var extendsClass = _parseExtendsClass(documentContents);
+    var extendsClass =
+        enableInheritance == true ? _parseExtendsClass(documentContents) : null;
```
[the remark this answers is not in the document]
Isakdl: Another comment here, do we actually need to toggle this at all in here?

I think the parsing logic is fine to keep right? As long as we throw an error in the analyzer we will block anyone from using this feature unless they turn it on. 

Having the parser extracting the value will not matter in this case. This means we could simplify the toggling to only a single place.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -42,6 +48,8 @@ class SerializableModelLibraryGenerator {
       config: config,
     );
 
+    bool isParentClass = classDefinition.childClasses.isNotEmpty;
```
Isakdl, line 51: Suggestion: Make this into a getter on the classDefinition instead. Then you do not need to pass this variable all the way down in the call stack and can instead access it by `classDefinition.isParentClass`

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/inheritance_test.dart
```diff
@@ -0,0 +1,249 @@
...
+      var parent = models.first as ClassDefinition;
+      var childClasses = parent.childClasses;
+      var isChildResolved = childClasses.first is ResolvedInheritanceDefinition;
+
+      var child = models.last as ClassDefinition;
+      var extendsClass = child.extendsClass;
+      var isExtendsResolved = extendsClass is ResolvedInheritanceDefinition;
+
+      expect(isChildResolved, isTrue);
+      expect(isExtendsResolved, isTrue);
```
Isakdl, line 55: Suggestion: Make this into a group and two destinct tests, where each field is in it's own test. 

While related I think these are two different values / features that could fail independently and it will be easier to debug / fix if they are reported as destinct tests.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/inheritance_test.dart
```diff
@@ -0,0 +1,249 @@
...
+
+    var error = collector.errors.first;
+    expect(
+      error.message,
+      'You can only extend classes from your own project.',
+    );
+  });
+
+  test(
+      'Given a child-class when inheritance is not enabled, then error is collected that experimental-features=inheritance needs to be enabled',
```
Isakdl, line 210: Suggestion: This test description no longer matches the implementation.

Isakdl (changes requested):
I did a pass on the entire code, have a few more comments here and there.

One thing I think would be nice to have included is a test in our test project.

E.g. enable the exprimental feature in our test project and create a model with this new feature. Generate the code and then implement a simple test that checks that one class extends the other by validating that it has the same type of the base class. 

With that I think we have pretty good coverage.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -66,17 +72,27 @@ class ClassDefinition extends SerializableModelDefinition {
     required super.serverOnly,
     required this.manageMigration,
     required this.isException,
+    childClasses,
```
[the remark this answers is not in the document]
BenAuerDev: With this `_resolveInheritance` throws an error at this part:

```dart
parentClass.childClasses.add(
  ResolvedInheritanceDefinition(classDefinition),
);
```

error: ` Unsupported operation: Cannot add to an unmodifiable list`

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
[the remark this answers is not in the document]
BenAuerDev: yeah you're right :) we can remove this :muscle:

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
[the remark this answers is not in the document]
Isakdl: I see, but in that case please add a type to this list in the constructor.

BenAuerDev:
I noticed a vulnerability with the import path of the parent class.

previously it was declared like:

```dart
p.joinAll([
    ...parentClass.subDirParts,
    '${parentClass.fileName}.dart',
]),
```

which works only if

- the files are in the same folder
  
  - `/models/child_class.spy.yaml`
    
  - `/models/parent_class.spy.yaml`
    

    OR

- the parent file is in a sub folder relative to the child class within models
  
  - `/models/child_class.spy.yaml`
    
  - `/models/extra_folder/parent_class.spy.yaml`



It worked NOT if we have for instance

- `/models/inheritance/child_class.spy.yaml`
  
- `/models/inheritance/parent_class.spy.yaml`
  

the resulting import looked like: `import 'inheritance/parent_class.dart' as _i1`

So I use an absolute path to import the parent and added some tests checking each case

- no sub dir 
- child in sub dir
- parent in sub dir
- both in sub dir

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -152,8 +152,7 @@ class SerializableModelLibraryGenerator {
           parentClass.className,
           p.joinAll([
             'package:${serverCode ? '${config.serverPackage}/src/generated' : '${config.dartClientPackage}/src/protocol'}',
-            ...parentClass.subDirParts,
-            '${parentClass.fileName}.dart',
+            'protocol.dart',
```
Isakdl: Suggestion: We already have code that manages these path inside the class `TypeDefinition`. But currently we only store this type definition for the class fields, but not for the class it self. 

This will be a little annoying to do, but I think it is the correct way to handle this case. 

Introduce a new field `type` on the `SerializableModelDefinition` class. In the model parser `e.g. model_parser.dart` you have to parse the type with the method `parseType` as the input of the type you should use `$defaultModulAlias:$className`.

e.g.
```dart
var classType = parseType(
      '$defaultModuleAlias:$className',
      extraClasses: extraClasses,
    );
```

Then you need to pass this type on into your parent resolver class you created. 

Then you should be able to call the method `reference` on the type to get the full URL automagically, instead of constructing it manually here.

E.g.

```dart
parentClass.type.reference(
          serverCode,
          subDirParts: subDirParts,
          config: config,
        );
```

Isakdl: With this, it would be nice with a small little test for the type field on the `SerializableModelDefinition` when parsing a yaml file. To verify that we create it correctly, will be enough to check that the class name is the type name.

Isakdl (changes requested):
Good work! Almost there I have one additional request to manage the import URL in the same way as the rest of the code, it is a little annoying but I think it will complete the functionality of our parser.

tools/serverpod_cli/lib/src/test_util/builders/enum_definition_builder.dart
```diff
@@ -30,7 +31,8 @@ class EnumDefinitionBuilder {
           ProtocolEnumValueDefinition('B'),
           ProtocolEnumValueDefinition('C'),
         ],
-        _documentation = [];
+        _documentation = [],
+        _type = TypeDefinitionBuilder().withClassName('Example').build();
```
Isakdl, line 35: Suggestion: This type does not have the enum definition on it, make sure that the builder represents the real state valid state by default!

tools/serverpod_cli/lib/src/test_util/builders/enum_definition_builder.dart
```diff
@@ -96,4 +98,9 @@ class EnumDefinitionBuilder {
     _documentation = documentation;
     return this;
   }
+
+  EnumDefinitionBuilder withType(TypeDefinition type) {
+    _type = type;
+    return this;
+  }
```
Isakdl, line 105: Suggestion: Remove this method you should never really set this manually, instead add an internal type definition construction if someone changes the enum definition class name.

Apply the same logic for the class definition builder too.

Isakdl (changes requested):
Almost there, the test builders do not produce a valid default state / management.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -211,6 +213,52 @@ class Restrictions {
...
+    if (parentClass == null) {
+      return [
+        SourceSpanSeverityException(
+          'The class "$parentClassName" was not found in any model.',
+          span,
+        )
+      ];
+    }
+
+    if (parentClass.moduleAlias != 'protocol') {
```
Isakdl, line 241: Nit: defaultAliasModule instead of 'protocol'

tools/serverpod_cli/lib/src/analyzer/models/model_analyzer.dart
```diff
@@ -39,8 +39,9 @@ class SerializableModelAnalyzer {
   /// Best effort attempt to extract a model definition from a yaml file.
   static SerializableModelDefinition? extractModelDefinition(
     ModelSource modelSource,
-    List<TypeDefinition> extraClasses,
-  ) {
+    List<TypeDefinition> extraClasses, {
+    bool enableInheritance = false,
```
Isakdl, line 43: Nit: Remove this variable is unused.

tools/serverpod_cli/lib/src/analyzer/models/stateful_analyzer.dart
```diff
@@ -85,6 +86,8 @@ class StatefulAnalyzer {
     var doc = SerializableModelAnalyzer.extractModelDefinition(
       state.source,
       config.extraClasses,
+      enableInheritance:
+          config.isExperimentalFeatureEnabled(ExperimentalFeature.inheritance),
```
Isakdl, line 90: Comment: As a consequence these should be removed as well.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -373,41 +416,44 @@ class SerializableModelLibraryGenerator {
...
-                _buildCopyWithAssignment(classDefinition, fields),
-              )
-              .returned
-              .statement;
+        m.name = 'copyWith';
+        if (!classDefinition.isParentClass) {
+          m.annotations.add(refer('override'));
+        }
+        m.optionalParameters.addAll(
+          [...classDefinition.parentFields, ...fields]
```
Isakdl, line 424: Question: Also another nit thing here, since you are destructing the classDefinition.parentFields and merge it with fields, can we simply not do this at the top level instead and just have the input fields be the merged fields? 

What is the difference between parentFields and fields?

Isakdl (changes requested):
I found some unused dead code and some very small nit things. But this should be the final change request!

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
BenAuerDev: I think we could, we only need to know which fields are from parent and which fields are from the child class because we have to pass them to `super` or `this` respectively

Isakdl: Got it, then my suggestion here is to lift it up wherever it is possible.

BenAuerDev: Resolved in Slack, this will be cleaned up when the table related classes will be activated

Isakdl (approved):
Good job @BenAuerDev, LGTM 🚀 🚀 🚀 let's go 😄

Issue #2494, opened 2024-07-16
RFC: Testing framework (server-side)


vlidholt:
Serverpod would benefit from including a framework that makes testing easier. Currently, to test endpoint methods, the user must either mock the `Session` class or set up end-to-end tests. A better solution would be to provide a way to call endpoint methods through a testing API.

![testing](https://github.com/user-attachments/assets/24c2478e-7393-4a88-b2bf-fcfa4706951b)

## Goals
It should be easy and intuitive to:
- Test endpoints
- Test relic endpoints
- Seed database state and corresponding clean-up
- Mock/override the following properties of `Session`:
  - `AuthenticationInfo? authenticated`
  - `Caches? caches`
  - `MessageCentral? messages`
  - `Storage? storage`
  - `Map<String,String>? passwords`
  - `dynamic userObject`
 
## Testing interface
Due to the way the Dart test framework works, it seems best and most idiomatic to create a HOC that provides the needed test objects and that can take care of set up and clean up that should be hidden to the end user.

Below follows a code example in which two objects, `TestServerpod` and `TestSession`, are provided to meet the requirements of the outlined goals:
```dart
withServerpod('Example Endpoint', (endpoints, session) {
  test('Hello', () async {
    final response =
        await endpoints.example.hello(session, 'Michael');
    expect(response, 'Hello, Michael!');
  });
  
  test('Hello with authenticated user', () async {
    AuthenticationInfo authInfo =
        AuthenticationInfo(12345, Set.of([Scope('admin')]));
  
    var session = session.withTestOverrides(
        userObject: 'data', authenticated: authInfo);
    final response =
        await endpoints.example.hello(session, 'Michael');
    expect(response, 'Hello, Michael!');
  });

// TODO: Add missing stream endpoint example

});

withRelic('Example Endpoint', (client, session) {
  test('Create user', () {
    var headers = {
      'Content-Type': 'application/json'
    };
    var body = {'name': 'Michael'};
    final response =
        client.post(session, '/users', body, headers);
    expect(response.body['id'], isA<String>);
    expect(response.status, equals(201));
  });
});
```
The names of properties and functions here are up for debate, but conceptually it makes sense to:
- Have a zero-conf way to start doing an endpoint test
- Some way to extend the baseline `testSession` like with `withTestOverrides` above (maybe `copyWith` is more idiomatic?)
[...]

PR #2713, opened 2024-09-05
feat: Add Serverpod test tools

hampuslavin:
## Description
Introduces the "alpha version" of the new Serverpod test tools, as outlined on [a high level here](https://github.com/serverpod/serverpod/issues/2494) (specifically item 2 and 3 on the roadmap at the bottom).

The main feature is the `withServerpod` helper:
```dart
withServerpod(
  (endpoints, session) {
    test(
        'when not authenticated and calling returnsString then throws UnauthenticatedEndpointCallTestException',
        () async {
      session = session.copyWith(authentication: AuthenticationOverride.authenticationInfo(1234, {}));

      final result =
          endpoints.authenticatedTestTools.returnsString(session, "Hello");
      await expectLater(
          result, throwsA(isA<UnauthenticatedEndpointCallTestException>()));
    });
  // ...
```
- Provides an `endoints` object that enables the user to easily test all their endpoints 
- Provides a test specific `session` object that lets the user call endpoints and the database just like they are used to. To customize the session state, the user simply calls `copyWith`.
  - The only thing that can be overridden with `copyWith` at the moment is `authenticationInfo`. As we go forward we will explore what the needs are and extend with more properties here, e.g. cache and message central.
-  By default, all `test`s are wrapped in database transactions which automatically rolls back the database after each test case

See the `tests/serverpod_test_server/test_integration/test_tools/*` folder for comprehensive examples of all capabilities.

💡 The test tools will be generated if the `server_test_tools_path` key is specified in `config/generator.yaml`. Since the templates are not updated in this PR, this is not exposed by default. Once it's production ready, the template will specify a default test tools path.

💡 As you will see, `withServerpod` is not exported by `serverpod_test` directly but from the _generated_ test tools file. The main benefit from doing it like this is that this completely removes the need for boiler plate code. Since using generated code is already done in other places, it doesn't go against the paradigm of the Serverpod framework.  

## Changes
- Adds test helpers, as described above
- Adds new test helper code generation module
- Adds new small `serverpod_test` package that contains the business logic for the test helper
- Adds code to prepare for running Serverpod in `test` mode, but will be finished in a follow-up PR (see Follow-up issues below)
- Adds test for the test helper in `tests/serverpod_test_server/test_integration/test_tools/*`. The goal is that these tests should be the source of truth of all the capabilities of the test helper. Therefore, these tests are written from the perspective of an end user.
- Adds unit tests for the test code generation
  - 💡 Note that I limited the tests to the public interfaces. That the internal implementation works is validated by all the integrations tests, as well as the linter which would not build our project if there were type errors.
- Adds `DatabaseAccessor` interface to enable the automatic rollback feature in tests

## Recommended reviewing
- Start by looking at the `tests/serverpod_test_server/test_integration/test_tools/*` folder to see how the test tool is used for different use cases
- I tried committing in logical order, so reviewing commit by commit should be best
- Please be explicit about wether you think something is missing in _this_ PR or if it's a suggestion for a follow-up PR.

## Follow-up issues [WIP]
There are a few things that needs to be fixed before this can be exposed publicly. See full list below:
- https://github.com/serverpod/serverpod/issues/2734
- https://github.com/serverpod/serverpod/issues/2725
- https://github.com/serverpod/serverpod/issues/2740
- https://github.com/serverpod/serverpod/issues/2739
- To discuss: "Global" setUpAll/tearDownAll
- To discuss: Enhance stack trace for errors thrown in tests (seems difficult)
 
## Pre-launch Checklist


packages/serverpod_test/lib/src/serverpod_test.dart
```diff
@@ -0,0 +1,415 @@
...
+/// A test specific session that is used to call database methods or pass to endpoints.
+abstract class TestSession implements DatabaseAccessor {
+  /// AuthenticationInfo for the session.
+  AuthenticationInfo? get authenticationInfo;
+
+  /// Creates a new unique session with the provided properties.
+  /// This is useful for setting up different session states in the tests
+  /// or simulating multiple users.
+  Future<TestSession> copyWith({
+    AuthenticationInfo? Function()? getAuthenticationInfo,
```
hampuslavin, line 17: 💡 This is a workaround for to basically simulate `Optional<AuthenticationInfo?>. In other places in the code we have a `Undefined` type to solve this, but in this case a callback seemed more straight forward.

tests/serverpod_test_server/test_integration/test_tools/rollback_database_test.dart
```diff
@@ -0,0 +1,253 @@
...
+              session,
+              where: (t) => Constant.bool(true),
+            );
+          });
+
+          test('then the database is updated according to the change',
+              () async {
+            await session.db.transaction((transaction) async {
+              await SimpleData.db.insert(session, [SimpleData(num: 111)],
+                  transaction: transaction);
```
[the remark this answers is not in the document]
SandPod: Same for other places.

SandPod (commented):
Dumping a first session of comments now after looking at the tests. Review is not complete but it seems like my review tab is slowing down.

packages/serverpod_test/lib/src/serverpod_test.dart
```diff
@@ -0,0 +1,415 @@
...
+          ? getAuthenticationInfo()
+          : _authenticationInfo,
+      enableLogging: enableLogging ?? _enableLogging,
+      transaction: transaction,
+    );
+    var newServerpodSession = await _testServerpod.createSession(
+      enableLogging: enableLogging ?? _enableLogging,
+    );
+
+    newSession.setAndConfigureServerpodSession(newServerpodSession);
```
SandPod, line 86: Question(For my understanding): The reason this can not be done in the constructor is because we preserve all test sessions between the tests, since we don't control their environment, but reset their internal session?

packages/serverpod_test/lib/src/serverpod_test.dart
```diff
@@ -0,0 +1,415 @@
...
+    unawaited(
+      serverpodSession.db.transaction((newTransaction) async {
+        localTransaction = newTransaction;
+
+        transactionStartedCompleter.complete();
+
+        await _endTransactionScopeCompleter.future
+            .timeout(const Duration(seconds: 60), onTimeout: () async {
+          await newTransaction.cancel();
+        });
```
SandPod, line 284: Question: What is the purpose of placing a timer on the lifetime of a transaction?

The lifetime of a test is configurable. We probably don't want to impose a test time limit with our framework.
Could this be removed? If not, it needs to be be configurable.

packages/serverpod_test/lib/src/serverpod_test.dart
```diff
@@ -0,0 +1,415 @@
...
+    await serverpodSession.db.unsafeExecute(
+      'SAVEPOINT $savePointId;',
+      transaction: _transaction,
+    );
+  }
+
+  String _getNextSavePointId() {
+    var postgresCompatibleRandomString =
+        const Uuid().v4obj().toString().replaceAll(RegExp(r'-'), '_');
+    var savePointId = 'savepoint_$postgresCompatibleRandomString';
```
SandPod, line 323: Suggestion: Add an index to the savepoint id so that they are easier to track.

packages/serverpod_test/lib/src/serverpod_test.dart
```diff
@@ -0,0 +1,415 @@
...
+) async {
+  try {
+    return await call();
+  } catch (e) {
+    throw _getException(e);
+  }
+}
+
+/// Test tools helper to not leak exceptions from functions that return streams.
+/// Used by the generated code.
```
SandPod, line 357: Suggestion: Add information that this enables the streams execution to start without having to wait for a listener.

One benefit in this case was to enable streams to run without having to think about subscriptions. We should clarify that here for future devs.

By the way, do we have a test validating this specific case?

packages/serverpod_test/lib/serverpod_test.dart
```diff
@@ -0,0 +1,3 @@
+library serverpod_test;
+
+export 'src/serverpod_test.dart';
```
SandPod, line 3: Question: Is it possible to split it up a bit to make this project a bit simpler to navigate?

tools/serverpod_cli/lib/src/generator/dart/library_generators/test_tools_generator.dart
```diff
@@ -0,0 +1,400 @@
+import 'package:code_builder/code_builder.dart';
+import 'package:path/path.dart' as p;
+import 'package:serverpod_cli/analyzer.dart';
+import 'package:serverpod_cli/src/analyzer/dart/definitions.dart';
+import 'package:serverpod_cli/src/generator/shared.dart';
+
+class TestToolsGenerator {
```
SandPod, line 7: Suggestion: Rename to `ServerTestToolsGenerator`.

Since this is not a general test tools generator and we later on might want some kind of "mock" that we generate.

tools/serverpod_cli/lib/src/test_util/builders/generator_config_builder.dart
```diff
@@ -103,6 +105,12 @@ class GeneratorConfigBuilder {
     return this;
   }
 
+  GeneratorConfigBuilder withRelativeServerTestToolsPathParts(
+      List<String> relativeServerTestToolsPathParts) {
```
SandPod, line 109: Suggestion(NIT): Add comma after last parameter.

```suggestion
  GeneratorConfigBuilder withRelativeServerTestToolsPathParts(
      List<String> relativeServerTestToolsPathParts,) {
```

tools/serverpod_cli/lib/src/test_util/builders/generator_config_builder.dart
```diff
@@ -42,7 +43,8 @@ class GeneratorConfigBuilder {
           ),
         ],
         _extraClasses = [],
-        _enabledFeatures = [ServerpodFeature.database];
+        _enabledFeatures = [ServerpodFeature.database],
+        _relativeServerTestToolsPathParts = ['integration_test', 'test_tools'];
```
SandPod, line 47: Suggestion: Remove default value here.

By default this is not generated, so this field should not be configured.
Should also have a test that validates nothing is generated when this is not set in the configuration.

SandPod (changes requested):
Awesome work with all of this.
I have a bunch of comments and questions.

Please look them through and ping me when it is time for a re-review or if you have any questions.

The changelog in the new project should inlcude all the other changelogs.

I think you can add these by running the `util/update_pubspecs` script

tests/serverpod_test_server/test_integration/test_tools/rollback_database_test.dart
```diff
@@ -0,0 +1,253 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+import 'package:test/test.dart';
+
+import 'serverpod_test_tools.dart';
+
+void main() {
+  group('Given rollbackDatabase set to afterEach', () {
```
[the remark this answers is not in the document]
hampuslavin: All the other tests would blow up if we did that, but you're right that this would be more explicit. I don't think it hurts to add this case 👍

tests/serverpod_test_server/test_integration/test_tools/rollback_database_test.dart
```diff
@@ -0,0 +1,253 @@
...
+
+          test(
+              'then automatically rolls back the transaction so that subsequent tests get the same setUp',
+              () async {
+            final result = await SimpleData.db.find(session);
+
+            expect(result.length, 2);
+
+            var nums = result.map((e) => e.num).toList();
+            expect(nums, containsAll([111, 222]));
```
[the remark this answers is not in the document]
hampuslavin: Ahh I changed these but apparently missed one, thanks!

tests/serverpod_test_server/test_integration/test_tools/rollback_database_test.dart
```diff
@@ -0,0 +1,253 @@
...
+    );
+
+    withServerpod(
+      (endpoints, session) {
+        tearDownAll(() async {
+          await SimpleData.db.deleteWhere(
+            session,
+            where: (t) => Constant.bool(true),
+          );
+        });
```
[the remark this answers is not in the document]
hampuslavin: This is a bit of a tricky scenario, but I think this is the simplest way to test it. This `withServerpod` just validates that data was _not_ cleaned up in the previous `withServerpod`.

This `withServerpod` doesn't create any data, so the rollback policy doesn't matter. The `tearDownAll` here cleans up data from the _previous_ `withServerpod` which did have `RollbackDatabase.never`.

packages/serverpod_test/lib/src/serverpod_test.dart
```diff
@@ -0,0 +1,415 @@
+import 'dart:async';
+
+import 'package:serverpod/serverpod.dart';
+import 'package:test/test.dart';
+
+export 'package:meta/meta.dart' show isTestGroup;
+
+/// A test specific session that is used to call database methods or pass to endpoints.
+abstract class TestSession implements DatabaseAccessor {
```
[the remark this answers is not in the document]
hampuslavin: Great suggestion!

packages/serverpod_test/lib/src/serverpod_test.dart
hampuslavin: That is one reason. Although the original reason was that the user-provided `testClosure` which takes the `InternalTestSession` has to be called inside our test `group`. Creating a Serverpod session is an `async` call, but test `group`s are not allowed to be async so that has to happen in the `setUpAll`. 

Maybe this could be done differently, but that seemed the most straight forward way!

packages/serverpod_test/lib/src/serverpod_test.dart
```diff
@@ -0,0 +1,415 @@
...
+
+        await testServerpod.shutdown();
+      });
+
+      testClosure(testServerpod.testEndpoints, mainTestSession);
+    });
+  };
+}
+
+class _TransactionManager {
```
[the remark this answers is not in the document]
hampuslavin: That makes total sense! 👍

packages/serverpod_test/lib/src/serverpod_test.dart
hampuslavin: Very easy to do, but I thought we said these aren't inspectable? Or do you mean if we need to debug this code in the future?

tools/serverpod_cli/lib/src/test_util/builders/generator_config_builder.dart
hampuslavin: Good point, will add that test! 🧪

tests/serverpod_test_server/test_integration/test_tools/authentication_validation_test.dart
```diff
@@ -0,0 +1,98 @@
...
+import 'serverpod_test_tools.dart';
+
+void main() {
+  group('Given AuthenticatedTestToolsEndpoint', () {
+    withServerpod(
+      (endpoints, session) {
+        test(
+            'when not authenticated and calling returnsString then throws UnauthenticatedEndpointCallTestException',
+            () async {
+          session = await session.copyWith(getAuthenticationInfo: () => null);
```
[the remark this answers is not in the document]
hampuslavin: [Discussed offline]: Agreed to rename this field simply `authentication` and simulate a union type by creating an abstract base class which we create appropriate subclasses of.

tests/serverpod_test_server/test_integration/test_tools/rollback_database_test.dart
```diff
@@ -0,0 +1,253 @@
...
+  group('Given rollbackDatabase set to afterEach', () {
+    group('when creating objects in a setUpAll', () {
+      withServerpod(
+        (endpoints, session) {
+          setUpAll(() async {
+            await SimpleData.db
+                .insert(session, [SimpleData(num: 111), SimpleData(num: 222)]);
+          });
+
+          test('then creates the objects in the database according to setUpAll',
```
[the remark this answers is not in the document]
hampuslavin: Good catch!

tests/serverpod_test_server/test_integration/database_operations/crud/create_test.dart
```diff
@@ -3,18 +3,28 @@ import 'package:serverpod_test_server/src/generated/protocol.dart';
...
 
+  tearDownAll(() async {
+    await SimpleData.db.deleteWhere(
+      session,
+      where: (t) => Constant.bool(true),
+    );
+    await EmptyModelWithTable.db.deleteWhere(
+      session,
+      where: (t) => Constant.bool(true),
+    );
```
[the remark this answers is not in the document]
hampuslavin: The other tear down is only for the specific group, but I can indeed enhance this code a bit. Will create groups and put relevant clean up in there.

packages/serverpod_test/.gitignore
```diff
@@ -0,0 +1,3 @@
+# https://dart.dev/guides/libraries/private-files
+# Created by `dart pub`
+.dart_tool/
```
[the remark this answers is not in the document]
hampuslavin: You're completely right, it can be removed 👍

packages/serverpod_test/lib/src/serverpod_test.dart
hampuslavin: This can be removed 👍

packages/serverpod_test/lib/serverpod_test.dart
hampuslavin: Good call out! I should definitely break this file up.

tests/serverpod_test_server/test_integration/test_tools/reset_test_session_test.dart
```diff
@@ -0,0 +1,100 @@
...
+
+            expect(newSession.authenticationInfo!.userId, 123);
+          });
+
+          test('then the second test has the reset newSession', () async {
+            expect(newSession.authenticationInfo, isNull);
+          });
+        });
+      },
+      resetTestSessions: ResetTestSessions.afterEach,
```
christerswahn, line 57: I think it would be really valuable to have these at the top of the test instead of the bottom. Reading test cases is naturally done top-down and having these semantics defined at the end seems to risk a gotcha everytime they're not the default.. 

In order to get around the problem with default params and named params having to be last, a couple of approaches could be considered:

- Leaving `withServerpod()` as-is, but adding wrapper functions with clear names that apply these settings, such as `withServerpodResetEach()`. The drawback is somewhat weird and long names and if there are many common permutations this gets unwieldy.

- Making the main argument function (endpoints, session){} a named parameter with a simple name like func. Not much more to write but clear and flexible:
```
    withServerpod(
      resetTestSessions: ResetTestSessions.afterEach,
      runMode: ServerpodRunMode.production,
      func: (endpoints, session) {
        ...
      },
    )
```

tests/serverpod_test_server/test_integration/test_tools/authentication_validation_test.dart
```diff
@@ -0,0 +1,98 @@
...
+
+          final result =
+              endpoints.authenticatedTestTools.returnsString(session, "Hello");
+          await expectLater(
+              result, throwsA(isA<UnauthenticatedEndpointCallTestException>()));
```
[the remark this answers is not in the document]
hampuslavin: Since we're stripping out the transport layer, we basically have to decide wether we'd like to either expose internal exceptions from Serverpod or create test domain specific ones. I opted for the latter here, since it seems more prudent to have a separation and avoid user dependencies on internal Serverpod Exceptions.

These two exceptions are mapping 1:1 to internal Serverpod exceptions (see `packages/serverpod_test/lib/src/function_call_wrappers.dart`) so their only function is to create a layer of indirection.

tests/serverpod_test_server/test_integration/test_tools/reset_test_session_test.dart
hampuslavin: This is great feedback and those are valid alternatives. The two main arguments for the current setup are:
- I think the 95% percentile use case will have _no_ options; so "no options" is the case that should be optimized for and should not contain unnecessary "boiler plate".
- Dart `test` and `group` have their options last just like this, and keeping the same structure will provide familiarity for the user and hopefully be more intuitive.
- (We're specifying `runMode` in each test, but this should actually be extremely rare in the wild, since going forward there shouldn't be many use cases for not using `test` mode. Our `serverpod_test_server.dart` was currently using `prod` so I opted here to not use `test` in some tests and `prod` in all current tests).

Would be great to get input from @Isakdl and @vlidholt as well!

packages/serverpod_test/lib/src/serverpod_test.dart
```diff
@@ -0,0 +1,415 @@
...
+    bool? enableLogging,
+  });
+}
+
+/// Internal test session that contains implementation details that should
+/// only be used internally by the test tools.
+class InternalTestSession extends TestSession {
+  final List<InternalTestSession> _allTestSessions;
+  final TestServerpod _testServerpod;
+  AuthenticationInfo? _authenticationInfo;
```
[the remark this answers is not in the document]
hampuslavin: As per other comment, this will instead be a "union" type.

packages/serverpod/lib/src/server/log_manager/log_manager.dart
```diff
@@ -127,7 +127,8 @@ class SessionLogManager {
       order: _nextLogOrderId,
     );
 
-    if (_session.serverpod.runMode == ServerpodRunMode.development) {
+    if (_session.serverpod.runMode == ServerpodRunMode.development ||
+        _session.serverpod.runMode == ServerpodRunMode.test) {
```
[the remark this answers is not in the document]
hampuslavin: I actually realized now that this (and on other place) doesn't respect the "enable logging" setting because it writes directly to `stdout`. I will remove this I think and we can revisit this.

packages/serverpod_test/lib/src/function_call_wrappers.dart
```diff
@@ -0,0 +1,76 @@
...
+}
+
+dynamic _getException(dynamic e) {
+  switch (e) {
+    case NotAuthorizedException():
+      return switch (e.authenticationFailedResult.reason) {
+        AuthenticationFailureReason.unauthenticated =>
+          UnauthenticatedEndpointCallTestException(),
+        AuthenticationFailureReason.insufficientAccess =>
+          InsufficientEndpointAccessTestException(),
```
Isakdl, line 63: Suggestion: As discussed offline, it would be nice to align the clientside exceptions with the exceptions in the tests.

tests/serverpod_test_server/test_integration/test_tools/database_operations_in_endpoint_test.dart
```diff
@@ -0,0 +1,76 @@
...
+    (endpoints, session) {
+      group('Given TestToolsEndpoint', () {
+        test(
+            'when calling createSimpleData then creates a SimpleData in the database',
+            () async {
+          await endpoints.testTools.createSimpleData(session, 123);
+
+          final result = await SimpleData.db.find(session);
+          expect(result.length, 1);
+          expect(result.first.num, 123);
```
Isakdl, line 18: Suggestion: Add an internal copyWith on the session inside the test method call. This is to force a new session id and a new state for things such as the logger to mimic the production behaviour and not trick people into thinking the session is persistent between calls.

tests/serverpod_test_server/test_integration/test_tools/reset_test_session_test.dart
Isakdl: I agree that you get a better overview with the option in the top, but as you point out it is a good idea to mirror the structure that already exists in the dart test tooling. I'm in favor of keeping it as is for this reason only.

tests/serverpod_test_server/test_integration/test_tools/database_operations_in_endpoint_test.dart
hampuslavin: [a84e1c5](https://github.com/serverpod/serverpod/pull/2713/commits/a84e1c5bf51d08574f40bf118f5e942011e97729)

packages/serverpod_test/lib/src/serverpod_test.dart
```diff
@@ -0,0 +1,415 @@
...
+  final bool _enableLogging;
+
+  /// The underlaying Serverpod session
+  late Session serverpodSession;
+
+  @override
+  Database get db => serverpodSession.db;
+
+  @override
+  AuthenticationInfo? get authenticationInfo => _authenticationInfo;
```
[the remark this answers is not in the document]
hampuslavin: [a84e1c5](https://github.com/serverpod/serverpod/pull/2713/commits/a84e1c5bf51d08574f40bf118f5e942011e97729)

packages/serverpod/lib/src/server/session.dart
```diff
@@ -87,7 +86,12 @@ abstract class Session {
   /// Access to the database.
   Database? _db;
 
+  /// Optional transaction to use for all database queries.
+  @override
+  Transaction? transaction;
```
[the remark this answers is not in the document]
hampuslavin: [a84e1c5](https://github.com/serverpod/serverpod/pull/2713/commits/a84e1c5bf51d08574f40bf118f5e942011e97729)

tests/serverpod_test_server/test_integration/test_tools/rollback_database_test.dart
```diff
@@ -387,4 +388,19 @@ void main() {
...
+  //         () async {
+  //       final result = await SimpleData.db.find(session);
+
+  //       expect(result.length, 2);
+  //       expect(result[0].num, 111);
+  //       expect(result[1].num, 222);
+  //     });
+  //   },
+  //   runMode: ServerpodRunMode.production,
+  // );
```
hampuslavin, line 405: Mistake, will fix in next commit

tests/serverpod_test_server/test_integration/test_tools/rollback_database_test.dart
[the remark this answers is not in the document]
hampuslavin: Bug fixed in separate PR: https://github.com/serverpod/serverpod/pull/2749

packages/serverpod_test/lib/src/function_call_wrappers.dart
hampuslavin: @Isakdl I think we ended up at the conclusion that the client exceptions probably should be named something that is not tied 1:1 to HTTP response statuses. So the question is, are these names okay or do we still want to change them?

Isakdl: Yes that is fair, but I do think they should map 1:1 with the names and exceptions we use in the client. Even if that means renaming / extending new exceptions for the client.

packages/serverpod_test/lib/src/test_session.dart
```diff
@@ -0,0 +1,142 @@
...
+        _authenticationOverride = authenticationOverride,
+        _testServerpod = testServerpod,
+        _enableLogging = enableLogging,
+        _transaction = transaction {
+    _allTestSessions.add(this);
+    _configureServerpodSession(serverpodSession);
+  }
+
+  @override
+  TestSession copyWith({
```
Isakdl, line 95: Comment: As discussed offline this name may not exactly be the expected behavior. Not sure if I got a better name in mind, but a more accurate name is something like `createFrom` or similar.

packages/serverpod_test/lib/src/test_session.dart
```diff
@@ -0,0 +1,142 @@
...
+  /// and recreates the underlying Serverpod session.
+  Future<void> resetState() async {
+    await serverpodSession.close();
+    _authenticationOverride = null;
+    serverpodSession = _testServerpod.createSession(
+      transaction: _transaction,
+      enableLogging: _enableLogging,
+    );
+    _configureServerpodSession(serverpodSession);
+  }
```
Isakdl, line 135: Question: Where is this used? This will also modify the id of the session! Perhaps that is ok?

packages/serverpod_test/lib/src/test_session.dart
hampuslavin: Lets discuss together offline!

packages/serverpod_test/lib/src/function_call_wrappers.dart
hampuslavin: [b88616b](https://github.com/serverpod/serverpod/pull/2713/commits/b88616b98a5f3eff87e430f097f3deb2d82afe48)

Isakdl (approved):
This is huge! Been on my whish list since day one ❤️!

LGTM 🚀 🚀 🚀

Issue #2008, opened 2024-03-13
RFC: Improvements to webserver


vlidholt:
Serverpod comes with a built-in webserver called Relic. It's still experimental, and it would be nice to clean it up and make it top-notch, like the rest of Serverpod. 🚀

There are many use cases for the web server:

- Serve a Flutter web app.
- Build websites that use the same database as the API server.
- Create webhooks.
- Create REST APIs that look a specific way. I.e., complete control over status code, request types, headers, etc.
- 
The main benefit of having a good web server bundled with Serverpod is that developers can use the Serverpod sessions and ORM. This means that a lot of code can be shared between the generated API and a website or custom API.

## Improvements
The overall architecture for Relic looks sound, but it would be nice to improve in certain areas.

- **Routing**. Providing an easier way to pattern-match paths and automatically extract components from them would be nice. We can take inspiration from packages like Shelf, Dart Frog, and Pharaoh here. Looking at packages from other languages would also be smart. Let's be best in class here!
- **Middleware**. Add support for middleware, as it can be used for things like custom logging, authentication, data validation, error handling, or other logic. Again, we should look at other frameworks to do something really good here.
- **Authentication**. It should be possible (through pre-built middleware?) to access the same authentication system as for Serverpod's generated endpoints.
- **Sharing static files**. There is a route made for sharing a directory of static files. It works, but the API could be clearer and easier to understand. We should also support more MIME types. It should also be possible to add additional types.
- **Make Flutter web easy**. Serving a Flutter app may be the most common reason to use Relic. We should make it very simple to set up, compile, and host a Flutter web app.
- **Support modules**. Make it possible to place web routes in Serverpod modules. This would make building modules for integrations like payments requiring custom webhooks easier.
- **Tests**. Improve the test suite for Relic. There are a couple of tests, but it would be nice to cover a larger surface.
Documentation. Add examples and proper documentation for the webserver.

## Ideas for the future
These are ideas that we can add in the future without breaking any of the new core features.

- **Improved templates**. Relic has a template engine that uses the Mustache library. A template combined with a Dart class forms a `Widget`,  which is analogous to a Flutter widget. It would be cool to support Relic widgets that resemble some of the static Flutter widgets (`Container`, `Row`, `Padding`, `Text`, etc). We could also generate widgets (possibly without a template file) that correspond to all the default tags in HTML.

codekeyz:
[...]

PR #2744, opened 2024-09-12
New version of the Relic webserver - adopted from Shelf

klkucaj:
This PR introduces **Relic**, a lightweight and strictly typed HTTP server designed to work seamlessly with [[Serverpod](https://serverpod.dev/)](https://serverpod.dev). Inspired by Shelf, Relic focuses on type safety, performance, and modularity while maintaining extensibility.

#### **Core Library (`lib/`)**
- **`relic.dart`**:
  - The primary entry point of the library, exposing the core server functionalities.
  
- **`lib/src/`**:
  This directory organizes the core logic into specialized modules to ensure clarity, scalability, and maintainability.

  - **Core Server Files**:
    - `relic_server.dart`: Implements the primary server functionality backed by Dart's `HttpServer`, handling connections, routing, and core features.
    - `relic_server_serve.dart`: Provides functionality for serving HTTP requests, with support for advanced features like request hijacking.

  - **Modules**:
    - **`handler/`**: Implements mechanisms for flexible request routing, delegation, and pipelines for chaining handlers.
    - **`middleware/`**: Contains utilities to integrate middleware, including extensions and logging functionality for monitoring requests and responses.
    - **`logger/`**: Provides structured logging for better observability and debugging.
    - **`util/`**: Shared utility functions for common operations within the library.
    - **`body/`**: Handles parsing, validation, and type-checking of request/response bodies.
    - **`message/`**: Manages HTTP message formatting, including `request.dart` and `response.dart` for processing requests and responses.
    - **`headers/`**:
      The **headers** module is a central component of the Relic library, offering robust functionality for handling HTTP headers:
      - **Core Functionality**:
        - `headers.dart`: Provides the main abstraction for managing HTTP headers, integrating with various submodules.
        - `extension/`: Adds utility methods for working with headers represented as strings, enabling ease of use and flexibility.
      - **Parsing**:
        - `parser/`: Includes `headers_parser.dart` and `common_types_parser.dart` to convert raw headers into structured, validated formats.
      - **Type-Safe Headers**:
        - `typed/`: Offers predefined and user-defined typed headers, improving reliability and ensuring correctness.
      - **Custom Headers**:
        - `custom/`: Allows for the definition and management of application-specific headers.
      - **Error Handling**:
        - `exception/`: Defines error types for invalid or improperly formatted headers, ensuring reliability.
    - **`static/`**: Implements the logic for serving static files, including custom directory listings and optimizations for asset handling.
    - **`hijack/`**: Provides request hijacking for advanced use cases requiring low-level control.
    - **`method/`**: Manages and validates HTTP methods, ensuring adherence to standard protocols.

### **Test Suite (`test/`)**
The `test/` directory includes a comprehensive set of unit and integration tests that validate Relic's features and reliability.

- **Core Test Files**:
  - `relic_server_test.dart`: Tests the primary server's functionality, including routing and connection handling.
  - `relic_server_serve_test.dart`: Focuses on validating the server's `serve` logic and advanced features like request hijacking.

- **Subdirectories**:
  - **`handler/`**: Verifies the routing, delegation, and pipeline mechanisms.
  - **`middleware/`**: Tests middleware integration, extensions, and logging functionality.
  - **`util/`**: Tests utility functions for correctness and performance.
  - **`message/`**: Focuses on request and response message formatting and parsing.
  - **`headers/`**:
    - Validates all aspects of the `headers/` module, including:
      - Parsing and validation of standard and custom headers.
      - Error handling for malformed or invalid headers.
      - Typed headers and extensions.
  - **`static/`**: Ensures proper handling of static file serving and directory listings.
  - **`hijack/`**: Tests advanced request hijacking logic for robustness.

Related: [#2008](https://github.com/serverpod/serverpod/issues/2008)

## Pre-launch Checklist

## Breaking changes


vlidholt:
Some notes for the review. We are primarily focusing on the code in the new `relic` package. The `relic` package is based on Dart Shelf, but with adding type safety, properties in HTTP headers, and improved `Request` and `Response` objects.

The changes in Serverpod's webserver (within the `serverpod` package) are just to make sure everything still works. The interface for these will be improved upon in the next set of PRs.

packages/relic/lib/src/util/util.dart
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  return value;
+}
+
+/// Adds a header with [name] and [value] to [headers], which may be null.
+///
+/// Returns a new map without modifying [headers].
+Map<String, Object> addHeader(
+  Map<String, Object>? headers,
```
vlidholt, line 51: Is this used? If so it should be typed (`List<String>` instead of `Object`). If it's no longer used it should be removed.

packages/relic/lib/src/util/util.dart
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:collection/collection.dart';
+import 'package:relic/src/map/shelf_unmodifiable_map.dart';
+
+/// Run [callback] and capture any errors that would otherwise be top-leveled.
```
vlidholt, line 10: It looks like these methods shouldn't be global, but belong to a class or possibly and extension on a class. If they are no longer used, they should be removed.

packages/relic/test/ssl_certs.dart
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
vlidholt, line 1: This file will need to be excluded in `.pubignore` or relic may not publish to pub (see the auth package).

packages/serverpod/lib/src/relic/static/static_handler.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:serverpod/serverpod.dart';
```
vlidholt, line 1: Ok, for now IMO, but this may need to be reworked with the new routing update.

packages/serverpod/lib/src/relic/web_server.dart
```diff
@@ -4,12 +4,22 @@ import 'dart:io';
...
   final Serverpod serverpod;
 
+  /// If a security context is provided an HTTPS server will be started.
+  final SecurityContext? securityContext;
+
+  /// The [address] can either be a [String] or an [InternetAddress]. If
+  /// [address] is a [String], the server will perform a
+  /// [InternetAddress.lookup] and use the first value in the list. Defaults to
+  /// [InternetAddress.anyIPv6].
+  final Object? address;
```
vlidholt, line 21: I don't really like the unspecified type, but it may be ok as the underlaying `HttpServer` uses `dynamic`.

vlidholt (changes requested):
Great work on this one! Lot's of great updates. I did a quick pass, but I think we'll need a second pair of eyes on this one.

There are some things that need to be fixed throughout the PR:

- There still may be code in here that is no longer used.
- A lot of the documentation is outdated. It need to be brought up to par with all the changes that have been made.
- There is a lot of missing documentation. I suggest that we do a second PR for this, once this one has been merged, where we move to the Serverpod linting rules (which also requires docs for everything).

There are also changes that will need to be made to the Serverpod side of things in this PR, but it may be better to do a review of that when we implement the improved routing.

packages/serverpod/lib/src/relic/web_server.dart
hampuslavin: +1 Would also prefer that this was typed.

hampuslavin (commented):
Left some initial comments, will have to do a second pass to dig into some more details 👍

packages/relic/lib/src/headers.dart
```diff
@@ -0,0 +1,1219 @@
...
+      wwwAuthenticate: wwwAuthenticate is List<String>
+          ? wwwAuthenticate
+          : this.wwwAuthenticate,
+      custom: custom ?? this.custom,
+    );
+  }
+
+  @override
+  @override
+  Headers withHeaders(Headers? other) {
```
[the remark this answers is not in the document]
klkucaj: done!

packages/serverpod/lib/src/relic/static/static_handler.dart
klkucaj: Fixed!

packages/serverpod/lib/src/relic/static/static_handler.dart
```diff
@@ -0,0 +1,61 @@
...
+      normalizedMountedPath = '$normalizedMountedPath/';
+    }
+
+    // Check if the request path starts with the mounted path
+    if (request.requestedUri.path.startsWith(normalizedMountedPath)) {
+      // Safely replace the mounted path prefix
+      var modifiedPath = requestedPath
+          .replaceFirst(normalizedMountedPath, '/')
+          // Prevent double slashes in the resulting path
+          .replaceAll(RegExp(r'//+'), '/');
```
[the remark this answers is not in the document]
klkucaj: @hampuslavin 
Since I'm using a raw string (`r''`), the forward slashes don't need to be escaped, so `RegExp(r'//+')` works also. But you're right, escaping them explicitly with `RegExp(r'\/+')` makes it clearer.

Should I go ahead and change it for clarity?

packages/relic/lib/src/headers.dart
```diff
@@ -0,0 +1,1219 @@
+import 'dart:collection';
```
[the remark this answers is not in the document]
klkucaj: @hampuslavin 
In the `headers` folder, I only include custom headers and the authorization header. The idea is to keep everything custom that's used specifically within this file, so this it the main headers file.
 Do you think it would be a good idea to include this as well?

packages/serverpod/lib/src/relic/static/static_handler.dart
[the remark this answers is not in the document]
hampuslavin: Sorry, I thought that a `/` was a reserved regex token but it is not! You can disregard this comment, better to keep as is 👍

packages/relic/lib/src/headers.dart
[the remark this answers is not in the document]
hampuslavin: No I think it's fine to keep as is 👍

packages/relic/lib/src/headers.dart
```diff
@@ -0,0 +1,1219 @@
...
+    List<String>? allow,
+    String? contentDisposition,
+    AuthorizationHeader? authorization,
+    List<String>? cacheControl,
+    List<String>? connection,
+    List<String>? contentEncoding,
+    List<String>? contentLanguage,
+    String? contentLocation,
+    String? contentMD5,
+    String? contentRange,
```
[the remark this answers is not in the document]
klkucaj: done

packages/relic/lib/src/body.dart
```diff
@@ -0,0 +1,113 @@
...
+  }) {
+    return Body._(
+      body,
+      null,
+      encoding: encoding,
+      mimeType: mimeType,
+    );
+  }
+
+  factory Body.fromIntStream(
```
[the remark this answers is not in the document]
klkucaj: We need it for https://github.com/serverpod/serverpod/pull/2744/files#diff-1b7bafb0ff8a0c00a7c583bf5c11799b8778893308a3db5ff8d1b951726ccd58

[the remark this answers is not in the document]
klkucaj: I used it here packages/relic/lib/src/static/static_handler.dart

Screenshot 2024-09-19 at 3 56 28 PM
Screenshot 2024-09-19 at 3 56 36 PM

packages/relic/lib/src/body.dart
```diff
@@ -0,0 +1,113 @@
...
+    Stream<List<int>> body, {
+    Encoding? encoding = utf8,
+    MimeType mimeType = MimeType.plainText,
+  }) {
+    Stream<Uint8List> byteStream = body.map((list) => Uint8List.fromList(list));
```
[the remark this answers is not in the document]
klkucaj: Already left a comment here https://github.com/serverpod/serverpod/pull/2744#discussion_r1766889467

vlidholt (changes requested):
Great work on this. We are closing in! 🥳

- There are still many headers that are using `String` over the correct type. I commented on two places, but there are more.
- We are missing tests for the new header types. We should definitely have unit tests for parsing these.

I haven't yet, played around with this in a real world setting. Creating a new project and doing some manual testing. So there may be more comments once I do that.

packages/relic/analysis_options.yaml
```diff
@@ -0,0 +1,30 @@
+# This file configures the static analysis results for your project (errors,
```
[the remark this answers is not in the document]
klkucaj: Done!

packages/relic/lib/src/body.dart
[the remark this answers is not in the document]
klkucaj: Done!

packages/relic/lib/src/request.dart
```diff
@@ -0,0 +1,404 @@
...
+  ///
+  /// [url]'s path is always relative. It may be empty, if [requestedUri] ends
+  /// at this handler. [url] will always have the same query parameters as
+  /// [requestedUri].
+  ///
+  /// [handlerPath] and [url]'s path combine to create [requestedUri]'s path.
+  final Uri url;
+
+  /// The HTTP request method, such as "GET" or "POST".
+  final Method method;
```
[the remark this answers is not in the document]
klkucaj: @vlidholt is this okay?

https://github.com/serverpod/serverpod/blob/a7c20c01d8a305962c057f4d40afa205b2b71147/packages/relic/lib/src/method/method.dart#L1-L31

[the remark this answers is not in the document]
klkucaj: @vlidholt is this good ?

https://github.com/serverpod/serverpod/blob/17a82f96ebc7f716e6dc43d4b022ba428a5ea981/packages/relic/lib/src/method/method.dart#L1-L50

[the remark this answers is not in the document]
vlidholt: The `parse` method must be able to parse custom methods, otherwise there is no reason to convert this to a class?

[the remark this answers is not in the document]
klkucaj: Done!

packages/relic/lib/src/headers/util/headers_util.dart
```diff
@@ -0,0 +1,10 @@
+part of '../../headers.dart';
+
+/// Helper method to check if a string is valid Base64.
+///
+/// This method is used internally to validate that the Base64-encoded credentials
+/// are correctly formatted before attempting to decode them.
+bool _isValidBase64(String str) {
```
vlidholt, line 7: Isn't this better as an extension to a `String`. Also, no need to make it private, this can be useful as a part of the public interface.

packages/relic/lib/src/headers/vary_header.dart
```diff
@@ -0,0 +1,65 @@
+part of '../headers.dart';
+
+/// A class representing the HTTP Vary header.
+///
+/// This class manages the list of headers that the response may vary on.
+/// It provides functionality to parse, manipulate, and generate Vary header values.
+class VaryHeader {
+  /// A list of headers that the response varies on.
+  final List<String> fields;
```
vlidholt, line 9: The name in the specification is `header-name`, not `field`. Also, this is just a `List<String>`, I think we should use that in our header, rather than wrapping it in a custom class. If we are keeping the class. We should add a second constructor `VaryHeader.uncachable()` that returns `['*']`.

packages/relic/lib/src/request.dart
```diff
@@ -0,0 +1,410 @@
...
+  /// An adapter can check whether a request was hijacked using [canHijack],
+  /// which will be `false` for a hijacked request. The adapter may throw an
+  /// error if a [HijackException] is received for a non-hijacked request, or if
+  /// no [HijackException] is received for a hijacked request.
+  ///
+  /// See also [hijack].
+  Request(
+    Method method,
+    Uri requestedUri, {
+    List<io.Cookie> cookies = const [],
```
vlidholt, line 118: Cookies should be part of the header. We should provide our own `CookieHeader` class and not rely on `dart:io`, as that may break things in the future.

vlidholt (changes requested):
## Important

Before starting to work on this, we are awaiting a second review by @SandPod.

## `parse` / `tryParse`

- `tryParse` should never throw an exception. If an error occurs, it should return null.
- `parse` can throw exceptions.

The parser methods should not take a nullable parameter. Instead, that should be handled in the calling code. Dart has generally good constructs for this. E.g., see `num` for reference:
https://api.dart.dev/stable/2.7.2/dart-core/num-class.html

This should be corrected everywhere in our code. Which means that there may be methods that should be renamed to `parse`.

## Headers

Many of the parsed headers do not match the HTTP specification. I marked a few, but they all need a second look. Use Mozilla as reference. (See the _HTTP headers_ section.)
https://developer.mozilla.org/en-US/docs/Web/HTTP

All examples listed for each header should work, at a minimum. We should include the examples in the tests.

Where possible, we should name the header parsers more by their type, rather than the name of the header. There may be code we can reuse between headers.

### x-powered-by header

The `xPoweredBy` field in the header is not used. In fact, it cannot be used in any way currently. We should use the `poweredByHeader` field parameter passed into the `mount` method to create the default `Response` objects, if possible. If it’s not possible, second best is to use the `xPoweredBy` field to override the value passed into `mount`.

The default value for the `poweredByHeader` should be `Relic`.

### Naming of parse methods

The naming is inconsistent for the different header types. Sometimes it's `fromHeaderValue` sometimes it's `parse` or `tryParse`. We should go with just one consistent name. I say `parse` and `tryParse` should be good.

## Data efficiency

We should use `Uint8List` everywhere in our code and in our interfaces. If there is an external interface using the old `List<int>` we should verify that it is in fact a `Uint8List` under the hood, otherwise we throw an exception. This is one of the biggest improvements I made compared to shelf, not sure why these changes was reverted?

It’s possible to cast a stream (without making a copy or copy the underlaying data). E.g.

```dart
Stream<List<int>> bar = Stream.empty();
Stream<Uint8List> foo = bar.cast();
```

## Misc

- The `util/pub_get_all` needs to be updated to include `relic`.

## Overall

Many of the comments I added may apply to many places in the code. Make sure that the changes reflect all places before we do another review.

klkucaj:
**x-powered-by header**:
I'm not sure if the 'X-Powered-By' header should be changeable. I believe this header should remain consistent after the server is mounted, though I could be mistaken.

**Data efficiency**:
I tried doing this the same way you described but i got the casting error, i will check it again.

packages/relic/lib/src/headers/accept_header.dart
```diff
@@ -0,0 +1,49 @@
...
+
+    return AcceptHeader(mediaTypes);
+  }
+
+  /// Static method that attempts to parse the Accept header and returns `null` if the value is `null`.
+  ///
+  /// This method safely parses the Accept header value or returns `null` if the input is invalid or `null`.
+  static AcceptHeader? tryParse(List<String>? value) {
+    if (value == null) return null;
+    return AcceptHeader.fromHeaderValue(value);
```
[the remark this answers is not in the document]
klkucaj: @vlidholt The idea is that if the value is null, we should return null. However, if it's not null and has some value, we should throw an error because the value is invalid. If they provide a value but see the header as null, they won’t understand what's going on.
What do you suggest i do here ?

packages/relic/lib/src/relic_server.dart
```diff
@@ -80,25 +81,33 @@ class RelicServer {
   /// Mounts a handler to the server. Only one handler can be mounted at a time.
   void mount(
     Handler handler, {
+    ExceptionHandler? exceptionHandler,
```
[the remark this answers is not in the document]
klkucaj: @vlidholt what if someone wants to return a different response when an error is thrown from the server ? 
Check how i used it for the tests:

https://github.com/klkucaj/serverpod/blob/795edf06b0025af58d8f0bbacf1e097e536eef0c/packages/relic/test/static/test_util.dart#L103-L106

packages/relic/lib/src/headers/accept_header.dart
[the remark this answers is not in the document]
vlidholt: Please see my overarching comments. A `tryParse` should never throw an exception. Having the null check inside the parse method adds little value. The parsers shouldn't take nullable values, instead do the null check when calling the method. E.g., with `?.`.

packages/relic/lib/src/relic_server.dart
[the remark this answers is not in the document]
vlidholt: This new from shelf. We may want to add something like this in the future, but then it should be a separate PR, with an issue that explains the problem so we can discuss the best way to solve implementation.

vlidholt:
> **x-powered-by header**: I'm not sure if the 'X-Powered-By' header should be changeable. I believe this header should remain consistent after the server is mounted, though I could be mistaken.

With the current implementation, the `X-Powered-By` has no function in the `Headers` class. If changing the header has no effect, it's just confusing to have it there. Making it possible to override the "main" setting makes the server more flexible, so why not?

> **Data efficiency**: I tried doing this the same way you described but i got the casting error, i will check it again.

The code I had written for this did work, so it should be possible. We should only use `Uint8List`. If the cast fails, that means that the underlaying stream is a `List<int>`, which is very bad and something we need to fix. What is causing the issue?

klkucaj:
@vlidholt file.openRead() is a Stream<List<int>> 

Screenshot 2024-10-09 at 11 16 27 AM


https://github.com/klkucaj/serverpod/blob/795edf06b0025af58d8f0bbacf1e097e536eef0c/packages/relic/lib/src/static/static_handler.dart#L218-L219

packages/relic/lib/src/body.dart
```diff
@@ -0,0 +1,120 @@
...
+      body.length,
+      encoding: encoding,
+      mimeType: mimeType,
+    );
+  }
+
+  /// Returns a [Stream] representing the body.
+  ///
+  /// Can only be called once.
+  Stream<List<int>> read() {
```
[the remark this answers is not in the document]
klkucaj: Reverted back to   Stream<List<int>>? _stream;

klkucaj:
@vlidholt your you are right!

packages/relic/lib/src/headers/exception/invalid_header_value_exception.dart
```diff
@@ -0,0 +1,9 @@
+/// Custom exception for invalid header values.
+class InvalidHeaderValueException implements Exception {
```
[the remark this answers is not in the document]
klkucaj: yes, ur right

packages/relic/lib/src/headers/from_header.dart
```diff
@@ -0,0 +1,58 @@
+part of '../headers.dart';
+
+/// A class representing the HTTP `From` header.
+///
+/// The `From` header is used to indicate the email address of the user making the request.
+/// It usually contains a single email address, but in edge cases, it could contain multiple
+/// email addresses separated by commas.
+class FromHeader {
```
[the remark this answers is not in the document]
klkucaj: @vlidholt should the class name contain the header key name or what it handles ? This is a question also for all headers.

packages/relic/lib/src/headers/special/accept_encoding_header.dart
```diff
@@ -0,0 +1,55 @@
...
+    var parts = values.splitTrimAndFilterUnique();
+
+    if (parts.isEmpty) {
+      throw FormatException(
+        'Invalid Accept-Encoding header: value cannot be empty',
+      );
+    }
+
+    var encodings = parts.map((e) {
+      var parts = e.split(';q=');
```
SandPod, line 24: Suggestion(nit): Defininig a variable named parts while iterating a map called parts. Please clarify what is what.

packages/relic/lib/src/headers/special/accept_encoding_header.dart
```diff
@@ -0,0 +1,55 @@
...
+      };
+      return EncodingQuality(encoding, quality);
+    }).toList();
+
+    return AcceptEncodingHeader(encodings);
+  }
+
+  /// Converts the [AcceptEncodingHeader] instance into a string representation suitable for HTTP headers.
+  @override
+  String toString() => encodings.map((e) => e.toString()).join(', ');
```
SandPod, line 38: Question: Should we have a prefix here that explains these encodings are part of the `AcceptEncodingHeader`? Just thinking if we print multiple headers at the same time that contain encodings it will not be clear that these specific encodings are part of the `AcceptEncodingHeader`.

packages/relic/lib/src/headers/special/accept_encoding_header.dart
```diff
@@ -0,0 +1,55 @@
...
+/// A class representing an encoding with an optional quality value.
+class EncodingQuality {
+  /// The encoding value.
+  final String encoding;
+
+  /// The quality value (default is 1.0).
+  final double quality;
+
+  /// Constructs an instance of [EncodingQuality].
+  EncodingQuality(this.encoding, [this.quality = 1.0]);
```
SandPod, line 50: Suggestion: Make quality a nullable parameter and don't assign it on the outside.

Since we have logic (in the toString) statement regarding the value of the quality, we should keep that codification on the inside.

packages/relic/lib/src/headers/special/accept_encoding_header.dart
```diff
@@ -0,0 +1,55 @@
...
+
+  /// Converts the [EncodingQuality] instance into a string representation suitable for HTTP headers.
+  @override
+  String toString() => quality == 1.0 ? encoding : '$encoding;q=$quality';
```
SandPod, line 54: Suggestion: Separate the human readable to string method from the "header" encoding method.

Feels like there is a risk we introduce something to help us debug that is not valid when encoding for header. Better to separate these.

packages/relic/test/headers/new/special/accept_encoding_header_test.dart
```diff
@@ -0,0 +1,331 @@
...
+        );
+
+        expect(
+          headers.acceptEncoding?.encodings.map((e) => e.encoding).toList(),
+          equals(['gzip', 'deflate', 'br']),
+        );
+        expect(
+          headers.acceptEncoding?.encodings.map((e) => e.quality).toList(),
+          equals([1.0, 1.0, 1.0]),
+        );
```
SandPod, line 37: Suggestion: Separate the default value validation from the "encodings is parsed correctly" test.

packages/relic/test/headers/new/special/accept_encoding_header_test.dart
```diff
@@ -0,0 +1,331 @@
...
+        server = await RelicServer.bind(InternetAddress.loopbackIPv6, 0);
+      } on SocketException catch (_) {
+        server = await RelicServer.bind(InternetAddress.loopbackIPv4, 0);
+      }
+    });
+
+    tearDown(() => server.close());
+
+    test(
+      'when a valid Accept-Encoding header is passed then it should parse correctly',
```
SandPod, line 22: Suggestion: Create separate test for single Accept-Encoding and multiple Accept-Encoding.

packages/relic/test/headers/new/special/accept_encoding_header_test.dart
```diff
@@ -0,0 +1,331 @@
+import 'dart:io';
+import 'package:test/test.dart';
+import 'package:relic/src/headers/headers.dart';
+import 'package:relic/src/relic_server.dart';
+import '../../headers_test_utils.dart';
+
+void main() {
+  group('Given an Accept-Encoding header with the strict flag true', () {
```
SandPod, line 8: Suggestion/Discussion: When strict mode is active we probably only need to valide how the behavior changes. So if we already validate the details around parsing correct headers in a non-strict mode, then we can only validate when we should reject the header in the strict case?

packages/relic/test/headers/new/special/accept_encoding_header_test.dart
```diff
@@ -0,0 +1,331 @@
...
+        );
+        expect(
+          headers.acceptEncoding?.encodings.map((e) => e.quality).toList(),
+          equals([1.0, 0.5, 0.8]),
+        );
+      },
+    );
+
+    test(
+      'when an invalid Accept-Encoding header is passed then it should throw TestInvalidHeaderException',
```
SandPod, line 64: Suggestion: Describe how it is invalid in the description.

packages/relic/test/headers/new/special/accept_encoding_header_test.dart
```diff
@@ -0,0 +1,331 @@
...
+        );
+        expect(
+          headers.acceptEncoding?.encodings.map((e) => e.quality).toList(),
+          equals([1.0, 0.5, 0.8]),
+        );
+      },
+    );
+
+    test(
+      'when an invalid Accept-Encoding header is passed then it should return null',
```
SandPod, line 226: Suggestion: assert the value is available on the other place for non-parsable headers.

packages/relic/lib/src/headers/special/accept_encoding_header.dart
klkucaj: like this:  `EncodingQuality(this.encoding, [double? quality]) : quality = quality ?? 1.0;` ?

packages/relic/lib/src/headers/special/accept_encoding_header.dart
klkucaj: something like this:

  ```
String toHeaderString() => quality == 1.0 ? encoding : '$encoding;q=$quality';

  @override
  String toString() => 'EncodingQuality(encoding: $encoding, quality: $quality)';
```

File: packages/relic/lib/src/headers/special/accept_encoding_header.dart (hunk 23 above)
Comment (klkucaj) [reply packages/relic/lib/src/headers/special/accept_encoding_header.dart:38]: something like this:

```
 /// Converts the [AcceptEncodingHeader] instance into a string representation suitable for HTTP headers.
  String toHeaderString() => encodings.map((e) => e.toHeaderString()).join(', ');

  @override
  String toString() => 'AcceptEncodingHeader(encodings: $encodings)';
```

Comment (SandPod) [reply packages/relic/lib/src/headers/special/accept_encoding_header.dart:38]: Yeah, that makes sense!

File: packages/relic/lib/src/headers/special/accept_encoding_header.dart (hunk 24 above)
Comment (SandPod) [reply packages/relic/lib/src/headers/special/accept_encoding_header.dart:50]: Yes, that makes sense! 👍

File: packages/relic/lib/src/headers/special/accept_encoding_header.dart (hunk 25 above)
Comment (SandPod) [reply packages/relic/lib/src/headers/special/accept_encoding_header.dart:54]: Yes, that sounds good! 👍

File: packages/relic/test/headers/new/special/accept_encoding_header_test.dart [hunk 31]
```diff
@@ -0,0 +1,262 @@
...
+import 'package:test/test.dart';
+import 'package:relic/src/headers/headers.dart';
+import 'package:relic/src/relic_server.dart';
+import '../../headers_test_utils.dart';
+
+void main() {
+  group('Given an Accept-Encoding header with the strict flag true', () {
+    late RelicServer server;
+
+    setUp(() async {
```
SandPod, line 11: Suggestion: Switch places for this test compared to the "detailed" tests, since it assumes that the detailed tests have already passed.

packages/relic/test/headers/new/special/accept_encoding_header_test.dart
```diff
@@ -0,0 +1,262 @@
...
+      try {
+        server = await RelicServer.bind(InternetAddress.loopbackIPv6, 0);
+      } on SocketException catch (_) {
+        server = await RelicServer.bind(InternetAddress.loopbackIPv4, 0);
+      }
+    });
+
+    tearDown(() => server.close());
+
+    test(
```
SandPod, line 21: Question: Is it enough with a single test here that reads `when invalid Accept-Encoding header is passed...`? Since we already validate what "invalid" means in the other tests.

The bad quality test should covered in the other tests right?

packages/relic/test/headers/new/special/accept_encoding_header_test.dart
```diff
@@ -0,0 +1,262 @@
...
+          );
+
+          expect(
+            headers.acceptEncoding?.encodings.map((e) => e.encoding).toList(),
+            equals(['gzip']),
+          );
+          expect(
+            headers.acceptEncoding?.encodings.map((e) => e.quality).toList(),
+            equals([1.0]),
+          );
```
SandPod, line 100: Suggestion: Break this out into a separate test explaining that the default value is assigned if no quality is present.

packages/relic/test/headers/new/special/accept_encoding_header_test.dart
```diff
@@ -0,0 +1,262 @@
...
```
SandPod, line 133: Suggestion: Remove this part from the test.

This just validates that the default value is set. We can do that in a single place and skip repeating it.

packages/relic/test/headers/new/special/accept_encoding_header_test.dart
```diff
@@ -0,0 +1,262 @@
+import 'dart:io';
+import 'package:test/test.dart';
+import 'package:relic/src/headers/headers.dart';
+import 'package:relic/src/relic_server.dart';
+import '../../headers_test_utils.dart';
+
+void main() {
```
SandPod, line 7: Question: Do we give a reason in the reject message why the parsing failed? If so, maybe "strict" should be the one we test all header variations in. Then, we just validate that "null" is set if we fail.

That way, we can validate that the validation fails for the expected reason. For example, an invalid quality value.

SandPod (commented):
Better! Some minor suggestions to limit the overlap a bit and ensure we fail on the right thing.

packages/relic/test/headers/special/accept_encoding_header_test.dart
```diff
@@ -0,0 +1,280 @@
+import 'dart:io';
+import 'package:test/test.dart';
+import 'package:relic/src/headers/headers.dart';
+import 'package:relic/src/relic_server.dart';
+import '../headers_test_utils.dart';
+
+void main() {
+  // Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding
```
klkucaj, line 8: @SandPod is the reference here enough ?

SandPod: No, I think I was thinking more about references for the specific tests. So for example when you had an empty `Accept-Encoding` header. Where is the source for that decision.

packages/relic/test/headers/special/accept_encoding_header_test.dart
```diff
@@ -0,0 +1,280 @@
...
+        server = await RelicServer.bind(InternetAddress.loopbackIPv6, 0);
+      } on SocketException catch (_) {
+        server = await RelicServer.bind(InternetAddress.loopbackIPv4, 0);
+      }
+    });
+
+    tearDown(() => server.close());
+
+    test(
+      'when an empty Accept-Encoding header is passed then it should throw TestInvalidHeaderException',
```
SandPod, line 23: Discussion: What we really want to test is that the server rejects the message with a specific error message. Would it be better to expect a 400 with the expected message instead of this exception?

SandPod: Suggestion: We include the message we are expecting in the description.

packages/relic/test/headers/special/accept_encoding_header_test.dart
```diff
@@ -0,0 +1,280 @@
...
+          expect(
+            headers.acceptEncoding?.encodings.map((e) => e.encoding).toList(),
+            equals(['gzip', 'deflate', 'br']),
+          );
+        },
+      );
+
+      test(
+        'Accept-Encoding headers with quality values are passed then they should parse the qualities correctly',
+        () async {
```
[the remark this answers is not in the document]
SandPod: I think this makes sense to keep :)

packages/relic/lib/src/headers/extension/http_headers_extension.dart
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  /// Parses and returns multiple header values associated with [key] as a `List<String>`.
+  ///
+  /// - For headers that can be split, splits the values by commas and trims any whitespace.
+  /// - Returns `null` if no values are found.
+  /// - Throws an [InvalidHeaderValueException] if `strict` is `true` and the header contains only empty values.
+  ///
+  /// [strict] determines whether an exception is thrown for invalid or unexpected header values.
+  List<String>? parseMultipleValue(
```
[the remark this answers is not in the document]
klkucaj: Now we have new code and implementation, so this makes sense to leave it like this.

packages/relic/lib/src/headers/extension/http_headers_extension.dart
```diff
@@ -0,0 +1,172 @@
...
+extension HttpHeadersExtension on HttpHeaders {
+  /// Returns the first header value associated with [key] as a `String`.
+  ///
+  /// - If a single value is found, it is returned.
+  /// - If multiple values are present, the first value is returned.
+  /// - Returns `null` if no value is found.
+  /// - Throws an [InvalidHeaderValueException] if `strict` is `true` and the header has an empty or unexpected value.
+  ///
+  /// [strict] determines whether an exception is thrown for invalid or unexpected header values.
+  String? parseSingleValue(
```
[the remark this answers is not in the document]
klkucaj: same for this

packages/relic/lib/src/body/body.dart
```diff
@@ -0,0 +1,120 @@
...
+  /// The contents of the message body.
+  ///
+  /// This will be `null` after [read] is called.
+  Stream<Uint8List>? _stream;
+
+  /// The length of the stream returned by [read], or `null` if that can't be
+  /// determined efficiently.
+  final int? contentLength;
+
+  final BodyType? contentType;
```
SandPod, line 24: Suggestion/Discussion: This is missing a description of a public field.

There are also no descriptions for the factory methods.

As a general linter rule we require descriptions for all public methods and fields, one should probably be added for all of these as well.

packages/relic/lib/src/body/types/mime_type.dart
```diff
@@ -0,0 +1,61 @@
...
+  static const pdf = MimeType('application', 'pdf');
+
+  static const rtf = MimeType('application', 'rtf');
+
+  final String primaryType;
+  final String subType;
+
+  const MimeType(this.primaryType, this.subType);
+
+  factory MimeType.parse(String type) {
```
SandPod, line 31: Suggestion: This will be hard to use without inspecting the contents of the method. Please add a description to these methods.

packages/relic/lib/src/body/types/mime_type.dart
```diff
@@ -0,0 +1,61 @@
...
+    return MimeType(
+      type.primaryType,
+      type.subType,
+    );
+  }
+
+  static MimeType? tryParse(String? value) {
+    if (value == null) return null;
+    return MimeType.parse(value);
+  }
```
SandPod, line 57: Chore: This method will throw if the format is incorrect.

That is not correct behavior for a `tryParse` method. 

1. Is it actually required or is it enought for us to supply a `parse` method.
2. Please ensure that any `tryParse` method behaves correctly and never throws.

packages/relic/lib/src/extension/base64_extension.dart
```diff
@@ -0,0 +1,24 @@
+import 'dart:convert';
+
+/// Helper method to check if a string is valid Base64.
+///
+/// This method is used internally to validate that the Base64-encoded credentials
+/// are correctly formatted before attempting to decode them.
+extension Base64Extension on String {
+  bool get isValidBase64 {
+    final base64Pattern = RegExp(r'^[A-Za-z0-9+/]+={0,2}$');
```
SandPod, line 9: Question/Discussion: This pattern definition is different from the one used to identify base64 in auth. Which one is correct? And can the implementation be shared? `packages/serverpod_serialization/lib/src/auth_encodingd.dart`

packages/relic/lib/src/middleware/logger.dart
```diff
@@ -0,0 +1,98 @@
...
+  if (isError) {
+    print('[ERROR] $msg');
+  } else {
+    print(msg);
+  }
+}
+
+void main() {
+  print("POST".padRight(7));
+}
```
SandPod, line 98: Chore: Remove this main method.

packages/relic/lib/src/middleware/logger.dart
```diff
@@ -0,0 +1,98 @@
...
+        .terse;
+  }
+
+  var msg = '$requestTime\t$elapsedTime\t${method.value}\t${requestedUri.path}'
+      '${_formatQuery(requestedUri.query)}\n$error';
+
+  return '$msg\n$chain';
+}
+
+void _defaultLogger(String msg, bool isError) {
```
SandPod, line 88: Suggestion/Discussion: It is strange to me that this is defined in a different place compared to `error_logger.dart`.

Should they be defined in the same place?

packages/relic/lib/src/message/message.dart
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  /// Returns the encoding specified in the Content-Type header, or null if not specified.
+  Encoding? get encoding {
+    if (body.contentType == null) return null;
+    return body.contentType!.encoding;
+  }
+
+  /// Reads the body as a stream of bytes. Can only be called once.
+  Stream<List<int>> read() => body.read();
```
SandPod, line 36: Suggestion: Shouldn't this be a stream of `Uint8List`?

Please double check all places where we return a list of int so that they shouldn't be a `Uint8List`.

packages/relic/lib/src/message/message.dart
```diff
@@ -0,0 +1,53 @@
...
+
+  /// Reads the body as a string, decoding it using the specified or detected encoding.
```
SandPod, line 38: Suggestion: Add to the description that `utf8` is used if no encoding is provided or detected.

SandPod (commented):
Some initial comments.

packages/relic/lib/src/extension/base64_extension.dart
klkucaj: i removed this, we dont actually need it, because this `utf8.decode(base64Decode(this))` is gonna throw an exception if the encoding is invalid, so we dont actually need it to check first.

util/run_tests_unit
```diff
@@ -14,6 +14,7 @@ declare -a projectPaths=(
     "packages/serverpod_client"
     "packages/serverpod_shared" 
     "packages/serverpod_serialization"
+    "packages/relic"
```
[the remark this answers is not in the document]
klkucaj: Since this package it is not gonna be user for now, i am skipping this.

packages/relic/lib/src/headers/headers.dart
```diff
@@ -0,0 +1,1688 @@
...
+    required this.failedHeadersToParse,
+  }) : custom = custom ?? CustomHeaders.empty();
+
+  factory Headers.fromHttpRequest(
+    io.HttpRequest request, {
+    bool strict = false,
+    required String? xPoweredBy,
+  }) {
+    Map<String, List<String>> failedHeadersToParse = {};
+    var dartIOHeaders = request.getHeadersParser(
```
SandPod, line 389: Suggestion: Remove extension and create the HeaderParser directly.

This is only called from this single place. No need to add an extention to request to support this.

packages/relic/lib/src/headers/parser/headers_parser.dart
```diff
@@ -0,0 +1,276 @@
...
+      _throwException(e, key: key);
+    }
+  }
+
+  /// Parses a URI from the header value associated with [key], or returns
+  /// `null` if no valid URI is found.
+  ///
+  /// - Throws an [InvalidHeaderException] if `strict` is `true` and the header
+  /// `key` is present but contains an invalid URI.
+  Uri? parseUri(String key) {
```
SandPod, line 112: Question(NIT - maybe we should not do anything): These methods don't seem to fit on this abstraction level. There isn't really a difference between this and `CacheControlHeade``parse` except that one is a dart supplied type? 

On this class method we should maybe only supply the `parseSingleValue` and `parseMultipleValue` methods. And then pass these type parsers in as values to the `parseMultipleValue` or `parseSingleValue` methods? (They can still be shared between different headers)

Now we have bunbled pulling one or more values our of the header and parsing them.

packages/relic/lib/src/headers/custom/custom_headers.dart
```diff
@@ -0,0 +1,113 @@
...
+    return CustomHeaders(updatedHeaders);
+  }
+}
+
+/// Converts a list of [MapEntry] to a case-insensitive map.
+CaseInsensitiveMap<List<String>> _toCaseInsensitiveMap(
+  Iterable<MapEntry<String, List<String>>> entries,
+) {
+  final validEntries = entries
+      .where((entry) => entry.value.isNotEmpty)
```
SandPod, line 96: Question: Why do we need to filter out the empty values?

packages/relic/lib/src/headers/headers.dart
```diff
@@ -0,0 +1,1688 @@
...
+      age: dartIOHeaders.parseInt(
+        _ageHeader,
+        allowNegative: false,
+      ),
+      allow: dartIOHeaders.parseMultipleValue(
+        _allowHeader,
+        onParse: (values) {
+          return values
+              .splitTrimAndFilterUnique(emptyCheck: false)
+              .map(Method.parse)
```
SandPod, line 451: Suggestion: Follow the general pattern and avoid implementation in this declaration.

packages/relic/lib/src/headers/headers.dart
```diff
@@ -0,0 +1,1688 @@
...
+        onParse: IfNoneMatchHeader.parse,
+      ),
+      ifRange: dartIOHeaders.parseSingleValue(
+        _ifRangeHeader,
+        onParse: IfRangeHeader.parse,
+      ),
+      maxForwards: dartIOHeaders.parseInt(
+        _maxForwardsHeader,
+        allowNegative: false,
+      ),
```
SandPod, line 490: Suggestion: Create a parser called `parsePositiveInt` instead of passing in bools from this place.

packages/relic/lib/src/headers/headers.dart
```diff
@@ -0,0 +1,1688 @@
...
+      // Response Headers
+      location: dartIOHeaders.parseUri(_locationHeader),
+      xPoweredBy: dartIOHeaders.parseString(
+            _xPoweredByHeader,
+          ) ??
+          xPoweredBy,
+      accessControlAllowCredentials: dartIOHeaders.parseBool(
+        _accessControlAllowCredentialsHeader,
+        allowFalse: false,
+      ),
```
SandPod, line 521: Suggestion: Create a specialized method to parse this.

Strange to only allow positive values. Should negative values be discarded rather than throwing?

packages/relic/test/ssl/ssl_certs.dart
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
SandPod, line 1: Question: This is the file that should probably be ignored podspec ignored right?

packages/relic/test/headers/special/accept_encoding_header_test.dart
```diff
@@ -0,0 +1,388 @@
+import 'dart:io';
```
SandPod, line 1: Question: Why are the headers folder named "special"?  Is that a common term? These headers seems to be more of the "standard" heeaders that are accepted by browsers.

What is the difference bewteen these and "basic" headers? Should they all be in a "standard" header folder alongside the "custom" folder?

packages/relic/test/headers/docs/strict_validation_docs.dart
```diff
@@ -0,0 +1,9 @@
+/// ## Strict Header Validation
+///
+/// This code will throw an exception when the header is empty and the strict flag is set to true.
+/// In many servers, an empty header is considered as null, which can lead to unexpected behavior.
+/// By enforcing strict validation, we ensure that any empty headers are caught and handled appropriately,
+/// preventing potential issues in the server's request handling process.
+class StrictValidationDocs {
```
SandPod, line 7: Question: What is this?

SandPod (commented):
Couple of comments. Let's have Viktor take a look at this as well.

If there are major changes we could also do these in separate PRs. But Then I would like to have one or more issues that track these changes.

templates/pubspecs/packages/relic/pubspec.yaml
```diff
@@ -0,0 +1,25 @@
+name: relic
+description: A lightweight web server inspired by Shelf.
+version: SERVERPOD_VERSION
```
SandPod, line 3: Suggestion: Remove the serverpod version from the relic package. This should be managed separately.

packages/serverpod/pubspec.yaml
```diff
@@ -34,6 +34,7 @@ dependencies:
   system_resources: ^1.6.0
   vm_service: ">=13.0.0 <15.0.0"
   yaml: ^3.1.1
+  relic: 2.2.0
```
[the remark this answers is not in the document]
SandPod: Suggestion: Use version `^0.1.0` for the relic package import.

This will be version controlled separately from Serverpod.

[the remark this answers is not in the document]
SandPod: (remember to update the template)

templates/pubspecs/packages/relic/pubspec.yaml
SandPod: Or even better, add a RELIC_VERSION variable here that is substituted from a separate file in the root.

SandPod (commented):
Last comments around the version.

packages/relic/lib/src/headers/special/cookie_header.dart
```diff
@@ -0,0 +1,68 @@
...
+/// A class representing a single cookie.
+class Cookie {
+  /// The name of the cookie.
+  final String name;
+
+  /// The value of the cookie.
+  final String value;
+
+  /// Constructs a [Cookie] instance with the specified name and value.
+  Cookie({required this.name, required this.value});
```
vlidholt, line 59: Should we validate that the Strings are valid?

Cookie Name:

Must not contain:
Any whitespace characters (space, tabs, etc.).
Any of the following characters: ()<>@,;:\"/[]?={} \t.
Generally, the name should only contain:
Alphanumeric characters (A-Z, a-z, 0-9).
Special characters like _, -, or ..
Cookie Value:

Can contain:
Alphanumeric characters (A-Z, a-z, 0-9).
Special characters like !#$%&'*+-.^_|~\`.
Encoded characters like %20 (for spaces).
Must not contain:
Control characters (ASCII 0–31 or 127).
Reserved characters without being encoded (e.g., \`;

vlidholt: This may apply to other headers, too.

packages/relic/lib/src/headers/special/cross_origin_resource_policy_header.dart
```diff
@@ -0,0 +1,48 @@
...
+
+/// A class representing the HTTP Cross-Origin-Resource-Policy header.
+///
+/// This header specifies the policy for sharing resources across origins.
+class CrossOriginResourcePolicyHeader {
+  /// The policy value of the header.
+  final String policy;
+
+  /// Constructs a [CrossOriginResourcePolicyHeader] instance with the specified value.
+  const CrossOriginResourcePolicyHeader(this.policy);
```
vlidholt, line 11: We should validate that the values are correct here. Only the defined values seem to be valid. Or, we can make the constructor private?

packages/relic/lib/src/message/message.dart
```diff
@@ -0,0 +1,55 @@
...
+  Message({
+    required this.body,
+    required this.headers,
+    this.context = const {},
+  });
+
+  /// Returns the MIME type from the Content-Type header, if available.
+  String? get mimeType {
+    if (body.contentType == null) return null;
+    return body.contentType!.mimeType.toHeaderValue();
```
vlidholt, line 27: This probably shouldn't return a `String`. Instead:

```dart
MimeType? get mimeType => body.contentType?.mimeType;
```

packages/relic/lib/src/message/message.dart
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /// Returns the encoding specified in the Content-Type header, or null if not specified.
+  Encoding? get encoding {
+    if (body.contentType == null) return null;
+    return body.contentType!.encoding;
+  }
```
vlidholt, line 34:
```dart
Encoding? get encoding => body.contentType?.encoding;
```

packages/relic/lib/src/relic_server.dart
```diff
@@ -0,0 +1,217 @@
...
+import 'package:relic/src/message/request.dart';
+import 'package:relic/src/message/response.dart';
+import 'package:relic/src/util/util.dart';
+
+import 'handler/handler.dart';
+
+/// A [Server] backed by a `dart:io` [HttpServer].
+class RelicServer {
+  /// The default powered by header to use for responses.
+  static const String defaultPoweredByHeader = 'Relic';
```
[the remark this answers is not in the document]
vlidholt: I think just `Relic` is more in line with most response headers. (The shelf variant is kind of odd.)

packages/relic/pubspec.yaml
```diff
@@ -0,0 +1,25 @@
+name: relic
+description: A lightweight web server inspired by Shelf.
+version: 2.2.0
```
vlidholt, line 3: `version: 0.2.0`.

util/update_pubspecs
vlidholt, line 1: Let's revert the changes in this one (as we're not going to use the templates).

templates/pubspecs/packages/relic/pubspec.yaml
vlidholt: Ah, a didn't see this when doing the review. A `RELIC_VERSION` would be 👌. Perhaps this is easier left to a follow up PR though?

SandPod: Yes, maybe just remove all templating for now.

packages/relic/lib/src/headers/special/cookie_header.dart
SandPod: The rule for these are here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

packages/serverpod/test/relic/templates_test.dart
```diff
@@ -8,6 +8,7 @@ void main() {
   test(
       'Given missing templates folder when loading templates then no templates are loaded',
       () async {
+    Templates templates = Templates();
```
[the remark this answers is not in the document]
klkucaj: Nope!

packages/relic/lib/src/body/body.dart
klkucaj: I tried to add comments on all public classes/methods/fields, but this will also be done on a separate pr when we add `serverpod_lints` package

packages/relic/test/headers/docs/strict_validation_docs.dart
klkucaj: These are just docs for the tests

packages/relic/lib/src/headers/headers.dart
klkucaj: Yes only positive, but you are right, we can set it to null instead of throwing 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials#directives

packages/relic/lib/src/headers/typed/headers/accept_language_header.dart
```diff
@@ -0,0 +1,86 @@
+import 'package:relic/src/headers/extension/string_list_extensions.dart';
+import 'package:relic/src/headers/typed/base/typed_header.dart';
+
+/// A class representing the HTTP Accept-Language header.
+///
+/// This header specifies the natural languages that are preferred in the response.
+class AcceptLanguageHeader extends TypedHeader {
```
jonasfj, line 7: Just curious:
 * Have you considered [extension types](https://dart.dev/language/extension-types) instead of creating an object for each header?
   * I get the impression that you'll be parsing headers that aren't even read.
   * It might be cheaper, then again might be slightly more dynamic.

Nit: Have you considered using [class modifiers](https://dart.dev/language/class-modifiers) to prevent users from implementing these class, thus, ensuring that you can extend the interface in the future, should you ever want to.

packages/relic/lib/src/headers/headers.dart
```diff
@@ -0,0 +1,1482 @@
...
+import 'package:relic/src/headers/extension/string_list_extensions.dart';
+import 'package:relic/src/headers/parser/headers_parser.dart';
+import 'package:relic/src/headers/parser/shared_parser.dart';
+import 'package:relic/src/method/method.dart';
+
+import 'typed/typed_headers.dart';
+
+import '../body/body.dart';
+
+abstract class Headers {
```
jonasfj, line 14: If you don't make this a `final class` then people outside this package can implement / extend it, and you'll have problems if you add new headers to it in the future.

It could potentially be a breaking change to add a new field to a class that isn't `final` or `sealed`.

packages/relic/lib/src/headers/typed/headers/accept_language_header.dart
```diff
@@ -0,0 +1,86 @@
...
+    if (splitValues.isEmpty) {
+      throw FormatException('Value cannot be empty');
+    }
+
+    if (splitValues.length == 1 && splitValues.first == '*') {
+      return AcceptLanguageHeader.wildcard();
+    }
+
+    if (splitValues.length > 1 && splitValues.contains('*')) {
+      throw FormatException('Wildcard (*) cannot be used with other values');
```
jonasfj, line 36: If my request handler never readers the `Accept-Language` header, a server using this framework will still reject requests if they carry an invalid `Accept-Language` header.

Is that wise? (I'm actually not sure, hence, why I'm asking :rofl: )

What do other server frameworks do here?

packages/relic/lib/src/headers/typed/headers/accept_language_header.dart
klkucaj: Hi @jonasfj,  
This update aims to eliminate dynamic typing entirely, which is the primary reason for this PR.

packages/relic/lib/src/headers/typed/headers/accept_language_header.dart
klkucaj: Hi @jonasfj,  
We have a `strict` flag for headers. When true, any invalid header rejects the request, ensuring valid headers. If false, invalid headers are added to the `failedHeadersToParse` map in `headers.dart`.

packages/relic/lib/src/headers/typed/headers/authorization_header.dart
```diff
@@ -0,0 +1,315 @@
...
+      throw FormatException('Value cannot be empty');
+    }
+
+    if (value.startsWith(BearerAuthorizationHeader.prefix.trim())) {
+      return BearerAuthorizationHeader.parse(value);
+    } else if (value.startsWith(BasicAuthorizationHeader.prefix.trim())) {
+      return BasicAuthorizationHeader.parse(value);
+    } else if (value.startsWith(DigestAuthorizationHeader.prefix.trim())) {
+      return DigestAuthorizationHeader.parse(value);
+    }
```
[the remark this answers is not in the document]
klkucaj: @jonasfj,  
This is just the beginning. The goal is to make all typed headers abstract later, allowing 'Relic' to provide default implementations while enabling devs to create custom classes.  

For now, this header supports `Bearer`, `Basic`, and `Digest`. The PR is still a work in progress with much left to do.

packages/relic/lib/src/headers/headers.dart
klkucaj: @jonasfj,  
I agree with you, but if someone implements this class, they must override and implement all abstract methods. If those methods are implemented incorrectly, it's beyond our responsibility.  

That said, we should review and decide if making it a `final` class is better to prevent unintended implementations.

packages/relic/lib/src/headers/typed/headers/accept_language_header.dart
SandPod: `strict` mode is our own contraption and can be configured per route in the server. It allows a developer always to reject any request that has invalid headers. The downside is that the header parser will need to parse all headers received when the request reaches the server.

Most other server frameworks do not supply such a feature but require this to be implemented as a middleware if wanted. However, since we provide types for all headers, it was pretty simple to supply the feature with the framework. But with that said, this is an optional feature that the developer can enable at will.

packages/relic/lib/src/headers/special/cookie_header.dart
klkucaj: @vlidholt fixed here: https://github.com/serverpod/serverpod/pull/2744/commits/cc673d4935d8cf1f34bd53317c16c607ab86cd7e

packages/relic/lib/src/headers/typed/headers/cookie_header.dart
```diff
@@ -0,0 +1,120 @@
...
+  String toHeaderString() => '$name=$value';
+
+  @override
+  String toString() {
+    return 'Cookie(name: $name, value: $value)';
+  }
+
+  /// Validates the cookie name according to the rules.
+  static bool isNameValid(String name) {
+    if (name.isEmpty) return false;
```
SandPod, line 94: Question: Is an empty name not allowed? Do you have a source on this?

packages/relic/lib/src/headers/typed/headers/cookie_header.dart
```diff
@@ -0,0 +1,120 @@
...
+
+    return true;
+  }
+
+  /// Validates the cookie value according to the rules.
+  static bool isValueValid(String value) {
+    if (value.isEmpty) return false;
+
+    // Check for control characters and whitespace
+    final invalidCharacters = RegExp(r'[\x00-\x1F\x7F]');
```
SandPod, line 112: Question: Aren't the control characters also forbidden for the cookie name?

packages/relic/pubspec.yaml
```diff
@@ -0,0 +1,25 @@
+name: relic
+description: A lightweight web server inspired by Shelf.
+version: 0.1.0
```
SandPod, line 3: chore: Update this to 0.2.0 as suggested in the review.

SandPod (commented):
Still a couple of things to fix. There where a couple of comments from the review that where resolved that where not fixed in the PR.

Please read through the comments carefully to ensure that the feedback corresponds to the fix.

packages/relic/lib/src/headers/typed/headers/authorization_header.dart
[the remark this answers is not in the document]
jonasfj: Interesting, it seems a bit hard to allow custom classes, if you're parsing them like this.

packages/relic/lib/src/headers/typed/headers/accept_language_header.dart
jonasfj: disclaimer: I don't have any answers here, so feel free to ignore my drive by comments :rofl: 

I can see it being attractive to have certain headers be `strict`.
Examples:
 * Ignoring a malformed `Content-Type`, `Content-Length`, `Host` could be bad.
 * Ignoring a malformed `Referer`, `Accept-Language` or `Accept` header, might be more reasonable. Indeed rejecting a request over malformed `Accept-Language` header might defeat the point of content-negotiation (especially, if the request handler never reads the header).

packages/relic/lib/src/headers/typed/headers/accept_language_header.dart
jonasfj: > This is something we are aiming to fix in upcoming changes; if strict mode is disabled, we should only parse headers that are actually accessed.

I suspect that if you do this, you might need to change the contract around what a `Handler` is.

I'm guessing that like in `shelf` a `Handler` is a function that takes a `Request` and returns a `FutureOr<Response>`.

But if you're parsing headers on-the-fly, you'll have a harder time communicating a 400 response from inside the request handler.

When working on `pub.dev` we found that having a `ResponseException` was pretty handy.

Basically, we made some middelware that would catch `ResponseException` and return `ResponseException.response`.

Then whenever we try to parse a header on-the-fly and it fails, we just throw a `ResponseException(400)` or if authorization failed throw `ResponseException(401)`.

This contract isn't part of `shelf`, and maybe it shouldn't be, but if you want to do on-the-fly parsing of headers and reject requests with invalid headers, you might need to make such a concept part of the `Handler` specification in relic.

(just some food for thought).

packages/relic/lib/src/headers/typed/base/typed_header.dart
```diff
@@ -0,0 +1,7 @@
+/// A typed header that can be converted to a header string.
```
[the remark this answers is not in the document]
klkucaj: @SandPod `typed_headers.dart` contains only the import of the all the typed headers, do u think is a good idea to put this there ?

packages/relic/lib/src/headers/typed/headers/cookie_header.dart
klkucaj: @SandPod here https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes

packages/relic/lib/src/headers/typed/headers/cookie_header.dart
klkucaj: yes you are right, thank you!

packages/relic/lib/src/headers/typed/headers/cookie_header.dart
```diff
@@ -0,0 +1,120 @@
...
+
+  /// The value of the cookie.
+  final String value;
+
+  /// Constructs a [Cookie] instance with the specified name and value.
+  Cookie({
+    required this.name,
+    required this.value,
+  })  : assert(isNameValid(name), 'Invalid cookie name'),
+        assert(isValueValid(value), 'Invalid cookie value');
```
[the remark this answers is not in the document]
klkucaj: @SandPod Good point, and yes, this was intentional. The goal is to ensure that invalid headers are caught as early as possible, regardless of how the cookies are constructed. Since someone could use the constructor directly to build cookies, it’s critical that the class validates inputs and throws exceptions when invalid headers are provided in any scenario.
So what do you think ?

packages/relic/lib/src/headers/typed/headers/accept_language_header.dart
klkucaj: @jonasfj a new issue is created for this https://github.com/serverpod/serverpod/issues/3034
Thank you for your feedback!

packages/relic/lib/src/headers/typed/headers/authorization_header.dart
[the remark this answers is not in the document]
klkucaj: @jonasfj you are right, and i just created an issue for this: https://github.com/serverpod/serverpod/issues/3035
thank you!

packages/relic/lib/src/headers/headers.dart
```diff
@@ -0,0 +1,1482 @@
...
+import 'package:relic/src/headers/extension/string_list_extensions.dart';
+import 'package:relic/src/headers/parser/headers_parser.dart';
+import 'package:relic/src/headers/parser/common_types_parser.dart';
+import 'package:relic/src/method/method.dart';
+
+import 'typed/typed_headers.dart';
+
+import '../body/body.dart';
+
+sealed class Headers {
```
SandPod, line 14: Suggestion: Use `base` class.

We don't want exhaustive search for the classes defined from this, we simply want to prohibit implmenetations of this class outside of the library. I think `base` is the right choice in that case (since we need to inherit from this class for the copyWith method)

SandPod: You can read more here: https://dart.dev/language/class-modifiers-for-apis#the-base-modifier

packages/relic/lib/src/method/method.dart
```diff
@@ -0,0 +1,57 @@
+import 'package:relic/src/headers/typed/typed_headers.dart';
+
+/// Represents the HTTP methods used in requests as constants.
+class Method extends TypedHeader {
```
SandPod, line 4: Suggestion: This isn't a header.

It is simply a value that is used in header.
It shouldn't extend the TypedHeader interface right?

SandPod (commented):
Still a couple of comments and suggestions.

packages/relic/lib/src/method/method.dart
klkucaj: Yes it is, the `Request Method` is a header.

packages/relic/lib/src/headers/typed/headers/cookie_header.dart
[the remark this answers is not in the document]
klkucaj: @SandPod you are right, this will be removed by the compiler in `release mode`, and we need to throw in `release mode` also. Thank you!

packages/relic/lib/src/headers/typed/headers/cookie_header.dart
klkucaj: Fixed!

packages/relic/lib/src/headers/typed/typed_headers.dart
```diff
@@ -38,3 +37,11 @@ export 'package:relic/src/headers/typed/headers/access_control_allow_origin_head
 export 'package:relic/src/headers/typed/headers/access_control_allow_headers_header.dart';
 export 'package:relic/src/headers/typed/headers/access_control_expose_headers_header.dart';
 export 'package:relic/src/headers/typed/headers/accept_header.dart';
+
+/// A typed header that can be converted to a header string.
+abstract class TypedHeader {
```
SandPod, line 42: Suggestion: Mark this as an `abstract interface class` and implement it instead.

Since inheritance is not really needed for the usecase, it should be enough to guarantee that a method is implemented right?

packages/relic/lib/src/headers/typed/headers/cookie_header.dart
```diff
@@ -52,71 +60,348 @@ class Cookie {
...
+
+  /// Whether the cookie is available from other sites.
+  ///
+  /// This value is `null` if the SameSite attribute is not present.
+  ///
+  /// See [SameSite] for more information.
+  SameSite? sameSite;
+
+  /// Constructs a [Cookie] instance with the specified name and value.
+  SetCookieHeader({
```
SandPod, line 137: Suggestion: Break this out into a different file.

To make the project easy to navigate we should only have a single header in each file.

Shared classes like `Cookie` can then be placed in some shared place.

packages/relic/lib/src/headers/typed/headers/content_encoding_header.dart
```diff
@@ -79,7 +79,7 @@ class ContentEncoding {
     if (trimmed.isEmpty) {
       throw FormatException('Name cannot be empty');
     }
-    switch (trimmed.toLowerCase()) {
+    switch (trimmed) {
```
SandPod, line 82: Question: Is the removal of `toLowerCase` intentional?

packages/relic/lib/src/headers/typed/headers/content_encoding_header.dart
```diff
@@ -93,7 +93,7 @@ class ContentEncoding {
       case _zstd:
         return zstd;
       default:
-        return ContentEncoding(trimmed);
+        throw FormatException('Invalid value');
```
SandPod, line 96: Question: Was the default case changed to throw intentional?

klkucaj: We want only to accept headers and values that are known, not custom values.

packages/relic/lib/src/headers/typed/headers/content_encoding_header.dart
klkucaj: There are constants like `includeSubDomains` that are case-sensitive. If we lowercase the input string, we would also need to lowercase the constants for the `switch` statement to work properly.

For example:

```dart
static const _includeSubDomains = 'includeSubDomains';
// code...
switch (trimmed.toLowerCase()) {
  case _includeSubDomains:
    return includeSubDomains;
}
// code...
```

The above won't work because `_includeSubDomains` is not lowercased, so it won't match the lowercased `trimmed`.

The only way to handle this would be to explicitly lowercase the constant in the `case`, like so:

```dart
// code...
switch (trimmed.toLowerCase()) {
  case _includeSubDomains.toLowerCase():
    return includeSubDomains;
}
// code...
```

However, this is not possible because `switch` statements require constants, and method calls like `.toLowerCase()` are not allowed.

packages/relic/lib/src/headers/typed/typed_headers.dart
klkucaj: Yes.

SandPod:
This PR was merged as an initial commit to a standalone git repository containing only the Relic web server.
This is the PR that copies over the changes in this branch: https://github.com/serverpod/relic/pull/3

PR #2745, opened 2024-09-12
Refactor GoogleCloudStorage to use gcloud

### LINKED ISSUE #2517: `verifyDirectFileUpload` always returns `false` even when upload succeeds

olivoil:
This PR refactors the implementation of the GoogleCloudStorage integration to use gcloud.

Authentication to Cloud Storage switches was previously done using HMAC keys, this PR uses service account credentials. Users are expected to provide the path to the credentials file  when creating a GoogleCloudStorage instance.

Fixes #2517.

This implementation is based on a gist I shared in https://github.com/serverpod/serverpod/issues/2517#issuecomment-2285029195, which I have been using on an app in development. It's been adapted to the current version of Serverpod.

I may need help to test this, and make sure it's working as expected, but will submit the PR and see what passes first.

## Pre-launch Checklist

## Breaking changes

The way to instantiate a GoogleCloudStorage instance is slightly different, as we need to run an `async` function to read the service credentials file and setup authentication to gcloud.

An example of usage has been added to the GoogleCloudStorage class:

```dart
final publicStorage = await GoogleCloudStorage.init(
  serverpod: pod,
  storageId: 'public',
  public: true,
  bucket: myBucket,
  serviceAccountPath: '/config/google_account_credentials.json',
);

pod.addCloudStorage(publicStorage);
```

Additionally, users are now expected to provide service account credentials instead of using HMAC keys to authenticate with storage. The documentation on the website likely needs to be updated to reflect it. I am assuming the website documentation falls outside of this PR, but happy to contribute to it if needed.


olivoil:

integrations/serverpod_cloud_storage_gcp/lib/src/google_cloud_storage.dart
```diff
@@ -0,0 +1,289 @@
...
+  ///     serviceAccountPath: '/config/google_account_credentials.json',
+  ///   );
+  ///
+  ///   pod.addCloudStorage(publicStorage);
+  static Future<GoogleCloudStorage> init({
+    required Serverpod serverpod,
+    required String storageId,
+    required bool public,
+    required String bucket,
+    required String serviceAccountPath,
```
Isakdl, line 48: Suggestion: 

This comment comes with a few implications. By passing in a path to the service account we are forcing people to use a service account stored on disc to authenticate with google. But there are otherways people may want to auth such as via the metaserver in a hosted environment. Google also has a method for default authentication, e.g. `clientViaApplicationDefaultCredentials` that will use an environment variable with the path to the service account, or the metaserver if running in cloud engine. 

So it would be nice to use this method instead of passing in the path to the service account.

On top of this, it would be great to take an optional http client. If set it would override the call to `clientViaApplicationDefaultCredentials`. Then the user can use any available Google auth client or even a mock client for testing.

There is one implication of this though, and that is that we are currently signing the URL in process because we have direct access to the private key. This is not possible if we have authed via the meta server. 

Luckily Google has an API to sign the URL though, but it requires the service account email. We could in other words take the service account email as input instead of the service account path and call this endpoint.

Like so:

```dart
var _iamCredentials = IAMCredentialsApi(authedClient);

final response = await _iamCredentials.projects.serviceAccounts.signBlob(
  SignBlobRequest(payload: base64.encode(stringToSignList)),
  'projects/-/serviceAccounts/$_serviceAccountEmail',
);
```

Where the the class `IamCredentials` comes from `import 'package:googleapis/iamcredentials/v1.dart';`

With the above change we can remove all service account management inside the code.

integrations/serverpod_cloud_storage_gcp/lib/src/google_cloud_storage.dart
```diff
@@ -0,0 +1,289 @@
+import 'dart:collection';
+import 'dart:convert';
+import 'dart:io';
+import 'dart:typed_data';
+
+import 'package:crypto/crypto.dart';
+import 'package:hex/hex.dart';
+import 'package:gcloud/storage.dart';
```
Isakdl, line 8: Suggestion: Reading up on this dependency, it is tagged as experimental by Google. I'm a little reluctant to add it because of that. It is also possible to implement this same functionality with `googleapis/storage`, it is a bigger deviation from the current solution, but I believe it would be more resilient.

Isakdl (changes requested):
Thank you very much for the PR @olivoil very helpful and good work putting this together 😄!

I have taken a look at the code and made a few suggestions. On top of these comments I would like to put this code into a complete new package rather than overriding the old one as you point out this is a breaking change. But if we create an additional package it is not.

On the tests, yes unfortunately we have 0 tests on this code today. There are two different approaches to test this, either using a mock for the google services, such as this https://github.com/fsouza/fake-gcs-server, or implement a real e2e test against a real bucket. 

It would be good to have coverage here, but I understand that this may be difficult to setup and I can help you with that if you like.

alexdess:
Hello,
I just tested your PR, it works well, thank you.

For the web integration, it’s necessary to configure CORS as follows:
```json
[
  {
    "origin": ["http://localhost:49423", [TO_BE_COMPLETED]],
    "method": ["DELETE", "GET", "HEAD", "POST","PUT"],
    "responseHeader": ["Content-Type", "x-goog-content-length-range"],
    "maxAgeSeconds": 3600
  }
]
```
I created a file named gcloud_cors.json containing this configuration and applied it using the following command:
```
gcloud storage buckets update gs://[YOUR_BUCKET_NAME] --cors-file=gcloud_cors.json
```

[Google doc for CORS infos](https://cloud.google.com/storage/docs/using-cors)

I believe it could be beneficial to include this information in the project's documentation once the PR is merged. I'd be happy to contribute and help with that if needed.

integrations/serverpod_cloud_storage_gcp/lib/src/google_cloud_storage.dart
olivoil: I think using the API to get a signature makes sense, and removes the burden of maintaining the signature logic. I'll update it

integrations/serverpod_cloud_storage_gcp/lib/src/google_cloud_storage.dart
olivoil: It does have a status of experimental, and I don't know what their timeline to GA is, or if it will make it to GA... Will look into `googleapis/storage` to see how it works

olivoil:
> Thank you very much for the PR @olivoil very helpful and good work putting this together 😄!
> 
> I have taken a look at the code and made a few suggestions. On top of these comments I would like to put this code into a complete new package rather than overriding the old one as you point out this is a breaking change. But if we create an additional package it is not.
> 
> On the tests, yes unfortunately we have 0 tests on this code today. There are two different approaches to test this, either using a mock for the google services, such as this https://github.com/fsouza/fake-gcs-server, or implement a real e2e test against a real bucket.
> 
> It would be good to have coverage here, but I understand that this may be difficult to setup and I can help you with that if you like.

Thank you, I'll go through the feedback and post an update

> Thank you very much for the PR @olivoil very helpful and good work putting this together 😄!
> 
> I have taken a look at the code and made a few suggestions. On top of these comments I would like to put this code into a complete new package rather than overriding the old one as you point out this is a breaking change. But if we create an additional package it is not.
> 
> On the tests, yes unfortunately we have 0 tests on this code today. There are two different approaches to test this, either using a mock for the google services, such as this https://github.com/fsouza/fake-gcs-server, or implement a real e2e test against a real bucket.
> 
> It would be good to have coverage here, but I understand that this may be difficult to setup and I can help you with that if you like.

Thank you for the review! I'll post an update addressing the feedback, then we can look at how to test it

Isakdl:
@olivoil good question, but some variant sounds nice. Not 100% sure on the naming but ideally it should look nice and be kind of obvious that this is the one you should use. Some ideas: `_native` `_bucket` `_iam`, but we don't have to set it in stone yet.

iampopal:
In which version of serverpod this will release because we really need to check is file uploaded or not and better control the security of files so each user only see files that belongs to them.

marcelomendoncasoares:
Closing now that the issue is solved with the refactor of the storage packages.

Issue #2709, opened 2024-09-04
Non-persistent data is lost when inserting a row into the database.


vlidholt:
Consider the following model (from the chat module):

```yaml
### A chat message.
class: ChatMessage
table: serverpod_chat_message
fields:
  ### The channel this message was posted to.
  channel: String

  ### The body of the message.
  message: String

  ### The time when this message was posted.
  time: DateTime

  ### The user id of the sender.
  sender: int

  ### Information about the sender.
  senderInfo: module:auth:UserInfoPublic?, !persist

  ### True, if this message has been removed.
  removed: bool

  ### The client message id, used to track if a message has been delivered.
  clientMessageId: int?, !persist

  ### True if the message has been sent.
  sent: bool?, !persist

  ### List of attachments associated with this message.
  attachments: List<ChatMessageAttachment>?
indexes:
  serverpod_chat_message_channel_idx:
    fields: channel
```

We are using the following code to insert a row in the database.

```dart
var chatMessage = ChatMessage(
    channel: channel,
    message: message.message,
    time: DateTime.now(),
    sender: userInfo.id!,
    senderInfo: userInfo,
    removed: false,
    clientMessageId: message.clientMessageId,
    sent: true,
    attachments: message.attachments,
  );

chatMessage = await ChatMessage.db.insertRow(session, chatMessage);
```

The `chatMessage` returned from `insertRow` has the `userInfo` property stripped. Perhaps the expected behaviour is that `!persist` fields should be preserved?

PR #2751, opened 2024-09-16
fix: Preserve non-persisted fields during database insert and update operations

klkucaj:
Closes: #2709

## Pre-launch Checklist


tests/serverpod_test_server/test_integration/database_operations/persist/non_persist_field_test.dart
```diff
@@ -0,0 +1,145 @@
...
+  var session = await IntegrationTestServer().session();
+
+  group('Given a class with fields that include non-persisted fields,', () {
+    tearDownAll(() async => ObjectFieldScopes.db.deleteWhere(
+          session,
+          where: (_) => Constant.bool(true),
+        ));
+
+    test(
+      'when inserting a single record into the database, non-persisted fields should retain their values after insertion, even though they are not stored in the database',
```
hampuslavin, line 16: NIT: add `then` instead of comma, applies to all tests
```suggestion
      'when inserting a single record into the database then non-persisted fields should retain their values after insertion, even though they are not stored in the database',
```

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -200,7 +210,13 @@ class DatabaseConnection {
...
 
     var result =
-        await _mappedResultsQuery(session, query, transaction: transaction);
+        (await _mappedResultsQuery(session, query, transaction: transaction))
+            .toList();
+
+    result = _mergeResultsWithNonPersistedFields(
+      rows: rows,
+      dbResults: result,
+    );
```
hampuslavin, line 219: Suggestion: Maybe this code (and the same code earlier) can be made a bit more functional to make it more declarative and easier to follow. Something along these lines should be possible:
```dart
    return (await _mappedResultsQuery(session, query, transaction: transaction))
        .map(_mergeResultWithNonPersistedFields(rows))
        .map(_poolManager.serializationManager.deserialize<T>)
        .toList();
```

Then `_mergeResultWithNonPersistedFields` would be something like:
```dart
  Map<String, dynamic> Function(T)
      _mergeResultWithNonPersistedFields<T extends TableRow>(
    Map<String, dynamic> dbResult,
  ) {
    return (T row) => {
          // Add non-persisted fields from the original object
          ...row.toJson(),
          // Override with database fields (common fields)
          ...dbResult,
        };
  }
```

klkucaj: Thanks for this, done!

tests/serverpod_test_server/test_integration/database_operations/persist/non_persist_field_test.dart
```diff
@@ -0,0 +1,145 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+import 'package:serverpod_test_server/test_util/test_serverpod.dart';
+import 'package:test/test.dart';
+
+void main() async {
```
Isakdl, line 6: Suggestion: Can you add a test with an object that has a relation that has a relation, e.g. `A <-> B <-> C`.

When doing an insert with A when B and C is present in the object then both B and C should be returned in the final object.

Isakdl: E.g. I would like to verify that nested objects that are not persistent also are taken into account.

Isakdl (changes requested):
Looking good, I have requested one addional test to cover.

tests/serverpod_test_server/test_integration/database_operations/persist/non_persist_field_test.dart
klkucaj: okay, on it

klkucaj: done!

Isakdl (approved):
LGTM

PR #2755, opened 2024-09-18
fix: Allow event listeners to remove themselves inside their handler

### LINKED ISSUE #2754: Revoked authentication does not work for multiple streams
**Describe the bug**
When revoking authentication for a user that has multiple streams open only the first stream will be closed. This is due to the fact that the revoked authentication event removes its own handler in the event callback, resulting in the following error:
```log
Concurrent modification during iteration: _Set len:0.
#0      _CompactIterator.moveNext (dart:collection-patch/compact_hash.dart:714:7)
#1      MessageCentral.postMessage (package:serverpod/src/server/message_central.dart:61:28)
```

**To Reproduce**
1. Start two streams with a single user
2. Revoke access for the user
3. Only the first stream connection is revoked and closed, the second still lives

**Expected behavior**
Both stream should be closed. Simply fixing the callback handlers iteration should fix this issue.

hampuslavin:
## Description
Fixes #2754. The revoked authentication handler removed itself inside its event handler, causing an exception to be thrown:
```log
Concurrent modification during iteration: _Set len:0.
#0      _CompactIterator.moveNext (dart:collection-patch/compact_hash.dart:714:7)
#1      MessageCentral.postMessage (package:serverpod/src/server/message_central.dart:61:28)
```
This stopped the iteration and prevented all handlers to be called. 

💡 Note that this exception was silently failing because the server will not shutdown on exceptions, but it was visible in the server log

## Changes
- Simply adds `.toList` to the callback list to create a copy. This allows the event handler to remove itself (or indeed add new handlers).
- Adds test cases which were failing before the fix (by timeout, since the second stream was never closed).

## Pre-launch Checklist

### DIFF
--- a/packages/serverpod/lib/src/server/message_central.dart
+++ b/packages/serverpod/lib/src/server/message_central.dart
@@ -58,7 +58,7 @@ class MessageCentral {
       var channel = _channels[channelName];
       if (channel == null) return true;
 
-      for (var callback in channel) {
+      for (var callback in channel.toList()) {
         callback(message);
       }
       return true;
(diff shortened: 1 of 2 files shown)


Isakdl (approved):
LGTM

PR #2775, opened 2024-09-24
refactor: TableRow implementation

### LINKED ISSUE #2711: Refactor TableRow and the way it is implemented to allow to use inheritance on Models with Database
With implementing #2710 we have the option to use the `extends` keyword in a yaml-file when declaring our model.

```yaml
class: BaseClass
fields: 
  fieldOne: String
```

```yaml
class: SubClass
extends: BaseClass
fields: 
  fieldTwo: String
```

Which generates a class that can be used like:

```dart
SubClass(fieldOne: '...', fieldTwo: '...');
```

BUT this is not yet possible with classes that have a `table` property, as a class with table-property is generated extending `TableRow` 

```dart 
abstract class MyClass extends TableRow ....
```

Change TableRow to be an interface kind of like:
```dart
abstract interface class NewTableRow {
  int? id;

  Table get table;
}
```

This way we can `implement` the TableRow instead of extending it.

```dart
abstract class SubClass extends BaseClass
    implements _i1.TableRow, _i1.ProtocolSerialization {
```

**Describe alternatives you've considered**
Allow inheritance only for data-classes but not for classes related to a database

**Additional context**

BenAuerDev:
This PR refactors the implementation of `TableRow` from:
an `abstract class` that was `extended`
to
an `abstract interface class` that is `implemented` 

with the following steps:
TableRow
1. `TableRow` was changed to be an interface
2.  Instead of the `constructor` a `getter` and a `setter` was added for `id`

ModelLibraryGenerator
1. The generated class now `implements` TableRow instead of extending it
2. The class now has a private field `_id` that is initialized in the constructor body instead of calling `super(id)`
3. Additionally it `overrides` the `getter` and `setter` from TableRow 

`entity_class_test.dart` 
1. Test no longer expects TableRow to be `extended` but `implemented`
2. Test no longer expects `id` to be passed to `super` but to be initialized in the constructor body

`insert_sql_query_test`
- Test util classes were changed in the same fashion

closes #2711 
connected to #1873 

I would recommend reviewing it commit by commit, as the PR includes the output of `generate_all` but note, I forgot to update the test in the database package at first. This is done in the last commit 

## Pre-launch Checklist

## Breaking changes
I don't think there are breaking changes as explained by @SandPod in #2711

### DIFF
--- a/packages/serverpod/lib/src/database/concepts/table.dart
+++ b/packages/serverpod/lib/src/database/concepts/table.dart
@@ -5,13 +5,11 @@ import 'package:serverpod_serialization/serverpod_serialization.dart';
 /// Holds data corresponding to a row in the database. Concrete classes are
 /// typically generated. Instances of [TableRow] can also be serialized and
 /// either passed to clients or cached.
-abstract class TableRow implements SerializableModel {
-  /// Create a new TableRow object.
-  TableRow([this.id]);
-
+abstract interface class TableRow implements SerializableModel {
   /// The id column of the row. Can be null if this row is not yet stored in
   /// the database.
-  int? id;
+  int? get id;
+  set id(int? value);
 
   /// The table that this row belongs to.
   Table get table;
--- a/packages/serverpod/lib/src/database/concepts/table.dart
+++ b/packages/serverpod/lib/src/database/concepts/table.dart
@@ -8,8 +8,7 @@ import 'package:serverpod_serialization/serverpod_serialization.dart';
 abstract interface class TableRow implements SerializableModel {
   /// The id column of the row. Can be null if this row is not yet stored in
   /// the database.
-  int? get id;
-  set id(int? value);
+  int? id;
 
   /// The table that this row belongs to.
   Table get table;
--- a/tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
@@ -160,14 +160,25 @@ class SerializableModelLibraryGenerator {
       }
 
       if (serverCode && tableName != null) {
-        classBuilder.extend =
-            refer('TableRow', 'package:serverpod/serverpod.dart');
+        classBuilder.implements.add(
+          refer('TableRow', 'package:serverpod/serverpod.dart'),
+        );
 
         classBuilder.fields.addAll([
           _buildModelClassTableField(className),
         ]);
 
         classBuilder.fields.add(_buildModelClassDBField(className));
+
+        classBuilder.fields.add(Field(
+          (f) => f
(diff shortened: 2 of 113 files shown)


packages/serverpod/lib/src/database/concepts/table.dart
```diff
@@ -5,13 +5,11 @@ import 'package:serverpod_serialization/serverpod_serialization.dart';
...
-abstract class TableRow implements SerializableModel {
-  /// Create a new TableRow object.
-  TableRow([this.id]);
-
+abstract interface class TableRow implements SerializableModel {
   /// The id column of the row. Can be null if this row is not yet stored in
   /// the database.
-  int? id;
+  int? get id;
+  set id(int? value);
```
SandPod, line 12: Question: Do we really need to change this to a setter and a getter?

It is essentially the same thing but seems a bit bloated.

Seems like it is fine to define a field directly in an interface that needs to be
 overridden by anything implementing the interface:
 
 ```dart
abstract interface class MyInterface {
  int? myInt;
}

class MyClass implements MyInterface {
  @override
  int? myInt;
}

void interfaceMethod<T extends MyInterface>(T myInterface) {
  print(myInterface.myInt);
}

void main() {
  MyClass myClass = MyClass();
  myClass.myInt = 5;
  interfaceMethod(myClass);
}
 ```

This prints out "5".

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -160,15 +160,26 @@ class SerializableModelLibraryGenerator {
       }
 
       if (serverCode && tableName != null) {
-        classBuilder.extend =
-            refer('TableRow', 'package:serverpod/serverpod.dart');
+        classBuilder.implements.add(
+          refer('TableRow', 'package:serverpod/serverpod.dart'),
```
SandPod, line 164: Question: Don't we have a thing that builds references to the serverpod package? Instead of having this hardcoded?

I think it is `serverpodUrl(serverCode)`. Maybe we should change this to use that method as well.

SandPod: This will help us if we ever change the path to the Serverpod package.

packages/serverpod/lib/src/database/concepts/table.dart
SandPod: Then we can also remove the code generated constructor logic for setting a private `_id` field.

SandPod (changes requested):
Nice clean work!

Couple of suggestions that would probably simplify this change and make the code diff a lot smaller.

SandPod (approved):
Nice clean change! 👍

PR #2784, opened 2024-09-26
Fix: Support named and optional args in test tools

### LINKED ISSUE #2783: Named and optional arguments are not generated in the test tools endpoints
**Describe the bug**
Named and optional arguments are not generated in the test tools endpoints.

hampuslavin:
## Description
Adds support for named and optional args in test tools endpoint generation.

💡 It is starting to seem like some of the generation logic should be broken out so that it can be used by both the test tools generator and the client generator. For example, if we will add support for default args that logic will need to be added in 2 places. However, I think that should be a separate PR (perhaps even as part of the default value PR).

Closes #2783 

## Changes
- Adds code generation for two more types of parameters
  - optional positional
  - named
- Adds both snapshot tests in `test_tools_test.dart` as well as interface tests from user POV in `protocol_test.dart`.
  - 💡 Note that there is an overlap between `protocol_test.dart` and `return_and_input_types_test.dart`. I was about to merge them, but in the end I thought it was a bit nice to have them separate because they are testing slightly different aspects. The idea with `protocol_test.dart` is to just test all permutations of inputs, but not advanced behaviors like two users posting to a stream. Let me know if you think they should be merged!

## Pre-launch Checklist


tests/serverpod_test_server/lib/src/endpoints/test_tools.dart
```diff
@@ -108,3 +108,63 @@ class AuthenticatedTestToolsEndpoint extends Endpoint {
     }
   }
 }
+
+class TestToolsWithArgVariationsEndpoint extends Endpoint {
```
SandPod, line 112: Discussion: Would it make sense to add all these permutations once and then test them both form the "real" generated client in e2e tests and also the test framework?

This will help us validate we actually have client and test framework support for these?

tests/serverpod_test_server/test_integration/test_tools/protocol_test.dart
```diff
@@ -0,0 +1,98 @@
...
+        () async {
+      final result =
+          await endpoints.testToolsWithArgVariations.echoOptionalArg(session);
+      expect(result, isNull);
+    });
+
+    test('when calling method with positional and named args then echoes args',
+        () async {
+      final result =
+          await endpoints.testToolsWithArgVariations.echoPositionalAndNamedArgs(
```
SandPod, line 54: Suggestion(NIT): Add test for when named arg is omitted.

To complete the matrix, this seems to have been missed.

tests/serverpod_test_server/test_integration/test_tools/protocol_test.dart
```diff
@@ -0,0 +1,98 @@
...
+      );
+
+      expect(result, ['PositionalArg', 'NamedArg']);
+    });
+
+    test(
+        'when calling method with positional and optional args then echoes both args',
+        () async {
+      final result = await endpoints.testToolsWithArgVariations
+          .echoPositionalAndOptionalArgs(
```
SandPod, line 67: Suggestion(NIT): Add test for when optional arg is omitted.

To complete the matrix, this seems to have been missed.

tools/serverpod_cli/lib/src/analyzer/dart/definitions.dart
```diff
@@ -58,6 +58,12 @@ abstract base class MethodDefinition {
   /// The named parameters of this method.
   final List<ParameterDefinition> parametersNamed;
 
+  List<ParameterDefinition> get allParameters => [
```
SandPod, line 61: Comment: Nice 🌟

SandPod: Suggestion: Would be nice to use this where possible in the library generator as well.

To ensure this refactor doesn't diverge the code.

SandPod (commented):
Nice work!

Couple of comments and suggestions.
Let me know what you think.

tests/serverpod_test_server/lib/src/endpoints/test_tools.dart
hampuslavin: Sounds like a good idea, will investigate!

tools/serverpod_cli/lib/src/analyzer/dart/definitions.dart
hampuslavin: Good suggestion, will fix!

tests/serverpod_test_server/test_integration/test_tools/protocol_test.dart
hampuslavin: 45245d4

tests/serverpod_test_server/test_integration/test_tools/protocol_test.dart
hampuslavin: 45245d4

tools/serverpod_cli/lib/src/analyzer/dart/definitions.dart
hampuslavin: b15a517

tests/serverpod_test_server/lib/src/endpoints/test_tools.dart
hampuslavin: Made an attempt here, let me know what you think of its placement/naming: 556ae12

tests/serverpod_test_server/test_integration/test_tools/protocol_test.dart
```diff
@@ -60,6 +60,31 @@ void main() {
       expect(result, ['PositionalArg', 'NamedArg']);
     });
 
+    test(
+        'when calling method with positional and nullable named args when passing value then echoes args',
```
SandPod, line 64: Suggestion: Change description to: `when calling method with positional and nullable named args with named arg then echoes args.`.

SandPod (approved):
LGTM! 🚀

tests/serverpod_test_server/test_integration/test_tools/protocol_test.dart
hampuslavin: 75fa6ce

PR #2791, opened 2024-09-27
Fix: Support transactions in endpoints with test tool

### LINKED ISSUE #2782: Doing transactions in endpoints is not compatible with test tools
**Describe the bug**
 Due to the below condition in the database repository, transactions that are done in the endpoints are not rolled back since the test tool doesn't know about it.
```dart
transaction: transaction ?? databaseAccessor.transaction,
```

**Suggested solution**
- The condition should be inverted to `transaction: databaseAccessor.transaction ?? transaction,` so that the test transaction takes precedence.
- To opt out of this behavior, automatic rollback with transactions has to be turned off for that specific `withServerpod` group.

hampuslavin:
## Description
The case where calls to `session.db.transaction` happens in endpoints is currently not handled in the best way in test tools. Those database changes will not be rolled back because the transaction is out of control of the test tools.

The options are:
- Not supporting this and throw an exception when this case happens
- Emulate "nested transactions" by hijacking the `transaction` method and use save points instead

This PR went for the second option with the following implementation.

Closes #2782

## Changes
- Adds a `DatabaseProxy` to be able to intercept calls to `transaction` 
- If the the `transaction` call is made by the user (i.e. not test tools) then one level of nesting is supported by using save points and try/catch. 
  - Emulated nested transactions are not supported in Serverpod, so one level of nesting from the test tools POV is simply normal behavior from the endpoint POV
  - Parallel calls to `transaction` becomes a hard problem to solve for nested transactions and is therefore excluded. The hypothesis is that this is somewhat rare (perhaps less than 5% of endpoints) so supporting the non-parallel case is still worthwhile.
- If parallel calls are made for any database setting other than "rollbacks disabled", then a runtime exception is thrown informing the user about this and recommending them to change the database test config setting
- Adds `packages/serverpod_test/lib/serverpod_test_public_exports.dart` to make it easier to overview what should be re-exported to end users

## Question for reviewer ❓ 
- I added `DatabaseTestConfig` initially for the first iteration. Technically it's not needed with the final solution presented here, but I ended up liking it more because it will be very extendable. It does feel like a premature optimization though, but thought I'd take a second opinion if it makes sense to keep it? (In this case maybe `resetTestSession` should be on the same format) 

## Pre-launch Checklist


tests/serverpod_test_server/test_integration/test_tools/serverpod_test_tools.dart
```diff
@@ -38,23 +38,14 @@ import 'package:serverpod_test_server/src/generated/scopes/scope_server_only_fie
...
-        AuthenticationOverride;
+export 'package:serverpod_test/serverpod_test_public_exports.dart';
 
 @_i1.isTestGroup
 withServerpod(
   String testGroupName,
   _i1.TestClosure<TestEndpoints> testClosure, {
   _i1.ResetTestSessions? resetTestSessions,
-  _i1.RollbackDatabase? rollbackDatabase,
+  _i1.DatabaseTestConfig? databaseTestConfig,
```
SandPod, line 48: Question/Discussion: Are we ok with mixing nouns and verbs here?

ResetTestSession described something that was done while databaseTestConfig is more a object we are passing in.

packages/serverpod_test/lib/src/with_serverpod.dart
```diff
@@ -49,6 +63,50 @@ typedef TestClosure<T> = void Function(
...
+
+  /// Rolls back the database after each test. This is the default.
+  /// Will throw an [InvalidConfigurationException] if several calls to `transaction` are made concurrently
+  /// since this is not supported by the test tools.
+  /// In this case, disable rolling back the database by setting `databaseTestConfig` to `DatabaseTestConfig.rollbacksDisabled()`.
+  factory DatabaseTestConfig.rollbackAfterEach() {
+    return DatabaseTestConfig._(
+      rollbackDatabase: _RollbackDatabase.afterEach,
+    );
+  }
```
SandPod, line 89: Discussion: This works in the case when we configure a single thing. But since we have created a TestConfig object we will probably want to add more things to it later?

How do you see this interface scale in that case?

I'm a bit concerned that the factory methods will be obselete as soon as we add another option.

tests/serverpod_test_server/test_integration/test_tools/database_operations_in_endpoint_test.dart
```diff
@@ -82,4 +82,327 @@ void main() {
...
+  withServerpod(
+    'Given TestToolsEndpoint and rollbackDatabase afterEach',
+    (endpoints, session) {
+      group('when calling createSimpleDatasInsideTransactions', () {
+        setUpAll(() async {
+          await endpoints.testTools
+              .createSimpleDatasInsideTransactions(session, 123);
+        });
+
+        test("then finds SimpleDatas using the test session's transaction",
```
SandPod, line 95: Suggestion: Remove "using the test session's transaction".

Since we are not specifically fetching the transaction in the test, this is more of an implementation detail. We just want to make sure that we can read the result of what has happened inside a transaction in an endpoint right?

tests/serverpod_test_server/test_integration/test_tools/database_operations_in_endpoint_test.dart
```diff
@@ -82,4 +82,327 @@ void main() {
     },
     runMode: ServerpodRunMode.production,
   );
+
+  withServerpod(
+    'Given TestToolsEndpoint and rollbackDatabase afterEach',
+    (endpoints, session) {
+      group('when calling createSimpleDatasInsideTransactions', () {
```
SandPod, line 89: Suggestion: Add to the description that his happens once, otherwise the second test description in this group doesn't really make sense on it's own.

The second test would fail if it is run in isolation right? Should we tag these "scenario" tests in some way to make it clear they are dependent on each other?

tests/serverpod_test_server/test_integration/test_tools/database_operations_in_endpoint_test.dart
```diff
@@ -82,4 +82,327 @@ void main() {
...
+
+          expect(simpleDatas, hasLength(0));
+        });
+      });
+
+      group('when calling createSimpleDataAndThrowInsideTransaction', () {
+        setUpAll(() async {
+          try {
+            await endpoints.testTools
+                .createSimpleDataAndThrowInsideTransaction(session, 123);
```
SandPod, line 114: Question: If a transaction throws and is not catched in the endpoint? Will the exception be propagated back to the test?

I'm guessing these end up in 500 if it happens in a server.

tests/serverpod_test_server/lib/src/endpoints/test_tools.dart
```diff
@@ -78,6 +78,77 @@ class TestToolsEndpoint extends Endpoint {
   Future<List<SimpleData>> getAllSimpleData(Session session) async {
     return SimpleData.db.find(session);
   }
+
+  Future<void> createSimpleDatasInsideTransactions(
+      Session session, int data) async {
```
SandPod, line 83: Suggestion(NIT): Add comma after last parameter to get a cleaner linebreak.

packages/serverpod_test/lib/serverpod_test_public_exports.dart
```diff
@@ -0,0 +1,14 @@
+library serverpod_test;
+
+export 'serverpod_test.dart'
```
SandPod, line 3: Suggestion: Flip the naming for the public and the internal exports.

We probably want the name for the export our users should be using to be as simple as possible. Then we can call the other one `_internal_exports`.

packages/serverpod_test/lib/src/with_serverpod.dart
```diff
@@ -49,6 +63,50 @@ typedef TestClosure<T> = void Function(
   TestSession testSession,
 );
 
+/// Internal helper extension for the [DatabaseTestConfig] class.
+extension ConfigRuntimeChecks on DatabaseTestConfig {
+  /// Returns true if rollbacks are disabled.
+  bool get areRollbacksDisabled => _rollbackDatabase == _RollbackDatabase.never;
+}
```
SandPod, line 70: Question: What is the motivation behind this indirection?

packages/serverpod_test/lib/src/database_proxy.dart
```diff
@@ -0,0 +1,234 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test/src/transaction_manager.dart';
+
+import 'with_serverpod.dart';
+
+/// A database proxy that forwards all calls to the provided database.
+class DatabaseProxy implements Database {
```
SandPod, line 7: Suggestion: Rename to `TestDatabaseProxy` so that it is not used in other production code.

packages/serverpod_test/lib/src/database_proxy.dart
```diff
@@ -0,0 +1,234 @@
...
+    if (localTransaction == null) {
+      throw StateError('No ongoing transaction.');
+    }
+
+    try {
+      await _transactionManager.addSavePoint(lock: true);
+    } on ConcurrentTransactionsException {
+      throw InvalidConfigurationException(
+        'Several calls to `transaction` was made concurrently which is not supported with the current database test configuration. '
+        'Disable rolling back the database by setting `databaseTestConfig` to `DatabaseTestConfig.rollbacksDisabled()`.',
```
SandPod, line 147: Suggestion: Rephrase to: "Several concurrent calls to `transaction` are not supported when database rollbacks are enabled. Disable rolling back the database by setting... to continue. This makes you responsible for restoring the database after the test as complete.".

I would like to be more distinct on what exactly is causing the exception, what action needs to be taken and possibly the consequence of this action.
Maybe the consequence could be left out of the message since it should be clear when configuring the flag?

packages/serverpod_test/lib/src/transaction_manager.dart
```diff
@@ -27,46 +34,55 @@ class TransactionManager {
     late Transaction localTransaction;
 
     unawaited(
-      serverpodSession.db.transaction((newTransaction) async {
-        localTransaction = newTransaction;
+      (serverpodSession.db as DatabaseProxy).transaction(
```
SandPod, line 37: Suggestion: Add private extention to serverpodSession that retrieve the proxy directly to remove the cast.

This is the only place we need to cast this right? Might as well just have a private extension.
I'm generally not a fan of casts so if we can avoid it, and it makes sense, I would rather pursue that option.

SandPod (changes requested):
Nice work and good research.

Really cool to be on par with other major frameworks functionality on this!

Have a couple of comments, suggestions and questions for you.

tests/serverpod_test_server/test_integration/test_tools/database_operations_in_endpoint_test.dart
hampuslavin: Yes, you are correct! This is intended as the strategy is that any exceptions that are thrown should bubble up just like when you are calling any other function.

packages/serverpod_test/lib/serverpod_test_public_exports.dart
hampuslavin: The strategy is that people should really never import anything directly from `serverpod_test`. They should always import from the generated file. We're just re-exporting this file in the generated file and in that sense I think this is very explicit.

packages/serverpod_test/lib/src/with_serverpod.dart
hampuslavin: Could be, yea. I'm not sure about this either. I will just revert back to the enum to reduce the scope of this PR.

packages/serverpod_test/lib/src/with_serverpod.dart
hampuslavin: It was needed to get the current setting, but will revert these changes: https://github.com/serverpod/serverpod/pull/2791#discussion_r1782467043

packages/serverpod_test/lib/src/database_proxy.dart
hampuslavin: Good input, that is better 👍

tests/serverpod_test_server/test_integration/test_tools/database_operations_in_endpoint_test.dart
hampuslavin: Maybe I can simply move it into the first test? And yes, I could tag these as scenario. We don't follow that convention elsewhere but maybe we want to start doing that?

tests/serverpod_test_server/test_integration/test_tools/serverpod_test_tools.dart
hampuslavin: Will revert this change as it's out of scope for the PR: https://github.com/serverpod/serverpod/pull/2791#discussion_r1782467043

tests/serverpod_test_server/test_integration/test_tools/database_operations_in_endpoint_test.dart
hampuslavin: Sorry I was unclear, I meant moving the `setUpAll` code into the first test.

tests/serverpod_test_server/test_integration/test_tools/serverpod_test_tools.dart
hampuslavin: 100408c

packages/serverpod_test/lib/src/with_serverpod.dart
hampuslavin: 100408c

packages/serverpod_test/lib/src/database_proxy.dart
```diff
@@ -0,0 +1,234 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test/src/transaction_manager.dart';
+
+import 'with_serverpod.dart';
+
+/// A database proxy that forwards all calls to the provided database.
+class TestDatabaseProxy implements Database {
```
SandPod, line 7: Suggestion: Rename file to match the class in the file.

SandPod (approved):
LGTM!

Filename change was missed that should be done before merging this.

packages/serverpod_test/lib/src/database_proxy.dart
hampuslavin: Ah shoot, thanks for catching!

PR #2799, opened 2024-10-02
feat: inheritance for DB class

BenAuerDev:
This PR adds the functionality allowing a class with DB Table to be extended aswell. 

Changes:
1. Removed restrictions disallowing inheritance on `DB classes`
2. Added restriction disallowing multiple `DB classes` in the same hierarchy
3. Removed tests checking for removed restrictions
4. Added tests checking for new restriction AND test checking if a child-class with table has both its own and its parents fields.
5. Simplifies how `fields`/`inheritedFields`/`fieldsIncludingInherited` is derived by adding a getter to `ClassDefinition` this includes a refactor of `model_library_generator.dart` 
6. Passes `fieldsIncludingInherited` instead of `fields` to `createDatabaseDefinitionFromModels` 
7. Adds e2e test checking if the migration is created correctly and table works as intended
8. fixed typo 

related to: #1873 and #1618

## Pre-launch Checklist

## Breaking changes

No breaking changes


tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -228,7 +228,8 @@ class Restrictions {
       ];
     }
 
-    var parentClass = parsedModels.findByClassName(parentClassName);
+    var parentClass =
+        parsedModels.findByClassName(parentClassName) as ClassDefinition?;
```
SandPod: Question: Is this cast to `ClassDefinition?` required? Seems like it was not needed before so we should be able to skip it here?

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -1378,4 +1384,29 @@ class Restrictions {
...
+  ClassDefinition? _getParentClass(ClassDefinition currentClass) {
+    if (currentClass.extendsClass is! ResolvedInheritanceDefinition) {
+      return null;
+    }
+
+    return (currentClass.extendsClass as ResolvedInheritanceDefinition)
+        .classDefinition;
+  }
+
+  ClassDefinition? _findOtherTableClassInHierarchy(
```
SandPod, line 1397: Suggestion: rename to `_findTableClassInParentClasses` or `_findTableClassInSuperClasses`.

Just to clarify we are only checking our parents and never our children or siblings.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -304,6 +296,20 @@ class Restrictions {
...
 
+    var currentModel =
+        parsedModels.findByTableName(tableName) as ClassDefinition;
+
+    var ancestorWithTable = _findOtherTableClassInHierarchy(currentModel);
+
+    if (ancestorWithTable != null) {
+      return [
+        SourceSpanSeverityException(
+          '${currentModel.className} cannot have a table, another class in its hierarchy already declares a table "${ancestorWithTable.className}".',
```
SandPod, line 269: Discussion: I would like to give better instructions for the action to take to fix this issue.

This is validation on the table name right? So we will be in the context of the file.


For example: `The "table" property is not allowed because another class, "${ancestorWithTable.className}", in the class hierarchy already has one defined. Only one table definition is allowed when using inheritance.`

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/inheritance_test.dart
```diff
@@ -92,21 +92,23 @@ void main() {
     });
 
     test(
-        'Given a child-class is defined with the table property, then an error is collected that child-classes cannot have a table',
+        'Given a child class of a external class, then an error is collected that only classes from within the project can be extended',
```
SandPod, line 95: Suggestion(NIT): Rename to `Given a class that extends an external class....`

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/inheritance_test.dart
```diff
@@ -126,18 +128,17 @@ void main() {
...
       expect(
         error.message,
-        'The "extends" property is mutually exclusive with the "table" property.',
+        'You can only extend classes from your own project.',
       );
     });
 
     test(
-        'Given a child-class extends a class with a table defined, then collect an error that parent class cannot have a table',
+        'Given a child-class when inheritance is not enabled, then error is collected that the "extends" property is not allowed',
```
SandPod, line 136: Suggestion(NIT): You changed `child-class` to `child class` in the previous test. Should this be changed for all tests in this file? (And possibly other files)

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -89,16 +89,32 @@ class ClassDefinition extends SerializableModelDefinition {
...
   }
 
+  /// Returns a list of all fields in the parent class.
+  /// If there is no parent class, an empty list is returned.
   List<SerializableModelFieldDefinition> get parentFields =>
       parentClass?.fields ?? [];
 
+  /// Returns a list of all fields in this class, including inherited fields.
+  /// It ensures that the 'id' field, if present, is always included at the beginning of the list.
+  List<SerializableModelFieldDefinition> get allFields {
```
SandPod, line 108: Discussion/Question: The `fields` getter and this getter are a bit confusing now. Should we rename the `fields` getter to better reflect that this is only includes fields that are defined in the class?

SandPod (changes requested):
Nice work! 🌟 

Getting closer!

I have a couple of comments and suggestions.

But I would also like to see some e2e tests that validate we can define models with the different hierarchies.

Say for example 3 tests where the table is defined in all different places in this hierarchy:
Grandparent -> Parent -> Child.

This would help us validate that the migrations are generated correctly.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
BenAuerDev: or we change `allFields` to be more descriptive?

tests/serverpod_test_server/test_integration/inheritance_test.dart
```diff
@@ -0,0 +1,44 @@
...
+      .deleteWhere(session, where: (element) => Constant.bool(true));
+}
+
+void main() async {
+  var session = await IntegrationTestServer().session();
+
+  tearDownAll(() async => await _deleteAll(session));
+
+  test(
+      'Given a class extends another class, then the child-class is a sub-type parent-class',
```
SandPod, line 17: Suggestion(NIT): 
```suggestion
      'Given a class that extends another class, then the child-class is a sub-type of the parent-class',
```

tests/serverpod_test_server/test_integration/inheritance_test.dart
```diff
@@ -0,0 +1,44 @@
...
+      grandParentField: 'grandParentField',
+      parentField: 'parentField',
+      childField: 2,
+    );
+
+    expect(childClass, isA<ParentClass>());
+  });
+
+  test(
+      'Given a ChildClass is instantiated and inserted into the ParentClass table, then the ChildClass should be retrievable from the ParentClass table',
```
SandPod, line 29: Suggestion: Use the same phrasing for all tests.

Here you use ChildClass, in the test above you use child-class. Please use one type.

SandPod: Suggestion: Use Given, When, Then structure.
```suggestion
      'Given an instantiated ChildClass when inserted into the ParentClass table, then the ChildClass should be retrievable from the ParentClass table',
```

The `When` part is phrased differently here.

tests/serverpod_test_server/test_integration/inheritance_test.dart
```diff
@@ -0,0 +1,44 @@
...
+      parentField: 'parentField',
+      childField: 2,
+    );
+
+    var childInParentDb = await ParentClass.db.insertRow(session, childClass);
+    var parentDbFirstRow = await ParentClass.db.findFirstRow(session);
+
+    expect(childInParentDb.id, parentDbFirstRow!.id);
+    expect(childInParentDb.grandParentField, parentDbFirstRow.grandParentField);
+    expect(childInParentDb.parentField, parentDbFirstRow.parentField);
```
SandPod, line 42: Suggestion: Compare to the instantiated child class, except for the ID, the id should come from the `childInParentDb` instance.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -89,15 +89,31 @@ class ClassDefinition extends SerializableModelDefinition {
...
-  List<SerializableModelFieldDefinition> get parentFields =>
-      parentClass?.fields ?? [];
+  /// Returns a list of all fields in the parent class.
+  /// If there is no parent class, an empty list is returned.
+  List<SerializableModelFieldDefinition> get inheritedFields =>
+      parentClass?.fieldsIncludingInherited ?? [];
+
+  /// Returns a list of all fields in this class, including inherited fields.
+  /// It ensures that the 'id' field, if present, is always included at the beginning of the list.
+  List<SerializableModelFieldDefinition> get fieldsIncludingInherited {
```
SandPod, line 108: Question: Should this really be `fields` and then `fields` should be `modelDefinedFields`?

SandPod: (This can always be changed later so not a blocker for this PR)

SandPod (commented):
Nice work, last couple of comments and suggestions!

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
BenAuerDev: cool I'll make a small PR for that change then :)

tests/serverpod_test_server/test_integration/inheritance_test.dart
```diff
@@ -0,0 +1,44 @@
...
+      grandParentField: 'grandParentField',
+      parentField: 'parentField',
+      childField: 2,
+    );
+
+    expect(childClass, isA<ParentClass>());
+  });
+
+  test(
+      'Given an instantiated child-class when inserted into the parent-class table, then the child-class should be retrievable from the parent-class table',
```
SandPod, line 29: Suggestion(NIT):
```suggestion
      'Given an instantiated child-class when inserted into the parent-class table, then inherited fields should be retrievable from the parent-class table',
```

SandPod (approved):
LGTM!

Once last thing to fix.

Issue #2800, opened 2024-10-02
Passwords from the env are ignored if the passwords.yaml does not exist.


Isakdl:
I have right now tried to run the whole setup, but experience an error where serverpod doesn't seem to read to Database Password from the environment properly.

```
Exception: Missing database password.
#0      new DatabaseConfig._fromJson (package:serverpod_shared/src/config.dart:300)
#1      new ServerpodConfig.loadFromMap (package:serverpod_shared/src/config.dart:113)
#2      new ServerpodConfig.load (package:serverpod_shared/src/config.dart:157)
#3      new Serverpod (package:serverpod/src/server/serverpod.dart:308)
#4      run (package:vps_deployment_server/server.dart:14)
#5      main (file:///app/bin/main.dart:6)
```

Currently using serverpod 2.1.1

~Any quick thoughts what might be behind this?~

**EDIT**

~Not entirely sure, but I think the issue is with this code~

```dart
  /// Load all passwords for the current run mode, or null if passwords fail
  /// to load.
  Map<String, String>? loadPasswords() {
    try {
      var passwordYaml = File('config/passwords.yaml').readAsStringSync();
      var data = (loadYaml(passwordYaml) as Map).cast<String, Map>();

      return loadPasswordsFromMap(data, environment: Platform.environment);
    } catch (e) {
      return null;
    }
  }
```

Since there is no passwords.yaml file, the whole try catch fails, even though all required values might just be in the environment.

**EDIT-2**

The missing passwords.yaml file is indeed the problem, once I create a passwords.yaml file, everything works. This also happens if you do not supply the production.yaml file or leave it empty (comments do not work). The yaml files must contain a value key-value pair to prevent this from breaking Serverpod.

_Originally posted by @inf0rmatix in https://github.com/serverpod/serverpod/issues/2124#issuecomment-2381312218_

PR #2801, opened 2024-10-02
fix: Load passwords from env even if the passwords file does not exist.

Isakdl:
# Changes

Ignore the passwords.yaml file if it does not exist but still load the passwords from env if possible.

Closes: https://github.com/serverpod/serverpod/issues/2800

## Pre-launch Checklist

## Breaking changes

none

### DIFF
--- a/packages/serverpod_shared/test_integration/assets/load_passwords/runner_empty_passwords.dart
+++ b/packages/serverpod_shared/test_integration/assets/load_passwords/runner_empty_passwords.dart
@@ -1,0 +1,10 @@
+import 'dart:io';
+import 'package:serverpod_shared/serverpod_shared.dart';
+
+void main() {
+  var passwords = PasswordManager(runMode: 'development').loadPasswords(
+    './config/empty_passwords.yaml',
+  );
+
+  stdout.write(passwords.toString());
+}
--- a/packages/serverpod_shared/test_integration/assets/load_passwords/runner_missing_passwords.dart
+++ b/packages/serverpod_shared/test_integration/assets/load_passwords/runner_missing_passwords.dart
@@ -1,9 +1,12 @@
 import 'dart:io';
 import 'package:serverpod_shared/serverpod_shared.dart';
+import 'package:uuid/uuid.dart';
 
 void main() {
+  var missingFileName = const Uuid().v4();
+
   var passwords = PasswordManager(runMode: 'development').loadPasswords(
-    './config/missing_passwords.yaml',
+    './config/$missingFileName.yaml',
   );
 
   stdout.write(passwords.toString());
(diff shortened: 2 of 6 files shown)


packages/serverpod_shared/test_integration/assets/load_passwords/runner_missing_passwords.dart
```diff
@@ -0,0 +1,10 @@
+import 'dart:io';
+import 'package:serverpod_shared/serverpod_shared.dart';
+
+void main() {
```
SandPod, line 4: Question: do we really need these two tests in different files? Seems like they are testing the same thing just with different configurations?

packages/serverpod_shared/test_integration/assets/load_passwords/runner_missing_passwords.dart
```diff
@@ -0,0 +1,10 @@
...
+  var passwords = PasswordManager(runMode: 'development').loadPasswords(
+    './config/missing_passwords.yaml',
```
SandPod, line 6: Suggestion: use a generated uuid name for the file to ensure it doesn't exist.

This test will start failing if an a file is created called this.

packages/serverpod_shared/test_integration/assets/load_passwords/runner_empty_passwords.dart
```diff
@@ -0,0 +1,10 @@
+import 'dart:io';
+import 'package:serverpod_shared/serverpod_shared.dart';
+
+void main() {
+  var passwords = PasswordManager(runMode: 'development').loadPasswords(
+    './config/empty_passwords.yaml',
```
SandPod, line 6: Suggestion: create this file as part of the test.

Since the file doesn't have a complex structure and is only used for a single test we might as well create it temporarily for the specific test.

SandPod (approved):
Good find! This probably fixes an issue around a discussion I answered.

Have a couple of minor suggestions.

Approving and trust your judgement on the comments. Let me know if you need another round of eyes.

packages/serverpod_shared/test_integration/assets/load_passwords/runner_missing_passwords.dart
Isakdl: The reason we need this is because we have to run this code in a different process to be able to inject the environment variables. To run them in a different process we need a main method, hence this file.

Isakdl: And I'm too lazy to build some arg parser for this, seems a little overkill for the current need.

packages/serverpod_shared/test_integration/assets/load_passwords/runner_empty_passwords.dart
Isakdl: We could but since the file location is hardcoded in this file which is disconnected from the test it self I think it is nice to keep the file checked in.

Issue #2804, opened 2024-10-02
MethodStreamManager: Type Mismatch in onTimeout callback for closeAllStreams() Method


KristijanZic:
**Describe the bug**
When a connection with a stream that has a controller defined in the endpoint gets cut, the closing errors out instead of being handled gracefully causing all the cancel or addWillCloseListener not to run. Specifically, the error is:

```sh
2024-10-02 13:15:35.798186Z Internal server error. Zoned exception.
type '() => void' is not a subtype of type '(() => Future<Null>?)?' of 'onTimeout'
#0      Future.timeout (dart:async/future_impl.dart:925:54)
#1      MethodStreamManager.closeAllStreams.<anonymous closure> (package:serverpod/src/server/websocket_request_handlers/helpers/method_stream_manager.dart:169:38)
#2      MappedListIterable.elementAt (dart:_internal/iterable.dart:425:31)
#3      ListIterator.moveNext (dart:_internal/iterable.dart:354:26)
#4      new _GrowableList._ofEfficientLengthIterable (dart:core-patch/growable_array.dart:189:27)
#5      new _GrowableList.of (dart:core-patch/growable_array.dart:150:28)
#6      new List.of (dart:core-patch/array_patch.dart:39:18)
#7      MethodStreamManager.closeAllStreams (package:serverpod/src/server/websocket_request_handlers/helpers/method_stream_manager.dart:175:23)
#8      MethodWebsocketRequestHandler.handleWebsocket (package:serverpod/src/server/websocket_request_handlers/method_websocket_request_handler.dart:89:33)
<asynchronous suspension>
```

**To Reproduce**
To reproduce the behavior create the following endpoint, listen to it on from the app and then quit the app.
```dart
class PokemonStreamEndpoint extends Endpoint {
  // ensure that the user is logged in (serverpod functionality)
  @override
  bool get requireLogin => true;

  // Stream Methods
  final pokemonStreamController = StreamController<PokemonDTO?>.broadcast(
    onCancel: () {
      print(" CANCELLING POKEMON STREAM");
    },
    onListen: () {
      print("LISTENING TO POKEMON STREAM");
    },
  );

  Stream<PokemonDTO?> pokemonStream(Session session) {
    return pokemonStreamController.stream;
  }
}
```

**Expected behavior**
The code should execute without a type mismatch, allowing the `onTimeout` function to run without causing an error.

**Screenshots**
If applicable, add screenshots to help explain the problem.

**Serverpod versions**
[...]

PR #2805, opened 2024-10-02
fix: Correct type mismatch in onTimeout callback for subscription cancellation in MethodStreamManager closeAllStreams()

KristijanZic:
It fixes the #2804 issue

## Pre-launch Checklist

### DIFF
--- a/packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
+++ b/packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
@@ -168,7 +168,10 @@ class MethodStreamManager {
     var closeSubscriptionFutures = outboundStreamContexts.map(
       (c) => c.subscription.cancel().timeout(
             _closeTimeout,
-            onTimeout: () => c.controller.onCancel?.call(),
+            onTimeout: () async {
+              await c.controller.onCancel?.call();
+              return null;
+            } as Future<Null> Function()?,
           ),
     );
 
--- a/packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
+++ b/packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
@@ -171,6 +171,8 @@ class MethodStreamManager {
             onTimeout: () async {
               await c.controller.onCancel?.call();
               return null;
+              // This type case is needed to avoid a runtime exception
+              // Filed as bug on dart-lang/sdk: https://github.com/dart-lang/sdk/issues/56846
             } as Future<Null> Function()?,
           ),
     );
--- a/tests/serverpod_test_server/test_integration/method_streaming_cancel_test.dart
+++ b/tests/serverpod_test_server/test_integration/method_streaming_cancel_test.dart
@@ -1,0 +1,32 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:test/test.dart';
+
+import 'test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod(
+    'Given call to MethodStreamingEndpoint',
+    (endpoints, session) {
+      test(
+          'when calling an endpoint returning a non-broadcast stream and cancelling '
+          'then will cancel', () async {
+        var stream = endpoints.methodStreaming.intStreamFromValue(session, 1);
+        var subscription = stream.listen((event) {});
+
+        await expectLater(subscription.cancel(), completes);
+      });
+
+      test(
+        'when calling an endpoint returning a broadcast stream and cancelling '
+        'then will cancel',
+        () async {
+          var stream = endpoints.methodStreaming.getBroadcastStream(session);
+          var subscription = stream.listen((event) {});
+
+          await expectLater(subscription.cancel(), completes);
+        },
+      );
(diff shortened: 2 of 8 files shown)


KristijanZic:

hampuslavin:
@KristijanZic Thanks for reporting this issue and pin pointing the location of the error 🙌 I'll just add a test to this as well so we can get it merged quickly!

tests/serverpod_test_server/test_integration/method_streaming_cancel_test.dart
```diff
@@ -0,0 +1,32 @@
...
+      });
+
+      test(
+        'when calling an endpoint returning a broadcast stream and cancelling '
+        'then will cancel',
+        () async {
+          var stream = endpoints.methodStreaming.getBroadcastStream(session);
+          var subscription = stream.listen((event) {});
+
+          await expectLater(subscription.cancel(), completes);
```
hampuslavin, line 26: @SandPod This test was throwing the reported error (see below) before the fix was applied. So the current assertion should be enough and I'm not convinced we need more. What we could do is as we discussed, namely add some variable to the endpoint that is set upon cancel to ensure it was executed as intended. However, this expands the scope of what we consider "correct" to be. By the same reasoning we should also then ensure the `session` was closed etc. WDYT?
```
type '() => void' is not a subtype of type '(() => Future<Null>?)?' of 'onTimeout'
```

packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
```diff
@@ -168,7 +168,10 @@ class MethodStreamManager {
     var closeSubscriptionFutures = outboundStreamContexts.map(
       (c) => c.subscription.cancel().timeout(
             _closeTimeout,
-            onTimeout: () => c.controller.onCancel?.call(),
+            onTimeout: () async {
+              await c.controller.onCancel?.call();
+              return null;
+            } as Future<Null> Function()?,
```
SandPod, line 174: Discussion: Should we add a comment here explaining the behavior? Or is our test enough? 

This is kind of funky so might be worth a comment.

tests/serverpod_test_server/test_integration/method_streaming_cancel_test.dart
SandPod: The test is probably enough!

Do we think it is enough that the test is applied through our test framework? Might be good enough?

SandPod (commented):
Big thanks @KristijanZic for finding and fixing this issue.

Just a couple of minor comments for you @hampuslavin.

packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
hampuslavin: Good point, I'll add a comment 👍

tests/serverpod_test_server/test_integration/method_streaming_cancel_test.dart
hampuslavin: I think we could consider it enough, since we are (at least trying to) guaranteeing that the test tool should take the same code path; in this case the MethodStreamManager behind the scenes.

hampuslavin: Did you have any idea for alternative?

SandPod: Not really, we should guarantee this on the Server side so this should be a concern there.

Since the issue was specifially broadcast stream related I think it makes sense to just make sure that codepath is exercided on the Server somewhere. Let's go with this!

hampuslavin: Found a bug in the test stream manager that was fixed. I also extended the tests a bit to ensure everything is cleaned up properly: 
23e28bd

tests/serverpod_test_server/test_integration/method_streaming_cancel_test.dart
```diff
@@ -0,0 +1,61 @@
...
+      test(
+          'when calling an endpoint returning a broadcast stream and cancelling '
+          'then it should close the session and call its will close listener',
+          () async {
+        var wasSessionWillCloseListenerCalled = endpoints.methodStreaming
+            .wasSessionWillCloseListenerCalled(session);
+
+        var stream = endpoints.methodStreaming.getBroadcastStream(session);
+        // Wait for the stream to be created, otherwise cancel is called before creation
+        await flushMicrotasks();
```
SandPod, line 48: Comment: Microtasks are separate from events in the flutter async loop. I think the intention is to flush all events in the event loop right? Might need to clarify the naming of this method in that case: https://web.archive.org/web/20170704074724/https://webdev.dartlang.org/articles/performance/event-loop

SandPod (approved):
LGTM! 🚀

tests/serverpod_test_server/test_integration/method_streaming_cancel_test.dart
hampuslavin: Good catch, will rename in this pr: https://github.com/serverpod/serverpod/pull/2816

PR #2816, opened 2024-10-04
fix: Clean up and refactor test tools

### LINKED ISSUE #2810: Refactor test tools and remove `resetTestSessions`
After reworking and adding features in multiple rounds, some abstractions no longer makes sense:
- `resetTestSessions` should be removed. Session should always be closed/recreated after each test case. It should maintain all its properties such as auth state.
- Close the copied session inside the generated test tools endpoint code (to trigger on closed listeners and logging)
- Create a way to expose the underlying "Serverpod session"

hampuslavin:
## Description
Closes #2810.

## Changes
- Adds missing ignore for file directive
- Removes a redundant leftover `await` in generated code
- For future returning functions the session is now correctly closed after the call is finished (for streams this is already handled by the `method_stream_manager`)
- Removes `resetTestSessions` as the behavior no longer makes sense
- Exposes the underlying Serverpod session. After reasoning about this, it really is quite limiting to not be able to get this ones. For instance, one might have business logic that needs a `Session` encapsulated outside the endpoint itself for better isolation. 

## Pre-launch Checklist


tests/serverpod_test_server/test_integration/test_tools/authentication_validation_test.dart
```diff
@@ -131,7 +131,7 @@ void main() {
             'and the authenticated user is revoked then stream is closed with ConnectionClosedException.',
             () async {
           await expectLater(
-            session.messages.authenticationRevoked(
+            session.serverpodSession.messages.authenticationRevoked(
```
SandPod, line 134: Question: Now that the "session" is not really returned, should we change the name of session in the `withServerpod` interface to reflect this?

SandPod (approved):
LGTM! 🚀

Single question around naming for the session.

tests/serverpod_test_server/test_integration/test_tools/authentication_validation_test.dart
hampuslavin: As discussed offline we'll create a follow-up PR for this 👍

PR #2820, opened 2024-10-04
fix: Refactor `withServerpod` to provide a session builder 

### LINKED ISSUE #2819: Refactor `withServerpod` test helper
As discussed offline, the following changes should be made as a final touch up:
- Change `TestSession` to a `TestSessionBuilder`
- Change argument order so that `TestSessionBuilder` comes first. Session might be used on its own but endpoints will always be used together with a session.

hampuslavin:
## Description
Closes #2819.

To create a better and more intuitive mental model the `session` argument is switched out for a `sessionBuilder`. This achieves a few things:
- It doesn't create a false expectation that the session that is passed to the endpoint call is the one actually being used (right now it's being copied to ensure it's unique for each endpoint call, but this is opaque to the developer). Passing a builder makes it explicit that the developer only creates the conditions for the session, but the creation of the Serverpod `Session` happens inside the test tools endpoint call.
- It solves the problem that the underlying Serverpod `Session` has to be exposed to enable a developer to test business logic that is separated out from an endpoint but needs a `Session` passed to it. Currently the mental model is very messy, because there is kind of a session (a `TestSession`) but then a `.serverpodSession` of type `Session` is also exposed to enable this use case. Only having a single session concept to keep in mind while still solving this case is a big win.

## Changes
- Changes `TestSession` into a `TestSessionBuilder` that has a `build` method that returns a `Session`.
- Switches the order of the test closure's arguments and puts `sessionBuilder` first

## Pre-launch Checklist


packages/serverpod_test/lib/src/test_session_builder.dart
```diff
@@ -4,135 +4,115 @@ import 'test_serverpod.dart';
...
-      serverpodSession: newServerpodSession,
+      mainServerpodSession: _mainServerpodSession,
     );
   }
 
+  /// Creates a new Serverpod session with the specified [endpoint] and [method].
+  /// Should only be used by generated code.
+  Session internalBuild({
+    String endpoint = '<not applicable: created by user in test>',
+    String method = '<not applicable: created by user in test>',
```
hampuslavin, line 63: I thought this could be handy for debugging purposes if for some reason someone had some logic around this.

SandPod (approved):
LGTM! :)

PR #2828, opened 2024-10-07
fix: Fixes #2826, HTTP code 500 instead of 400 upon param errors

### LINKED ISSUE #2826: Serverpod server responds with HTTP code 500 instead of 400 upon wrong endpoint parameters
## Version:

SERVERPOD version: 2.1.1, dart: 3.5.3 (stable) (Wed Sep 11 16:22:47 2024 +0000) on "macos_arm64", time: 2024-10-07 13:43:46.766788Z

## To reproduce:

Start a serverpod mini project, e.g. the example project with the endpoint `example`.

Invoke the endpoint method using curl with first correct parameters and then incorrect parameters. The second case should return the HTTP error code 400 (bad request) instead of 500 (internal server error).

```
$ curl -i "localhost:8080/example" -H "Content-Type: application/json" -d '{"method": "hello", "name": "Christer"}'
HTTP/1.1 200 OK
transfer-encoding: chunked
access-control-allow-origin: *
access-control-allow-methods: POST
x-frame-options: SAMEORIGIN
content-type: application/json; charset=utf-8
x-xss-protection: 1; mode=block
x-content-type-options: nosniff

"Hello Christer, the time is 2024-10-07 15:44:42.711563"

$ curl -i "localhost:8080/example" -H "Content-Type: application/json" -d '{"method": "hello"}'
HTTP/1.1 500 Internal Server Error
transfer-encoding: chunked
access-control-allow-origin: *
access-control-allow-methods: POST
x-frame-options: SAMEORIGIN
content-type: text/plain; charset=utf-8
x-xss-protection: 1; mode=block
x-content-type-options: nosniff

Internal server error. Call log id: 1051605082

```

## Server error log is:

```
Instance of 'InvalidParametersException'
#0      parseParameters (package:serverpod/src/server/endpoint_parameter_helper.dart:26:7)
#1      EndpointDispatch._getEndpointMethodConnector (package:serverpod/src/server/endpoint_dispatch.dart:157:27)
<asynchronous suspension>
#2      EndpointDispatch.getMethodCallContext (package:serverpod/src/server/endpoint_dispatch.dart:120:9)
<asynchronous suspension>
#3      Server._handleUriCall (package:serverpod/src/server/server.dart:466:31)
<asynchronous suspension>
#4      Server._handleRequest (package:serverpod/src/server/server.dart:293:18)
<asynchronous suspension>
#5      Server._handleRequestWithErrorBoundary (package:serverpod/src/server/server.dart:182:5)
<asynchronous suspension>

```

christerswahn:
Fixes the issue where Serverpod server responds with HTTP code 500 instead of 400 upon wrong endpoint parameters.

Closes #2826 

## Pre-launch Checklist

## Breaking changes
Breaks previous error response behavior, but that behavior was wrong and upon discussion in team this was deemed ok.

### DIFF
--- a/packages/serverpod/lib/src/server/endpoint_dispatch.dart
+++ b/packages/serverpod/lib/src/server/endpoint_dispatch.dart
@@ -449,9 +449,24 @@ class ResultInvalidParams extends Result {
   }
 }
 
+/// The result of a failed [Endpoint] method call where the
+/// endpoint was not found.
+class ResultNoSuchEndpoint extends Result {
+  /// Description of the error.
+  final String errorDescription;
+
+  /// Creates a new [ResultNoSuchEndpoint] object.
+  ResultNoSuchEndpoint(this.errorDescription);
+
+  @override
+  String toString() {
+    return errorDescription;
+  }
+}
+
 /// The result of a failed [EndpointDispatch.getMethodStreamCallContext],
 /// [EndpointDispatch.getMethodCallContext] or [EndpointDispatch.getEndpointConnector] call.
-abstract class GetEndpointMethodException implements Exception {
+abstract class EndpointDispatchException implements Exception {
   /// Description of the error.
   String get message;
 
@@ -462,7 +477,7 @@ abstract class GetEndpointMethodException implements Exception {
 }
 
 /// The user is not authorized to access the endpoint.
-class NotAuthorizedException extends GetEndpointMethodException {
+class NotAuthorizedException extends EndpointDispatchException {
   @override
   String message;
 
@@ -475,7 +490,7 @@ class NotAuthorizedException extends GetEndpointMethodException {
 }
 
 /// The endpoint was not found.
-class EndpointNotFoundException extends GetEndpointMethodException {
+class EndpointNotFoundException extends EndpointDispatchException {
   @override
   String message = 'Endpoint not found';
 
@@ -484,7 +499,7 @@ class EndpointNotFoundException extends GetEndpointMethodException {
 }
 
 /// The endpoint method was not found.
-class MethodNotFoundException extends GetEndpointMethodException {
+class MethodNotFoundException extends EndpointDispatchException {
   @override
   String message = 'Method not found';
 
@@ -493,7 +508,7 @@ class MethodNotFoundException extends GetEndpointMethodException {
 }
 
 /// The found endpoint method was not of the expected type.
-class InvalidEndpointMethodTypeException extends GetEndpointMethodException {
+class InvalidEndpointMethodTypeException extends EndpointDispatchException {
   @override
   String get message =>
       'Endpoint method $_methodName in $_endpointPath is not of the expected type.';
@@ -507,7 +522,7 @@ class InvalidEndpointMethodTypeException extends GetEndpointMethodException {
 }
 
 /// The input parameters were invalid.
-class InvalidParametersException extends GetEndpointMethodException {
+class InvalidParametersException extends EndpointDispatchException {
   @override
   String message = 'Invalid parameters';
 
--- a/packages/serverpod/lib/src/server/endpoint_dispatch.dart
+++ b/packages/serverpod/lib/src/server/endpoint_dispatch.dart
@@ -449,14 +449,20 @@ class ResultInvalidParams extends Result {
   }
 }
 
-/// The result of a failed [EndpointDispatch.getMethodStreamCallContext], [EndpointDispatch.getMethodCallContext] or [EndpointDispatch.getEndpointConnector] call.
-abstract class GetAuthorizedEndpointMethodException implements Exception {
+/// The result of a failed [EndpointDispatch.getMethodStreamCallContext],
+/// [EndpointDispatch.getMethodCallContext] or [EndpointDispatch.getEndpointConnector] call.
+abstract class GetEndpointMethodException implements Exception {
   /// Description of the error.
   String get message;
+
+  @override
+  String toString() {
+    return 'Endpoint dispatch error: $message';
+  }
 }
 
 /// The user is not authorized to access the endpoint.
-class NotAuthorizedException implements GetAuthorizedEndpointMethodException {
+class NotAuthorizedException extends GetEndpointMethodException {
   @override
   String message;
 
@@ -469,8 +475,7 @@ class NotAuthorizedException implements GetAuthorizedEndpointMethodException {
 }
 
 /// The endpoint was not found.
-class EndpointNotFoundException
-    implements GetAuthorizedEndpointMethodException {
+class EndpointNotFoundException extends GetEndpointMethodException {
   @override
   String message = 'Endpoint not found';
 
@@ -479,7 +484,7 @@ class EndpointNotFoundException
 }
 
 /// The endpoint method was not found.
-class MethodNotFoundException implements GetAuthorizedEndpointMethodException {
+class MethodNotFoundException extends GetEndpointMethodException {
   @override
   String message = 'Method not found';
 
@@ -488,8 +493,7 @@ class MethodNotFoundException implements GetAuthorizedEndpointMethodException {
 }
 
 /// The found endpoint method was not of the expected type.
-class InvalidEndpointMethodTypeException
-    implements GetAuthorizedEndpointMethodException {
+class InvalidEndpointMethodTypeException extends GetEndpointMethodException {
   @override
   String get message =>
       'Endpoint method $_methodName in $_endpointPath is not of the expected type.';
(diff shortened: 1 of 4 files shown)


packages/serverpod/lib/src/server/server.dart
```diff
@@ -526,11 +527,8 @@ class Server {
       return ResultInvalidParams(e.message);
     } on NotAuthorizedException catch (e) {
       return e.authenticationFailedResult;
-    } on InvalidParametersException catch (e, stackTrace) {
```
christerswahn, line 529: This change could in theory be a breaking change. Is the sessionLogId used / needed in this error scenario? Note that this code seems to be rather old, >4 ys, if I parsed the history correctly.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -449,14 +449,20 @@ class ResultInvalidParams extends Result {
   }
 }
 
-/// The result of a failed [EndpointDispatch.getMethodStreamCallContext], [EndpointDispatch.getMethodCallContext] or [EndpointDispatch.getEndpointConnector] call.
-abstract class GetAuthorizedEndpointMethodException implements Exception {
+/// The result of a failed [EndpointDispatch.getMethodStreamCallContext],
+/// [EndpointDispatch.getMethodCallContext] or [EndpointDispatch.getEndpointConnector] call.
+abstract class GetEndpointMethodException implements Exception {
```
vlidholt, line 454: Is `EndpointDispatchException` a better name?

vlidholt (commented):
LGTM, with suggested change. But someone else should have a look too. 🙂

packages/serverpod/lib/src/server/server.dart
SandPod: TLDR; Change looks safe to me! 👍 

### Logging Id
Logically speaking I don't think logging the Session Id makes a lot of sense until the request has reached the endpoint. Before that the session hasn't really been "opened" even if the object is created.

The session id is also unique for each method call session created (the override is not used in this codepath): https://github.com/serverpod/serverpod/blob/a19e1660f236423349a65340a8b6a2fa9630a4a8/packages/serverpod/lib/src/server/session.dart#L145

So it won't help with debugging calls to the client.

In my opinion this is fine to remove.

### Logging in general
Logging changes level, we won't automatically print these to the server logs if a flag (configured through insights or server start) is not enabled. But I don't think this is a bad thing.

Now: https://github.com/serverpod/serverpod/blob/a19e1660f236423349a65340a8b6a2fa9630a4a8/packages/serverpod/lib/src/server/server.dart#L298-L302

Before: 
https://github.com/serverpod/serverpod/blob/a19e1660f236423349a65340a8b6a2fa9630a4a8/packages/serverpod/lib/src/server/server.dart#L320-L322

### Client representation
On the client side we will change the representation of these errors: https://github.com/serverpod/serverpod/blob/a19e1660f236423349a65340a8b6a2fa9630a4a8/packages/serverpod_client/lib/src/serverpod_client_shared_private.dart#L46-L53

But since this is an already known exception that had to be handled before I think this is completely fine.

SandPod: My only though is if we should return a 404 if the endpoint is not found? I'm not sure that makes a lot of sense since the endpoint dispatching depends on request body?

tests/serverpod_test_server/test_e2e/http_codes_test.dart
```diff
@@ -0,0 +1,73 @@
...
+        Uri.parse('${serverUrl}simple'),
+        body: jsonEncode({
+          'method': 'hello',
+          'name': 'Starbase Alpha',
+          'extra': 'spurious value',
+        }),
+      );
+
+      expect(response.statusCode, 400);
+    }, skip: 'desired behavior unclear');
```
SandPod, line 56: Suggestion: The desired behavior here is to ignore any additional parameters. We have tests validating this behavior for the parameter parsing.

I think this is usually common practice when parsing query parameters, but it also enables us to be backwards compatible with old clients even if a parameter is removed from an endpoint.

tests/serverpod_test_server/test_e2e/http_codes_test.dart
```diff
@@ -0,0 +1,73 @@
...
+
+    test(
+        'when calling an endpoint method with misspelled method name '
+        'then it should respond with 400 bad request', () async {
+      var response = await http.post(
+        Uri.parse('${serverUrl}simpleMistake'),
```
SandPod, line 62: Suggestion: To make this test more robust you could generate the method name with a UUID.

"misspelled" or "non-existing" does not really matter for this case. To make this test resistant towards someone actually introducing an endpoint called "simpleMistake" we could just as well generate a UUID for the method name and prove the same point.

SandPod (commented):
Nice work!

Glad to get correct HTTP codes for invalid requests. This should reduce invaluable error logging on the server quite a bit.

Just had one though around the test for additional params. We might as well document the current desired behavior.

tests/serverpod_test_server/test_e2e/http_codes_test.dart
christerswahn: Playing devil's advocate:
On the other hand, removing a parameter is a customer-space decision, and perhaps the framework shouldn't force that sort of backwards compatibility and thereby preventing what can be desired error handling.

Also, parameters are sent in the message body so it's not a case of HTTP query convention, rather it is a Serverpod protocol decision.

And finally, getting an intelligable error message when misspelling in Postman or on the Curl command line is superior UX! :)

christerswahn: BUT even if so, perhaps now is not the time to change this design which has been decided upon and there are tests for!

SandPod: Good points!

> ...parameters are sent in the message body so it's not a case of HTTP query convention rather it is a Serverpod protocol decision.

Very true, this is a design decision in the Serverpod protocol. We can deviate from normal query parsing conventions here if we have reason to do so.

> ...getting an intelligable error message when misspelling in Postman or on the Curl...

I'm not sure I follow how it would help in this case? 

Regardless, changing this would be a breaking change for existing clients so it is not actionable at this time.

packages/serverpod/lib/src/server/server.dart
christerswahn: 404 when the path name (endpoint class name) is wrong would actually be more compliant with HTTP. Actual method name should be 400 as you say. I'll look into it!

packages/serverpod/lib/src/server/server.dart
```diff
@@ -425,7 +436,8 @@ class Server {
     var path = uri.pathSegments.join('/');
     var endpointComponents = path.split('.');
     if (endpointComponents.isEmpty || endpointComponents.length > 2) {
-      return ResultInvalidParams('Endpoint $path is not a valid endpoint name');
+      return ResultNoSuchEndpoint(
```
SandPod, line 439: Suggestion: Revert this change.

As far as I understand, this is more of a formatting exception/invalid params exception rather than a no such endpoint exception right? And how does this tie together with the error validation done on line 472?

SandPod: Hampus made some effort consolidating our endpoint parsing logic so that it is a shared path regardless of how the protocol layer is implemented. So endpoint dispatch currently owns the knowledge about possible endpoints.

I think it is enough returning this exception based on that knowledge, as you have already implemented on line 538.

tests/serverpod_test_server/test_e2e/http_codes_test.dart
```diff
@@ -0,0 +1,92 @@
...
+          'method': 'hello',
+        }),
+      );
+
+      expect(response.statusCode, 400);
+      expect(response.body, contains('Missing required query parameter: name'));
+    });
+
+    test(
+        'when calling an endpoint method with misspelled endpoint path '
```
SandPod, line 45: Suggestion(NIT): 
```suggestion
        'when calling an endpoint method with non-existing endpoint path '
```

tests/serverpod_test_server/test_e2e/http_codes_test.dart
```diff
@@ -0,0 +1,92 @@
...
+          'name': 'Starbase Alpha',
+        }),
+      );
+
+      expect(response.statusCode, 404);
+      expect(response.body, contains('Endpoint $nonExistingPath not found'));
+    });
+
+    test(
+        'when calling an endpoint method with misspelled method name '
```
SandPod, line 62: Suggestion(NIT):
```suggestion
        'when calling an endpoint method with non-existing method name '
```

SandPod (commented):
One question around the recent changes made. I think we can just revert that specific change and the PR is good to go with nice new 404s!

PR #2834, opened 2024-10-09
fix: inheritance for serverOnly scoped classes and fields

BenAuerDev:
This PR tries to adjust the `SerializableModelLibraryGenerator` to handle inherited serverOnly fields correctly. As the parentClass declares them as `Object?` but the child-class parses their real type (`int`, `String` etc.) we now check in the methods building `_buildCopyWithMethod` and `_buildAbstractCopyWithMethod` and pass them to the childs copyWith method as expected

Furthermore a test case was added to `json_protocol_test.dart` making sure `serverOnly` fields are not leaked when inherited. 

Another test was added to `server_code_generator/inheritance_class_test.dart` checking if nullable fields are typed as expected in the `copyWith` method. 

And a `restriction` and `test` was added to only allow extending a `serverOnly` class if the child is marked as `serverOnly` aswell. 

Selection_003

linked to: #1873 and #1618 

## Pre-launch Checklist

## Breaking changes

no breaking changes

### DIFF
--- a/tests/serverpod_test_server/lib/src/endpoints/server_only_scoped_field_model_child.dart
+++ b/tests/serverpod_test_server/lib/src/endpoints/server_only_scoped_field_model_child.dart
@@ -1,0 +1,18 @@
+import 'dart:async';
+
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+
+class ServerOnlyScopedFieldChildModelEndpoint extends Endpoint {
+  Future<ScopeServerOnlyField> getProtocolField(
+    Session session,
+  ) async {
+    return ScopeServerOnlyFieldChild(
+        serverOnlyScope: Types(anInt: 2),
+        nested: ScopeServerOnlyField(
+          allScope: Types(anInt: 1),
+          serverOnlyScope: Types(anInt: 2),
+        ),
+        childFoo: 'childFoo');
+  }
+}
--- a/tests/serverpod_test_server/lib/src/endpoints/server_only_scoped_field_model_child.dart
+++ b/tests/serverpod_test_server/lib/src/endpoints/server_only_scoped_field_model_child.dart
@@ -8,11 +8,12 @@ class ServerOnlyScopedFieldChildModelEndpoint extends Endpoint {
     Session session,
   ) async {
     return ScopeServerOnlyFieldChild(
+      serverOnlyScope: Types(anInt: 2),
+      nested: ScopeServerOnlyField(
+        allScope: Types(anInt: 1),
         serverOnlyScope: Types(anInt: 2),
-        nested: ScopeServerOnlyField(
-          allScope: Types(anInt: 1),
-          serverOnlyScope: Types(anInt: 2),
-        ),
-        childFoo: 'childFoo');
+      ),
+      childFoo: 'childFoo',
+    );
   }
 }
--- a/tests/serverpod_test_server/lib/src/endpoints/server_only_scoped_field_model_child.dart
+++ b/tests/serverpod_test_server/lib/src/endpoints/server_only_scoped_field_model_child.dart
@@ -4,7 +4,7 @@ import 'package:serverpod/serverpod.dart';
 import 'package:serverpod_test_server/src/generated/protocol.dart';
 
 class ServerOnlyScopedFieldChildModelEndpoint extends Endpoint {
-  Future<ScopeServerOnlyField> getProtocolField(
+  Future<ScopeServerOnlyFieldChild> getProtocolField(
     Session session,
   ) async {
     return ScopeServerOnlyFieldChild(
--- a/tests/serverpod_test_server/lib/src/generated/scopes/scope_server_only_field_child.dart
+++ b/tests/serverpod_test_server/lib/src/generated/scopes/scope_server_only_field_child.dart
@@ -1,0 +1,117 @@
+/* AUTOMATICALLY GENERATED CODE DO NOT MODIFY */
+/*   To generate run: "serverpod generate"    */
+
+// ignore_for_file: implementation_imports
+// ignore_for_file: library_private_types_in_public_api
+// ignore_for_file: non_constant_identifier_names
+// ignore_for_file: public_member_api_docs
+// ignore_for_file: type_literal_in_constant_pattern
+// ignore_for_file: use_super_parameters
+
+// ignore_for_file: no_leading_underscores_for_library_prefixes
+import '../protocol.dart' as _i1;
+import 'package:serverpod/serverpod.dart' as _i2;
+
+abstract class ScopeServerOnlyFieldChild extends _i1.ScopeServerOnlyField
+    implements _i2.SerializableModel, _i2.ProtocolSerialization {
+  ScopeServerOnlyFieldChild._({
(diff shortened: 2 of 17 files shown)


tests/serverpod_test_server/lib/src/generated/scopes/scope_server_only_field_child.dart
```diff
@@ -0,0 +1,117 @@
...
+
+  String childFoo;
+
+  @override
+  ScopeServerOnlyFieldChild copyWith({
+    Object? allScope,
+    Object? serverOnlyScope,
+    Object? nested,
+    String? childFoo,
+  });
```
SandPod, line 58: Question: Shouldn't a typed interface be used here? Isn't it only inherited classes that do not have support for type copy with interfaces?

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -308,6 +308,15 @@ class Restrictions {
       ];
     }
 
+    if (parentClass.serverOnly && !documentDefinition!.serverOnly) {
```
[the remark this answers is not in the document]
SandPod: Please add a test for this as well.

tests/serverpod_test_server/lib/src/endpoints/server_only_scoped_field_model_child.dart
```diff
@@ -0,0 +1,18 @@
+import 'dart:async';
+
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+
+class ServerOnlyScopedFieldChildModelEndpoint extends Endpoint {
+  Future<ScopeServerOnlyField> getProtocolField(
```
SandPod, line 7: Suggestion: This endpoint doesn't return `ScopeServerOnlyFieldChild` which is what we want to test.

Please change the return type to catch this specific case.

SandPod (commented):
Nice work!

A couple of suggestions for improvements and a couple of questions around the work.

It seems like we might have broken the typed copy with interface a bit.

tests/serverpod_test_server/lib/src/endpoints/server_only_scoped_field_model_child.dart
```diff
@@ -0,0 +1,18 @@
...
+  Future<ScopeServerOnlyFieldChild> getProtocolField(
+    Session session,
+  ) async {
+    return ScopeServerOnlyFieldChild(
+        serverOnlyScope: Types(anInt: 2),
+        nested: ScopeServerOnlyField(
+          allScope: Types(anInt: 1),
+          serverOnlyScope: Types(anInt: 2),
+        ),
+        childFoo: 'childFoo');
```
SandPod, line 16: Suggestion(NIT): Add comma for better formatting here.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -308,6 +308,21 @@ class Restrictions {
...
+    var currentModel =
+        parsedModels.findByClassName(documentDefinition!.className);
+
+    var ancestorServerOnlyClass =
+        _findServerOnlyClassInParentClasses(currentModel);
+
+    if (!documentDefinition!.serverOnly && ancestorServerOnlyClass != null) {
+      return [
+        SourceSpanSeverityException(
+          'The class "${documentDefinition!.className}" cannot inherit from the server-only class "${ancestorServerOnlyClass.className}" unless it is also marked as "serverOnly".',
```
SandPod, line 320: Suggestion: How about `Cannot extend a "serverOnly" class in the inheritance chain ("${ancestorServerOnlyClass.className}") unless class is marked as "serverOnly".`?

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/inheritance_test.dart
```diff
@@ -213,4 +213,90 @@ void main() {
...
+
+    var error = collector.errors.first;
+    expect(
+      error.message,
+      'The class "ExampleChildClass" cannot inherit from the server-only class "Example" unless it is also marked as "serverOnly".',
+    );
+  });
+
+  test(
+      'Given a serverOnly child-class, When the parent-class not serverOnly but the grandparent-class is, then error is collected that a client class cannot extend a serverOnly class',
```
SandPod, line 257: Suggestion(NIT):
```suggestion
      'Given a serverOnly child-class, When the parent-class is not serverOnly but the grandparent-class is, then error is collected that a client class cannot extend a serverOnly class',
```

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -1138,19 +1139,23 @@ class SerializableModelLibraryGenerator {
...
         subDirParts: classDefinition.subDirParts,
         config: config,
       );
 
+      var isInheritedField = classDefinition.inheritedFields.contains(field);
+
+      var type = field.type.nullable && isInheritedField
+          ? refer('Object?')
+          : fieldType;
+
```
SandPod, line 1154: Suggestion: Could we add some test for this change?

SandPod (commented):
almost the same result but with less changes. I like that! A couple of more comments and suggestions.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
BenAuerDev: Yes I'll add one in `server_code_generator > inheritance_class_test.dart` checking if 
inherited-nullable => `Object?` 
local-nullable => typed (int?, String?, ...)

tools/serverpod_cli/test/generator/dart/server_code_generator/inheritance_class_test.dart
```diff
@@ -348,4 +348,68 @@ void main() {
...
+        );
+
+        test(
+            'with named params where nullable-inherited fields are "Object?" and nullable-local fields are typed',
+            () {
+          expect(
+            copyWithMethod?.parameters?.toSource(),
+            '({int? id, Object? name, int? age})',
+          );
+        }, skip: copyWithMethod == null);
```
SandPod, line 411: Suggestion: Please split these two assertions into two different tests. One that validated the inherited nullable field and one that validatesthe nullable local field.

tools/serverpod_cli/test/generator/dart/server_code_generator/inheritance_class_test.dart
```diff
@@ -348,4 +348,68 @@ void main() {
...
+
+    var codeMap = generator.generateSerializableModelsCode(
+      models: models,
+      config: config,
+    );
+
+    var compilationUnit =
+        parseString(content: codeMap[childExpectedFilePath]!).unit;
+
+    group('Then the $childClassName', () {
```
SandPod, line 392: Suggestion(NIT): change `Then` to `then`.

SandPod (approved):
LGTM with the last modifications that I commented on!

Nice work!

Issue #2393, opened 2024-06-17
Make it possible to log `session.log` output to console.


SandPod:
**Is your feature request related to a problem? Please describe.**
Logging to the console in a cloud hosting environment is crucial because it provides real-time feedback essential for immediate troubleshooting and debugging. Cloud platforms often integrate seamlessly with console logs, enabling real-time monitoring, alerting, and automated response to issues.

**Describe the solution you'd like**
Make it configurable to also always log to the console.

**Additional context**
This is partially already supported when running in development mode: https://github.com/serverpod/serverpod/blob/18d006395a846674e73453e3ad29841157cd2965/packages/serverpod/lib/src/server/session.dart#L199-L203

This implementation needs to be modified in two ways:
1. Logging to standard out should be toggled based on a configuration flag and not a run mode. This should also be configurable through environment variables when the server is started.
2. Multiple logging outputs should be configurable (defaults to current database logging).
3. `stderr` should be used for log messages that have `LogLevel.error`.

lukehutch:
This is now in `log_manager.dart`... I had to make this logging unconditional, because I only have a Linux dev machine, and Serverpod Insights is still not available for Linux, so I need another way to view the server logs easily...

SandPod:
Two updates:

1. We are focusing on improving the logging right now. It's quite tangled so we are working on simplifying it and then adding a separate stoud log writer that is configurable. This should give you good and valid logs for requests directly to the console.

2. We had someone help us validate the Linux build on insights a couple of weeks ago and it seemed to be mostly functional. It is downloadable through this super secret link if you would like to try it out: https://downloads.serverpod.dev/linux/x64/Serverpod-2.0.0.zip

Just swap out `linux/x64` for the version of Serverpod you are running.

lukehutch:
@SandPod Thanks, I tried out Serverpod Insights for Linux.

It works for my development server, but it doesn't correctly interpret the server setting for the production server:

![image](https://github.com/user-attachments/assets/309c360f-bc9f-4d58-80cf-0d4131480cfd)

The server config files use the GCP-internal (private) domains, which are not visible outside that network.

Serverpod Insights can probably predict the correct external domains, based on the internal domains.

PR #2843, opened 2024-10-10
Add Configuration for Controlling `session.log` Output to DB and Console

klkucaj:
### **Description**

This PR introduces the ability to control `session.log` output through both environment variables and configuration files. Logs can now be written to the database, the console, both, or neither, depending on configuration.

### **Key Changes:**
- Added support for two new environment variables:
  - `SERVERPOD_SESSION_PERSISTENT_LOG_ENABLED`
  - `SERVERPOD_SESSION_CONSOLE_LOG_ENABLED`
  
- The `sessionLogs` section in the configuration file now supports:
  - `persistentEnabled`: Enable logging to the database.
  - `consoleEnabled`: Enable logging to the console.

### **Important Note:**
- If `persistentEnabled` is set to `true` but **no database configuration** is provided, a `StateError` will be thrown, as persistent logging requires database support. This ensures that invalid configurations are caught during initialization, preventing silent overrides.

### **How to Configure:**
- **Environment Variables**:
  - Control session logging with:
    - `SERVERPOD_SESSION_PERSISTENT_LOG_ENABLED=true/false`
    - `SERVERPOD_SESSION_CONSOLE_LOG_ENABLED=true/false`
  
- **Configuration File**:
  - Modify the `sessionLogs` section:
    ```yaml
    sessionLogs:
      persistentEnabled: true
      consoleEnabled: true
    ```

Closes: #2393

## Pre-launch Checklist

### DIFF
--- a/packages/serverpod/lib/src/server/log_manager/log_writers.dart
+++ b/packages/serverpod/lib/src/server/log_manager/log_writers.dart
@@ -229,3 +229,45 @@ class StdOutLogWriter extends LogWriter {
     return _logId;
   }
 }
+
+@internal
+class MultipleLogWriter extends LogWriter {
+  final List<LogWriter> _logWriters;
+
+  MultipleLogWriter(this._logWriters);
+
+  @override
+  Future<int> closeLog(SessionLogEntry entry) async {
+    var response =
+        await Future.wait(_logWriters.map((writer) => writer.closeLog(entry)));
+    return response.firstOrNull ?? 0;
+  }
+
+  @override
+  Future<void> logEntry(LogEntry entry) async {
+    await Future.wait(
+      _logWriters.map((writer) => writer.logEntry(entry)),
+    );
+  }
+
+  @override
+  Future<void> logMessage(MessageLogEntry entry) async {
+    await Future.wait(
+      _logWriters.map((writer) => writer.logMessage(entry)),
+    );
+  }
+
+  @override
+  Future<void> logQuery(QueryLogEntry entry) async {
+    await Future.wait(
+      _logWriters.map((writer) => writer.logQuery(entry)),
+    );
+  }
+
+  @override
+  Future<void> openLog(SessionLogEntry entry) async {
+    await Future.wait(
+      _logWriters.map((writer) => writer.openLog(entry)),
+    );
+  }
+}
--- a/packages/serverpod/lib/src/server/log_manager/log_writers.dart
+++ b/packages/serverpod/lib/src/server/log_manager/log_writers.dart
@@ -1,5 +1,6 @@
 import 'dart:io';
 
+import 'package:collection/collection.dart';
 import 'package:meta/meta.dart';
 import 'package:serverpod/protocol.dart';
 import 'package:serverpod/serverpod.dart';
@@ -236,11 +237,33 @@ class MultipleLogWriter extends LogWriter {
 
   MultipleLogWriter(this._logWriters);
 
+  //// Closes logs for all writers and returns the log ID.
+  ///
+  /// The log ID for persistent logging is prioritized if present.
+  /// This is because persistent logs are more critical for long-term record-keeping
+  /// compared to non-persistent logs, such as console logs.
+  /// If no persistent log is available, the first log ID from any writer is returned.
   @override
   Future<int> closeLog(SessionLogEntry entry) async {
-    var response =
-        await Future.wait(_logWriters.map((writer) => writer.closeLog(entry)));
-    return response.firstOrNull ?? 0;
+    int? databaseLogId;
+
+    // Close logs for all writers and collect their log IDs
+    var responses = await Future.wait(_logWriters.map((writer) async {
+      // Each writer closes the log and returns a logId
+      var logId = await writer.closeLog(entry);
+
+      // If this writer is a DatabaseLogWriter, prioritize its logId
+      if (writer is DatabaseLogWriter) {
+        databaseLogId = logId;
+      }
+
+      // Return the logId for this writer to include in the results
+      return logId;
+    }));
+
+    // Return the DatabaseLogWriter's logId if available,
+    // otherwise use the first logId from any writer
+    return databaseLogId ?? responses.firstOrNull ?? 0;
   }
 
   @override
--- a/packages/serverpod/lib/src/server/log_manager/log_writers.dart
+++ b/packages/serverpod/lib/src/server/log_manager/log_writers.dart
@@ -247,22 +247,16 @@ class MultipleLogWriter extends LogWriter {
   Future<int> closeLog(SessionLogEntry entry) async {
     int? databaseLogId;
 
-    // Close logs for all writers and collect their log IDs
     var responses = await Future.wait(_logWriters.map((writer) async {
-      // Each writer closes the log and returns a logId
       var logId = await writer.closeLog(entry);
 
-      // If this writer is a DatabaseLogWriter, prioritize its logId
       if (writer is DatabaseLogWriter) {
         databaseLogId = logId;
       }
 
-      // Return the logId for this writer to include in the results
       return logId;
     }));
 
-    // Return the DatabaseLogWriter's logId if available,
-    // otherwise use the first logId from any writer
     return databaseLogId ?? responses.firstOrNull ?? 0;
   }
 
--- a/packages/serverpod_shared/lib/src/config.dart
+++ b/packages/serverpod_shared/lib/src/config.dart
@@ -38,6 +38,9 @@ class ServerpodConfig {
   /// Authentication key for service protocol.
   late final String? serviceSecret;
 
+  /// Configuration for Session logs.
+  final SessionLogConfig? sessionLogs;
+
   /// Creates a new [ServerpodConfig].
   ServerpodConfig({
     required this.apiServer,
@@ -49,6 +52,7 @@ class ServerpodConfig {
     this.database,
     this.redis,
     this.serviceSecret,
+    this.sessionLogs,
   }) {
     apiServer._name = 'api';
     insightsServer?._name = 'insights';
@@ -126,6 +130,15 @@ class ServerpodConfig {
           )
         : null;
 
+    var sessionLogsConfig = _sessionLogsConfigMap(configMap, environment);
+    var losessionLogs = sessionLogsConfig != null
+        ? SessionLogConfig._fromJson(
+            sessionLogsConfig,
+            ServerpodConfigMap.sessionLogs,
+            databaseEnabled: database != null,
+          )
+        : null;
+
     return ServerpodConfig(
       runMode: runMode,
       serverId: serverId,
@@ -136,6 +149,7 @@ class ServerpodConfig {
       database: database,
       redis: redis,
       serviceSecret: serviceSecret,
+      sessionLogs: losessionLogs,
     );
   }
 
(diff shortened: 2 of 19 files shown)


packages/serverpod_shared/lib/src/config.dart
```diff
@@ -386,6 +401,52 @@ class RedisConfig {
...
+
+  /// Creates a new [SessionLogConfig].
+  SessionLogConfig({
+    required this.persistentEnabled,
+    required this.consoleEnabled,
+  });
+
+  factory SessionLogConfig._fromJson(
+    Map logsSetup,
+    String name, {
```
hampuslavin, line 420: Question: Am I missing something or is this just unused?

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -386,6 +401,52 @@ class RedisConfig {
...
+  final bool consoleEnabled;
+
+  /// Creates a new [SessionLogConfig].
+  SessionLogConfig({
+    required this.persistentEnabled,
+    required this.consoleEnabled,
+  });
+
+  factory SessionLogConfig._fromJson(
+    Map logsSetup,
```
hampuslavin, line 419: Suggestion: should we maybe just call this `sessionLogConfigJson` to be consistent with naming?

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -449,6 +510,15 @@ Map? _redisConfigMap(Map configMap, Map<String, String> environment) {
   ]);
 }
 
+Map? _sessionLogsConfigMap(Map configMap, Map<String, String> environment) {
```
hampuslavin, line 513: Suggestion: Add `build` to the beginning to be consistent
```suggestion
Map? _buildSessionLogsConfigMap(Map configMap, Map<String, String> environment) {
```

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -126,6 +130,15 @@ class ServerpodConfig {
           )
         : null;
 
+    var sessionLogsConfig = _sessionLogsConfigMap(configMap, environment);
+    var losessionLogs = sessionLogsConfig != null
```
hampuslavin, line 134: Suggestion: Rename the first variable to `sessionLogsConfigJson` and the second one to `sessionLogsConfig`

packages/serverpod/lib/src/server/log_manager/log_writers.dart
```diff
@@ -229,3 +229,45 @@ class StdOutLogWriter extends LogWriter {
...
 }
+
+@internal
+class MultipleLogWriter extends LogWriter {
+  final List<LogWriter> _logWriters;
+
+  MultipleLogWriter(this._logWriters);
+
+  @override
+  Future<int> closeLog(SessionLogEntry entry) async {
```
hampuslavin, line 240: Question: What are we using the log id for? Is it a problem that we only return the id of the first log?

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -386,6 +401,52 @@ class RedisConfig {
...
+    Map logsSetup,
+    String name, {
+    required bool databaseEnabled,
+  }) {
+    var persistent =
+        logsSetup[ServerpodEnv.sessionPersistentLogEnabled.configKey] ?? false;
+
+    if (persistent && !databaseEnabled) {
+      stdout.writeln(
+        'Warning: Persistent session logging is enabled in the configuration, but this project was created without database support. '
```
hampuslavin, line 428: NIT/Suggestion: Maybe it could be helpful to be super explicit about what key is causing this warning:
```suggestion
        'Warning: The `persistentEnabled` setting was is enabled in the configuration, but this project was created without database support. '
```

packages/serverpod_shared/test/session_logs_config_test.dart
```diff
@@ -0,0 +1,186 @@
...
+        loadYaml(serverpodConfig),
+      );
+
+      expect(config.sessionLogs?.persistentEnabled, isFalse);
+      expect(config.sessionLogs?.consoleEnabled, isTrue);
+    },
+  );
+
+  test(
+    'Given a Serverpod config with sessionLogs and database when environment variables override them, sessionLogs config reflects the environment overrides.',
```
hampuslavin, line 152: Suggestion: use "then" instead of comma, applies to whole file
```suggestion
    'Given a Serverpod config with sessionLogs and database when environment variables override them then sessionLogs config reflects the environment overrides.',
```

hampuslavin (changes requested):
Nice work! 👍 Left a bunch of minor suggestions and questions

packages/serverpod_shared/lib/src/config.dart
klkucaj: You are right, thanks!

packages/serverpod_shared/lib/src/config.dart
klkucaj: Yes, done, thank u!

packages/serverpod_shared/lib/src/config.dart
klkucaj: @hampuslavin if i do this for this one then we should do for the others, what do u think ?
`_databaseConfigMap`
`_redisConfigMap`
ect ect

packages/serverpod/lib/src/server/log_manager/log_writers.dart
klkucaj: @hampuslavin Since we want to minimize changes, @Isakdl and I agreed to keep it this way. In the current implementation, the database log writer has priority because, when persistent logging is enabled, the database writer is always added first to the list of log writers. This ensures that the database log ID is prioritized. You can see this behavior in the following code snippet:

https://github.com/serverpod/serverpod/blob/c65546bbf0627b06effd0cbd0db2e1227f388427/packages/serverpod/lib/src/server/session.dart#L177-L189

hampuslavin: Ah I see, I wonder if it would make sense to do a check for `is DatabaseLogWriter` and chose the log id based on that. Then it will be agnostic to the order the loggers are added and would be more robust. It would also show the intent clearer in the `closeLog` method. We could even add a comment in the `closeLog` method for why the `DatabaseLogWriter` id is chosen.

packages/serverpod/lib/src/server/log_manager/log_writers.dart
```diff
@@ -236,11 +237,33 @@ class MultipleLogWriter extends LogWriter {
...
+      if (writer is DatabaseLogWriter) {
+        databaseLogId = logId;
+      }
+
+      // Return the logId for this writer to include in the results
+      return logId;
+    }));
+
+    // Return the DatabaseLogWriter's logId if available,
+    // otherwise use the first logId from any writer
```
hampuslavin, line 265: Nit: remove comment, it basically says the same as the code

packages/serverpod/lib/src/server/log_manager/log_writers.dart
```diff
@@ -236,11 +237,33 @@ class MultipleLogWriter extends LogWriter {
...
   @override
   Future<int> closeLog(SessionLogEntry entry) async {
-    var response =
-        await Future.wait(_logWriters.map((writer) => writer.closeLog(entry)));
-    return response.firstOrNull ?? 0;
+    int? databaseLogId;
+
+    // Close logs for all writers and collect their log IDs
+    var responses = await Future.wait(_logWriters.map((writer) async {
+      // Each writer closes the log and returns a logId
```
hampuslavin, line 252: Nit: remove comment, it basically says the same as the code

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -386,6 +402,61 @@ class RedisConfig {
...
+
+    var isSessionPersistentLogEnabled = sessionLogConfigJson[
+            ServerpodEnv.sessionPersistentLogEnabled.configKey] ??
+        false;
+
+    if (isSessionPersistentLogEnabled && !databaseEnabled) {
+      stdout.writeln(
+        'Warning: The `persistentEnabled` setting was enabled in the configuration, but this project was created without database support. '
+        'Persistent logging is only available when the database is enabled, so the value will be overridden and disabled.',
+      );
```
hampuslavin, line 441: Suggestion: when looking at this again, I think we should throw a `StateError` here since it's an invalid state. We should also add a test for it.

hampuslavin (commented):
Looks good! One more thing that I saw now 👁️

packages/serverpod_shared/lib/src/config.dart
klkucaj: Done. Additionally, the validation logic has now been moved to a dedicated `_validate` method, which you can view here:

- https://github.com/serverpod/serverpod/blob/b1e86124ce4a4122d9e099f33c124bd42a640035/packages/serverpod_shared/lib/src/config.dart#L447-L456

This method is called after initialization here:

- https://github.com/serverpod/serverpod/blob/b1e86124ce4a4122d9e099f33c124bd42a640035/packages/serverpod_shared/lib/src/config.dart#L60-L62

The reason for this change is to ensure that the validation is enforced regardless of how the `SessionLogConfig` is initialized. Initially, the validation was embedded in the `_fromJson` factory method, but if someone directly instantiated `SessionLogConfig` outside of this factory, the validation would be skipped.

By moving the validation to a separate `_validate` method, it can be called every time the configuration is initialized, ensuring that the check for persistent logging (i.e., it can only be enabled if the database is available) is always performed, no matter how or when the config is created. This guarantees that the configuration remains consistent and avoids any invalid states during initialization.

Isakdl (approved):
Very nice PR description! Good job!

LGTM! But please also add a PR in the docs repo for this feature.

hampuslavin (approved):
👍

tests/serverpod_test_server/test_integration/logging/session_logging_test.dart
```diff
@@ -214,67 +215,39 @@ void main() async {
...
-            publicScheme: 'http',
-          ),
-          database: DatabaseConfig(
-            host: 'postgres',
-            port: 5432,
-            user: 'postgres',
-            password: 'password',
-            name: 'serverpod_test',
+    test(
+        'when the server is started, then a StateError should be thrown due to lack of database support',
```
hampuslavin: @klkucaj I think this group can be merged into the test below now, right?

tests/serverpod_test_server/test_integration/logging/session_logging_test.dart
```diff
@@ -0,0 +1,318 @@
...
+      var logs = await LoggingUtil.findAllLogs(session);
+      expect(logs, isEmpty);
+      expect(record.output.contains('Test message'), isFalse);
+    });
+  });
+
+  group(
+      'Given persistent logging is enabled but the database support is not available',
+      () {
+    test('a StateError should be thrown due to lack of database support',
```
hampuslavin, line 218: Suggestion: I think these two can be merged into a single `test`.

klkucaj: @hampuslavin done:

https://github.com/serverpod/serverpod/blob/cea16cfc2c0a9b38c93dfed2b27836a04c24bd53/tests/serverpod_test_server/test_integration/logging/session_logging_test.dart#L215-L244

hampuslavin (approved):
🚀 🎸

PR #2846, opened 2024-10-10
fix: Correctly encode parameters passed to test tools.

### LINKED ISSUE #2845: Serialize models before passing to EndpointDispatch in generated test tools
**Describe the bug**
The following test throws an exception because the parameters are not serialized before being sent into `EndpointDispatch`:
```dart
test('when calling echoSimpleData then should echo the object', () async {
        final data = SimpleData(num: 42);
        var result =
            await endpoints.testTools.echoSimpleData(sessionBuilder, data);
        expect(result.num, 42);
      });
```

To solve this, parameters should pass through `SerializationManager.encodeForProtocol(parameter)` to ensure it's serialized first.

hampuslavin:
## Description
- Fixes #2845 

## Changes
- Adds a test serializer helper `testObjectToJson` that uses the same code paths as production code

## Question to reviewer ❓ 
- It's inefficient to go to String and then to JSON, but unless I missed something we didn't have any helpers that takes an object and returns JSON? I thought this performance hit was okay because a) it's a test and b) it's  great to know that we're taking the same code path as prod.

## Pre-launch Checklist

### DIFF
--- a/tools/serverpod_cli/lib/src/generator/dart/library_generators/server_test_tools_generator.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/library_generators/server_test_tools_generator.dart
@@ -171,12 +171,12 @@ class ServerTestToolsGenerator {
                   ..body = refer('_localUniqueSession').code).closure,
                 'endpointPath': literalString(endpoint.name),
                 'methodName': literalString(method.name),
-                'parameters': literalMap({
-                  for (var parameter in method.allParameters)
-                    literalString(parameter.name):
-                        refer('testObjectToJson', serverpodTestUrl)
-                            .call([refer(parameter.name)]).code,
-                }),
+                'parameters': refer('testObjectToJson', serverpodTestUrl).call([
+                  literalMap({
+                    for (var parameter in method.allParameters)
+                      literalString(parameter.name): refer(parameter.name).code,
+                  })
+                ]),
                 'serializationManager': refer('_serializationManager'),
               }))
               .statement,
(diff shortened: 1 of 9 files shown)


tools/serverpod_cli/lib/src/generator/dart/library_generators/server_test_tools_generator.dart
```diff
@@ -173,7 +173,9 @@ class ServerTestToolsGenerator {
                 'methodName': literalString(method.name),
                 'parameters': literalMap({
                   for (var parameter in method.allParameters)
-                    literalString(parameter.name): refer(parameter.name).code,
+                    literalString(parameter.name):
+                        refer('testObjectToJson', serverpodTestUrl)
+                            .call([refer(parameter.name)]).code,
```
SandPod, line 178: Question: Why can't we just encode the parameters the same way as we do in the client?

https://github.com/serverpod/serverpod/blob/3b1594b732fd8c49a0fe3d6ccfaa6db1271ae6bf/packages/serverpod_client/lib/src/serverpod_client_shared_private.dart#L8-L15

SandPod: This might not be a problem but I suspect this should actually be `.encodeForProtocol` right?

SandPod (commented):
Single question around this.

tools/serverpod_cli/lib/src/generator/dart/library_generators/server_test_tools_generator.dart
hampuslavin: > This might not be a problem but I suspect this should actually be `.encodeForProtocol` right?

I guess so? 🧐

SandPod: OK! Makes sense, can we create a general method that handles all parameters instead of handling them specifically?

hampuslavin: @SandPod Fixed: [3433cc8](https://github.com/serverpod/serverpod/pull/2846/commits/3433cc81c5a5fd11d1c9c0451c6a0c915ba6a9c9) 👍

SandPod (approved):
LGTM! 🚀

PR #2849, opened 2024-10-11
fix: Rename `databaseAccessor` arg to `session` to clarify intent.

hampuslavin:
## Description
- When the `DatabaseAccessor` interface was added, all the arguments where also renamed to `databaseAccessor`. This was a mistake, however, since the intent is still to only pass `session` objects. 
- To clarify this intent to users of Serverpod, this PR renames the argument to `session` again.

## Changes
- Renames `databaseAccessor` arg to `session` to clarify intent

## Pre-launch Checklist


SandPod (approved):
LGTM! 👍 

Would have been nice if the interface was "session" typed but I think this is OK!

Issue #2850, opened 2024-10-11
Test database proxy is not handing database exceptions in the same way as a real transaction


hampuslavin:
**Describe the bug**
When the database throws exceptions during a transactions, it will close the transaction. See first example:

```dart
var transactionFuture = session.db.transaction((tx) async {
  await UniqueData.db.insertRow(session, data, transaction: tx);

  try {
    await UniqueData.db.insertRow(session, duplicatedData, transaction: tx);
  } catch (_) {
    // Ignore duplicate key exception from DB
  }

 // This will throw with "transaction is closed"
  await SimpleData.db.insertRow(
    session,
    SimpleData(num: 1),
    transaction: tx,
  );
});
```

However, if the the transaction is allowed to complete, it will re-throw the error:

```dart
var transactionFuture = session.db.transaction((tx) async {
  await UniqueData.db.insertRow(session, data, transaction: tx);

  try {
    await UniqueData.db.insertRow(session, duplicatedData, transaction: tx);
  } catch (_) {
    // Ignore duplicate key exception from DB
  }
});

await expectLater(transactionFuture, throwsA(isA<DatabaseException>()));
```

This is reasonable considering that PostgreSQL considers the transaction to be erroneous. 
As a side note, the try-catch in the transaction should therefore never be used to ignore errors, but only to re-throw a domain specific exception:

```dart
try {
  await UniqueData.db.insertRow(session, duplicatedData, transaction: tx);
} catch (_) {
  throw MyDuplicateEntryException();
}
```

The problem is that the test framework currently cannot emulate this behavior, since the test database proxy is not aware of wether a database exception was thrown or not (or wether the transaction is aborted):

```dart
// test_database_proxy.dart:151
try {
  var result = await transactionFunction(localTransaction);
  await _transactionManager.removePreviousSavePoint(unlock: true);
  // This would be needed to emulate the described behavior:
  // if database error occurred
  // then throw database error
  return result;
} catch (e) {
  await _transactionManager.rollbackToPreviousSavePoint(unlock: true);
  rethrow;
}
```
There is no easy way to know this, so it's unclear how this would be implemented.

**Suggested solution**

I would suggest to document this edge case and consider it out of scope for the moment.

hampuslavin:
Closed by https://github.com/serverpod/serverpod/pull/2851 and https://github.com/serverpod/serverpod_docs/pull/184

PR #2851, opened 2024-10-11
fix: Catch database exceptions at the end of transactions.

hampuslavin:
## Description
As explained in detail in this issue: #2850 , database exceptions are re-thrown if a transaction is allowed to complete when a database exception has occurred. This PR ensures those errors are caught when the top level test transaction completes.

We should also create a small caveat in the test docs about this. Currently the following code will throw in production code, but will complete successfully in test code:
```dart
session.db.transaction((tx) async {
  var data = UniqueData(number: 1, email: 'test@test.com');
  try {
    await UniqueData.db.insertRow(session, data);
    await UniqueData.db.insertRow(session, data);
  } catch (_) {
    // Ignore duplicate key exception from DB
  }
});
```
While this is a theoretical inconsistency, it should very rarely (if ever) be a problem because this type of code does not make sense and should be considered an anti-pattern. The transaction is aborted when errors happen, so no more operations can be performed anyway.

## Changes
- Adds `catchError` to the top level test transaction [4abee2ac7531924c80acaf6df151a674c646a4ae]
- Adds a test config for `serverpod_test_server` to avoid having to pass a production run mode to the `withServerpod` helper [a05fe24]

## Pre-launch Checklist

### DIFF
--- a/tests/serverpod_test_server/test_integration/test_tools/transaction_correctness_test.dart
+++ b/tests/serverpod_test_server/test_integration/test_tools/transaction_correctness_test.dart
@@ -25,7 +25,7 @@ void main() {
 
       test(
           'when next test is run '
-          'then should still work', () async {
+          'then database operations should still work', () async {
         await SimpleData.db.insertRow(session, SimpleData(num: 1));
 
         expect(await SimpleData.db.find(session), hasLength(1));
--- a/tests/serverpod_test_server/test_integration/test_tools/transaction_correctness_test.dart
+++ b/tests/serverpod_test_server/test_integration/test_tools/transaction_correctness_test.dart
@@ -19,6 +19,11 @@ void main() {
           await UniqueData.db.insertRow(session, data);
         });
 
+        // Even though this exception is caught in this test, due to how transactions work
+        // the exception will be re-thrown when the top level test transaction is canceled.
+        // If the top level transaction error is not caught this test will fail.
+        // Therefore, this test validates that the exception is caught on the top level
+        // and does not fail the dart test runner.
         await expectLater(future, throwsA(isA<DatabaseException>()));
       });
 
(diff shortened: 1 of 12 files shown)


tests/serverpod_test_server/test_integration/test_tools/transaction_correctness_test.dart
```diff
@@ -0,0 +1,35 @@
...
+
+void main() {
+  withServerpod(
+    'Given transaction call in test',
+    (sessionBuilder, endpoints) {
+      var session = sessionBuilder.build();
+
+      test(
+          'when database exception occurs '
+          'then should close transaction after test suite and should not fail `dart test` by leaking exceptions',
```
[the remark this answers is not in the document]
hampuslavin: Good question! 
Canceling the transaction is already verified in a bunch of other test suites, but for explicitness I could add a separate case for that when something throws (by checking that the next `withServerpod` is rolled back) This test should probably be rephrased to: `then should not fail `dart test` by leaking exceptions`

hampuslavin:
@SandPod Added docs here: https://github.com/serverpod/serverpod_docs/pull/184

tests/serverpod_test_server/test_integration/test_tools/transaction_correctness_test.dart
```diff
@@ -0,0 +1,84 @@
...
+
+import 'serverpod_test_tools.dart';
+
+void main() {
+  withServerpod(
+    'Given transaction call in test',
+    (sessionBuilder, endpoints) {
+      var session = sessionBuilder.build();
+
+      test(
```
SandPod, line 13: Suggestion: Please add a comment that explains how this test would fail if we had not applied our current fix.

since this is a nitty implementation detail I think it makes sense to help future developers out a bit.

SandPod (approved):
LGTM given that the comment is added! 🚀

tests/serverpod_test_server/test_integration/test_tools/transaction_correctness_test.dart
hampuslavin: b4fdc1a

PR #2862, opened 2024-10-14
fix: Change repository class arg type back to `Session`

hampuslavin:
## Description
To not degrade the DevEx of using the repository classes, the `Session` type should be used as the argument type to make it intuitive what to pass.

## Changes
- Reverts the change to `DatabaseAccessor` type and uses `Session` as it used to be

## Pre-launch Checklist
- [x] Any breaking changes are documented below.nges, make sure to outline them here, so that they can be included in the notes for the next release._

### DIFF
--- a/examples/chat/chat_server/lib/src/generated/channel.dart
+++ b/examples/chat/chat_server/lib/src/generated/channel.dart
@@ -8,6 +8,8 @@
 // ignore_for_file: type_literal_in_constant_pattern
 // ignore_for_file: use_super_parameters
 
+// ignore_for_file: invalid_use_of_visible_for_testing_member
+
 // ignore_for_file: no_leading_underscores_for_library_prefixes
 import 'package:serverpod/serverpod.dart' as _i1;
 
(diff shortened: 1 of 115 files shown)


examples/chat/chat_server/lib/src/generated/channel.dart
```diff
@@ -8,6 +8,8 @@
 // ignore_for_file: type_literal_in_constant_pattern
 // ignore_for_file: use_super_parameters
 
+// ignore_for_file: invalid_use_of_visible_for_testing_member
```
SandPod, line 11: Question: Are we sure we want to nuke this for the whole file? Would it be helpful for us to note if we add dependencies toward testing members in the generated code?

SandPod (commented):
Single question for discussion.

examples/chat/chat_server/lib/src/generated/channel.dart
hampuslavin: I was mainly thinking we are using file-level scopes for all other linting rules. The same reasoning could apply to all of those: maybe we actually did a mistake somewhere. But in the end I chose this for consistency reasons.

PR #2878, opened 2024-10-17
fix: Support unsafe queries in test tools.

### LINKED ISSUE #2874: Default to session.transaction in `database.dart` methods
**Describe the bug**
The database methods in `src/database/database.dart` should use the same transaction fallbacks as the repository classes. Right now the transaction is not applied to these database calls. The unsafe queries currently does not have any alternative and the test level transaction is not including them right now.

hampuslavin:
## Description
The way we set up the transaction fallback to support the tests was actually overly complicated. It is better and more robust to only do this fallback at the source of truth. This also fixes the issue that unsafe queries was not included in the test transaction.

Closes #2874 

## Changes
- Moves transaction fallback to the database class instead

## Pre-launch Checklist

### DIFF
--- a/examples/chat/chat_server/lib/src/generated/channel.dart
+++ b/examples/chat/chat_server/lib/src/generated/channel.dart
@@ -8,8 +8,6 @@
 // ignore_for_file: type_literal_in_constant_pattern
 // ignore_for_file: use_super_parameters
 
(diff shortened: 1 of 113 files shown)


SandPod (approved):
LGTM!

Double checked all the session passing in `database.dart` and it looks good! 👍

PR #2888, opened 2024-10-21
feat: Sealed Classes

BenAuerDev:
This PR allows users to use the 'sealed' keyword in their yaml files to generate sealed classes/hierarchies.

This was possible thanks to two custom allocators `PartAllocator` and `PartOfAllocator` that @Isakdl created in [dart/code_builder/issue#442](https://github.com/dart-lang/code_builder/issues/442)

Changes:
1. `sealed` keyword was added, with parsing method, node-validation and test checking that `sealed` and `table` are mutually exclusive
2. (temporary) restriction was added disallowing a sealed hierarchy to be defined when not all models are located in the same sub-directory + test
3. - `ModelLibraryGenerator` now also generates sealed hierarchy files like intended
    - `GenerateCode` now accepts `Allocators` (nullable) as parameter and passes appropriate allocator to `DartEmitter`
    - client-/ and server_code_generator: generate sealed hierarchies separate from the rest of the modesl and create a custom-allocator-pair per hierarchy.
    - `LibraryGenerator` exports children of sealed hieararchies through their parent and does not expect sealed classes to have a `fromJson` method
    - `CompliationUnitHelpers` now has methods to find `part`, `part of`, `export` directives and also `methodAnnotations`
4. tests were added to check the `protocol` and `sealed classes` being generated correctly 

linked to: #1873 and #1618

## Pre-launch Checklist

## Breaking changes


tools/serverpod_cli/lib/src/generator/dart/library_generators/util/model_filter_util.dart
```diff
@@ -0,0 +1,56 @@
+import 'package:path/path.dart' as p;
+
+import 'package:serverpod_cli/src/analyzer/models/definitions.dart';
+import 'package:serverpod_cli/src/config/config.dart';
+
+List<ClassDefinition> getSealedClasses(
```
SandPod, line 6: Suggestion: Rename to `getSealedTopNodeClasses`.

Can this method also be private?

tools/serverpod_cli/lib/src/generator/dart/library_generators/util/model_filter_util.dart
```diff
@@ -0,0 +1,56 @@
...
+  List<SerializableModelDefinition> models,
+) {
+  return models
+      .whereType<ClassDefinition>()
+      .where((element) => element.sealedTopNode == element)
+      .toList();
+}
+
+List<List<ClassDefinition>> getSealedHierarchies(
```
SandPod, line 15: Suggestion: Add an `abstract final class` around these helper methods to capture the intention of them.

tools/serverpod_cli/lib/src/generator/dart/library_generators/util/model_filter_util.dart
```diff
@@ -0,0 +1,56 @@
+import 'package:path/path.dart' as p;
+
+import 'package:serverpod_cli/src/analyzer/models/definitions.dart';
+import 'package:serverpod_cli/src/config/config.dart';
+
```
SandPod, line 5: Suggestion: For this file: Change the return type to Iterator where possible.

This allows users of the methods to combine them and still iterate the lists efficiently.

tools/serverpod_cli/lib/src/generator/dart/library_generators/util/model_filter_util.dart
```diff
@@ -0,0 +1,56 @@
...
+      return [...element.descendantClasses, element];
+    },
+  ).toList();
+}
+
+List<ClassDefinition> getSealedHierarchyClasses(
+  List<SerializableModelDefinition> models,
+) {
+  return getSealedHierarchies(models).expand((e) => e).toList();
+}
```
SandPod, line 31: Question: Why is this step needed? Isn't the return of `getSealedHierarchies` good enough?

tools/serverpod_cli/lib/src/generator/dart/library_generators/util/model_filter_util.dart
```diff
@@ -0,0 +1,56 @@
...
+
+List<SerializableModelDefinition> getClassesWithoutSealedHierarchies(
+  List<SerializableModelDefinition> models,
+) {
+  return models
+      .where((e) => !getSealedHierarchyClasses(models).contains(e))
```
SandPod, line 37: Suggestion: Change this to only loop the list 1 on 2 times.

Since `getSealedHierachyClasses` loops the full list each time it is invoked, this will loop the list for every model in `models`.
Figure out what are the sealed classes first and then check if they contain the entry.

tools/serverpod_cli/lib/src/generator/dart/client_code_generator.dart
```diff
@@ -19,16 +21,36 @@ class DartClientCodeGenerator extends CodeGenerator {
...
-            ...model.subDirParts,
-            '${model.fileName}.dart',
-          ]): clientSideGenerator.generateModelLibrary(model).generateCode(),
-    };
+    var sealedHierarchies = getSealedHierarchies(models);
+
+    var modelsWithoutSealedHierarchies =
+        getClassesWithoutSealedHierarchies(models);
+
+    for (var sealedHierarchy in sealedHierarchies) {
```
SandPod, line 31: Discussion: Radical thought, but instead of actually generating the code here. How about you simple create a list that contains the model and a nullable allocator. Then we can loop through that list at the of this method to generate the code and not have to repeat the `createFilePath` stuff?

SandPod: I think this could help simplify the code a bit here.

SandPod: Same should be applicable on the `server_code_generator`.

Maybe we then don't even need to break out the `createFilePath` method?

tools/serverpod_cli/test/generator/dart/server_code_generator/inheritance_protocol_test.dart
```diff
@@ -0,0 +1,290 @@
...
+
+void main() {
+  var expectedFileName = path.join(
+    'lib',
+    'src',
+    'generated',
+    'protocol.dart',
+  );
+
+  group('Given a sealed hierarchy when generating protocol files', () {
```
SandPod, line 25: Suggestion: Please describe the hierarchy in the description of the test.

tools/serverpod_cli/test/generator/dart/server_code_generator/inheritance_protocol_test.dart
```diff
@@ -0,0 +1,290 @@
...
+    var parentClassName = 'Example';
```
SandPod, line 26: Question: How does this differentiate from the validation we are doing in the `simple` case in `sealed_class_test.dart`?

SandPod (changes requested):
Nice work getting the sealed stuff working. This was more complicated than it seemed at first sight :)

A general though around this PR is that we can be more effective when iterating or investigating structures. Think about how many times you might visit each note in some hierarchy and if this is required.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -324,6 +324,19 @@ class Restrictions {
           )
         ];
       }
+
+      // This is temporary until [PartAllocator] and [PartOfAllocator] can automatically allocate the imports from the sub-classes relative to the parent imports.
```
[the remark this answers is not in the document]
BenAuerDev: Of course :rocket: #2893

tests/serverpod_test_server/lib/src/generated/protocol.dart
```diff
@@ -6582,214 +6596,223 @@ class Protocol extends _i1.SerializationManagerServer {
...
       return 'ParentClass';
     }
-    if (data is _i48.CityWithLongTableName) {
+    if (data is _i48.SealedChild) {
+      return 'SealedChild';
+    }
+    if (data is _i48.SealedGrandchild) {
+      return 'SealedGrandchild';
+    }
+    if (data is _i48.SealedParent) {
```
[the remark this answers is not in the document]
BenAuerDev: no don't think so :)

tools/serverpod_cli/test/generator/dart/server_code_generator/inheritance_protocol_test.dart
BenAuerDev: This doesn't check the classes like in `sealed_class_test.dart` it tests the protocol file, to make sure we are not exporting sub-classes of a sealed hierarchy by themselves. 

And that sealed classes are not serialized

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -324,6 +324,19 @@ class Restrictions {
...
+      if (currentModel.sealedTopNode != null) {
+        if (currentModel.subDirParts.join() !=
+            currentModel.sealedTopNode!.subDirParts.join()) {
+          return [
+            SourceSpanSeverityException(
+              'All models in a sealed library must be in the same subdirectory. The class "${currentModel.className}" needs to be located in the same subdirectory as "${currentModel.sealedTopNode!.className}".',
```
SandPod, line 334: Suggestion: Please remove the dangerous deref `!`.

SandPod (commented):
Getting closer!

Still a couple of things to fix. Especially the dangerous derefs that are scattered around.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -115,7 +121,59 @@ class ClassDefinition extends SerializableModelDefinition {
...
+    }
+
+    return parent.everyParentIsSealed;
+  }
+
+  /// Returns a list of all descendant classes.
+  /// This includes all child classes and their descendants.
+  /// If the class has no child classes, an empty list is returned.
+  List<ClassDefinition> get descendantClasses {
+    return _descendantClasses ??= _computeDescendantClasses();
```
SandPod, line 161: Comment: Neat! 🌟

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -49,6 +49,20 @@ class SerializableModelLibraryGenerator {
...
+          for (var child in classDefinition.descendantClasses) {
+            libraryBuilder.directives
+                .add(Directive.part('${child.fileName}.dart'));
+          }
+        }
+
+        if (!classDefinition.isSealedTopNode &&
+            classDefinition.sealedTopNode != null) {
+          libraryBuilder.directives.add(
+            Directive.partOf('${classDefinition.sealedTopNode?.fileName}.dart'),
```
SandPod, line 62: Suggestion: Assign sealedTopNode to a variable on the outside and nullcheck it.

This way we remove unecessary case where this value COULD return something different the second time we check it.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -49,6 +49,20 @@ class SerializableModelLibraryGenerator {
...
+          for (var child in classDefinition.descendantClasses) {
+            libraryBuilder.directives
+                .add(Directive.part('${child.fileName}.dart'));
+          }
+        }
+
+        if (classDefinition.sealedTopNode != null &&
+            classDefinition.sealedTopNode != classDefinition) {
+          libraryBuilder.directives.add(
+            Directive.partOf('${classDefinition.sealedTopNode!.fileName}.dart'),
```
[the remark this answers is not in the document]
SandPod: Question: Why did you choose not to ask `!classDefinition.isSealedTopNode`?

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
SandPod: Suggestion: Please validate the value exists before using it.

Assign `currentModel.sealedTopNode` to a variable outside and nullcheck that variable instead of fetching it with the getter multiple times.

SandPod (commented):
Looks like this could be the last round.

There is a merge conflict with main now as well. Please resolve that with a main merge.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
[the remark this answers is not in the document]
BenAuerDev: I did :) 

```dart
   if (!classDefinition.isSealedTopNode && sealedTopNode != null) {
          libraryBuilder.directives.add(
            Directive.partOf('${sealedTopNode.fileName}.dart'),
          );
        }
```

but we also have to check if `sealedTopNode != null` otherwise every class would have a part-of directive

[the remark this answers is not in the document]
SandPod: My bad, this was already fixed :)

SandPod (approved):
Awesome work Ben! 🎉

Thank you for all the great work! This is a great testament to your skill as a software dev :)

Thank you 🙏

PR #2904, opened 2024-10-24
refactor: Build Serverpod lazily to enable better concurrency support.

### LINKED ISSUE #2894: Lazily create the Serverpod Instance in `withServerpod`
Since the Serverpod constructor tries to connect to the database, and an instance is created for each call to `withServerpod`, this cause the max connections limit to be hit on the postgres database very quickly for larger projects.

**Suggested enhancement**
Create Serverpod lazily. It should still be possible to build a `Session` outside of any `setUp`/`setUpAll` and that will trigger the instance creation. 

We could also consider making `buildSession` async to force usage inside either set up functions or tests themselves.

hampuslavin:
## Description
Closes #2894.

Tries to better support 

## Question/Discussion ❓ 
The other option here would be to actually make `sessionBuilder.build` async (even though it's not needed). This would force users to use it in either `setUp`, `setUpAll` or `test` contexts. This would then avoid too many concurrent instances by design.

The downside is that instead of simply doing this on the top level:
```dart
var session = sessionBuilder.build();
```
a user would instead need to do this, which is slightly more cumbersome:
```dart
late Session session;
setUpAll(() {
  session = sessionBuilder.build();
});
```

## Changes
- Creates a closure that allows Serverpod to be lazily constructed when needed

## Pre-launch Checklist

### DIFF
--- a/packages/serverpod_test/lib/src/test_serverpod.dart
+++ b/packages/serverpod_test/lib/src/test_serverpod.dart
@@ -94,7 +94,19 @@ class TestServerpod<T extends InternalTestEndpoints> {
   /// The test endpoints that are exposed to the user.
   T testEndpoints;
 
-  late final Serverpod _serverpod;
+  late final Serverpod Function() _buildServerpodAndInitializeEndpoints;
+
+  Serverpod? _serverpodInstance;
+  Serverpod get _serverpod {
+    var serverpodInstance = _serverpodInstance;
+    if (serverpodInstance != null) {
+      return serverpodInstance;
+    }
+
+    serverpodInstance = _buildServerpodAndInitializeEndpoints();
+    _serverpodInstance = serverpodInstance;
+    return serverpodInstance;
+  }
 
   /// Whether the database is enabled and supported by the project configuration.
   final bool isDatabaseEnabled;
@@ -112,22 +124,23 @@ class TestServerpod<T extends InternalTestEndpoints> {
     // Ignore output from the Serverpod constructor to avoid spamming the console.
     // Should be changed when a proper logger is implemented.
     // Tracked in issue: https://github.com/serverpod/serverpod/issues/2847
-    IOOverrides.runZoned(
-      () {
-        _serverpod = Serverpod(
-          _getServerpodStartUpArgs(
-            runMode: runMode,
-            applyMigrations: applyMigrations,
-            loggingMode: serverpodLoggingMode,
-          ),
-          serializationManager,
-          endpoints,
+    _buildServerpodAndInitializeEndpoints = () => IOOverrides.runZoned(
+          () {
+            var serverpod = Serverpod(
+              _getServerpodStartUpArgs(
+                runMode: runMode,
+                applyMigrations: applyMigrations,
+                loggingMode: serverpodLoggingMode,
+              ),
+              serializationManager,
+              endpoints,
+            );
+            endpoints.initializeEndpoints(serverpod.server);
+            return serverpod;
+          },
+          stdout: () => NullStdOut(),
+          stderr: () => NullStdOut(),
         );
-      },
-      stdout: () => NullStdOut(),
-      stderr: () => NullStdOut(),
-    );
-    endpoints.initializeEndpoints(_serverpod.server);
     testEndpoints.initialize(serializationManager, endpoints);
   }
 


hampuslavin:
> async seems reasonable IMHO, it also allows us to introduce async stuff in the builder in the future without breaking stuff. But does this approach really reduce parallel connections much in practice?

@christerswahn Darts' test suites use half of the host's CPU cores ([source](https://pub.dev/packages/test#test-concurrency)) when running concurrently. It should be safe to say this should be less than 100 (default max connection limit for the postgres image). So as long as we don't connect to the database until the test group is actually running, then we should be in the clear.

The problem is that dart loads all test files that it is supposed to run first. If our code instantiates Serverpod instances at this time (which is currently the case) then the there will be as many connection attempts as there are `withServerpod` declarations, which of course will easily surpass 100 for a bigger project.

SandPod:
Comment before review:

I like the suggestion of simply making the session builder async to get around the problem by design. This should futureproof us as well if we, as @christerswahn suggested, add async operations to the build method in the future.

packages/serverpod_test/lib/src/test_serverpod.dart
```diff
@@ -94,7 +94,19 @@ class TestServerpod<T extends InternalTestEndpoints> {
...
+  Serverpod get _serverpod {
+    var serverpodInstance = _serverpodInstance;
+    if (serverpodInstance != null) {
+      return serverpodInstance;
+    }
+
+    serverpodInstance = _buildServerpodAndInitializeEndpoints();
+    _serverpodInstance = serverpodInstance;
+    return serverpodInstance;
+  }
```
SandPod, line 109: Suggestion: I think you can replace this with this:

```suggestion
  Serverpod get _serverpod {
    return _serverpodInstance ??= _buildServerpodAndInitializeEndpoints();
  }
```

packages/serverpod_test/lib/src/test_serverpod.dart
```diff
@@ -112,22 +124,23 @@ class TestServerpod<T extends InternalTestEndpoints> {
...
+          },
+          stdout: () => NullStdOut(),
+          stderr: () => NullStdOut(),
         );
-      },
-      stdout: () => NullStdOut(),
-      stderr: () => NullStdOut(),
-    );
-    endpoints.initializeEndpoints(_serverpod.server);
     testEndpoints.initialize(serializationManager, endpoints);
```
SandPod, line 144: Question: Just to make sure we didn't miss something, is it fine iof endpoints are passed here to the initialize method even if we have not initialized them?

SandPod (commented):
One question and a suggestion for code simplification.

I also like the idea to make the build method async so we get protection around this.

packages/serverpod_test/lib/src/test_serverpod.dart
hampuslavin: Ahh, amazing! 🎉

packages/serverpod_test/lib/src/test_serverpod.dart
hampuslavin: The endpoints only needs to be initialized before being called. Since we will always build out a session before calling an endpoint, which will trigger this lazy build, it can never happen. 

As a side note, this case is if the user does an un-awaited call to an endpoint before the setups/tests which _may_ enter the code path just described depending on execution order.

packages/serverpod_test/lib/src/test_serverpod.dart
hampuslavin: 62d0959

tests/serverpod_test_server/test_integration/test_tools/start_timeout_test/test_that_will_timeout.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:test/test.dart';
+
+import '../serverpod_test_tools.dart';
+
+void main() {
+  withServerpod(
+    'Given that runMode is set to staging when calling withServerpod',
+    // The staging config has intentionally been sedt to not find the database and provoke timeout.
```
SandPod, line 9:
```suggestion
    // The staging config has intentionally been set to not find the database and provoke timeout.
```

tests/serverpod_test_server/test_integration/test_tools/start_timeout_test/start_timeout_test.dart
```diff
@@ -0,0 +1,78 @@
+import 'dart:io';
+
+import 'package:path/path.dart' as path;
+import 'package:test/test.dart';
+
+enum TestTag { include, exclude }
+
+void main() async {
```
SandPod, line 8: Question: Arn't these tests going to be pretty slow? Any possibility to get around the 4 second wait and still timeout on the right thing?

Is the (last) passing test required for example?

tests/serverpod_test_server/test_integration/test_tools/start_timeout_test/test_that_will_timeout.dart
SandPod: This is wrong in `test_that_will_timeout_after_default.dart`.

tests/serverpod_test_server/test_integration/test_tools/start_timeout_test/start_timeout_test.dart
hampuslavin: Yes, this is not good, especially since we're running tests with `concurrency=1`. 

I looked at the [`fake_async` package](https://pub.dev/packages/fake_async), but for some reason this doesn't actually work: 
```dart
  fakeAsync((async) async {
    withServerpod(
      'Given that runMode is set to staging when calling withServerpod',
      // The staging config has intentionally been set to not find the database and provoke timeout.
      runMode: ServerpodRunMode.staging,
      serverpodStartTimeout: Duration(seconds: 20),
      (sessionBuilder, endpoints) {
        test(
            'then dummy test should not be run and exit code should be non-zero',
            () async {
          expect(true, true);
        });
      },
    );

    async.elapse(Duration(seconds: 20));
  });
```

But I think the issue here might the timing of when we're calling `await Future.delayed(const Duration(seconds: 10));` internally in the database connection logic. Especially since we're doing that periodically, so a single `async.elapse` wouldn't work anyway probably. I tried adding in some `await Future.delayed` before the elapse to see if that did some difference but no success.

What we could do is maybe tag these slow tests and run them concurrently to the other integration tests?

hampuslavin: > Is the (last) passing test required for example?

It's more for completeness, since we don't have a passing test with that config anywhere. The passing test it really quick compared to the others though.

tests/serverpod_test_server/test_integration/test_tools/start_timeout_test/test_that_will_timeout.dart
hampuslavin: 98fb7cb

hampuslavin:
@SandPod Noticed while testing the timeout now that we get some unrelated follow-on exceptions in the log. This is just confusing to the user and should not happen: 
f0705ee
Before change:
image

After change:
image

tests/serverpod_test_server/test_integration/test_tools/start_timeout_test/start_timeout_test.dart
SandPod: Discussed IRL: there isn't much to do regarding speeding up the time. In a separate PR, we will separate these test that are long running so that they can be run in parallel.

SandPod (approved):
LGTM!

packages/serverpod_test/lib/src/with_serverpod.dart
```diff
@@ -66,6 +66,7 @@ void Function(TestClosure<T>)
   required RollbackDatabase? maybeRollbackDatabase,
   required bool? maybeEnableSessionLogging,
   required List<String>? maybeTestGroupTagsOverride,
+  required Duration? maybeServerpodStartTimeout,
```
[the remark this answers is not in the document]
hampuslavin: @SandPod I actually think that should be in the generated code to be helpful via intellisense. Otherwise it's just for curious people that follow the code into the library. I can create an issue for that as well.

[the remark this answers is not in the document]
SandPod: Yes, makes sense! I might have commented on the wrong place. But I think it makes sense to provide documentation for the interface as well :)

PR #2907, opened 2024-10-25
fix: Support DB exceptions outside transactions and parallel DB calls.

### LINKED ISSUE #2895: Detect database exceptions and rollback if they happen
**Describe the bug**
The following code is currently not supported in the test tools. The first create can throw a database exception which then invalidates the entire current transaction that the test is running in:
```dart
try {
  return await Resource.create();
} catch (e) {
  if (e is DatabaseException && e.isDuplicateKey()) {
    return await Resource.readFirst()
  }

  rethrow;
}
```
**Suggested solution**
- Wrap each database call inside the database proxy in a save point
- Rollback if a database exception happened
- Re-throw the database exception

It's impossible for the test tools to support this inside a transaction, which is already documented. But supporting it outside a transaction like the example above should be possible (as suggested).

hampuslavin:
## Description
Closes #2895 .

This PR fixes another issue related to emulated nested transaction in the test tools. When an error is raised inside a transaction, the transaction is no longer valid. Therefore the test tools needs to rollback before the error happened but still re-throw it out to the code under test.

Additionally, this PR introduces a synchronized lock to enable code like this:
```dart
test(
    'when creating multiple SimpleData in parallel '
    'then should have inserted all', () async {
  await Future.wait([
    SimpleData.db.insertRow(
      session,
      SimpleData(num: 1),
    ),
    SimpleData.db.insertRow(
      session,
      SimpleData(num: 2),
    ),
    SimpleData.db.insertRow(
      session,
      SimpleData(num: 3),
    ),
  ]);

  var simpleDatas = await SimpleData.db.find(session);

  expect(simpleDatas, hasLength(3));
  expect(simpleDatas.map((s) => s.num), containsAll([1, 2, 3]));
});
```

For rollbacks to work, there can never be more than a single operation that can fail going on at a single time. Therefore the test database proxy only allows a single database operation at a time. 

As soon as the user enters a concurrent mode which is not supported by this approach, an exception is thrown. These cases are:
- Concurrent transaction calls (already threw exception before this change)
- Non-transaction call concurrently happening during an ongoing transaction

## Changes
- Add synchronized lock to database operations in the database test proxy
- Add additional checks for invalid concurrent states that the test tools can't handle
- Unrelated fix to the issue, but this PR also releases save points when they are not needed anymore

## Pre-launch Checklist

### DIFF
--- a/tests/serverpod_test_server/test_integration/test_tools/database_operations_test.dart
+++ b/tests/serverpod_test_server/test_integration/test_tools/database_operations_test.dart
@@ -416,4 +416,99 @@ void main() {
       rollbackDatabase: RollbackDatabase.disabled,
     );
   });
+
+  withServerpod(
+    'Given rollbackDatabase is not disabled ',
+    rollbackDatabase: RollbackDatabase.afterEach,
+    (sessionBuilder, _) {
+      var session = sessionBuilder.build();
+
+      group('when creating UniqueData with the same unique value', () {
+        late Future failingInsert;
+        setUp(() async {
+          await UniqueData.db.insertRow(
+            session,
+            UniqueData(email: 'test@test.com', number: 1),
+          );
+          failingInsert = UniqueData.db.insertRow(
+            session,
+            UniqueData(email: 'test@test.com', number: 1),
+          );
+        });
+
+        test('then should throw database exception', () async {
+          await expectLater(
+            failingInsert,
+            throwsA(
+              allOf(
+                isA<DatabaseException>(),
+              ),
+            ),
+          );
+        });
+
+        test(
+            'then catching database exception should not prevent further database operations',
+            () async {
+          SimpleData? simpleData;
+          try {
+            await failingInsert;
+          } on DatabaseException catch (_) {
+            simpleData =
+                await SimpleData.db.insertRow(session, SimpleData(num: 123));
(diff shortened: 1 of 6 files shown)


Isakdl (approved):
LGTM 🚀, Nice job! Clear tests and easy to follow the logic!

tests/serverpod_test_server/test_integration/test_tools/database_operations_test.dart
```diff
@@ -416,4 +416,99 @@ void main() {
...
+            simpleData =
+                await SimpleData.db.insertRow(session, SimpleData(num: 123));
+          }
+
+          expect(simpleData, isNotNull);
+          expect(simpleData?.num, 123);
+        });
+      });
+
+      test(
```
SandPod, line 466: Question: Are we guaranteed that the first transaction will always happen first? Since they are happening in "parallel" we shouldn't know or care which of the inserts happen first right?

hampuslavin: You're right! The test description should be updated to reflect that.

hampuslavin: 4bbe168

PR #2914, opened 2024-10-29
Add transaction modes

SandPod:

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/packages/serverpod/lib/src/database/concepts/transaction.dart
+++ b/packages/serverpod/lib/src/database/concepts/transaction.dart
@@ -8,3 +8,34 @@ abstract interface class Transaction {
   /// exception depending on driver.
   Future<void> cancel();
 }
+
+/// Isolation levels for transactions.
+enum IsolationLevel {
+  /// A statement can only see rows committed before it began.
+  readCommitted,
+
+  /// Transaction may see uncommitted changes made by other transactions.
+  /// Not supported in PostgreSQL and is treated as readCommitted.
+  readUncommitted,
+
+  /// A transaction can only see rows committed before the first query or
+  /// data-modification statement was executed.
+  repeatableRead,
+
+  /// A transaction can only see rows committed before the first query or
+  /// data-modification statement was executed.
+  ///
+  /// If reads and writes among concurrent serializable transactions occur blablabla
+  serializable,
+}
+
+/// Settings for a transaction.
+class TransactionSettings {
+  /// The isolation level of the transaction.
+  final IsolationLevel isolationLevel;
+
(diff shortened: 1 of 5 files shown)


tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_rollback_test.dart
```diff
@@ -42,4 +42,97 @@ void main() async {
...
+      update(data.id!, session, 'bob'),
+      update(data.id!, session, 'alex'),
+      update(data.id!, session, 'jim'),
+      update(data.id!, session, 'hank'),
+    ]);
+
+    var success = result.where((e) => e).length;
+    expect(success, 1);
+    var dataAfter = await UniqueData.db.findById(session, data.id!);
+    expect(dataAfter?.email, isNot(initialValue));
```
Isakdl, line 93: Question: What is the expected behavior here? One of the update will be successful and the others fails and we cannot know which one it will be?

packages/serverpod/lib/src/database/concepts/transaction.dart
```diff
@@ -8,3 +8,34 @@ abstract interface class Transaction {
...
 }
+
+/// Isolation levels for transactions.
+enum IsolationLevel {
+  /// A statement can only see rows committed before it began.
+  readCommitted,
+
+  /// Transaction may see uncommitted changes made by other transactions.
+  /// Not supported in PostgreSQL and is treated as readCommitted.
+  readUncommitted,
```
christerswahn, line 19: Should we remove this if it is not supported anyway?

christerswahn (commented):
Nice interface to enable settings!

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_rollback_test.dart
SandPod: Exactly! We just want to validate that a single update is successful and that this has been committed to the database.

packages/serverpod/lib/src/database/concepts/transaction.dart
SandPod: Yes maybe, My thinking was that the "interface" should be agnostic to if this is supported in the database or not, while the adapter takes care of it. But I'm not sure there.

SandPod: Or reading a bit more, it is actually possible to set the transaction level `readUncommitted` in postgres, its only that the behavior will be exactly the same as `readCommitted`. So my description here is slightly incorrect.

Issue #2495, opened 2024-07-16
Add abstraction of requests and responses in Relic


vlidholt:
Relic is Serverpod's built-in web server. Adding an abstraction layer on top of `dart:io` could be beneficial for a number of reasons. It would make it easier to work with Relic and make the interface future-proof if we want to move away from `dart:io`.

The `shelf` package has done a pretty good job with this, but it has some shortcomings. We could create a new package based on `shelf` (to preserve its licence, and give them proper credit). We can then tighten up some of its shortcomings. The new package could be named `serverpod_relic_helpers` or if we come up with a better name. From the `shelf` package we can copy and modify the following classes: `Message`, `Request`, `Response`, and `Body`.

The most significant issues with the types from `shelf` are that they are largely untyped (my guess because of legacy), and they are using `List<int>` to represent data (which is much less efficient than using `Uint8List`). Adding type-safety and using `Uint8List` will improve performance and memory usage (especially for large requests).

## Modifications to the shelf `Body` class
Instead of having a generic constructor that takes any type of object, we could have three distinct, typed constructors:

```dart
Body.fromString(String body, {Encoding? encoding});

Body.fromDataStream(Stream<Uint8List> body);

Body.fromData(Uint8List body);
```

The underlying stream can be accessed through either the read method, which should return a `Stream<Uint8List>` (instead of the `Stream<List<int>>`) or `readAsString`, which decodes the full stream into a `String`.

## Modifications to the shelf `Request` class
We can take the opportunity to add the parameters that are parsed by the new `Router` (#2489) to the `Request` class, making them easy to access:

```dart
Map<String, String> pathParameters;
```

## Modifications to Relic's Route class
To support the new abstractions, the `handleRequest` method of the Route class can be structured like this:

```dart
Future<Response> handleRequest(Session session, Request request);
```

SandPod:
My understanding is that the biggest shortcoming of shelf is that it does not provide type-safety and uses `List<int>´ instead of `Uint8List`.

How big of an undertaking is it to introduce these changes? Is there any possibility for us to add these changes as contributions directly to shelf instead of creating a fork?

We kind of end up needing to copy or mimic the classes regardless if we want an abstraction layer, but it could help us make it smaller.

SandPod:
This issue is implemented as the new Relic library, a standalone web server that will be used to drive the web server in Serverpod: https://github.com/serverpod/relic

SandPod:
[...]

PR #2939, opened 2024-11-06
Make `Route.getBody` handle `application/json` content type body

marcelomendoncasoares:
I'm currently developing a backend where I need to build a webhook endpoint that adheres to a specific contract that sends POST messages with an `application/json` content type. For this, I saw on the #2140 discussion that there is the experimental `webserver`, but the body parsing is limited.

Currently, the `Route.getBody` method only handles `application/x-www-form-urlencoded` body types. This PR allows the method to also parse `application/json` types, which are broadly used. Although this change is far from the major abstraction proposed on #2495, it aims to temporarily improve the coverage of a very common use case while the #2008 improvements are still ongoing.

I understand that the `relic` package is one of the main focuses right now and this change is on its experimental version, so I will also understand if approving this PR makes no much sense for the timeline. Anyway, it is pleasure to contribute.

## Pre-launch Checklist

- [ ] ~I added new tests to check the change I am making.~
  _(Because (1) the webserver is still experimental, (2) there are nearly no tests for the `relic` module and (3) the change consists in a very simple 3-line addition, it seemed to me that there was no reason to include a test. But I may be wrong.)_

## Breaking changes

The only breaking change would be the typing change of the `getBody` return, from `Map<String, String>` to `Map<String, dynamic>`. However, this method is not used inside the `serverpod` code itself. The change may only break external code during compile time if there are variables strictly typed to expect `Map<String, String>`.

### DIFF
--- a/packages/serverpod/lib/src/relic/web_server.dart
+++ b/packages/serverpod/lib/src/relic/web_server.dart
@@ -240,13 +240,17 @@ abstract class Route {
   // types of responses too. Or at least clarify the naming of the method.
 
   /// Returns the body of the request, assuming it is standard URL encoded form
-  /// post request.
-  static Future<Map<String, String>> getBody(HttpRequest request) async {
+  /// post request or JSON.
+  static Future<Map<String, dynamic>> getBody(HttpRequest request) async {
     var body = await _readBody(request);
 
     var params = <String, String>{};
 
     if (body != null) {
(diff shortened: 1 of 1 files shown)


SandPod:
Hey @marcelomendoncasoares!

Sorry for the late reply on this one.

We are working on improving the webserver now, but I'll see if we can get this in to add support for this specific usecase.

Thank you for keeping the branch up to date.

Issue #2936, opened 2024-11-05
Support configuring transaction isolation level


SandPod:
**Is your feature request related to a problem? Please describe.**
Postgres supports multiple transaction isolation levels: https://www.postgresql.org/docs/current/transaction-iso.html

Serverpod should make it possible to configure these for each transaction.

**Describe the solution you'd like**
Typed interface for configuring the transaction isolation level used for a transaction.

The isolation levels that should be supported are:
- readCommited
- readUncommitted
- repetableRead
- serializable

PR #2943, opened 2024-11-07
feat: Add support for transaction isolation level

SandPod:
Closes: https://github.com/serverpod/serverpod/issues/2936

Add support for transaction isolation levels for the database.

## Pre-launch Checklist

## Breaking changes

None - database interface is extended in a backwards compatible way.


tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_isolatation_level_test.dart
```diff
@@ -0,0 +1,292 @@
+import 'dart:async';
+
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+import 'package:serverpod_test_server/test_util/test_serverpod.dart';
+import 'package:test/test.dart';
+
+void main() async {
+  var session = await IntegrationTestServer().session();
```
hampuslavin, line 9: Question: Why couldn't we use our test tools but and disable the database rollbacks?

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_isolatation_level_test.dart
```diff
@@ -0,0 +1,292 @@
...
+      isolationLevel: IsolationLevel.readCommitted,
+    );
+
+    test(
+        'when row is modified after first statement in transaction'
+        'then transaction observes the updated value.', () async {
+      var testData = await SimpleData.db.insertRow(
+        session,
+        SimpleData(num: 1),
+      );
```
hampuslavin, line 28: Suggestion: Move this out to `group` level to match the Given statement.

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_isolatation_level_test.dart
```diff
@@ -0,0 +1,292 @@
...
+
+  group(
+      'Given read committed transaction isolation level and single row in database',
+      () {
+    var settings = TransactionSettings(
+      isolationLevel: IsolationLevel.readCommitted,
+    );
+
+    test(
+        'when row is modified after first statement in transaction'
```
hampuslavin, line 23: nit: when breaking test descriptions up like this (which I really like) then a space has to be added to the end of each line otherwise the resulting text is concatenated.

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_isolatation_level_test.dart
```diff
@@ -0,0 +1,292 @@
...
+        testData.copyWith(num: 2),
+      );
+      c2.complete();
+
+      var result = await transactionFuture;
+      expect(result?.num, 1);
+    });
+
+    test(
+        'when read row is concurrently modified by other transaction'
```
hampuslavin, line 110: Question: Should there be an additional test for the isolation level Repeatable Read where updates are made to the row in which case it should throw?

hampuslavin (commented):
Nice work! 👏 Some small questions and suggestions.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -542,12 +542,29 @@ class DatabaseConnection {
...
+    pg.TransactionSettings? transactionSettings;
+    if (settings != null) {
+      transactionSettings = pg.TransactionSettings(
+        isolationLevel: switch (settings.isolationLevel) {
+          IsolationLevel.readCommitted => pg.IsolationLevel.readCommitted,
+          IsolationLevel.readUncommitted => pg.IsolationLevel.readUncommitted,
+          IsolationLevel.repeatableRead => pg.IsolationLevel.repeatableRead,
+          IsolationLevel.serializable => pg.IsolationLevel.serializable,
+          null => null,
+        },
```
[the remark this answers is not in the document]
SandPod: The TransactionSettings object lives inside of the "Serverpod" domain. It shouldn't have any knowledge of domains using it. Therefore, the adapter should capture the conversion, which is the `database_connection.dart` class. I opted to just do it inline now since there is only a single setting.

But a cleaner solution would maybe be to add a `fromServerpodTransaction` constructor to the driver settings object.

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_isolatation_level_test.dart
SandPod: Good catch! ⭐

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_isolatation_level_test.dart
SandPod: Aah thanks! I stole this from you I think, i also really like it! Will fix 🙏

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_isolatation_level_test.dart
SandPod: I wanted to stay out from testing the actual behavior of these things and focus on us setting the right configuration in the database. That is why I have developed the tests so that a set is unique for each configuration flag.

I don't think we need that test to be honest, we should be able to trust postgres to behave correctly.

packages/serverpod/lib/src/database/concepts/transaction.dart
```diff
@@ -8,3 +8,49 @@ abstract interface class Transaction {
...
+/// Isolation levels for transactions.
+enum IsolationLevel {
+  /// Each statement in the transaction sees a snapshot of the database as of
+  /// the beginning of the statement. This means each statement might observe
+  /// a different version of the database.
+  readCommitted,
+
+  /// Allow transaction to see uncommitted changes made by other transactions.
+  /// Though in PostgreSQL, this behaves like read committed.
+  readUncommitted,
```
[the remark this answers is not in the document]
SandPod: Fixed: e63c6f53acf504666b8baedad58678eedf7ac0a6

packages/serverpod/lib/src/database/database.dart
```diff
@@ -316,9 +316,13 @@ class Database {
   }
 
   /// Executes a [Transaction].
-  Future<R> transaction<R>(TransactionFunction<R> transactionFunction) async {
+  Future<R> transaction<R>(
+    TransactionFunction<R> transactionFunction, {
+    TransactionSettings? settings,
```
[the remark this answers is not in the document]
SandPod: 6d19369465d23f5b1c8099d475b99766165047cf

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_isolatation_level_test.dart
SandPod: 787086293005dec16c4d270ae70baf326e0a69da

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_isolatation_level_test.dart
SandPod: 787086293005dec16c4d270ae70baf326e0a69da

packages/serverpod/lib/src/database/database.dart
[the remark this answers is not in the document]
hampuslavin: Ah yes, I commented in the wrong place! Nice 👍

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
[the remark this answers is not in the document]
hampuslavin: Makes sense! Maybe something to do if this method would grow. I think it's fine then to keep as is for now 👍

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_isolatation_level_test.dart
hampuslavin: That's a good point!

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_isolatation_level_test.dart
hampuslavin: I think it makes sense to do the switch so that we're moving toward the `withServerpod` test paradigm (even though it will be a long migration process). So in this case you would have to add the following and it will work:
```dart
import 'package:serverpod_test_server/test_util/test_tags.dart';
// ...
rollbackDatabase: RollbackDatabase.disabled,
testGroupTagsOverride: [concurrencyOneTestTag],
```

hampuslavin: (as a bonus, acutally I think we should do like i did in cloud and declare the test tag like this):
```dart
abstract final class TestTags {
  concurrencyOne = 'concurrencyOne';
}
```

hampuslavin (approved):
Approved given that we change to `withServerpod` in the test 👍 🚀

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_isolatation_level_test.dart
SandPod: 60a9018f02e19879afb9fddbceda5d686c4767eb and 2165a9d334f91ba32fa4cb112a9565eb8025ffc6

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -542,12 +542,26 @@ class DatabaseConnection {
   }
 
   /// For most cases use the corresponding method in [Database] instead.
-  Future<R> transaction<R>(TransactionFunction<R> transactionFunction) {
+  Future<R> transaction<R>(
```
SandPod, line 545: Intentionally left as an unresolved conversation.

Don't merge this until 2.2 has been released.

hampuslavin (approved):

PR #2955, opened 2024-11-13
feat: Improved Serverpod startup and stream lifecycle events logging

christerswahn:
Improved Serverpod initialization log messages, and added verbose logging of method streams lifecycle events.

## Pre-launch Checklist

## Breaking changes

n/a

### DIFF
--- a/packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
+++ b/packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
@@ -157,6 +157,10 @@ class MethodStreamManager {
         _onOutputStreamError = onOutputStreamError,
         _onOutputStreamValue = onOutputStreamValue;
 
+  int get openInputStreamCount => _inputStreamContexts.length;
+
+  int get openOutputStreamCount => _outputStreamContexts.length;
+
   Future<void> closeAllStreams() async {
     var inputControllers =
         _inputStreamContexts.values.map((c) => c.controller).toList();
@@ -336,9 +340,12 @@ class MethodStreamManager {
       /// or a request from the client.
       if (isCancelled) return;
       isCancelled = true;
+      session.serverpod.logVerbose(
+          'Cancelling method output stream for ${session.endpoint}.${session.method}, id $methodStreamId');
       await revokedAuthenticationHandler?.destroy(session);
       await _closeOutboundStream(methodStreamCallContext, methodStreamId);
-      await session.close();
+      await session
+          .close(); // Will this also close the input streams? If so is this the desired behavior?
     });
 
     late StreamSubscription subscription;
@@ -431,6 +438,9 @@ class MethodStreamManager {
     for (var streamParam in callContext.inputStreams) {
       var parameterName = streamParam.name;
       var controller = StreamController(onCancel: () async {
+        callContext.endpoint.pod.logVerbose(
+            'Cancelling method input stream for ${callContext.fullEndpointPath}.'
+            '${callContext.method.name}.$parameterName, id $methodStreamId');
         var context = _inputStreamContexts.remove(_buildStreamKey(
           endpoint: callContext.fullEndpointPath,
           method: callContext.method.name,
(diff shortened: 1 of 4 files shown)


packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
```diff
@@ -336,9 +340,12 @@ class MethodStreamManager {
       /// or a request from the client.
       if (isCancelled) return;
       isCancelled = true;
+      session.serverpod.logVerbose(
+          'Cancelling method output stream for ${session.endpoint}.${session.method}, id $methodStreamId');
       await revokedAuthenticationHandler?.destroy(session);
       await _closeOutboundStream(methodStreamCallContext, methodStreamId);
-      await session.close();
+      await session
+          .close(); // Will this also close the input streams? If so is this the desired behavior?
```
SandPod, line 348: Suggestion: Remove comment.

Answer to the comment: These are closed in the _closeOutboundStream method and yes, this is the desired behavior. The lifecycle of a stream is determined by the output stream. If it is closed, all input streams should also be closed.

packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
```diff
@@ -336,9 +340,12 @@ class MethodStreamManager {
       /// or a request from the client.
       if (isCancelled) return;
       isCancelled = true;
+      session.serverpod.logVerbose(
```
SandPod, line 343: Question/Suggestion: Should we log these for the user with the debug log level?

Since this "action" is associated with a specific user it would make sense to track the log together with the other user logs.

packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
```diff
@@ -431,6 +438,9 @@ class MethodStreamManager {
     for (var streamParam in callContext.inputStreams) {
       var parameterName = streamParam.name;
       var controller = StreamController(onCancel: () async {
+        callContext.endpoint.pod.logVerbose(
```
SandPod, line 441: Question: Same suggestion here as above: should these be logged for the user?

A tricky thing here is that the session can be closed as a result of closing the output stream. So if we want to log something for the user we need to be mindful about that.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -36,7 +36,13 @@ class MethodWebsocketRequestHandler {
         }
 
         switch (message) {
-          case OpenMethodStreamCommand():
+          case OpenMethodStreamCommand(
+              endpoint: var endpoint,
+              method: var method,
+              connectionId: var connectionId,
+            ):
```
SandPod, line 43: Question(For my knowledge): This was neat; I'm not that familiar with this. Does this actually create a new instance of `OpenMethodStreamCommand` or does it just assign a variable to the fields from the message received?

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -87,13 +105,18 @@ class MethodWebsocketRequestHandler {
         }
       }
     } catch (e, stackTrace) {
-      if (server.serverpod.runtimeSettings.logMalformedCalls) {
+      if (e is! UnknownMessageException ||
```
SandPod, line 108: Discussion: This is the only change that scares me a bit. I'm not really sure in what cases we might have initiated a connection that then fails becuase the client has bad intent.

We might need to expose a server and try this out a bit before including it in the real release.

packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
christerswahn: Thx!

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
christerswahn: It destructures the object being matched, i.e. does not create a new object, just assigns the specified vars from public properties of the matched object.

Actually the previous code was already using destructuring, we just didn't extract any attributes!

SandPod (approved):
LGTM! 👍

PR #2969, opened 2024-11-19
test: Skip starting server in legacy test framework.

SandPod:
### Problem
Because the legacy test framework binds to the same ports, starting up multiple instances of Serverpod is not possible.

The session object in the legacy test framework utilizes a finalizer to shut down the Server. Because the finalizer's runtime is unpredictable, it sometimes does not have time to run before the next test starts. This means that an instance of Serverpod is already running when an attempt to start a new one is made.

And with the old code, this would simply exit the process with exit code 1, preventing any further test execution.

This causes flakiness in integration tests where they might pass in the PR branch but then, later on, fail in the main.

### Fix
This PR removes the start of the server in the legacy test framework. 

The real solution is to migrate these tests to the new test framework.

### Additional minor fixes

- Adds a missing test tag definition from the module test directory.
- Migrated a couple of email auth provider tests to the new framework because for some reason the finalizer was triggered before the test could complete (closing the session).

## Pre-launch Checklist

## Breaking changes

None - simple test framework fix.

### DIFF
--- a/tests/serverpod_test_server/test_integration/auth_module/email_auth_provider/account_request_test.dart
+++ b/tests/serverpod_test_server/test_integration/auth_module/email_auth_provider/account_request_test.dart
@@ -2,7 +2,7 @@ import 'package:serverpod/database.dart';
 import 'package:serverpod_auth_server/module.dart';
 import 'package:test/test.dart';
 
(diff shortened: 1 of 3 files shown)


Isakdl (approved):
LGTM

PR #2971, opened 2024-11-19
fix(Inheritance): Sealed Class Hierarchy Generation Limited to Same Directory

### LINKED ISSUE #2893: [Inheritance] Sealed Class Hierarchy Generation Limited to Same Subdirectory
As of implementing #2888, we have introduced functionality to generate `sealed` classes and hierarchies.

**Current Limitation**
It only works when all member classes of a sealed hierarchy are located in the same sub-directory.

**Problem** Given the case:
```
models/
  └── sealed_top_node.dart
  └── someSubDir/
          └── child_of_top_node.dart
```
then the automatic created imports are **located** in `sealed_top_node.dart` but are **relative to** `child_of_top_node.dart`. 

To avoid running into errors I added a restriction throwing an `Exception` during generation, informing the user that all model files in a sealed hierarchy have to be located in the same sub directory.

We might be able to fix that by passing the `top node` or it's `path` to the `PartAllocator` and `PartOfAllocator`, which can be used to adjust the import created while allocating the `sub class` to be relative to the `top node`. 

**Describe alternatives you've considered**
Keep the restriction throwing an exception when a file in the sealed hierarchy has a different sub dir.

**Additional context**
`client_code_generator.dart` & `server_code_generator.dart`: Here we instantiate and assign `PartOfAllocator` and `PartAllocator` for each model in the sealed hierarchy.

```dart
 ...
 for (var sealedHierarchy in sealedHierarchies) {
      var partOfAllocator = PartOfAllocator([]);
      var partAllocator = PartAllocator(partOfAllocator);

      for (var model in sealedHierarchy) {
        modelAllocatorContext.add((
          model: model,
          allocator: model.isSealedTopNode ? partAllocator : partOfAllocator
        ));
      }
    }
 ...
```

`custom_allocators.dart`
```dart
import 'package:code_builder/code_builder.dart';

class PartOfAllocator implements Allocator {
  static final _doNotPrefix = ['dart:core'];

  final _imports = <String, int>{};
  var _keys = 1;

  PartOfAllocator(List<String>? doNotPrefix) {
    _doNotPrefix.addAll(doNotPrefix ?? []);
  }

  @override
  String allocate(Reference reference) {
    var symbol = reference.symbol;
    var url = reference.url;

    if (url == null || _doNotPrefix.contains(url)) {
      return symbol!;
    }

    return '_i${_imports.putIfAbsent(url, _nextKey)}.$symbol';
  }

  int _nextKey() => _keys++;

  @override
  Iterable<Directive> get imports => [];
}

class PartAllocator implements Allocator {
  PartOfAllocator _partOfAllocator;

  PartAllocator._(this._partOfAllocator);

  factory PartAllocator(PartOfAllocator partOfAllocator) {
    return PartAllocator._(partOfAllocator);
  }

  @override
  String allocate(Reference reference) {
    return _partOfAllocator.allocate(reference);
  }

  @override
  Iterable<Directive> get imports => _partOfAllocator._imports.keys.map(
        (u) => Directive.import(u, as: '_i${_partOfAllocator._imports[u]}'),
      );
}

```

BenAuerDev:
This fixes the limitation that all members of a sealed hierarchy can only be located in the same directory.

Changes:
- Remove the `restriction` and corresponding `test` disallowing sealed hierarchy members in separate directories

- `custom_allocators.dart`  extract the functionality collecting and aliasing `imports`  into a `ImportCollector`, which is to be instantiated with the allocators and is passed the `path` of the corresponding `model`. It normalizes the path to avoid duplication and then set's it relative to the top node

- `class_generators_util.dart` adds a function `getGeneratedModelPath(...)` returning a `String` with the models path 

- `model_library_generator.dart` add `part` and `part of` directive with relative paths

- `server_code_generator.dart` and `client_code_generator.dart` instantiate `ImportCollector` and for each sealed hierarchy and use `getGeneratedModelPath` to avoid repetition.

- `sealed_class_test.dart` adds 3 tests checking if the directives (`part` and `part of`) are generated with the correct uri

- `compilation_unit_helpers.dart` created `applyWindowsBugfix(uri)` as separate method used in `tryFindPartDirective` and `tryFindPartOfDirective`

closes #2893

## Pre-launch Checklist

## Breaking changes
No breaking changes


BenAuerDev:
Sorry for all the failed workflows :x: 

it was a bit hard to debug the failing tests only for windows. 

I ended up adding the following to `tryFindPartDirective` and `tryFindPartOfDirective`:

```dart
 // Windows-specific: separator fix
if (Platform.isWindows) {
    directiveUri = directiveUri.replaceAll('/', '');
    uri = uri.replaceAll('/', '');
 }
```

as `CompilationUnit.directive.uri.stringValue` on `windows` returns a path without separators, so 

so what is `../filename.dart` on `linux` 
is returned as `..filename.dart` on `windows`

therefore all conventional attempts  to normalize and then compare the paths failed, so I went with a less conventional attempt that is only used on windows

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -52,15 +53,32 @@ class SerializableModelLibraryGenerator {
       (libraryBuilder) {
         if (classDefinition.isSealedTopNode) {
           for (var child in classDefinition.descendantClasses) {
-            libraryBuilder.directives
-                .add(Directive.part('${child.fileName}.dart'));
+            var childPath = p.relative(
+              p.joinAll([
+                ...child.subDirParts,
+                '${child.fileName}.dart',
+              ]),
```
SandPod, line 60: Suggestion: Make this an extension on `classDefinitions`.

We seem to be building this same path over and over again. Might as well make it an extension and use it everywhere in the generator.

tools/serverpod_cli/test/generator/dart/server_code_generator/sealed_class_test.dart
```diff
@@ -750,24 +750,23 @@ void main() {
...
+        parseString(content: codeMap[grandparentPath]!).unit;
+    var parentCompilationUnit = parseString(content: codeMap[parentPath]!).unit;
+    var childCompilationUnit = parseString(content: codeMap[childPath]!).unit;
 
     group('then ${grandparent.className}', () {
       test('has a part directive with ${parent.className} uri', () {
         var partDirective = CompilationUnitHelpers.tryFindPartDirective(
           grandparentCompilationUnit,
-          uri: '../${parent.fileName}.dart',
+          uri: p.relative(parentPath, from: p.dirname(grandparentPath)),
```
SandPod: Question(looking at the commit that introduced this): This is where we introduce the windows bug right? When the path is no longer "dart" compatible.

tools/serverpod_cli/lib/src/generator/dart/server_code_generator.dart
```diff
@@ -33,14 +34,26 @@ class DartServerCodeGenerator extends CodeGenerator {
         ModelFilterUtil.getClassesWithoutSealedHierarchies(models);
 
     for (var sealedHierarchy in sealedHierarchies) {
-      var partOfAllocator = PartOfAllocator([]);
-      var partAllocator = PartAllocator(partOfAllocator);
+      var topNode =
```
SandPod, line 37: Question/Discussion: This code needs to be kept in sync between the client and the server. Maybe it is time to break it out into a helper that manages this? Seems unecessary to repeate the code

tools/serverpod_cli/lib/src/test_util/compilation_unit_helpers.dart
```diff
@@ -129,10 +130,17 @@ abstract class CompilationUnitHelpers {
...
   }) {
-    return unit.directives
-        .whereType<PartDirective>()
-        .where((directive) => directive.uri.stringValue == uri)
-        .firstOrNull;
+    return unit.directives.whereType<PartDirective>().where((directive) {
+      String directiveUri = directive.uri.stringValue!;
+
+      // Windows-specific: separator fix
+      if (Platform.isWindows) {
```
SandPod, line 137: Suggestion: Break this out into a shared helper between the methods.

Also add documentation and possibly a link go the bug that you created around why this is needed.

SandPod (changes requested):
Nice work!

Will be neat to wrap this up! :)
I have a couple of comments and suggestions for improvement. Please have a look and let me know if anything is unclear.

tools/serverpod_cli/test/generator/dart/server_code_generator/sealed_class_test.dart
BenAuerDev: No the tests on the commit before failed too. 

But according to [this comment on the bug report](https://github.com/dart-lang/sdk/issues/59629#issuecomment-2512094587) it should have already worked with `../${parent.fileName}.dart`.

SandPod (changes requested):
Nice to see us harmonizing the code paths between the client and the server generator.


Couple of minor last comments.

SandPod (approved):
LGTM! 👍

Issue #964, opened 2023-05-29
Support different project structures


Isakdl:
Today you have to put the endpoints inside the endpoint folder, this is fine and good for smaller projects and is okay for most simple use cases. However, it would be nice to support different project layouts.

Default structure
```
- lib
  - business 
  - endpoints
  - generated
  - protocol
  ...
```

You can today expand this into a subfolder structure, like so:

```
- lib
  - business 
    - feat1
    - feat2
  - endpoints
    - feat1
    - feat2
  - generated
    - feat1
    - feat2
  - protocol
    - feat1
    - feat2
  ...
```

But when a project starts to grow it is often beneficial to order it by feature rather than by type, since you group your code that logically belongs together in the same folder. This makes it easier to jump between the different files when working on a specific feature. Today you cannot use this type of structure as we are looking for the file types in the respective folder.

Example of structure to support, group by feature:
```
- lib
  - feat1
     - business 
     - endpoints
     - generated
     - protocol
  - feat2
     - business 
     - endpoints
     - generated
     - protocol
  ...
```

Allowing full customization of where you can place the files would also enable the programmer(s) to make whatever project structure they may want, even something different than the group by feature proposed here.

**The problems to solve:**
* Finding the files to parse.
  - Endpoints
  - yaml files (protocols)

* Knowing where to place the generated dart files.

PR #2973, opened 2024-11-21
fix: Use statefull analyzers in generator

SandPod:
Uses stateful analyzers to analyze both endpoints and model files in the generator.

This PR is a preparation for supporting endpoint and model files located anywhere in the project that is tracked here  https://github.com/serverpod/serverpod/issues/964

This PR also introduces additional filtering on the watched files to only trigger a regeneration when either an endpoint or a model file is updated. This will be a necessary change once we listen to all files in the project directory to prevent unnecessary regenerations for other file changes.

A small delay was added to generate files so that multiple changes are batched together. For example, when a file is moved from one folder to another, this creates an `add` and a `remove` event, which could introduce an invalid state if they are processed separately. We already have tests that validate this in the e2e generate watch tests.

## Pre-launch Checklist

## Breaking changes

None - the changes in this PR should preserve the current generated behavior.

### DIFF
--- a/tools/serverpod_cli/lib/src/analyzer/dart/endpoints_analyzer.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/dart/endpoints_analyzer.dart
@@ -27,6 +27,8 @@ class EndpointsAnalyzer {
           resourceProvider: PhysicalResourceProvider.INSTANCE,
         );
 
+  Set<EndpointDefinition> _endpointDefinitions = {};
+
   /// Analyze all files in the [AnalysisContextCollection].
   /// Use [changedFiles] to mark files, that need reloading.
   Future<List<EndpointDefinition>> analyze({
@@ -95,6 +97,7 @@ class EndpointsAnalyzer {
       ));
     }
 
+    _endpointDefinitions = endpointDefs.toSet();
     return endpointDefs;
   }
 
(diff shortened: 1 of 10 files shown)


tools/serverpod_cli/lib/src/commands/generate.dart
```diff
@@ -59,18 +61,25 @@ class GenerateCommand extends ServerpodCommand {
         Directory(path.joinAll(config.endpointsSourcePathParts));
     var endpointsAnalyzer = EndpointsAnalyzer(endpointDirectory);
 
+    var protocols = await ModelHelper.loadProjectYamlModelsFromDisk(config);
```
Isakdl, line 64: Suggestion: Rename protocols to models, as that is what we are loading.

Isakdl: Or perhaps yamlModels is more accurate

tools/serverpod_cli/lib/src/generator/generator_continuous.dart
```diff
@@ -32,6 +33,8 @@ Future<bool> performGenerateContinuously({
 
   var modelSourcePath = p.joinAll(config.modelSourcePathParts);
   var protocolSourcePath = p.joinAll(config.protocolSourcePathParts);
+
+  Timer? generateDispatch;
```
Isakdl, line 37: Suggestion(nit): Rename to debounceDispatch as that is what we are doing!

Isakdl (approved):
LGTM, but I have added some nit suggestions!

tools/serverpod_cli/lib/src/analyzer/dart/endpoints_analyzer.dart
```diff
@@ -176,6 +201,17 @@ class EndpointsAnalyzer {
     }
   }
 
+  bool _isEndpointFile(File file) {
+    if (!file.absolute.path.startsWith(absoluteIncludedPaths)) return false;
+    if (!file.path.endsWith('.dart')) return false;
+    if (!file.existsSync()) return false;
+
+    var contents = file.readAsStringSync();
+    if (!contents.contains('extends Endpoint')) return false;
```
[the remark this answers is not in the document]
SandPod: Very true! This should work as our source of truth around what files we analyze. So we could update this spot when that happens.

PR #2987, opened 2024-11-27
feat: Support free model definition placement

SandPod:
Closes: #964 

Add support for placing model definitions anywhere in the `lib` directory if they use a `*.spy`, `*.spy.yaml` or `*.spy.yml` extension. 

The model files are still generated in the `lib/src/generated` directory, and the subdirectory path for the generated model files will have a base from either of these paths:
- `lib`
- `lib/src`
- `lib/src/protocol`
- `lib/src/models`

This means that a new restriction has been added to prevent file collisions for the generated files. For example if a model file with the same name is placed in `lib/src/protocol` or `lib/src/models/`, this would result in the generated model file having the same name.

## Pre-launch Checklist

## Breaking changes

None - this introduces more locations to place model files with the `spy` extension.


tools/serverpod_cli/lib/src/util/model_helper.dart
```diff
@@ -34,31 +34,17 @@ class ModelHelper {
     var modelSources = <ModelSource>[];
 
     var modelSource = await _loadYamlModelsFromDisk(
-      defaultModuleAlias,
-      _absolutePathParts(config.modelSourcePathParts),
+      moduleAlias: defaultModuleAlias,
       loadConfig: config,
+      makeConfigPathsAbsolute: true,
```
Isakdl, line 47: Comment: I think this is a little dirty solution, but I also know the struggle you had with the code so perhaps we need to have a rethink about the structure here in the future.

tools/serverpod_cli/lib/src/util/model_helper.dart
```diff
@@ -142,30 +136,40 @@ class ModelHelper {
         ));
   }
 
-  static List<String> extractPathFromConfig(ModelLoadConfig config, Uri uri) {
-    List<List<String>> paths = [
+  static List<String> extractPathFromConfig(
```
Isakdl, line 151: Comment: Just a note, do we have tests for this?

tools/serverpod_cli/lib/src/test_util/builders/model_source_builder.dart
```diff
@@ -2,30 +2,33 @@ import 'package:path/path.dart';
...
+        yaml = '''
     class: Example
     fields:
       name: String
-    ''');
-    _yamlSourcePathParts = ['lib', 'src', 'model', 'example.yaml'];
-    protocolRootPathParts = [];
-    moduleAlias = defaultModuleAlias;
-  }
+    ''';
```
Isakdl, line 23: Nit: spaces

tools/serverpod_cli/lib/src/analyzer/models/validation/model_validator.dart
```diff
@@ -42,6 +43,29 @@ List<SourceSpanSeverityException> validateTopLevelModelType(
...
+  var parsedModels = restrictions.parsedModels;
+
+  if (model == null) return [];
+
+  if (!parsedModels.isFilePathUnique(model)) {
+    var otherClass = parsedModels.findByFilePath(model, ignore: model);
+
+    return [
+      SourceSpanSeverityException(
+        'File collision with "${otherClass.className}" was detected for the generated model, please provide a unique path or filename for the model.',
```
Isakdl, line 60: Suggestion: This could be made a little more clear for the end user.

"File path", "File name" collision for "classname, with path: $path" collides with "classname, with path $path"

Something like that.

tools/serverpod_cli/test/integration/util/model_helper/model_file_load_test.dart
```diff
@@ -458,6 +458,61 @@ void main() {
...
+        var modelFile = File(join(
+          testProject.path,
+          'lib',
+          modelFileName,
+        ));
+        modelFile.createSync(recursive: true);
+        modelFile.writeAsStringSync('''
+  class: Example
+  fields:
+    name: String
```
Isakdl, line 474: Nit: The real world file does not have spacing at the start.

Isakdl (changes requested):
Good job, this was harder than expected! 💪 

I some smaller comments!

tools/serverpod_cli/lib/src/util/model_helper.dart
SandPod: Yes, these are covered in `model_helper_test` as well as in the model file load tests (that use this internally).

tools/serverpod_cli/lib/src/util/model_helper.dart
SandPod: Fixed! Reverted back to the original code where the absolute path is configured in the call of the method.

tools/serverpod_cli/lib/src/analyzer/models/validation/model_validator.dart
SandPod: To make this really clear for the user we should include the path to the colliding model definition. In order to to that we need to refactor how we store models in the stateful analyzer. I'll give this a stab in a separate PR and improve the message once that is done.

Isakdl (approved):
LGTM

Issue #2990, opened 2024-11-27
errors running serverpod generate using v2.2.0


starklord:
Hi, while updating serverpod from 2.1.5 to 2.2.0
there is a problem when running "serverpod generate"
the file lib/src/generated/endpoints.dart
create contraslashes instead slasche for the import routes. For example:
import '..\endpoints\branch_service.dart' as _i2;
import '..\endpoints\enterprise_service.dart' as _i3;
import '..\endpoints\finance\cash_service.dart' as _i4;
import '..\endpoints\hr\employee_service.dart' as _i5;

that gives me the error: "Target of URI doesn´t exists"

so i am changing manually to this:

import '../endpoints/branch_service.dart' as _i2;
import '../endpoints/enterprise_service.dart' as _i3;
import '../endpoints/finance/cash_service.dart' as _i4;
import '../endpoints/hr/employee_service.dart' as _i5;

and everythigs seems running ok. But while its a generated file i shouldnt do this change manually.

My Operating System is Windows 11, using VSCode

SandPod:
Thank you for the report!

We will create a fix for this ASAP!

Issue #785, opened 2023-03-25
Move scripts from `utils/` to `serverpod_cli`


fischerscode:
**Is your feature request related to a problem? Please describe.**
I'm currently developing on Windows. Since the `utils/` scripts don't work on Windows, it's hard to `pub get` all or `serverpod generate` all.

**Describe the solution you'd like**
It would be great if those things would be added to `serverpod_cli`. There could be an `internal` command with all the sub-commands that are needed when contributing to Serverpod.

**Describe alternatives you've considered**
At the moment, I'm using `melos` to make my life easier. But I always have to make sure I don't commit those extra changes.

**Additional context**

PR #2991, opened 2024-11-27
ci: Add melos and melos lint job for project

Isakdl:
# Changes

- Adds melos to the root of the project
  - With lint command with the 3 options, normal, strict and loose, normal allows info but not warning, loose allows warnings, strict fails on all.
  - With downgrade command to install the oldest dependencies.
- Add lint job with melos across all projects
- Add CI job to run the new melos command on linux and windows machines

Melos also comes with the benefit of the `bootstrap` command e.g. installing dependencies. I think it is a good idea to transition all our commands to melos instead of bash scripts. Much easier to understand and work with and we can make them platform-agnostic which is a huge plus.

But I think we can convert one script at a time rather than bundle them in a big PR.

Related issue: #2990 

## Pre-launch Checklist

## Breaking changes
none

### DIFF
--- a/.github/workflows/dart-tests.yaml
+++ b/.github/workflows/dart-tests.yaml
@@ -50,10 +50,11 @@ jobs:
 
   dart_analyze_latest:
     name: dart analyze latest
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         flutter-version: ['3.27.0']
+        os: [windows-latest, ubuntu-latest]
     steps:
       - uses: actions/checkout@v3
       - uses: subosito/flutter-action@v2
@@ -68,10 +69,11 @@ jobs:
 
   dart_analyze_downgrade:
     name: dart analyze downgrade
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         flutter-version: ['3.19.0', '3.27.0']
+        os: [windows-latest, ubuntu-latest]
     steps:
       - uses: actions/checkout@v3
       - uses: subosito/flutter-action@v2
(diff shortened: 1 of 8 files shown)


.github/workflows/dart-tests.yaml
```diff
@@ -48,37 +48,43 @@ jobs:
...
         with:
           flutter-version: ${{ matrix.flutter-version }}
-      - name: Install dependencies
-        run: bash util/pub_get_all
-      - name: Analyze
-        run: bash util/run_tests_analyze --allow-infos
+      - name: Install melos
+        run: dart pub global activate melos
+      - name: Bootstrap
+        run: melos bootstrap
```
SandPod, line 85: Suggestion: Skip the bootstrap.

We don't need the bootstrap when we do the downgrade right? They should be equivalent. Since it takes a while for us to pull all dependencies we can save some time if we skip it.

SandPod (commented):
This looks great!

Is it safe to remove the `run_tests_analyze` script as well?

SandPod:
This PR also relates to this issue: https://github.com/serverpod/serverpod/issues/785

.github/workflows/dart-tests.yaml
Isakdl: The thing is that bootstrap is doing more than just installing the dependencies. ´dart pub downgrade´ is the same as `dart pub get` and is just the first step of bootstrap. the linking of local dependencies etc are done later.

SandPod: Aaah, you are right. We have hardcoded the linking right now though... But Bootstrap will be required later if we can remove that part.

SandPod (approved):
👍

PR #3004, opened 2024-11-28
refactor: Model load refactors and tests

SandPod:
This PR contains refactors and tests required to support placing models anywhere in the Serverpod project `lib` directory.

## Pre-launch Checklist

## Breaking changes

None - Should not change the behavior of our generator, these changes are added in an upcoming PR.


tools/serverpod_cli/test/integration/util/model_helper/model_file_load_test.dart
```diff
@@ -0,0 +1,823 @@
...
+    serverPackage: 'test_server',
+    dartClientPackage: 'test_client',
+    dartClientDependsOnServiceClient: true,
+    serverPackageDirectoryPathParts: serverPackageDirectoryPathParts,
+    relativeDartClientPackagePathParts: [],
+    modules: [],
+    extraClasses: [],
+    enabledFeatures: [],
+  );
+}
```
Isakdl, line 24: Suggestion: Change to a builder, I believe there should be one.

tools/serverpod_cli/test/integration/util/model_helper/model_file_load_test.dart
```diff
@@ -0,0 +1,823 @@
...
+  setUpAll(() {
+    testDirectory = Directory(join(Directory.current.path, const Uuid().v4()));
+    testDirectory.createSync(recursive: true);
+  });
+
+  tearDownAll(() {
+    testDirectory.deleteSync(recursive: true);
+  });
+
+  group('Given a valid model class with *.yaml file ending', () {
```
Isakdl, line 37: Comment: As discussed offline, it may be okay to loop the different file endings, assuming we are testing the exact same thing (e.g. expect). 

Again we have to be careful with this but as long as it is declarative and we are not introducing ANY logic in the test it might be okay.

SandPod: OK! I'll take this into consideration for these tests.

tools/serverpod_cli/test/integration/util/model_helper/model_file_load_test.dart
SandPod: Fixed! 👍

PR #3011, opened 2024-12-02
ci: Add windows to bootstrap test platforms.

SandPod:
Adds windows as a platform that the bootstrap tests are run on.

Unfortunately Github actions runs docker with windows containers and the postgres does not supply a windows container. Therefore, the tests that start the container are skipped on windows: https://github.com/docker-library/postgres/issues/324

This should still provide us with 90% of the confidence since the only difference between starting the database connection in windows and UNIX is the dart runtime which we are not changing.

We still have a test that validates running the mini project on windows and executing all of the tests.

**Additional fix** 
- Fixes an issue where symlinks would break serverpod version compatibility check. We really only need to do this check one level down so this should not be an issue.

## Pre-launch Checklist

## Breaking changes

None - CI test platform addition.

### DIFF
--- a/.github/workflows/dart-tests.yaml
+++ b/.github/workflows/dart-tests.yaml
@@ -246,7 +246,7 @@ jobs:
       fail-fast: false
       matrix:
         flutter-version: ['3.19.0', '3.24.0']
-        os: [ubuntu-latest]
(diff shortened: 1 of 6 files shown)


christerswahn (approved):
Nice fixes and cleanup! 🤘

PR #3018, opened 2024-12-03
fix: Only assign project model definitions to project models.

SandPod:
Implements a stricter check to ensure that the projectModelDefinition field is only assigned to a model that is part of the project.

If this check is missing, an invalid reference might be generated. This could happened if there was a class name collision with a Serverpod model.

One such example was a project with a `Filter` model, which was used as a field in another model. When the project was loaded, the Serverpod definitions of the `filter`model were loaded first, so it was assigned as the 'projectModelDefinition` value.

This fix should also be applied to the `Stable-2.2` branch.

Created an issue to fix this "potential" issue in the other places as well: https://github.com/serverpod/serverpod/issues/3019

## Pre-launch Checklist

## Breaking changes

None - Simple bugfix.

### DIFF
--- a/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_datatype_test.dart
+++ b/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_datatype_test.dart
@@ -119,6 +119,69 @@ fields:
       });
     });
 
+    group(
+        'Given a class with a field containing a model first defined in a module and then the project (order matters)',
+        () {
+      var containedClassName = 'User';
+      var testClassName = 'Example';
+      var models = [
+        ModelSourceBuilder()
+            .withFileName('user.spy.yaml')
+            .withModuleAlias('module')
+            .withYaml(
+          '''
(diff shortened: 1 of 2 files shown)


tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_datatype_test.dart
```diff
@@ -119,6 +119,69 @@ fields:
       });
     });
 
+    group(
```
hampuslavin, line 122: nit: This test group is missing a `when` action declaration

hampuslavin (approved):

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_datatype_test.dart
SandPod: You are right, but looking at the file it doesn't seem like this is used for any of the tests, so I will opt to keep the pattern here.

PR #3025, opened 2024-12-05
fix: Introduce end of stream message to close stream gracefully.

SandPod:
In a cloud environment, streams sometimes seem to drop the last packages of a stream.

A guess is that this is caused by a data race between the subscription cancel call and the `onData` processing.

This fix mitigates the risk of that data race by introducing a last message that is sent through the stream to ensure that the buffer has been completely processed before cancel is called.

The message is passed as an error to circumvent branching in the "hot path" of the stream.

All the awaits have also been removed from the `listen` callback. This ensures that a new message is not processed before all methods have been called, ensuring that the subscription is canceled correctly if required.

Out tests caught this last edge-case by calling an endpoint that threw an exception. Because of the awaits in the `onError` callback, a new message was processed before the subscription could be canceled, ending the stream with `CloseReason.done` instead of `CloseReason.error`.

### Tests

I tried, but failed, to create a test that reproduces the issue observed in the cloud environment. 
To validate the fix, I would like to release it as a 2.3-beta and observe if the issue is resolved.

## Pre-launch Checklist

## Breaking changes

None - This should fix an issue but the behavior should stay the same.

### DIFF
--- a/packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
+++ b/packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
@@ -351,24 +351,31 @@ class MethodStreamManager {
       await session.close();
     });
 
+    var streamKey = _buildStreamKey(
+      endpoint: methodStreamCallContext.fullEndpointPath,
+      method: methodStreamCallContext.method.name,
+      methodStreamId: methodStreamId,
+    );
+
     late StreamSubscription subscription;
     subscription = outputController.stream.listen(
-      (value) {
+      (value) async {
         _onOutputStreamValue?.call(
             methodStreamId, value, methodStreamCallContext);
       },
-      onError: (e, s) async {
+      onError: (e, s) {
+        if (e is _StreamComplete) {
+          _updateCloseReason(streamKey, CloseReason.done);
+          subscription.cancel();
+          return;
+        }
+
         _onOutputStreamError?.call(
             methodStreamId, e, s, methodStreamCallContext);
 
-        var streamKey = _buildStreamKey(
-          endpoint: methodStreamCallContext.fullEndpointPath,
-          method: methodStreamCallContext.method.name,
-          methodStreamId: methodStreamId,
-        );
         _updateCloseReason(streamKey, CloseReason.error);
 
-        await session.close(error: e, stackTrace: s);
+        session.close(error: e, stackTrace: s);
 
         /// Required to close stream when error occurs.
         /// This will also close the input streams.
@@ -376,7 +383,7 @@ class MethodStreamManager {
         /// for the listen method because this cancels
         /// the stream before the onError callback has
         /// been called.
-        await subscription.cancel();
+        subscription.cancel();
       },
     );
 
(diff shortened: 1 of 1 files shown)


packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
```diff
@@ -571,21 +578,13 @@ class MethodStreamManager {
...
-          _updateCloseReason(streamKey, CloseReason.done);
-        }
-
-        await subscription.cancel();
-      },
-    );
   }
 }
+
+// Passed as the last message on a stream to indicate that the stream is
```
christerswahn, line 588: Suggestion: Write as doc comment with 3 /

packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
```diff
@@ -571,21 +578,13 @@ class MethodStreamManager {
...
 
     outputController.addStream(methodStream).whenComplete(
-      () async {
-        var streamKey = _buildStreamKey(
-          endpoint: methodStreamCallContext.fullEndpointPath,
-          method: methodStreamCallContext.method.name,
-          methodStreamId: methodStreamId,
+          // The stream complete message is sent as an error to circumvent
+          // branching when passing along stream events to the the handler.
+          () => outputController.addError(_StreamComplete()),
```
christerswahn, line 583: At least on the face of it this appears to play more nice with the underlying stream! :+1:

packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
```diff
@@ -351,32 +351,39 @@ class MethodStreamManager {
...
 
-        var streamKey = _buildStreamKey(
-          endpoint: methodStreamCallContext.fullEndpointPath,
-          method: methodStreamCallContext.method.name,
-          methodStreamId: methodStreamId,
-        );
         _updateCloseReason(streamKey, CloseReason.error);
 
-        await session.close(error: e, stackTrace: s);
+        session.close(error: e, stackTrace: s);
```
christerswahn, line 378: Suggestion: Add unawaited to indicate this is intentional, and maybe also add your comment from the PR as code comment here to explain why.

christerswahn (commented):
Looking forward to getting to use this! :) Just had a couple of minor comments!

packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
SandPod: Good catch! will fix!

packages/serverpod/lib/src/server/websocket_request_handlers/helpers/method_stream_manager.dart
SandPod: Good suggestion! Will fix! 👍

christerswahn (approved):
LGTM, snyggt!

Issue #2952, opened 2024-11-12
Quick opening and closing of multiple streams from stream endpoint methods fails


christerswahn:
**Describe the bug**
Quick opening and closing of multiple streams from stream endpoint methods fails with `Unhandled exception: Method stream WebSocket connection closed`.

**To Reproduce**

Assuming there is a stream endpoint method `laika.exercise.pump()`, this client code causes the stream failure:

```
  final pumps = <Stream<String>>[];
  for (int i = 0; i < 10; i++) {
    pumps.add(laika.exercise.pump());
  }
  for (final pump in pumps) {
    await for (final str in pump) {
      print(str);
    }
  }
```

An alternative reproduction, creating multiple streams quickly while keeping them open (just pausing them):

```
  final pumps = <Stream<String>>[];
  for (int i = 0; i < load; i++) {
    pumps.add(laika.exercise.pump());
    // For comparison, when instead creating the following local stream, it works:
    // pumps.add(Stream.fromIterable(List.generate(
    //   pumpLength,
    //   (final i) => '$i',
    // )));
  }
  for (final pump in pumps) {
    StreamSubscription<String>? sub;
    var count = 0;
    sub = pump.listen(
      (final String str) {
        print(str);
        if (++count >= 2 && sub != null) {
          sub.pause();
        }
      },
    );
  }
```

**Expected behavior**

The streams should succeed normally. Clients should be able to open and close multiple streams in quick succession.

**Serverpod versions**

Serverpod version: 2.2.0-beta.1

**Platform information**

MacOS, bash, simple CLI client towards Serverpod server.

hampuslavin:
Could possibly be related to this previous issue that I created: #2818

PR #3029, opened 2024-12-06
fix: Remove server closing of websocket connections.

SandPod:
Closes: #2952

The Serverpod streaming protocol supported closing the connection from the Server side if no open method stram connections were detected. However, this made it impossible for the client to synchronize its state since a connection might have been established and closed from the Server while the client was waiting for input.

This PR removes the server's possibility to close the WebSocket connections and makes the client authoritative around the lifecycle of WebSockets.

Note that the server can still close method streaming connections, for example, when a streaming endpoint has returned. It is simply not allowed to close the WebSocket connection as a result of this.

#### Test fixes

Some tests were removed as they validated the server's possibility of closing WebSocket connections. Other tests used the closing of the WebSocket connection to indicate that the streaming endpoint had returned.

These tests are now modified to rely on the session object or a StreamController, depending on what suited the test case best,  in the endpoint to provide that validation.

## Pre-launch Checklist

## Breaking changes

None - Client always closed WebSocket connections as well. Allowing the server to also close the connection was simply redundant.


tests/serverpod_test_server/test_e2e/method_streaming/open_stream_test.dart
```diff
@@ -0,0 +1,21 @@
...
+      'when calling an endpoint multiple times in quick succession '
+      'then all streams complete successfully', () async {
+    var client = Client(serverUrl);
+    var numMessages = 1000;
+    var stream;
+    for (var i = 0; i < 10; i++) {
+      stream = client.methodStreaming.intStreamFromValue(numMessages);
+    }
+
+    await expectLater(stream.last, completes);
```
christerswahn, line 19: Suggestion: Verify all the streams complete normally with emitsDone matcher

tests/serverpod_test_server/lib/src/endpoints/method_streaming.dart
```diff
@@ -14,15 +14,19 @@ class MethodStreaming extends Endpoint {
...
     Session session,
     int millisecondsDelay,
-  ) async* {
-    int i = 0;
-    while (true) {
-      await Future.delayed(Duration(milliseconds: millisecondsDelay));
-      yield i++;
-    }
+  ) {
+    var controller = StreamController<int>();
```
christerswahn, line 22: Question: Is the controller "middle man" needed here to "keep it alive", as opposed to just returning result of Stream.periodic()?

tests/serverpod_test_server/test_e2e/method_streaming/open_stream_test.dart
```diff
@@ -0,0 +1,21 @@
...
+  });
+}
```
christerswahn, line 21: Suggestion: Also add a test case where a bunch of streams from the same endpoint method are opened, and then read, paused, and resumed, in order to check pause-resume behavior with multiple streams open. (Unless that exists elsewhere already?)

tests/serverpod_test_server/test_e2e/method_streaming/open_stream_test.dart
```diff
@@ -0,0 +1,21 @@
+@Timeout(Duration(minutes: 5))
```
hampuslavin, line 1: Question: How come this needs a 5 minute timeout?

christerswahn (commented):
Looking forward to getting this in Ground Control! 😄

tests/serverpod_test_server/lib/src/endpoints/method_streaming.dart
hampuslavin: This is to enable the test to hook into the stream's lifecycle and assert that it was indeed cancelled in the end.

hampuslavin (commented):
Nothing to add, just a small question 👍

tests/serverpod_test_server/test_e2e/method_streaming/open_stream_test.dart
SandPod: Good suggestion! Will fix! 👍

tests/serverpod_test_server/test_e2e/method_streaming/open_stream_test.dart
SandPod: This is not required, will remove it!

tests/serverpod_test_server/test_e2e/method_streaming/open_stream_test.dart
SandPod: Good suggestion, added! 👍

hampuslavin (approved):
🏎️

PR #3047, opened 2024-12-11
fix: Support typed database exceptions.

SandPod:
Introduces a new database exception type called `DatabaseQueryException` that contains information related to an exception that occurred during a database query.

The fields exposed through the `DatabaseQueryException` class have been minimized to contain what is most relevant to the user.

**Please provide feedback if any important field has been left out.**

The `code` field, which is probably what most developers will add code to parse, would have been nice to have typed. But since codes are not shared between different database types, the field is simply a nullable String.

This PR supersedes the original PR introduced here: https://github.com/serverpod/serverpod/pull/2546

### Database exception abstract interfaces
New interfaces have been added that allow the framework to extend the interfaces in a backward-compatible way.

By using abstract interface class we reserve the right to extend the interface in the future and prevent code outside of the library to implement it.

An implementation of the interface has been added to the Postgres adapter. This continues to set the foundation for allowing other adapters in the future.

### Additional fixes
Where applicable, all legacy tests were converted to be more precise in their type checking.

The last commit also contains a minor refactor to flatten out the if statements when catching exceptions in a database query.

## Pre-launch Checklist

## Breaking changes

None - the intention behind these changes is to be implemented in a backwards-compatible way.
New subtypes in the adapter have been added, but these should not interfere with logic that catches current exceptions.


SandPod:
@lukehutch - Please have a look at this to ensure everything you need is included.

packages/serverpod/lib/src/database/concepts/exceptions.dart
```diff
@@ -0,0 +1,44 @@
...
+  String? get detail;
+
+  /// A hint on how to remedy an error, if provided by the database.
+  String? get hint;
+
+  /// Returns the name of the table where the error occurred.
+  String? get tableName;
+
+  /// Returns the name of the column where the error occurred.
+  String? get columnName;
```
hampuslavin, line 25: Question: Should we make a distinction on that this class refers to relational databases and not document based one? If not, how do we see that potentially fit in in the future?

tests/serverpod_test_server/test_integration/database_operations/crud/create_test.dart
```diff
@@ -43,7 +43,8 @@ void main() async {
 
       expect(
         UniqueData.db.insert(session, data),
-        throwsA(isA<DatabaseException>()),
+        throwsA(isA<DatabaseQueryException>()
+            .having((e) => e.code, 'code', '23505' /* Unique violation */)),
```
hampuslavin, line 47: Question: Would it make sense to have known [PG exception codes ](https://www.postgresql.org/docs/current/errcodes-appendix.html) defined in an abstract final class?

hampuslavin (commented):
Looks good! 🏎️  Just some Qs for my understanding.

packages/serverpod/lib/src/database/concepts/exceptions.dart
SandPod: Hmm, good question.

We could take a pragmatic approach here and treat the introduction of a document based database as a breaking change in the future? These values are nullable so if they are not used it should be fine.

This also kind of makes sense since our Expression queries are pretty much tailored toward a relational databases.

tests/serverpod_test_server/test_integration/database_operations/crud/create_test.dart
SandPod: Yes, it does... I took a shortcut here.

Let me introduce them as statics.

packages/serverpod/lib/src/database/exceptions.dart
```diff
@@ -1,10 +1,84 @@
...
+        trace = null,
+        schemaName = null,
+        tableName = null,
+        columnName = null,
+        dataTypeName = null,
+        constraintName = null,
+        fileName = null,
+        routineName = null;
+
+  /// Creates a new [DatabaseException] from a [PgException].
```
christerswahn: Suggestion, change PgException in doc comment to ServerException.

packages/serverpod/lib/src/database/concepts/exceptions.dart
```diff
@@ -0,0 +1,44 @@
...
+
+  /// Returns the name of the constraint that was violated.
+  String? get constraintName;
+
+  /// Returns the position in the query where the error occurred.
+  int? get position;
+}
+
+/// Exception thrown when an insert row operation fails.
+abstract interface class DatabaseInsertRowException
```
christerswahn, line 50: Discussion: Should we really have different exceptions for different operation types? The exception typically represents the kind of error, not the kind of operation that the error prevented. 

For example, the duplicate value error can happen both on update and insert operations. No such key error can happen both on select, update, and delete operations.

I would propose to have exception subclasses for categories of errors - if we need them at all at this point. Typical categories could then be for example constraint violation (including duplicate and null-for-non-nullable field) and no-such-record (when refering to a specific record id in select, update, or delete).

tests/serverpod_test_server/test_integration/database_operations/crud/create_test.dart
christerswahn: This would be super! Suggestion to see if it makes sense to enumerate some categories as well (for example constraint violation). This is just a thought, not sure if it is worth while!

christerswahn (commented):
This is a really great "maturity" improvement! :) Left some comments and discussion point!

packages/serverpod/lib/src/database/concepts/exceptions.dart
hampuslavin: I agree I think that makes sense!

packages/serverpod/lib/src/database/concepts/exceptions.dart
hampuslavin: Good points here, I was thinking about this as well.

tests/serverpod_test_server/test_integration/database_operations/crud/create_test.dart
SandPod: They don't seem to be separated in error code sense, so I opted for representing the table as it is.

Since this boils down to a string comparison that is different for MySQL, I opted to keep the solution simple.

lukehutch:
> @lukehutch - Please have a look at this to ensure everything you need is included.

LGTM, other than the open comments.

packages/serverpod/lib/src/database/exceptions.dart
SandPod: This file has been removed, but I think that the comment is still valid! Will double check it

SandPod: No, this was completely removed so I'll resolve it.

hampuslavin (approved):
Looks good on my end 👍

packages/serverpod/lib/src/database/concepts/exceptions.dart
christerswahn: IMO this principle is valid for our internal exceptions as well. The example you linked to perhaps shows need for having exception types for error categories rather than operation categories. In that example, there is a significant difference between the number of affected rows being zero, meaning "no such row / id found", and greater than 1, meaning ambiguous/wrong filter input. Both are presumably "caller issues" rather than DB insert issues?

tests/serverpod_test_server/test_integration/database_operations/crud/create_test.dart
christerswahn: The PG codes are hierarchical, with the first 2 digits being the class, e.g:
Class 23 — Integrity Constraint Violation

These classes could conceivably be maintained across DBs to some extent. But not essential right now! 👍

packages/serverpod/lib/src/database/concepts/exceptions.dart
SandPod: Discussed IRL, the principle of the argument is correct here. These exceptions should probably have been defined on a different level. Unfortunately, this would be a breaking change to modify, so we can't really fix it at this point.

tests/serverpod_test_server/test_integration/database_operations/crud/create_test.dart
SandPod: Aah yes, that is correct! We will opt for the simple solution for now.

christerswahn (approved):
LGTM!

PR #3065, opened 2024-12-18
fix: Improve request body parsing.

LeonidVeremchuk:
## Summary of Changes
- **Added** a faster and more efficient solution for byte data handling.
- **Improved** the handling of request size with a correct and optimized approach.

---

## Pre-launch Checklist
- [ ] Verify that all unit tests pass successfully.
- [ ] Ensure that request size validation functions as expected.
- [ ] Confirm that performance benchmarks show improvement with the updated implementation.

---

## Original Issue
The original implementation used a `List<int>` (`data`) to accumulate incoming byte segments. The `+=` operator creates a new list every time data is appended. This approach is inefficient for handling large amounts of data, leading to excessive memory allocations and unnecessary copying of existing data.

---

## Optimized Solution
The updated code replaces the list with a `BytesBuilder`, a utility specifically designed for efficient byte accumulation. This ensures a more memory- and performance-efficient implementation.

---

### Key Improvements
1. **Efficient Byte Accumulation:**
   - `BytesBuilder` dynamically grows its internal buffer, avoiding the need to create a new object each time data is appended.
   - Reduces memory usage and improves performance significantly.

2. **Avoiding Redundant Memory Allocation:**
   - Previously, using `+=` created a new list and copied all existing data into it. `BytesBuilder` manages memory internally, preventing these redundant operations.

3. **Efficient Data Conversion:**
   - The `BytesBuilder.takeBytes()` method retrieves the final byte array without additional copying, enabling more efficient conversion to a UTF-8 string.

4. **Preserving Size Check Logic:**
   - The logic for checking the request size (`len > serverpod.config.maxRequestSize`) is retained.
   - This ensures oversized requests are detected early and stopped before unnecessary data accumulation.

---

### Benefits of the Optimization
- **Improved Performance:** 
  - `BytesBuilder` is designed for scenarios requiring dynamic byte accumulation, reducing time complexity and memory overhead.
- **Cleaner Code:**
  - The optimized implementation is easier to read and understand, as `BytesBuilder` is semantically appropriate for this use case.
- **Scalability:** 
  - The solution is more suitable for handling larger data streams without performance degradation.

---

This release improves the efficiency, maintainability, and scalability of handling HTTP request body data.


LeonidVeremchuk:

SandPod (commented):
Thank you for the great writeup and implementation 🌟 

This is a great contribution.

Could you add a test that validates that the max request size message is propagated to the client? It is enough to validate this on an endpoint basis. A suggestion is to create a test where the `maxRequestSize` is set really low (1), and then make a normal RPC request to the server with some data.

You can take inspiration from this test to have both the server and client in the same test: https://github.com/serverpod/serverpod/blob/main/tests/serverpod_test_server/test_integration/logging/override_logging_test.dart

Then you could configure the server by passing in a configuration to the `IntegrationTestServer.create()` method.

No test is needed for the web_server as we are about to update that implementation soon.

packages/serverpod/lib/src/server/server.dart
```diff
@@ -402,15 +402,17 @@ class Server {
   }
 
   Future<String?> _readBody(HttpRequest request) async {
```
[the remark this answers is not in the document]
LeonidVeremchuk: good point. I'll fix it

LeonidVeremchuk:
# Update Summary

## Changes and Fixes

1. **Added a new `ResultRequestTooLarge` class**:
   - Implemented a class to handle errors related to request body size exceeding the maximum limit.
   - Provides detailed information about the maximum allowed size and the actual size of the request.

2. **Enhanced Error Handling**:
   - Improved handling for `ResultRequestTooLarge` exceptions.
   - Clear and informative error messages are now sent to users when their request size exceeds the allowed limit.

3. **Optimized `_readBody` Function**:
   - Simplified logic for improved readability and performance.
   - Resolved issues with redundant checks and stream reuse, ensuring more reliable behavior.

4. **Added Comprehensive Tests**:
   - Created tests to verify the system's behavior when the request size exceeds the configured limit.
   - Tests validate both the returned HTTP status code and the error message sent to the user.

## Benefits
- Improved system stability.
- Enhanced code readability and maintainability.
- Better user experience with clear feedback on errors.

If you have any questions or suggestions, feel free to reach out.

Thank you! 🚀

packages/serverpod/lib/src/server/server.dart
```diff
@@ -401,16 +410,16 @@ class Server {
     );
   }
 
-  Future<String?> _readBody(HttpRequest request) async {
-    // TODO: Find more efficient solution?
-    var len = 0;
-    var data = <int>[];
+  Future<String> _readBody(HttpRequest request) async {
+    var builder = BytesBuilder();
+    var len = request.headers.contentLength;
```
SandPod, line 415: Suggestion: If transfer encoding chunked is used, the contentLength header will have the value -1.

I will revert this to the old implementatation to ensure consistency regardless of if the contentLenght header is set.

packages/serverpod/lib/src/server/endpoint_dispatch.dart
```diff
@@ -629,3 +629,47 @@ class ExceptionResult<T extends SerializableException> extends Result {
...
+/// ```dart
+/// var maxSize = serverpod.config.maxRequestSize;
+/// if (requestSize > maxSize) {
+///   throw ResultRequestTooLarge(maxSize, requestSize);
+/// }
+/// ```
+///
+/// This error provides details about the maximum allowed size and the actual
+/// size of the request, allowing the client to adjust their request accordingly.
+class ResultRequestTooLarge extends Result {
```
SandPod, line 650: Suggestion: Implement this as an exception instead of a result.

This result class is used specifically when method call. All the details here are great but it should be implemented as an exception.

tests/serverpod_test_server/test_integration/request/result_request_too_large_test.dart
```diff
@@ -0,0 +1,15 @@
...
+    test('should create error with correct description', () {
+      final error = ResultRequestTooLarge(1024, 2048);
+
+      expect(error.maxSize, 1024);
+      expect(error.actualSize, 2048);
+      expect(error.toString(),
+          'Request size exceeds the maximum allowed size of 1024 bytes. Actual size: 2048 bytes.');
+    });
+  });
+}
```
SandPod, line 15: Suggestion: This test is not specifically required, we can validate this as part of the other `read_body_test.dart`.

SandPod: For maintainability reasons we the number of tests that fail when a change is made to the project.

SandPod (commented):
Thank you for the added tests.

I'll implement the last suggested changes here to get these in. 

I added the comments to highlight what will be modified.

SandPod (approved):
LGTM! 👍 🚀 

Thank you 🙏

Issue #3082, opened 2024-12-25
Wrong generated constructor for models that inherit from classes that have fields with default values


marcelomendoncasoares:
**Describe the bug**
When using the model inheritance experimental feature, there is a bug if any of the base class fields have default values. The generated constructor tries to initialize the default fields as if they belong to the class itself instead of delegating to super.

**To Reproduce**
1. Create a `base.spy.yaml` file with the contents:
```yaml
class: BaseModel
fields:
  noDefaultField: String  # This field works as expected.
  uid: UuidValue, default=random
  localId: int, default=0
  updatedAt: DateTime, default=now
```

2. Create another model file `model.spy.yaml` that extends this model, such as:
```yaml
class: Model
extends: BaseModel
fields:
  anyOtherField: int
  fieldWithDefault: int, default=0
```

3. Run `serverpod generate --experimental-features=all` to generate the models.
4. Go to the `<package>_client/lib/src/protocol/model.dart` to see the error on the generated class. The error will be the same for all fields that have a default value:
> `'<...>' isn't a field in the enclosing class. Try correcting the name to match an existing field, or defining a field named '<...>'.`

The class will be generated like the following (comments added by me):
```dart
// ignore_for_file: no_leading_underscores_for_library_prefixes
import 'protocol.dart' as _i1;
import 'package:serverpod_client/serverpod_client.dart' as _i2;
import 'package:uuid/uuid.dart' as _i3;

abstract class Model extends _i1.BaseModel implements _i2.SerializableModel {
  Model._({
    required super.noDefaultField,  // Correct initialization of super field.
    _i2.UuidValue? uid,
    int? localId,
    DateTime? updatedAt,
    required this.anyOtherField,
    int? fieldWithDefault,
  })  : uid = uid ?? _i3.Uuid().v4obj(),  // <-- Error on this initialization sequence, as the fields come from the base class.
        localId = localId ?? 0,
        updatedAt = updatedAt ?? DateTime.now(),
        fieldWithDefault = fieldWithDefault ?? 0;  // This line is correct, as the field belong to the class.
```

**Expected behavior**
The generated class should delegate the initialization of the base class values to super, but without the `required` keyword. It is already correct for required fields, as shown for the `noDefaultField`:
```dart
// ignore_for_file: no_leading_underscores_for_library_prefixes
import 'protocol.dart' as _i1;
import 'package:serverpod_client/serverpod_client.dart' as _i2;

[...]

PR #3089, opened 2024-12-31
fix: [Inheritance] Invalid code generation when base class have parameters with default values

marcelomendoncasoares:
This PR closes the issue #3082, where the problem is described in detail. As a brief summary, when using the model inheritance experimental feature, there is a bug if any of the base class fields have default values. The generated constructor tries to initialize the default fields as if they belong to the class itself instead of delegating to super.

## Pre-launch Checklist

## Breaking changes

No breaking changes, only the fix for a case not covered on the tests.

### DIFF
--- a/tools/serverpod_cli/test/generator/dart/client_code_generator/inheritance_class_test.dart
+++ b/tools/serverpod_cli/test/generator/dart/client_code_generator/inheritance_class_test.dart
@@ -35,19 +35,21 @@ void main() {
   );
 
   group(
-      'Given a child-class named $childClassName with one primitive var extending a parent-class named $parentClassName with one primitive var when generating code',
+      'Given a child-class named $childClassName with one primitive var and a var with default value extending a parent-class named $parentClassName with one primitive var and a var with default value when generating code',
       () {
     var models = [
       ClassDefinitionBuilder()
           .withClassName(parentClassName)
           .withFileName(parentClassFileName)
           .withSimpleField('name', 'String')
+          .withSimpleField('parentDefault', 'int', defaultValue: '0')
           .withChildClasses(
         [
           ClassDefinitionBuilder()
               .withClassName(childClassName)
               .withFileName(childClassFileName)
               .withSimpleField('age', 'int')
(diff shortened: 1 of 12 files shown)


tools/serverpod_cli/test/generator/dart/client_code_generator/inheritance_class_test.dart
```diff
@@ -35,31 +35,35 @@ void main() {
   );
 
   group(
-      'Given a child-class named $childClassName with one primitive var extending a parent-class named $parentClassName with one primitive var when generating code',
+      'Given a child-class named $childClassName with one primitive var and a var with default value extending a parent-class named $parentClassName with one primitive var and a var with default value when generating code',
```
SandPod, line 38: Suggestion: Break this test out into a separate test and add a similar test for the `server_code_generator`.

We try to isolate each test so that there is a single reason for failure. We believe this will guide future developers when implementing new features. Please break the test and setup validating the changes in this PR out into a separate group to make the setup as simple as possbile to validate the test scenario.

We also want to validate that this is working correctly for both the server and client generator. Regardless of how we refactor the code.

SandPod (changes requested):
Awesome bug report and nice and isolated fix.

Apart from the comment on the tests, could you please add a model in the test project that validates this scenario? We generally try to have each feature of our models generated once in the project to catch edge cases.

tools/serverpod_cli/test/generator/dart/client_code_generator/inheritance_class_test.dart
marcelomendoncasoares: Done! Following the hint from the other PR, I also simplified the test to remove validations that would become redundant when separating the tests.

marcelomendoncasoares:
> Awesome bug report and nice and isolated fix.
> 
> Apart from the comment on the tests, could you please add a model in the test project that validates this scenario? We generally try to have each feature of our models generated once in the project to catch edge cases.

No problem. Just added two models (the parent and the child) with the exact same configuration from the server and client tests.

Issue #1557, opened 2023-11-13
Prints from serverpod_client.


SandPod:
**Describe the bug**
When exceptions happen in the serverpod client, sometimes these exceptions are printed.

**To Reproduce**
1. Call an endpoint from the client when the server is not running and observer that the exception is printed.

Examples of this can be observed in [serverpod_client_browser.dart](https://github.com/serverpod/serverpod/blob/534a76241c4349780b8cb39b86914e67784d4803/packages/serverpod_client/lib/src/serverpod_client_browser.dart#L73) and [serverpod_client_io.dart](https://github.com/serverpod/serverpod/blob/534a76241c4349780b8cb39b86914e67784d4803/packages/serverpod_client/lib/src/serverpod_client_io.dart#L90). 

**Expected behavior**
The client should only throw exceptions and leave it to the user to determine if and how these should be printed.

**Serverpod versions**
Pre 1.2 release.

hampuslavin:
Additional context: One part of the problem here is this part in the `client_shared` file: 
https://github.com/serverpod/serverpod/blob/7e9aa21845c7cb22aa7f74e91c472beb2548e5d7/packages/serverpod_client/lib/src/serverpod_client_shared.dart#L502-L507

While we should also consider if this makes sense at all, we should at least expose the `logFailedCalls` flag in the generated client. Currently it is not exposed and cannot be turned off.

PR #3117, opened 2025-01-13
fix: remove debug print messages for failed client calls

hampuslavin:
## Description
- Removes debug prints for failed client calls, as it doesn't make any sense.
- Closes https://github.com/serverpod/serverpod/issues/1557.

## Pre-launch Checklist

### DIFF
--- a/packages/serverpod_client/lib/src/serverpod_client_shared.dart
+++ b/packages/serverpod_client/lib/src/serverpod_client_shared.dart
@@ -499,12 +499,6 @@ abstract class ServerpodClientShared extends EndpointCaller {
     } catch (e, s) {
       onFailedCall?.call(callContext, e, s);
 
-      if (logFailedCalls) {
-        // ignore: avoid_print
-        print('Failed call: $endpoint.$method');
-        // ignore: avoid_print
-        print('$e');
-      }
       rethrow;
     }
   }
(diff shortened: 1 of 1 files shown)


packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -499,12 +499,6 @@ abstract class ServerpodClientShared extends EndpointCaller {
     } catch (e, s) {
       onFailedCall?.call(callContext, e, s);
 
-      if (logFailedCalls) {
```
SandPod, line 502: Question: Isn't it safe to remove the `logFailedCalls` constructor variable as well?

SandPod (commented):
Nice to clean this up, single question!

packages/serverpod_client/lib/src/serverpod_client_shared.dart
hampuslavin: Taking pragmatic approach and yanking this member variable as there is no use-case for it: 3d4b1d4

SandPod: I think it is fine to remove it completely. This had practically no function for any user.

SandPod (approved):
LGTM! 🙌

PR #3123, opened 2025-01-14
fix: Add `moduleAlias` checks for accurate class resolution

klkucaj:
###  Description:
This PR reviews all the places where class resolution is used. Added `moduleAlias` checks where needed to ensure correct behavior and left unchanged the places where the existing logic was already sufficient. Below is a breakdown of what was updated and what was not.

### 1. **Inheritance Resolution**
- **Unchanged**  
  The existing code already works correctly by checking only the `className`. Since we dont allow inheritance across modules, no changes were needed.
https://github.com/serverpod/serverpod/blob/0009d80b52417669e80dd0a728fc00dae409bcea/tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart#L47

### 2. **Enum Resolution**
- **Modified**  
  We added `moduleAlias` checks because enums can have the same `className` in different modules. Both `className` and `moduleAlias` are now used to uniquely identify enums.
 https://github.com/serverpod/serverpod/blob/8ba5561eef83cce6af96587a91f85f199ce3c414/tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart#L96-L97

### 3. **Object Relation Resolution**
- **Modified**  
  We added `moduleAlias` checks to ensure object relations are resolved correctly when there are classes with the same `className` in different modules.
 https://github.com/serverpod/serverpod/blob/8ba5561eef83cce6af96587a91f85f199ce3c414/tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart#L116-L117

### 4. **List Relation Resolution**
- **Unchanged**  
  We dont allow list relations on modules. We already have this check:
https://github.com/serverpod/serverpod/blob/e1fae905d1fdba2d72e1920da20100baa23b96a3/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart#L1057-L1065
https://github.com/serverpod/serverpod/blob/0009d80b52417669e80dd0a728fc00dae409bcea/tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart#L114

### 5. **Extra Class Restrictions**
- **Unchanged**  
  This code only checks if a `className` exists in `extraClasses`. Since no module ambiguity exists here, it didn’t need any changes.
https://github.com/serverpod/serverpod/blob/0009d80b52417669e80dd0a728fc00dae409bcea/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart#L172

### 6. **Custom Type Check**
- **Unchanged**  
  This function only checks if a type is in `extraClasses` by `className`. The existing logic was already correct, so no changes were made.
https://github.com/serverpod/serverpod/blob/0009d80b52417669e80dd0a728fc00dae409bcea/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart#L172

### 7. **Type Parsing for Extras**
- **Unchanged**  
  This part determines if a `className` exists in `extraClasses`. No changes were needed because the logic is simple and sufficient.
https://github.com/serverpod/serverpod/blob/0009d80b52417669e80dd0a728fc00dae409bcea/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart#L1415

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/model_name_conflict_test.dart
+++ b/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/model_name_conflict_test.dart
@@ -1,0 +1,116 @@
+import 'package:serverpod_cli/src/analyzer/models/definitions.dart';
+import 'package:serverpod_cli/src/analyzer/models/stateful_analyzer.dart';
+import 'package:serverpod_cli/src/generator/code_generation_collector.dart';
+import 'package:serverpod_cli/src/test_util/builders/generator_config_builder.dart';
+import 'package:serverpod_cli/src/test_util/builders/model_source_builder.dart';
+import 'package:test/test.dart';
+
+void main() {
+  var config = GeneratorConfigBuilder().build();
+
+  test(
+      'Given multiple enums with the same name in different modules '
+      'when an enum is referenced '
+      'then the enum is resolved to the correct module', () {
+    var commonEnumName = 'CommonEnum';
+    var firstModuleAlias = 'module1';
+    var secondModuleAlias = 'module2';
+    var modelSources = [
+      ModelSourceBuilder()
+          .withYaml(
+            '''
+        enum: $commonEnumName
+        values:
+          - value1
+          - value2
+        ''',
+          )
+          .withFileName('common_enum.yaml')
+          .withModuleAlias(firstModuleAlias)
+          .build(),
+      ModelSourceBuilder()
+          .withYaml(
+            '''
+        enum: $commonEnumName
+        values:
+          - value1
+          - value2
+        ''',
+          )
+          .withFileName('common_enum.yaml')
+          .withModuleAlias(secondModuleAlias)
+          .build(),
+      ModelSourceBuilder().withYaml(
+        '''
+        class: Example
+        fields:
+          enumField: module:$secondModuleAlias:$commonEnumName
+        ''',
+      ).build(),
+    ];
+
+    var collector = CodeGenerationCollector();
+    var models =
+        StatefulAnalyzer(config, modelSources, onErrorsCollector(collector))
+            .validateAll();
+
+    var model = models.last as ClassDefinition;
+    var enumField = model.fields.first;
+
+    expect(enumField.type.enumDefinition, isA<EnumDefinition>());
+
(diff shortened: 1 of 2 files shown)


tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/model_name_conflict_test.dart
```diff
@@ -0,0 +1,116 @@
...
+import 'package:serverpod_cli/src/generator/code_generation_collector.dart';
+import 'package:serverpod_cli/src/test_util/builders/generator_config_builder.dart';
+import 'package:serverpod_cli/src/test_util/builders/model_source_builder.dart';
+import 'package:test/test.dart';
+
+void main() {
+  var config = GeneratorConfigBuilder().build();
+
+  test(
+      'Given multiple enums with the same name in different modules '
```
SandPod, line 12: Suggestion: Rename to `Given an enum with the same name first define in a module and then the project (order matters) when that enum is referenced then the type is resolved to the project enum` and update the test accordingly.

I think we could have both of these tests to validate that we can reference different modules as well. But they should both mention that order matters since that is what introduced the bug in the first place.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/model_name_conflict_test.dart
```diff
@@ -0,0 +1,116 @@
...
+    var model = models.last as ClassDefinition;
+    var enumField = model.fields.first;
+
+    expect(enumField.type.enumDefinition, isA<EnumDefinition>());
+
+    expect(enumField.type.enumDefinition?.moduleAlias, secondModuleAlias);
+  });
+
+  test(
+      'Given multiple classes with the same name in different modules '
```
SandPod, line 66: Suggestion: Same comment as above, please mention that order matters and test it for the "protocol" case as well when the project model is defined after a module model.

SandPod (commented):
Implementation and tests are on point 🌟 

Some suggestions for a different test scenario that I think most of our users have encountered. And to ensure that order is integrated into the test since that is what made the bug surface in the first place.

SandPod (approved):
LGTM! 🎉 

Thanks for the fix 🙏

Issue #3145, opened 2025-01-20
Wrong relative import path for sealed classes one level below protocol file on Windows


marcelomendoncasoares:
### Describe the bug

On Windows, when generating models for a child sealed class that is located on a directory one level below the `protocol.dart` generated file, the relative import is generated with backslash, which generates a [`uri_does_not_exist`](https://dart.dev/diagnostics/uri_does_not_exist) error:

```dart
import '..\protocol.dart' as _i1;
```

> Target of URI doesn't exist: '..protocol.dart'.
> Try creating the file referenced by the URI, or try using a URI for a file that does exist.

This same model arrangement exists for the models of the `serverpod_test_server`, where the test sealed classes are declared on `lib/src/models/inheritance`. The error occurs if generating the models for this project on Windows.

### To reproduce

Simply run the generate command on the `serverpod_test_server` on Windows. Or, to reproduce elsewhere:

1. Create a directory inside the `models` folder of a serverpod project.
```bash
mkdir -p lib/src/models/example
```

2. Inside the directory, create a `sealed_parent.spy.yaml` file with the contents:
```yaml
class: SealedParent
sealed: true
fields: 
  sealedInt: int
  sealedString: String
```

3. Also in the same directory, create a `sealed_child.spy.yaml` file with the contents:
```yaml
class: SealedChild
extends: SealedParent
fields:
  nullableInt: int?
```

4. Run `serverpod generate --experimental-features=all` to generate the models.

5. Go to any generated class, either on server or client, to see the error on the line 12:
```dart
import '..\protocol.dart' as _i1;
```

### Expected behavior

The import path should be generated with forward slash:

```dart
import '../protocol.dart' as _i1;
```

Manually fixing the slash solves the error.

### Library version

- _CLI version_: 2.3.1.
- _Library version_: 2.3.1.

### Platform information

Running the generation code from Windows.

### How experienced are you with this library?

Intermediate - Familiar with the basics or have used it in a few projects

### Are you interested in working on a PR for this?

- [x] I want to work on this

Isakdl:
Fix is merged

PR #3147, opened 2025-01-20
fix: Relative imports sometimes generated with backward slashes on Windows

marcelomendoncasoares:
Closes issue #3145.

## Pre-launch Checklist

## Breaking changes

No breaking changes.

### DIFF
--- a/tools/serverpod_cli/lib/src/generator/dart/library_generators/util/custom_allocators.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/library_generators/util/custom_allocators.dart
@@ -14,10 +14,11 @@ class ImportCollector {
     }
 
     var absolutePath = p.join(p.dirname(currentPath), topNodePath);
-    return p.relative(
-      absolutePath,
-      from: p.dirname(basePath),
-    );
+    var relativePath = p.relative(absolutePath, from: p.dirname(basePath));
+
+    // If on Windows, top level paths could appear with backslashes and break
+    // the import clause, such as `import '..\protocol.dart' as _i1;`.
+    return relativePath.replaceAll('\\', '/');
   }
 
   int getOrCreateAlias(String topNodePath, String currentPath) {
--- a/tools/serverpod_cli/lib/src/generator/dart/library_generators/util/custom_allocators.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/library_generators/util/custom_allocators.dart
@@ -18,7 +18,7 @@ class ImportCollector {
 
     // If on Windows, top level paths could appear with backslashes and break
     // the import clause, such as `import '..\protocol.dart' as _i1;`.
-    return relativePath.replaceAll('\\', '/');
+    return p.split(relativePath).join('/');
   }
 
   int getOrCreateAlias(String topNodePath, String currentPath) {
--- a/tools/serverpod_cli/test/generator/dart/client_code_generator/imports_test.dart
+++ b/tools/serverpod_cli/test/generator/dart/client_code_generator/imports_test.dart
@@ -1,0 +1,94 @@
+import 'package:analyzer/dart/analysis/utilities.dart';
+import 'package:path/path.dart' as p;
+import 'package:serverpod_cli/src/analyzer/models/definitions.dart';
+import 'package:serverpod_cli/src/generator/dart/client_code_generator.dart';
+import 'package:serverpod_cli/src/test_util/builders/class_definition_builder.dart';
+import 'package:serverpod_cli/src/test_util/builders/generator_config_builder.dart';
+import 'package:serverpod_cli/src/test_util/builders/serializable_entity_field_definition_builder.dart';
+import 'package:serverpod_cli/src/test_util/builders/type_definition_builder.dart';
+import 'package:serverpod_cli/src/test_util/compilation_unit_helpers.dart';
+import 'package:serverpod_cli/src/util/model_helper.dart';
+import 'package:test/test.dart';
+
+const projectName = 'example_project';
+final config = GeneratorConfigBuilder().withName(projectName).build();
+const generator = DartClientCodeGenerator();
+
+void main() {
+  String getExpectedFilePath(String fileName, {List<String>? subDirParts}) =>
+      p.joinAll([
+        '..',
+        'example_project_client',
+        'lib',
+        'src',
+        'protocol',
+        ...?subDirParts,
+        '$fileName.dart',
+      ]);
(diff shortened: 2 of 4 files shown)


tools/serverpod_cli/lib/src/generator/dart/library_generators/util/custom_allocators.dart
```diff
@@ -14,10 +14,11 @@ class ImportCollector {
...
     var absolutePath = p.join(p.dirname(currentPath), topNodePath);
-    return p.relative(
-      absolutePath,
-      from: p.dirname(basePath),
-    );
+    var relativePath = p.relative(absolutePath, from: p.dirname(basePath));
+
+    // If on Windows, top level paths could appear with backslashes and break
+    // the import clause, such as `import '..\protocol.dart' as _i1;`.
+    return relativePath.replaceAll('\\', '/');
```
Isakdl, line 21: Suggestion: The above code will fail if a file is named with '\' with in the file name, change this line to the following instead:

```dart
return p.split(relativePath).join('/');
```

Isakdl (changes requested):
Thank you for the PR! 

I have made a suggestion to make your solution a little more robust, I would also like to add two tests for this case.

Add a tests in this file: https://github.com/serverpod/serverpod/blob/fc8e4e2a26e7301ae45ec07b98f3b1f7f7993c82/tools/serverpod_cli/test/generator/dart/server_code_generator/sealed_class_test.dart#L18

And copy the test for the client as well, in the sibling folder `client_code_generator`. Currently there is no sealed_class_test.dart file on the client, but create one and only append your test there!

The case to check for is validating that the import to protocol.dart is correct, e.g. the output code should contain `import '../protocol.dart'`

tools/serverpod_cli/lib/src/generator/dart/library_generators/util/custom_allocators.dart
marcelomendoncasoares: You're right, good point. Just changed to your suggestion.

marcelomendoncasoares:
> The case to check for is validating that the import to protocol.dart is correct, e.g. the output code should contain `import '../protocol.dart'`

@Isakdl, I added the suggested test, but there is a problem. Although the real code generation have the protocol import - that is used in the part directives for extension of the parent class -, the test code generation is missing it.

The actual parent class on `tests\serverpod_test_server\lib\src\generated\inheritance\sealed_parent.dart` declares the protocol import as `_i1`:
```dart
import '../protocol.dart' as _i1;
import 'package:serverpod/serverpod.dart' as _i2;
part 'sealed_child.dart';
part 'sealed_grandchild.dart';
part 'sealed_other_child.dart';
```

And the actual child class on `tests\serverpod_test_server\lib\src\generated\inheritance\sealed_child.dart` uses it to reference the parent class on the extends:
```dart
part of 'sealed_parent.dart';

class SealedChild extends _i1.SealedParent
    implements _i2.SerializableModel, _i2.ProtocolSerialization {
  SealedChild({
    required super.sealedInt,
    required super.sealedString,
    this.nullableInt,
  });
```

But the `parentCompilationUnit` variable of the tests returns:
> "import 'package:serverpod/serverpod.dart' as _i1; part 'example_child.dart'; sealed class Example implements _i1.SerializableModel, _i1.ProtocolSerialization {Example({required this.name}); String name;}"

And the `childCompilationUnit` extends the class directly:
> "part of 'example.dart'; abstract class ExampleChild extends Example implements _i1.SerializableModel, _i1.ProtocolSerialization {ExampleChild._({required super.name, required this.age}); ...

The child is extending the `Example` class directly, without referencing it from a protocol import.

I can confirm that the relative path fix solves the problem because the `serverpod generate` command no longer breaks the import on the sealed classes. But the added test will fail, so I added it only on the server for now and with a skip clause. Any clues on why the generated code for the test is different from the real one?

Isakdl:
Hey @marcelomendoncasoares 

I took a look a bit deeper on this, I don't think the bug is related to sealed classes at all. The problem is with having the model in nested folders. 

The reason that the test is not including the `protocol` import is that the test data is not configured to need this import. This import is needed when referencing another model as a class type.


So first we have to configure the test data to set the file to be inside a folder. This can be done with the method `withSubDirParts` on the `ClassDefinitionBuilder`.

Then we need to configure a class reference to another model as a field on the current class.

After that we will get the `'../protocol.dart'` import.

Full test setup:


```dart
import 'package:serverpod_cli/src/util/model_helper.dart'; // imports defaultModuleAlias

    var example = ClassDefinitionBuilder()
        .withClassName('Example')
        .withFileName('example')
        .withSubDirParts(['folder'])
        .withField(FieldDefinitionBuilder()
            .withName('myClass')
            .withType(
              TypeDefinitionBuilder()
                  .withClassName('RefClass')
                  .withUrl(defaultModuleAlias)
                  .build(),
            )
            .build())
        .build();

    var refClass = ClassDefinitionBuilder()
        .withClassName('RefClass')
        .withFileName('ref_class')
        .withSimpleField('name', 'String')
        .build();
```

But I think this test actually belongs in a new test file, if you could create one that we call `imports_test.dart` and move the test there with this setup! Then copy the entire file into the `client_code_generator` as well. 

E.g. we end up with

- client_code_generator
  - imports_test.dart
- server_code_generator
  - imports_test.dart

Isakdl:
@marcelomendoncasoares 

Sorry I made one mistake here it is indeed related to sealed classes.

This is the test construction needed, but I still think it makes sense to put this in a standalone file.

```dart
   var parent = ClassDefinitionBuilder()
        .withClassName('Example')
        .withFileName('example')
        .withField(FieldDefinitionBuilder()
            .withName('user')
            .withType(
              TypeDefinitionBuilder()
                  .withClassName('User')
                  .withUrl(defaultModuleAlias)
                  .build(),
            )
            .build())
        .withIsSealed(true) 
        .build();

    var child = ClassDefinitionBuilder()
        .withClassName('ExampleChild')
        .withFileName('example_child')
        .withSimpleField('age', 'int')
        .withExtendsClass(parent)
        .build();

    var user = ClassDefinitionBuilder()
        .withClassName('User')
        .withFileName('user')
        .withSimpleField('name', 'String')
        .build();
```

marcelomendoncasoares:
Thanks for the help, @Isakdl! Just made the suggested changes, with the tests on standalone files.

Isakdl (approved):
LGTM, thank you for the work on this @marcelomendoncasoares 🚀 🚀 🚀

Issue #649, opened 2023-02-12
Make UUID default to UserUnfo table and Object


yahu1031:
## Feature request description ##
@fischerscode Instead making uuid optional for the user's custom DB. Also, we can add UUID as a default `ColumnString` in the user_info table.

> NOTE: This may need a whole table to alter if any DB/Server is in production. Lemme know your thoughts.

fischerscode:
*To be honest, I never  myself whether the `int` id is an issue.*

It'd definitely be easier to implement. But I'd prefer a `ColumnUUID`. 😆

Unfortunately this is a **breaking change**.

It might be best to give the modules versions independent of the main Serverpod version. Due to the way `pub get` resolves, it should be safe.
@vlidholt What do you think about this?

yahu1031:
Yes it is a breaking change but not for 2.*, can go to 1.*

vlidholt:
All breaking changes are saved for 2.x, unless there is a security issue.

glourencosilva:
I believe it should be configurable to the user's liking. Personally, I only use the gen_random_uuid () function in postgres to generate my IDs because it makes it easy for a possible migration to another database. So it would be interesting for the dev to choose how they want to generate the table's ID.

https://www.postgresql.org/docs/current/functions-uuid.html

fischerscode:
> So it would be interesting for the dev to choose how they want to generate the table's ID.

Might be related to #236. Or at least defaults might be a solution.

tgrushka:
Serverpod should have a config option to use UUID instead of serial integer for the `id` field if the developer requires it, especially for distributed apps. Somehow there could be a huge warning that changing this option would totally break their database. You could have it as a config flag i.e. `--uuid` in the serverpod CLI. It would be complicated to implement, but worth it.

SandPod:
@yahu1031 - Am I correct I interpret this issue as using UUIDs for all id fields instead of the current int id solution?

That would also differentiate it from https://github.com/serverpod/serverpod/issues/554, that focuses more on ensuring that ids are not guessable.

marcelomendoncasoares:
> @yahu1031 - Am I correct I interpret this issue as using UUIDs for all id fields instead of the current int id solution?
> 
> That would also differentiate it from #554, that focuses more on ensuring that ids are not guessable.

[...]

Issue #635, opened 2023-02-09
Add support for UUID


vlidholt:
It would be nice to be able to use UUID types with the database.

fischerscode:
Would the UUID be a `String` at the dart side?

fischerscode:
Related to #351

fischerscode:
Just thought about this, and it is not trivial, since we don't have a dedicated dart type representing an uuid. I think wit would be best, if the developer would specify `_uuid` as the type in the yaml file. With a bit of reworking, we could then support uuids.

PR #3152, opened 2025-01-20
feat: option to use UUIDs as primary keys

marcelomendoncasoares:
This PR implements the possibility to use UUIDs as primary keys for any table. It addresses a recurrent request that was discussed on the issues #351, #635 and #2641. It is also a pre-requisite for the issue #649. In these discussions, it is suggested to create unique indexes on UUID fields instead of using it as PK, but this only ensures uniqueness through constraints, while relationships between entities remain on sequential fields.

Although the common best practice is to use sequential as PK, there are some use cases where using UUIDs is preferred. In distributed systems, for example, the main advantage is the possibility to create several related entities on the client and feed them to the server without round-trips to get ID's. On offline-first applications, in which local and remote database must be a copy of each other, using UUIDs as PK also simplifies upserts without dealing with unique indexes violations (specially for related entities).

It is known that using UUIDs as PKs have several drawbacks, such as lack of PK sorting, increased storage, index bloat, etc. As alternative to reduce some of the drawbacks, the PR also includes the possibility of using the newly approved UUIDv7 spec, that solves most performance problems by being lexicographically sortable. I thought in separating this to another PR, but it is dependent on this PR and the only use case for using UUIDv7 seems to be as primary key.

## Implemented features:

- New experimental flag `changeIdType`. All features described below are behind this flag.
- New model property `idType` that depends on existing the `table` property. Admits any valid ID type (currently `int`, `uuidV4` or `uuidV7`). Changes the type of the ID field for the class table only. Any relation to this table will use this ID as FK if needed.
- New `--defaultIdType` option on the `serverpod create` command that changes the default ID type for all tables from `int` to the chosen type. Useful in case the majority of the tables will use such ID type. Admits the same options as the `idType` class property.
- New `defaultIdType` key on the `generator.yaml` file that does the same as the above description. Command line option takes precedence over the value stored by this method.
- Fields of type `UuidValue` now have a new possible default value `random_v7` that will create UUIDs randomly according to the v7 spec. This was included due to the advantage of using lexicographically sortable UUIDs when working with them as PKs. On the database level, a `gen_random_uuid_v7` function was added to support this (in native pgsql, extracted from [this gist](https://gist.github.com/kjmph/5bd772b2c2df145aa645b837da7eca74)).

## Pre-launch Checklist

## Breaking changes

As the `Table` and `TableRow` concept classes contains the typed `id` field, to maintain type safety, those classes had to be turned generic with the id type. In consequence, all `Database` methods had to be turned generic to the id type together with the `TableRow<T_ID>`. Because of such changes, all server models had to be re-generated.

However, after a user updates the serverpod version and call `serverpod generate`, only direct usages of the `session.db` methods should be broken, because they are now generic of two types instead of one. In such cases, adding the id type to the repository method before the TableRow type fixes the problem (i.e.: change `session.db.find<Model>` to `session.db.find<int, Model>`). Calls to regular repository methods are not affected (i.e.: `Model.db.find`).

It is also necessary to publish new versions of the `serverpod_chat` and `serverpod_auth` modules with the re-generated server models.

## Major changes on existing code:

To facilitate the revision, below is the logical order of the changes on existing code to support other id types on the database:

- A new class `SupportedIdType` concentrate the spec of all available id types. To support a new type in the future, the change must be done only in this class, that is referenced whenever a validation or parsing is needed. The PR implements two UUID types as option, but the feature was designed to allow extension for other types as well. If adding a new type other than `int` or `UuidValue` (such as `String`), the `Table` constructor must be updated. This is referenced on the `SupportedIdType` class doc comment.
- Base column types were rearranged and there is a new public column type `ComparableColumn<T>`. This is the type of valid id columns and implements equality and inequality comparison. All types that have [such comparison options on postgres](https://www.postgresql.org/docs/current/functions-comparison.html#FUNCTIONS-COMPARISON-OP-TABLE) are now `ComparableColunn`. With this, the string and uuid types can now be compared using less than/more than (and the equal to variant) operators, as supported by Postgres.
- The `Table` and `TableRow` concept classes are now generic to the type of the ID attribute, with the `TableRow<T_ID>.id` being of type `ComparableColunn<T_ID>`. This is why the majority of methods that interact with a table are now generic of two references: `<T_ID, T extends TableRow<T_ID>`. This change was the responsible for most of the changes on existing code, as `Database` and `DatabaseConnection` had several methods updated and all server models had to be re-generated.
- The `ObjectRelationDefinition` and `ListRelationDefinition` objects now store the id type of the table that owns the FK field. This reference is used to pass the `T_ID` generic reference to build the attach/detach repository methods on the server models.
- To support the `uuidV7` id type, a pgsql function named `gen_random_uuid_v7` is declared at the top of `definition.sql` and `migration.sql` on all `create-migration` calls. This function is used as default persist for such id type and any other `UuidValue` column that declares default as `random_v7`. As the function was extracted from [this gist](https://gist.github.com/kjmph/5bd772b2c2df145aa645b837da7eca74), its license is properly referenced on the `extensions.dart` module.

## Tests considerations

On all new tests, I tried to follow the practices recommended on the `testing-guideline` branch of the `serverpod_docs` repo and the examples of existing tests. The only test abstractions that I used was to make tests robust to all possible id types, including any new types implemented in the future - and to keep such tests maintainable, as copying and pasting all test files for every id type didn't seem to me as a good idea.

Regarding existing tests, the only changes were to fix code that broke due to other changes. All new features are tested on new tests, that I wrote trying to avoid overlap with existing tests.

## Final considerations

I tried to keep the changes to the minimum for this implementation, but the primary key as `int` was a strong premise of the project. If there is anything that I failed to adhere to the practices of the repo, just point it out. I can also help with a PR on the `serverpod_docs` repo once/if this PR gets approved.


tests/serverpod_test_server/lib/src/models/changed_id_type/one_to_one/address.spy.yaml
```diff
@@ -0,0 +1,11 @@
+class: AddressUuid
+table: address_uuid
+idType: uuidV4
```
SandPod, line 3: Suggestion: Use the reserved `id` field to apply these types of configuration. We could have stricter restrictions for that field specifically. But that would allow us to define `scope` and possible `colum name` for the id field later when that is added.

tools/serverpod_cli/lib/src/commands/create.dart
```diff
@@ -36,6 +39,11 @@ class CreateCommand extends ServerpodCommand {
       help: 'Template to use when creating a new project, valid options are '
           '"mini", "server" or "module".',
     );
+    argParser.addOption(
```
marcelomendoncasoares, line 42: This change is missing a test that I'm working on to ensure that the command reaches the `performCreate` call with and without this option being specified.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -49,7 +49,7 @@ class DatabaseConnection {
   }
 
   /// For most cases use the corresponding method in [Database] instead.
-  Future<List<T>> find<T extends TableRow>(
```
SandPod, line 52: Suggestion/Exploration: Check if it is possible to retrieve the id type from the TableRow class.

That would allow us to introduce this change without it being a breaking change in the framework.

SandPod (commented):
Once we have explored the two opens questions, will separate out the chnages in this PR to scoped down changes in the framework. This will help future engineers understand why the system has changed and for us to improve the review quality of each change.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
marcelomendoncasoares: For reference, I guess the main challenge is keeping the type safety on methods such as `findById` - which was the main one that led me to make `TableRow` generic in the end, beside also trying to avoid it at all costs. The only other alternative I found was to make such references to the id type `dynamic`, but then users calling `session.db.findById<TableRowClass>` would not know the id type and could face runtime errors. But I hope there is some other way that I haven't seen, as this would simplify the implementation and avoid the breaking change.

tests/serverpod_test_server/lib/src/models/changed_id_type/one_to_one/address.spy.yaml
marcelomendoncasoares: I'm with an already working version of it on the branch `configurable-id-field` of my fork. I'll just fix/add the tests and then merge into this one.

marcelomendoncasoares: @SandPod, change merged to the PR as specified. I restricted the id field to only accept the `default` key when needed (for `UuidValue` types, but not for the `int` type). All other keys are forbidden until there are proper use cases.

tools/serverpod_cli/lib/src/commands/create.dart
marcelomendoncasoares: Done!

tools/serverpod_cli/lib/src/create/create.dart
```diff
@@ -418,9 +418,9 @@ void _copyServerUpgrade(
           replacement: generateRandomString(),
         ),
         Replacement(
-          slotName: 'defaultIdType: int\n',
+          slotName: 'defaultIdType: int',
```
marcelomendoncasoares, line 421: This change was necessary because on some Window machines the newline is being generated as `\r\n` (strangely, not in all runs). By removing the newline, in case the user calls the create command with no defaultIdType option, the `generator.yaml` file will have two empty lines in a row:

1. Template file:
```yaml
type: server
defaultIdType: int

client_package_path: ../projectname_client
server_test_tools_path: test/integration/test_tools
```

2. Generated file when there is no defaultIdType option:
```yaml
type: server


client_package_path: ../projectname_client
server_test_tools_path: test/integration/test_tools
```

To avoid this, we could change the replacement by something like:
```dart
// The following replacement must be on a valid string instead of a
// placeholder to avoid generating an invalid file. It includes the EOL
// to avoid leaving two newlines in a row. The EOL loop is necessary
// for Windows, where line endings are generated as '\r\n'.
for (var platformEOL in ['\r\n', '\n'])
  Replacement(
    slotName: 'defaultIdType: int$platformEOL',
    replacement: defaultIdType != null
        ? 'defaultIdType: ${defaultIdType.aliases.first}$platformEOL'
        : '',
  ),
```

However, I felt it to be too dirty on the code for such a small detail. It is valid to mention that the template must contain a valid default option instead of a template name to avoid breaking existing tests that just copy the file, without calling the create command. 

@SandPod, do you have a take on what would be preferred here for the library?

tests/serverpod_test_server/lib/src/models/changed_id_type/one_to_one/address.spy.yaml
SandPod: In the PR, I can see that we now have a default defined for the int type as well. Did this implementation detail change? I think it is fine to allow the default key to be defined. But it should not be required for the `int` id type. We will just use a set of defaults there since it has special meaning in the framework.

tools/serverpod_cli/lib/src/create/create.dart
SandPod: We would prefer not to have `defaultIdType` defined at all in the config file. The reason for this is that we try and keep the context as small as possible for new users to Serverpod. We can then document how to change this if that is desired.

tests/serverpod_test_server/lib/src/models/changed_id_type/one_to_one/address.spy.yaml
marcelomendoncasoares: You're right, it makes sense for the `int` type to be specifiable without a default. I changed to make the implementation cleaner, but there is no need to force the declaration. Just reverted the behavior for `int` (while still keeping the simplification).

tools/serverpod_cli/lib/src/create/create.dart
marcelomendoncasoares: The reason why I put this optional key in the file is to allow the configuration to be persistent for the project (configured once as a flag on the `create` command). So that a `--defaultIdType` key should not need to be present on all subsequent calls to `serverpod generate` - as forgetting it once would change everything on the models. However, I implemented it as an optional key in the `generator.yaml`, making the project default to `int` if it is missing (as would happen with already existing projects). Do you think that even as an optional key with this behavior it would be a problem?

marcelomendoncasoares: The config in the `generator.yaml` is also needed for the language server to understand correctly the id type for the models if they were generated with a different default id type.

SandPod: We **SHOULD** have `defaultIdType` as an optional key in the `generator.yaml` file. It should be configurable and persistent for a project if the developer would like that.

But key should not be present in the `generator.yaml` template, with the above motivation (keep required initial context as small as possible). If we change the default value for this key for Serverpod. This will be listed as a breaking change with a clear migration step.

marcelomendoncasoares: Ok, think we are aligned on this. With the proposed implementation, the key will only be present on the `generator.yaml` after project creation if the user adds the `defaultIdType` option on the create command. Otherwise (which is the default), the key will be removed from the template during the project creation and the file will be the same as it is today.

I only added it to the template to allow the create command to inject the configuration on the correct place (or remove it) using the existing `Copier` + `Replacement` mechanism. But it would not need to be there before the create command if we have a better approach to inject a key on this file during the create.

With this behavior (of removing the key when the user does not use it), do you think the small context for the user is preserved? Or should we look for another configuration injection strategy on the `generator.yaml` file?

SandPod: Because Serverpod `mini` is not compatible with `--defaultIdTyp` I don't think we should add it to the top-level `create` command as `mini` is generated without the `generator.yaml` file. 

To still get some visibility around this configurable option, we might want to keep your original implementation of adding the keyword to the `generator.yaml` but comment it out with an explanation that includes its purpose and valid options.

The context for the user grows a bit, but ideally, it still provides enough value.

SandPod:
@marcelomendoncasoares - Closing this as this has now been implemented and will be part of the 2.9.0 release.

Issue #3154, opened 2025-01-20
Reduce/Eliminate 2-way dependencies


cannon-fodder:
### Problem to Solve

2-way dependencies have high potential for causing memory leaks (objects with cyclic dependencies are never garbage collected) and tend to make code hard to modify.

### Proposal

Introduce a service locator pattern so classes that need access to configures services don't need a reference to a Serverpod or Server instance for access to configured services.

### Use Case

In serverpod/lib/src/server/session.dart each Session instance has a MessageCentralAccess instance (which has a reference to the session it belongs to).  This cycle doesn't seem to be broken during cleanup resulting in objects that are never garbage collected.  The reason for the 2-way dependency seems only to get access to session.server.messageCentral.  This could be access through a service locator instead, breaking the 2-way dependency and allowing the garbage collector to clean up discarded session objects.

### Alternatives

There are other ways to cyclic dependencies, but this one seems to fit the use case the best.

### How experienced are you with this library?

Beginner - Just getting started with this library

### Are you interested in working on a PR for this?

- [x] I want to work on this

SandPod:
Hey @cannon-fodder 👋 

### Service locator pattern

I believe you are correct in the way that some services should be accessed using a service locator pattern.

For the specific case that you listed, I'm wondering if it could not be solved in

SandPod:
Hey @cannon-fodder 👋 

2-way dependencies is a code smell, and I'm not suprised if it affects performance of the system in some way.

### Service locator pattern

I believe that you are correct in that there needs to be a way to retrieve services in Serverpod. The question is how to provide that functionality.

I'm generally very cautious about introducing singletons as these introduce a global stored state and make testing difficult. We can already see that in your tests where each test needs to use a globally unique key or they would fail.

I would also like such a system to be pluggable for modules so that they can tag on additional services that are unique for them with the configuration they require. The auth module is an example of that.

### MessageCentralAccess

For the specific case of message central, I'm wondering if that couldn't be solved in a different way. There really is no need for the message central to exist or keep an internal reference to the session. What we want is the interface so that developers interacting with session get a typed interface around what is available.

### Logging

We have an RFC that starts the discussion around logging in Serverpod: https://github.com/serverpod/serverpod/issues/2847

The solution we are looking for there is to consolidate both the request logging through session and the "framework" logging such as verbose logs or exception logging.

cannon-fodder:
[...]

PR #3155, opened 2025-01-20
Begin removing 2-way dependencies through the use of a service locator pattern

cannon-fodder:
A large amount of the code in this project has 2-way dependencies, which tend to cause memory leaks and make it more difficult to modify code.  This change introduces a service locator pattern as a way to begin eliminating 2-way dependencies where possible.  As an example case, this change introduces a new ConsoleLogger service class which replaces Serverpod.logVerbose(String).  The serverpod instance initializes a ConsoleLogger in the ServiceManager and then other classes that want access to console logging calls lookup the service instead of using a reference to a Serverpod instance.  This is only a first step in addressing the issue

Issue #3154 

This is my first contribution to the project, so I appologize if I've missed something.

## Pre-launch Checklist

## Breaking changes

removing Serverpod.logVerbose shouldn't be a breaking change, unless downstream apps are relying on it (hopefully not)  We could put a stub back that uses the configured console logger, but that seems counter to the purose of the change.


cannon-fodder:

cannon-fodder:
I've modified service locator mechanism to be pluggable and modified serverpod and server to configure ServiceHolder/ServiceLocator instances, as well as modified code using references like session.server.serverpod to use serviceLocator.locate.  I also added an optional name to configured values. since I found that there were cases where you had multiple instances of the same type (particularly String and Map).  I extracted the FutureCall mechanisms from serverpod into a new FutureCallFacade class, but left the methods on serverpod with a call to a configured facade instance.  Code downstream in the configuration chain was modified to locate a facade instance.  I have some other typedef functions that should be replaced with classes, that I haven't gotten to yet.  Comments and suggestions on the work so far is welcome.

Issue #3104, opened 2025-01-07
serverOnly object relations allow required foreign key column


Isakdl:
**Describe the bug**
If a model has an object relation that is server only no error is displayed in the analyzer. But the generated code will generate a fromJson method that expects the required foreign key column to be present, even though the client model does not contain said field.

**To Reproduce**

```dart
class: Example
table: example
fields:
  name: String
  user: User?, relation, scope=serverOnly
```

Send the class as a param in an endpoint and there will be a run time exception.

See full thread here: https://github.com/serverpod/serverpod/discussions/3100#discussioncomment-11759472

**Expected behavior**
An error should be given by the analyzer if a `serverOnly` object relation is declared without an `optional` foreign key column.

**Serverpod versions**
 -  _CLI Version:_ main
 - _Library version:_ main

PR #3171, opened 2025-01-24
feat: Add validation for relations with serverOnly scope to ensure they are optional

klkucaj:
### Description
This PR introduces validation to ensure that relations with the `serverOnly` scope are explicitly marked as optional. Without this validation, the analyzer fails to detect invalid model definitions, leading to runtime exceptions during deserialization.

Previously, if a relation with the `serverOnly` scope was not marked as optional, the generated `fromJson` method would expect the foreign key column to be present, even though it is absent in the client model. This mismatch caused runtime errors when the class was passed as a parameter to an endpoint.

### Example
**Invalid Definition:**
```yaml
class: Example
table: example
fields:
  name: String
  user: User?, relation, scope=serverOnly
```

**Valid Definition:**
```yaml
class: Example
table: example
fields:
  name: String
  user: User?, relation(optional), scope=serverOnly
```

Closes: #3104 

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_relation_scope_test.dart
+++ b/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_relation_scope_test.dart
@@ -11,13 +11,13 @@ void main() {
   var parentClassModel = ModelSourceBuilder()
       .withYaml(
         '''
-        class: ParentClass
-        table: parent
+        class: Post
+        table: post
         fields:
-          name: String
+          title: String
         ''',
       )
-      .withFileName('parent_class')
+      .withFileName('post_class')
       .build();
 
   group('Given a class with a non-optional relation and "serverOnly" scope',
@@ -30,13 +30,13 @@ void main() {
           ModelSourceBuilder()
               .withYaml(
                 '''
-                class: Example
-                table: example
+                class: Comment
+                table: comment
                 fields:
-                  parentClass: ParentClass?, relation, scope=serverOnly
+                  post: Post?, relation, scope=serverOnly
                 ''',
               )
-              .withFileName('example_class')
+              .withFileName('comment_class')
               .build(),
         ];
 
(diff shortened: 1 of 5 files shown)


tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_relation_scope_test.dart
```diff
@@ -0,0 +1,147 @@
...
+        expect(collector.errors, isNotEmpty);
+
+        var error = collector.errors.first as SourceSpanSeverityException;
+        expect(error.severity, SourceSpanSeverity.error);
+        expect(
+          error.message,
+          'The relation with scope "serverOnly" requires the relation to be optional.',
+        );
+      },
+    );
```
Isakdl, line 90: Suggestion: This case should not generate an error as the field `relationId` is not serverOnly.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_relation_scope_test.dart
```diff
@@ -0,0 +1,147 @@
...
+        var models = [
+          parentClassModel,
+          ModelSourceBuilder()
+              .withYaml(
+                '''
+                class: Example
+                table: example
+                fields:
+                  relationId: int
+                  parentClass: ParentClass?, relation(optional, field=relationId), scope=serverOnly
```
Isakdl, line 132: Question: is this really possible? Having these two keywords at the same time looks wrong to me, `relation(optional, field=relationId)`. Since `optional` only makes sense if there is no "manual" field defined.

Isakdl: `optional` and `field` should probably be mutually exclusive

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_relation_scope_test.dart
```diff
@@ -0,0 +1,147 @@
...
+import 'package:serverpod_cli/src/test_util/builders/model_source_builder.dart';
+import 'package:test/test.dart';
+
+void main() {
+  var config = GeneratorConfigBuilder().build();
+
+  var parentClassModel = ModelSourceBuilder()
+      .withYaml(
+        '''
+        class: ParentClass
```
Isakdl, line 14: Nit: Rename to something else, Parent could be confused with inheritance which this is not.

Isakdl (changes requested):
There is one case that should be inverted but otherwise we have full coverage I think! 

See comments :).

Isakdl (approved):
LGTM

PR #3176, opened 2025-01-27
fix: `LocalCache` returning `null` for expired keys instead of invoking `cacheMissHandler`

### LINKED ISSUE #2953: `CacheMissHandler` doesn't work as expected
### Discussed in https://github.com/serverpod/serverpod/discussions/2951

<div type='discussions-op-text'>

<sup>Originally posted by **lukasbalaz** November 11, 2024</sup>
I am putting this in Q&A instead of creating an issue because there is high chance there is some good reason it works like this but for me it's just really unintuitive.

If I use something like this:
```dart
final userData = session.caches.localPrio.get(
  'User-$id',
  CacheMissHandler(
    () async {
      User user = await getUserFromDb(id);
      return user;
    },
    lifetime: const Duration(minutes: 10),
  ),
);
```
I would expect `userData` never to be `null`. Either it gets the data from cache or from the database.

But for some reason it actually returns `null` if the entry exists in the cache but is expired (see line 91 below). The `cacheMissHandler` is only invoked on a subsequent call.
https://github.com/serverpod/serverpod/blob/main/packages/serverpod/lib/src/cache/local_cache.dart#L88-L92

For me it looks weird and unexpected. Is there some good reason for this?</div>

tp:
Closes #2953

Additionally ensures only one `cacheMissHandler` for a given key is invoked, and that they won't be called in parallel for a given key. All `get`s will then resolve to the same value (or error).

### DIFF
--- a/packages/serverpod/lib/src/cache/local_cache.dart
+++ b/packages/serverpod/lib/src/cache/local_cache.dart
@@ -97,7 +97,6 @@ class LocalCache extends Cache {
 
     if (entry != null) {
       if (entry.isExpired) {
-        // invalidate key and generate a new value below
         await invalidateKey(key);
       } else {
         return serializationManager.decode<T>(entry.serializedObject);
--- a/packages/serverpod/lib/src/cache/local_cache.dart
+++ b/packages/serverpod/lib/src/cache/local_cache.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:collection/collection.dart';
 import 'package:serverpod/src/cache/cache_miss_handler.dart';
 import 'package:serverpod_serialization/serverpod_serialization.dart';
@@ -10,6 +12,10 @@ class LocalCache extends Cache {
   final List<_KeyListKey> _keyList = <_KeyListKey>[];
   final Map<String, _CacheEntry> _entries = <String, _CacheEntry>{};
   final Map<String, Set<String>> _groups = <String, Set<String>>{};
+
+  /// New cache values currently being computed by a [CacheMissHandler]
+  // The future values in here must not be resolved (at which state the value should just be in the cache), but just pending
+  final _inProgressCacheValues = <String, Future<SerializableModel?>>{};
 
   /// Creates a new [LocalCache].
   LocalCache(super.maxEntries, super.serializationManager);
@@ -81,6 +87,10 @@ class LocalCache extends Cache {
   @override
   Future<T?> get<T extends SerializableModel>(
     String key, [
+    /// Handler to generate a new value in case there is no active value in the cache
+    ///
+    /// In case a value computation from a previous [get] call is already running, the caller will receive the value from
+    /// that call and the `cacheMissHandler` from this call will not be invoked.
     CacheMissHandler<T>? cacheMissHandler,
   ]) async {
     var entry = _entries[key];
@@ -94,10 +104,30 @@ class LocalCache extends Cache {
       }
     }
 
+    var pendingEntry = _inProgressCacheValues[key];
+    if (pendingEntry != null) {
+      return pendingEntry.then((v) => v as T);
+    }
+
     if (cacheMissHandler == null) return null;
 
-    var value = await cacheMissHandler.valueProvider();
-    if (value == null) return null;
+    T? value;
+    var completer = Completer<T?>();
+    try {
+      _inProgressCacheValues[key] = completer.future;
+
+      value = await cacheMissHandler.valueProvider();
+
+      completer.complete(value);
+
+      if (value == null) return null;
+    } catch (e, stackTrace) {
+      completer.completeError(e, stackTrace);
+
+      rethrow;
+    } finally {
+      unawaited(_inProgressCacheValues.remove(key));
+    }
 
     await put(
       key,
--- a/packages/serverpod/lib/src/cache/local_cache.dart
+++ b/packages/serverpod/lib/src/cache/local_cache.dart
@@ -70,7 +70,7 @@ class LocalCache extends Cache {
 
     if (entry == null) return false;
 
-    if ((entry.expirationTime?.compareTo(DateTime.now()) ?? 0) < 0) {
+    if (entry.isExpired) {
       await invalidateKey(key);
       return false;
     }
@@ -85,14 +85,13 @@ class LocalCache extends Cache {
   ]) async {
     var entry = _entries[key];
 
-    if (entry != null &&
-        (entry.expirationTime?.compareTo(DateTime.now()) ?? 0) < 0) {
-      await invalidateKey(key);
-      return null;
-    }
-
     if (entry != null) {
-      return serializationManager.decode<T>(entry.serializedObject);
+      if (entry.isExpired) {
+        // invalidate key and generate a new value below
+        await invalidateKey(key);
+      } else {
+        return serializationManager.decode<T>(entry.serializedObject);
+      }
     }
 
     if (cacheMissHandler == null) return null;
@@ -193,6 +192,11 @@ class _CacheEntry {
     required this.serializedObject,
     this.lifetime,
   }) : creationTime = DateTime.now();
+
+  bool get isExpired {
+    var expirationTime = this.expirationTime;
+    return expirationTime != null && expirationTime.isBefore(DateTime.now());
+  }
 }
 
 class _KeyListKey {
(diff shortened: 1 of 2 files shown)


packages/serverpod/lib/src/cache/local_cache.dart
```diff
@@ -81,24 +87,47 @@ class LocalCache extends Cache {
   @override
   Future<T?> get<T extends SerializableModel>(
     String key, [
+    /// Handler to generate a new value in case there is no active value in the cache
+    ///
+    /// In case a value computation from a previous [get] call is already running, the caller will receive the value from
```
tp, line 92: Seems like an important edge-case to mention, though I would think that all calls to `get` for a given type `T` will always use the same `cacheMissHandler`.

(Would be interesting to have an "overlay"/"view" on the cache to have everything cast to `T` and thus also avoid the duplication of lifetimes in `put` and `cacheMissHandler` as I would assume users would always want those to be aligned – but alas just my thoughts of the moment and outside the scope of the bug.)

packages/serverpod/lib/src/cache/local_cache.dart
```diff
@@ -81,24 +87,47 @@ class LocalCache extends Cache {
...
+      } else {
+        return serializationManager.decode<T>(entry.serializedObject);
+      }
     }
 
-    if (entry != null) {
-      return serializationManager.decode<T>(entry.serializedObject);
+    var pendingEntry = _inProgressCacheValues[key];
+    if (pendingEntry != null) {
+      return pendingEntry.then((v) => v as T);
```
tp, line 109: This obviously would not survive changing the model's type between invocations.

You could have written some code before that did just that (successfully), but I don't think that's a real-world behavior anyone wants or would rely upon…

packages/serverpod/lib/src/cache/local_cache.dart
```diff
@@ -85,14 +85,13 @@ class LocalCache extends Cache {
...
-    if (entry != null &&
-        (entry.expirationTime?.compareTo(DateTime.now()) ?? 0) < 0) {
-      await invalidateKey(key);
-      return null;
-    }
-
     if (entry != null) {
-      return serializationManager.decode<T>(entry.serializedObject);
+      if (entry.isExpired) {
+        // invalidate key and generate a new value below
```
Isakdl, line 100: Suggestion: Remove the comment, generally think comments should be avoided within the code. Very often these go out of sync with the code when it is changed.

Prefer refactoring the flow if it isn't obvious what is going on.

tests/serverpod_test_server/test/local_cache_test.dart
```diff
@@ -173,6 +175,35 @@ void main() {
...
+
+    final retrieved2Future = cache.get<SimpleData>(
+      key,
+      CacheMissHandler(
+        () async => throw '`cacheMissHandler` should not be called twice',
+        lifetime: Duration(minutes: 10),
+      ),
+    );
+
+    completer.complete(SimpleData(num: 100));
```
Isakdl, line 200: Comment: Neat solution!

tests/serverpod_test_server/test/local_cache_test.dart
```diff
@@ -173,6 +175,35 @@ void main() {
     expect(retrieved?.num, 2);
   });
 
+  test(
+      'get `cacheMissHandler` will only be called once when multiple requests are made',
```
Isakdl, line 178: Suggestion: Same as other comment on test description.

packages/serverpod/lib/src/cache/local_cache.dart
```diff
@@ -94,10 +104,30 @@ class LocalCache extends Cache {
...
+
+      completer.complete(value);
+
+      if (value == null) return null;
+    } catch (e, stackTrace) {
+      completer.completeError(e, stackTrace);
+
+      rethrow;
+    } finally {
+      unawaited(_inProgressCacheValues.remove(key));
```
Isakdl, line 129: Question: Just for my understanding, we need this as the value stored is a future?

packages/serverpod/lib/src/cache/local_cache.dart
Isakdl: Suggestion: I think you should type check this variable instead of casting, as casting is potentially dangerous. Return null if it doesn't match.

nit: use await here instead.

tests/serverpod_test_server/test/local_cache_test.dart
```diff
@@ -157,14 +157,13 @@ void main() {
...
-    await cache.put(
-      key,
-      SimpleData(num: 1),
-      // Object is expired from the start (but still gets added to the cache); we could also use a mocked "now" to have it expire on read
-      lifetime: Duration(minutes: -1),
-    );
+    await cache.put(key, SimpleData(num: 1),
+        lifetime: Duration(milliseconds: 100));
+
+    // Expire the newly added key
```
Isakdl, line 163: Suggestion: I think the comment is redundant.

Isakdl (changes requested):
Good job! And nice addition with the consolidation of cache misses. 

I have made some smaller suggestions.

packages/serverpod/lib/src/cache/local_cache.dart
tp: yes, because we get a `Future?` from the map, the linter wants us to explicitly state that we don't want to handle it.

alternatively we could write `map[key] = null` to avoid this.

packages/serverpod/lib/src/cache/local_cache.dart
tp: Type-checking and returning `null` might hide an error (in the rare case that the user really does change the type between calls).

In my view this is not different from the `serializationManager.decode<T>` used here. If the stored data does not match the current `T`'s "shape" the parser will also just crash with an internal type assertion, e.g.

CleanShot 2025-01-27 at 16 33 04@2x

If we really wanted to avoid the cast, we should also be able to check the runtime type of the "in progress" element, and in case it differs we would then use the new `cacheMissHandler` to generate the next value. (But then again, the `get` would fail if it's ever called again with the previous `T`, which would cause usage of the previous serializer…)

tests/serverpod_test_server/test/local_cache_test.dart
```diff
@@ -150,6 +150,29 @@ void main() {
     expect(retrieved, isNull);
   });
 
+  test(
+      'get object which is expired in cache will call given `cacheMissHandler`',
```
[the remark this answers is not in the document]
tp: Changed all the tests in this file to the new format.

packages/serverpod/lib/src/cache/local_cache.dart
tp: Removed the "obvious" comments.

tests/serverpod_test_server/test/local_cache_test.dart
```diff
@@ -8,55 +8,100 @@ import 'package:test/test.dart';
...
+    await cache.put(
+      key,
+      entry,
+      lifetime: const Duration(milliseconds: 100),
+    );
+
+    var retrieved = await cache.get<SimpleData>(key);
+    expect(retrieved?.num, equals(0));
 
     expect(cache.localSize, equals(1));
```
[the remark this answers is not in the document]
Isakdl: Apply to the other tests too!

Isakdl (changes requested):
Thank you for that test clean up! Very good! Two smaller requests on them!

packages/serverpod/lib/src/cache/local_cache.dart
Isakdl: You might be right! let's keep it as is.

tests/serverpod_test_server/test/local_cache_test.dart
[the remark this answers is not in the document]
tp: Okay, will remove them. I think initially this was probably intended as a “double check” so that we query the correct (and only) key, especially in some of the other cases where “return is `null`” is checked, which you could also achieve by passing the wrong key instead of an expired one etc. But now with all the tests using a `const key = …` such a mix up is no longer possible (for the tests that just operate on a single key at least).

So I will remove this. Overall, with regards to the new testing document, it seems like `localSize` is more of an “implementation test” than a behavior test anyway. The better approach would probably to e.g. set a size of 5, add 6 items, and then check that you can read only the last 5 out again.

Anyway, for this specific suggestion I’ll push the removal next.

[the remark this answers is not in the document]
tp: Removed that extra assertion from tests which are not primarily concerned with the size / eviction count behavior.

PR #3178, opened 2025-01-27
fix: Do not leak internal exception message for web routes

### LINKED ISSUE #2516: Server serves up raw exception message as a response
**Describe the bug**

I deleted `templates/default.html`, which was in use by my web app, and Serverpod ended up serving up the raw exception message as a response: `Null check operator used on a null value`. A raw exception message like that should never be served as an HTTPS response!

This is the exception from the log:

```
INFO 2024-07-21T06:36:29.747372345Z [resource.labels.instanceId: serverpod-production-28gh] startup-script: ERROR: Null check operator used on a null value
INFO 2024-07-21T06:36:29.748434269Z [resource.labels.instanceId: serverpod-production-28gh] startup-script: #0 new Widget (package:serverpod/src/relic/web_widget.dart:34)
INFO 2024-07-21T06:36:29.748489428Z [resource.labels.instanceId: serverpod-production-28gh] startup-script: #1 new DefaultPageWidget (package:click_server/src/web/widgets/default_page_widget.dart:4)
INFO 2024-07-21T06:36:29.748510924Z [resource.labels.instanceId: serverpod-production-28gh] startup-script: #2 RouteRoot.build (package:click_server/src/web/routes/root.dart:9)
INFO 2024-07-21T06:36:29.748527268Z [resource.labels.instanceId: serverpod-production-28gh] startup-script: #3 WidgetRoute.handleCall (package:serverpod/src/relic/web_server.dart:301)
INFO 2024-07-21T06:36:29.748541658Z [resource.labels.instanceId: serverpod-production-28gh] startup-script: #4 WebServer._handleRouteCall (package:serverpod/src/relic/web_server.dart:172)
INFO 2024-07-21T06:36:29.748557188Z [resource.labels.instanceId: serverpod-production-28gh] startup-script: #5 WebServer._handleRequest (package:serverpod/src/relic/web_server.dart:153)
INFO 2024-07-21T06:36:29.748571805Z [resource.labels.instanceId: serverpod-production-28gh] startup-script: #6 WebServer._start (package:serverpod/src/relic/web_server.dart:94)
```

So there are two issues here: (1) that the `!` operator is being used dangerously; and (2) that the web server is directly delivering exception messages back to the browser. These should be logged, and something more generic reported to the user.

tp:
Closes #2516

### DIFF
--- a/packages/serverpod/lib/src/relic/web_server.dart
+++ b/packages/serverpod/lib/src/relic/web_server.dart
@@ -164,11 +164,7 @@ class WebServer {
       logError(e, stackTrace: stackTrace);
 
       request.response.statusCode = HttpStatus.internalServerError;
-      if (serverpod.runMode == ServerpodRunMode.production) {
-        request.response.write('Internal Server Error');
-      } else {
-        request.response.write('$e');
-      }
+      request.response.write('Internal Server Error');
 
       await request.response.close();
     }
--- a/packages/serverpod/lib/src/relic/web_server.dart
+++ b/packages/serverpod/lib/src/relic/web_server.dart
@@ -103,7 +103,7 @@ class WebServer {
   }
 
   void _handleRequest(HttpRequest request) async {
-    if (serverpod.runMode == 'production') {
+    if (serverpod.runMode == ServerpodRunMode.production) {
       request.response.headers.add('Strict-Transport-Security',
           'max-age=63072000; includeSubDomains; preload');
     }
@@ -164,7 +164,12 @@ class WebServer {
       logError(e, stackTrace: stackTrace);
 
       request.response.statusCode = HttpStatus.internalServerError;
-      request.response.write('$e');
+      if (serverpod.runMode == ServerpodRunMode.production) {
+        request.response.write('Internal Server Error');
+      } else {
+        request.response.write('$e');
+      }
+
       await request.response.close();
     }
     return true;


packages/serverpod/lib/src/relic/web_server.dart
```diff
@@ -164,7 +164,12 @@ class WebServer {
       logError(e, stackTrace: stackTrace);
 
       request.response.statusCode = HttpStatus.internalServerError;
-      request.response.write('$e');
+      if (serverpod.runMode == ServerpodRunMode.production) {
```
tp, line 167: For now we just keep the error internally on production deployments, but I also wondered whether this should be configurable.

Then we would need to somehow set it to enabled for development (where it's very helpful), but should make extra sure that consumers never enable it in production.

Relying on the "Run mode" enum here seems best, as we would be aware when changing this in the future (as there have been discussions to have more flexible run modes).

Isakdl: Suggestion: Remove the production mode check, I would like to move away from doing hard checks on the mode we are running in. This should mostly be used by the application layer and selection of configuration.

We are already logging the exception so the programmer would have access to the error during development.

packages/serverpod/lib/src/relic/web_server.dart
```diff
@@ -103,7 +103,7 @@ class WebServer {
   }
 
   void _handleRequest(HttpRequest request) async {
-    if (serverpod.runMode == 'production') {
+    if (serverpod.runMode == ServerpodRunMode.production) {
```
Isakdl, line 106: Comment: Okay to leave this here as that was already present, and this change is an improvement.

Isakdl (changes requested):
I have added a modification request, could you also check if we are still throwing the null check cast error if you delete the template folder as described in the issue. 

This should also be handled gracefully if that is still a problem! :)

packages/serverpod/lib/src/relic/web_server.dart
tp: @Isakdl So we would then move this into the configuration? E.g. something like `returnErrorsOnWebserver: true`?

> could you also check if we are still throwing the null check cast error if you delete the template folder as described in the issue.

The `!` has been removed a while back, and now we return a specific error message (like "template not found") in that case.

But still I think the overall issue was not that specific error message or the `!` (which is not really worse than doing the null check ourselves and then throwing an exception), but rather the leaking of any internals, which is now prevented.

tp: @Isakdl Okay, to summarize your suggestion would be to never return the “raw error message” (even during development), as it’s visible in the logs and printed to the console.

Makes sense to me as well, and then we don’t add another dependency on the obsolete mode. And indeed, the logs would be the first thing the developer looks at anyway when encountering such an error.

Will change this then accordingly.

Isakdl (approved):
LGTM

Issue #1237, opened 2023-08-02
Add documentation to generated code


SandPod:
**Is your feature request related to a problem? Please describe.**
Code that is generated does not have any automatically generated documentation that explains the interfaces. This makes it hard to understand how the interfaces can be used and generates questions to the Serverpod Forum.

**Describe the solution you'd like**
Include documentation as part of the code generation that explains:
- Parameters
- Return value
- Method/Function purpose and responsibility.

tp:
@SandPod Do you recall any specific classes that are still lacking from your point of view?

I checked the current code generation, and it seems that we do copy model field comments (from `YAML` to `Dart`) and also the comments for methods exposed on `Endpoint`s.

In the latter case I only saw that we did not transfer comment from the parameters to the generated client code. Maybe that would be a good place for me to start, and then build upon if we can list more cases where we should copy over existing comments.

Isakdl:
@tp Looking at the time this was written I believe this is referencing the methods that we supply the users with. e.g:

```dart
MyModel.db.insert(...);

MyModel.db.update(...);

model.copyWith(...);

... etc
```

Isakdl:
Closing as completed.

PR #3181, opened 2025-01-28
Generate docs for `TableRow` model repositories

tp:
Relates to #1237

The comments are adapted from the underlying `Database` class.


tests/serverpod_test_server/lib/src/generated/unique_data.dart
```diff
@@ -183,6 +183,10 @@ class UniqueDataIncludeList extends _i1.IncludeList {
 class UniqueDataRepository {
   const UniqueDataRepository._();
 
+  /// Find a list of [UniqueData]s from a table, using the provided [where]
```
tp, line 186: Probably should be `finds`, as most methods use that verb form.

Though I think even more common in Dart code is always starting the comment with `Returns …`.

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
```diff
@@ -209,6 +209,11 @@ class BuildRepositoryClass {
   Method _buildFindMethod(String className,
       Iterable<SerializableModelFieldDefinition> objectRelationFields) {
     return Method((m) => m
+      ..docs.add(
+          '/// Find a list of [$className]s from a table, using the provided [where]\n'
+          '/// expression, optionally using [limit], [offset], and [orderBy]. To order by\n'
+          '/// multiple columns, use [orderByList]. If [where] is omitted, all rows in\n'
+          '/// the table will be returned.')
```
tp, line 216: This reads a little convoluted, as it seems like `where` was made optional later on.



```dart
  /// Returns a list of [$className]s matching the given query parameters.
  /// 
  /// Use [where] to specify which items to include in the return value.
  /// If none is specified, all items will be returned.
  ///
  /// To specify the order of the items use [orderBy] or [orderByList]
  /// when sorting by multiple columns.
  ///
  /// The maximum number of items can be set by [limit]. If no limit is set,
  /// all items matching the query will be returned.
  ///
  /// [offset] defines how many items to skip, after with [limit] (or all)
  /// items are read from the database.
```

Tried to align that with my experience reading Dart comments and https://dart.dev/effective-dart/documentation#do-use-prose-to-explain-parameters-return-values-and-exceptions (which does not give very many examples though, so a bit of judgement was necessary). I could not find a relevant similar cases of ORMs that might give some hint of how to best document such a flexible method.

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
```diff
@@ -307,6 +312,8 @@ class BuildRepositoryClass {
     Iterable<SerializableModelFieldDefinition> objectRelationFields,
   ) {
     return Method((m) => m
+      ..docs.add(
+          '/// Find a single [$className] from a table, using the provided [where]')
```
tp, line 316: Depending on what we decide to do with `find` we should copy the same comment over (+ adjustments for the single return value of course).

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
```diff
@@ -893,6 +928,8 @@ class BuildRepositoryClass {
       var relation = field.relation as ListRelationDefinition;
 
       methodBuilder
+        ..docs
```
tp, line 931: I hope I understood the attach/detach behavior here correctly.

Was not sure why that was provided as a special implementation, instead of just using `update(model.copyWith(referenceType: null))`. Would need to read up on the original ticket, in order to provide a more useful comment here.

modules/serverpod_auth/serverpod_auth_server/lib/src/generated/email_auth.dart
```diff
@@ -272,6 +281,7 @@ class EmailAuthRepository {
     );
   }
 
+  /// Inserts a single [EmailAuth] and returns the inserted row.
```
tp, line 284: here we should hint that only the returned entity will have the entity.

The same also probably applies for attach/detach.

tests/serverpod_test_server/lib/src/generated/unique_data.dart
Isakdl: Agreed! update it please :)

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
Isakdl: Yes, your suggestion is much better, please feel free to reword all documentation comments here. But apply it to the original and the new one!

modules/serverpod_auth/serverpod_auth_server/lib/src/generated/email_auth.dart
Isakdl: Agreed!

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
Isakdl: In addition it could be nice with some example code too, this is quite common in other dartdocs comments in flutter/dart.

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
Isakdl: In essence, it is a convenience method for exactly that yes. But there is a special case where this method will be the only way to add a relation.

We have support for what we call an "implicit" one-to-many where you only define the many side (e.g. a list). This will add a hidden column on the other model that is only accessible in the database layer. To be able to still "attach" objects to the list in the database, you have to use this attach/detach method.

Isakdl (commented):
Good job! I agree with all your suggestions for improving the docs comments, I think you can go ahead and use your best judgement here. From the suggestions we are getting a huge improvement. Do one loop on that and we should be good to go.

For the commit history, I would recommend committing the generated code in a standalone commit, (e.g. with no manual edits). This way it will be easier for me to filter out the changes when reviewing.

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
tp: > In addition it could be nice with some example code too, this is quite common in other dartdocs comments in flutter/dart.

That would be super nice, but I think the most interesting one would the `where` clause (with the callback), but since that depends on the fields on the model and the types of the field to be used, that might get tricky to generate.

Or should we use a generic example like `Person` that most users could make sense of, e.g.

```dart
      var persons = await Persons.db.find(
        session,
        where: (t) => t.lastName.equal('Jones'),
        orderBy: (t) => t.firstName,
        limit: 100,
      );
```

Isakdl: Yeah doing custom examples seem pretty hard, but a static generic example is good enough I think. Still pretty valuable.

tp: Added static code examples to find*.

packages/serverpod/lib/src/database/database.dart
```diff
@@ -39,10 +39,19 @@ class Database {
   })  : _session = session,
         _databaseConnection = DatabaseConnection(poolManager);
 
-  /// Find a list of [TableRow]s from a table, using the provided [where]
-  /// expression, optionally using [limit], [offset], and [orderBy]. To order by
-  /// multiple columns, user [orderByList]. If [where] is omitted, all rows in
-  /// the table will be returned.
+  /// Returns a list of [$TableRow]s matching the given query parameters.
```
Isakdl, line 42: Suggestion: There shouldn't be a $ infront of TableRow.

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
```diff
@@ -210,10 +210,19 @@ class BuildRepositoryClass {
...
+          '/// If none is specified, all items will be returned.\n'
+          '///\n'
+          '/// To specify the order of the items use [orderBy] or [orderByList]\n'
+          '/// when sorting by multiple columns.\n'
+          '///\n'
+          '/// The maximum number of items can be set by [limit]. If no limit is set,\n'
+          '/// all items matching the query will be returned.\n'
+          '///\n'
+          '/// [offset] defines how many items to skip, after which [limit] (or all)\n'
+          '/// items are read from the database.')
```
Isakdl: Suggestion: You could also do this as a multiline string and you can remove the \n e.g: 

```dart
'''/// Returns a list of [$className]s matching the given query parameters.
///
/// Use [where] to specify which items to include in the return value.
/// If none is specified, all items will be returned.
'''
```

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
```diff
@@ -1285,7 +1302,7 @@ class BuildRepositoryClass {
 
       methodBuilder
         ..docs.add(
-            '/// Remove the `${field.name}` field from the [$classFieldName]')
+            '/// Removes the `${field.name}` field from the [$classFieldName]')
```
Isakdl, line 1327: Comment: Perhaps we should add something about the relation here too.

Isakdl (changes requested):
Very nice this is looking so much better than the previous docs comments! 🚀 

I made two smaller suggestions! Then we should be good to go!

packages/serverpod/lib/src/database/database.dart
tp: my bad, removed that.

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
tp: @Isakdl What do you have in mind?

A longer explanation (for which I might not be firm enough on the details), or just add "relation" to these 6 methods like 

```
'/// Removes the "`${field.name}`" relation from the [$classFieldName]'
```

(thus replacing the technical (or storage-oriented) "field" with the logical "relation".

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
tp: Right, in the end this will probably read better even if the first line will be off (let's see, maybe even having an empty-line start would work fine here, since the code should get formatted in the end and thus there would be no double blank line).

Will try this out.

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
tp: Updated the comments with the latest version we agreed upon in Slack.

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
tp: Switched to multiline. Looks much better indeed.

Isakdl (approved):
LGTM 🚀 

Great commit history in the latest revision too!

PR #3184, opened 2025-01-28
Add doc comments to generated `copyWith` methods

tp:
Annotate `copyWith` with `@useResult` to ensure that callers actually make use of the newly created object

Re-export of the `meta` `useResult` annotation was added to `serverpod` and `serverpod_client` such that consumers don't have to manually add a reference to `meta/meta` in their application project (as otherwise you'd get the warning that the generated code is importing a package which is not a dependency).

Relates to #1237

### DIFF
--- a/tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
@@ -1,7 +1,6 @@
 import 'package:code_builder/code_builder.dart';
 import 'package:path/path.dart' as p;
 import 'package:recase/recase.dart';
-
 import 'package:serverpod_cli/analyzer.dart';
 import 'package:serverpod_cli/src/analyzer/models/definitions.dart';
 import 'package:serverpod_cli/src/analyzer/models/utils/duration_utils.dart';
@@ -74,6 +73,10 @@ class SerializableModelLibraryGenerator {
             from: p.dirname(classDefinition.filePath),
           );
           libraryBuilder.directives.add(Directive.partOf(topNodePath));
+        } else if (!(classDefinition.isSealedTopNode &&
+            classDefinition.childClasses.isEmpty)) {
+          libraryBuilder.directives
+              .insert(0, Directive.import('package:meta/meta.dart'));
         }
 
         libraryBuilder.body.addAll([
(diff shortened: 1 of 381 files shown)


tools/serverpod_cli/test/generator/dart/client_code_generator/class_copy_with_test.dart
```diff
@@ -138,8 +137,12 @@ void main() {
           expect(copyWithMethod?.returnType?.toSource(), testClassName);
         }, skip: copyWithMethod == null);
 
+        test('annotated with @useResult', () {
```
tp, line 140: I see that this file also should be migrated to the current pattern, but I deemed this out of scope for now to get the quick-win in and focus on the other TODOs next.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -74,6 +73,10 @@ class SerializableModelLibraryGenerator {
             from: p.dirname(classDefinition.filePath),
           );
           libraryBuilder.directives.add(Directive.partOf(topNodePath));
+        } else if (!(classDefinition.isSealedTopNode &&
+            classDefinition.childClasses.isEmpty)) {
+          libraryBuilder.directives
+              .insert(0, Directive.import('package:meta/meta.dart'));
```
Isakdl: Question: Why the `classDefinition.isSealedTopNode` check here? I think an explicit test for this case would be nice!

tools/serverpod_cli/test/generator/dart/client_code_generator/class_copy_with_test.dart
Isakdl: It does actually have the description style, but perhaps a little too segmented in different layers.

Isakdl (approved):
LGTM!

But one comment on the commit history here. As I mentioned on the last PR, please do all the generated code in a single commit separated from the code that produces the generated code.

E.g.

- commit introducing changes in code generator
- commit generated dart code (handle migrations separately)

In this case I think you could have rebased the PR too as the first change was later reverted!

tools/serverpod_cli/test/generator/dart/client_code_generator/class_copy_with_test.dart
tp: Yeah, I just thought we would avoid `group` for what is really a (dependent) test, and `test` for individual assertions.

But happy to refactor this some other time, then we can clarify any open question in there :)

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
tp: There was some special logic to see when to add the import directive for `meta`, which must only be added if a `copyWith` was being implemented. Otherwise consumers would get a warning for an unused import.

But then I changed it so that `@useResult` is exported by the Serverpod libs, so that we don't get that warning on using `meta` in the generated code (which might not be a dependency of the consuming package). Now we don't need any check anymore, as the "Serverpod import" is available in all files.

Issue #1784, opened 2024-01-19
Add support for additional dart:core types


Isakdl:
Add support for additional dart core types in our serialization and model files

- [BigInt](https://api.dart.dev/stable/dart-core/BigInt-class.html) 
- [Iterable](https://api.dart.dev/stable/dart-core/Iterable-class.html) 
- [num](https://api.dart.dev/stable/dart-core/num-class.html) 
- [Record](https://api.dart.dev/stable/dart-core/Record-class.html) 
- [Set](https://api.dart.dev/stable/dart-core/Set-class.html)
- [Uri](https://api.dart.dev/stable/dart-core/Uri-class.html)

Isakdl:
Not necessarily, I mean the serialized version would still probably be a list, but an iterator can be much more than just a list. It could add value to be able to use an iterator on the class rather than a list and it shouldn't be too hard to implement.

In any case the above types are the ones we do not support that the Google maintained package: https://pub.dev/packages/json_serializable supports.

SandPod:
Working method:
In separate PRs for each type, first, add it so that it can be sent and received from an endpoint. After that, we can consider adding support for it in the model definition. But this will be a case-by-case decision.

Isakdl:
@SandPod Database support also needs to be considered even if they just map to an existing database type.

tp:
A few questions / comments about the remaining types to be implemented.

# `Iterable`

What would be the expectations of when to read the iterator? Personally I would've started out with consuming the iterator once when creating the model class. That means it would only ever be consumed once (per model class), and internally (to the model class and database) it could just behave like a `List<T>`.

We could of course also retain the `Iterable` itself on the model class, but then we might get different values at different points (depending on how it/what is implemented), or succeed once and fail another time (e.g. when saving it to the database vs. sending the seemingly "same" model to the client afterwards).

# `num`

Would you go for `double precision` in the database here, or some more specific type like `decimal` with a higher precision, such that we could best map all possible numeric Dart types?

How important would it be to retain the original "source" type when round-tripping under the hood? I wonder if a `3.0 (double)` might not become a `3 (int)`. But maybe that would be fine, as people who opted into this more flexible type will just be OK treating it as `3 (num)` all the time.

# Record

This should be fine overall, I'll just have to dig into how we handle the case when the framework things "the type of this field has changed".

I think we would need to be a lot more flexible at that stage then, to e.g. allow additions of `null` positional and/or named parameters.

[...]

PR #3186, opened 2025-01-28
Support type `Uri` in endpoint methods ("top-level" parameter / return value)

### LINKED ISSUE #426: No deserialization found for type `Uri?`
I tried uploading a file to `serverpod_cloud_storage`. It works fine. But when I try getting the public url, I ran into this exception.

### Debugger traces

```console
args: {
  "path" : "PhotoViewer.m",
  "auth" : "1:P4OpgPmsZ289mQnWII0cET17c5tW0CnJ",
  "method" : "getDownloadUrl"
}

e: FormatException (FormatException: No deserialization found for type Uri?)

endpoint: "assets"

method: "getDownloadUrl"
```

### Endpoint code

```dart

class AssetsEndpoint extends Endpoint {
  Future<String?> getUploadDescription(Session session, String path) async {
    try {
      session.log('Path: $path');
      return session.storage.createDirectFileUploadDescription(
        storageId: 'public',
        path: path,
      );
    } on Exception catch (e) {
      session.log(e.toString(), level: LogLevel.error);
      return null;
    }
  }

  Future<bool> verifyUpload(Session session, String path) async {
    return await session.storage.verifyDirectFileUpload(
      storageId: 'public',
      path: path,
    );
  }

  Future<Uri?> getDownloadUrl(Session session, String path) async {
    session.log('Path: $path');
    return session.storage.getPublicUrl(
      storageId: 'public',
      path: path,
    );
  }
}
```

### I use in flutter code

```dart
uploading
? const CircularProgressIndicator.adaptive()
: ElevatedButton(
    onPressed: () async {
      FilePickerResult? result =
          await FilePicker.platform.pickFiles(
        allowMultiple: false,
      );
      String? uploadDescription;
      if (result != null) {
        setState(() {
          uploading = true;
        });
        var name = result.files.single.name;
        uploadDescription = await AppConstants.client.assets
            .getUploadDescription(name);
        var bytes =
            File(result.files.single.path!).readAsBytesSync();
        bool success = false;
        if (uploadDescription != null) {
          var uploader = FileUploader(uploadDescription);
          var a = ByteData.sublistView(bytes);
          success = await uploader.uploadByteData(a);
          var uploaded =
              await AppConstants.client.assets.verifyUpload(name);
          var exists = await AppConstants.client.assets
              .checkFileExists(name);
          print('Uploaded: $uploaded');
          print('Exists: $exists');
          if (exists) {
            var a = await AppConstants.client.assets
                .getDownloadUrl(result.files.single.name);
            print(a?.toString());
            await Fluttertoast.showToast(
              msg: '$uploadDescription  $success',
              toastLength: Toast.LENGTH_SHORT,
              gravity: ToastGravity.BOTTOM,
              timeInSecForIosWeb: 5,
              backgroundColor:
                  success ? Colors.green : Colors.red,
              textColor: Colors.white,
              fontSize: 16.0,
            );
          }
          setState(() {
            uploading = false;
          });
        }
      }
    },
    child: const Text('Upload and get URL'),
  ),
```

### serverpod_cloud_storage table 

| id      | storageId    | path    | addedTime    | expiration    | byteData    | verified    |
| :-----: | :---: | :---: | :---: | :---: | :---: | :---: |
| 1 | public   | PhotoViewer.m   | 2022-12-03 21:39:10.476641   | NULL   | BINARY (13 KB)   | TRUE   |

tp:
Closes #426

Support for usage inside models and database storage will be done as part of #1784


tests/serverpod_test_server/test_e2e/connection_test.dart
```diff
@@ -818,6 +820,16 @@ void main() {
       var result = await client.basicTypes.testUuid(null);
       expect(result, isNull);
     });
+
+    test('Type Uri', () async {
+      var result = await client.basicTypes.testUri(uri);
+      expect(result, equals(uri));
+    });
```
Isakdl, line 827: Suggestion: Add the same test in the file `test/serialization_test.dart`

Isakdl: And while we are at it, lets break out these type e2e tests into a standalone file and update the test description. There are a lot of old tests in here I can see.

I think you can move the URI setup into the test too, instead of having it declared so far up in the code. Then it becomes easier to see what we are testing.

Isakdl (changes requested):
Nice job! One small requests on the tests + an addition of unit tests.

tests/serverpod_test_server/test_e2e/connection_test.dart
tp: > Suggestion: Add the same test in the file test/serialization_test.dart

Are you referring to
https://github.com/tp/serverpod/blob/f6b6b8983e50152f697fccbc897b994353cd5ae5/tests/serverpod_test_server/test/serialization_test.dart#L102

I wanted to add the support for "in model" later with the other ticket, but I could also just add it now.

Looking at places where for example UUID is tested, I saw though that I was missing one in https://github.com/tp/serverpod/blob/f6b6b8983e50152f697fccbc897b994353cd5ae5/packages/serverpod_serialization/test/decode_with_type_test.dart#L251 that I have just added.

Also moved the basic type tests into their own file and aligned the descriptions with the pattern.

Isakdl: Ahh you are right! Sorry I was getting a head of my self there! 

But good find on the other test file!

tests/serverpod_test_server/test_e2e/basic_types_test.dart
```diff
@@ -0,0 +1,160 @@
...
+  var client = Client(serverUrl);
+
+  test(
+      'Given the test server, when the global int is set to 10 and 1 is added, then 11 is returned',
+      () async {
+    await client.simple.setGlobalInt(10);
+    await client.simple.addToGlobalInt();
+    var value = await client.simple.getGlobalInt();
+    expect(value, equals(11));
+  });
```
Isakdl, line 17: Question: This doesn't seem like a types test, what are we testing here? I can see that it was there in the old test suite, I think we can simply move this one back and deal with it another day.

Isakdl (changes requested):
Very nice! Small little request, then we should be ready to merge.

tests/serverpod_test_server/test_e2e/basic_types_test.dart
tp: Right. I moved this back out.

Isakdl (approved):
LGTM 🚀

Issue #2512, opened 2024-07-19
DataDog implementation with Openmetrics or Opentelemetry


fstof:
**Is your feature request related to a problem? Please describe.**
Having insights in a standalone app is great and all, but once the app is deployed in the cloud its difficult / impossible to configure everything for the insights app to access the port because of company policies that only allow certain ports to be exposed.

**Describe the solution you'd like**
Having the current insights made available with Openmetrics or integrate with Opentelemetry. This will allow better integration with enteprise tools like DataDog

**Describe alternatives you've considered**
We did try to integrate this manually using  [prometheus_client](https://pub.dev/packages/prometheus_client) and [opentelemetry](https://pub.dev/packages/opentelemetry) but these packages are either not feature complete or just quite invasive to implement

**Additional context**
Having something like this natively integrated would be great

PR #3187, opened 2025-01-29
feat: Exception event hooks in Serverpod

### LINKED ISSUE #1352: Add exception handler to add support for things like Sentry server-side
We may want to add an exception handler that can be set to get notifications when an exception is thrown on the server side. This would make it easier to integrate with Sentry or other services for logging exceptions.

### LINKED ISSUE #3078: Add exceptionHandler to log exceptions
We should have a way to catch the (unhandled) exceptions and log them to your preferred crash logging service like sentry.

Related issues:
improving logs: https://github.com/serverpod/serverpod/issues/2847
slightly related to hooks: https://github.com/serverpod/serverpod/issues/3066

I think a seperate callback on the server object should be a good way to catch (unhandled) exceptions. Integrating this with the better loggin provided in https://github.com/serverpod/serverpod/issues/2847 we would have a more robust crashlogging system

```dart
  void _handleRequestWithErrorBoundary(HttpRequest request) async {
    // [Future.sync] ensures no synchronous error is accidentally thrown from the handler
    // https://dart.dev/libraries/async/futures-error-handling#solution-using-future-sync-to-wrap-your-code
    await Future.sync(() {
      return _handleRequest(request);
    }).catchError((e, stackTrace) async {
      stderr.writeln(
        '${DateTime.now().toUtc()} Internal server error. _handleRequest failed with exception.',
      );
      stderr.writeln('$e');
      stderr.writeln('$stackTrace');
      request.response.statusCode = HttpStatus.internalServerError;
      return request.response.close();
    });
  }
```

**Example**

```dart
  final pod = Serverpod(
    args,
    Protocol(),
    Endpoints(),
    authenticationHandler: auth.authenticationHandler,
    exceptionHandler: (session, error, trace) {
       //log to whatever service you want
    }
  );
```

params we would need for the handler
- error
- trace
- session (if the error came from an endpoint?)
- request (if the error came from an endpoint?)

I was thinking do we only want unhandled exceptions? Or also the other ones that are following this structure? -> https://docs.serverpod.dev/concepts/exceptions

christerswahn:
Adds support for exception event hooks throughout the lifecycle of a Serverpod server, both for application-code generated exceptions and exceptions raised within the framework.

This may be used to get all exceptions reported to services such as Sentry. It is easy to implement handlers and to implement custom filters in them.

Any number of hooks can be added. They are run asynchronously and should not affect the behavior or response times of the server.

Event hooks are for diagnostics only, they do not allow any behavior-changing action such as suppressing exceptions or converting them to another type.

Closes #3078 
Closes #1352 

> This feature does not address the general transaction monitoring / trace spans telemetry (referenced in issue #2512). Supporting that will require a tighter coupling with the request dispatching code and the lifecycle of the Session.

## Pre-launch Checklist

## Breaking changes

This does not introduce breaking changes.

## Experimental

This is introduced as an experimental feature. It is opt-in by providing event handlers to the Serverpod constructor. 

The API should currently be regarded as unstable which means there may be breaking changes to it between minor versions of the Serverpod package. To make this clear, the important parameters and methods in Serverpod class have their names prefixed with `unstable`.

When this feature and its interface has matured, proper names without `unstable` will be introduced. If possible, the `unstable` names will remain for some time as `@deprecated`, and then removed.

## Interface

The Serverpod constructor now accepts the named argument unstableDiagnosticEventHandlers which is an optional array of diagnostic event handlers.

Example:

```dart
  var serverpod = Serverpod(
    ...
    experimentalFeatures: ExperimentalFeatures(
      unstableDiagnosticEventHandlers: [
        AsEventHandler((event, space, {required context}) {
          print('$event  Origin is $space\n  Context is ${context.toJson()}');
        }),
      ],
    ),
  );
```

## Guidelines for handlers

An [DiagnosticEvent] represents an event that occurs in the server.
[DiagnosticEventHandler] implementations can react to these events
in order to gain insights into the behavior of the server.

As the name suggests the handlers should perform diagnostics only,
and not have any responsibilities that the regular functioning
of the server depends on.

The registered handlers are typically run concurrently,
can not depend on each other, and asynchronously -
they are not awaited by the operation they are triggered from.

If a handler throws an exception it will be logged to stderr
and otherwise ignored.


packages/serverpod/lib/src/relic/web_server.dart
```diff
@@ -80,12 +80,12 @@ class WebServer {
...
           ),
       };
-    } catch (e) {
-      stderr.writeln(
-        '${DateTime.now().toUtc()} ERROR: Failed to bind socket, Webserver '
-        'port $_port may already be in use.',
+    } catch (e, stackTrace) {
+      logError(
+        'Failed to bind socket, '
+        'Webserver port $_port may already be in use.',
```
SandPod: Suggestion(NIT): Should we remove `Webserver` here since it is already reflected in the scope of the message logged?

packages/serverpod/lib/src/server/diagnostic_events/event_function.dart
```diff
@@ -0,0 +1,45 @@
...
+/// Function signature type for functions that handle a [DiagnosticEvent].
+///
+/// Use together with [AsEventHandler] to adapt such a function to a
+/// [DiagnosticEventHandler].
+///
+/// {@macro diagnostic_event_handler}
+typedef EventHandlerFunction<T extends DiagnosticEvent> = void Function(
+  T event,
+  OriginSpace space, {
+  required DiagnosticEventContext context,
```
SandPod, line 12: Question(For my knowledge): Why is context a required named parameter as opposed to just a required parameter?

packages/serverpod/lib/src/server/diagnostic_events/event_function.dart
```diff
@@ -0,0 +1,45 @@
...
+  /// The function that handles the event.
+  final EventHandlerFunction<T> handler;
+
+  /// Optional filter to only handle events of a specific [OriginSpace].
+  final OriginSpace? spaceFilter;
+
+  /// Creates a new [AsEventHandler] wrapping .
+  const AsEventHandler(
+    this.handler, [
+    this.spaceFilter,
```
SandPod, line 28: Question: Do we envision this class to grow with other optional filters? If so, would it make sense to use a named optional paramter instead?

packages/serverpod/lib/src/server/diagnostic_events/event_handler.dart
```diff
@@ -0,0 +1,165 @@
...
+///
+/// [DiagnosticEventHandler] implementations are typically provided by the
+/// developer to the [Serverpod] constructor upon startup.
+///
+/// Implementations can handle [DiagnosticEvent] instances of a specific
+/// type or context, or all events.
+///
+/// ## Guidelines
+///
+/// An [DiagnosticEvent] represents an event that occurs in the server.
```
SandPod, line 92: Suggestion(NIT): Should this be `A [DiagnosticEve...`?

packages/serverpod/lib/src/server/diagnostic_events/event_handler.dart
```diff
@@ -0,0 +1,165 @@
...
+/// Handlers should not run for an extended period of time.
+/// Permitting that would risk accumulating resource consumption
+/// and make the long-running performance of the server degrade.
+/// This is especially true in situations of high load
+/// or implementation issues causing a large number of events.
+///
+/// By default they are given a maximum of 30 seconds to complete.
+/// This can be overridden by setting the diagnostic handler timeout
+/// in the constructor of the [Serverpod] instance.
+/// {@endtemplate}
```
SandPod, line 118: Question: Could there be a need to have this configurable per handler?

packages/serverpod/lib/src/server/diagnostic_events/operation_context.dart
```diff
@@ -0,0 +1,198 @@
...
+
+  @override
+  String toString() {
+    return '$runtimeType($serverName, $serverId, $serverRunMode, '
+        '$operationType, $sessionId, $userAuthInfo)';
+  }
+}
+
+/// The context of a future call operation.
+base class FutureCallOpContext extends OperationEventContext {
```
SandPod, line 71: Question/Discussion: Should the leaf implementations of `OperationEventContext` be final? That way we get more freedom as library developers to evolve these classes.

packages/serverpod/lib/server.dart
```diff
@@ -12,3 +12,8 @@ export 'src/server/serverpod.dart' hide ServerpodInternalMethods;
 export 'src/server/session.dart' hide SessionInternalMethods;
 export 'src/server/command_line_args.dart' show ServerpodLoggingMode;
 export 'src/config/security_context_config.dart' show SecurityContextConfig;
+export 'src/server/diagnostic_events/connection_info.dart';
```
SandPod, line 15: Question: What is the reason for not exporting the "barrel" file and hide the specific import that is not wanted?

packages/serverpod/lib/src/server/session.dart
```diff
@@ -69,6 +69,12 @@ abstract class Session implements DatabaseAccessor {
     return _authenticated;
   }
 
+  /// The authentication information for the session, if set.
+  /// This will be null if the session is not authenticated or not initialized.
+  AuthenticationInfo? get authInfoOrNull {
```
SandPod, line 74: Suggestion: Remove this as a public interface.

I don't think this is something we should publically expose.

What is this used for? When trying to access "internal" fields, we have added extensions that are not exported instead. Would that pattern be possible to use here as well?

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -294,18 +306,57 @@ class Serverpod {
     this.httpResponseHeaders = _defaultHttpResponseHeaders,
     this.httpOptionsResponseHeaders = _defaultHttpOptionsResponseHeaders,
     SecurityContextConfig? securityContextConfig,
-  }) : _securityContextConfig = securityContextConfig {
+    List<DiagnosticEventHandler>? unstableDiagnosticEventHandlers,
```
SandPod, line 309: Suggestion: Create an optional parameter named `experimentalFeatures` and add `diagnosticEventHandlers` there.

My main concern here is polluting the consturctor interface, if we would create some form of wrapper, we would be free to add any number of experiments without changing the constructor.

I also think we should use the term `experimentalFeatures` since that is what we have used previously to communicate that we are testing something out, both for the test framework and for the ongoing inheritance feature.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -881,3 +969,50 @@ extension ServerpodInternalMethods on Serverpod {
...
+    var futures = handlers.map((handler) => Future(
+          () => handler.handleEvent(event, space, context: context),
+        ));
+
+    var to = timeout;
+    if (to != null) {
+      futures = futures.map((future) => future.timeout(to));
+    }
+
+    futures.wait.onError((ParallelWaitError e, stackTrace) {
```
SandPod, line 1005: Question: Is this on error triggered for all of them or only for the first one? I remember we stumbled a bit on that once before.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -819,6 +897,16 @@ class Serverpod {
     }
   }
 
+  /// Submits an event to registered event handlers.
+  /// They will execute asynchrously.
+  void submitEvent(
```
SandPod: Suggestion: Mark this as experimental or unstable.

Serverpod is a public class and is always accessible, and this will become a public method. Let's mark it as part of the experiment until we know that it should be preserved.

We also might want to think about the naming of "Server Events" in our documentation (feature is named messages) and this.

packages/serverpod/lib/src/server/server.dart
```diff
@@ -520,19 +526,40 @@ class Server {
...
-        result,
-        sendByteDataAsRaw: methodCallContext.endpoint.sendByteDataAsRaw,
-      );
+        return ResultSuccess(
+          result,
+          sendByteDataAsRaw: methodCallContext.endpoint.sendByteDataAsRaw,
+        );
+      } catch (e, stackTrace) {
+        // Note: In case of malformed argument, the method connector may throw,
+        // which may be argued is not an "application space" exception.
```
SandPod, line 554: Question: Is it possible for us to add a specific exception to catch that and then report it correctly?

packages/serverpod/lib/src/relic/web_server.dart
```diff
@@ -186,6 +193,31 @@ class WebServer {
...
+
+    serverpod.submitEvent(
+      ExceptionEvent(e, stackTrace, message: message),
+      space,
+      context: session != null
+          ? contextFromSession(session, httpRequest: httpRequest)
+          : httpRequest != null
+              ? contextFromHttpRequest(
+                  serverpod.server, httpRequest, OperationType.web)
+              : contextFromServer(serverpod.server),
```
SandPod, line 217: Suggestion(NIT): Resolve this before submitting the event for readability :)

packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
```diff
@@ -170,10 +176,40 @@ abstract class EndpointWebsocketRequestHandler {
...
+
+    server.serverpod.submitEvent(
+      ExceptionEvent(e, stackTrace, message: message),
+      space,
+      context: session != null
+          ? contextFromSession(session, httpRequest: httpRequest)
+          : httpRequest != null
+              ? contextFromHttpRequest(
+                  server, httpRequest, OperationType.stream)
+              : contextFromServer(server),
```
SandPod, line 212: Suggestion(NIT): Resolve this before calling submit event for more readable code.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -232,17 +260,25 @@ class MethodWebsocketRequestHandler {
...
+        server,
+        e,
+        stackTrace,
+        message: 'Failed to parse arguments for open stream request: $message',
+        streamCommandMessage: message,
+        webSocketIntermediary: webSocket,
       );
+      // server.serverpod.logVerbose(
+      //   'Failed to parse arguments for open stream request: $message ($e)',
+      // );
```
SandPod, line 281: Suggestion(NIT): Remove

SandPod: Same for the other cases in this file.

SandPod (commented):
Awesome work on this! 🌟 

Overall, it looks great! This is going to be a great addition!

Tomorrow morning I'll take a second look at the MethodStream class just to make sure we covered all the cases there, and I didn't have time to look at the last test commit either.

I mostly had a couple of questions and some minor suggestions, perhaps the bigger one was around how we take in the handlers and the naming we should use.

packages/serverpod/lib/src/server/serverpod.dart
SandPod: This was fixed in a later commit :)

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
```diff
@@ -0,0 +1,267 @@
...
+        ),
+      );
+      pod = IntegrationTestServer.create(
+        unstableDiagnosticEventHandlers: [exceptionHandler],
+        config: config,
+      );
+      pod?.webServer.addRoute(ExceptionRoute(), '/exception');
+      final result = pod
+          ?.start(runInGuardedZone: false)
+          .timeout(const Duration(seconds: 5));
```
SandPod, line 63: Question: If this timeout is reached, what is the test output then?

If I remember correctly, when these timeouts throw, the print from the test is not very helpful. I think it needs to be handled explicitly to "fail" the test gracefully with valuable output.

SandPod: Same for the other potential timeout failures.

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
```diff
@@ -0,0 +1,267 @@
...
+      final record = await exceptionHandler.events.first.timeout(timeout);
+      expect(record.event.exception, isA<Exception>());
+      expect(record.space, equals(OriginSpace.framework));
+      expect(record.context, isA<DiagnosticEventContext>());
+      expect(
+          record.context.toJson(),
+          allOf([
+            containsPair('serverId', 'default'),
+            containsPair('serverName', 'Server default'),
+            containsPair('serverRunMode', 'production'),
```
SandPod, line 75: Question: In the test we configure the `runMode` to be `invalid`, why is it `production` in the diagnostics event?

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
```diff
@@ -0,0 +1,267 @@
...
+        config: config,
+      );
+      pod?.webServer.addRoute(ExceptionRoute(), '/exception');
+      final result = pod
+          ?.start(runInGuardedZone: false)
+          .timeout(const Duration(seconds: 5));
+      await expectLater(result, throwsA(isA<ExitException>()));
+
+      final record = await exceptionHandler.events.first.timeout(timeout);
+      expect(record.event.exception, isA<Exception>());
```
SandPod, line 67: Question: We don't seem to have any tracability on where the exception originated from, wouldn't it be valuable to know that this came from the web server?

Maybe that is supposed to be reflected in the `serverName`?

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
```diff
@@ -0,0 +1,267 @@
...
+            containsPair('endpoint', 'exceptionTest'),
+            containsPair('methodName', 'throwNormalException'),
+          ]));
+    });
+
+    test(
+        'when a client calls streaming method outStreamThrowsException '
+        'then the diagnostic event handler gets called', () async {
+      final stream = client.methodStreaming.outStreamThrowsException();
+      // print('Stream: ${await stream.first}');
```
SandPod, line 165: Suggestion(NIT): Remove comment.

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
```diff
@@ -0,0 +1,267 @@
...
+
+    test(
+        'when a client calls method url with malformed json '
+        'then the diagnostic event handler gets called', () async {
+      var response = http.post(
+        Uri.parse('http://localhost:8080/simple/hello'),
+        body: '{"name": [42]}',
+      );
+      await response;
+      // print('Response: ${response.statusCode} ${response.body}');
```
SandPod, line 223: Suggestion(NIT): Remove comment.

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
```diff
@@ -0,0 +1,267 @@
...
+    test(
+        'when a client calls web url with malformed json '
+        'then the diagnostic event handler gets called', () async {
+      var response = http.post(
+        Uri.parse('http://localhost:8082/exception'),
+      );
+      await response;
+
+      final record = await exceptionHandler.events.first.timeout(timeout);
+      // print(record.toString());
```
SandPod, line 261: Suggestion(NIT): Remove comment.

packages/serverpod/lib/server.dart
christerswahn: I wanted to make a distinction between the "public" exports here and all the exports including the internal ones (in the directory's own barrel file). There is only one file that is "internal", still I'd like to not make that part of the official interface.

packages/serverpod/lib/src/relic/web_server.dart
christerswahn: :+1:

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -351,6 +433,7 @@ class MethodWebsocketRequestHandler {
 
     methodStreamManager.createStream(
       session: session,
+      httpRequest: webSocket.httpRequest,
```
SandPod, line 436: Suggestion: Store as a final field of the MethodStreamManager.

The HTTP request will be the same for all calls to the method stream manager, since a single manager is created for each websocket connection.
We might as well store it as an optional final field inside of the manager, then we don't need to pass it around on the inside from the `createStream` method call.

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
```diff
@@ -232,17 +260,25 @@ class MethodWebsocketRequestHandler {
...
+    } catch (e, stackTrace) {
+      _reportFrameworkException(
+        server,
+        e,
+        stackTrace,
+        message: 'Failed to parse arguments for open stream request: $message',
+        streamCommandMessage: message,
+        webSocketIntermediary: webSocket,
       );
+      // server.serverpod.logVerbose(
```
SandPod, line 279: Discussion: What do we think about the verbose logging here? By default nothing will be logged now, assuming we remove the commented out code.

packages/serverpod/lib/src/server/diagnostic_events/event_handler.dart
christerswahn: Yes, not sure what you mean?

SandPod (commented):
Found some minor things with the tests and method stream manager on my second pass. But otherwise it looks solid! :)

packages/serverpod/lib/src/server/diagnostic_events/event_handler.dart
christerswahn: Sure! The central timeout config is open for continued design. And it is easy to implement a handler wrapper that does this (just like a handler filter). We could for example include some handler examples in a kind of toolbox.

packages/serverpod/lib/src/server/diagnostic_events/event_function.dart
christerswahn: This function follows the signature of handleEvent() in DiagnosticEventHandler exactly. If one should change then both should.

Dart guidelines advice against using too many positional parameters. In addition, for public APIs, named parameters make evolution of the interface a lot easier - adding new parameters, deprecating and maybe removing old ones, order being independent, and so on. For me the question is perhaps whether space should also be a named parameter...

packages/serverpod/lib/src/server/diagnostic_events/event_function.dart
christerswahn: Not really, and I'm on the fence if this class should even do space filtering. It's generally better if a class does a single thing, especially API "toolbox helpers" such as this.

packages/serverpod/lib/src/server/diagnostic_events/operation_context.dart
christerswahn: Agree!

packages/serverpod/lib/src/server/server.dart
christerswahn: I'm not sure, it seems to me a handful of Dart language errors might be thrown by the runtime, and they might also be thrown by user code. I discussed it with @Isakdl and the solution would either be

- Implement our own full argument and value validation (this is the ideal, but quite a bit of work)

- Analyze the stack trace here to determine if it originated from our generated method-invoking code

For now we said that we ignore this since it should be very rare. If using our generated client it would be our bug and hopefully rare. If they are hand-crafting URL requests then this might happen but in worst case this error is simply reported with the wrong space value.

packages/serverpod/lib/src/server/serverpod.dart
christerswahn: I like the idea with bundling the experimental stuff!

Regarding naming there is a subtle distinction between "experimental feature" and "unstable interface". This is an experimental feature yes, AND it's interface is unstable, so it's clear when naming the parameters and methods from user code that they might change in the next minor version. next.js uses `unstable` in the naming, for example.

packages/serverpod/lib/src/server/serverpod.dart
christerswahn: It is triggered if at least one throws. All are still run. From the `.wait` documentation:

If any future completes with an error, the returned future completes with a [ParallelWaitError].

packages/serverpod/lib/src/server/session.dart
christerswahn: It is used when submitting exception events to include the auth state IF if it is available, but not trigger anything such as initialization.

christerswahn: This needs to be accessed from a couple of event handling source files. An extension in a separate file can't access private members. Do you have an idea how to make this "public" but for internal framework use only?

packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
christerswahn: :+1:

packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
christerswahn: I forgot to remove the old commented-out code.

_reportFrameworkException will perform the same logVerbose(). A goal here was to change the existing behavior as little as possible.

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
christerswahn: Good point, I'll add conditions for that.

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
christerswahn: The runmode was something I experimented with and forgot to remove. (Invalid runmode values just result in production being set.)

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
christerswahn: Here we get the following, which is quite clear IMHO:

```
Expected: throws <Instance of 'ExitException'>
  Actual: <Instance of 'Future<void>'>
   Which: threw TimeoutException:<TimeoutException after 0:00:00.000000: Future not completed>
          stack test_integration/diagnostics/diagnostic_events_test.dart 65:7  main.<fn>.<fn>
                
          which is not an instance of 'ExitException'

test_integration/diagnostics/diagnostic_events_test.dart 65:7  main.<fn>.<fn>
```

packages/serverpod/lib/src/server/diagnostic_events/event_handler.dart
Isakdl: Grammer in the comment is **An** and not **A**

packages/serverpod/lib/src/server/session.dart
Isakdl: At the bottom of this file there is an extension called `SessionInternalMethods`, this extension is hidden from the global export, effectively enabling "protected" methods/fields.  E.g. they are visible for internal use, but not exported.

packages/serverpod/lib/src/server/server.dart
Isakdl: Yeah I'm advocating for fixing this problem with the Relic integration and implementing proper input validation at the endpoint level. Meaning we can deal with malformed requests before we pass it on to user land.

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
christerswahn: In the other case when we don't have an expectLater on the result we get:
```
TimeoutException after 0:00:00.000000: Future not completed
test_integration/diagnostics/diagnostic_events_test.dart 62:22  main.<fn>.<fn>
```

I experimented a bit with getting better printout but wasn't able to do that without adding plenty of custom code for each occurrence unforunately.

packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
Isakdl: Second on that, I don't think a ternary operator should ever be nested as it becomes very hard to read.

packages/serverpod/lib/src/server/diagnostic_events/event_handler.dart
christerswahn: Of course, thanks!

packages/serverpod/lib/src/server/session.dart
christerswahn: Thanks!

packages/serverpod/lib/src/server/diagnostic_events/operation_context.dart
```diff
@@ -0,0 +1,198 @@
...
+    return {
+      ...super.toJson(),
+      'futureCallName': futureCallName,
+    };
+  }
+}
+
+/// The context of a client call operation.
+/// Subclasses represent web, method, and method streaming calls.
+base class ClientCallOpContext extends OperationEventContext {
```
Isakdl, line 96: Question/Discussion: Do we need different context objects for each call context? To me this almost looks like what the Session object should have been.

Would it be possible to consolidate to a single class? Is that even desirable?

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -794,17 +858,31 @@ class Serverpod {
...
+      submitEvent(
+        ExceptionEvent(e, stackTrace),
+        OriginSpace.framework,
+        context: DiagnosticEventContext(
+          serverId: serverId,
+          serverRunMode: server.runMode,
+          serverName: '',
+        ),
+      );
+      rethrow;
```
Isakdl: Question: Would it make sense to run each call in its own try-catch? If one of these fails we are not shutting down the other services gracefully. 

I think the future call manager in particular is a bad one to not shut down gracefully as we could lose work.

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
```diff
@@ -0,0 +1,268 @@
...
+      expect(record.event.exception, isA<SocketException>());
+      expect((record.event.exception as SocketException).message,
+          contains('Failed to create server socket'));
+      expect(record.space, equals(OriginSpace.framework));
+      expect(record.context, isA<DiagnosticEventContext>());
+      expect(
+          record.context.toJson(),
+          allOf([
+            containsPair('serverId', 'default'),
+            containsPair('serverName', 'Server default'),
```
Isakdl, line 77: Suggestion: I would argue for a group with a setup here and then test each slice on individual tests. 

The origin space could be wrong without anything else being wrong for example, and then it is nice to have an isolated test for that, becomes very clear what you broke when the test fails.

Isakdl (changes requested):
Good job, and super nice git history honestly the best one you made, great work on that was very easy to follow what is happening. 

Not too many comments from me here, just one on the test and some discussion point.

packages/serverpod/lib/src/server/serverpod.dart
christerswahn: We should run these using List.wait! Otherwise, if we get a SIGTERM and one takes a long time, the remaining ones won't have a chance to close down! But that's a separate PR.

packages/serverpod/lib/src/server/server.dart
```diff
@@ -569,6 +571,20 @@ class Server {
 
     _running = false;
   }
+
+  Future<void> _reportFrameworkException(
+    Object e,
+    StackTrace stackTrace, {
+    String? message,
+    HttpRequest? httpRequest,
```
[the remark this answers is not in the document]
christerswahn: Good catch! We should include it in the context if available, will fix.

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
christerswahn: Ok

Isakdl (approved):
LGTM

packages/serverpod/lib/src/server/diagnostic_events/event_function.dart
SandPod: Yes, I think perhaps that was the distinction I was after. Feels like "event" is the only parameter that is implicit to the method, the rest could probably be required named.

And yes, if this method changes the interface "DiagnosticEventHandler " would also need to change.

packages/serverpod/lib/src/server/diagnostic_events/event_function.dart
SandPod: Yes, it seems a bit out of place with the description. What would the recommended pattern be here if more filters are applied? A wrapper?

If so, should we apply that already?

packages/serverpod/lib/server.dart
SandPod: Would it make sense to create a "public" barrel file to make this easy to manage over time?

Seems like there might still be "internal" files even when the feature is done and this brings a lot of "implementation" detail into the public `server` export barrel.

packages/serverpod/lib/src/server/serverpod.dart
SandPod: I think "experimental" are implicitly "unstable" features. In our documentation, we also mention this explicitly.
Screenshot 2025-02-24 at 06 23 02

In the `next.js` documentation, It seems like the difference is that "unstable" could be used for production while experimental should not. It that distinction important for the state of this feature? If not, I would keep it simple and simply use experimental.

packages/serverpod/lib/src/server/serverpod.dart
SandPod: Yes, but is it repeatedly triggered if multiple of the futures throws?

My concern is that we might silence errors and only print a warning for the first Future that throws.

The documentation for `wait` states:

> If any future completes with an error, then the returned future completes with that error. If further futures also complete with errors, those errors are discarded.

https://api.flutter.dev/flutter/dart-async/Future/wait.html

That is not desired behavior right?

packages/serverpod/lib/src/server/server.dart
SandPod: Sounds reasonable!

For context, I think there are two ways this can fail. Either we don't recognize the type when deserializing and we will throw a `DeserializationTypeNotFoundException` from protocol.dart`.

Or the cast fails because we deserialized to a type the endpoint did not expect.

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
SandPod: Aah, TIL... ^^

packages/serverpod/lib/src/server/diagnostic_events/op_context_helpers.dart
```diff
@@ -0,0 +1,150 @@
+/// These helpers are for internal framework use only.
```
SandPod, line 1: Suggestion: Rename file to `operation_context_helpers.dart` to make it as simple as possible to find when searching for it.

SandPod (commented):
Last concern is the futures that might not be reported when they fail. Otherwise it is mostly NIT suggestions

packages/serverpod/lib/server.dart
christerswahn: Yes makes sense. Current pattern seems to be having the public barrel files next to this one, I'll put it there

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -294,18 +306,57 @@ class Serverpod {
     this.httpResponseHeaders = _defaultHttpResponseHeaders,
     this.httpOptionsResponseHeaders = _defaultHttpOptionsResponseHeaders,
     SecurityContextConfig? securityContextConfig,
-  }) : _securityContextConfig = securityContextConfig {
+    ExperimentalFeatures? experimentalFeatures,
```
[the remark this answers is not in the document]
christerswahn: It's a good idea. It would be like a staging area for new methods/members, and they may get promoted to first-class members after the experimental period.

However there is both an internal method and a public one for this experimental feature. The Experimental container is the public one - to be used by user code. Question is what to do with the internal "unstable" method.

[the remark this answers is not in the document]
christerswahn: I moved the internal method to the internal subclass of Serverpod.

packages/serverpod/lib/src/server/serverpod.dart
christerswahn: That's the wrong wait, it's this one:
https://api.flutter.dev/flutter/dart-async/FutureIterable/wait.html

Yes all the futures are completed.

SandPod: Aah, my mistake! thanks for the clarification!

packages/serverpod/lib/src/server/diagnostic_events/event_function.dart
christerswahn: Yes I'll streamline this now, and we can do wrappers separately!

packages/serverpod/lib/src/server/serverpod.dart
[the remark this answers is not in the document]
christerswahn: There's a snag: I want the `ExperimentalFeatures` container to be a setup container - it is constructed by the user and provided to Serverpod in the constructor. As a setup container it should not contain runtime methods IMO...

We could create a separate `ExperimentalApi` type with the getter `experimental` in `Serverpod` and which contains such methods?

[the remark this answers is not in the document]
christerswahn: (This applies to the next PR, which contains the API for user code!)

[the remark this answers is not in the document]
SandPod: Yeah, I though about that as well, we don't want to bundle configuration and the actual feature.

That sounds like a sensible solution to the problem! 👍

SandPod (approved):
LGTM! 👍 

Nice and clean work with a neat interface! 🚀

Issue #2964, opened 2024-11-17
Setup SSL Certificate for ServerPod


codeeuser:
if the server is running over https, we need SecurityContext on the ServerPod constructor.

We found the ServerPod constructor has missing the SecurityContext parameter for setup the https feature.

Package: packages/serverpod/lib/src/server/serverpod.dart
Code:-
  /// Creates a new Serverpod.
  Serverpod(
    List<String> args,
    this.serializationManager,
    this.endpoints, {
    ServerpodConfig? config,
    this.authenticationHandler,
    this.healthCheckHandler,
    this.httpResponseHeaders = _defaultHttpResponseHeaders,
    this.httpOptionsResponseHeaders = _defaultHttpOptionsResponseHeaders,
    this.securityContext, // <=Please add this
  }) {...

SandPod:
Additional requirements:
- The Security context should be configurable for each running server (web server, insights, and API server).
- If more web servers are added in the future, it should be possible to configure these without modifying the Serverpod constructor or breaking existing configurations.

PR #3188, opened 2025-01-29
feat: Implement `SecurityContext` for API, Web, and Insights Servers  

klkucaj:
### **PR Description:**  
- Introduced `SecurityContext` configuration for API, Web, and Insights servers to enhance security.  
- Updated necessary configurations to integrate `SecurityContext` seamlessly.  

Closes: #2964 

## Pre-launch Checklist

## Breaking changes


tests/serverpod_test_server/test_integration/ssl/test_certificate.pem
```diff
@@ -0,0 +1,20 @@
+-----BEGIN CERTIFICATE-----
+MIIDOTCCAiGgAwIBAgIUUmZAMo/6EO73Uh8UPTS0pTnUFwYwDQYJKoZIhvcNAQEL
+BQAwRTELMAkGA1UEBhMCQVUxEzARBgNVBAgMClNvbWUtU3RhdGUxITAfBgNVBAoM
+GEludGVybmV0IFdpZGdpdHMgUHR5IEx0ZDAeFw0yNTAxMjgxMjUyMzJaFw0yNjAx
+MjgxMjUyMzJaMEUxCzAJBgNVBAYTAkFVMRMwEQYDVQQIDApTb21lLVN0YXRlMSEw
+HwYDVQQKDBhJbnRlcm5ldCBXaWRnaXRzIFB0eSBMdGQwggEiMA0GCSqGSIb3DQEB
+AQUAA4IBDwAwggEKAoIBAQC1zLj0HJJXYu064zbphsMX1EtGYzTOKkKjuPG4eWKN
+ZLritjanmUcYdvBaEdSayHj9E6/TMrRIcqzomjlpz10WRSgO6tucCJ5nGVLMqnpP
```
Isakdl, line 8: Suggestion: Put these files in an `asset` folder or similar so we differentiate them from the tests them selfs.

tests/serverpod_test_server/test_integration/security_context_test.dart
```diff
@@ -0,0 +1,292 @@
...
+import 'package:serverpod_test_server/src/web/routes/root.dart';
+
+import 'package:serverpod_test_server/test_util/test_serverpod.dart';
+import 'package:test/test.dart';
+
+void main() {
+  const certificatePath = 'test_integration/ssl/test_certificate.pem';
+  const keyPath = 'test_integration/ssl/test_key.pem';
+
+  group('Given a api server', () {
```
Isakdl, line 13: Nit: Given **an** api server

packages/serverpod/lib/src/relic/web_server.dart
```diff
@@ -20,9 +20,13 @@ class WebServer {
   /// A list of [Route] which defines how to handle path passed to the server.
   final List<Route> routes = <Route>[];
 
+  /// Security context if the web server is running over https.
+  final SecurityContext? securityContext;
```
Isakdl, line 24: Question: Do we get any value from having this public? I think it might make more sense to just make it private.

tests/serverpod_test_server/test_integration/security_context_test.dart
```diff
@@ -0,0 +1,292 @@
...
+        await serverpod.shutdown(exitProcess: false);
+      });
+
+      test('then server is running over https', () {
+        expect(serverpod.server.securityContext, isNotNull);
+      });
+
+      test('then server can be accessed over https', () async {
+        var client = HttpClient();
+        client.badCertificateCallback = (cert, host, port) => true;
```
Isakdl, line 37: Suggestion: Considering that you are loading real certificates and keys I think it would make more sense to test the validation too and not just accept any certificate given.

E.g.

```dart
final context = SecurityContext();

context.setTrustedCertificates(...)
HttpClient(context: context);
```

tests/serverpod_test_server/test_integration/security_context_test.dart
```diff
@@ -0,0 +1,292 @@
...
+          Uri.parse('https://localhost:8080'),
+        );
+        var response = await request.close();
+
+        expect(response.statusCode, 200);
+      });
+
+      test('then server cannot be accessed over http', () async {
+        var client = HttpClient();
+        client.badCertificateCallback = (cert, host, port) => true;
```
Isakdl, line 49: Question: Is this callback needed in this case?

tests/serverpod_test_server/test_integration/security_context_test.dart
```diff
@@ -0,0 +1,292 @@
...
+      test('then server cannot be accessed over https', () async {
+        var client = HttpClient();
+
+        expect(
+          () async => await await client.getUrl(
+            Uri.parse('https://localhost:8080'),
+          ),
+          throwsA(isA<HandshakeException>()),
+        );
+      });
```
Isakdl, line 102: Comment: This test is basically the exact same as above but we are just not ignoring the error. Modifying the above test with the real cert this test makes more sense.

Isakdl (changes requested):
I have made a few suggestions! 

I think we have the coverage needed but would like to test with the real certificate flow in the client.

packages/serverpod/lib/src/relic/web_server.dart
klkucaj: Yes i need it for the tests and also we do the same thing here:
https://github.com/serverpod/serverpod/blob/1cda9faecdbf658b06c5595962bd15a8a843c32e/packages/serverpod/lib/src/server/server.dart#L66

which already exists.

Isakdl: Hmm, for the tests all you are doing is checking that it isn't null. I don't think that is necessary the important part is validating that we are using the cert, and you already have tests for that!

But does the securityContext contain anything useful to expose to the users? If so fair enough, but it not I think it will just pollute the interface.

tests/serverpod_test_server/test_integration/security_context_test.dart
klkucaj: Fixed!

packages/serverpod/lib/src/relic/web_server.dart
klkucaj: Okay, made these fields private

tests/serverpod_test_server/test_integration/security_context_test.dart
klkucaj: Which test above you are referring here?

tests/serverpod_test_server/test_integration/ssl/ssl_cert.dart
```diff
@@ -0,0 +1,95 @@
+import 'dart:convert';
+
+List<int> certChainBytes = utf8.encode('''
```
Isakdl, line 3: Nit: put these in an `abstract final class` to namespace the variables, this makes it more clear where the variable is coming from when you see it in the code.

E.g.

```dart

abstract final class Certificate {
   static List<int> certChanBytes = ....
}

Isakdl (changes requested):
Good job looks good! Just one small nit :)

tests/serverpod_test_server/test_integration/security_context_test.dart
Isakdl: disregard this

tests/serverpod_test_server/test_integration/ssl/ssl_cert.dart
klkucaj: Done

Isakdl (approved):
LGTM

PR #3189, opened 2025-01-29
Support `BigInt` type

tp:
Relates to #1784


tests/serverpod_test_server/lib/src/models/nullability.spy.yaml
```diff
@@ -24,6 +24,9 @@ fields:
   aUuid: UuidValue
   aNullableUuid: UuidValue?
 
+  aBigInt: BigInt
+  aNullableBigInt: BigInt?
```
Isakdl, line 28: Suggestion: Commit the yaml files separately from the generated code.

tests/serverpod_test_server/test/serialization_legacy_test.dart
```diff
@@ -291,6 +294,12 @@ void main() {
           UuidValue.fromString('a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11'),
           null,
         ],
+        aBigInt: BigInt.one,
+        aBigIntList: [BigInt.one, BigInt.two],
+        aListWithNullableBigInts: [BigInt.one, null],
+        aNullableBigIntList: [BigInt.one],
+        aNullableBigInt: BigInt.one,
+        aNullableListWithNullableBigInts: [BigInt.one, null],
```
Isakdl, line 302: Suggestion: I think you can remove this test and just disregard anything in here. These are legacy tests that we have not spent time on migrating yet, but we are consciously not adding new things in here! Instead all the tests inside `serialization_test.dart` should provide the full coverage for the types.

Isakdl (commented):
So far looking good.

There is one missing test though, and that is for the yaml parser. We have a special test suite that checks all valid types inside the cli (unit test). Should be trivial to add BigInt there.

tests/serverpod_test_server/lib/src/models/nullability.spy.yaml
tp: (It's a bit tricky to untangle this locally, as I first check everything working together and then do the 3 commits, but I also can't think of a better way right now (except maybe just 2 commits for YAML and "manual" code changes, and then the generated files – either way we'll have some "partial" commits in between that would not work by themselves).)

tests/serverpod_test_server/test/serialization_legacy_test.dart
tp: Reverted the addition to the legacy tests.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions/default.dart
```diff
@@ -309,6 +311,27 @@ class DefaultValueRestriction extends ValueRestriction {
     return errors;
   }
 
+  List<SourceSpanSeverityException> _bigIntValueValidation(
+    dynamic value,
+    SourceSpan? span,
+  ) {
+    if (value is Duration) return [];
```
tp, line 318:
```suggestion
    if (value is BigInt) return [];
```

Did not come up in my testing, but maybe this could happen?

packages/serverpod/test/database/columns/column_bigint_test.dart
```diff
@@ -0,0 +1,99 @@
+import 'package:serverpod/database.dart';
```
tp, line 1: There is also an integration test in `tests/serverpod_test_server/test_integration/column_operations/column_duration_test.dart` that should be added for `BigInt` as well.

tests/serverpod_test_client/test/copy_with_mutable_test.dart
```diff
@@ -96,6 +96,21 @@ void main() {
       );
```
tp, line 96: There is also the file `tests/serverpod_test_server/test/generated_methods/copy_with_mutable_test.dart` (test server vs. test client).


Since the models are not shared, we'll have to copy the same test over there as well.

tests/serverpod_test_server/test/serialization_test.dart
```diff
@@ -97,6 +97,34 @@ void main() {
     );
   });
 
+  test(
```
tp, line 100: There is also `from_json_test` which we should extend as well

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions/default.dart
tp: Fixed type check.

packages/serverpod/test/database/columns/column_bigint_test.dart
tp: Added integration test.

tests/serverpod_test_server/test/serialization_test.dart
tp: Added the `from_json` test case.

tests/serverpod_test_client/test/copy_with_mutable_test.dart
tp: Added new tests for `copyWith` on `BigInt`.

Isakdl (changes requested):
Good job, this is a huge one!

Looks solid overall I think but aside from the comments in the code, I would like to add tests in the follow files:

- serverpod/serverpod/packages/serverpod_serialization/test/encode_with_type_test.dart
- serverpod/serverpod/tools/serverpod_cli/test/database/migration/defaults/big_int_default_migration_test.dart (this file does not exist).
- serverpod/serverpod/tools/serverpod_cli/test/database/extentions/defaults/big_int_default_column_test.dart (the file does not exist).


Then for the future the types implementation, I think splitting this into multiple PRs would be helpful for the reviewing. Something like:

- Serialization
- Model
- Database

tests/serverpod_test_server/lib/src/models/types.spy.yaml
```diff
@@ -9,5 +9,6 @@ fields:
   aByteData: ByteData?
   aDuration: Duration?
   aUuid: UuidValue?
+  aBigInt: BigInt?
```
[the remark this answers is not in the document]
tp: Has been added to `types_list` and `types_map` spies (and respective tests) in a later commit.

tools/serverpod_cli/test/database/create_definition/defaults/bigint_default_definition_test.dart
```diff
@@ -0,0 +1,339 @@
...
+        );
+      });
+
+      test('then the table should have two columns', () {
+        var table = databaseDefinition.tables.first;
+        expect(
+          table.columns,
+          hasLength(2),
+        );
+      });
```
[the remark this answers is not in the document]
tp: > Question: Do you think this adds value to repeat? My guess is that there is no real risk of deviation of these fields for default values, but I could be wrong.

I think we could agree on not checking the table and column count as well as the name, as I also don't see why they should be impacted from the other parameters, and thus we need not "recheck" them over and over.

tests/serverpod_test_server/test_e2e/connection_test.dart
```diff
@@ -737,6 +737,7 @@ void main() {
       var dateTime = DateTime.utc(1976, 9, 10, 2, 10);
       var duration = const Duration(seconds: 1);
       var uuid = UuidValue.fromString('a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11');
+      var bigInt = BigInt.parse('18446744073709551615999');
```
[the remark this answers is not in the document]
tp: I think this is the only test that is executes `client -> server (endpoint) -> database` and back again.

We have some other partial tests (e.g. serialization), and an integration test for writing into the database (and querying with various filters), but I don't see any other test was would test the connection from the client model all the way to the database and back out.

Personally this is the test I gain most confidence from (wrongly?), and from my point of view the others all just test some smaller parts of this (which should obviously be correct when the E2E passes – assuming it is correctly asserting).

tests/serverpod_test_server/test_integration/column_operations/column_bigint_test.dart
```diff
@@ -0,0 +1,120 @@
...
+    test('when filtering using empty notInSet then all rows are returned.',
+        () async {
+      var result = await Types.db.find(
+        session,
+        where: (t) => t.aBigInt.notInSet({}),
+      );
+
+      expect(result.length, 4);
+    });
+  });
```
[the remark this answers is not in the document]
tp: These numeric operations are not supported right now. They are also not available on the column as it doesn't have the "number operations" mixed in.

I took a note to investigate, but since the data is not backed by a numeric field in PostgreSQL but rather a string, I think we'd have to "manually" add them and could not rely on the default implementation for numbers like e.g. `Duration(Column)` does.

tests/serverpod_test_server/test/defaults/bigint/bigint_default_mix_test.dart
```diff
@@ -0,0 +1,84 @@
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+import 'package:test/test.dart';
+
+void main() async {
```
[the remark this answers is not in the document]
tp: Yeah, I think overall (also referring to the other question for individual PRs) this is not ideal, but it's really tricky to get this locally, as I need to work on everything side-by-side and then pick the changes (core, tests, model generation) later into individual commits where it's easy to miss one or assign it wrongly (as e.g. the generated files are lying around from the start).

I think for the initial read-through some approach to read through by folder/file type (core, tests, gen) would be best, because as you also saw, looking at smaller commits is misleading as we're likely to both identify some issues which are then fixed in later commits (that you have not seen yet).

So ideally I would give you just 3 commits: core, tests, gen in the PR, but I don't see any way to do that manageably, as I can not assigned "later changes" an earlier commit (without a lot of extra work going back and then re-applying the other changes over and over).

tools/serverpod_cli/test/database/create_definition/defaults/bigint_default_definition_test.dart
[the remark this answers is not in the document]
Isakdl: Great lets remove them!

tests/serverpod_test_server/test_e2e/connection_test.dart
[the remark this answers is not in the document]
Isakdl: Ahh okay, I thought we had some other e2e test for this, but in that case could you extract the types checking for the e2e case into a different file, following the test styles!

tests/serverpod_test_server/test_integration/column_operations/column_bigint_test.dart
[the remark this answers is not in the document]
Isakdl: I see, hmm could you see if you can solve this in some clean way? I think it make sense to support this type of filtering for this type.

tp:
Thanks for the pointers, added the 4 missing tests:

* serverpod/serverpod/packages/serverpod_serialization/test/encode_with_type_test.dart
* serverpod/serverpod/tests/serverpod_test_server/test/generated_methods/to_json_for_protocol.dart
* serverpod/serverpod/tools/serverpod_cli/test/database/migration/defaults/big_int_default_migration_test.dart (this file does not exist).
* serverpod/serverpod/tools/serverpod_cli/test/database/extentions/defaults/big_int_default_column_test.dart (the file does not exist).

And will also check that those are present in the `Uri` PR.

tools/serverpod_cli/test/database/migration/defaults/bigint_default_migration_test.dart
```diff
@@ -0,0 +1,134 @@
+import 'package:serverpod_cli/analyzer.dart';
+import 'package:serverpod_cli/src/test_util/builders/database/column_definition_builder.dart';
+import 'package:serverpod_cli/src/test_util/builders/database/database_definition_builder.dart';
+import 'package:serverpod_cli/src/test_util/builders/database/table_definition_builder.dart';
+import 'package:serverpod_service_client/serverpod_service_client.dart';
+import 'package:test/test.dart';
+
+void main() {
```
tp, line 8: Not sure how useful this test is, as there does not seem to be anything tested from the `BigInt` part, except the the demo string is a "serialized `BigInt`" instead of any random text.

I think these tests only differ for the SQL column type (not Dart type), and thus this test is equivalent to the normal String (== `::text` column) test.

But since we also had this test for `Duration` (which is just a numeric SQL column, and not its own type), I put it in place for now.

tools/serverpod_cli/test/database/create_definition/defaults/bigint_default_definition_test.dart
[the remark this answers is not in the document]
tp: Removed the checks in all but the first test.

Maybe this would be a decent case for a test helper / custom matcher though, so we can have the check in each test, without all the repetition (even across cases here, as this is just a tiny "side assert" we need, but not the core of the logic under test): like `expect(databaseDefinition, isSingleColumnTestDefinition)`.

tests/serverpod_test_server/test_integration/column_operations/column_bigint_test.dart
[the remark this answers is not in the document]
tp: Sure, I can look into that. I would suggest doing it as a follow up though, as we probably can not re-use the current numeric comparison helper, as we'd need to build something custom for the "BigInt in strings", as the default would be this…

CleanShot 2025-01-31 at 12 45 40@2x

So I'll have to think about this, if this is solvable by checking the sign and length, and only using a string comparison in the DB after that.

Ideally I can find some prior art doing this in another DB project, so that we can be reasonably sure this won't have unanticipated edge-cases.

tests/serverpod_test_server/test_e2e/connection_test.dart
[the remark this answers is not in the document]
tp: Okay, I will move out the `Types`-roundtripping check to `database_roundtrip_test.dart`.

tests/serverpod_test_server/test_integration/column_operations/column_bigint_test.dart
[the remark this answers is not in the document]
Isakdl: That is fair, let's do it in a follow up.

tools/serverpod_cli/test/database/migration/defaults/bigint_default_migration_test.dart
Isakdl: You are probably right that we in practice already have coverage for this. But I think it makes sense to add for completeness.

tests/serverpod_test_server/test_e2e/database_roundtrip_test.dart
```diff
@@ -0,0 +1,259 @@
...
+
+Future<void> setupTestData(Client client) async {
+  await client.basicDatabase.deleteAllSimpleTestData();
+  await client.basicDatabase.createSimpleTestData(100);
+}
+
+void main() {
+  var client = Client(serverUrl);
+
+  group('Database', () {
```
Isakdl, line 13: Suggestion: Could you split out each type into individual tests, to align with the general test style. 

This will also be helpful when things break as we can see directly what is going wrong without having to inspect and understand these convoluted tests. 

Rename this file to include the word "type" like, `database_roundtrip_type_test.dart` or similar. 

Anything that is not a type test shouldn't be located in this file.

Isakdl (changes requested):
I got one more request on the last test, but I think that should be the last thing! For the query thing we can solve that in a follow up!

tests/serverpod_test_server/test_e2e/database_roundtrip_test.dart
tp: Broke out all the individual cases into distinct test files.

tests/serverpod_test_server/test_e2e/database_roundtrip_enum_test.dart
```diff
@@ -0,0 +1,39 @@
...
+      expect(returnedObject, isNotNull);
+      expect(returnedObject!.testEnum, equals(TestEnum.two));
+      expect(returnedObject.enumListList.length, equals(2));
+      expect(returnedObject.enumListList[0].length, equals(2));
+      expect(returnedObject.enumListList[0][0], equals(TestEnum.one));
+      expect(returnedObject.enumListList[0][1], equals(TestEnum.two));
+      expect(returnedObject.enumListList[1].length, equals(2));
+      expect(returnedObject.enumListList[1][0], equals(TestEnum.two));
+      expect(returnedObject.enumListList[1][1], equals(TestEnum.one));
+    });
```
Isakdl, line 37: Suggestion: Sorry to be annoying here, but I think this test is testing way to many different things. 

I would like each case to be split into its own isolated test. There are several problems with the current style.

First being that if this test fails on the first expect, we may still have other unrelated bugs and may fail on the next expect after fixing the first. Meaning we are not getting the full overview of what is broken when running the test suite.

It is harder to understand what we are testing, which we can also see on the test description on this test. This in turn makes it harder to know what failed when reading the test description, ideally I should understand what the bug is by just reading the description of a failing test.

Last point I like to treat my test like immutable code, if I add functionality in the production code I should not have to modify existing tests. All old behavior that is still expected to work should be covered and all I should have to do is add an additional test. This makes it easier to modify code as I don't have to understand every single previous case that was handled. It also minimizes the risk of modifying the wrong thing, creating a bug in the old test.

E.g. what I expect a test like this to look like is something like this:

```dart
test(
        'Given an enum when sending and writing it to the database then the returned value is unmodified',
        () async {
      var object = ObjectWithEnum(
        testEnum: TestEnum.two,
      );

      var result = await client.basicDatabase.storeObjectWithEnum(object);
      expect(result.testEnum, TestEnum.two);
}
```

Apply this logic on all test files in this commit.

Isakdl (changes requested):

tests/serverpod_test_server/test_e2e/database_roundtrip_enum_test.dart
tp: I adapted that test as well to the new pattern.

Since `ObjectWithEnum` has a bunch of non-nullable required fields, I added a `ObjectWithEnumBuilder` so that each test only needs to set the one field it cares about.

So if the object is ever changed in the future, we would only need to add a new default there for required fields.

(Speaking of the builders, I wondered whether we could just have a `ObjectWithEnum` "test default" instance instead, and then use the generate `copyWith`? As manually implementing the builder was a bit cumbersome, even for this small of a class, especially in these cases where it's likely single use and the builder won't offer real convenience methods (like some of the DB types do).)

tests/serverpod_test_server/test_e2e/database_roundtrip_operations_test.dart
```diff
@@ -0,0 +1,115 @@
+import 'package:serverpod_test_client/serverpod_test_client.dart';
+import 'package:serverpod_test_server/test_util/config.dart';
+import 'package:test/test.dart';
+
+Future<void> _setupTestData(Client client) async {
+  await client.basicDatabase.deleteAllSimpleTestData();
+  await client.basicDatabase.createSimpleTestData(100);
+}
+
+void main() {
```
Isakdl, line 10: Suggestion: Could you validate that we have this type of check in an integration test. I think an integration test is better and good enough for database operations.

Previously we did not have database query tests outside of end-to-end tests. But it is also hard to get full coverage of all different database queries when doing it end-to-end. Hence we have moved into putting this into integration tests, skipping the serialization in endpoints.

tests/serverpod_test_server/test_e2e/database_roundtrip_type_test.dart
```diff
@@ -0,0 +1,194 @@
...
+
+      types = await client.basicDatabase.insertTypes(types);
+      expect(types.id, isNotNull);
+
+      var newCount = await client.basicDatabase.countTypesRows();
+      expect(newCount, isNotNull);
+      expect(newCount, equals(count! + 1));
+    });
+
+    // TODO: Support ByteData in database store
```
Isakdl, line 25: Suggestion: fix this todo.

Isakdl (changes requested):
Some last comments on the tests!

tests/serverpod_test_server/test_e2e/database_roundtrip_enum_test.dart
```diff
@@ -0,0 +1,155 @@
...
+    var object = ObjectWithEnumBuilder().withTestEnum(TestEnum.two).build();
+
+    object = await client.basicDatabase.storeObjectWithEnum(object);
+    expect(object.id, isNotNull);
+
+    var result = await client.basicDatabase.getObjectWithEnum(object.id!);
+
+    expect(result, isNotNull);
+    expect(result!.testEnum, equals(TestEnum.two));
+  });
```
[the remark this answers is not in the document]
tp: > This seem like more of a database operation thing than an end-to-end test validation. 

I think it's just a safeguard someone put in place to it does not crash on `getObjectWithEnum(object.id!)` further down in case the object has not been round-tripped through the database.

> Every single test will fail with the same problem too, if we would stop storing the id.

True, but would that be so bad now to have 10 tests being failed for the same reason? The repetition is not ideal, granted, but overall I wouldn't see a problem to have many very related/similar tests fail if some implementation broke. In practice one would just go in, fix the issue in the first time and then re-run the file and see all of them passing.


Personally I would have to say I liked the previous "whole object tested in one batch" better, because at least I could see what the test did and be reasonably that all cases where covered by reading a single page. Now, as the test gets spread out further and further I would have to read a whole file (plus the test object definition and builder) to make sure that all the fields are tested at least once.

Likewise with dropping the extra read now, because that looks like a lot of repetition now that we have the tiny tests. I think before it was super clear that we did a write and DB-backed read (unless the server tricked us with in-memory storage…), and thus we could build a reasonable model of what the server was doing. If I switch to reading the returned object, I have to trust the server a bit more to actually return a persisted object instead of just returning the incoming one verbatim.

But if we say the implementation details of the server are always part of this test, and we don't want to check the entire outside behavior (that someone can actually retrieve a piece of data later by ID), I can shrink it down.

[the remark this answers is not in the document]
tp: > apply the changes you did in the last commit and this comment to all the test files database_roundtrip_*_test.dart

For the field scope test, we don't have a immediate return value from the server (`void`), so we'd have to change the server there as well to make the test just work with the immediate return value.


CleanShot 2025-02-03 at 11 40 25@2x

tests/serverpod_test_server/test_e2e/database_roundtrip_operations_test.dart
tp: I did not find a similar integration test to all the cases we did with `Types` here, so extended `tests/serverpod_test_server/test_integration/database_operations/crud/read_test.dart` in that regard.

tests/serverpod_test_server/test_e2e/database_roundtrip_enum_test.dart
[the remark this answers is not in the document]
Isakdl: For the test project, we should always consider the "production code" as part of the test. The intention is not to test our implementation in this project but rather that the framework behaves correctly. I mean arguably even doing an endpoint to db roundtrip might be overkill in the sense that we already have tests for the db side. The biggest issue is of course that we are hiding test details outside of the test. But this is unfortunately the only way to do a real e2e test for the endpoints. To me, the important part is that the endpoints are as simple as possible, ideally without logic. 

We want to make sure that the code generator is creating working code! With this said I think it is okay to assume that we know the code in the endpoint, as that is not the code we want to test here!

I can see where you are coming from with grouping the checks, but I think the drawback with that approach is with maintenance, it is generally harder to understand what you are testing when there are a lot of checks in a test. It generally is harder to make modifications to existing tests rather than add additional ones. It is generally easier to read/modify the code when all the code follows the same style. Hence why I argue for it here.

Hope that makes sense :)

tests/serverpod_test_server/test_e2e/database_roundtrip_type_test.dart
Isakdl: I think this should be fully supported! But looks scary if we do not have any tests in place! I prefer to have these "TODOs" as issues in the repo, over comments like this, as todos are easily forgotten! Feel free to create issues and delete todos if you run into them!

tp: Hmm, maybe the reason this wasn't done, was because it doesn't work yet.

At least my basic test cases result in a surprising behavior

CleanShot 2025-02-03 at 12 18 24@2x


CleanShot 2025-02-03 at 12 17 13@2x


Maybe I will have to carve this out into a follow-up, as I'll need to dig deeper to see what's going on here.

tp: Interestingly the E2E test going through the database and back out works, while the integration test going just into the database and back out does not 🤯

Maybe something in the extra to/from JSON that might be happening over the protocol corrects something 🤔 (Because the long data above looks like it would be looking at some encoded/wrapped info.)

tp: Ah, turns out that the `ByteData` in the "read from DB" case is pointing to a shared buffer, and my test so far was wrongly checking the whole buffer instead of just reading the assigned part (accessing which is trickier than I think such a standard operation should be, as this would also be what every consumer has to do).

With this updated assertion the DB integration tests now also passes:

```dart
    expect(
      Uint8List.view(
        retrieved!.aByteData!.buffer,
        retrieved.aByteData!.offsetInBytes,
        retrieved.aByteData!.lengthInBytes,
      ).toList(),
      [1, 2, 3],
    );
```

But I wonder if we couldn't provide a better "wrapper" on that field, such that such a mixup would not happen.

tests/serverpod_test_server/test_e2e/database_roundtrip_enum_test.dart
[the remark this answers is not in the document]
tp: All good, I think we both agree that is a trade-off in both directions.

Now adapted as discussed :)

tests/serverpod_test_server/test_e2e/database_roundtrip_type_test.dart
Isakdl: Yeah, ByteData is easy to get wrong when it comes to the views. Maybe possible to make some abstraction but I have no good ideas for that 😅.

tests/serverpod_test_server/test_e2e/basic_types_test.dart
```diff
@@ -148,4 +148,20 @@ void main() {
...
   });
+
+  test(
+      'Given the test server, when a `BigInt`` is sent to the server, then it is returned verbatim',
+      () async {
+    var bigInt = BigInt.parse('-12345678901234567890');
+
+    var result = await client.basicTypes.testBigInt(bigInt);
+    expect(result, equals(bigInt));
+  });
```
Isakdl, line 159: Suggestion: I just realized that we have nothing like this for streams! The streaming serialization is also using a different code path for serialization under the hood so it is relevant to replicate these tests for them!

This means duplicating this file and test cases but using streams instead, there are not streaming endpoints like this today in the project!

An echo stream endpoint could be implemented like this:
```dart
Stream<int?> testInt(Session session, Stream<int?> value) async* {
    await for (var v in value) {
      yield v;
    }
  }
```

Isakdl (changes requested):
Looks good! 

But I just realized that we have one more case we haven't tested! That is streaming methods, no types tests exists for this today. Should be fairly straight forward to implement however, but please add that and then we can merge this! :)

tests/serverpod_test_server/test_e2e/basic_types_test.dart
tp: Got it, will add that :)

tp: Added test for the basic types streaming, so we have them all covered.

There were a few cases with a handful of types in https://github.com/tp/serverpod/blob/6d209b322398548cd4abd70f3d8f407018073253/tests/serverpod_test_server/test_e2e/method_streaming, but not all of them covered.

So good that you thought of that, so that we also have new types like `Duration`, `BigInt`, and `Uri` covered :)

PR #3195, opened 2025-01-31
fix: Cascade recreation of table to related tables.

### LINKED ISSUE #3130: Failing application of force migrations when table name changes
### Describe the bug

So far I've observed in many to many relationships like [official example](https://docs.serverpod.dev/concepts/database/relations/many-to-many) (student, course, enrollment (junction table)) the following error:
```sh
dart bin/main.dart --apply-migrations
SERVERPOD version: 2.3.1, dart: 3.6.1 (stable) (Tue Jan 7 09:50:00 2025 -0800) on "macos_arm64", time: 2025-01-16 22:14:21.717659Z
mode: development, role: monolith, logging: normal, serverId: default
SERVERPOD initialized, time: 2025-01-16 22:14:21.770151Z
Failed to apply migration 20250116221410499.
DatabaseQueryException: { message: constraint "enrollment_fk_0" of relation "enrollment" does not exist, code: 42704 }
Failed to apply database migrations. DatabaseQueryException: { message: constraint "enrollment_fk_0" of relation "enrollment" does not exist, code: 42704 }
Insights listening on port 8081
Server default listening on port 8080
Webserver listening on port 8082
```

If we create the initial migration with the tables mentioned above and then rename the `student` table to something else like `custom_student` and do the force migration (because the `student` table needs to be dropped in this case) the migration will fail to apply.

This happens because the force migration first drops the `student` table with `CASCADE` behaviour clause which will drop all the dependent objects of `student` table including but not limited to `enrollment_fk_0` (the foreign key in the enrollment junction table pointing to the primary key in the `student` table). But then it tries to drop the constraint to the `enrollment_fk_0` again in a later step.

Problematic part of the force migration:
```SQL
BEGIN;

--
-- ACTION DROP TABLE
--
DROP TABLE "student" CASCADE;

--
-- ACTION CREATE TABLE
--
CREATE TABLE "custom_student" (
    "id" bigserial PRIMARY KEY,
    "name" text NOT NULL
);

--
-- ACTION ALTER TABLE
--
ALTER TABLE "enrollment" DROP CONSTRAINT "enrollment_fk_0"; -- THIS IS THE PROBLEMATIC STEP THAT SHOULD NOT EXIST IN THIS CASE
ALTER TABLE ONLY "enrollment"
    ADD CONSTRAINT "enrollment_fk_0"
    FOREIGN KEY("studentId")
    REFERENCES "custom_student"("id")
    ON DELETE NO ACTION
    ON UPDATE NO ACTION;
```

### To reproduce

1. replicate [official Serverpod example for many to may relation](https://docs.serverpod.dev/concepts/database/relations/many-to-many)*
2. generate models and create initial migration and apply it
3. in the `student.spy.yaml` rename the `table` form `student` to `custom_student`
4. generate models and create migration with `--force` flag
5. run the server with `--apply-migration` flag
6. observe the error

\* ***Exact model files are:***
<details><summary>student.spy.yaml</summary>
<pre>
# student.yaml
class: Student
table: student
fields:
  name: String
  enrollments: List<Enrollment>?, relation(name=student_enrollments)
</pre>
</details>
<details><summary>course.spy.yaml</summary>
<pre>
# course.yaml
class: Course
table: course
fields:
  name: String
  enrollments: List<Enrollment>?, relation(name=course_enrollments)
</pre>
</details>
<details><summary>enrollment.spy.yaml</summary>
<pre>
# enrollment.yaml
class: Enrollment
table: enrollment
fields:
  student: Student?, relation(name=student_enrollments)
  course: Course?, relation(name=course_enrollments)
indexes:
  enrollment_index_idx:
    fields: studentId, courseId
    unique: true
</pre>
</details>

### Expected behavior

I would expect the ACTION ALTER TABLE step to look like this:

```SQL
BEGIN;

--
-- ACTION DROP TABLE
--
DROP TABLE "student" CASCADE;

--
-- ACTION CREATE TABLE
--
CREATE TABLE "custom_student" (
    "id" bigserial PRIMARY KEY,
    "name" text NOT NULL
);

--
-- ACTION ALTER TABLE
--
ALTER TABLE ONLY "enrollment"
    ADD CONSTRAINT "enrollment_fk_0"
    FOREIGN KEY("studentId")
    REFERENCES "custom_student"("id")
    ON DELETE NO ACTION
    ON UPDATE NO ACTION;
```

### Library version

Dart SDK version: 3.6.1 (stable) (Tue Jan 7 09:50:00 2025 -0800) on "macos_arm64"
Serverpod version: 2.3.1

### Platform information

```sh
                    'c.          kristijanzic@MacBookPro
                 ,xNMM.          -----------------------
               .OMMMMo           OS: macOS 15.2 24C101 arm64
               OMMM0,            Host: MacBookPro18,3
     .;loddo:' loolloddol;.      Kernel: 24.2.0
   cKMMMMMMMMMMNWMMMMMMMMMM0:    Uptime: 2 days, 4 mins
 .KMMMMMMMMMMMMMMMMMMMMMMMWd.    Packages: 110 (brew)
 XMMMMMMMMMMMMMMMMMMMMMMMX.      Shell: zsh 5.9
;MMMMMMMMMMMMMMMMMMMMMMMM:       Resolution: 1800x1169
:MMMMMMMMMMMMMMMMMMMMMMMM:       DE: Aqua
.MMMMMMMMMMMMMMMMMMMMMMMMX.      WM: Quartz Compositor
 kMMMMMMMMMMMMMMMMMMMMMMMMWd.    WM Theme: Blue (Dark)
 .XMMMMMMMMMMMMMMMMMMMMMMMMMMk   Terminal: /dev/ttys020
  .XMMMMMMMMMMMMMMMMMMMMMMMMK.   CPU: Apple M1 Pro
    kMMMMMMMMMMMMMMMMMMMMMMd     GPU: Apple M1 Pro
     ;KMMMMMMMWXXWMMMMMMMk.      Memory: 3160MiB / 16384MiB
       .cooc,.    .,coo:.

```

### Additional context

_No response_

None

- [ ] I want to work on this

tp:
Changes the database migration code generation to handle the case when table `X` has a foreign key into `Y` and `Y`'s table name changes.

As the migration handles renames by dropping and recreating the table (in this case `Y`), we also need to drop and recreate `X` in this case, as IDs stored in `X` would be not point to any data in `Y` (and thus the previous migration code would fail to enable the foreign key constraint in case where the `X` contained any IDs pointing to `Y`).

Closes #3130


tools/serverpod_cli/test/database/extentions/migration_with_many_to_many_relation_pgsql_code_generation_test.dart
```diff
@@ -0,0 +1,121 @@
...
+
+import '../../generator/dart/client_code_generator/class_constructors_test.dart';
+
+void main() {
+  test(
+      'Given a table that is not managed by serverpod that changes to be managed',
+      () {
+    DatabaseDefinition sourceDefinition;
+    {
+      var models = [
```
tp, line 17: Started from the high-level model sources here, because the set-up is rather complex and I would not trust (myself) getting the low-level classes setup exactly in the way they would be from this definition.

So a more higher level (integration-ish) test makes sense for me, as that's also what the user sees.

tools/serverpod_cli/lib/src/database/migration.dart
```diff
@@ -85,6 +86,30 @@ DatabaseMigration generateDatabaseMigration({
     }
   }
 
+  var tablesToBeDeleted = actions
```
tp, line 89: This can be moved up, as we collect the tables to be deleted early. So we might just be able to pass that into the "alter table" computation, and then avoid the "double drop".

tools/serverpod_cli/test/database/extentions/migration_with_many_to_many_relation_pgsql_code_generation_test.dart
```diff
@@ -0,0 +1,121 @@
...
+import 'package:serverpod_cli/src/analyzer/models/stateful_analyzer.dart';
+import 'package:serverpod_cli/src/database/create_definition.dart';
+import 'package:serverpod_cli/src/generator/code_generation_collector.dart';
+import 'package:serverpod_cli/src/test_util/builders/model_source_builder.dart';
+import 'package:serverpod_service_client/serverpod_service_client.dart';
+import 'package:test/test.dart';
+
+import '../../generator/dart/client_code_generator/class_constructors_test.dart';
+
+void main() {
```
tp, line 11: Seeing the code causing the problem, I think this should also be reproducible with a simpler 2 table setup. One refers to the other, and when we rename the "target table" it should occur as well (as the "Course" part in this example seems to be not relevant).

tools/serverpod_cli/lib/src/database/migration.dart
```diff
@@ -11,14 +12,26 @@ DatabaseMigration generateDatabaseMigration({
...
+
+    if (!databaseTarget.containsTableNamed(srcTable.name) ||
+        srcTable.foreignKeys.any(
+            (foreignKey) => deleteTables.contains(foreignKey.referenceTable))) {
+      if (!deleteTables.contains(srcTable.name)) {
+        deleteTables.add(srcTable.name);
+
+        // Start the loop over, as an earlier table might point to the newly deleted
+        // table, and thus also needs to get dropped now.
+        i = -1; // loop will add +1 again, making this 0
```
tp, line 30: This is not as nice as I would've like, but I think just starting the loop over in the case a new table is marked for deletion and then finding all dependent tables is the simplest and thus most solid approach (especially as the relationship could go through multiple table/foreign key relations, so we can not just easily look for "all the other tables pointing here").

tools/serverpod_cli/lib/src/database/migration.dart
```diff
@@ -43,12 +56,14 @@ DatabaseMigration generateDatabaseMigration({
         (table) => table?.name == dstTable.name,
```
tp, line 56: Or should I better extend this condition to check whether the source table is marked for deletion? Because then it does not really "exist" anymore.

tools/serverpod_cli/lib/src/database/migration.dart
```diff
@@ -11,14 +12,26 @@ DatabaseMigration generateDatabaseMigration({
...
+        deleteTables.add(srcTable.name);
+
+        // Start the loop over, as an earlier table might point to the newly deleted
+        // table, and thus also needs to get dropped now.
+        i = -1; // loop will add +1 again, making this 0
+      }
     }
   }
+
   for (var tableName in deleteTables.reversed) {
```
tp, line 35: Would've loved to understand why this is being forked on in reverse (as the `sourceTables` and thus `deleteTables` does not have a special order as far as I can see).

tools/serverpod_cli/lib/src/database/migration.dart
christerswahn: Suggestion: Make a general helper function that given an iterable of target tables finds all dependent tables in an iterable of source tables. Should be straight-forward to do recursively. That would maybe also be resusable and not just for deletion.

tools/serverpod_cli/lib/src/database/migration.dart
christerswahn: This is the condition added below right? Sounds reasonable

christerswahn (commented):
Great to get this issue fixed! Made a couple of comments. Would be good also if @Isakdl has a look!

tools/serverpod_cli/lib/src/database/migration.dart
tp: @christerswahn Just pushed a change with your suggestion.

Hopefully it's a little clearer due to the function name. Overall that's still tricky, as we have to be careful to check that we don't add any table twice / get stuck in a recursive loop. That's why the `Set<String>` is passed around, and we can not just merge the return values for all tables together in this case.

But it works just as before, and I also added a special case for a recursive dependency (or a chain would also do) to make sure we don't have any issues there.

tools/serverpod_cli/lib/src/database/migration.dart
```diff
@@ -107,6 +100,27 @@ DatabaseMigration generateDatabaseMigration({
   );
 }
 
+/// Returns the set of table names for all tables which have any relation into the table mentioned by [tableName]
+Set<String> _findDependentTables(
+  String tableName,
+  List<TableDefinition> tables, {
+  Set<String>? relatedTables,
+}) {
+  relatedTables ??= {};
```
christerswahn, line 109: Nit: Suggest to rename to dependentTables, that's what they are right?

tools/serverpod_cli/lib/src/database/migration.dart
christerswahn: Good stuff! And good catch regarding guarding against circular dependency loops.

christerswahn (approved):
Looks good from my end. Perhaps @Isakdl should approve as well as he's more familiar with this code base.

SandPod (changes requested):
As discussed in slack - Before merging this I would like to understand how this change would interact with migrations when the column referencing the dropped table is removed  as well.

tools/serverpod_cli/test/database/extentions/migration_with_many_to_many_relation_pgsql_code_generation_test.dart
```diff
@@ -0,0 +1,226 @@
...
```
tp, line 11: as per @SandPod's suggestion: Having tables `A(id, name)` and `B(id, a -> A)`, then do *not* drop `A` when both `A` and `B.a` are dropped by the migration already (in which case there is nothing pointing into `A` anymore anyway).

Thus we need to take the "target tables" (new schema) into account for this check as well.

tp: Added a test and support for this cases.

tools/serverpod_cli/lib/src/database/migration.dart
```diff
@@ -92,6 +102,38 @@ DatabaseMigration generateDatabaseMigration({
...
+    if (!dependentTables.contains(table.name) &&
+        table.foreignKeys.any((foreignKey) =>
+            foreignKey.referenceTable == tableName &&
+            // Check whether the reference will also be upheld in the target table.
+            // otherwise the target table will already be modified and does not need to have be fully dropped
+            targetTables.any((targetTable) =>
+                targetTable.name == table.name &&
+                targetTable.foreignKeys.any((targetForeignKey) =>
+                    targetForeignKey.constraintName ==
+                    foreignKey.constraintName)))) {
```
tp, line 124: @SandPod Is checking the constraint here sufficient, or would it better to compare the columns (or both)?

For the columns I suppose we'd have to import `ListEquality` from package `collection`. (Which is already an indirect dependency anyway.)

tools/serverpod_cli/lib/src/database/migration.dart
tp: Renamed as suggested, @christerswahn.

tools/serverpod_cli/test/database/extentions/migration_with_many_to_many_relation_pgsql_code_generation_test.dart
```diff
@@ -0,0 +1,327 @@
...
+  name: String
+          ''',
+        ).build(),
+        ModelSourceBuilder().withFileName('source').withYaml(
+          '''
+class: Source
+table: source
+fields:
+  name: String
+  target: Target?, relation
```
SandPod, line 247: Suggestion/Discussion: Add a test for when the relation is optional.

In that case, it becomes a bit more tricky to decide if we should drop the table or not. The user might have "nulled" out all the rows ensuring that the rename is fine.

tools/serverpod_cli/lib/src/database/migration.dart
SandPod: I think this should be sufficient. If I remember this correcly we aggregate all the column knowledge around foreign keys in the table either way. So if there is no information left in the table there shouldn't exist a column with a foreign key reference.

SandPod (commented):
I think the approach looks good, there is another case that I would like to see covered in the tests and that we should maybe discuss.

tools/serverpod_cli/test/database/extentions/migration_with_many_to_many_relation_pgsql_code_generation_test.dart
tp: Here we discussed that while indeed all values might be `null` in the database (in which case we would not need to drop), it's probably not worth the complexity to handle this presumably rare case until it's requested.

(The workaround for the consumer would also be easy enough, they would just have to drop (and later recreate) the column if they want to, in which case we leave the table in place.)

tools/serverpod_cli/lib/src/database/migration.dart
```diff
@@ -92,6 +102,38 @@ DatabaseMigration generateDatabaseMigration({
...
+Set<String> _findDependentTables(
+  String tableName, {
+  required List<TableDefinition> sourceTables,
+  required List<TableDefinition> targetTables,
+  Set<String>? dependentTables,
+}) {
+  dependentTables ??= {};
+
+  for (var table in sourceTables) {
+    if (!dependentTables.contains(table.name) &&
```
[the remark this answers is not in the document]
tp: Yes. I rewrote this to break up the a) the `!contains` (to not handle a table to be deleted again), and then the actual "should this table be deleted" into an inline function to hopefully make this clearer.

tools/serverpod_cli/test/database/extentions/migration_with_many_to_many_relation_pgsql_code_generation_test.dart
```diff
@@ -0,0 +1,327 @@
...
+          ''',
+        ).build(),
+        ModelSourceBuilder().withFileName('c').withYaml(
+          '''
+class: C
+table: c
+fields:
+  name: String?
+          ''',
+        ).build(),
```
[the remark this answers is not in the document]
tp: Split this out into a separate test.

tools/serverpod_cli/test/database/extentions/migration_with_many_to_many_relation_pgsql_code_generation_test.dart
SandPod: I couldn't find a test that covers when the relation is optional.

Even if we ignore it, I still think we should have a test that documents the expected behavior.

SandPod (commented):
Single comment around the missing test, otherwise the changes in the PR looks good! 👍

tools/serverpod_cli/test/database/extentions/migration_with_many_to_many_relation_pgsql_code_generation_test.dart
tp: Added the extra test case :)

Issue #3148, opened 2025-01-20
Exception: Not a server package (server) when creating migration


Isakdl:
The CLI is crashing when running the `create-migration`command if the server folder was renamed.
I think it would make sense not to rely on the folder name as a marker for a Serverpod project to be less opinionated. But if we do we shouldn't crash.

### Discussed in https://github.com/serverpod/serverpod/discussions/3141

<div type='discussions-op-text'>

<sup>Originally posted by **Hanibachi** January 19, 2025</sup>
when running the server i get 
```
SERVERPOD version: 2.3.1, dart: 3.5.4 (stable) (Wed Oct 16 16:18:51 2024 +0000) on "linux_x64", time: 2025-01-18 23:28:58.843345Z
Failed to parse command line arguments. Using default values. FormatException: Could not find an option named "--apply-repair-migrations".
mode: development, role: monolith, logging: normal, serverId: default
SERVERPOD initialized, time: 2025-01-18 23:28:58.951629Z
WARNING: The database does not match the target database:
 - Table "users" is missing.
Hint: Did you forget to apply the migrations (--apply-migrations) or run a repair migration (--apply-repair-migration)?
Insights listening on port 8081
Server default listening on port 8080
Webserver listening on port 8082
```
but when i try to run `serverpod create-migration` i get this
```
Yikes! It is possible that this error is caused by an internal issue with the Serverpod tooling. We would appreciate if you filed an issue over at
Github. Please include the stack trace below and describe any steps you did to trigger the error.
https://github.com/serverpod/serverpod/issues

Exception: Not a server package (server)

#0      moduleNameFromServerPackageName (package:serverpod_cli/src/util/locate_modules.dart:169:5)
#1      MigrationGenerator._extractModuleNameFromPath (package:serverpod_cli/src/migrations/generator.dart:392:12)
#2      MigrationGenerator._loadMigrationVersionsFromModules.<anonymous closure> (package:serverpod_cli/src/migrations/generator.dart:336:26)
#3      WhereIterator.moveNext (dart:_internal/iterable.dart:451:13)
#4      MigrationGenerator._loadMigrationVersionsFromModules (package:serverpod_cli/src/migrations/generator.dart:343:28)
<asynchronous suspension>
#5      MigrationGenerator._loadModuleDatabaseDefinitions (package:serverpod_cli/src/migrations/generator.dart:150:20)
<asynchronous suspension>
[...]

PR #3206, opened 2025-02-06
Fix: for issue 3148

henycave:
This PR fixes issue #3148.

Previously, Serverpod determined the server package by checking if the package name had the _server suffix. However, if a user renamed the package, this check would fail and result in a crash.

With this update, the system will:

- Still check for the _server suffix.

- Additionally, inspect the config/generator.yaml file.

- Look for a field that explicitly assigns type: server.

- If this field is present, it will recognize the package as a server package, even if the name has been changed.

- If the package name is changed and type: server is missing from generator.yaml, it will throw an exception.

### DIFF
--- a/tools/serverpod_cli/lib/src/migrations/generator.dart
+++ b/tools/serverpod_cli/lib/src/migrations/generator.dart
@@ -331,17 +331,18 @@ class MigrationGenerator {
       (module) => module.migrationVersions.isNotEmpty,
     );
 
-    var selectedPaths = modulePaths.where(
-      (modulePath) {
-        var moduleName = _extractModuleNameFromPath(modulePath);
-        return selectedModules.any((module) => module.name == moduleName);
-      },
-    );
+    var selectedPaths = (await Future.wait(
+      modulePaths.map((modulePath) async {
+        var moduleName = await _extractModuleNameFromPath(modulePath);
+        return selectedModules.any((module) => module.name == moduleName)? modulePath
+            : null;
+      }),
+    )).whereType<Uri>().toList();
 
     var moduleMigrationVersions = <MigrationVersion>[];
 
     for (var modulePath in selectedPaths) {
-      var moduleName = _extractModuleNameFromPath(modulePath);
+      var moduleName = await _extractModuleNameFromPath(modulePath);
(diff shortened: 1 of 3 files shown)


henycave:

henycave:
recheck

henycave:
I have applied the necessary Dart formatting and pushed the updated commit. Please let me know if there are any other adjustments or if you require further changes. Thank you.

Isakdl (changes requested):
Hello @henycave 

Thank you for the PR :) 

But this PR is not addressing the issue linked. The bug is still present after these code changes, and while I agree that we should address the naming requirement it is not necessary to resolve the bug reported in the issue. The main point is to not crash the CLI when creating a migration.

If you create a new project, rename the server folder by removing the `_server` and then create a migration you will still see the crash even after your changes.

Screenshot 2025-02-07 at 14 31 02

To implement this fix for real you first have to setup a test scenario that demonstrates the bug, then when you have a failing test you can fix the bug and you should end up with a test that passes!

The problem to solve is the crash during the `create-migration` command, we should instead print a "friendly" message, instead of crashing. But the logical flow would still be the same.

---

The code you have changed here is related to how we read modules, not how we parse the users own project. With that said I think we shouldn't deal with the naming issue if your intention is to solve the crash it self as the naming problem is a little more involved than resolving this crash.

For example it will not be enough to look for the generator config file as we also have mini projects where this config file is not present. If we solve the naming problem we should find a solution that doesn't care about the name of the folder at all, and it should work in all scenarios, e.g. every type of Serverpod project. Hence this is a much bigger undertaking and needs much more test scenarios to validate the implementation.

henycave:
Hi @Isakdl, 
Thank you for the feedback! I identified the issue, it was a mix-up between absolute and relative paths. When the file path was absolute, I needed to add a `/` at the beginning. The changes should now work as expected. 

However, as you mentioned, if a server package doesn’t have a generator file, it could still cause a failure. To properly handle this, I’ll submit another PR that ensures the CLI displays a friendly error message instead of crashing.

henycave:
I have opened another PR that addresses this issue. #3210

PR #3210, opened 2025-02-07
fix: Gracefully handle unexpected serverpod module and package names.

henycave:
This PR fixes issue  #3148.

This PR introduces a change to gracefully handle errors by logging a friendly error message instead of allowing the CLI to crash. Additionally, a test has been added to ensure this error handling works as expected.

### DIFF
--- a/packages/serverpod_shared/lib/src/migration_exceptions.dart
+++ b/packages/serverpod_shared/lib/src/migration_exceptions.dart
@@ -68,14 +68,3 @@ class MigrationVersionAlreadyExistsException implements Exception {
     required this.directoryPath,
   });
 }
-
-/// Exception thrown when a module name cannot be determined from the server package name.
-class LocateModuleNameFromServerPackageNameException implements Exception {
-  /// The package name that doesn't have a suffix of '_server'.
-  final String packageName;
-
-  /// Creates a new [LocateModuleNameFromServerPackageNameException].
-  LocateModuleNameFromServerPackageNameException({
-    required this.packageName,
-  });
-}
--- a/packages/serverpod_shared/lib/src/migration_exceptions.dart
+++ b/packages/serverpod_shared/lib/src/migration_exceptions.dart
@@ -68,3 +68,14 @@ class MigrationVersionAlreadyExistsException implements Exception {
     required this.directoryPath,
   });
 }
+
+/// Exception thrown when a module name cannot be determined from the server package name.
+class LocateModuleNameFromServerPackageNameException implements Exception {
+  /// The package name that doesn't have a suffix of '_server'.
+  final String packageName;
+
+  /// Creates a new [LocateModuleNameFromServerPackageNameException].
+  LocateModuleNameFromServerPackageNameException({
+    required this.packageName,
+  });
+}
--- a/tools/serverpod_cli/lib/src/commands/create_migration.dart
+++ b/tools/serverpod_cli/lib/src/commands/create_migration.dart
@@ -52,11 +52,7 @@ class CreateMigrationCommand extends ServerpodCommand {
     GeneratorConfig config;
     try {
       config = await GeneratorConfig.load();
-    } catch (e) {
-      if (e is LocateModuleNameFromServerPackageNameException) {
-        log.error('Not a server package (${e.packageName}). Please '
-            'make sure your server package name ends with \'_server\'.');
-      }
+    } catch (_) {
       throw ExitException(ExitCodeType.commandInvokedCannotExecute);
     }
 
--- a/tools/serverpod_cli/lib/src/commands/create_migration.dart
+++ b/tools/serverpod_cli/lib/src/commands/create_migration.dart
@@ -96,9 +96,6 @@ class CreateMigrationCommand extends ServerpodCommand {
           'Unable to create migration. A directory with the same name already '
           'exists: "${e.directoryPath}".',
         );
-      } on LocateModuleNameFromServerPackageNameException catch (e) {
-        log.error('Not a server package (${e.packageName}). Please '
-            'make sure your server package name ends with \'_server\'.');
       }
 
       return migration != null;
--- a/tools/serverpod_cli/lib/src/commands/create_migration.dart
+++ b/tools/serverpod_cli/lib/src/commands/create_migration.dart
@@ -49,11 +49,14 @@ class CreateMigrationCommand extends ServerpodCommand {
         throw ExitException(ExitCodeType.commandInvokedCannotExecute);
       }
     }
-
     GeneratorConfig config;
     try {
       config = await GeneratorConfig.load();
-    } catch (_) {
+    } catch (e) {
+      if (e is LocateModuleNameFromServerPackageNameException) {
+        log.error('Not a server package (${e.packageName}). Please '
+            'make sure your server package name ends with \'_server\'.');
+      }
       throw ExitException(ExitCodeType.commandInvokedCannotExecute);
     }
 
--- a/tools/serverpod_cli/lib/src/commands/create_migration.dart
+++ b/tools/serverpod_cli/lib/src/commands/create_migration.dart
@@ -97,6 +97,9 @@ class CreateMigrationCommand extends ServerpodCommand {
           'Unable to create migration. A directory with the same name already '
           'exists: "${e.directoryPath}".',
         );
+      } on LocateModuleNameFromServerPackageNameException catch (e) {
+        log.error('Not a server package (${e.packageName}). Please '
+            'make sure your server package name ends with \'_server\'.');
       }
 
       return migration != null;
--- a/tools/serverpod_cli/lib/src/migrations/generator.dart
+++ b/tools/serverpod_cli/lib/src/migrations/generator.dart
@@ -21,6 +21,7 @@ class MigrationGenerator {
     required this.directory,
     required this.projectName,
   });
+
   final Directory directory;
   final String projectName;
 
@@ -324,7 +325,9 @@ class MigrationGenerator {
     required Directory directory,
   }) async {
     var modulePaths = await locateAllModulePaths(
-        directory: directory, currentProjectName: projectName);
+      directory: directory,
+      currentProjectName: projectName,
+    );
 
     var selectedModules = modules.where(
       (module) => module.migrationVersions.isNotEmpty,
--- a/tools/serverpod_cli/lib/src/migrations/generator.dart
+++ b/tools/serverpod_cli/lib/src/migrations/generator.dart
@@ -324,8 +324,7 @@ class MigrationGenerator {
     required Directory directory,
   }) async {
     var modulePaths = await locateAllModulePaths(
-      directory: directory,
-    );
+        directory: directory, currentProjectName: projectName);
 
     var selectedModules = modules.where(
       (module) => module.migrationVersions.isNotEmpty,
--- a/tools/serverpod_cli/lib/src/migrations/generator.dart
+++ b/tools/serverpod_cli/lib/src/migrations/generator.dart
@@ -9,7 +9,6 @@ import 'package:serverpod_cli/src/config/config.dart';
 import 'package:serverpod_cli/src/config_info/config_info.dart';
 import 'package:serverpod_cli/src/database/create_definition.dart';
 import 'package:serverpod_cli/src/migrations/migration_registry.dart';
-import 'package:serverpod_cli/src/util/locate_modules.dart';
 import 'package:serverpod_cli/src/util/model_helper.dart';
 import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart';
 import 'package:serverpod_serialization/serverpod_serialization.dart';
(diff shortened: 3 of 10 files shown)


tools/serverpod_cli/test/locate_modules_test.dart
```diff
@@ -19,5 +20,13 @@ void main() {
       moduleName = moduleNameFromServerPackageName('serverpod-1.1.1');
       expect(moduleName, 'serverpod');
     });
+
+    test('moduleNameFromServerPackageName - throws exception for invalid input',
```
SandPod, line 24: Comment(non-blocking): This test validates that the unit locate_modules behaves correctly but will not validate that the CLI does not crash when an exception is encountered.

tools/serverpod_cli/lib/src/util/locate_modules.dart
```diff
@@ -166,7 +166,8 @@ String moduleNameFromServerPackageName(String packageDirName) {
     return 'serverpod';
   }
   if (!packageName.endsWith(_serverSuffix)) {
-    throw Exception('Not a server package ($packageName)');
+    throw LocateModuleNameFromServerPackageNameException(
```
SandPod, line 169: Question/Discussion: Upon quick inspection of the code it seems like this method is called from multiple places. Please ensure that we gracefully handle the exception in all the other places as well. Otherwise, I'm concerned that we will just hide the fact that we might crash.

tools/serverpod_cli/lib/src/util/locate_modules.dart
```diff
@@ -166,7 +166,8 @@ String moduleNameFromServerPackageName(String packageDirName) {
...
+        packageName: packageName);
```
SandPod, line 170: Suggestion: add a comment to this method that communicates this method can throw this exception and the reason it would do so.

SandPod (commented):
Apraoch looks valid!

Just ensure that all the other cases that call this method are also covered. I'm concerned that we might run into this issues later on in other places.

henycave:
@SandPod
You are right; the locate_modules function was being called from many places. However, I have added error handling to every place where it was used and included a comment explaining why the exception occurs.

tools/serverpod_cli/lib/src/commands/create_migration.dart
```diff
@@ -49,11 +49,14 @@ class CreateMigrationCommand extends ServerpodCommand {
         throw ExitException(ExitCodeType.commandInvokedCannotExecute);
       }
     }
-
     GeneratorConfig config;
     try {
       config = await GeneratorConfig.load();
-    } catch (_) {
+    } catch (e) {
```
SandPod, line 55: Question(For my knowledge): Could you explain how the `GeneratorConfig.load()` method depends on the `moduleNameFromServerPackageName`?

There is also a bunch of exception handling inside of the `GenereatorConfig.load()`, would it make more sense to handle this exception inside of there like other cases are handled? It seems like we deliberately do not log the exceptions from the config load here, to be consistent, all of them should be handled.

henycave: The GeneratorConfig.load() depends on locateModules() to find modules like serverpod module and the official _server package. And locateModules() loops through the packages and depends on moduleNameFromServerPackageName() to find the name of packages but on further investigation, the locateModules() checks to see if any packages passed to moduleNameFromServerPackageName() are either named ’serverpod’ or ends with a suffix of ‘_server’. So there would not be a scenario where the exception is thrown. Also, inside the locateModules() loop of the packages, if an exception happens, on the catch() it already uses ‘continue’ to go to the next package. So I think we don’t need any exception handling on the GeneratorConfig.load().  

The above scenario is true only for the GeneratorConfig.load(), but for the generator.createMigration() the exception handling is necessary so that the CLI doesn’t crash.  Let me know what you think and I will make the changes.

henycave: The GeneratorConfig.load() depends on locateModules() to find modules like serverpod module and the official _server package. And locateModules() uses package_config to look for these packages in the package_config.json file which it then loops through the packages and depends on moduleNameFromServerPackageName() to find the names of packages. But on further investigation, the locateModules() checks to see if any packages passed to moduleNameFromServerPackageName() are either named ’serverpod’ or end with a suffix of ‘_server’. If we try to handle the error here, we will face a problem where we are showing an error message on every module that isn’t named ’serverpod’ or ends with a suffix of ‘_server’. Which is a lot of error messages that pop up on the CLI because package_config.json file holds lots of packages. So with the current implementation, there would not be a scenario where the exception is thrown. In Addition, Inside the locateModules() loop of the packages, if an exception happens, on the catch(), it already uses ‘continue’ to go to the next package. So I think we don’t need any exception handling on the GeneratorConfig.load().  

The above scenario is true only for the GeneratorConfig.load(), but for the generator.createMigration() the exception handling is necessary so that the CLI doesn’t crash. I also found that if a user edits the name of the package inside the pubspec yaml, it would throw a LocateModuleNameFromServerPackageNameException exception. This is happening inside the getProjectName() which I will add the exception handling.

Let me know what you think and I will make the changes.

henycave: All the specific catches on GeneratorConfig.load() have been removed. Also, the LSP was using  GeneratorConfig.load() which was why the catch was added but has been removed now.  

The catch is necessary for the generator.createMigration() because it needs the database definitions it gets from the migration versions of the modules. So to get those modules it calls the locateAllModulePaths() which gets all the packages from package_config.json after that, it checks whether the package has a name that ends with ‘_server’ or is ’serverpod’. The problem here is that if a user changes the name of the server module folder, it doesn’t affect the name on the package_config.json so even tho the name of the folder has been changed it will pass the checks that the locateAllModulePaths() was making. After that when moduleNameFromServerPackageName() receives the name it throws the exception which crashes the CLI. 

I also added exception handling on the getProjectName. This function gets the name of the project from the pubspec.yaml and gives it to moduleNameFromServerPackageName() which throws an error if the user changes the project name on the pubspec.yaml to a name that doesn’t match the naming convention.

henycave: All the specific catches on GeneratorConfig.load() have been removed. Also, the LSP was using  GeneratorConfig.load() which was why the catch was added but has been removed now.  

The catch is necessary for the generator.createMigration() because it needs the database definitions it gets from the migration versions of the modules. So to get those modules it calls the locateAllModulePaths() which gets all the packages from package_config.json after that, it checks whether the package has a name that ends with ‘_server’ or is ’serverpod’. The problem here is that if a user changes the name of the server module folder, it doesn’t affect the name on the package_config.json so even tho the name of the folder has been changed it will pass the checks that the locateAllModulePaths() was making. After that when moduleNameFromServerPackageName() receives the name it throws the exception which crashes the CLI. 

I also added exception handling on the getProjectName. This function gets the name of the project from the pubspec.yaml and gives it to moduleNameFromServerPackageName() which throws an error if the user changes the project name on the pubspec.yaml to a name that doesn’t match the naming convention.

SandPod: When inspecting THAT code, this again seems like the wrong solution.

Looking at the stack trace of the error in the issue, we see that the exception is triggered by `_loadMigrationVersionsFromModules`. The repro states that the exception is triggered from us renaming our current project, not a module. So we shouldn't really trigger any exception from there when renaming our project.

It feels like something funky is going on when our current project ends up in the list of `...FromModules`.

There is possibly some other filtering that should happen there so that we never add a module that does not fit the naming convention of a Serverpod project.

Could you inspect that part of the code closely to understand why this is happening?

I'm concerned that the current solution would simply silence a different error.

henycave: Could you explain a bit more? so we shouldn't even be getting the current project listed inside the `...FromModules`? In that case, we can filter it by checking if the module name is the same as the current project name and excluding it. 
For `getProjectName()`, an exception will only occur if the user manually changes the project name inside pubspec.yaml. In that case, we will leave the exception handling as is.

henycave: We are using `package_config` to find all the modules. Since the current project gets listed in package_config.json along with the modules, it was incorrectly included in the list of `...FromModules`. To fix this, I’ve added a filter that excludes the current project from the list of modules. This ensures that the exception will not happen.

SandPod: > Could you explain a bit more? so we shouldn't even be getting the current project listed inside the...
Yes, that is correct. Looking at the call here: https://github.com/serverpod/serverpod/blob/eadf36d3b52051a4ff01157182703273e129280e/tools/serverpod_cli/lib/src/migrations/generator.dart#L77-L86

We handle the project separate from the modules, then we explicitly merge the definitions here: https://github.com/serverpod/serverpod/blob/eadf36d3b52051a4ff01157182703273e129280e/tools/serverpod_cli/lib/src/migrations/generator.dart#L94-L98

So I think it is unexpected behavior if the project is included when loading the modules. Adding the filter is probably a better approach.

packages/serverpod_shared/lib/src/migration_exceptions.dart
```diff
@@ -68,3 +68,14 @@ class MigrationVersionAlreadyExistsException implements Exception {
     required this.directoryPath,
   });
 }
+
+/// Exception thrown when a module name cannot be determined from the server package name.
+class LocateModuleNameFromServerPackageNameException implements Exception {
```
SandPod, line 73: Suggestion: Move this exception from this file.

This file contains exceptions that are thrown during migration. Since this is not an exception that happens during a migration we should place it in a context of where it can happen.

SandPod: I think it could be defined directly in `locate_modules.dart`

tools/serverpod_cli/lib/src/migrations/generator.dart
```diff
@@ -324,8 +324,7 @@ class MigrationGenerator {
     required Directory directory,
   }) async {
     var modulePaths = await locateAllModulePaths(
-      directory: directory,
-    );
+        directory: directory, currentProjectName: projectName);
```
SandPod, line 327: Suggestion(NIT): Add comma at the end for better formatting of method call.

Apply this to all your changes that ended up with this type of line break.

tools/serverpod_cli/lib/src/commands/create_migration.dart
SandPod: But if we inspect `config.modulesDependent` that should only contain modules not including our self right? So there is already a filtering in place, the renaming probably makes us not compatible with our self anymore.

Therefore, I would suggest that we print a helpful message when the error occurs, but we ignore the failing module for now and continue with the operation of the method.

SandPod (commented):
Couple of minor comments.

tools/serverpod_cli/lib/src/commands/create_migration.dart
henycave: So, should I continue making the minor changes?

SandPod: OK, thank you for the explanation. The interaction between `locateAllModulePaths` and `locateModules` seems a bit strange. I'll need to take a look on that.

tools/serverpod_cli/lib/src/util/locate_modules.dart
```diff
@@ -110,9 +110,8 @@ List<String> findAllMigrationVersionsSync({
   }
 }
 
-Future<List<Uri>> locateAllModulePaths({
-  required Directory directory,
-}) async {
+Future<List<Uri>> locateAllModulePaths(
```
SandPod, line 113: Suggestion: Slim down this method significantly and only operate on a single `List<ModuleConfig>` parameter.

This method is basically a duplication of `locateModules` with the exception that the output is different.

It is only used in one place in the generator so we can make it single purpose.
If I understand the code correctly here, what we produce is very similair to the `ModuleConfig.serverPackageDirectoryPathParts` field, and since that is already available in the generator context, we should be able to simply read that field.

The field `ModuleConfig.name` should also be the same as the package name we are calculating in the generator. So we should be able to remove that as well and simply use the field in the config.

This would remove the two reasons for this to crash in `generator.dart` and also remove a lot of duplicated code.

SandPod (changes requested):
Please slim down or completely remove `locateAllModulePaths` and use the already calculated information available in the `ModuleConfig` object.

tools/serverpod_cli/lib/src/util/locate_modules.dart
henycave: I am able to completely remove the `locateAllModulePaths()` function and use the modules from the config which are passed as an argument to the `createMigration()`. The ModuleConfig.serverPackageDirectoryPathParts has a list of path segments, but it doesn't include the scheme so in the code we have to assume that it's a local File.

```
var uri = Uri(
        scheme: 'file', // assuming the module is local
        pathSegments: moduleConfig.serverPackageDirectoryPathParts,
      );
```

The other option is to edit the `ModuleConfig` constructor and include another field that holds the file URI, which would also include the scheme. 

Let me know which option is better, and I will make the change.

SandPod: Great! I think it is fine to assume this is a `file` scheme; we don't really need that additional information in the `ModuleConfig`.

SandPod (commented):
PR looks good to me now, just ensure that all the CI runs pass and we can merge it! :)

SandPod (approved):
LGTM! 👍

PR #3218, opened 2025-02-10
feat: Support `Record` types for endpoint return value and parameters.

tp:
Relates to #1784

### DIFF
--- a/packages/serverpod_serialization/lib/src/serialization.dart
+++ b/packages/serverpod_serialization/lib/src/serialization.dart
@@ -241,6 +241,12 @@ abstract class SerializationManager {
         } else if (encodeForProtocol && nonEncodable is ProtocolSerialization) {
           return nonEncodable.toJsonForProtocol();
         } else {
+          if (object is Record) {
+            throw Exception(
+              'Records are not supported in `encode`. They must be converted beforehand via `Protocol.mapRecordToJson` or the enclosing `SerializableModel`.',
+            );
+          }
+
           // ignore: avoid_dynamic_calls
           return nonEncodable?.toJson();
           // throws NoSuchMethodError if toJson is not implemented
@@ -274,7 +280,7 @@ abstract class SerializationManager {
   }
 
   /// Encode the provided [object] to a json-formatted [String], include class
-  /// name so that it can be decoded even if th class is unknown.
+  /// name so that it can be decoded even if the class is unknown.
   /// If [formatted] is true, the output will be formatted with two spaces
   /// indentation.
   String encodeWithType(
--- a/tests/serverpod_test_server/lib/src/endpoints/record_parameters.dart
+++ b/tests/serverpod_test_server/lib/src/endpoints/record_parameters.dart
@@ -1,0 +1,235 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+
+typedef _MixedRecordTypeDef = (int number, {SimpleData data});
+
+class RecordParametersEndpoint extends Endpoint {
+// #region Records with single positional parameter (required and nullable)
+  Future<(int,)> returnRecordOfInt(Session session, (int,) record) async {
+    return record;
+  }
+
+  Future<(int,)?> returnNullableRecordOfInt(
+    Session session,
+    (int,)? record,
+  ) async {
+    return record;
+  }
+
+  Future<(int?,)> returnRecordOfNullableInt(
+    Session session,
+    (int?,) record,
+  ) async {
+    return record;
+  }
+
+  Future<(int?,)?> returnNullableRecordOfNullableInt(
+    Session session,
+    (int?,)? record,
+  ) async {
+    return record;
+  }
+// #endregion
+
+// #region Records with multiple positional parameters
+  Future<(int, String)> returnIntStringRecord(
+    Session session,
+    (int, String) record,
+  ) async {
+    return record;
+  }
+
+  Future<(int, String)?> returnNullableIntStringRecord(
+    Session session,
+    (int, String)? record,
+  ) async {
+    return record;
+  }
+
+  Future<(int, SimpleData)> returnIntSimpleDataRecord(
+    Session session,
+    (int, SimpleData) record,
+  ) async {
+    return record;
+  }
+
+  Future<(int, SimpleData)?> returnNullableIntSimpleDataRecord(
+    Session session,
+    (int, SimpleData)? record,
+  ) async {
+    return record;
+  }
+// #endregion
+
+// #region Records with named parameters
+  Future<({int number, String text})> returnNamedIntStringRecord(
+    Session session,
+    ({int number, String text}) record,
+  ) async {
+    return record;
+  }
+
+  Future<({int number, String text})?> returnNamedNullableIntStringRecord(
+    Session session,
+    ({int number, String text})? record,
+  ) async {
+    return record;
+  }
+
+  Future<({int number, SimpleData data})> returnRecordOfNamedIntAndObject(
+    Session session,
+    ({int number, SimpleData data}) record,
+  ) async {
+    return record;
+  }
+
+  Future<({int number, SimpleData data})?>
+      returnNullableRecordOfNamedIntAndObject(
+    Session session,
+    ({int number, SimpleData data})? record,
+  ) async {
+    return record;
+  }
+
+  Future<({int? number, SimpleData? data})>
+      returnRecordOfNamedNullableIntAndNullableObject(
+    Session session,
+    ({int? number, SimpleData? data}) record,
+  ) async {
+    return record;
+  }
+// #endregion
+
+// #region Records using a `typedef`, pointing to positional and named parameters
+  Future<_MixedRecordTypeDef> returnRecordTypedef(
+    Session session,
+    _MixedRecordTypeDef record,
+  ) async {
+    return record;
+  }
+
+  Future<_MixedRecordTypeDef?> returnNullableRecordTypedef(
+    Session session,
+    _MixedRecordTypeDef? record,
+  ) async {
+    return record;
+  }
+// #endregion
+
+// #region Records inside `List`s
+  Future<List<(int, SimpleData)>> returnListOfIntSimpleDataRecord(
+    Session session,
+    List<(int, SimpleData)> recordList,
+  ) async {
+    return recordList;
+  }
+
+  Future<List<(int, SimpleData)?>> returnListOfNullableIntSimpleDataRecord(
+    Session session,
+    List<(int, SimpleData)?> record,
+  ) async {
+    return record;
+  }
+// #endregion
+
+// #region Records inside `Set`s
+  Future<Set<(int, SimpleData)>> returnSetOfIntSimpleDataRecord(
+    Session session,
+    Set<(int, SimpleData)> recordSet,
+  ) async {
+    return recordSet;
+  }
+
+  Future<Set<(int, SimpleData)?>> returnSetOfNullableIntSimpleDataRecord(
+    Session session,
+    Set<(int, SimpleData)?> set,
+  ) async {
+    return set;
+  }
+
+  Future<Set<(int, SimpleData)>?> returnNullableSetOfIntSimpleDataRecord(
+    Session session,
+    Set<(int, SimpleData)>? recordSet,
+  ) async {
+    return recordSet;
+  }
+// #endregion
+
+// #region Records inside `Map`s
+  Future<Map<String, (int, SimpleData)>> returnStringMapOfIntSimpleDataRecord(
+    Session session,
+    Map<String, (int, SimpleData)> map,
+  ) async {
+    return map;
+  }
+
+  Future<Map<String, (int, SimpleData)?>>
+      returnStringMapOfNullableIntSimpleDataRecord(
+    Session session,
+    Map<String, (int, SimpleData)?> map,
+  ) async {
+    return map;
+  }
+
+  Future<Map<(String, int), (int, SimpleData)>>
+      returnRecordMapOfIntSimpleDataRecord(
+    Session session,
+    Map<(String, int), (int, SimpleData)> map,
+  ) async {
+    return map;
+  }
+// #endregion
+
+// #region Complex nested container structure
+  /// Returns the first and only input value mapped into the return structure (basically reversed)
+  Future<Map<String, List<Set<(int,)>>>> returnStringMapOfListOfRecord(
+    Session session,
+    // This type is only used in a parameter postion, ensuring that we generate those as well
+    Set<List<Map<String, (int,)>>> input,
+  ) async {
+    var key = input.single.single.keys.single;
+    var value = input.single.single.values.single.$1;
+
+    return {
+      key: [
+        {
+          (value,),
+        }
+      ],
+    };
+  }
+// #endregion
+
+// #region Records with nested records
+  Future<({(SimpleData, double) namedSubRecord})> returnNestedNamedRecord(
+    Session session,
+    ({(SimpleData, double) namedSubRecord}) record,
+  ) async {
+    return record;
+  }
+
+  Future<({(SimpleData, double)? namedSubRecord})>
+      returnNestedNullableNamedRecord(
+    Session session,
+    ({(SimpleData, double)? namedSubRecord}) record,
+  ) async {
+    return record;
+  }
+
+  Future<((int, String), {(SimpleData, double) namedSubRecord})>
+      returnNestedPositionalAndNamedRecord(
+    Session session,
+    ((int, String), {(SimpleData, double) namedSubRecord}) record,
+  ) async {
+    return record;
+  }
+
+  Future<List<((int, String), {(SimpleData, double) namedSubRecord})>>
+      returnListOfNestedPositionalAndNamedRecord(
+    Session session,
+    List<((int, String), {(SimpleData, double) namedSubRecord})> recordList,
+  ) async {
+    return recordList;
+  }
+// #endregion
+}
--- a/tests/serverpod_test_server/lib/src/endpoints/test_tools.dart
+++ b/tests/serverpod_test_server/lib/src/endpoints/test_tools.dart
@@ -183,6 +183,20 @@ class TestToolsEndpoint extends Endpoint {
     return simpleDatas;
   }
 
+  Future<(String, (int, bool))> echoRecord(
+    Session session,
+    (String, (int, bool)) record,
+  ) async {
+    return record;
+  }
+
+  Future<List<(String, (int, bool))>> echoRecords(
+    Session session,
+    List<(String, (int, bool))> records,
+  ) async {
+    return records;
+  }
+
   Future<void> logMessageWithSession(Session session) async {
     session.log('test session log in endpoint');
   }
--- a/tests/serverpod_test_server/lib/src/generated/protocol.dart
+++ b/tests/serverpod_test_server/lib/src/generated/protocol.dart
@@ -6752,18 +6752,8 @@ class Protocol extends _i1.SerializationManagerServer {
     if (t == List<int>) {
       return (data as List).map((e) => deserialize<int>(e)).toList() as T;
     }
-    if (t == _i1.getType<List<String>?>()) {
-      return (data != null
-          ? (data as List).map((e) => deserialize<String>(e)).toList()
-          : null) as T;
-    }
     if (t == List<List<int>>) {
       return (data as List).map((e) => deserialize<List<int>>(e)).toList() as T;
-    }
-    if (t == _i1.getType<List<int>?>()) {
-      return (data != null
-          ? (data as List).map((e) => deserialize<int>(e)).toList()
-          : null) as T;
     }
     if (t == _i1.getType<List<int>?>()) {
       return (data != null
@@ -6779,16 +6769,6 @@ class Protocol extends _i1.SerializationManagerServer {
           ? (data as List).map((e) => deserialize<int>(e)).toList()
           : null) as T;
     }
-    if (t == _i1.getType<List<int>?>()) {
-      return (data != null
-          ? (data as List).map((e) => deserialize<int>(e)).toList()
-          : null) as T;
-    }
-    if (t == _i1.getType<List<List<int>>?>()) {
-      return (data != null
-          ? (data as List).map((e) => deserialize<List<int>>(e)).toList()
-          : null) as T;
-    }
     if (t == _i1.getType<List<List<int>>?>()) {
       return (data != null
           ? (data as List).map((e) => deserialize<List<int>>(e)).toList()
@@ -6796,11 +6776,6 @@ class Protocol extends _i1.SerializationManagerServer {
     }
     if (t == List<int?>) {
       return (data as List).map((e) => deserialize<int?>(e)).toList() as T;
-    }
-    if (t == _i1.getType<List<int?>?>()) {
-      return (data != null
-          ? (data as List).map((e) => deserialize<int?>(e)).toList()
-          : null) as T;
     }
     if (t == _i1.getType<List<int?>?>()) {
       return (data != null
@@ -6846,18 +6821,6 @@ class Protocol extends _i1.SerializationManagerServer {
           ? (data as List).map((e) => deserialize<_i135.SimpleData>(e)).toList()
           : null) as T;
     }
-    if (t == _i1.getType<List<_i135.SimpleData>?>()) {
-      return (data != null
-          ? (data as List).map((e) => deserialize<_i135.SimpleData>(e)).toList()
-          : null) as T;
-    }
-    if (t == _i1.getType<List<_i135.SimpleData?>?>()) {
-      return (data != null
-          ? (data as List)
-              .map((e) => deserialize<_i135.SimpleData?>(e))
-              .toList()
-          : null) as T;
-    }
     if (t == _i1.getType<List<_i135.SimpleData?>?>()) {
       return (data != null
           ? (data as List)
@@ -6881,12 +6844,6 @@ class Protocol extends _i1.SerializationManagerServer {
               (k, v) => MapEntry(deserialize<String>(k), deserialize<int>(v)))
           : null) as T;
     }
-    if (t == _i1.getType<Map<String, int>?>()) {
-      return (data != null
-          ? (data as Map).map(
-              (k, v) => MapEntry(deserialize<String>(k), deserialize<int>(v)))
-          : null) as T;
-    }
     if (t == Map<String, Map<String, int>>) {
       return (data as Map).map((k, v) => MapEntry(
           deserialize<String>(k), deserialize<Map<String, int>>(v))) as T;
@@ -6895,12 +6852,6 @@ class Protocol extends _i1.SerializationManagerServer {
       return (data as Map).map(
               (k, v) => MapEntry(deserialize<String>(k), deserialize<int?>(v)))
           as T;
-    }
-    if (t == _i1.getType<Map<String, int?>?>()) {
-      return (data != null
-          ? (data as Map).map(
-              (k, v) => MapEntry(deserialize<String>(k), deserialize<int?>(v)))
-          : null) as T;
     }
     if (t == _i1.getType<Map<String, int?>?>()) {
       return (data != null
@@ -6979,18 +6930,6 @@ class Protocol extends _i1.SerializationManagerServer {
               deserialize<String>(k), deserialize<_i135.SimpleData>(v)))
           : null) as T;
     }
-    if (t == _i1.getType<Map<String, _i135.SimpleData>?>()) {
-      return (data != null
-          ? (data as Map).map((k, v) => MapEntry(
-              deserialize<String>(k), deserialize<_i135.SimpleData>(v)))
-          : null) as T;
-    }
-    if (t == _i1.getType<Map<String, _i135.SimpleData?>?>()) {
-      return (data != null
-          ? (data as Map).map((k, v) => MapEntry(
-              deserialize<String>(k), deserialize<_i135.SimpleData?>(v)))
-          : null) as T;
-    }
     if (t == _i1.getType<Map<String, _i135.SimpleData?>?>()) {
       return (data != null
           ? (data as Map).map((k, v) => MapEntry(
@@ -7009,11 +6948,6 @@ class Protocol extends _i1.SerializationManagerServer {
       return (data as List).map((e) => deserialize<_i3.UserInfo>(e)).toList()
           as T;
     }
-    if (t == _i1.getType<List<_i135.SimpleData>?>()) {
-      return (data != null
-          ? (data as List).map((e) => deserialize<_i135.SimpleData>(e)).toList()
-          : null) as T;
-    }
     if (t == Set<int>) {
       return (data as List).map((e) => deserialize<int>(e)).toSet() as T;
     }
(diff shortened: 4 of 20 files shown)


tp:
The simplest case of records with primitives can round-trip now, but only by putting a lot of serialization logic into the encoding before sending to the server/client respectively.

And this also does not help when the record is nested (in a list, object, etc.).

To handle those cases I will have to change to using the project's generated serialization manager instead of the `static` one, as discussed in Slack. But that should not be a huge thing, as we then just need to copy the serialization code from the current endpoint methods into the `switch` over there, and then it should support arbitrarily nested structures.


CleanShot 2025-02-11 at 14 30 46@2x

tp:
Now the protocol encoding code is changed such that we can also supported nested "model objects" and thus more cases pass E2E.

CleanShot 2025-02-12 at 11 27 30@2x


This was achieved by adding support for the actually used `Record` types in the generated `Protocol` class, and then redirecting all calls to `encode` etc. to `protocol(instance).encode` instead of the static base class's `SerializationManager.encode`.

For the "consumer's" generated code this works fine, but I worry that there might be some code paths in the database or caching that before used the single shared `SerializationManger`, and would not have to discern between the various generated ones (like the consumer ones, `serverpod` package's one (insights), `UserInfo` from auth etc. pp.).

So maybe this solution is not tenable for all the other types which are not part of the consumer project. For now it might work fine in all cases as long as we have a single consumer-generated `Protocol` and always use that – but only as long as all the other packages do not implement a `Record`-type that the "consumer project" does not use…

I wonder if there's a way to roll back to the previous `static` `SerializationManager` and have a `registerRecordMapper` hook, that each used package would hook into (e.g. when the `Protocol` instance is created). That way each package could add support for its `Records` types globally, and it wouldn't really matter if the same type is defined in 2 (or more) helpers, as the result would always be the same.

cc @SandPod (FYI, the updated code doing what's described above is not yet pushed, I have to pick that out from the hundreds of generated files). But you can probably already confirm my suspicion that it'd get tricky to pick the "correct" `Protocol` for encode now, given the various types flowing through some shared places like the database (plus, they might even have arbitrary nested, which I just saw in a demo "to JSON widget", where the caller might not even know what package/Protocol the type came from).

tp:
The new to transform on the sending / receiving side to / from "JSON `Maps`" works out.

It's not implemented for all of the cases shown before as well as these newly nested (models inside records):

CleanShot 2025-02-13 at 11 44 17@2x


The only remaining issue right now is that a double-nested record like `(int, (int, Foo?))` can loose the runtime information and thus the instance `(0, (0, null))` does not get deserialized. It does work with a proper instance of `Foo` just fine, but somehow the nested record literal above gets the type `(int, Null)` instead of retaining `(int, Foo?)`…

I'll have to see whether any of the tricks used in deserialize would help here, like passing the expected type as a generic parameter. Just need to find any solution that does not require extra `Null`-using checks (as that could be quite a ton of Model-type/Null-type combinations in a record e.g. like `(Foo?, Bar?, Baz?)`…).

tp:
The latest version now also supports `Records` in "default container types" like `List`, `Map` and `Set`.

This should now cover all possible endpoint method cases (top-level and container-nested records in both parameter and return value position).

I would like to tacke `Stream` support separately in https://github.com/serverpod/serverpod/pull/3219. But the learnings from this PR should already help quite a bit, as we generate all the encode/decode for containers with records, and then we would just need to apply the same for all containers and create the appropriate `className`s for the String-based look-up in streaming methods.

Next up for `Record`s specifically would then be their support in the models, and persistence to the database (which should not be a big problem, as it'll be the JSON serialization we're already using in the protocol E2E).

packages/serverpod_serialization/lib/src/serialization.dart
```diff
@@ -325,3 +331,51 @@ const extensionSerializedTypes = [
 extension<K, V> on Map<K, V> {
   Type get keyType => K;
 }
+
+// extension<T> on Iterable<T> {
+//   Type get valueType => T;
+// }
+
+/// TODO
+dynamic mapRecordContainingContainerToJson(
```
tp, line 340: I added this generic method here that takes any "container" and traverses it to find and map any nested records.

Alternatively, now that we know it works we could also add this inline into the protocol endpoint definition, but for nested types this generates quite convoluted code. It would be type-checked, but not sure if it's an improvement overall, as here we can unit-test the "generic" implementation and be confident that it works.

It's probably more a matter of preference.

We could also look into making this more type-safe/easier to follow (e.g. by having extension methods). But naming will get a bit tricky for the records, and then I'd rather do this in a concerted effort, also looking at possible improvements to e.g. `deserialize` which grows a bit unwieldy (as least when having to work with the generated code in detail), and where you don't see from the outside whether a calling type is even supported.

tp:
FYI: All the `as ClassTypeDefinition` inserted here in the model code generation will disappear in a follow-up when we tackle Record support in the models. Then we'll have to handle each of those cases and discern between class and record types. For now one may only define class types, so that casting is just fine.

packages/serverpod_serialization/lib/src/serialization.dart
```diff
@@ -325,3 +331,53 @@ const extensionSerializedTypes = [
...
+/// It should not be called for [SerializableModel] types. These handle the "[Record] in container" mapping internally already.
+///
+/// It is only supposed to be called from generated protocol code.
+///
+/// Returns either a `List<dynamic>` (for List, Sets, and Maps with non-String keys) or a `Map<String, dynamic>` in case the input was a `Map<String, …>`.
+dynamic mapRecordContainingContainerToJson(
+  Object obj,
+  Map<String, dynamic>? Function(Record?) mapRecord,
+) {
+  assert(obj is List || obj is Map || obj is Set);
```
SandPod, line 346: Suggestion: Change this to a real throw.

Since we are working with serialization and deserialization, the actual values passed in here might dynamically change, meaning that production code might exhibit inputs that are not valid in the debug case. In those cases, we still want to give the same level of validation that the input is correct.

packages/serverpod_serialization/lib/src/serialization.dart
SandPod: Just started the review, so this might be subject to change, but my hunch tells me that this should be inlined in the generated protocol file, that way we don't expose a new interface we promise to support in our serialization package.

All tests for this method should also play well in endpoints so we should be able to create these cases there directly.

tests/serverpod_test_server/lib/src/endpoints/record_parameters.dart
```diff
@@ -0,0 +1,401 @@
...
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+
+typedef _MixedRecordTypeDef = (int number, {SimpleData data});
+
+class RecordParametersEndpoint extends Endpoint {
+  Future<(int,)> returnIntRecord(Session session, (int,) record) async {
+    return record;
+  }
+
+  Future<(String, String)?> returnNullableStringRecord(
```
SandPod, line 11: Suggestion: Move this definition to either be before or after the `int` definitions to preserve the groups.

tests/serverpod_test_server/lib/src/endpoints/record_parameters.dart
```diff
@@ -0,0 +1,401 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+
+typedef _MixedRecordTypeDef = (int number, {SimpleData data});
+
+class RecordParametersEndpoint extends Endpoint {
```
SandPod, line 6: Suggestion: Could you add comments to this file to differentiate the groups?

I think that would help readability and make it easier to asses we have covered all cases.

tests/serverpod_test_server/lib/src/endpoints/record_parameters.dart
```diff
@@ -0,0 +1,401 @@
...
+  }
+
+  Future<(String, String)?> returnNullableStringRecord(
+    Session session,
+    (String, String)? record,
+  ) async {
+    return record;
+  }
+
+  Stream<(int,)> streamIntRecordEcho(
```
SandPod, line 18: Question: I can't seem to find any tests for the streaming methods, is that deliberate?

Please validate that all endpoints defined here are called in a test. If we will add validation for streaming methods later. Please remove the method definitions from this PR and add them in the PR that covers that context.

tools/serverpod_cli/lib/src/generator/types.dart
```diff
@@ -17,8 +18,155 @@ const _moduleRef = 'module:';
 const _projectRef = 'project:';
 const _packageRef = 'package:';
 
+sealed class TypeDefinition {
```
SandPod, line 21: Suggestion: Revert the changes to `TypeDefinition`.

The intention of TypeDefinition was to be able to represent any dart type. Architecturally Records should be one of them.

The class needs some love and refactoring. But I think this should be done in isolation in a separate PR since it touches a lot of code in a lot of places.

As far as I can understand, this specialization was added because we had specific `reference` and `generateDeserailzation` logic. Could you investigate adding that to the already existing structure instead? The logic should be able to be scoped in private method to make clear that it is only connected to Record types.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -766,7 +769,9 @@ class SerializableModelLibraryGenerator {
             );
 
             if (field.relation is ListRelationDefinition) {
-              type = field.type.generics.first.reference(
+              type =
+                  // NOTE(tp): Cast is fine, as relations are to model classes
```
SandPod, line 773: Suggestion: Remove comment.

Please remove any other TODOs or Notes added if there are any. We prefer filing issues over storing TODOs in the code.

SandPod (changes requested):
The inflation of the protocol file from ~4000 lines of code to ~65000 lines of code does not seem to be proportional to the 20 or so endpoints added that used records.

The `deserialize` method seems to have grown significantly there. Are we sure all that code is required?

I would also like to see a rework of the commits in this PR. PRs work as living documentation for future developers around how a change was introduced and why specific decision in the implementation where made. Ideally each commit changes one part of the code (with tests) and documents what is being achieved with the commit message.

This makes the change easy to follow and easier to review. Especially when it comes to these larger changes.

And when having that perspective, fixup commits don't add any value for future reviewers, so small touch up changes should be baked in with the reason for it to change. There is no need for us to preserve that we introduced a bug and fixed it in the same PR.

So please:
- Reshape the commit history so that each commit captures a single change change to the codebase.
- Embedd the fixup commits with the commits where they belong.

tests/serverpod_test_server/lib/src/endpoints/record_parameters.dart
tp: > Question: I can't seem to find any tests for the streaming methods, is that deliberate?

That is correct. Stream support for containers is pending review in https://github.com/serverpod/serverpod/pull/3219.

Once that is merged I would apply a similar adaption for `Record`s (or merge `main` in here, depending on the timing).

If it can not be test in here, I can remove the stream methods from the endpoint file though.

tools/serverpod_cli/lib/src/generator/types.dart
tp: I did not want to further extend `TypeDefinition` to begin with as it covers classes, primitives (String, int, etc.), and enums together, but this combined approach always means you have to beware of the internals of the class to know when calling with field is valid (e.g. something as simple as `retrieveGenericType` would throw for primitives, enum, and now `Record`s by design now (and interestingly also does not handle e.g. multiple generics 🤷🏻‍♂️)).

So I thought a clean split for the `Record`s would be a good start, and then we could later do a similar thing for `enums` (so it's not just a nullable field on the common type), and likely also primitives would benefit from this simplification (them having no generics for example).

Also having the extra class helped me quite a bit to have a "useful package" to pass to other methods. As otherwise we will always have to handle/support the "big type definition" being passed in and then checking that it's actually configured in a way to support our sub-case.

But if you want I can also merge the classes together again. That would mean:
- Adding 2 new fields for positional and named records fields (or add 1 combined field for the record specification)
  - This is needed because during the model parsing we don't have the `DartType` available, and we need to store the structure, just like classes with with `className` + `generics` (which would be redundant if it had the `DartType`))
- Either making the `className` optional (or allow/expect `''`, as records don't have a name), or using a special `"Record"` constant here.

Having the methods like `reference` generate the appropriate code is not a big deal, as the class internally should know about the "special cases" and when using which field(s) makes sense. I rather worry that on the outside it might not be as clear which fields can/should be used.

tp:
> Just started the review, so this might be subject to change, but my hunch tells me that this should be inlined in the generated protocol file, that way we don't expose a new interface we promise to support in our serialization package.

I can share that concern. Initially I thought it was neat to have this method as plain code, so we can iterate on it faster and also know that each instance is using exactly the same code and not some one-off variant.

But I can also have it created in the protocol, especially now that we have a working version.

The bigger question for me on the final version was whether this "general" implementation is preferred, or if we would rather have "type switched" concrete implementation for each container pair (like the reverse of `deserialize` with its type checks).

tp:
> The deserialize method seems to have grown significantly there. Are we sure all that code is required?

Some type checks are duplicated currently, though only was often as they are used in the endpoints. So as that is currently just in the test-cases I don't deduplicating this will reduce the amount a lot, but let's see, maybe this is bigger :)

Also found a few seeming duplicates like this in the latest version:
CleanShot 2025-02-18 at 13 56 26@2x

Here I'll have to dig into whether they are already on `main` or not, and whether really both variants are needed.

Overall for `Record`s it just gets quite verbose, as we generate a new `if` for every (nested) type, and then we have 1 line for each field like this:

CleanShot 2025-02-18 at 13 58 46@2x


Another small improvement could be having top-level nullable type support, so like `if T is "nullable" && data == null -> return null`. That might simplify a bunch of the mappings, but not yet sure if we could map like that.

For now I'll try the deduplication and see if I spot any other obvious errors that could lead to less code being generated.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
SandPod: Aah, that makes sense. In that case, I would simply leave a comment.

Tagging it with `NOTE(tp)` makes me assume there is some form of action that you specifically need to handle and makes it hard to understand if it is safe to remove or if it becomes outdated.

SandPod:
> The bigger question for me on the final version was whether this "general" implementation is preferred, or if we would rather have "type switched" concrete implementation for each container pair (like the reverse of deserialize with its type checks).

One thought now, is it possible to extend records? If so, maybe we could place it with the other extensions that allows us to serialize complex types? `packages/serverpod_serialization/lib/src/extensions/serialization_extensions.dart`

tp:
> If so, maybe we could place it with the other extensions that allows us to serialize complex types?

@SandPod Yes, we could use extension methods on records, but they'd have to live in the per-project generated code, as we'd need one for each concrete type, like e.g. `(bool, MyModel)`.

I thought about using `extension`s instead of "inlining" the implementation in the `Protocol` methods, but in the end it'll be even more code, just in another place.

This is further complicated by the `deserialize`, where it'd have to be a static method on the extension, but there is no default name to even use to call that method. So I'd have to turn e.g. `(bool, MyModel)` into something like:

```dart
extension _Record_Bool_MyModel on (bool, MyModel) {
  static (bool, MyModel) deserialize(Map) { … }
}
```

This would work okay for now inside the `Protocol`'s file (or a special one with the record mappings). `deserialize` would still have to forward to this one.

In order to support records in models (from YAML, in #3239) though we'd have to make the extension "public" so that the model's method could call into it as well.
We could still hide them externally though (not exporting them via the generated code), so that they would not clutter up the consuming code (or even worse conflict with similar records types defined in other projects).

tools/serverpod_cli/lib/src/generator/types.dart
tp: Just FYI (from our discussion): As discussed in person, there is really not much to refactor on `TypeDefinition` right now, as the "class" and "enum" support utilize most of the same functionality.

So a split is only becoming useful now that we add the record subtype, which uses a different structure, and this can not really be implemented separately, as without any usage / tests, we wouldn't see that it's a useful and correct split.

So we'll leave that in this PR for now.

The `RecordTypeDefinition` will also grow and get bit more interesting in #3239 when we support records in models, as we then have to support the whole structure "in our code" and not just in the underlying `RecordType` (which we don't have at that stage), much like we currently do with `generics` for class types.

packages/serverpod_serialization/lib/src/serialization.dart
tp: As discussed, will remove this from the public API and rather do it in the protocol.

Either like this as a global helper, or as extension methods should we proceed with that approach in https://github.com/serverpod/serverpod/pull/3218#issuecomment-2668491712.

tp:
> The inflation of the protocol file from ~4000 lines of code to ~65000 lines of code does not seem to be proportional to the 20 or so endpoints added that used records.

Ups, my bad there was a nesting error in the generated loop, so that we generated the record types for each endpoint method (across the whole server)…

Now that is corrected :)

tests/serverpod_test_server/lib/src/generated/protocol.dart
```diff
@@ -6752,11 +6752,6 @@ class Protocol extends _i1.SerializationManagerServer {
     if (t == List<int>) {
       return (data as List).map((e) => deserialize<int>(e)).toList() as T;
     }
-    if (t == _i1.getType<List<String>?>()) {
```
tp, line 6755: FYI @SandPod The deduplication fix also found some other instance where `deserialize` code was generated for the same type multiple times (because return types and parameters where each iterated individually, without any deduplication).

Now we first fine all the types and then generated in a single loop.

I preserved the iteration order, so that now the diff in here is minimal.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
tp: Okay, removed the initials and `NOTE` prefix from such general / long-term comments.

People can always use `git blame` to see by whom (and when/why) this was added.

tests/serverpod_test_server/lib/src/endpoints/record_parameters.dart
tp: removed all the streaming methods from the tests for now.

Will re-add them in a follow up, as that requires further code-gen support.

packages/serverpod_serialization/lib/src/serialization.dart
tp: Changed to `throw` (and also moved to the protocol, so we don't expose this globally).

tests/serverpod_test_client/lib/src/protocol/record_serialization.dart
```diff
@@ -0,0 +1,337 @@
...
+// ignore_for_file: public_member_api_docs
+// ignore_for_file: type_literal_in_constant_pattern
+// ignore_for_file: use_super_parameters
+
+// ignore_for_file: no_leading_underscores_for_library_prefixes
+import 'package:serverpod_test_client/src/protocol/simple_data.dart' as _i1;
+
+typedef _RecordOfint = (int,);
+
+extension Ext__RecordOfint on _RecordOfint {
```
tp, line 16: @SandPod I think due to the requirement (in order to use it externally) of naming the record type, this is a dead end (as it's super ugly 😅)

Also the call side is not as nice as possible, as we probably don't just want to do `record.mapRecordToJson()`, but rather `_i999.Ext__RecordOfint(record).mapRecordToJson()` in order to avoid conflicts if multiple imports had defined the same extension. (Though for our own generated code we can practically prevent this, still it might be worse than before.)

Due to my current suspicions I did not polish this up as usual, but I think it already shows how it would generally look.

Let me know what you think, maybe there is a way to make it way nicer.

For the call side (endpoint transformations) I don't think there is a huge gain in the end from `mapRecord(record)` to `record.mapRecord()`, and in this file is rather seems like a regression with those weird type names.

(Technically the `typedef` is optional though. We could also declare the extension on the "real" type, but still the extension would need some kind of name.)

SandPod: I agree, this doesn't look very pretty.

The previous solution, where we handled this in the generated code, might actually have been better. Seems like we didn't gain that much with this.

tests/serverpod_test_server/lib/src/endpoints/record_parameters.dart
tp: Moved all endpoint methods into `#region`s and used matching `group`s in tests.

Now all declared endpoint methods are also called at least once.

tests/serverpod_test_client/lib/src/protocol/record_serialization.dart
tp: Reverted the `extension` approach and instead used "free" functions in the `protocol.dart` file.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -1048,6 +1148,240 @@ class LibraryGenerator {
...
+          _addRecordType(type, recordTypes, handledTypes);
+        } else {
+          _addTypeAndCollectRecords(type, recordTypes, handledTypes);
+        }
+      }
+    }
+
+    // TODO(tp): Once we allow `Record`s in the model definitions, we need to extract them here as well
+    //           Currently the `model.type.dartType` is always `null` here though, so we can not iterate over the models' fields
+    // for (var model in models) {}
```
SandPod, line 1233: Suggestion: Remove personalized comment and incomplete `for each` code.

Is it worth saving any form of hint here?

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -1542,12 +1550,14 @@ class SerializableModelLibraryGenerator {
         subDirParts: subDirParts,
         config: config,
         nullable: false,
-      );
-      TypeReference tableType = field.type.reference(serverCode,
-          subDirParts: subDirParts,
-          config: config,
-          nullable: false,
-          typeSuffix: 'Table');
+      ) as TypeReference; // relations must point to another model type
```
SandPod, line 1553: Suggestion: Remove the comment.

The responsibility for validating this is in the analyzer. Do we have a test that validate that we can't use a record in a relation?

tools/serverpod_cli/test/generator/types_test.dart
```diff
@@ -0,0 +1,145 @@
...
+
+void main() {
+  var emptyRecordType = RecordType(
+    nullabilitySuffix: NullabilitySuffix.none,
+    positional: [],
+    named: {},
+  );
+
+  test(
+      'When creating a `TypeDefinition` from a `RecordType` with a positional field, then the fields are stored in `generics`',
```
SandPod, line 14: Suggestion: Add `Given` part to these tests.

`Given`, `When`, `Then` is the pattern we would like to enforce in the codebase. It is not perfectly reflected in all placed but we try to apply and reshape existing tests when practical.

I think this would be helpful for these tests as well.

 `Given "RecordType" with a positional field when creating "TypeDefinition" then positional field is stored in "generics" without field name`

tools/serverpod_cli/test/generator/types_test.dart
```diff
@@ -0,0 +1,145 @@
...
+      () {
+    var typeDefinition = TypeDefinition.fromDartType(
+      RecordType(
+        nullabilitySuffix: NullabilitySuffix.none,
+        positional: [emptyRecordType],
+        named: {},
+      ),
+    );
+
+    expect(typeDefinition.className, 'Record');
```
SandPod, line 24: Suggestion: Break this out into a separate test.

Seems like this could have a different reason to fail and is a different check compared to the generics part.
This is now also repeated in all tests, it should be enough to validate this once right?

tests/serverpod_test_server/test_integration/test_tools/serverpod_test_tools.dart
```diff
@@ -7578,6 +7584,845 @@ class _OptionalParametersEndpoint {
   }
 }
 
+class _RecordParametersEndpoint {
```
SandPod, line 7587: Suggestion: Could you add tests to the test tools that validate the record parameters and return types work as expected?

The test tools have a different code generation flow, I'm concerned the solution might only work for the http case.

SandPod (commented):
Nice work!

I would like to see tests that validate we can use these in the test framework. I know that the code generation solution there is different so we might need to validate it works as expected.

tools/serverpod_cli/test/generator/types_test.dart
tp: > Seems like this could have a different reason to fail and is a different check compared to the generics part.
> This is now also repeated in all tests, it should be enough to validate this once right?

That is true. But since the input is different in every case, I personally would like to check all invariants that the code relies upon.

Because if we only test this once we'd only do it with some "basic" definition, because we don't want to repeat the complicated test setup 3 times to assert 3 different properties.

To me it would be fine if say 3 tests say for the same (and obvious) reason, then you fix one, re-run the test and everything goes green again. (Especially if you have a chain of assertion, you could say: Oh, invariants 1 and 2 hold up, but 3 does somehow not anymore. Whereas in the tiny test-case version you only see "3 is not true", and then I'd have to dig into "what is even correct until now" to find out where the whole thing went off the rails.)

But this request to have the tiniest possible number of assertions also came up before, so in order to align with the project structure, I will trim them down again.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -346,6 +335,85 @@ class LibraryGenerator {
 
     library.body.add(protocol.build());
 
+    if (recordTypesToDeserialize.isNotEmpty) {
+      library.body.addAll([
```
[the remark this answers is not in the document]
tp: Extracted method.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -752,12 +820,36 @@ class LibraryGenerator {
...
+    );
+
     return refer('caller').property('callServerEndpoint').call([
       literalString('$modulePrefix${endpointDef.name}'),
       literalString(methodDef.name),
       literalMap({
         for (var parameterDef in params)
-          literalString(parameterDef.name): refer(parameterDef.name),
+          // The generated classes implement `ProtocolSerialization` and get handle by `serverpod_serialization` later
+          // For the records we need to transform then into a map that can be handled by the shared (non-project specific) serialization code
```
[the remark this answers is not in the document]
tp: fixed typo

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
tp: removed commented out code

tools/serverpod_cli/test/generator/types_test.dart
tp: Used strict `given, when, then` pattern.

tools/serverpod_cli/test/generator/types_test.dart
tp: removed repeated assertions.

tests/serverpod_test_server/test_integration/test_tools/serverpod_test_tools.dart
tp: There was indeed some work needed for support the mapping of incoming / outgoing record types in the test tool generated code paths.

This is now support and an integration test is included in https://github.com/tp/serverpod/blob/fc5c9724757bfc01107d66fc93c7e1ae6949d4e8/tests/serverpod_test_server/test_integration/test_tools/return_and_input_types_test.dart#L123

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
SandPod: Sounds good! 👍

tests/serverpod_test_server/lib/src/endpoints/test_tools.dart
```diff
@@ -183,6 +183,20 @@ class TestToolsEndpoint extends Endpoint {
     return simpleDatas;
   }
 
+  Future<(String, (int, bool))> echoRecord(
```
SandPod, line 186: Question: Why are these additional endpoints required in this file as opposed to the other record type endpoints in the normal `record_parameters.dart` file?

tests/serverpod_test_server/test_integration/test_tools/return_and_input_types_test.dart
```diff
@@ -119,6 +119,28 @@ void main() {
         expect(result[0].num, 1);
         expect(result[1].num, 2);
       });
+
+      test('when calling echoRecord then should return the record', () async {
```
SandPod, line 123: Question/Suggestion: Please reflect all calls that where tested through the client.

Is there any reason for only testing a minimal set here? Is is harder for us to introduce breaking changes in the test tools? If not, please reflect all the client endpoints through the test tools as well.

We could probably break these out into a separate file.

SandPod (commented):
Implementation looks good now! Single concern around test coverage for the test tools!

tools/serverpod_cli/test/generator/types_test.dart
SandPod: We try to mitigate this by utilizing groups to branch out and validate multiple things for a specific setup. Then, we can define what is shared for the group and validate each property or assertion in isolation when they are not related.

tests/serverpod_test_server/lib/src/endpoints/test_tools.dart
tp: I just followed the existing pattern where there is 1 method / method pair for each "kind" of type in here.

I could also re-use some of the methods defined in `record_parameters.dart`, but then this would be only methods called from another endpoint.

I think the idea is that one can easily see the comprehensiveness of the test (e.g. the server is providing the `n` methods for test-tool testing, and here we are using them all in a single file).

tests/serverpod_test_server/test_integration/test_tools/return_and_input_types_test.dart
tp: "record" and "container of record" are the only 2 distinct types of new cases we need to handle (as they generate different (de)serialization code).

The other tests try out various combinations like more deeply nested stuff etc. to make sure that the code generated also handles cases many levels down, but on the endpoint level (sending / receiving) the code is always that 1 line transformation for plain records and containers with records.

tp: As discussed, no need to test all type combinations again, as the top-level mappings would be the same, and there each kind of structure is covered now.

tests/serverpod_test_server/lib/src/endpoints/test_tools.dart
tp: Since we only keep these 2 and there is a clean mapping between endpoint and test file, we'll keep them in here.

SandPod: Makes sense! :)

SandPod (approved):
LGTM! 🚀

PR #3219, opened 2025-02-10
Support containers in Stream endpoints, like Stream<List<Foo>>

tp:

## Pre-launch Checklist

## Breaking changes


tests/serverpod_test_server/lib/src/generated/protocol.dart
```diff
@@ -7597,6 +7606,219 @@ class Protocol extends _i1.SerializationManagerServer {
...
+      return 'Map<String,SimpleData?>?';
+    }
+    if (data is Map<String, Duration>) {
+      return 'Map<String,Duration>';
+    }
+    if (data is Map<String, Duration?>) {
+      return 'Map<String,Duration?>';
+    }
+    if (data is List<_i3.UserInfo>) {
+      return 'List<UserInfo>';
```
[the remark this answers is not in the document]
tp: @SandPod Do you have a hint from the top of your head to get to the module name?

Else I would try to take the list of linked modules, and try to determine the right one from the import path of the model class (probably needing to take client/server paths into account).

tp:
@SandPod Sorry about the bunch of commits, but I figured out some edge-case corrections and simplifications after the initial E2E success. Since it's not that many files (and even less non-generated), I would just leave it like this and don't spend a ton of time re-committing the same code in different steps if that's fine with you.

SandPod:
> @SandPod Sorry about the bunch of commits, but I figured out some edge-case corrections and simplifications after the initial E2E success. Since it's not that many files (and even less non-generated), I would just leave it like this and don't spend a ton of time re-committing the same code in different steps if that's fine with you.

Totally fine, I'll just review the final diff :)

tests/serverpod_test_server/lib/src/endpoints/method_streaming.dart
```diff
@@ -178,6 +179,81 @@ class MethodStreaming extends Endpoint {
...
+    Stream<Set<SimpleData>> simpleDataSetStream,
+  ) async* {
+    await for (var data in simpleDataSetStream) {
+      yield data;
+    }
+  }
+
+  Stream<Set<SimpleData>> nestedSetInListInOutDataStream(
+    Session session,
+    // NOTE(tp): Important that this type is only declared once as a parameter, but still code is generated for it
```
SandPod, line 247: Chore: Change to a normal comment.

tests/serverpod_test_server/test_e2e/method_streaming/out_stream_test.dart
```diff
@@ -75,4 +76,223 @@ void main() {
...
+          Stream.fromIterable([
+            [1, 2, 3],
+            [4, 5, 6],
+          ]),
+        )
+        .toList();
+
+    expect(
+      response,
+      equals([
```
SandPod, line 94: Suggestion: Change to `emitsInOrder` since this is a specialized matcher for streams and events.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -52,6 +53,22 @@ class LibraryGenerator {
...
+    var topLevelStreamContainerTypes = <TypeDefinition>[];
+    for (var topLevelType in protocolDefinition.endpoints
+        .expand((e) => e.methods)
+        .expand((m) => [m.returnType, ...m.allParameters.map((p) => p.type)])
+        .where((t) => t.isStreamType)) {
+      var valueType = topLevelType.generics.first;
+      if (valueType.isSetType || valueType.isListType || valueType.isMapType) {
+        if (valueType.dartType == null ||
+            !topLevelStreamContainerTypes.any((type) =>
+                // TODO(tp): Use `isStructurallyEqualTo` in the future (not supported by current lower bound of Flutter 3.19)
```
SandPod, line 65: Suggestion(NIT): Remove comment.

If this is functionally the same as the code that is already here, this optimization won't be important for future readers of this code.

tests/serverpod_test_server/lib/src/generated/protocol.dart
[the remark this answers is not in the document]
SandPod: Great! Is there need for a specific test that validates we make the right decision here?

SandPod (approved):
Awesome!

PR looks good to me!
I only had a couple of NIT suggestions.

Approving on the basis that these are fixed.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
tp: Removed un-actionable (for now) `TODO`

tests/serverpod_test_server/test_e2e/method_streaming/out_stream_test.dart
tp: Switched to `emitsInOrder`. Initially it did not work for me, hence the `toList` workaround. But now it was fine for all tests, so I must've done something wrong that lead me to this workaround in the first place.

Fixed now.

tests/serverpod_test_server/lib/src/generated/protocol.dart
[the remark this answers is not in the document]
tp: So far that method was only tested "on the side" in a websocket (streaming method) test:

https://github.com/tp/serverpod/blob/6ed324e82ab36ffb7ba9c554841710add7262df9/tests/serverpod_test_server/test_e2e/websocket_test.dart#L46

I now added https://github.com/tp/serverpod/blob/bef3a6243f639db0a77ad28d1280061d40536ead/tests/serverpod_test_server/test/class_name_for_object_test.dart#L9 to test the name building for containers with primitives and modules types, and ensuring that they are consistent across client and server. (Which of course they are since E2E is working, but will be great to have them explicitly spelled out in a test case, such that we don't accidentally change the format in the future.)

PR #3221, opened 2025-02-10
Speed up local E2E tests in Docker container

tp:
by only copying the local files after the container is set up (apt-get being done)

That way Docker can use a cache for the previous build steps, and just needs to re-run `pub get` on each run as the files might influence that.


Isakdl (approved):
LGTM

PR #3238, opened 2025-02-18
feat: Support setting serverId with SERVERPOD_SERVER_ID env var

henycave:
Added the ability to set the serverId from environment variables (`SERVER_ID`) in addition to the command line argument 
(`--server-id`). If both options are provided, the command line argument takes precedence. 

Changes: 
Serverpod Constructor Update: Modified the Serverpod constructor to accept the serverId as a parameter. 
Template Changes: Updated templates to reflect the changes.
Test Addition: Added a test to verify that the `serverId` can be successfully set from the `SERVER_ID` environment variable.


tests/serverpod_test_server/test_integration/server_set_server_id_from_env.dart
```diff
@@ -0,0 +1,23 @@
+import 'dart:convert';
+
+import 'package:serverpod_test_server/test_util/server_test_util.dart';
+import 'package:test/test.dart';
+
+/// This test checks if the server starts with the SERVER_ID set from environment variable.
+/// Before running this test, ensure that Docker (docker-local) is up and that the server is not running.
+void main() {
+  test('Server should start with SERVER_ID set', () async {
```
christerswahn, line 9: Please have tests for having either the env var or the command line arg set, and both while verifying the correct one is chosen.

templates/serverpod_templates/projectname_server/lib/server.dart
```diff
@@ -6,10 +7,19 @@ import 'src/generated/endpoints.dart';
 // This is the starting point of your Serverpod server.
 void run(List<String> args) async {
   // Initialize Serverpod and connect it with your generated code.
+
+  // You can set the serverId using either:
+  // 1. A command-line flag: --server-id=<value>
+  // 2. The 'SERVER_ID' environment variable
```
christerswahn, line 13: Please follow the naming convention with the SERVERPOD_ prefix, this variable should probably be named SERVERPOD_SERVER_ID.

christerswahn (changes requested):
This is a nice feature to have!

It seems the command line argument handling this PR describes is missing from the code? This needs to be added in a manner consistent with the other command line parsing.

The name of the environment variable needs to follow the naming convention with the SERVERPOD_ prefix, i.e. SERVERPOD_SERVER_ID.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -307,7 +310,11 @@ class Serverpod {
     stdout.writeln(commandLineArgs.toString());
 
     _runMode = commandLineArgs.runMode;
-    serverId = commandLineArgs.serverId;
+
+    // Set the serverId to the value of commandLineArgs.serverId if it is not 'default'.
+    if (commandLineArgs.serverId != 'default') {
```
christerswahn, line 315: This code section needs to handle the cases where the command line arg is provided, the env var is set, both are set, or none of them is set.

henycave: The default value from the command line is 'default'.
`..addOption(
          'server-id',
          abbr: 'i',
          defaultsTo: 'default',
        )`

and the default value from the Serverpod contractor is 'default'

`Serverpod(
    List<String> args,
    this.serializationManager,
    this.endpoints, {
    this.serverId = 'default',
  })`
  
 So if the serverId is set from the command line it would be different from 'default' in that case we assign it to that value. if it's default then we ignore it and use the environment SERVER_ID value which also can be a different value or the default 'default'. In the case where both are set it will give precedence to the command line arg `--server-d` So, in this code, it handles all the cases. I will also add tests for all the cases, which will help clear all of this.

henycave: I added test cases to cover all the cases.

templates/serverpod_templates/projectname_server/lib/server.dart
henycave: I have changed the name to SERVERPOD_SERVER_ID.

tests/serverpod_test_server/test_integration/server_set_server_id_from_env.dart
henycave: Added tests for all the cases. Please check now.

templates/serverpod_templates/projectname_server/lib/server.dart
```diff
@@ -6,10 +7,19 @@ import 'src/generated/endpoints.dart';
...
 void run(List<String> args) async {
   // Initialize Serverpod and connect it with your generated code.
+
+  // You can set the serverId using either:
+  // 1. A command-line flag: --server-id=<value>
+  // 2. The 'SERVERPOD_SERVER_ID' environment variable
+  //
+  // If both are set, the command-line flag takes precedence.
+  // If neither is set, the default value 'default' will be used.
+  final serverId = Platform.environment['SERVERPOD_SERVER_ID'] ?? 'default';
```
christerswahn, line 17: Handling of environment variables is centralized and we want to avoid special handling of individual variables.

Please avoid modifications to the run() function templates.

Instead, incorporate the arg and env var handling together with the other configuration handling, see e.g. config.dart.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -288,6 +290,7 @@ class Serverpod {
     List<String> args,
     this.serializationManager,
     this.endpoints, {
+    this.serverId = 'default',
```
christerswahn, line 293: We don't want to modify the constructor interface for this type of change. Please see the configuration handling instead.

christerswahn (changes requested):

templates/serverpod_templates/projectname_server/lib/server.dart
henycave: Can you explain a bit more? So should I add a new configuration.yaml just for serverId, or add serverId field inside all the phases like development.yaml, production.yaml, staging.yaml, and test.yaml? And inside the config.dart I should read the config files and assign the serverId? I am not sure But if we do that wouldn't it defeat the purpose of assigning different serverIds for replicated servers in k8s cluster since this configuration.yaml file is static?

henycave: Modifications to the templates have been removed, and assigning the `serverId` is now handled inside the `config.dart`.

henycave: Also, the highest precedence is given to the command line arg `--server-id`, after that, it's given to the environment variable SERVERPOD_SERVER_ID lastly, it's given to the config file field `serverId`. If all are not assigned it will be assigned to 'default'.

packages/serverpod/lib/src/server/serverpod.dart
henycave: It has been removed in the new implementation.

henycave:
In the latest implementation, we can assign the server ID from the command line arg, the environment, or the config files. the highest precedence is given to the command line arg `--server-id`, after that, it's given to the environment variable `SERVERPOD_SERVER_ID` lastly, it's given to the config file field `serverId`. If all are not assigned it will be assigned to `'default'`.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -363,6 +363,7 @@ class Serverpod {
         );
     logVerbose(this.config.toString());
 
+    serverId = this.config.serverId;
```
christerswahn, line 366: This needs to be moved up to before _innerInitializeServerpod is invoked, since serverId is required to be set before then (and it should not have a different temporary value before that). So please move up the config load to _initializeServerpod and replace the line `serverId = commandLineArgs.serverId;` there.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -594,6 +596,18 @@ int _readMaxRequestSize(
...
 }
 
+String _readServerId(Map<dynamic, dynamic> configMap,
+    Map<String, String> environment, String serverIdFromCommandLineArg) {
+  if (serverIdFromCommandLineArg != 'default') {
+    return serverIdFromCommandLineArg;
+  }
+  var serverId = configMap[ServerpodEnv.serverId.configKey];
+
+  serverId = environment[ServerpodEnv.serverId.envVariable] ?? serverId;
```
christerswahn, line 606: Please chain these assignments into one expression chain using ??, it makes it easier to follow.

christerswahn (changes requested):
Good job! Just a couple more comments. :+1:

packages/serverpod/lib/src/server/serverpod.dart
henycave: Sure! So we can still catch some errors or rethrow them, I have moved it inside the try catch.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -330,10 +330,25 @@ class Serverpod {
     stdout.writeln(commandLineArgs.toString());
 
     _runMode = commandLineArgs.runMode;
-    serverId = commandLineArgs.serverId;
 
     try {
-      _innerInitializeServerpod(commandLineArgs, config: config);
+      // Load passwords
```
christerswahn, line 335: Sorry I wasn't entirely clear, the password initialization currently needs to remain as it was.

The code in `_initializeServerpod` should probably be:

```dart
...
    _runMode = commandLineArgs.runMode;

    // Load config
    this.config = config ??
        ServerpodConfig.load(
          _runMode,
          commandLineArgs.serverId,
          _passwords,
        );
    logVerbose(this.config.toString());
    serverId = this.config.serverId;

    try {
      _innerInitializeServerpod();
    } catch (e, stackTrace) {
...
```

henycave: no worries, but don't we need the loaded `_password` inside the `ServerpodConfig.load`?

christerswahn: You're right. Ok move the password init up before the config load. But it needs to be before the try - catch unfortunately. (We're going to overhaul this init sequence later.)

henycave: ok great

christerswahn (approved):
LGTM, good job! 👍 

We're probably cutting a new release this week and this can be included.

PR #3239, opened 2025-02-18
Support Record declarations in YAML model files

tp:
Update model class generation, including support for `copyWith` on record types.

Relates to #1784

#### TODO
- [ ] Rebase once #3218
- [ ] Apply applicable suggestions from #3218 also to the new code written in here
- [ ] Finish up model class generation


tools/serverpod_cli/lib/src/generator/dart/library_generators/server_test_tools_generator.dart
```diff
@@ -565,6 +568,12 @@ extension on ParameterDefinition {
           ]),
         ]).code,
       ]);
+    } else if (isModelFromProject) {
+      if (type.nullable) {
```
tp, line 572: A bit surprised that I had to add this, but otherwise when using a plain model class as an argument to an endpoint method, the test would crash, because the model (not it's JSON representation) would be given to `Protocol.deserialize`.

SandPod:
@tp - Please split this PR up into separate commits for Model validation and parsing changes, code generation changes, the added models requird to test this functionality and then lastly a commit with the generated code.

Each commit should also contains the tests added for each modified part of the code.

The context of this single commit is to large to communicate any meaningfull context.
By splitting it up, you speed up and improve the review that I can deliver, and you also make sure this change is easily understood for future developers when we inveitably revisit this feature.

tests/serverpod_test_client/test/copy_with_mutable_test.dart
```diff
@@ -111,6 +121,21 @@ void main() {
...
+      var record = ('1', optionalUri: null);
+
+      var types = Types(aRecord: record);
+      var typesCopy = types.copyWith();
+      types.aRecord = ('2', optionalUri: null);
+
+      expect(
+        typesCopy.aRecord,
+        record,
+      );
```
SandPod, line 136: Discussion: This seems to suggest that my previous comment around deep copy for records is not correct.

tests/serverpod_test_server/lib/src/endpoints/record_parameters.dart
```diff
@@ -255,4 +255,30 @@ class RecordParametersEndpoint extends Endpoint {
     }
   }
 // #endregion
+
+// #region Records inside model class
+  Stream<TypesRecord> streamOfModelClassWithRecordField(
```
SandPod, line 260: Question: Shouldn't we also test a non-streaming endpoint that returns this model?

SandPod: Same for the test tools.

tests/serverpod_test_server/lib/src/generated/types_set.dart
```diff
@@ -124,6 +127,12 @@ abstract class TypesSet
               itemFromJson: (e) => (e as List)
                   .map((e) => _i5.Types.fromJson((e as Map<String, dynamic>)))
                   .toList()),
+      aRecord: jsonSerialization['aRecord'] == null
+          ? null
+          : _i1.SetJsonExtension.fromJson(
+              (jsonSerialization['aRecord'] as List),
+              itemFromJson: (e) => _i6.Protocol()
```
SandPod, line 134: Suggestion: This instantiation seems unnecessary to repeat each time we parse deserialize the class.

tests/serverpod_test_server/test_integration/database_operations/crud/update_test.dart
```diff
@@ -166,6 +167,7 @@ void main() async {
       aDuration: Duration(milliseconds: 1000),
       aUuid: UuidValue.fromString(Uuid().v4()),
       anEnum: TestEnum.one,
+      aRecord: ('test', optionalUri: Uri.parse('https://serverpod.dev')),
```
SandPod, line 170: Question: Should we test this for the other CRUD operations as well? At least the read tests that validate each type in isolation?

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/model_record_types_test.dart
```diff
@@ -0,0 +1,106 @@
...
+import 'package:serverpod_cli/src/generator/code_generation_collector.dart';
+import 'package:test/test.dart';
+
+import '../../../../test_util/builders/generator_config_builder.dart';
+import '../../../../test_util/builders/model_source_builder.dart';
+
+void main() {
+  var config = GeneratorConfigBuilder().build();
+
+  group('Given a model class with a record fields,', () {
```
SandPod, line 12: Suggestion: Break this up so that we validate one style at a time. If this starts to fail, it would be interesting to know which of these still work.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/model_record_types_test.dart
```diff
@@ -0,0 +1,106 @@
...
+      var testField11 = modelDefinition.findField('testField11');
+      expect(testField11?.type.isRecordType, isTrue);
+      expect(testField11?.type.generics, hasLength(2));
+      expect(testField11?.type.generics.first.className, 'String');
+      expect(testField11?.type.generics.first.recordFieldName, isNull);
+      expect(testField11?.type.generics.last.className, 'int');
+    });
+  });
+
+  test(
```
SandPod, line 83: Suggestion: Move this test to `model_types_test.dart` and update description to match the other onces there.

tools/serverpod_cli/test/generator/types_test.dart
```diff
@@ -152,4 +152,263 @@ void main() {
       ),
     ]);
   });
+
+  test(
+      'Given a field type in parentheses, when it is parsed, then the it is used as `className` verbatim',
```
SandPod, line 157: chore: Fix scentence in description, `...then the it is used...`

tools/serverpod_cli/test/generator/types_test.dart
```diff
@@ -152,4 +152,263 @@ void main() {
...
+      () {
+    var type = parseType('(int)', extraClasses: []);
+
+    // This is practically wrong, but at least it's not parsed as a record
+    // The analyzer will then have to warn about this not being a known type (as you can only use extra `()` on values, not types)
+    expect(type.className, '(int)');
```
SandPod, line 163: Suggestion: Please add a test for the analyzer that validates an error message is written to the user for this type.

SandPod (changes requested):
Nice to see that it is possible to get this working with the system.

Main concern is with us creating a `Protocol` for each deserialization attempt of a model that holds a record. That needs to be fixed.

I also have a couple of suggestions to harden this a bit and ensure that we can continue to work with the code. There are some questions that need answering and some additional test coverage that should be added.

tools/serverpod_cli/lib/src/generator/dart/library_generators/util/class_generators_util.dart
```diff
@@ -420,6 +428,37 @@ Expression _buildClassTypeFromJson(
...
+        : 'package:${config.dartClientPackage}/src/protocol/protocol.dart',
+  );
+
+  return CodeExpression(Block.of([
+    if (type.nullable) ...[
+      valueExpression.code,
+      const Code('== null ? null : '),
+    ],
+    protocolRef
+        .newInstance([])
```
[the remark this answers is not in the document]
tp: @SandPod It reads weird, but in practice all those `Protocol()` calls are just a getter onto the shared `static` instance.

CleanShot 2025-03-14 at 09 33 55@2x


So I don't think it's something we need to worry about. Ideally the `Protocol` interface would just be "honest" about being a singleton (e.g. making `instance` public and the constructor private), but such a change is probably out of scope here.

packages/serverpod_serialization/lib/src/copy_with.dart
```diff
@@ -36,4 +36,5 @@ const _nonMutableTypes = [
   UuidValue,
   Uri,
   BigInt,
+  Record,
```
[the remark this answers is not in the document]
tp: Right, if the record contains any mutable containers we should have the same behavior as e.g. `Map<String, Map<…>>` and deep copy both the outer and inner `Map`'s values.

tools/serverpod_cli/lib/src/generator/dart/library_generators/server_test_tools_generator.dart
tp: > How does this work with models that are part of a module?

I'll add a test for that. Will be interesting how I can identify those, as the problem here was that no part of the stack seems to call `toJson` on the model class, but on the receiving end it is passed to `deserialize` which expect the JSON "map" representation (and not a class instance, which would already be the result).

This was the best I could come up with to detect when it would be allowed (and necessary) to call `toJson()` on the parameter.

> Or `Custom` models?

What do you mean with "custom" here? It currently does work fine (and is tested) for models defined in the same project.

tp: Changed this to generate "toJSON" (though we have to use `SerializationManager.encode` (which gives a `String`) + `jsonDecode` in order to get the JSON "Map" representation, as the model's `toJson` is not "deep").

Now this also works for external models.

tools/serverpod_cli/lib/src/generator/types.dart
```diff
@@ -681,6 +695,10 @@ TypeDefinition parseType(
 }) {
   var trimmedInput = input.trim();
 
+  if (_describesRecord(trimmedInput)) {
+    return _parseRecord(trimmedInput, extraClasses: extraClasses);
+  }
```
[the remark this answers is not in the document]
tp: okay, grouped implementation and commits in a single commit from now on.

tools/serverpod_cli/test/generator/types_test.dart
tp: Added a test for this (the analyzer was already correctly err-ing, as to be expect since the type defined in `className` could not be resolved).

packages/serverpod_serialization/lib/src/copy_with.dart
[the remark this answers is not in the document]
tp: Records will now be deep copied when no replacement is given to `copyWith`.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -954,6 +957,21 @@ class Restrictions {
...
+      if (fieldType.generics.isNotEmpty) {
+        errors.addAll(
+          fieldType.generics.expand(
+            (field) => _validateFieldDataType(field, span),
+          ),
+        );
+      } else {
+        errors.add(
+          SourceSpanSeverityException(
+            'A record type must have at least one field defined (e.g. (int,)).',
```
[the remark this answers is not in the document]
tp: As of now you can not create this by any model definition, as the code would not recognize a "zero field record" like `(,)` as a record.

So should we then remove this here, or keep it in, as the type definition object would allow for this invalid state and thus we should check that it not (somehow) got into it?

tools/serverpod_cli/lib/src/generator/types.dart
```diff
@@ -716,6 +734,115 @@ TypeDefinition parseType(
...
+
+  return (fields, nullable);
+}
+
+TypeDefinition _parseRecord(
+  String trimmedRecordInput, {
+  List<TypeDefinition>? extraClasses,
+}) {
+  assert(trimmedRecordInput.startsWith('('));
+  assert(trimmedRecordInput.endsWith(')') || trimmedRecordInput.endsWith('?'));
```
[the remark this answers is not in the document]
tp: changed `assert`s to runtime error to guard against invalid inputs (just to make sure that this code is only ever called with valid inputs, and we don't have any weird downstream effects).

tools/serverpod_cli/lib/src/util/string_manipulation.dart
```diff
@@ -1,10 +1,11 @@
 import 'package:super_string/super_string.dart';
 
 /// Splits a string on the separator token unless the token is inside
-/// brackets, angle brackets, ( ) and < >, single quotes '', or double quotes "".
+/// brackets, angle brackets, ( ) and < >, single quotes '', or double quotes "", or curly braces {}.
 List<String> splitIgnoringBracketsAndQuotes(
   String input, {
   String separator = ',',
+  bool includeEmpty = false,
```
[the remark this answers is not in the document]
tp: I think "include" was better than "allow" because the method would still work (just differently) with empty string and the flag being `true`. It would not err, what "allow" might imply.

But I changed it to `returnEmptyParts`, which would be pretty much what I would leave as a doc comment on this parameter, so that should simplify and clarify the situation.

Also added a test for this case.

tools/serverpod_cli/lib/src/util/string_manipulation.dart
```diff
@@ -49,7 +50,7 @@ List<String> splitIgnoringBracketsAndQuotes(
     }
   }
 
-  if (current.isNotEmpty) {
+  if (current.toString().trim().isNotEmpty || includeEmpty) {
     result.add(current.toString().trim());
```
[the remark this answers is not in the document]
tp: changed to only `trim` once

tests/serverpod_test_server/lib/src/endpoints/record_parameters.dart
tp: added test for `Future` in test tools as well.

tests/serverpod_test_server/test_integration/database_operations/crud/update_test.dart
tp: I am not sure what you're asking here. I modeled the test based on the others, and they all do the same for each field.

But now that I re-read the test in detail, I am questioning its premise.

It always write a "full" `Types` object into the database and then replaces it with a `null`-ed one (with only the current "in focus" field being set to `null`, but all others being so implicitly), and then check on the return value (which does not tell one much, as the function could just be `(model) => model`).

So I think a more interesting test would be to first always read the "default" object by ID and then assert the field is there, and then update it.

Bonus points if one already wants to include the "column update" functionality, check that only the desired column is affected (though there might be more specific tests for that).

So the way the test are currently written IMHO is just testing `insert` once and `update` a bunch of times, but basically always with the same flow (as all the fields are always null in each test).

Maybe this would be more fitting as a follow up issue to see what cases are covered, and either find a way to make those tests turn up theoretical implementation issues, or, if we agree that they all just test the same "update Types(null)" scenario, to reduce it down to 1.

CleanShot 2025-03-17 at 13 52 56@2x

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/model_record_types_test.dart
tp: Makes sense, moved it.

I think that test case is also a prime example to be something like a [table driven](https://go.dev/wiki/TableDrivenTests) with a list of keywords to check, and then iterating all of them.

Seeing that some tests of those kinds now made it to `main` (even though my previous information was that we should rather have small, explicit cases), I thought I'd bring it up again for consideration.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/model_record_types_test.dart
tp: broke it up into 1 test case per field

tests/serverpod_test_server/lib/src/endpoints/record_parameters.dart
SandPod: I might be missing something but I can't seem to find the e2e tests added for this that are not run through the test tools?

I was expecting them to be added to `record_types_test.dart`.

tests/serverpod_test_server/test_integration/database_operations/crud/update_test.dart
SandPod: I think that is the purpose. The file you are inspecting is called `update_test.dart` so `update` is the focus of this test file.

The other files in the same folder test other aspects of the database operations. My suggestion was that we should ensure that records are also represented in these, where it makes sense. I think it should at least be represented in the `read_test.dart` file.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/model_record_types_test.dart
SandPod: Yes, this is actually an example where it might make sense to use an iterator for the test since we are validating the exact same thing.

But I still believe that it should be handled with care, as complexity in tests is easy to understand when you write them but makes the intent harder to get in the future.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/stateful_analyzer_test.dart
```diff
@@ -309,4 +309,30 @@ and neither is this line
...
+        StatefulAnalyzer(config, [yamlSource], (uri, errors) {
+      reportedErrors = errors;
+    });
+
+    statefulAnalyzer.validateAll();
+
+    expect(reportedErrors?.errors, hasLength(1));
+    expect(
+      reportedErrors?.errors.single.message,
+      contains('invalid datatype "(String)"'),
```
SandPod, line 335: Question/Discussion: Is this the normal message we display for unknown types? Is it easy for us to give a to the user here? Like `Did you mean "String" or "(String,)"?

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/model_record_types_test.dart
```diff
@@ -0,0 +1,438 @@
...
+import 'package:serverpod_cli/src/generator/code_generation_collector.dart';
+import 'package:test/test.dart';
+
+import '../../../../test_util/builders/generator_config_builder.dart';
+import '../../../../test_util/builders/model_source_builder.dart';
+
+void main() {
+  var config = GeneratorConfigBuilder().build();
+
+  group('Given a model class with a record field,', () {
```
SandPod, line 12: Suggestion: The given are all different for these tests.

Please reflect the difference in the group description so that it is clear under which circumstances the test is failing if it fails in our CI.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/model_record_types_test.dart
```diff
@@ -0,0 +1,438 @@
...
+        fields:
+          testField: (String, int)
+        ''',
+      ).build()
+    ];
+
+    var collector = CodeGenerationCollector();
+    final analyzer =
+        StatefulAnalyzer(config, models, onErrorsCollector(collector));
+    analyzer.validateAll();
```
SandPod, line 26: Suggestion: For all new test files that use the StatefulAnalyzer, make each test possible to run in isolation.

Statements placed in the `group` are executed directly when parsing the test file. That means that if I would place a breakpoint in the analyzer, and then try to run a specific test in this file. The analyzer would still execute from all tests getting stuck in my breakpoint.

To fix this, we can simply make the analyzer either execute as part of a `setUp` or a `late` variable.

SandPod (commented):
Nice fixes! A couple of more minor comments, and answered a couple of your questions as well.

tools/serverpod_cli/test/util/string_manipulation_test.dart
```diff
@@ -59,10 +60,46 @@ void main() {
...
+      '',
+      '',
+      '',
+    ]);
+  });
+
+  test(
+      'Given a string where some separators are right next to each other when it is split with returnEmptyParts=false then no empty parts are returned',
+      () {
+    var result = splitIgnoringBracketsAndBracesAndQuotes('a,,');
```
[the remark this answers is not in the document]
tp: `a,,b,,` should be `['a', 'b']`.

Just meaning that whenever separators are right next to each other (or only separated by whitespace), then no entry or that part is returned.

Likewise `a,   ,     b   ,     ,       ` would also result in `['a', 'b']`.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/stateful_analyzer_test.dart
tp: Yeah, any type would result in this message with just the input "type" in quotes. Likewise also e.g. `"bol"` if one misspelled that etc.

Sure for the parentheses case we should handle that specifically and give suggestions (probably to drop them instead of using a single field record), but I think the developer also has enough info here and can see that the `()` were interpreted as part of the type and thus cause errors.

(I think the better approach, if we spent more time here, would just to drop the `()`, just like Dart.)

tests/serverpod_test_server/test_integration/database_operations/crud/update_test.dart
tp: > I think it should at least be represented in the read_test.dart file.

Agreed, I also think we only need to extend the `read_test`, because all others do not operate on a field type level (and `insert` and `delete` are implicitly tested in setup and tear-down).

tp:
> I might be missing something but I can't seem to find the e2e tests added for this that are not run through the test tools?

There are these E2E test for "record fields in models"

The Streaming once are specifically here in the file you would expect: https://github.com/tp/serverpod/blob/da1e2a9c7ec423ba562a10efe64703a5aa2f8250/tests/serverpod_test_server/test_e2e/record_types_test.dart#L686

Where the "simple" parameters / `Future` return value is already covered in the `Types` E2E persistence test: https://github.com/tp/serverpod/blob/da1e2a9c7ec423ba562a10efe64703a5aa2f8250/tests/serverpod_test_server/test_e2e/database_roundtrip_type_test.dart#L225

As discussed in our call today, we could also break out simple "echo" (non persisting) tests for the model types, but the question remains whether that really helps to find an issue earlier as you'd have to be lucky to look at the "smaller" test case first.

But if you'd like I can add `Future<TypesRecord> echoModel*` endpoints to the test endpoint and add those cases in E2E.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/model_record_types_test.dart
tp: Okay. I was just thinking we'd go by line number, because if I re-describe the field in the test case name, it could just as easily diverge as is the standing rationale for not having code comments in "simple cases".

But I'll copy the field definitions to the title then.

tp:
> > I might be missing something but I can't seem to find the e2e tests added for this that are not run through the test tools?
> 
> There are these E2E test for "record fields in models"
> 
> The Streaming once are specifically here in the file you would expect: https://github.com/tp/serverpod/blob/da1e2a9c7ec423ba562a10efe64703a5aa2f8250/tests/serverpod_test_server/test_e2e/record_types_test.dart#L686
> 
> Where the "simple" parameters / `Future` return value is already covered in the `Types` E2E persistence test: https://github.com/tp/serverpod/blob/da1e2a9c7ec423ba562a10efe64703a5aa2f8250/tests/serverpod_test_server/test_e2e/database_roundtrip_type_test.dart#L225
> 
> As discussed in our call today, we could also break out simple "echo" (non persisting) tests for the model types, but the question remains whether that really helps to find an issue earlier as you'd have to be lucky to look at the "smaller" test case first.
> 
> But if you'd like I can add `Future<TypesRecord> echoModel*` endpoints to the test endpoint and add those cases in E2E.

Ah, I suppose this referred to the model from another module. Added E2E tests for that as well now :)

tools/serverpod_cli/lib/src/generator/types_record.dart
```diff
@@ -0,0 +1,116 @@
+import 'package:serverpod_cli/src/generator/keywords.dart';
+import 'package:serverpod_cli/src/generator/types.dart';
+import 'package:serverpod_cli/src/util/string_manipulation.dart';
+
+/// Attempts to parse a record type from a field type definition
+///
+/// Returns `null` in case the type does not describe a valid record
+TypeDefinition? tryParseRecord(
```
SandPod, line 8: Discussion: Could we achieve this encapsulation as an extension instead?

The filename suggests that this would be a separate type of `TypeDefinition`.

Could we encapsulate this in an extension on `TypeDefinition`?
We should reflect that in the file name and the file location as well.

tools/serverpod_cli/lib/src/generator/types_record.dart
```diff
@@ -0,0 +1,116 @@
...
+Iterable<TypeDefinition> _parseNamedRecordFields(
+  String namedRecordFieldsPart, {
+  List<TypeDefinition>? extraClasses,
+}) sync* {
+  if (!namedRecordFieldsPart.startsWith('{') ||
+      !namedRecordFieldsPart.endsWith('}')) {
+    throw Exception(
+      '"$namedRecordFieldsPart" does not described the named parameters of a Record',
+    );
+  }
```
SandPod, line 82: Suggestion: Remove this.

As far as I can tell this is already validated on the only callsite.

Are there more synergies that can be simplified now that this is an implementation detail of `tryParseRecord`?

SandPod: This could actually be converetd to a `_tryParseNamedRecordFields` that returns null if it does not describe the named parameters of a Record.

Then we only need to have this check in a single place and we don't need any implicit information.

SandPod: (And the existance of these could be validated using the `start` and `end` variables.)

tests/serverpod_test_server/lib/src/endpoints/record_parameters.dart
SandPod: > But if you'd like I can add Future<TypesRecord> echoModel* endpoints to the test endpoint and add those cases in E2E.

Thanks for adding the tests. I think it makes sense to cover this explicitly as we might remove or modify the other tests for different reasons resulting in us not covering this scenario at all anymore.

SandPod (commented):
Minor comments on the `types_record.dart` file then we are done!

tools/serverpod_cli/lib/src/generator/types_record.dart
tp: > (And the existance of these could be validated using the start and end variables.)

That's the thing, if the `{}` do not exist, the substring parsing would fail.

So what we could do then is `try/catch` everything, and return `null` for the entire record (I would not silently skip over any sub-parts which turn out not to be valid record (parts)).

tools/serverpod_cli/lib/src/generator/types.dart
```diff
@@ -764,3 +787,164 @@ enum DefaultValueAllowedType {
...
+    }
+
+    var positionalFieldTypes = _getTypesFromPositionalFields(
+      recordParts.positionalFieldStrings,
+    );
+
+    var namedFields = _tryParseNamedFieldsString(recordParts.namedFieldsString);
+    if (namedFields == null) {
+      return null;
+    }
```
SandPod, line 858: Suggestion(NIT): Move this up before `_getTypesFromPositionalFields` to minimize processing before the early return.

SandPod (approved):
Nice work!

Thanks for taking the time to clarify how the record is parsed in `types.dart`.

tools/serverpod_cli/lib/src/generator/types.dart
tp: Done. While performance wise this could save some computation, `_tryParseNamedFieldsString` really only would fail if any named field did not have a name, which I would expect to be very rare for invalid input (most other cases we'd have already caught above).

I think a better reason for the structure before was that it matches the record declaration (positional fields before named fields).

But not feeling strongly about this either way, as its just 3 lines apart, so made the change :)

PR #3240, opened 2025-02-18
Split `TypeDefinition` into a union with specific types for "class" and "enum"

tp:
Paving the path forward to add "record" types next (which have yet another shape)

For the primary use-cases of parsing and analyzing endpoints and model files, use parallel unions which hold onto the underlying Dart types and consumer-defined model definitions respectively.

Relates to #1784


tools/serverpod_cli/lib/src/generator/types.dart
```diff
@@ -17,39 +16,100 @@ const _moduleRef = 'module:';
 const _projectRef = 'project:';
 const _packageRef = 'package:';
 
+/// Base class for type definitions
+///
+/// One of
+/// - [EnumTypeDefinition]
+/// - [ClassTypeDefinition]
```
tp, line 23: Other than discussed not using a "primitive" type for now, as I can not draw a clear boundary.

Today those might be `int`/`String`, but also `DateTime` and `UuidValue`. And they might possibly also have generics in the future as well.

So just splitting enum and class, where we can then slot in the record support next.

What goes where on each interface is still TBD as I work on getting the code to type-check and run again.

As shown below I would add the same split below for the wrapper classes that handle `TypeDefinition`s, so that we can easily switch over the container and be sure to have the right `type` inside (as these wrappers additionally contain further fields, which must also match the type, see `ClassDefinition` and `EnumDefinition` from the model parsing for example, which should be backed by a `ClassTypeDefinition` and `EnumTypeDefinition` respectively).

tp:
Discarding this effort for now, as the split between `class` and `enum` does not make as much sense as I initially thought. The only difference between the 2 is really the extra "cases" field from the enum definition, all other properties are shared after all (name, generics, etc.).

So instead we'll just split the `TypeDefinition` into `Class*` and `Record*`, much like the distinct type (and element) classes from the Dart analyzer.

PR #3243, opened 2025-02-19
Fix serverOnly + one-to-one id relation regression on main

### LINKED ISSUE #3225: Serverpod CLI generate command from master throws errors when using serverOnly + one to one relation to id
### Describe the bug

After switching to cli from main branch my project no longer generates  (works fine on 2.3.1)

```
Error on line 5, column 32 of lib/src/features/comment/data/comment_data.spy.yaml: The relation with scope "serverOnly" requires the relation to be optional.
  ╷
1 │ parent=submission
  │ ^^^^^^^^^^^^^^^^^
  ╵
```

My model is a table server only with a simple one to one id relation.

`submissionId: int?, relation(parent=submission)`

Putting an issue to prevent regression

### To reproduce

Setup simple project 
add 2 simple table models
add to one of them one to one to id relation to the other, like in above description

### Expected behavior

model should generate

### Library version

Master ref: f852fd7a0637c3541cc27b94285a7b0ac4c13319
Project created using 2.3.1

### Platform information

macos

### Additional context

looks like adding "customField" - https://docs.serverpod.dev/concepts/database/relations/one-to-one#custom-foreign-key-field
workarounds this issue but it forces us to use it every time

- [ ] I want to work on this

nielsenko:
Fix #3225 

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tools/serverpod_cli/lib/src/analyzer/models/checker/analyze_checker.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/checker/analyze_checker.dart
@@ -3,12 +3,6 @@ import 'package:serverpod_cli/src/analyzer/models/validation/keywords.dart';
 import 'package:yaml/yaml.dart';
 
 class AnalyzeChecker {
-  static bool isIdType(dynamic type) {
-    if (type is! String) return false;
-
-    return type == 'int' || type == 'int?';
-  }
-
   static bool isParentDefined(dynamic node) {
     if (node is! YamlMap) return false;
     return node.containsKey(Keyword.parent);
--- a/tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
@@ -359,7 +359,7 @@ class ResolvedInheritanceDefinition extends InheritanceDefinition {
   ResolvedInheritanceDefinition(this.classDefinition);
 }
 
-abstract class RelationDefinition {
+sealed class RelationDefinition {
   String? name;
 
   bool isForeignKeyOrigin;
--- a/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
@@ -731,35 +731,38 @@ class Restrictions {
     dynamic content,
     SourceSpan? span,
   ) {
-    var errors = <SourceSpanSeverityException>[];
-    var definition = documentDefinition;
+    var classDefinition = documentDefinition;
+    if (classDefinition is! ClassDefinition) return const [];
 
-    if (definition is! ClassDefinition) return errors;
+    var field = classDefinition.findField(parentNodeName);
+    if (field == null) return const [];
 
-    var field = definition.findField(parentNodeName);
-    if (field == null) return errors;
-    var type = field.type.className;
-
-    if (AnalyzeChecker.isIdType(type) &&
-        !AnalyzeChecker.isParentDefined(content)) {
-      errors.add(SourceSpanSeverityException(
-        'The "parent" property must be defined on id fields.',
-        span,
-      ));
+    var type = field.type;
+    if (type.isIdType && !AnalyzeChecker.isParentDefined(content)) {
+      return [
+        SourceSpanSeverityException(
+          'The "parent" property must be defined on id fields.',
+          span,
+        )
+      ];
     }
 
-    if (!AnalyzeChecker.isFieldDefined(content)) {
-      var isOptional = AnalyzeChecker.isOptionalDefined(content);
-      var isServerOnly = field.scope == ModelFieldScopeDefinition.serverOnly;
-      if (isServerOnly && !isOptional) {
-        errors.add(SourceSpanSeverityException(
+    var relation = field.relation;
+    if (relation is! ObjectRelationDefinition) return const [];
+
+    if (!AnalyzeChecker.isFieldDefined(content) &&
+        !classDefinition.serverOnly &&
+        field.scope == ModelFieldScopeDefinition.serverOnly &&
+        !relation.nullableRelation) {
+      return [
+        SourceSpanSeverityException(
           'The relation with scope "${field.scope.name}" requires the relation to be optional.',
           span,
-        ));
-      }
+        )
+      ];
     }
 
-    return errors;
+    return const [];
   }
 
   List<SourceSpanSeverityException> validateParentName(
--- a/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_one_to_one_test.dart
+++ b/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_one_to_one_test.dart
@@ -1000,4 +1000,75 @@ void main() {
     expect(collector.errors, hasLength(variant.errors.length));
     expect(collector.errors.map((e) => e.message), variant.errors);
   });
+
+  testWithVariants(
+      'Given a class with an object relation,',
+      describeVariant: (v) =>
+          'when the class is${v.classIsServerOnly ? '' : ' not'} serverOnly'
+          ' and the field is${v.optional ? '' : ' not'} optional'
+          '${v.fieldScope == null ? '' : ' and the field scope is ${v.fieldScope!.name}'}'
+          ', then it should cause ${v.errors.length} error(s)',
+      const [
+        (classIsServerOnly: true, fieldScope: null, optional: true, errors: []),
+        (
+          classIsServerOnly: true,
+          fieldScope: null,
+          optional: false,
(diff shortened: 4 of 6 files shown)


nielsenko:

nielsenko:
recheck

tools/serverpod_cli/lib/src/analyzer/models/checker/analyze_checker.dart
```diff
@@ -3,12 +3,6 @@ import 'package:serverpod_cli/src/analyzer/models/validation/keywords.dart';
 import 'package:yaml/yaml.dart';
 
 class AnalyzeChecker {
-  static bool isIdType(dynamic type) {
-    if (type is! String) return false;
-
-    return type == 'int' || type == 'int?';
-  }
-
```
nielsenko, line 11: This was only used in one place. And there it was used like `AnalyzeChecker.isIdType(field.type.className)`. The better way to do that check is simply `field.type.isIdType`. Also, as `className` doesn't actually contain the `?` when called on a `TypeDefinition` of a nullable. I decided this was just better removed altogether.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -359,7 +359,7 @@ class ResolvedInheritanceDefinition extends InheritanceDefinition {
   ResolvedInheritanceDefinition(this.classDefinition);
 }
 
-abstract class RelationDefinition {
+sealed class RelationDefinition {
```
nielsenko, line 362: Can be, should be.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -731,35 +731,38 @@ class Restrictions {
...
+      ];
     }
 
-    if (!AnalyzeChecker.isFieldDefined(content)) {
-      var isOptional = AnalyzeChecker.isOptionalDefined(content);
-      var isServerOnly = field.scope == ModelFieldScopeDefinition.serverOnly;
-      if (isServerOnly && !isOptional) {
-        errors.add(SourceSpanSeverityException(
+    var relation = field.relation;
+    if (relation is! ObjectRelationDefinition) return const [];
```
nielsenko, line 751: This is core of the fix. Id relations are explicit integer fields, and the correctness wrt. `serverOnly` is already handled. We only need to consider object relations.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_one_to_one_test.dart
```diff
@@ -923,4 +924,151 @@ void main() {
...
     }, skip: errors.isEmpty);
   });
+
+  testWithVariants(
+      'Given a class with an id relation,',
+      describeVariant: (v) =>
+          'when the class is${v.classIsServerOnly ? '' : ' not'} serverOnly'
+          ' and the id field is${v.nullableField ? '' : ' not'} nullable'
+          '${v.fieldScope == null ? '' : ' and the field scope is ${v.fieldScope!.name}'}'
+          ', then it should cause ${v.errors.length} error(s)',
```
nielsenko, line 934: We discussed this on slack. It breaks a bit with the current pattern, but I hope you agree to it. 

BTW: Currently we do a lot of stuff in `group`s `body` callback. I'm not a fan of that as that callback is invoked during test setup, not on test run. This slow down the start of individual tests, and makes it harder to use break-points, as they may be hit before the test starts.

tools/serverpod_cli/test/test_variants.dart
```diff
@@ -0,0 +1,32 @@
+import 'package:meta/meta.dart';
```
nielsenko, line 1: Not sure where to put this file, so for now it is at the root of the `serverpod_cli` test directory.

tools/serverpod_cli/test/test_variants.dart
```diff
@@ -0,0 +1,32 @@
...
+import 'package:test/test.dart';
+
+///
```
nielsenko, line 4: Ups, will fix.

tools/serverpod_cli/lib/src/test_util/builders/relation_model_source_builder.dart
```diff
@@ -0,0 +1,56 @@
+import 'package:serverpod_cli/analyzer.dart';
+import 'package:serverpod_cli/src/test_util/builders/model_source_builder.dart';
+import 'package:serverpod_cli/src/util/model_helper.dart';
+
+class RelationModelSourceBuilder {
```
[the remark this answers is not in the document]
SandPod: Building these objects can be a bit tricky the first time, but once learned, we get clear documentation around what cases we have under test for the analyzer (You can simply search for some part of a YAML file and find tests that cover it).

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_one_to_one_test.dart
```diff
@@ -930,115 +929,126 @@ void main() {
...
+            ]);
+          });
+        });
+      });
+    });
   });
 }
+
+extension on List<ModelSource> {
+  void dumpOnFailure() {
```
SandPod: Question: This doesn't seem to be used (in the commit that introduces it), can it be removed?

tools/serverpod_cli/analysis_options.yaml
```diff
@@ -12,7 +12,6 @@ linter:
   rules:
     # Temporary disable the following so that each can be fixed one by one.
     avoid_escaping_inner_quotes: false
-    directives_ordering: false
```
SandPod, line 15: Comment(For future reference): Since this linter fix became quite a big change it might have been best introduced in a separate PR :)

SandPod (changes requested):
A couple of more thing to align on regarding tests which is natural when we start working together.

We probably don't need most of the fixes in the last commit either, maybe it is best to revert it?

tools/serverpod_cli/analysis_options.yaml
nielsenko: I will move the last two commits to a separate PR

tools/serverpod_cli/lib/src/test_util/builders/relation_model_source_builder.dart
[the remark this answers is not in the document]
nielsenko: Done, if a bit grudgingly 😄

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_one_to_one_test.dart
```diff
@@ -923,4 +923,362 @@ void main() {
...
+    var analyzer = StatefulAnalyzer(
+      config,
+      models,
+      onErrorsCollector(collector),
+    );
+    analyzer.validateAll();
+    var errors = collector.errors;
+    expect(errors, isEmpty);
+  });
+  test(
```
SandPod, line 956: Suggestion(NIT): add a line break between the tests.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_one_to_one_test.dart
```diff
@@ -923,4 +923,362 @@ void main() {
...
+      (ModelSourceBuilder().withFileName('parent').withYaml('''
+class: Parent
+table: parent
+''').build()),
+      ModelSourceBuilder().withFileName('child').withYaml('''
+class: Child
+table: child
+serverOnly: true
+fields:
+  parent: Parent?, relation(optional)
```
SandPod, line 943: Suggestion(NIT): Indent so that it reads like the previous tests.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_one_to_one_test.dart
```diff
@@ -923,4 +923,362 @@ void main() {
...
+''').build()
```
SandPod, line 944: Suggestion(NIT): Add a comma after each build for better formatting.

SandPod (approved):
Changes look good now! Please fix the last NITS and we will get this merged.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_one_to_one_test.dart
SandPod: Yes, that is correct. But the analyzer is built to remove leading whitespace in files. Therefore it is possible to format these prints to nicely read together with the tests.

Issue #1146, opened 2023-07-12
Improved lints in all packages


vlidholt:
It would be nice to improve our workflow with a stricter set of linting rules. We are currently using the Dart recommended rules plus:
- avoid_print
- prefer_const_constructors
- prefer_const_constructors_in_immutables
- prefer_const_declarations
- prefer_const_literals_to_create_immutables
- always_declare_return_types
- prefer_single_quotes
- unawaited_futures
- unnecessary_final

I suggest we also add:
- directives_ordering (orders imports)
- always_use_package_imports (makes relative imports invalid)
- sort_pub_dependencies (makes sure imports are in alphabetical order in pubspec files)

Potentially also:
- require_trailing_commas (how do you feel about this one?)

Any thoughts/additions before I do a massive PR. 😅

SandPod:
This looks good to me!
I'm for adding the trailing comma requirement as well since I easily forget it.

SandPod:
Stumbled upon this today: According to effective dart relative imports are preferred https://dart.dev/effective-dart/usage#prefer-relative-import-paths 🤷 

Don't know if we should take that into consideration.

christerswahn:
We should also apply a fairly strict linting rule set in our serverpod test project to catch this sort of thing directly.

SandPod:
Tracking fixing the exceptions of the rules in #3251

PR #3245, opened 2025-02-20
chore: Update Serverpod shared lints.

SandPod:
Relates to: #1146 

This PR simply configures our wanted linter rules but does not fix any linter issues.

The fixes for the linter issues will be fixed in subsequent PRs.

## Pre-launch Checklist

## Breaking changes

Maybe - These are simple lint changes but this might require action from downstream users.

### DIFF
--- a/packages/serverpod_lints/lib/internal.yaml
+++ b/packages/serverpod_lints/lib/internal.yaml
@@ -2,13 +2,13 @@ include: package:lints/recommended.yaml
 
 linter:
   rules:
-    - avoid_print
-    - prefer_const_constructors
-    - prefer_const_constructors_in_immutables
-    - prefer_const_declarations
-    - prefer_const_literals_to_create_immutables
-    - always_declare_return_types
-    - prefer_single_quotes
-    - unawaited_futures
-    - unnecessary_final
-    - avoid_dynamic_calls
+    avoid_print: true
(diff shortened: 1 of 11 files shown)


packages/serverpod_lints/lib/internal.yaml
```diff
@@ -2,13 +2,13 @@ include: package:lints/recommended.yaml
...
+    avoid_print: true
+    prefer_const_constructors: true
+    prefer_const_constructors_in_immutables: true
+    prefer_const_declarations: true
+    prefer_const_literals_to_create_immutables: true
+    always_declare_return_types: true
+    prefer_single_quotes: true
+    unawaited_futures: true
+    unnecessary_final: true
+    avoid_dynamic_calls: true
```
christerswahn: Perhaps we should consider including these analyzer rules, preventing "indirect" use of dynamic:

```yaml
analyzer:
  language:
    strict-casts: true
    strict-inference: true
    strict-raw-types: true
```

See https://dart.dev/tools/analysis#enabling-additional-type-checks

SandPod: Aaah yes, good catch! I think this makes sense as well. I'm all for adding that as well 👍

vlidholt (approved):
LGTM, great to get this done!

PR #3247, opened 2025-02-20
Rename `TypeDefinition` to `ClassTypeDefinition`

tp:
in preparation of the upcoming `RecordTypeDefinition`

Also rename the associated builder to `ClassTypeDefinitionBuilder`

Relates to #1784


nielsenko:
I'm curious and probably just out of the loop, but why do we need a separate `RecordTypeDefinintion`, and hence this renaming. And assuming we do, shouldn't we still have a base `TypeDefinition` class, that `ClassTypeDefinition` and `RecordTypeDefinition` extends?

tp:
> I'm curious and probably just out of the loop, but why do we need a separate RecordTypeDefinintion, and hence this renaming. 

Because records are so different, that otherwise we'd have to support a ton of flexibility on the current `TypeDefinition` to support both classes and records.

This also matches the type-split in the Dart analyzer.

For example records don't have a `className` (or any name) and neither `generics`. But we have to have the structure (positional and named parameters) in order to compare them (where for classes we would just look at the import + name).

> And assuming we do, shouldn't we still have a base TypeDefinition class, that ClassTypeDefinition and RecordTypeDefinition extends?

The "root" `TypeDefinition` actually came in commit #2 in this PR.

Overall this PR is not intended to support records yet, but rather just get the basic plumbing out of the way, such that adding actual record support in #3218 is much more focused on that addition and does not include a ton of renames etc.

tp:
@nielsenko We've decided to not pursue this split, and rather extend the current `TypeDefinition` to support `Record`s as well.

PR #3248, opened 2025-02-20
feat: Interface Model Support

### LINKED ISSUE #2842: "Interface" model support.
As an alternative to inheritance, there might be fields that I would like to share between different unrelated models.

A use case for this could be audit fields that standardizes the format of when a record was created or updated.

By supporting interfaces, we would enable multiple definitions of standardized fields that could be added to a model.

Add a new model type called `interface` that can be `implemented` by other models.

The definition for an `Interface` could look like this:

```yaml
interface: AuditLog
fields:
  createdAt: Datetime, default=now
  updatedAt: Datetime, default=now
```

And then be used in `Class` model like so:

```yaml
class: MyUser
implements: AuditLog
fields:
    name: String
    age: Int
```

The fields `createdAt` and `updatedAt` are then accessible for the user.

**Additional context**
- It should be possible to implement multiple interfaces.
- SerializableExceptions should also have support for implementing interfaces.

BenAuerDev:
This PR tries to enable the use of the `interface` and `implements` keyword in model files.

This is done in 3 steps/commits 

1. Enable `interface` and `implements` keywords
2. Enable generating interfaces and classes implementing them
3. Exclude interfaces from serialization in `protocol`

I described every change ordered by file in the commit descriptions. 

This PR tries to close #2842. 

## Pre-launch Checklist

## Breaking changes

No breaking changes

### DIFF
--- a/tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
@@ -56,6 +56,9 @@ class ClassDefinition extends SerializableModelDefinition {
   /// `true` if this is an exception and not a class.
   final bool isException;
 
+  /// `true` if this is an interface and not a class.
+  final bool isInterface;
+
   /// If set to true the class is sealed.
   final bool isSealed;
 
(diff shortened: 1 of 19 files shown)


SandPod:
Closing as #3292 replaces it.

PR #3256, opened 2025-02-23
feat: Handle SIGTERM and shutdown gracefully

### LINKED ISSUE #2147: Handle SIGTERM and shutdown gracefully
Serverpod ignores the SIGTERM command and does nothing.

In cloud run as an example the SIGTERM is sent ~10 seconds before the container will be force killed leaving room for a graceful shutdown. 

But since we ignore the signal the server may be executing commands with the expectation that it can finish them. Instead we should clean up all resources when the signal is received.

christerswahn:
Orderly shutdown of Serverpod when SIGTERM or SIGINT (ctrl-c in the terminal) are received.

If ctrl-c is pressed a second time before shutdown is complete then the process exits immediately.

(On Windows SIGTERM is not supported.)

Closes #2147 

## Pre-launch Checklist

## Breaking changes

n/a

### DIFF
--- a/packages/serverpod/lib/src/server/health_check_manager.dart
+++ b/packages/serverpod/lib/src/server/health_check_manager.dart
@@ -23,6 +23,7 @@ class HealthCheckManager {
 
   bool _running = false;
   Timer? _timer;
+  Completer<void>? _completer;
 
   /// Creates a new [HealthCheckManager].
   HealthCheckManager(this._pod, this.onCompleted);
@@ -44,9 +45,31 @@ class HealthCheckManager {
   Future<void> stop() async {
     _running = false;
     _timer?.cancel();
+    await _completer?.future;
   }
 
-  void _performHealthCheck() async {
+  Future<void> _performHealthCheck() async {
+    final completer = Completer<void>();
+    _completer = completer;
+
+    try {
+      await _innerPerformHealthCheck();
+      completer.complete();
+    } catch (e, stackTrace) {
+      _pod.internalSubmitEvent(
+        ExceptionEvent(e, stackTrace, message: 'Error in health check'),
+        space: OriginSpace.framework,
+        context: DiagnosticEventContext(
+          serverId: _pod.serverId,
+          serverRunMode: _pod.commandLineArgs.role.name,
+          serverName: '',
+        ),
+      );
+      completer.completeError(e, stackTrace);
+    }
+  }
+
+  Future<void> _innerPerformHealthCheck() async {
     if (_pod.commandLineArgs.role == ServerpodRole.maintenance) {
       stdout.writeln('Performing health checks.');
     }
--- a/packages/serverpod/lib/src/server/health_check_manager.dart
+++ b/packages/serverpod/lib/src/server/health_check_manager.dart
@@ -23,7 +23,7 @@ class HealthCheckManager {
 
   bool _running = false;
   Timer? _timer;
-  Completer<void>? _completer;
+  Completer<void>? _pendingHealthCheck;
 
   /// Creates a new [HealthCheckManager].
   HealthCheckManager(this._pod, this.onCompleted);
@@ -45,12 +45,12 @@ class HealthCheckManager {
   Future<void> stop() async {
     _running = false;
     _timer?.cancel();
-    await _completer?.future;
+    await _pendingHealthCheck?.future;
   }
 
   Future<void> _performHealthCheck() async {
     final completer = Completer<void>();
-    _completer = completer;
+    _pendingHealthCheck = completer;
 
     try {
       await _innerPerformHealthCheck();
--- a/packages/serverpod/lib/src/server/health_check_manager.dart
+++ b/packages/serverpod/lib/src/server/health_check_manager.dart
@@ -41,7 +41,7 @@ class HealthCheckManager {
   }
 
   /// Stops the health check manager.
-  void stop() {
+  Future<void> stop() async {
     _running = false;
     _timer?.cancel();
   }
--- a/packages/serverpod/lib/src/server/serverpod.dart
+++ b/packages/serverpod/lib/src/server/serverpod.dart
@@ -543,7 +543,9 @@ class Serverpod {
       var serversStarted = true;
 
       ProcessSignal.sigint.watch().listen(_onShutdownSignal);
-      ProcessSignal.sigterm.watch().listen(_onShutdownSignal);
+      if (!Platform.isWindows) {
+        ProcessSignal.sigterm.watch().listen(_onShutdownSignal);
+      }
 
       // Serverpod Insights.
       if (Features.enableInsights) {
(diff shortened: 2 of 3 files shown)


packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -842,24 +842,52 @@ class Serverpod {
...
       await _databasePoolManager?.stop();
     } catch (e, stackTrace) {
-      _reportException(e, stackTrace, message: 'Error in Serverpod shutdown');
-      rethrow;
+      shutdownError = e;
+      _reportException(e, stackTrace,
+          message: 'Error in database pool manager shutdown');
     }
 
+    stdout.writeln('SERVERPOD has shutdown, time: ${DateTime.now().toUtc()}');
```
SandPod, line 898: Suggestion(NIT): Rephrase to `SERVERPOD shutdown completed, time:...`

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -842,24 +842,52 @@ class Serverpod {
...
+    ].nonNulls;
+
+    Object? shutdownError;
+    await futures.wait.onError((ParallelWaitError e, stackTrace) {
+      shutdownError = e;
+      var errors = e.errors;
+      if (errors is Iterable<AsyncError?>) {
+        for (var error in errors.nonNulls) {
+          _reportException(error.error, error.stackTrace,
+              message: 'Error in server shutdown');
```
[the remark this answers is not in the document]
SandPod: We could maybe bump our formatter (in a separate commit) to use Dart 3.7 to get this formatting out of the box.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -542,6 +542,9 @@ class Serverpod {
         commandLineArgs.role == ServerpodRole.serverless) {
       var serversStarted = true;
 
+      ProcessSignal.sigint.watch().listen(_onShutdownSignal);
```
SandPod, line 545: Question: Isn't the contract for `SIGINT` as opposed to `SIGTERM` that we are not expecting a graceful shutdown?  Do we have any reference documentation for treating `SIGINT` as `SIGTERM`?

My concern is that `SIGINT` might be used by the system to instantly shut down the process, such as a user pressing `ctrl+c`.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -710,6 +713,12 @@ class Serverpod {
     }
   }
 
+  void _onShutdownSignal(ProcessSignal signal) {
+    stdout.writeln('${signal.name} (${signal.signalNumber}) received'
+        ', time: ${DateTime.now().toUtc()}');
+    shutdown(signalNumber: signal.signalNumber);
```
SandPod, line 721: Suggestion: Add `exitProcess: true` as an argument here.

If we ever change the default behavior of that method, this is the behavior we would like to preserve.

tests/serverpod_test_server/test_integration/shutdown_test.dart
```diff
@@ -0,0 +1,237 @@
+// ignore_for_file: dead_code
```
SandPod, line 1: Question: Why is this needed? Can we remove it?

tests/serverpod_test_server/test_integration/shutdown_test.dart
```diff
@@ -0,0 +1,237 @@
...
+        ['bin/main.dart', '--mode=test'],
+        verbose: verbose,
+      );
+
+      await expectLater(
+        processOutput.outQueue,
+        emitsThrough(contains('SERVERPOD initialized')),
+      );
+
+      await Future.delayed(signalDelay);
```
SandPod, line 67: Question: Is it possible to remove this delay and simply await some output from the server that informs us that it is ready?

Motivation is to remove unnecessary time delays in tests.

tests/serverpod_test_server/test_integration/shutdown_test.dart
SandPod: Is this connected to the "verbose" prints?

tests/serverpod_test_server/test_integration/shutdown_test.dart
```diff
@@ -0,0 +1,237 @@
...
+      print('response received with code ${response.statusCode}');
+      expect(response.statusCode, 200);
+
+      processOutput.outQueue.rest.listen((s) {});
+      processOutput.errQueue.rest.listen((s) {});
+      var exitCode = await processOutput.process.exitCode.timeout(
+        terminationTimeout,
+      );
+      expect(exitCode, 143);
+    }, skip: 'Dart HTTP server does not support this graceful shutdown');
```
SandPod, line 232: Question: This test appears to have been skipped. Is this intentional? Why in that case?

packages/serverpod/lib/src/server/serverpod.dart
SandPod: It feels like I have encountered that a double `ctrl+c` exhibits instant shutdown.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -975,6 +976,28 @@ class Serverpod {
   }
 }
 
+Future<void>? _shutdownTestAuditor() {
```
SandPod, line 981: Discussion: I'm not a big fan of this. We introduce behavior in production that is related to us creating tests for the system.

 If I understand this correctly, this gives us the ability to "throw" during shutdown? Let's discuss if there are any options to adding this to the server.

SandPod (commented):
Nice clean work!

Just have a couple of questions regarding best practices when treating SIGINT, also would like to explore if we can achieve the test auditor test without adding reading a hidden `env` to production code. It feels like that is the same thing as adding test methods which we generally stay away from.

tests/serverpod_test_server/test_integration/shutdown_test.dart
christerswahn: I tried to do this, but it was not possible, at least not with the stdout/stderr events. We don't print a specific log message for "services up and ready to receive calls". We could poll the http port every 0.1 secs for example but this was simpler.

packages/serverpod/lib/src/server/health_check_manager.dart
```diff
@@ -23,6 +23,7 @@ class HealthCheckManager {
 
   bool _running = false;
   Timer? _timer;
+  Completer<void>? _completer;
```
SandPod, line 26: Suggestion(NITTY NITT NITT): Rename to `_pendingHealthCheck` as `_completer` is simply the type but it doesn't give us any information on what we are waiting on in `stop()`.

SandPod (approved):
LGTM! One minor NIT suggestion.

Issue #3255, opened 2025-02-23
feat: Option to use UUIDs as primary keys


marcelomendoncasoares:
### Problem to Solve

This issue proposes a comprehensive feature of how to allow the usage of UUID as primary keys instead of int. The feature is already discussed and implemented in the draft PR #3152. However, as the PR touches large parts of the serverpod code, we will break it into several small PRs that will slowly incorporating the feature into the framework. All such PRs will reference this issue.

For reference, the feature is a recurrent request that was previously discussed on the issues #351, #635 and #2641. It is also a pre-requisite for the issue #649.

### Proposal

The proposed implementation should:

- Allow the user to declare the `id` field type in serverpod models, changing its type to one of the supported id types (with the correct default attribution). The `int` type should not need default values, although it can be optionally specified. Any relation to the model will use this id as FK if needed.
- Expose an alternative to change the default id type for the project. This will make all tables be generated with the chosen option instead of the default `int` type. The option will be persisted in the `generator.yaml` config file for the language server and subsequent calls to `serverpod generate` be able to interpret the project correctly.
- Add a new default option for `UuidValue` fields to create UUIDs randomly according to the v7 spec, that solves most performance issues of using UUIDs as PK (lack of PK sorting, increased storage, index bloat, etc.) by generating lexicographically sortable ids. This option should be supported as a valid id type.

### Use Case

Although the common best practice is to use sequential as PK, there are some use cases where using UUIDs is preferred. In distributed systems, for example, the main advantage is the possibility to create several related entities on the client and feed them to the server without round-trips to get ID's. On offline-first applications, in which local and remote database must be a copy of each other, using UUIDs as PK also simplifies upserts without dealing with unique indexes violations (specially for related entities).

### Alternatives

In the referenced discussions, it is suggested to create unique indexes on UUID fields instead of using it as PK, but this only ensures uniqueness through constraints, while relationships between entities remain on sequential fields.

### Additional context

Additional context regarding the proposal can be found in the discussion of the PR #3152.

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [x] I want to work on this

SandPod:
Fixed by: https://github.com/serverpod/serverpod/pull/3299

marcelomendoncasoares:
From the complete spec on the proposal, we still have two features pending from this issue:
- Expose an alternative to change the default id type for the project.
[...]

PR #3257, opened 2025-02-24
refactor: Enable table primary keys to have other types than int

marcelomendoncasoares:
This PR enables the feature #3255 implementation by removing the core premise of the serverpod package that table ids are exclusively of int type. To do this, the following changes were made:

- Base column types were rearranged and there is a new public column type `ComparableColumn<T>`. This is the type of valid id columns and implements equality and inequality comparison. All types that have [such comparison options on postgres](https://www.postgresql.org/docs/current/functions-comparison.html#FUNCTIONS-COMPARISON-OP-TABLE) are now `ComparableColunn`. With this, the string and uuid types can now be compared using less than/more than (and the equal to variant) operators, as supported by Postgres.
- The `Table` and `TableRow` concept classes are now generic to the type of the ID attribute, with the `TableRow<T_ID>.id` being of type `ComparableColunn<T_ID>`. On the `Database` level and below, references to the id column were changed to `Object` instead of `int` and all server models were re-generated.
- The `ObjectRelationDefinition` and `ListRelationDefinition` objects now store the id type of the table that owns the FK field. This reference is used to pass the `T_ID` generic reference to build the attach/detach repository methods on the server models.

The PR contains tests of the changes, with an exception of one test that will be added on a subsequent PR of the feature #3255: the test that the stored FK field type on the relation objects is being collected from the correct model. The test exists on the #3152 draft PR, but it requires that more than one type of id is already implemented. Otherwise, it is not be possible to assert that the collected `int` type comes from one or other model of the relation.

## Pre-launch Checklist

## Breaking changes

After the PR, users will have to run `serverpod generate` again to update the models. However, it is already part of the update routine and not considered a breaking change.


SandPod:
Awesome work! Will take a look at this first thing tomorrow morning! 👍

packages/serverpod/lib/src/database/sql_query_builder.dart
```diff
@@ -238,7 +238,8 @@ class SelectQueryBuilder {
 
     var relationFieldName = tableRelation.foreignFieldBaseQuery;
 
-    var whereAddition = Expression('$relationFieldName IN (${ids.join(', ')})');
+    var strIds = [for (var id in ids) id is num? ? '$id' : "'$id'"].join(', ');
+    var whereAddition = Expression('$relationFieldName IN ($strIds)');
```
SandPod, line 242: Question: Would it be possible to connect this implementation detail to the `IdColumnIterable` in `packages/serverpod/lib/src/database/concepts/columns.dart`?

It feels like this might be easily forgotten if additional id fields should be supported.

tests/serverpod_test_server/test_integration/column_operations/column_string_test.dart
```diff
@@ -150,5 +150,51 @@ void main() async {
 
       expect(result, hasLength(2));
     });
+
+    test('when filtering using greater than then matching rows are returned.',
```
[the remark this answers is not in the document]
marcelomendoncasoares: Agreed, the existing values can be misleading to future developers. But maybe a better test description is enough? Just made a commit renaming the tests for both String and UUID types to describe the lexicographical behavior. Do you think this addresses the problem?

packages/serverpod/lib/src/database/sql_query_builder.dart
marcelomendoncasoares: Good idea! Done.

tests/serverpod_test_server/test_integration/column_operations/column_string_test.dart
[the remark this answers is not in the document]
SandPod: Yes, that should be enough! 👍

packages/serverpod/lib/src/database/concepts/columns.dart
```diff
@@ -72,9 +72,32 @@ abstract class _ValueOperatorColumn<T> extends Column<T> {
   Expression _encodeValueForQuery(T value);
 }
 
+/// A [Column] whose values can be compared equal to other values.
+abstract class ColumnComparableEquals<T> extends _ValueOperatorColumn<T>
```
SandPod, line 76: Suggestion(NIT): Make this class private.

Doesn't seem like this is used outside of this file currently.

packages/serverpod/lib/src/database/concepts/columns.dart
```diff
@@ -733,3 +752,23 @@ class _NotInSetExpression extends _SetColumnExpression {
...
+
+    throw Exception('Unsupported id column type: ${first.runtimeType}');
+  }
+
+  /// Converts each id element to a string that can be used in a SQL query.
+  Iterable<String> castIdsToSqlString() => map(castIdToSqlString);
+}
+
+/// Converts an id to a string that can be used in a SQL query.
+String castIdToSqlString(dynamic id) => id is num? ? '$id' : "'$id'";
```
SandPod, line 774: Suggestion/Discussion: This made it clear that we are adding more SQL query building to our columns context.

Ideally we would like to keep this clean from the query building so that it could be adopted for different query generators. (Breaking out the strings from here would be a required step for that, but at least the tree like structure allows us to analyze it dissegarding the strings)

How about we instead make `_extractPrimaryKeyForRelation` in `database_connection.dart` return both the column type and the value in some struct, that should give us the knowledge required to convert to valid sql syntax regarless of what the type is?

That would give us flexibility to easily support more columns in relations without necesseraly having them as primary key?

SandPod (commented):
Changes look good now! Have a last possible change to ensure that we don't add more coupling between our columns class and the sql query builder.

We want to ensure we don't forget about the type, but we don't necessarily want to build (more) sql strings outside of the sql query generator.

packages/serverpod/lib/src/database/concepts/columns.dart
SandPod: Since this is an isolated thing I think we could also break this out into a separate PR that follows directly after this.

Then we could tackle this in isolation, as long as we fix it.

packages/serverpod/lib/src/database/concepts/columns.dart
marcelomendoncasoares: You're right, it's only used inside this file. I had only left this as public to be consistent with the `ColumnComparable`, in case someone would want to reference all columns that implement `equals` (includes enum and bool beside all `ColumnComparable`). But I do not have an use case, so no problem making it private.

packages/serverpod/lib/src/database/concepts/columns.dart
marcelomendoncasoares: I guess I've found a way to remove this string conversion. The `DatabasePoolManager.encoder.convert` already contains the logic to safely convert any dart type to the appropriate SQL string, completely removing the need for the two added methods. Do you think that this solves the problem?

SandPod: Aah even better, that should probably have been the solution to start with! 👍

SandPod (approved):
Changes look good to me! Nice diligent work! 🚀

Issue #3258, opened 2025-02-24
Flaky integration test


nielsenko:
A test [run](https://github.com/serverpod/serverpod/actions/runs/13493625408/attempts/1) surfaced a flaky integration test.

_Given that withServerpod can not find the database and has a timeout set to 0 seconds
when running the test
then should timeout immediately_

PR #3259, opened 2025-02-24
fix: Validate timeout, but don't bound how long.

nielsenko:
If timeout exceeded, process should exit 1, and stdout indicate the reason, but don't mandate how long it took.

Mandating how long it took is flaky, and can cause false positives.

fix: #3258 

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tests/serverpod_test_server/test_integration/test_tools/start_timeout_test/start_timeout_test.dart
+++ b/tests/serverpod_test_server/test_integration/test_tools/start_timeout_test/start_timeout_test.dart
@@ -9,17 +9,12 @@ void main() async {
       'Given that withServerpod can not find the database and has a timeout set to 0 seconds '
       'when running the test '
       'then should timeout immediately', () async {
-    var timer = Stopwatch()..start();
     final result = await runTest('test_that_will_timeout.dart');
 
     expect(result.exitCode, 1);
     expect(
-        result.stdout,
-        contains(
-            'Serverpod did not start within the timeout of 0:00:00.000000'));
-    expect(
-      timer.elapsed.inSeconds,
-      lessThan(10), // sanity check
+      result.stdout,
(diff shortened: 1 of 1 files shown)


tests/serverpod_test_server/test_integration/test_tools/start_timeout_test/start_timeout_test.dart
```diff
@@ -37,15 +32,11 @@ void main() async {
           contains(
               'Serverpod did not start within the timeout of 0:00:30.000000'));
       expect(
-          timer.elapsed.inSeconds,
-          allOf(
-            greaterThanOrEqualTo(30),
-            lessThan(40), // sanity check
-          ));
+        timer.elapsed.inSeconds,
+        greaterThanOrEqualTo(30),
```
Isakdl, line 36: Question: Are we actually waiting 30 seconds in this test?

Isakdl (approved):
LGTM! Just got a question on time, it seems a little excessive if we are actually waiting 30 seconds in the test. Especially since I believe we are running these test in concurrent mode. 

But we don't have to fix that in this PR.

tests/serverpod_test_server/test_integration/test_tools/start_timeout_test/start_timeout_test.dart
nielsenko: Yes, when not specifying a timeout, it seems we are hardcoded to use 30s:

The test checks how long it takes until `withServerpod`  fails, when not specifying a timeout, and no database exists to be found.

from `with_serverpod.dart:80`
```dart
var startTimeout = maybeServerpodStartTimeout ?? const Duration(seconds: 30);
```

Smells a bit, but that is what is happening today.

PR #3266, opened 2025-02-25
Support `Stream<T>` endpoints containing records

tp:
For return type and arguments, including records nested inside records

Relates to #1784


packages/serverpod_serialization/lib/src/websocket_messages.dart
```diff
@@ -457,7 +459,20 @@ class MethodStreamMessage extends WebSocketMessage
             data[WebSocketMessageDataKey.connectionId]),
         parameter = data[WebSocketMessageDataKey.parameter],
         object = serializationManager
-            .deserializeByClassName(data[WebSocketMessageDataKey.object]);
+            .deserializeByClassName(data[WebSocketMessageDataKey.object]) {
+    // TODO(tp): Either this or we'd need to keep `serialization manager around`
```
tp, line 463: The old `toString` was calling into the serialization, which now only works if we use the project's one, not the global one.

Do you have a preference here? Either we'd need to build the string upfront (like this), or keep the serialization manager on the class.

tests/serverpod_test_server/test_e2e/record_types_test.dart
```diff
@@ -529,31 +545,115 @@ void main() {
...
+          .returnListOfNestedPositionalAndNamedRecord([record, record]);
+
+      expect(result, hasLength(2));
+      expect(
+        result.first.$1,
+        (1, '2'),
+      );
+      expect(
+        result.first.namedSubRecord.$1,
+        isA<SimpleData>().having((d) => d.num, 'num', 3),
```
tp, line 570: All these assertions are a bit complicated (using `isA`), because models do not implement equality.

So the only way to test combinations with data classes is to unwrap them manually (instead of being able to assert something like `expect(response, emitsInOrder(requestParameters))`…)

packages/serverpod_serialization/lib/src/websocket_messages.dart
SandPod: Since we will only store a reference to the SerializationManager, and the SerializationManager has the same lifecycle has the Serverpod instance. I think it is fine to store a local reference to it and then use it in the toString method.

tests/serverpod_test_server/lib/src/endpoints/test_tools.dart
```diff
@@ -249,4 +249,58 @@ class AuthenticatedTestToolsEndpoint extends Endpoint {
       yield value;
     }
   }
+
+  Stream<(String, (Map<String, int>, {SimpleData simpleData, bool flag}))>
```
SandPod, line 253: Question(For my knowledge): There are more scenarios we validate in the `test_tools.dart` compared to `record_parameters.dart`. What is the reason for that? Wouldn't these scenarios be interesting in both cases?

(I think you helped me understand this the last time as well, but unfortunately, I can't remember the motivation)

tests/serverpod_test_server/test_integration/test_tools/return_and_input_types_test.dart
```diff
@@ -141,6 +142,218 @@ void main() {
...
+          ('hello', ({'world': 1}, flag: true, simpleData: SimpleData(num: 2))),
+          ('hi', ({'world': 3}, flag: true, simpleData: SimpleData(num: 4))),
+        ];
+
+        var authenticatedSessionBuilder = sessionBuilder.copyWith(
+          authentication: AuthenticationOverride.authenticationInfo(
+            1,
+            {Scope('user')},
+          ),
+        );
```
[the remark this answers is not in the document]
SandPod: Since the tests are quite complex in both their setup and their expects, I reducing the complexity is significant.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -59,10 +59,14 @@ class LibraryGenerator {
...
         .where((t) => t.isStreamType)) {
       var valueType = topLevelType.generics.first;
-      if (valueType.isSetType || valueType.isListType || valueType.isMapType) {
-        if (valueType.dartType == null ||
-            !topLevelStreamContainerTypes.any((type) =>
-                type.dartType.toString() == valueType.dartType.toString())) {
+      if (valueType.isSetType ||
+          valueType.isListType ||
+          valueType.isMapType ||
+          valueType.isRecordType) {
```
SandPod, line 65: Suggestion: Break the group in the if statement out into a private method and give it a suitable name that explains why these specific types get this special treatment.

SandPod: Might also want to include the `any` check there as well to explain the purpose of the nested if statement. Then we can remove it from this outer scope and make this easier to understand.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -332,6 +336,37 @@ class LibraryGenerator {
             ..returns = TypeReference((t) => t..symbol = 'String')
             ..body = literalString(config.name).code,
         ),
+      if (recordTypesToDeserialize.isNotEmpty)
+        Method(
```
SandPod, line 340: Question: Do we have a unit test for the code generator that validates when this override is generated?

When working with the generator, it would be good to get validation that this still works as early as possible.

tools/serverpod_cli/lib/src/generator/dart/library_generators/server_test_tools_generator.dart
```diff
@@ -292,9 +301,32 @@ class ServerTestToolsGenerator {
                   ..body = refer('_localUniqueSession').code).closure,
                 'endpointPath': literalString(endpoint.name),
                 'methodName': literalString(method.name),
-                'arguments': literalMap({
+                'arguments': literalMap(<Expression, Code>{
                   for (var parameter in parameters)
-                    literalString(parameter.name): refer(parameter.name).code,
+                    // Ensure record containing data structures are converted to JSON Map representation all the way down
```
SandPod, line 306: Suggestion: If a comment is needed, please break this out into a private method with a suitable name instead.

Comments have a tendency to grow stale as code evolves, but method are more commonly kept up to date.

tests/serverpod_test_server/test_integration/test_tools/return_and_input_types_test.dart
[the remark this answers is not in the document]
tp: Absolutely right, I should move them to the unauthenticated one. I missed that there are 2 endpoint classes in the file (`TestToolsEndpoint` and `AuthenticatedTestToolsEndpoint`), and just added them at the end, which then ended up being the authenticated ones. Will move this.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
tp: No, there is not a unit test, though I could look into that.

For now the primary motivator for this `if` was to not generate any unused code in projects that don't use it. (Otherwise we'd also see changes (or CI failures) for our example projects.)

tests/serverpod_test_server/test_integration/test_tools/return_and_input_types_test.dart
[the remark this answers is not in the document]
tp: Moved them to the unauthenticated endpoint.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
tp: Added a unit test that the overwrite of `wrapWithClassname` is only generated when using `Stream` + `Record` (directly or nested inside a container).

tools/serverpod_cli/lib/src/generator/dart/library_generators/server_test_tools_generator.dart
tp: Good suggestions. Factored it out and left a detailed comment on the method for what is done to the arguments.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
tp: Factored this out and added explanation of why we pick containers + records (which are container in that they also need custom serialization).

tests/serverpod_test_server/lib/src/endpoints/test_tools.dart
tp: There are more individual test cases in `record_parameters.dart` (32 as of today), because they use much smaller units (e.g. don't combine positional + named + container + model as above).

In the stream tests tools we only have 4 distinct methods for the primary types: a complex record and a complex record in a container plus the nullable variant for each.

We could also explode the number of test cases for the test tools endpoint to cover smaller variants and more container types (eg. Set and List in addition to Map), but I think that wouldn't gain much as the current setup already ensures that the major "branches" are taken.

Since the test tools are build not to verify by itself but rather help people to build a test, it'd also be obvious (in ours and theirs) if for any "real world" usage there would be an issue (that e.g. a `Set<List<Map<int, SimpleData>>>` would not serialize) that we'd need to fix that. Since the number of possible combinations is so large, I would opt to just focus on the major branches here and trust that further nesting works just as well :)

SandPod: Sounds good to me! Then we can build specialized tests if we encounter any deviation later on :)

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -1291,6 +1277,36 @@ extension on ProtocolDefinition {
 
     return false;
   }
+
+  //// Returns all non-model/non-primiate types used with `Streams`
```
SandPod, line 1281: Suggestion(NIT): Typo, change `non-primiate` to `non-primitive`.

SandPod (approved):
LGTM! 🚀

Single NIT fix then we can merge.

PR #3270, opened 2025-02-26
refactor: Separate ClassDefinition into specialization classes.

SandPod:
This PR lays the groundwork for [introducing interfaces](https://github.com/serverpod/serverpod/pull/3248) by defining explicit types for parsed model structures.

Key changes:
- Establishes a clear inheritance structure for parsed definitions:
- Previously, `ModelClassDefinition` and `ExceptionClassDefinition` were both represented as `ClassDefinition`, distinguished only by a boolean flag.
- Now, these types are fully separated in both parsing and code generation, improving type safety and maintainability.

The inheritance structure allows us to target parse rules for specific definitions while sharing logic for all definitions that have fields.

### New inheritance structure

The new inheritance structure looks like this:
```bash
ClassDefinition  
├── ModelClassDefinition  
└── ExceptionClassDefinition  
EnumDefinition  
```

It will now be simple to add an `InterfaceClassDefintion` while preserving all the validation rules that already apply to `ClassDefinition` fields.

#### Open question

There is an argument for renaming `ClassDefinition` to `ModelDefinition` and then having the specializations communicate what type of model definition we have so we would have the following structure:

The new inheritance structure looks like this:
```bash
ModelDefinition  
├── ClassModelDefinition  
└── ExceptionModelDefinition
└── InterfaceModelDefinition  
EnumDefinition  
```

### Guidance for review

Best reviewed commit by commit.

A new test was added at the end for a specific case that I encountered that was not previously covered.

## Pre-launch Checklist

## Breaking changes

None - These changes here are intended to be a strictly internal refactor.


tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -70,12 +70,64 @@ class ModelParser {
...
       subDirParts: protocolSource.subDirPathParts,
       documentation: classDocumentation,
-      isException: documentTypeName == Keyword.exceptionType,
+      isException: false,
       serverOnly: serverOnly,
       type: classType,
     );
   }
 
+  static SerializableModelDefinition? serializeExceptionClassFile(
```
christerswahn, line 78: Question for my understanding: This change kind of looks like we've added exception serialization, but we've just broken it out right? There is significant overlap between these functions. May it make sense to combine the commonalities in shared subfunction(s)?

christerswahn (commented):
Good stuff and easy to grasp overview in the PR description! Just had a question for discussion!

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
SandPod: Yes, that is true! And I think that makes sense. I'll have a look at sharing their logic.

SandPod: Broke out parsing everything that is shared for a `ClassDefintion` in a private shared subfunction with the possibility to add the extra fields that need to be parsed for the specializations.

christerswahn (approved):
LGTM! 🚀

Issue #2762, opened 2024-09-20
Error when running `serverpod generate`


ChrisCRCB:
Hi,
Firstly, thank you so much for making such an awesome framework! I love Serverpod with a passion! It has made many of the projects I work on a breeze.

Now for the issue: I got this while running `serverpod`, with or without the `-w` switch.

```shell
Yikes! It is possible that this error is caused by an internal issue with the Serverpod tooling. We would appreciate if you filed an issue over at Github. Please include the stack trace below and describe     
any steps you did to trigger the error.                                                                                                                                                                          
https://github.com/serverpod/serverpod/issues                                                                                                                                                                    
                                                                                                                                                                                                                 
PathNotFoundException: Directory listing failed, path =                                                                                                                                                          
'C:\Users\chris\src\talking_noticeboard\talking_noticeboard_flutter\linux\flutter\ephemeral\.plugin_symlinks\flutter_soloud\android\.cxx\Debug\3j3u5064\arm64-v8a\CMakeFiles\flutter_soloud_plugin.dir\33da5992842d4b024c1554ca2f93d5a0\src\soloud\src\audiosource\ay\*'                                                                                                                                                          
(OS Error: The system cannot find the path specified.                                                                                                                                                            
, errno = 3)                                                                                                                                                                                                     
                                                                                                                                                                                                                 
#0      _Directory._fillWithDirectoryListing (dart:io-patch/directory_patch.dart:42:24)                                                                                                                          
#1      _Directory.listSync (dart:io/directory_impl.dart:228:5)                                                                                                                                                  
#2      findPubspecsFiles (package:serverpod_cli/src/util/pubspec_helpers.dart:20:24)                                                                                                                            
#3      performServerpodPackagesAndCliVersionCheck (package:serverpod_cli/src/serverpod_packages_version_check/serverpod_packages_version_check.dart:26:22)                                                      
#4      GenerateCommand.run (package:serverpod_cli/src/commands/generate.dart:45:20)                                                                                                                             
<asynchronous suspension>                                                                                                                                                                                        
#5      CommandRunner.runCommand (package:args/command_runner.dart:212:13)                                                                                                                                       
<asynchronous suspension>                                                                                                                                                                                        
#6      BetterCommandRunner.runCommand (package:cli_tools/src/better_command_runner/better_command_runner.dart:140:7)                                                                                            
<asynchronous suspension>                                                                                                                                                                                        
[...]

PR #3273, opened 2025-02-28
fix: Limit how many projects are considered for version check validation on generate

nielsenko:
This PR changes how `serverpod_cli` validates versions during `serverpod generate`.

Previously it would consider all **pubspec.yaml** files below the parent of the generated server project. This could potentially include a lot of projects, and is not needed.

It is important (for now) that the **serverpod** dependencies used between the generated server and client library match the version of `serverpod_cli` exactly. We ensure that by pining the version, and have `serverpod generate` check that the versions matches its own, and are still pinned. This guarantees that the generated code, and protocol matches.

However we don't need to run this check on any other projects. If any other project (fx. the flutter project setup during `serverpod create`) depends on the generated client, then normal package resolution will ensure that the project uses the same version, as it is pinned.

This change came about while investigating #2762 . There is a known [issue](https://github.com/flutter/flutter/issues/106953) with `Directory.listSync` and friends on windows, when traversing directories containing symlinks. 

```sh
PathNotFoundException: Directory listing failed, path =                                                                                                                                                          
'C:\Users\chris\src\talking_noticeboard\talking_noticeboard_flutter\linux\flutter\ephemeral\.plugin_symlinks\flutter_soloud\android\.cxx\Debug\3j3u5064\arm64-v8a\CMakeFiles\flutter_soloud_plugin.dir\33da5992842d4b024c1554ca2f93d5a0\src\soloud\src\audiosource\ay\*'                                                                                                                                                          
(OS Error: The system cannot find the path specified.                                                                                                                                                            
, errno = 3)                                                                                                                                                                                                     
                                                                                                                                                                                                                 
#0      _Directory._fillWithDirectoryListing (dart:io-patch/directory_patch.dart:42:24)                                                                                                                          
#1      _Directory.listSync (dart:io/directory_impl.dart:228:5)                                                                                                                                                  
#2      findPubspecsFiles (package:serverpod_cli/src/util/pubspec_helpers.dart:20:24)    
...
```

Flutter uses symlinks (**..\ephemeral\.plugin_symlinks\..**) in platforms projects, if an app uses native plugins. Since we no longer invoke `Directory.listSync` we implicitly sidestep this issue. 

fix: #2762 

## Pre-launch Checklist

## Breaking changes

None, this will produce less warnings than before.


tools/serverpod_cli/lib/src/util/pubspec_plus.dart
```diff
@@ -0,0 +1,90 @@
...
+import 'package:source_span/source_span.dart';
+import 'package:yaml/yaml.dart';
+
+import 'yaml_util.dart';
+
+/// A pubspec plus the yaml map it was read from, and a list of all augmented
+/// dependencies regardless of kind.
+class PubspecPlus {
+  final Pubspec pubspec;
+  final YamlMap yaml;
```
nielsenko, line 13: Purpose of this class is to keep the `Pubspec` and associated yaml together. Previously a lot of methods took both a `Pubspec` and a `YamlMap` as argument, and it was implicit that the two had to be related.

tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -259,13 +259,10 @@ class GeneratorConfig implements ModelLoadConfig {
...
     var file = File(p.join(serverRootDir, 'config', 'generator.yaml'));
-    Map generatorConfig = {};
-    try {
-      var yamlStr = file.readAsStringSync();
-      generatorConfig = loadYaml(yamlStr);
-    } catch (_) {}
-
-    PackageType type = getPackageType(generatorConfig);
+    YamlMap generatorConfig = await file.exists()
+        ? loadYaml(await file.readAsString(), sourceUrl: file.uri) as YamlMap
```
SandPod: Question: Aren't there other reasons that this might crash than simply handling if the file doesn't exist? For example if the `loadYaml` fails or if the `cast` is invalid?

tools/serverpod_cli/lib/src/serverpod_packages_version_check/serverpod_packages_version_check.dart
```diff
@@ -29,18 +29,12 @@ List<SourceSpanSeverityException> performServerpodPackagesAndCliVersionCheck(
...
-    } catch (e) {
-      // TODO: Log to debug output
-      // Failed to open or parse pubspec file
-    }
+    var file = pubspecFile.readAsStringSync();
+    var pubspec = Pubspec.parse(file);
+    var yaml = loadYaml(file, sourceUrl: pubspecFile.uri) as YamlMap;
+    var warnings = _validateServerpodPackagesVersion(cliVersion, pubspec, yaml);
+
+    accumulatedWarnings.addAll(warnings);
```
SandPod: Question: Not generating the client is a valid option for a Serverpod project. Do we really need to report a warning for that? Maybe we should first filter the list for files that exist?

I think it makes sense to add warnings for files that do exists, but don't have our expected format.

tools/serverpod_cli/lib/src/serverpod_packages_version_check/serverpod_packages_version_check.dart
```diff
@@ -18,102 +19,115 @@ class ServerpodPackagesVersionCheckWarnings {
...
+        yaml: yaml['dev_dependencies'] as YamlMap?
+      ),
+      (
+        deps: pubspec.dependencyOverrides,
+        type: DepKind.override,
+        yaml: yaml['dependency_overrides'] as YamlMap?
+      ),
+    ]) {
+      for (var e in x.deps.entries) {
+        var span = x.yaml!.nodes[e.key]!.span; // ! safe as we loop over entries
```
SandPod: Suggestion: We should always avoid `!` even if we know it's safe.

Please fetch the value using `?` and throw an explicit exception if the value is null.

tools/serverpod_cli/test/integration/serverpod_packages_version_check/serverpod_packages_version_check_test.dart
```diff
@@ -210,35 +187,5 @@ void main() {
...
-          () {
-        var cliVersion = Version(1, 1, 0);
-        var packageWarnings = performServerpodPackagesAndCliVersionCheck(
-          cliVersion,
-          corruptedPubspecPath,
-        );
-
-        expect(packageWarnings.isEmpty, isTrue);
-      });
-    });
```
SandPod, line 226: Question: Why was this test removed? It feels like this is still an important scenario to cover?

tools/serverpod_cli/test/util/pubspec_plus_test.dart
```diff
@@ -0,0 +1,121 @@
...
+    group('when calling PubspecPlus.parse', () {
+      late PubspecPlus pubspecPlus = PubspecPlus.parse(pubspecString);
+      test('then it succeeds', () {
+        expect(
+          () => pubspecPlus,
+          returnsNormally,
+        );
+      });
+
+      test('then it returns PubspecPlus with correct dependencies', () {
```
SandPod, line 32: Suggestion: Please break these out into different tests validating each dependency type.

This test now has multiple independent reasons to fail.

tools/serverpod_cli/test/util/pubspec_plus_test.dart
```diff
@@ -0,0 +1,121 @@
...
+            pubspecPlus.deps.firstWhere((d) => d.name == 'serverpod');
+        expect(serverpodDep.kind, DepKind.normal);
+        expect(serverpodDep, isA<HostedDep>());
+        expect(
+          (serverpodDep as HostedDep).dependency.version,
+          VersionConstraint.parse('2.3.1'),
+        );
+      });
+
+      test('then the span of each dependency is correct', () {
```
SandPod, line 56: Suggestion: Please break these out into multiple tests.

Since we are doing the exact same validation this might be the case where I would be more accepting to introduce a loop of some sort.

But I still think it makes more sense to have these as separate tests.

Might be good to isolate the scenario even. A file with a single dependency as opposed to a file with two?

tools/serverpod_cli/test/integration/serverpod_packages_version_check/serverpod_packages_version_check_test.dart
SandPod: Is this already validated on a Unit test level? If so, we might still want to have at least one test that validates our behavior when the pubspec parsing goes wrong.

SandPod (commented):
Really neat cleanup and improvements of that part of the system!

It really feels like there were a couple of hot patches in that code that could be removed.

I have a couple of questions and comments, main focus is ensuring that we don't introduce unhandled crashes that we did not have before.

tools/serverpod_cli/lib/src/config/config.dart
nielsenko: I want us to fail if we cannot parse the pubspec. Previously we just silently ignored that we had no clue, if the versions where correct.

tools/serverpod_cli/lib/src/serverpod_packages_version_check/serverpod_packages_version_check.dart
nielsenko: What exception do we throw, when it would be an internal bug if a situation ever happens?

tools/serverpod_cli/test/integration/serverpod_packages_version_check/serverpod_packages_version_check_test.dart
nielsenko: Loading the pubspecs are handled higher up. Validation happens on loaded pubspecs.

tools/serverpod_cli/test/util/pubspec_plus_test.dart
nielsenko: Okay 👍

tools/serverpod_cli/test/util/pubspec_plus_test.dart
nielsenko: Will look into it..

tools/serverpod_cli/lib/src/config/config.dart
SandPod: That makes sense! What would the formatted output be to the user? I think we should inform them in a meaningful way that the file is corrupted and that we can't read it.

tools/serverpod_cli/lib/src/serverpod_packages_version_check/serverpod_packages_version_check.dart
SandPod: Sorry, it's not an exception; we should throw an error since it is not actionable from the user. Perhaps a `StateError` with a message explaining the unexpected condition we are in?

tools/serverpod_cli/test/util/pubspec_plus_test.dart
nielsenko: Done

tools/serverpod_cli/lib/src/serverpod_packages_version_check/serverpod_packages_version_check.dart
nielsenko: Okay. Server project must exists, but client will only be checked if it exists. Currently the config will always report a clientPackagePath.

Going forward I think we should consider to explicit record in config if what features are enabled and always have (also in the mini case). I think we have too much semantics riding on missing files currently.

But alas, let us save that for another PR:

nielsenko: We have code in `GeneratorConfig.load` that insists on finding a client project. This is independent of an actual config, as it will just use default values if `generatorConfig['client_package_path'] == null`.

We raise:
```dart
      ...
      throw const ServerpodProjectNotFoundException(
        'Failed to load client pubspec.yaml. If you are using a none default '
        'path it has to be specified in the config/generator.yaml file!',
      );
```
If not found, and terminate generate.

So.. excluding a missing client pubspec from version validaiton will still fail on `serverpod generate`

tools/serverpod_cli/test/integration/serverpod_packages_version_check/test_assets/invalid_pubspec/pubspec.yaml
```diff
@@ -0,0 +1,9 @@
+version: 1.1.0
```
SandPod, line 1: Suggestion(non-blocking): Since we are touching all these assets, perhaps they should be inlined and created in the test instead of storing these on disk?

tools/serverpod_cli/test/integration/serverpod_packages_version_check/load_config_test.dart
```diff
@@ -0,0 +1,55 @@
...
+    await missingPubspecDir.delete();
+  });
+
+  test(
+      'Given an invalid server pubspec.yaml '
+      'then a ServerpodProjectNotFoundException is thrown', () {
+    // TODO(nielsenko):
+    // This is current behavior, but I think it should be
+    // the SourceSpanException raised by Pubspec.parse to give the user
+    // a more detailed error message.
```
SandPod, line 34: Discussion: I agree with this, would be better to give some more details here so that the developer can figure out if we did something wrong or if the file is corrupted.

tools/serverpod_cli/test/integration/serverpod_packages_version_check/load_config_test.dart
```diff
@@ -0,0 +1,55 @@
...
+    expectLater(
+      GeneratorConfig.load(path.join(testAssetsPath, 'invalid_pubspec')),
+      throwsA(isA<ServerpodProjectNotFoundException>()),
+    );
+  });
+
+  test(
+      'Given a valid server pubspec.yaml'
+      ' and a missing client pubspec.yaml '
```
SandPod, line 43: Suggestion(NIT): Style fix.

```suggestion
      'Given a valid server pubspec.yaml '
      'and a missing client pubspec.yaml '
```

tools/serverpod_cli/lib/src/serverpod_packages_version_check/serverpod_packages_version_check.dart
SandPod: And just to double-check, this is what would happen in the latest release if we run Serverpod generate for a full project(config exists) when the client has been removed?

tools/serverpod_cli/test/integration/serverpod_packages_version_check/load_config_test.dart
```diff
@@ -0,0 +1,55 @@
...
+      'then a ServerpodProjectNotFoundException is thrown', () {
+    // TODO(nielsenko):
+    // This is current behavior, but I perhaps it should depend on config
+    // if a client project is generated or not.
+    // (see https://github.com/serverpod/serverpod/pull/3273#discussion_r1982987504)
```
SandPod, line 48: chore: Remove the personal TODO and create an issue that captures the change we would like to introduce here and reference that issue.

That way, we have all the context required to take action and the action is not tied to you personally.

tools/serverpod_cli/test/integration/serverpod_packages_version_check/load_config_test.dart
SandPod: Suggestion: Either fix this or remove the personal TODO and create and issue that capture the action required to resolve this issue and reference that issue here instead.

tools/serverpod_cli/test/integration/serverpod_packages_version_check/serverpod_packages_version_check_test.dart
```diff
@@ -13,230 +14,150 @@ void main() {
...
       });
 
-      test('performServerpodPackagesAndCliVersionCheck() with newer version',
-          () {
+      test(
+          'when calling validateServerpodPackagesVersion with newer version'
+          'then one warning is returned'
+          ' and the message is correct'
+          ' and the span is correct'
+          ' and severity is warning', () {
```
SandPod, line 141: Suggestion: Create a group and assert each `and` in a single test.

There are multiple reasons this single expect might fail. Capture each of the `having` in a single test.

SandPod: Same for the other tests in this file.

SandPod: Giving test descriptions such as `when calling validateServerpodPackagesVersion with newer version then one warning is returned with approximate version warning message.`

tools/serverpod_cli/lib/src/commands/generate.dart
```diff
@@ -44,9 +44,22 @@ class GenerateCommand extends ServerpodCommand {
       throw ExitException(ServerpodCommand.commandInvokedCannotExecute);
     }
 
+    // Directory.current is the server directory
+    var clientPubspecFile = File(path.joinAll([
+      ...config.clientPackagePathParts,
+      'pubspec.yaml',
+    ]));
+    var pubspecsToCheck = [
+      File('pubspec.yaml'), // server
```
SandPod, line 53: Suggestion: create a variable named serverPubspec, connect it to the comment above and remove the inlined comment.

tools/serverpod_cli/lib/src/commands/generate.dart
```diff
@@ -44,9 +44,22 @@ class GenerateCommand extends ServerpodCommand {
...
+      if (await clientPubspecFile.exists()) clientPubspecFile,
```
SandPod, line 54: Question: doesn't this ensure that we will not validate the client pubspec if it does not exist? (Thinking about the comment in the test)

SandPod (commented):
Nice cleanup of the code!

tools/serverpod_cli/test/integration/serverpod_packages_version_check/load_config_test.dart
nielsenko: Okay. It was on purpose though. It makes it easier to see the where the switch from given to then happens.

SandPod: Aah, sorry missed the intentional indentation! This makes sense! 👍 

Ignore my comment.

tools/serverpod_cli/lib/src/commands/generate.dart
nielsenko: Yes. I introduced this behavior because of your feeback. But `GeneratorConfig.load` has its own opinion on this.

tools/serverpod_cli/test/integration/serverpod_packages_version_check/load_config_test.dart
SandPod: But then could simply be on a single line as well right since it is part of the "Given"?(NIT...)

tools/serverpod_cli/test/integration/serverpod_packages_version_check/test_assets/invalid_pubspec/pubspec.yaml
nielsenko: I agree. I was just following the existing tests.

tools/serverpod_cli/test/integration/serverpod_packages_version_check/load_config_test.dart
nielsenko: Okay

tools/serverpod_cli/lib/src/serverpod_packages_version_check/serverpod_packages_version_check.dart
nielsenko:
```sh
serverpod generate 
An error occurred while parsing the server config file: Failed to load client pubspec.yaml. If you are using a none default path it has to be specified in the
config/generator.yaml file!

serverpod version
Serverpod version: 2.4.0
```

tools/serverpod_cli/test/integration/serverpod_packages_version_check/test_assets/invalid_pubspec/pubspec.yaml
nielsenko: Actually  this is used in the new tests for `GeneratorConfig.load` which reads directly from the filesystem.

SandPod: My though was that we could write them to a temp location directly in the test. Would that be possible?

nielsenko: Yes, we can. I'll rework to do that.

tools/serverpod_cli/test/integration/serverpod_packages_version_check/load_config_test.dart
nielsenko: I removed the prefix ' ' but kept the multiline

tools/serverpod_cli/test/integration/serverpod_packages_version_check/serverpod_packages_version_check_test.dart
nielsenko: I have split each .having into its own test

tools/serverpod_cli/lib/src/config/config.dart
nielsenko: Assuming a `config/generator.yaml` containing just:
```
invalid yaml
```

In 2.4 we hapily ignore any parse errors:
```
serverpod generate 
✓ Generating code (10.0s)
✅ Done.
```
going forward we will error out like:
```
serverpod generate
Building package executable... (12.6s)
Built serverpod_cli:serverpod_cli.
An error occurred while parsing the server config file: Error on line 1, column 1 of config/generator.yaml: Expected a map
  ╷
1 │ invalid yaml
  │ ^^^^^^^^^^^^
  ╵
```

tools/serverpod_cli/test/integration/serverpod_packages_version_check/load_config_test.dart
nielsenko: Now I removed use of 'and' and multi-line as well

tools/serverpod_cli/test/integration/serverpod_packages_version_check/serverpod_packages_version_check_test.dart
nielsenko: I avoid the use of 'and' as well

tools/serverpod_cli/test/integration/serverpod_packages_version_check/test_assets/invalid_pubspec/pubspec.yaml
nielsenko: All test asset files has been removed.

SandPod (approved):
LGTM! 🚀

Issue #3149, opened 2025-01-20
Error in protocol in client for "Client" named model.


SandPod:
### Describe the bug

If a model is created with the class name `Client`, errors are reported in the client generated protocol file reporting that the class is missing `toJson()` and `fromJson()`.

### To reproduce

1. Create a model with a class named client:

```yaml
class: Client
fields:
    name: String
```

2. Generate the project.

### Expected behavior

If `Client` needs to be reserved by the framework then a warning when generating the code should inform me that I need to change the name, if not, the code should generate without error.

### Library version

Serverpod version: 2.3.1
Dart version: 3.6.0

### Platform information

macOS: Sonoma 14.6.1

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

PR #3288, opened 2025-03-06
fix: Fail early on invalid class and file names

nielsenko:
Some file and class names are not useful when defining models as they will conflict with other files and classes we generate.
In particular the file names:

- _client.dart_
- _endpoints.dart_
- _protocol.dart_

and the matching class names:

- **Client**
- **Endpoints**
- **Protocol**

We need to check for both as there are no guarantees that class and file names are aligned for models.

This PR adds extra validation so that `serverpod generate` will catch this, instead of generating invalid code.

We can get rid of these restrictions, but it will be a breaking change as it will impact the import paths used in client code.

fix: #3149

## Pre-launch Checklist

## Breaking changes

None. Trying to use these names would already leave your project in a broken state. This fix merely adds validation.

### DIFF
--- a/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
@@ -156,7 +156,16 @@ class Restrictions {
       ];
     }
 
-    var reservedClassNames = const {'List', 'Set', 'Map', 'String', 'DateTime'};
+    const reservedClassNames = {
+      'List',
+      'Set',
+      'Map',
+      'String',
+      'DateTime',
+      'Client',
+      'Endpoints',
+      'Protocol',
+    };
     if (reservedClassNames.contains(className)) {
       return [
         SourceSpanSeverityException(
--- a/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_conflicts_test.dart
+++ b/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_conflicts_test.dart
@@ -1,4 +1,5 @@
 import 'package:serverpod_cli/analyzer.dart';
+import 'package:serverpod_cli/src/analyzer/code_analysis_collector.dart';
 import 'package:serverpod_cli/src/analyzer/models/stateful_analyzer.dart';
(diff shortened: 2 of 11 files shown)


tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -156,7 +156,16 @@ class Restrictions {
       ];
     }
 
-    var reservedClassNames = const {'List', 'Set', 'Map', 'String', 'DateTime'};
+    const reservedClassNames = {
```
SandPod, line 159: Suggestion: Please add a test Clients Endpoints and Protocol in `model_types_test.dart`.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_conflicts_test.dart
```diff
@@ -105,4 +106,32 @@ void main() {
       reason: 'Expected no errors but some were generated.',
     );
   });
+
+  for (var className in ['Client', 'Endpoints', 'Protocol']) {
```
SandPod, line 110: Comment: Aaah, the test was added here. Please consolidate this with the tests in `model_types_test.dart`.

SandPod (commented):
The changes look good! 👍 

Single comment around consolidating the location for the reserved keywords tests.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_conflicts_test.dart
nielsenko: I have moved all tests to `model_types_test`.dart and aligned with the flat layout used there. We could consider doing the opposite? It is a _class_conflict_.

SandPod (approved):
LGTM! 🚀

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_conflicts_test.dart
SandPod: The way it is now looks good to me! 👍

Issue #3213, opened 2025-02-09
ByteData Serialization Bug (Image Upload)


Lightgaming:
### Describe the bug

When I try to upload some images the storeFile Function will fail and Serverpod closes the connection to the Flutter Client. 

I already increased the maxRequestSize to 2000000000000 - which should be plenty ^^

It also happens when I try to upload GIF files. 

Since the storeFile Function returns void and Serverpod is not logging any Error Message I have trouble debugging the cause of this issue. 

Client Error Message:

flutter: Failed call: images.uploadEventImage
flutter: HttpException: Connection closed before full header was received, uri = http://[redacted]:8080/images
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: HttpException: Connection closed before full header was received, uri = http://[redacted]:8080/images

### To reproduce

Here is the Code I use:

Client: 

```dart
      FilePickerResult? result = await FilePicker.platform.pickFiles(
        type: FileType.image,
        allowMultiple: false,
        withData: true,
      );
      if (result == null) return;
      final url = await uploadImage(result.files.first);
```

```dart
Future<String?> uploadImage(PlatformFile file) async {
  if (file.bytes == null) {
    debugPrint('File bytes is null');
    return null;
  }

  final fileExists = await client.images.checkFileExists(file.name);
  if (fileExists) {
    final publicUrl = await client.images.getPublicImageUrl(file.name);
    if (publicUrl != '') {
      return publicUrl;
    }
    return null;
  }

  await client.images.uploadEventImage(
    file.name,
    file.bytes!.buffer.asByteData(),
  );

  final fileExistsAfterUpload = await client.images.checkFileExists(file.name);
  if (fileExistsAfterUpload) {
    final publicUrl = await client.images.getPublicImageUrl(file.name);
    if (publicUrl != '') {
      return publicUrl;
    }
    return null;
  }
  return null;
}
```

Server:

```dart
Future<bool> checkFileExists(Session session, String path) async {
    return await session.storage.fileExists(
      storageId: 'public',
      path: path,
    );
  }

  Future<String> getPublicImageUrl(Session session, String path) async {
    final uri = await session.storage.getPublicUrl(
      storageId: 'public',
      path: path,
    );
    return uri.toString();
  }

  Future<void> deleteImage(Session session, String path) async {
    try {
      return await session.storage.deleteFile(
        storageId: 'public',
        path: path,
      );
    } catch (e) {
      print(e);
    }
  }

[...]

PR #3291, opened 2025-03-07
fix: Ensure we drain the request data stream completely, even if we realize it is too large

nielsenko:
This is a workaround to https://github.com/dart-lang/sdk/issues/60271

fix: #3213 

## Pre-launch Checklist

### DIFF
--- a/packages/serverpod/lib/src/server/server.dart
+++ b/packages/serverpod/lib/src/server/server.dart
@@ -428,16 +428,25 @@ class Server {
   }
 
   Future<String> _readBody(HttpRequest request) async {
-    var builder = BytesBuilder();
+    var builder = BytesBuilder(copy: false);
     var len = 0;
-    await for (var segment in request) {
-      len += segment.length;
-      if (len > serverpod.config.maxRequestSize) {
-        throw _RequestTooLargeException(serverpod.config.maxRequestSize);
+    var maxRequestSize = serverpod.config.maxRequestSize;
+    var tooLargeForSure = request.contentLength > maxRequestSize;
+    if (!tooLargeForSure) {
+      await for (var segment in request) {
+        if (tooLargeForSure) continue; // always drain request, if reading begun
+        len += segment.length;
+        tooLargeForSure = len > maxRequestSize;
+        builder.add(segment);
       }
-      builder.add(segment);
     }
-    return const Utf8Decoder().convert(builder.toBytes());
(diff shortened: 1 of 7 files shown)


packages/serverpod/lib/src/server/server.dart
```diff
@@ -430,15 +430,24 @@ class Server {
...
-    await for (var segment in request) {
-      len += segment.length;
-      if (tooLarge) continue; // ensure we drain the request stream regardless
-      if (len > serverpod.config.maxRequestSize) tooLarge = true;
-      builder.add(segment);
+    var maxRequestSize = serverpod.config.maxRequestSize;
+    var tooLarge = request.contentLength > maxRequestSize;
+    // if request.contentLength is -1, it means the length is unknown
+    // and we may realize later that the request is still too large
+    if (!tooLarge) {
```
SandPod, line 437: Suggestion/Discussion(non-blocking NIT): Would readability be helped by using an early return here?

I'm generally not a big fan of having logic placed inside of if statements. Although this would give two reasons to throw. But the first one could simply be 
```dart
if (request.contentLength > maxRequestSize) throw _RequestTooLargeException(maxRequestSize);
```

And then the `tooLarge` variable would only be connected to the workaround.

This would perhaps also make the `-1` detail implicit, but we could make that explicit with a comment just like now.

SandPod (approved):
LGTM! 👍 

Single suggestion/discussion for improved readability

packages/serverpod/lib/src/server/server.dart
nielsenko: What if I renamed the `tooLarge` to `tooLargeForSure` to indicate that it may be a false negative?

nielsenko: So goes like
```dart
var tooLargeForSure = ...
if (!tooLargeForSure) // attempt to read, update tooLargeForSure as we go along
if (tooLargeForSure) // throw
```
I updated the PR to use this approach.

SandPod: Hmm, I don't think this really improves readability. The early return would target carrying context through the method. Making it clear that nothing below the first check of `request.contentLength > maxRequestSize` would run.

But it is also such a small method that this context probably doesn't matter.

SandPod: Feel free to resolve if you dissagree since this is a non-blocking NIT :)

PR #3299, opened 2025-03-13
feat: Allow changing a table id type between int and uuid

marcelomendoncasoares:
This PR implements the first feature described at #3255: allow the user to declare the id field type in serverpod models, changing its type to one of the supported id types (with the correct default attribution).

The implementation changes the following:

- There is now a new experimental feature flag named `changeIdType`. The user is only allowed to declare the id field and change its type if this experimental feature is enabled.
- The id field is no longer reserved for table classes, but only the `default` key can be specified. For the `int` type, the `default` key is optional with the value `serial`; for all others, it is required.
- All supported id types are declared in the new `SupportedIdType` class. Currently supports `int` (default) and `uuid` (with `default=random` to generate UUIDv4 values).

The missing test for the PR #3257 is implemented in this PR. It validates that the stored FK field type on the relation objects is being collected from the correct model. Its the file `relation_changed_field_id_type_test.dart` of the `serverpod_cli` model validation tests.

Most of the id type tests were implemented with a little abstraction to ensure that all supported id types (current and future) are working as expected - both in isolation and in relations. This way, new supported id types in the future will be already 99% covered by tests, pending only to add one single line in the `migration_sql_test.dart` file to complete the coverage.

## Pre-launch Checklist

## Breaking changes

No breaking changes.


tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -1103,6 +1116,28 @@ class Restrictions {
     return [];
   }
 
+  List<SourceSpanSeverityException> validateScopeKey(
```
SandPod, line 1131: Suggestion: Please validate the keywords in the test file designated to the keyword.

When finding tests, it will be hard to understand that parts of the `scope` restrictions are actually tested in the `field_id_type_test.dart`. It is simpler if the tests are grouped for their respective keyword.

Then the `id` column is just a different `Given`.

SandPod: Please apply this principle for the tests of all modified keywords (Persist, Default)

tools/serverpod_cli/test/generator/dart/client_code_generator/changed_id_type_test.dart
```diff
@@ -0,0 +1,86 @@
...
+    'protocol',
+    '$testClassFileName.dart',
+  );
+
+  for (var idType in SupportedIdType.all) {
+    var idClassName = idType.className;
+    var idTypeAlias = idType.aliases.first;
+
+    group('Given the id type is $idTypeAlias', () {
+      group('Given a class with table name when generating code', () {
```
SandPod, line 33: Suggestion: Build a valid test description.

`Given` is repeated multiple times for this test. Please make sure the description is valid and helpful for future developers.

tools/serverpod_cli/test/generator/dart/client_code_generator/changed_id_type_test.dart
```diff
@@ -0,0 +1,86 @@
...
+              .build()
+        ];
+
+        var codeMap = generator.generateSerializableModelsCode(
+          models: models,
+          config: config,
+        );
+
+        var compilationUnit =
+            parseString(content: codeMap[expectedFileName]!).unit;
```
SandPod, line 48: Suggestion: Place this in a setup or a late so that a single test can be run in isolation.

tools/serverpod_cli/test/generator/dart/client_code_generator/changed_id_type_test.dart
SandPod: Same for all the tests in `server_code_generator/changed_id_type`.

tools/serverpod_cli/test/test_util/builders/database/column_definition_builder.dart
```diff
@@ -24,12 +26,21 @@ class ColumnDefinitionBuilder {
     );
   }
 
-  ColumnDefinitionBuilder withIdColumn() {
+  ColumnDefinitionBuilder withIdColumn(
+    String tableName, {
+    String type = 'int',
```
SandPod, line 31: Suggestion: Take in a `SupportedIdType` to give this a nice typed interface.

tools/serverpod_cli/test/test_util/builders/database/table_definition_builder.dart
```diff
@@ -29,6 +30,10 @@ class TableDefinitionBuilder {
         _managed = true;
 
   TableDefinition build() {
+    _columns.insert(
+      0,
+      ColumnDefinitionBuilder().withIdColumn(_name, type: _idType).build(),
+    );
```
SandPod, line 36: Suggestion: Don't insert this as part of the build step.

This has changed the functionality of the builder since all columns where overwritable before through `withColumns`.

tools/serverpod_cli/lib/src/config/experimental_feature.dart
```diff
@@ -12,9 +12,13 @@ class CommandLineExperimentalFeatures {
 
 enum ExperimentalFeature {
   all,
+
   // TODO: Remove when inheritance is enabled by default.
   // Tracked by issue: https://github.com/serverpod/serverpod/issues/2711
-  inheritance;
+  inheritance,
+
+  // TODO: Remove when the feature is considered stable.
```
SandPod, line 20: Suggestion: Add a link to the UUID issue here.

tests/serverpod_test_server/lib/src/models/changed_id_type/many_to_many/course.spy.yaml
```diff
@@ -0,0 +1,6 @@
+class: CourseUuid
```
SandPod, line 1: Suggestion: Please rename `changed_id_type` to `uuid_id_type`.

tests/serverpod_test_server/lib/src/generated/changed_id_type/many_to_many/course.dart
```diff
@@ -0,0 +1,575 @@
...
+// ignore_for_file: public_member_api_docs
+// ignore_for_file: type_literal_in_constant_pattern
+// ignore_for_file: use_super_parameters
+
+// ignore_for_file: no_leading_underscores_for_library_prefixes
+import 'package:serverpod/serverpod.dart' as _i1;
+import 'package:uuid/uuid.dart' as _i2;
+import '../../changed_id_type/many_to_many/enrollment.dart' as _i3;
+
+abstract class CourseUuid
```
SandPod, line 16: Suggestion Add tests that run db queries towards models that use UUID.

I might have missed it but I didn't see any tests that actually run relations queries towards the models with uuid?

tools/serverpod_cli/lib/src/generator/types.dart
```diff
@@ -672,6 +673,56 @@ class TypeDefinition {
   }
 }
 
+/// Supported ID type definitions.
+/// All configuration to support other types is done only on this class. For
+/// new variants of id types, add a new static getter and update the [all]
+/// getter. For different types, it is necessary to also update the `Table`
+/// constructor at `packages:serverpod/src/database/concepts/table.dart`.
```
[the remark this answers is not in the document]
marcelomendoncasoares: The word `variant` was to indicate that it is possible to have multiple entries of the same column type, but with different default values (and aliases). This is what will enable supporting both uuidV4 and uuidV7 lately, and is necessary for the user to be able to configure a default id type for the project (using the alias) in a next PR.

Just enhanced the text to better explain the difference:

```dart
/// Supported ID type definitions.
/// All configuration to support other types is done only on this class. A
/// supported id type is a combination of a type and a default value. It is
/// possible to support multiple default values for a same type (e.g. uuid type
/// with default of versions 4 and 7), with each having its getter. For new
/// entries, add a new static getter and update the [all] getter. If the entry
/// contains a type not yet used by existing id types, it's necessary to update
/// the `Table` constructor at `packages:serverpod/src/database/concepts/table.dart`.
```

Do you think like this is clearer?

tools/serverpod_cli/lib/src/generator/types.dart
```diff
@@ -672,6 +673,56 @@ class TypeDefinition {
...
+  final List<String> aliases;
+
+  /// The default value for the column on the database definition. Must be one
+  /// of the supported defaults for the type.
+  final String defaultValue;
+
+  /// The class name of the id type.
+  String get className => type.className;
+
+  /// If no id type is specified, the default id type is [int].
```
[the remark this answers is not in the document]
marcelomendoncasoares: This facilitated the exploration of the solution during development (changing between using a direct `TypeDefinition` or the new `SupportedIdType` class), but it is really not necessary. Just removed.

tools/serverpod_cli/lib/src/database/extensions.dart
```diff
@@ -357,6 +343,21 @@ extension ColumnDefinitionPgSqlGeneration on ColumnDefinition {
...
+      if (isNullable != false) {
+        throw (const FormatException('The id column must be non-nullable'));
+      }
+      if (((type == 'integer') || (type == 'bigint')) &&
+          (columnDefault?.startsWith('nextval') ?? false)) {
+        type = 'bigserial';
+        defaultValue = '';
+      }
+
+      type = '$type PRIMARY KEY';
```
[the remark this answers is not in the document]
marcelomendoncasoares: Done.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -245,28 +245,12 @@ class ModelParser {
...
-            '/// database or if it has been fetched from the database. Otherwise,',
-            '/// the id will be null.',
-          ],
-        ),
-      );
-    }
 
     var fieldsNode = documentContents.nodes[Keyword.fields];
-    if (fieldsNode is! YamlMap) return fields;
+    if (fieldsNode is! YamlMap?) return fields;
```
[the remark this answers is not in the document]
marcelomendoncasoares: Indeed, this return now is placed before the id field inclusion. Just changed the fields expansion to inside the `YamlMap` check, removing the early return.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -275,6 +259,39 @@ class ModelParser {
...
+        defaultValue ??= SupportedIdType.int.defaultValue;
+      }
+
+      var defaultIdFieldDoc = [
+        '/// The database id, set if the object has been inserted into the',
+        '/// database or if it has been fetched from the database. Otherwise,',
+        '/// the id will be null.',
+      ];
+
+      fields.removeWhere((f) => f.name == 'id');
```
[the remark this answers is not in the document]
marcelomendoncasoares: It already exists, actually: `defaultPrimaryKeyName`, on `definitions.dart`. But its rarely used, so I ended up following what was most common on the package. But we're better off without magic strings indeed.

Just replaced all occurrences of 'id' by the constant on the PR changes only, because replacing everywhere will require a careful revision and might be better to be refactored apart (specially due to interpolations, tests, etc.). Do you agree?

tools/serverpod_cli/lib/src/database/extensions.dart
```diff
@@ -357,6 +343,21 @@ extension ColumnDefinitionPgSqlGeneration on ColumnDefinition {
...
+    // The id column is special.
+    if (name == 'id') {
+      if (isNullable != false) {
+        throw (const FormatException('The id column must be non-nullable'));
+      }
+      if (((type == 'integer') || (type == 'bigint')) &&
+          (columnDefault?.startsWith('nextval') ?? false)) {
+        type = 'bigserial';
+        defaultValue = '';
+      }
```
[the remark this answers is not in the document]
marcelomendoncasoares: Created two getters with the conditions to improve the readability. Better this way?

```dart
/// Whether the column is the default primary key column.
bool get isIdColumn => name == defaultPrimaryKeyName;

/// Whether the column is a primary key of type int serial.
bool get isIntSerialIdColumn =>
    isIdColumn &&
    (columnType == ColumnType.integer || columnType == ColumnType.bigint) &&
    (columnDefault?.startsWith('nextval') ?? false);
```

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -249,8 +249,10 @@ class ModelParser {
       fields.add(
         SerializableModelFieldDefinition(
           name: 'id',
-          type: TypeDefinition.int.asNullable,
+          type: SupportedIdType.int.type.asNullable,
           scope: ModelFieldScopeDefinition.all,
+          defaultModelValue: SupportedIdType.int.defaultValue,
```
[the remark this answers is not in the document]
marcelomendoncasoares: Having the model here makes the id declaration more predictable internally: both model and database default are expected to be set for all possible id types. Though for int it is not to be used, for uuid it is good to have the value being created on the class.

We can bubble up the int special handling from the `model_library_generator.dart` to here, but I preferred to push it down so that all tests could treat id fields the same - with both model and database default set -, regardless of the int special case, to avoid conditional checks at maximum.

Do you think that having the special case handling in the `model_library_generator.dart` is a deal breaker?

marcelomendoncasoares:
Still pending to follow-up in the tests revision, but will get to them very soon!

tests/serverpod_test_server/lib/src/models/changed_id_type/many_to_many/course.spy.yaml
marcelomendoncasoares: I've chosen `changed_id_type` because all models in this dir have the `id` field declared, but not exclusively to `uuid`. Some are `uuid`, others are declared to `int` (with and without default value). When added support for uuidv7, some will be changed from v4 to v7 (as there are a lot of uuid models). And, with the support for other id types in the future, some of this models could be used for the new types. Do you think we should change to `uuid_id_type` either way?

tools/serverpod_cli/lib/src/config/experimental_feature.dart
marcelomendoncasoares: Done. I've written as `Feature tracked by issue` to make clear that the issue will not necessarily result in the feature being considered stable when gets closed.

tools/serverpod_cli/test/test_util/builders/database/column_definition_builder.dart
marcelomendoncasoares: Done.

tools/serverpod_cli/test/test_util/builders/database/table_definition_builder.dart
marcelomendoncasoares: Done.

tools/serverpod_cli/lib/src/generator/types.dart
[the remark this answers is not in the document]
SandPod: Yes! Thank you! 🙏

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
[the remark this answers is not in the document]
SandPod: I'm not completely convinced, but you might be right. Let's take this up in our discussion tomorrow! 👍

SandPod (commented):
Specifically looked at the changes regarding the comments that have been addressed.

tools/serverpod_cli/lib/src/generator/types.dart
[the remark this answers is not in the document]
marcelomendoncasoares: Done.

```dart
  /// Id type that generates sequential integer values.
  static SupportedIdType get int => SupportedIdType(...);

  /// Id type that generates UUID v4 values.
  static SupportedIdType get uuidV4 => SupportedIdType(...);
```

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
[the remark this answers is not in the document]
marcelomendoncasoares: I have just removed the default model value for the int id type, but allowing other defaults (such as `random` for UUID, to allow for the id creation on the client). With this, there is no more special handling in the `model_library_generator.dart` and all manipulation of default values is performed directly on the fields parsing.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_id_type_test.dart
```diff
@@ -56,34 +190,35 @@ void main() {
...
-      var definition = definitions.first as ClassDefinition;
+      StatefulAnalyzer(config, models, onErrorsCollector(collector))
+          .validateAll();
 
-      expect(definition.fields.first.name, isNot('id'));
-      expect(definition.fields, hasLength(1));
+      expect(
+        collector.errors.first.message,
+        contains(
+          'The default value "1" is not supported for the id type "int".',
```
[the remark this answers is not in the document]
marcelomendoncasoares: Done.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
marcelomendoncasoares: Done. I've refactored the `field_id_type_test.dart` to move each keyword validation to its place and also remove the abstractions.

tools/serverpod_cli/test/generator/dart/client_code_generator/changed_id_type_test.dart
marcelomendoncasoares: Done. Also removed the abstractions.

tools/serverpod_cli/test/generator/dart/client_code_generator/changed_id_type_test.dart
marcelomendoncasoares: No longer necessary, as we do not have loops anymore.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_id_type_test.dart
```diff
@@ -1,51 +1,185 @@
...
+    test('then the id type is nullable.', () {
+      expect(definition.fields.first.type.nullable, true);
+    });
+
+    test('then the default persist is "serial".', () {
+      expect(definition.fields.first.defaultPersistValue, defaultIntSerial);
+    });
+  });
+
+  for (var idType in SupportedIdType.all) {
```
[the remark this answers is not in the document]
marcelomendoncasoares: Guess we can close this, since the test has been refactored and all abstractions/automations were removed.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_changed_field_id_type_test.dart
```diff
@@ -0,0 +1,614 @@
...
+
+  for (var idType1 in SupportedIdType.all) {
+    var idTypeAlias1 = idType1.aliases.first;
+    var idClassName1 = idType1.className;
+
+    for (var idType2 in SupportedIdType.all) {
+      var idClassName2 = idType2.className;
+      var idTypeAlias2 = idType2.aliases.first;
+
+      group('For id type combination of $idTypeAlias1 and $idTypeAlias2', () {
```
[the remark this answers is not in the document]
marcelomendoncasoares: Just refactored the test to:
1. Remove the matrix, as the test aims to validate relations with different ids, but we can take one combination for all.
2. Remove tests redundancy, as pointed in this comment - there was indeed overlap between some of them.
3. Improve the test descriptions to match `Given`, `When`, `Then` (comment below).

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_changed_field_id_type_test.dart
```diff
@@ -0,0 +1,614 @@
...
+        group('For manual relations', () {
```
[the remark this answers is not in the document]
marcelomendoncasoares: Done.

tests/serverpod_test_server/lib/src/generated/changed_id_type/many_to_many/course.dart
marcelomendoncasoares: Done. To generate an even greater coverage, I have copied the tests from some of the existing relation tests. This way, we simultaneously ensure querying towards models with UUID and relations between models with different ids. Do you think this is enough?

tests/serverpod_test_server/test_integration/database_operations/changed_id_type_relations/one_to_many/count/where_every_test.dart
```diff
@@ -0,0 +1,203 @@
...
+            description: 'Prem: OrderUuid 5', customerId: customers[3].id!),
+        OrderUuid(
+            description: 'Prem: OrderUuid 6', customerId: customers[3].id!),
+      ]);
+
+      var customerCount = await CustomerInt.db.count(
+        session,
+        // All customers where every order starts with 'prem'
+        where: (c) => c.orders.every((o) => o.description.ilike('prem%')),
+      );
```
SandPod, line 46: Question: seems like most of these tests have an `int` that is querying a `uuid` model. Do we need to add a test for a `uuid` model that queries an `int` model or do we think that is covered by the other tests, since they all query towards an `int` model?

tests/serverpod_test_server/migrations/migration_registry.txt
```diff
@@ -8,4 +8,6 @@
 20240827080004999
 20241007090553465
 20250211073932976
-20250313015420884
+20250312134724494
+20250401060242031
+20250407042334770
```
SandPod: Suggestion: Feel free to squash all of these migrations into the already existing one.

SandPod (commented):
This is looking good to me!

Had a last couple of questions then we are ready to go I think :)

tests/serverpod_test_server/lib/src/generated/changed_id_type/many_to_many/course.dart
marcelomendoncasoares: I used `changed_id_type_relations` for coherence with the `changed_id_type` folder under `serverpod_test_server/lib/src/models` and because, when new id types gets created, some of these classes can be changed to the new id type - as there are many for each type and more than one id type under the folder. Do you think that it would be better still to narrow the folder name to `uuid_id` tests?

tools/serverpod_cli/test/generator/dart/client_code_generator/changed_id_type_test.dart
```diff
@@ -25,62 +25,97 @@ void main() {
...
+    var codeMap = generator.generateSerializableModelsCode(
+      models: models,
+      config: config,
+    );
+
+    var compilationUnit = parseString(content: codeMap[expectedFileName]!).unit;
+    var maybeClassNamedExample = CompilationUnitHelpers.tryFindClassDeclaration(
+      compilationUnit,
+      name: testClassName,
+    );
```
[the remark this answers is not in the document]
marcelomendoncasoares: Done.

tests/serverpod_test_server/test_integration/database_operations/changed_id_type_relations/one_to_many/count/where_every_test.dart
marcelomendoncasoares: I'd say we can consider this case covered by the other tests indeed. The most important here, in my opinion, is querying relations to models with an id different than `int` (which can only be `uuid` for now).

tests/serverpod_test_server/migrations/migration_registry.txt
SandPod: I think we can leave it as is since the PR is in good enough shape to merge now :)

tests/serverpod_test_server/lib/src/generated/changed_id_type/many_to_many/course.dart
SandPod: I think it is probably OK to leave as is in that case. `changed_id_type` covers what is important.

SandPod (approved):
Big change and really nice improvement to the framework!

Thank you for all the great work here! 🙏

PR #3306, opened 2025-03-14
Explore APIs for next version of authentication module

tp:


serverpod_auth_2/lib/additional_data.dart
```diff
@@ -0,0 +1 @@
+typedef AdditionalData = Map<String, dynamic>;
```
tp, line 1: Developers will need to be careful to basically only put ids in here, so that they can not be tempered with on the client (eg. use `{"invitation_code": "asdf"}` instead of `{"team": "7"}`).

Else maybe we should facilitate signing of the data (though lookups would also be nicer to smaller payload), such that developers get secure defaults.

examples/auth_example/auth_example_server/lib/src/generated/protocol.dart
```diff
@@ -52,14 +64,22 @@ class Protocol extends _i1.SerializationManagerServer {
...
-    if (data is _i4.Example) {
+    if (data is _i6.Example) {
       return 'Example';
     }
     className = _i2.Protocol().getClassNameForObject(data);
     if (className != null) {
       return 'serverpod.$className';
     }
     className = _i3.Protocol().getClassNameForObject(data);
+    if (className != null) {
```
tp, line 75: This is problematic, as the `className` used for serialization would change with the file being moved to the new package (from `serverpod_auth.X` to `serverpod_auth_email.X`).

Thus this would make this a breaking change towards the client.

Also tricky is the fact that since this protocol defers to each package's helper, it wasn't initially obvious, that the server lost the e-mail endpoints, because the protocol was unchanged (and no methods visibly removed).

The good news is, that the database storage survived the move without issue, as no migration was created (supposedly the tables are not prefixed, to be double checked).

modules/serverpod_auth/serverpod_auth_server/pubspec.yaml
```diff
@@ -7,22 +7,20 @@ version: 2.4.0
...
   openid_client: ^0.4.8
   meta: ^1.8.0
-  googleapis: '>=11.0.0 <14.0.0'
+  googleapis: ">=11.0.0 <14.0.0"
   googleapis_auth: ^1.6.0
-  http: '>=1.0.0 <2.0.0'
+  http: ">=1.0.0 <2.0.0"
   image: ^4.0.15
   jose: ^0.3.3
-  pointycastle: ^3.7.4
```
tp, line 25: Check client generation when we add the email auth & endpoint dependencies here.

Would the client be properly generated? Likely at least the `protocol.dart` would exhibit the same issue because the model class `EmailLogin` from `email_auth` is used without being a direct import.

examples/auth_example/auth_example_server/lib/src/generated/protocol.dart
tp:
```suggestion
    if (className != null && !className.contains('.')) {
```

Could be a quick-fix to support fully qualified name (if they want to "pretend" to be from a different module). But the issue would also occur on deserialization, so maybe a better structure to explicitly list these "renames" would make sense, such that the same can also be used for incoming messages.

examples/auth_example/auth_example_client/lib/src/protocol/client.dart
```diff
@@ -30,10 +34,16 @@ class EndpointExample extends _i1.EndpointRef {
...
 class Modules {
   Modules(Client client) {
-    auth = _i3.Caller(client);
+    serverpod_auth_email = _i3.Caller(client);
+    serverpod_auth_email_endpoint = _i4.Caller(client);
+    auth = _i5.Caller(client);
   }
 
-  late final _i3.Caller auth;
+  late final _i3.Caller serverpod_auth_email;
```
tp, line 42: TBD; would be nice not to have this, if no endpoints are defined, but since this is a straight important, it would require code-gen in this project if the dependency adds or removes an endpoint (at least for changes between 0 and !0).

examples/auth_example/auth_example_client/lib/src/protocol/client.dart
```diff
@@ -30,10 +34,16 @@ class EndpointExample extends _i1.EndpointRef {
...
+
+  late final _i4.Caller serverpod_auth_email_endpoint;
```
tp, line 44: This would of course be a breaking change for the client.
A small fix for sure (if we were to get the endpoint mapping & model parsing to get rewired on the backend to support older clients), but nonetheless it'd require a few 1-line changes if the app (client) uses its own UI instead of the `serverpod_email_flutter` package with pre-built component (in which case one would still required to update those packages in unison though).

PR #3316, opened 2025-03-18
test: Add compilation unit matchers.

SandPod:
This PR introduces a Compilation Unit Matcher, which enables matching the structure of a compilation unit rather than its raw textual representation.

By focusing on structural validation, this matcher allows us to write tests that verify the properties of the generated code without being affected by formatting changes. This makes our tests more resilient to updates in the code generator that may alter formatting but not the actual structure.

### Matcher structure
The matchers introduced in this PR are sufficient enough to replace all existing tests in `entity_implicit_class_test.dart`. The plan is to extend these matchers as needed for future use cases.

The currently available matchers are structured as follows:

Screenshot 2025-03-18 at 15 16 09

### FormattedCompilationUnit

When using the `parseCode` method to parse a string representation of Dart code, the resulting compilation unit is wrapped in a `FormattedCompilationUnit`. This ensures that the actual output is nicely formatted, making it as easy as possible to inspect the `actual` output of the test..

## Pre-launch Checklist

## Breaking changes

None - This is a simple test change that should not anything else.


tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
```diff
@@ -0,0 +1,34 @@
...
+/// ```
+ClassMatcher containsClass(String className) => _ClassMatcherImpl._(className);
+
+/// Parses a string of Dart code into a FormattedCompilationUnit.
+FormattedCompilationUnit parseCode(String code) =>
+    FormattedCompilationUnit(parseString(content: code).unit);
+
+/// A matcher that checks if a CompilationUnit contains a class that matches
+/// certain criteria.
+abstract interface class ClassMatcher {}
```
nielsenko: Why not have the interface class `implement Matcher`?

There are various examples later where `containsClass` is cast `as Matcher`.

tools/serverpod_cli/test/test_util/compilation_unit_matcher/class_matcher/class_matcher_test.dart
```diff
@@ -0,0 +1,85 @@
...
+    test(
+        'when matching with non-compilation unit then mismatch description is correct',
+        () {
+      final matcher = containsClass('User') as Matcher;
+      final description = StringDescription();
+      matcher.describeMismatch('Not a CompilationUnit', description, {}, false);
+
+      expect(
+        description.toString(),
+        equals('"String" is not a CompilationUnit'),
```
nielsenko, line 68: Question: Where does `"String"` come from here?

tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
```diff
@@ -20,7 +23,26 @@ FormattedCompilationUnit parseCode(String code) =>
...
+  /// );
+  /// ```
+  ///
+  /// Use [isNullable] to match field nullability. If the value is not set, the
+  /// matcher will ignore the nullability of the field.
+  FieldMatcher withField(String fieldName, {bool? isNullable});
+}
+
+/// A chainable matcher that matches a field in a compilation unit.
+abstract interface class FieldMatcher {}
```
nielsenko, line 143: Why not implement `Matcher`?

tools/serverpod_cli/test/test_util/compilation_unit_matcher/field_matcher/field_matcher_test.dart
```diff
@@ -0,0 +1,236 @@
...
+        compilationUnit,
+        containsClass('User').withField('name', isNullable: true),
+      );
+    });
+
+    test('when negate matching class and non-nullable field then test passes',
+        () {
+      expect(
+        compilationUnit,
+        isNot(containsClass('User').withField('name', isNullable: false)),
```
nielsenko, line 176: Question: Why should this test pass?

nielsenko (commented):
Looks great so far. Some initial comments/questions. Will look at remaining commits later today.

tools/serverpod_cli/test/test_util/compilation_unit_matcher/class_matcher/class_matcher.dart
```diff
@@ -54,6 +54,25 @@ class _ClassMatcherImpl implements Matcher, ClassMatcher {
...
+
+    return _withConstructor(constructorName, isFactory: isFactory);
+  }
+
+  @override
+  ConstructorMatcher withUnnamedConstructor({
+    bool? isFactory,
+  }) {
+    return _withConstructor('', isFactory: isFactory);
+  }
```
nielsenko, line 93: A little gold plating 😄 .. 

Just exposing `_withConstructor` and pass empty string for unnamed would have kept the method names shorter. 

Also `isFactory` is a flag - why no `withNamedFactoryConstructor`? Not a recommendation, just an argument for exposing `_withConstructor` instead.

tools/serverpod_cli/test/test_util/compilation_unit_matcher/class_matcher/class_matcher_test.dart
```diff
@@ -82,4 +82,14 @@ void main() {
...
+
+  test(
+      'Given class and named constructor matcher when matching on empty name then exception is thrown',
+      () {
+    expect(
+      () => containsClass('User').withNamedConstructor(''),
+      throwsA(isA<ArgumentError>().having((e) => e.message, 'message',
+          equals('constructorName cannot be empty'))),
+    );
+  });
```
nielsenko, line 94: This test is a bit weird, since it tests that the matcher throws on its input. Disappears if `withNamedConstrutor` is dropped (see previous comment).

tools/serverpod_cli/test/test_util/compilation_unit_matcher/constructor_matcher/constructor_matcher.dart
```diff
@@ -0,0 +1,106 @@
...
+    var classConstructors = parent.matchedFeatureValueOf(actual);
+    if (classConstructors == null) return null;
+
+    return classConstructors
+        .where((c) => c._hasMatchingName(_name))
+        .firstOrNull;
+  }
+}
+
+extension _ConstructorDeclarationExtensions on ConstructorDeclaration {
```
nielsenko, line 170: No need to name private extension classes.
`extension on ConstructorDeclaration {` will do.

tools/serverpod_cli/test/test_util/compilation_unit_matcher/constructor_matcher/constructor_matcher_mismatch_description_test.dart
```diff
@@ -0,0 +1,108 @@
...
+        () {
+      final matcher = containsClass('User').withUnnamedConstructor() as Matcher;
+      final description = StringDescription();
+      matcher.describeMismatch(compilationUnit, description, {}, false);
+
+      expect(
+        description.toString(),
+        equals('does not contain an unnamed constructor'),
+      );
+    });
```
nielsenko, line 24: Actually, since it contains no constructors at all, it does contain an implicit unnamed constructor.

tools/serverpod_cli/test/test_util/compilation_unit_matcher/constructor_matcher/constructor_matcher_test.dart
```diff
@@ -0,0 +1,197 @@
...
+    });
+  });
+
+  group('Given compilation unit with class and unnamed factory constructor',
+      () {
+    late final compilationUnit = parseCode(
+      '''
+      class User {
+        factory User() => User();
+      }
```
nielsenko, line 114: I was wondering if this was a compile error or an infinite recursion. Turns out it is an infinite recursion.

tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
```diff
@@ -173,3 +178,6 @@ abstract interface class MethodMatcher {
 
 /// A chainable matcher that matches a parameter in a compilation unit.
 abstract interface class ParameterMatcher {}
+
+/// A chainable matcher that matches a super initializer in a compilation unit.
+abstract interface class SuperInitializerMatcher {}
```
nielsenko: Could this be extended to supporting any form of re-directing constructor?

nielsenko (approved):
Great job 🚀 . I have a few questions/suggestions, but nothing that I think is mandatory to address.

tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
SandPod: My thinking was that this is the interface that the developer will encounter in the test, and this should only expose the possibilities this matcher has.

If this implements `Matcher`, methods like `describe`, `match` and `describeMismatch` would be added to the interface.

By only implementing the `Matcher` interface on the `...Impl` class we limit their visibility to the environment where it matters.

tools/serverpod_cli/test/test_util/compilation_unit_matcher/class_matcher/class_matcher_test.dart
SandPod: This test was in the wrong place.

It is quite unclear but the `"String"` comes from the literal string `' Not a CompilationUnit'` that is provided to the matcher.

Will move it out to the other place where we test with a string.

tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
SandPod: Same motivation as for the `ClassMatcher`.

Will resolve this and then we can have the discussion there :)

tools/serverpod_cli/test/test_util/compilation_unit_matcher/field_matcher/field_matcher_test.dart
SandPod: Since the field is defined as `nullable` and the matcher matches a non-nullable field(not a match to the definition), the test passes when we negate the response from the matcher.

```dart
      class User {
        final String? name;
      }
```

tools/serverpod_cli/test/test_util/compilation_unit_matcher/class_matcher/class_matcher.dart
SandPod: The main motivation behind having `named` and `unnamed` as specific methods is to remove any ambiguity about which constructor we might match in the class.

By having a strict requirement on the name, we ensure that we always match a single constructor.

`isFactory` is something we can tag on as an additional optional requirement, but as far as I could tell, there was no ambiguity around what we are matching if it was not set.

The pattern from the other matchers are that we ignore any null optional parameters. So allowing `name` to be an optional parameter would require us to ignore matching on name completely, and in that case, we might match multiple constructor in the class.

tools/serverpod_cli/test/test_util/compilation_unit_matcher/class_matcher/class_matcher_test.dart
SandPod: The thinking is that the `ConstructorMatcher` always has a name set, so this isn't an issue there. But more a requirement for the `withNamedConstructor` matcher.

Does this in combination with the motivation to the previous question make more sense?

tools/serverpod_cli/test/test_util/compilation_unit_matcher/constructor_matcher/constructor_matcher.dart
SandPod: Removed all names for private extensions: [5dec490](https://github.com/serverpod/serverpod/pull/3316/commits/5dec490d69f3dbf47dee6345ac87d6f14de5eefb)

tools/serverpod_cli/test/test_util/compilation_unit_matcher/constructor_matcher/constructor_matcher_mismatch_description_test.dart
SandPod: Wouldn't we want to match this explicitly in that case? Since we are inspecting the compilation unit, we are only interested in what is defined.

So I believe this is wanted behavior when we are matching the compilation unit.

tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
SandPod: Do you have an example of a different kind?

If they are specifically defined I believe a different `<kind>InitializerMatcher` would be the answer. But if these can be dynamically defined we might need a different approach here.

tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
nielsenko: True, and I'm fine with it. However the standard matchers do not go out of their way to hide that they are matchers.

tools/serverpod_cli/test/test_util/compilation_unit_matcher/class_matcher/class_matcher.dart
nielsenko: I would not make name optional. I would just equate name == '' with unnamed, as the underlying implementation.

tools/serverpod_cli/test/test_util/compilation_unit_matcher/constructor_matcher/constructor_matcher_mismatch_description_test.dart
nielsenko: Maybe. Let us leave as is for now.

tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
nielsenko: I was thinking about examples such as:
```dart
class User {
  final String name)
  User._(this.name);
  User.randomName() : this._(Random().nextInt(1000).toString());
}
```

tools/serverpod_cli/test/test_util/compilation_unit_matcher/class_matcher/class_matcher_test.dart
nielsenko: I agree this test is needed, if we don't drop the split between `withNamed..` and `withUnnamed..` 😄 

Anyway, this is definitely in the nit category. If you are not convinced then please leave as is 👍

tools/serverpod_cli/test/test_util/compilation_unit_matcher/class_matcher/class_matcher.dart
SandPod: I'll keep it as is and well see if this becomes an issue later.

Given this class definition:
```
class MyClass {
  MyClass._(String name);
}
```

I think it would be unexpected if `containsClass('MyClass').withConstructor().withParameter('name')` would fail because there was a constructor but it was private.

tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
SandPod: Given the current pattern, this would be solved with a `ThisInitializerMatcher`.

If there are only `This` and `Super` available, I think the approach is OK since it gives us typed options.

But I'll keep it as is for now and well see if we run into any issues :)

Issue #3312, opened 2025-03-17
FutureCalls - enable running in parallel


yourny:
### Problem to Solve

I've got a FutureCall that runs for a long time. While it runs, no other future call is ran.
I also have a lot of FutureCalls in general which are used to run logic on the server independent of clients.
Having these run all serial would slow down my service massively since clients are waiting for their results.

### Proposal

I would like FutureCalls to run in parallel.

### Use Case

Having frequent FutureCall runs.

### Alternatives

none

### Additional context

After looking into the implementation, this seems to be due to future calls always waiting for the current one to finish before starting another.

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

yourny:
FutureCallManager - _checkQueue seems to be where this will have to be implemented

inf0rmatix:
Also adding to documentation more detail about how FutureCalls are processed is vital to anyone relying on FutureCalls more then once a day.

SandPod:
Hey @yourny 👋 

Makes perfect sense. We are looking at finalizing future calls during this spring. Then we will take a performance overview of the feature a long with completing the documentation.

inf0rmatix:
Hey @SandPod thanks for the quick response! :)

Do you think, just implementing the change to run future calls in parallel up to a defined max amount of concurrent calls per registered FutureCall would be a good solution?

This way, FutureCalls that need to be queued could still be queued while others that need to be run asap be run in parallel while being able to secure performance by specifying a set limit for every future call (and maybe one for all future calls in total)?

inf0rmatix:
I'd implement this if you don't mind :)

inf0rmatix:
I did a first iteration on a possible solution, please have a look @SandPod  :)

inf0rmatix:
Thanks for taking the time :)

My current implementation will ensure the behavior as is currently.

I implemented the feature to be configurable to have a global concurrent future call limit (which is by default = 1) as well as a per future call limit, which is by default = 1.

Global Future Call limit is specified using an environment variable.
Per Future Call limit is specified within the FutureCall where you can override a getter which by default is implemented to return 1.

I will look into your concerns regarding the session, did not think about that yet :)

[...]

PR #3326, opened 2025-03-19
feat: Implement concurrency for future call manager

inf0rmatix:
- Added a maximum concurrency limit for running future calls, which takes precedence over individual future call limits.
- Updated FutureCallManager to handle concurrency and ensure that the total number of concurrent future calls does not exceed the defined limit.
- Enhanced configuration to allow setting the concurrency limit via ServerpodConfig.
- Updated relevant classes and methods to support the new concurrency management features.
- Default behavior is unchanged

Solves #3312 

## Pre-launch Checklist

## Breaking changes


packages/serverpod/lib/src/server/future_call_manager.dart
```diff
@@ -88,91 +102,244 @@ class FutureCallManager {
...
     _shuttingDown = true;
 
-    await _pendingFutureCall.future;
+    await _waitForRunningFutureCalls();
   }
 
-  void _run() async {
+  void _run() {
     unawaited(_checkQueue());
   }
```
SandPod, line 110: Suggestion: Delete the `_run()` method.

This is only used inside of `start()`.

packages/serverpod/lib/src/server/future_call.dart
```diff
@@ -15,6 +15,9 @@ abstract class FutureCall<T extends SerializableModel> {
   /// The type of the data provided by when calling [invoke].
   Type get dataType => T;
 
+  /// The maximum number of concurrent running calls with the same [name].
+  int get concurrentLimit => 1;
```
SandPod, line 19: Discussion: I don't think this is a reasonable approach. Scheduled tasks should have no opinion on the environment they are run in. And when we expand this problem to multiple servers running in a cluster the guarantee would be different depending on what server picks up the scheduled task.

packages/serverpod_shared/lib/src/environment_variables.dart
```diff
@@ -141,6 +144,7 @@ enum ServerpodEnv {
       (ServerpodEnv.webPublicPort) => ServerpodServerConfigMap.publicPort,
       (ServerpodEnv.webPublicScheme) => ServerpodServerConfigMap.publicScheme,
       (ServerpodEnv.maxRequestSize) => 'maxRequestSize',
+      (ServerpodEnv.futureCallConcurrencyLimit) => 'futureCallConcurrencyLimit',
```
SandPod, line 147: Suggestion: Create a `futureCall` namespace and place `concurrencyLimit` under that.

This makes it possible to add the timeout configuration as well.

SandPod (changes requested):
Thank you for finding and attempting to tackle multiple issues with the Future call approach.

As this part of the code does not have exhaustive tests we need to tread carefully to ensure we don't brake existing functionality.

You are attempting at solving multiple issues with future calls in the same PR. Please break these apart into multiple PRs (suggest you start with running future calls concurrently), and add them together with tests that validate the changes work as expected.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -610,6 +624,26 @@ String _readServerId(
   return serverId;
 }
 
+int _readFutureCallConcurrencyLimit(
```
SandPod, line 627: Suggestion: A value less than 1 should mean that no concurrency limit is put in place.

Long term, we want to allow users to configure no specific limit for the number of future calls that are running concurrently. So short term we can use any negative or 0 value to represent that.

SandPod: These two do not contradict each other. Adding a "future call" concurrency limit should still be possible. It should just also be possible to configure "no limit" if that is desired.

A use case for this would be If the user is running many lightweight but slow future calls (say, fetching updates from multiple slow servers and updating the database state); this limitation would be redundant.

inf0rmatix:
> Thank you for finding and attempting to tackle multiple issues with the Future call approach.
> 
> As this part of the code does not have exhaustive tests we need to tread carefully to ensure we don't brake existing functionality.
> 
> You are attempting at solving multiple issues with future calls in the same PR. Please break these apart into multiple PRs (suggest you start with running future calls concurrently), and add them together with tests that validate the changes work as expected.

Thanks for taking the time @SandPod :)

Please clarify where you identify multiple features, in my opinion this is one single feature. The moment future calls can be run concurrently, a (configurable) limit becomes necessary too. 
First and foremost, to keep the current default behavior and second, to enable load limiting as scheduled jobs tend to be used for work-heavy tasks.

packages/serverpod_shared/lib/src/config.dart
inf0rmatix: That was not the right thread for my response xD 
Yes, I second your suggestion for unlimited power/ concurrency.

packages/serverpod/lib/src/server/future_call_manager.dart
inf0rmatix: Done :)

packages/serverpod/lib/src/server/future_call.dart
inf0rmatix: I see your reasoning, especially with a use case where we'd want to have specific FutureCall running only once across all instances because of some database manipulations for example. 

So maybe just having a per server "in total" limit of running FutureCalls to enable resource preservation would be better. 

In my current use-case, I need a specific FutureCall to be running once concurrently. To ensure this, I'd then have to implement a solution on my own. The approach I implemented would work for this use-case, until there's more then one instance. Did not think of that ^^

inf0rmatix: I removed the per FutureCall concurrency limit as I understand and follow your reasoning :)

packages/serverpod_shared/lib/src/config.dart
inf0rmatix: Done :)

packages/serverpod_shared/lib/src/environment_variables.dart
inf0rmatix: I'm not sure would exactly you mean. Similar to `ServerpodServerConfigMap.publicScheme` i.e. `ServerpodFutureCallCOnfigMap.concurrencyLimit` ?

SandPod: Yes, exactly!

Or how `sessionLogs` are configurable.

Example configuration here: https://github.com/serverpod/serverpod/blob/a6f77e6914cb25fbd9ef784e641621c931505f16/tests/serverpod_test_server/config/development.yaml#L30-L32

With implementations here:
- https://github.com/serverpod/serverpod/blob/a6f77e6914cb25fbd9ef784e641621c931505f16/packages/serverpod_shared/lib/src/environment_variables.dart#L18-L19
- https://github.com/serverpod/serverpod/blob/a6f77e6914cb25fbd9ef784e641621c931505f16/packages/serverpod_shared/lib/src/config.dart#L420

SandPod:
> Please clarify where you identify multiple features, in my opinion this is one single feature. The moment future calls can be run concurrently, a (configurable) limit becomes necessary too.

I view these as different fixes:
1. Ensure future calls can run concurrently.
2. Ensure future calls are not removed from the database without being processed.

The second one needs some special care when thinking about multiple servers interacting with the same source. I think it would be best to tackle this one in isolation.

packages/serverpod/lib/src/util/mutex.dart
```diff
@@ -0,0 +1,45 @@
+import 'dart:async';
+
+/// A mutex implementation that allows for synchronized execution of code.
+///
+/// This class provides a synchronized block that ensures that only one
+/// operation can execute at a time. It uses a list of waiting operations
+/// to manage the order of execution.
```
SandPod, line 7: Suggestion: In the server package we already depend on https://pub.dev/packages/synchronized, could we use their `Lock` for this instead?

inf0rmatix: OOooh, I did not see that. Of course :) Will do!

packages/serverpod_shared/lib/src/environment_variables.dart
inf0rmatix: If no FutureCallConfig is present, it will be null afaik. This however would enable unlimited concurrent FutureCalls. How should I handle this to ensure default behavior is maintained when no config is present?

SandPod: You can use the same approach here as for the sessionLogs, if the values are missing we assing then to the default values.

So no config: 
- `concurrencyLimit=1`
- `pollFrequency(find a better name than this (: )=5` 

Then the FutureCallManager always expects these values to be passed in.

inf0rmatix: I have implemented it so that FutureCallConfig is always present and if nothing configured by the user, current default values are set.

Also added tests :)

packages/serverpod/lib/src/server/future_call_manager.dart
```diff
@@ -88,91 +102,244 @@ class FutureCallManager {
...
     // again.
-    if (_server.serverpod.commandLineArgs.role == ServerpodRole.monolith &&
-        !_shuttingDown) {
-      // Check the queue again in 5 seconds
-      _timer = Timer(const Duration(seconds: 5), _checkQueue);
-    } else if (_server.serverpod.commandLineArgs.role ==
-        ServerpodRole.maintenance) {
+    if (isMonolith && !_shuttingDown) {
+      // Check the queue again in 1 second
+      _timer = Timer(const Duration(seconds: 1), _checkQueue);
```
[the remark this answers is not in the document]
inf0rmatix: Done :)

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -45,6 +46,12 @@ class ServerpodConfig {
   /// Default is 30 seconds.
   final Duration? experimentalDiagnosticHandlerTimeout;
 
+  /// The maximum number of concurrent running future calls.
+  /// Takes precedence over the [FutureCall.concurrentLimit] setting.
+  /// Even if a single FutureCall has a higher limit, the total number of
+  /// concurrent future calls will not exceed this limit.
+  final int futureCallConcurrencyLimit;
```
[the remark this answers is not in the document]
inf0rmatix: Done :)

inf0rmatix:
> > Please clarify where you identify multiple features, in my opinion this is one single feature. The moment future calls can be run concurrently, a (configurable) limit becomes necessary too.
> 
> I view these as different fixes:
> 
> 1. Ensure future calls can run concurrently.
> 2. Ensure future calls are not removed from the database without being processed.
> 
> The second one needs some special care when thinking about multiple servers interacting with the same source. I think it would be best to tackle this one in isolation.
> 
> Give me a ping when it is ready for review again :)

@SandPod 🏓 😄 

Please take a look at my implementation of `_tryRunFutureCallEntry`. I think the database access and deletion is straightforward and simple enough to be kept within this PR. 
Since I am using a transaction, any writes / deletes to the rows accessed will be blocked on the database until the transaction completes, making this race-condition-proof. 

What do you think?

packages/serverpod_shared/test/future_call_config_test.dart
```diff
@@ -0,0 +1,161 @@
...
+    () {
+      var serverpodConfig = '''
+apiServer:
+  port: 8080
+  publicHost: localhost
+  publicPort: 8080
+  publicScheme: http
+futureCall:
+  concurrencyLimit: 5
+  queueDelay: 2000
```
SandPod, line 53: Discussion/Suggestion: `queueDelay` makes me think about a job queue, which this is not. Maybe `scanInterval`, `checkInterval`, `runInterval` or `pollInterval`, or is there something a big more Fluttery that could be used?

I think my favorite is `scanInterval`.

packages/serverpod_shared/test/future_call_config_test.dart
```diff
@@ -0,0 +1,161 @@
+import 'package:serverpod_shared/serverpod_shared.dart';
+import 'package:test/test.dart';
+import 'package:yaml/yaml.dart';
+
+void main() {
```
SandPod, line 5: Suggestion: Add a test that only configures one of the options as well.

This asserts that we have not accidentally created an "all or nothing" solution :)

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -416,6 +430,54 @@ class RedisConfig {
   }
 }
 
+/// Configuration for future call handling.
+class FutureCallConfig {
+  /// The maximum number of concurrent running future calls. If the limit is
+  /// reached, future calls will be postponed until a slot is available.
+  ///
+  /// If the limit is set to a value <= 0, the concurrency limit is disabled.
```
SandPod, line 438: Suggestion(typo): If the limit is set to a value **<= 1**.

Otherwise zero seems like a reasonable configuration option.

packages/serverpod/lib/src/server/future_call_manager.dart
```diff
@@ -88,91 +101,263 @@ class FutureCallManager {
...
   Future<void> _checkQueue() async {
-    var pendingFutureCall = Completer<void>();
-    _pendingFutureCall = pendingFutureCall;
+    if (_shuttingDown) {
+      return;
+    }
+
+    final serverpodRole = _server.serverpod.commandLineArgs.role;
+    final isMaintenance = serverpodRole == ServerpodRole.maintenance;
+    final isMonolith = serverpodRole == ServerpodRole.monolith;
```
SandPod, line 114: Suggestion: Break this out and make different calls from the `start()` method based on the server configuration.

Feels like we can simplify the logic quite a bit here if we would separate the code path of the roles completely.

One path simply does a single check and one path instantiates a loop that executes the pull and run logic.

packages/serverpod/lib/src/server/future_call_manager.dart
```diff
@@ -88,91 +101,263 @@ class FutureCallManager {
...
     // If we are running as a maintenance task, we shouldn't check the queue
     // again.
-    if (_server.serverpod.commandLineArgs.role == ServerpodRole.monolith &&
-        !_shuttingDown) {
-      // Check the queue again in 5 seconds
-      _timer = Timer(const Duration(seconds: 5), _checkQueue);
-    } else if (_server.serverpod.commandLineArgs.role ==
-        ServerpodRole.maintenance) {
+    if (isMonolith && !_shuttingDown) {
+      final queueDelay = _server.serverpod.config.futureCall.queueDelay;
```
SandPod, line 160: Suggestion: Pass this in through the constructor to the FutureCallManager.

Same for the concurrency limit.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -416,6 +430,54 @@ class RedisConfig {
...
+
+  /// Creates a new [FutureCallConfig].
+  const FutureCallConfig({
+    this.concurrencyLimit = defaultFutureCallConcurrencyLimit,
+    this.queueDelay =
+        const Duration(milliseconds: defaultFutureCallQueueDelayMs),
+  });
+
+  /// The default concurrency limit for future calls.
+  static const int defaultFutureCallConcurrencyLimit = 1;
```
SandPod, line 452: Suggestion: Make this nullable so that if a value less than 0 is set, we set it to null.

Then we don't need to interpret "magic values" in other places that read this configuration. Null means nothing.

SandPod (changes requested):
Couple of more things to fix, but I have not reviewed the internal code of the change yet since I still think we should split the fixes for removing future calls.

> Since I am using a transaction, any writes / deletes to the rows accessed will be blocked on the database until the transaction completes, making this race-condition-proof.

Unfortunately that won't help us, that will only ensure that "this" server will modify the database in a single atomic operation. But while the transaction is ongoing, there is nothing stopping multiple servers from picking up the same future calls while the transaction is still ongoing.

Therefore, I still think this is a bit harder problem to solve. So let's tackle it in isolation and with testing.

packages/serverpod_shared/lib/src/config.dart
inf0rmatix: You mean < 1 right? Because setting it to 0 can be a reasonable option if you want to have unlimited concurrent future calls. Setting it to 1 will just disable concurrent calls. I will adjust the documentation to be more specific 🙂

packages/serverpod_shared/lib/src/config.dart
inf0rmatix: So no default value, not setting the limit will just disable concurrent future calls altogether. Same behaviour as now, but instead of 1 being default, null is.

SandPod: No, I was thinking that `null` means that the `limit` is disabled, meaning no configuration is set as a limit. But then, in our configuration, we set the default to `1`. But in the 3.0 release we remove the default `1` value.

packages/serverpod_shared/test/future_call_config_test.dart
inf0rmatix: Good point, adjusted it :)

packages/serverpod_shared/test/future_call_config_test.dart
inf0rmatix: Good catch! I actually did not know, that the _validateJsonConfig is used for required fields solely. 

I removed the call to that and added tests as you suggested :)

packages/serverpod_shared/lib/src/config.dart
inf0rmatix: aaaah, ok. But I'd leave the `defaultFutureCallConcurrencyLimit` as 1. Why change the type now then?

inf0rmatix:
> Couple of more things to fix, but I have not reviewed the internal code of the change yet since I still think we should split the fixes for removing future calls.
> 
> > Since I am using a transaction, any writes / deletes to the rows accessed will be blocked on the database until the transaction completes, making this race-condition-proof.
> 
> Unfortunately that won't help us, that will only ensure that "this" server will modify the database in a single atomic operation. But while the transaction is ongoing, there is nothing stopping multiple servers from picking up the same future calls while the transaction is still ongoing.
> 
> Therefore, I still think this is a bit harder problem to solve. So let's tackle it in isolation and with testing.

I have read up and consulted GPT about the issue, now I know better :)

Re-implemented the single `deleteWhere` alongside an explanatory TODO.

packages/serverpod/lib/src/server/future_call_manager.dart
inf0rmatix: Why not make use of the server configuration? This class already depends on the Server object, so in regards of testability, I don't see any advantage. This way, it is very clear where the value comes from. Moving this into class fields and the assignment into the constructor call would just add a (unnecessary?) level of complexity.

packages/serverpod/lib/src/server/future_call_manager.dart
inf0rmatix: Love the idea! I adjusted accordingly :)

packages/serverpod/lib/src/server/future_call_manager.dart
```diff
@@ -88,91 +112,244 @@ class FutureCallManager {
...
 
-      var tempSession = _server.serverpod.internalSession;
+    _timer = Timer(scanInterval, _checkQueue);
+  }
 
-      var rows = await FutureCallEntry.db.deleteWhere(
-        tempSession,
-        where: (t) => t.time <= now,
-      );
+  Future<void> _checkQueue() async {
```
SandPod, line 140: Implementation suggestion: Separate scanning for tasks from executing them.

Let's say we have two parts of the system, `TaskScanner` and `TaskScheduler`.

The responsibility of the `TaskScanner` is to scan the database with a certain interval and add tasks to the `TaskScheduler`.

The responsibility of the `TaskScheduler` is to execute all scheduled tasks with respect to a given concurrency limit.

If the `TaskScheduler` is at max capacity when the `TaskScanner` scan interval is up, it will not add any `Tasks`.

When the server is stopped the `TaskScanner` is stopped and we wait for the `TaskScheduler` to empty its queue.

I think this separation will greatly simplify our around this problem and probably make the code a bit easier to read.

This is a GPT(So not to be trusted but a source of inspiration) suggestion of a how a `TaskScheduler` could look:

```dart
class TaskScheduler {
  final int maxConcurrentTasks;
  final Queue<Future<void> Function()> _taskQueue = Queue();
  int _runningTasks = 0;

  TaskScheduler({this.maxConcurrentTasks = 10});

  // Call this to schedule a task
  void schedule(Future<void> Function() task) {
    _taskQueue.add(task);
    _tryRunNext();
  }

  void _tryRunNext() {
    // If we have slots and tasks to run
    while (_runningTasks < maxConcurrentTasks && _taskQueue.isNotEmpty) {
      final task = _taskQueue.removeFirst();
      _runningTasks++;

      // Run the task and handle completion
      task().whenComplete(() {
        _runningTasks--;
        _tryRunNext(); // Try running next task after this finishes
      });
    }
  }
}
```

SandPod (changes requested):
Couple of more comments, feel free to reach out if you have any more questions :)

packages/serverpod_shared/lib/src/environment_variables.dart
```diff
@@ -177,6 +199,10 @@ enum ServerpodEnv {
       (ServerpodEnv.webPublicPort) => 'SERVERPOD_WEB_SERVER_PUBLIC_PORT',
       (ServerpodEnv.webPublicScheme) => 'SERVERPOD_WEB_SERVER_PUBLIC_SCHEME',
       (ServerpodEnv.maxRequestSize) => 'SERVERPOD_MAX_REQUEST_SIZE',
+      (ServerpodEnv.futureCallConcurrencyLimit) =>
+        'SERVERPOD_FUTURE_CALL_CONCURRENCY_LIMIT',
+      (ServerpodEnv.futureCallQueueDelay) =>
+        'SERVERPOD_FUTURE_CALL_QUEUE_DELAY',
```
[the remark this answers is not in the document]
inf0rmatix: image

packages/serverpod/lib/src/server/future_call_manager.dart
```diff
@@ -88,91 +112,244 @@ class FutureCallManager {
...
         }
 
-        dynamic object;
-        if (entry.serializedObject != null) {
-          object = _serializationManager.decode(
-              entry.serializedObject!, call.dataType);
+        // If we are shutting down, stop processing future calls.
+        if (_shuttingDown) {
+          break;
         }
```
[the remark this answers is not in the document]
inf0rmatix: ah ye, was a left over from when I wanted to only delete a FutureCall once it is actually invoked.

packages/serverpod/lib/src/server/future_call_manager.dart
inf0rmatix: done

packages/serverpod_shared/lib/src/environment_variables.dart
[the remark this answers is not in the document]
inf0rmatix: done

packages/serverpod/lib/src/server/future_call_manager.dart
```diff
@@ -88,91 +112,244 @@ class FutureCallManager {
...
+  }) async {
+    var internalSession = _server.serverpod.internalSession;
+
+    var overdueFutureCalls = List<FutureCallEntry>.from(
+      futureCallEntries ??
+          await FutureCallEntry.db.deleteWhere(
+            internalSession,
+            where: (row) => row.time <= due,
+          ),
+    )..sort((a, b) => a.time.compareTo(b.time));
```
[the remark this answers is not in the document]
inf0rmatix: deleteWhere has no sort parameter 🙈

packages/serverpod/lib/src/server/future_call_manager/future_call_queue.dart
```diff
@@ -0,0 +1,233 @@
...
+    futureCall.initialize(_server, name);
+    _futureCalls[name] = futureCall;
+  }
+
+  /// Stops the [FutureCallQueue].
+  ///
+  /// This method will wait for all running future calls to complete.
+  /// This will discard any queued future calls.
+  Future<void> stop() async {
+    await Future.wait(_runningFutureCallFutures);
```
SandPod, line 56: Suggestion: This will only wait for the `ongoing` future calls and give incorrect results.

If we have a long queue, more entries might be added after we have called for this wait and we would return from this method even if there are still running future calls.

We need another mechanism to ensure that all future calls have been processed before shutting down.

Ideally we would like to know when the number of running tasks AND the number of queued tasks is 0.
And during this time, the queue should probably not accept more entries.

Perhaps having a `stop` on the queue is the wrong aproach? Maybe an simple `drain()` method is more fitting? Since stopping is a collaboration with what is using the queue (ensuring that no new tasks are scheduled).

packages/serverpod/lib/src/server/future_call_manager/future_call_queue.dart
```diff
@@ -0,0 +1,233 @@
...
+          // Future Call not found, ignore.
+          // TODO this should be logged as an error.
+          continue;
+        }
+
+        final result = await _tryRunFutureCallEntry(
+          session: _server.serverpod.internalSession,
+          futureCallEntry: futureCallEntry,
+          futureCall: futureCall,
+        );
```
SandPod, line 132: Suggestion: Before this line, increase `_runningFutureCalls`, then call `_runFutureCall` directly with a `whenComplete` block that decreases `_runningFutureCalls` and calls `handleQueue`.

This allows us to remove the `_tryRunFutureCallEntry` middle step.

This also removes the need for having this as a synchronized block and saving futures in `_runningFutureCallFutures` since the `_runningFutureCalls` will tell us the number of running future calls (since it is modified in sync)

packages/serverpod/lib/src/server/future_call_manager/future_call_queue.dart
```diff
@@ -0,0 +1,233 @@
...
+import 'package:serverpod/src/server/serverpod.dart';
+import 'package:synchronized/synchronized.dart';
+
+enum _FutureCallInvocationResult {
+  success,
+  postponed;
+}
+
+/// A class responsible for scheduling and executing future call entries.
+class FutureCallQueue {
```
SandPod, line 17: Suggestion(style): If we add methods like `drain()` instead of `stop()`, I would suggest renaming this to `FutureCallScheduler` again, since it is not only a simple `Queue` but actually executes the tasks as well.

Seems like that is the most common term used for something that schedules and executes tasks.

SandPod (changes requested):
The split now made this much easier to reason around, thank you 🙏

Still a couple of issues to address to ensure this fuctions correctly.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -416,6 +433,75 @@ class RedisConfig {
...
+    int? concurrencyLimit;
+
+    if (hasConcurrencyLimitKey && concurrencyLimitJson != null) {
+      try {
+        concurrencyLimit = concurrencyLimitJson is int
+            ? concurrencyLimitJson
+            : int.parse(concurrencyLimitJson);
+      } catch (_) {
+        throw Exception(
+          'The `concurrencyLimit` setting was set to $concurrencyLimitJson, but must be an integer.',
```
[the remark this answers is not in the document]
inf0rmatix: done

packages/serverpod_shared/test/future_call_config_test.dart
```diff
@@ -0,0 +1,313 @@
...
+    'Given a negative concurrencyLimit when loading from Map then throws Exception',
+    () {
+      var serverpodConfig = '''
+apiServer:
+  port: 8080
+  publicHost: localhost
+  publicPort: 8080
+  publicScheme: http
+futureCall:
+  concurrencyLimit: -1
```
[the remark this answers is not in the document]
inf0rmatix: ok. got it :)

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -416,6 +433,75 @@ class RedisConfig {
...
+        );
+      }
+
+      if (concurrencyLimit < 1) {
```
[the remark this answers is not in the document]
inf0rmatix: done

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -545,6 +631,17 @@ Map? _buildSessionLogsConfigMap(
   ]);
 }
 
+Map? _buildFutureCallConfigMap(Map configMap, Map<String, String> environment) {
+  var futureCallConfig = configMap[ServerpodConfigMap.futureCall] ?? {};
+
+  return _buildConfigMap(futureCallConfig, environment, [
+    // passing null to the convert function will not convert the value, since
+    // the config value itself can be null
+    (ServerpodEnv.futureCallConcurrencyLimit, null),
```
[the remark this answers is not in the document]
inf0rmatix: done! Since the user themself is allowed to pass negative values or just none this made the whole thing easier and allowed to have the whole parsing done here. Nice :)

packages/serverpod/lib/src/server/future_call_manager/future_call_queue.dart
inf0rmatix: i think I'd just remove the other method and for the time being, move the logic from there into here. There I already implemented what you suggested. Idk why I did two. ^^

packages/serverpod/lib/src/server/future_call_manager/future_call_queue.dart
inf0rmatix: Removing the middle step, I agree with.
Removing the _runningFutureCalls, I'd rather not.

Because having this list of futures allows the stop method to be better implemented, using Future.any instead of active polling to run _handleQueue again.

packages/serverpod/lib/src/server/future_call_manager/future_call_queue.dart
inf0rmatix: done

inf0rmatix:
@SandPod I also went ahead and created some tests for the Manager, added mocktail to dev dependencies in the process. Feel free to adjust the tests to your liking. Might also do some for the Scheduler later. There's these times where I actually enjoy writing tests, strange, isn't it? ^^

packages/serverpod/test/server/future_call_manager/future_call_manager_test.dart
```diff
@@ -0,0 +1,125 @@
...
+
+class MockFutureCall extends Mock implements FutureCall<SerializableModel> {}
+
+abstract class OnCompleted {
+  void call();
+}
+
+class MockOnCompleted extends Mock implements OnCompleted {}
+
+void main() {
```
SandPod, line 28: Suggestion: Remove or reformat these tests to not know about the implementation.

These tests are now written in a way that we have locked the implementation of FutureCallManager.

If we would refactor the implementation with new internal classes all these tests would break and we wouldn't have any guarantees that previously supported functionality would continue to function.

The importance for tests towards the FutureCallManager is not how we set up the structure or what specific methods are called, but that the public API gives the results we are expecting.

The external API is:
- scheduleFutureCall (Object is added to database)
- cancelFutureCall (Object is removed from database)
- registerFutureCall (With current implementation, should be handled vs not handled when task is picked up from database)
- start (Process tasks in database, either continously or once)
- stop (Stops processing tasks in database even if new are added)

Test should be written in a way that the result from calling these are validated without knowing the internal code structure. I have given examples on how they could be validated.

SandPod: Mock are usually a bad sign that we either have strctured our code in a bad way or that we are testing on the wrong level.

packages/serverpod/lib/src/server/future_call_manager/future_call_scanner.dart
```diff
@@ -0,0 +1,89 @@
...
+/// Scans the database for overdue future calls and queues them for execution.
+class FutureCallScanner {
+  final Server _server;
+
+  Timer? _timer;
+
+  final Duration _scanInterval;
+
+  final IsConcurrentLimitReached _isConcurrentLimitReached;
+  final QueueFutureCallEntries _queueFutureCallEntries;
```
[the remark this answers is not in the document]
SandPod: Note that these comments are in the `future_call_scanner.dart`. I think you modified the wrong file.

The `Scanner` shouldn't know about the `Scheduler` and vice versa. Now the `Scheduler` has a `shouldSkipScan` that means that it knows about the `Scanner`.

packages/serverpod/test/server/future_call_manager/future_call_manager_test.dart
SandPod: The `forTesting` method should also be removed.

packages/serverpod/lib/src/server/future_call_manager/future_call_scheduler.dart
```diff
@@ -0,0 +1,191 @@
...
+      ),
+    );
+
+    unawaited(_handleQueue());
+  }
+
+  /// Handles the queue of future call entries.
+  /// While the concurrency limit is not reached, the queue is processed.
+  /// If the concurrency limit is reached, the queue is not processed further.
+  Future<void> _handleQueue() async {
```
SandPod, line 102: Suggestion: This method doesn't need to be async.

The `synchronized` block doesn't contain any `await` statements. This means that all actions inside of the sync block are executed in sync. Therefore, I don't think this is needed?

If the `lock` is removed, then this method doesn't need to be `async` anymore either better clarifying it's intent.

packages/serverpod/lib/src/server/future_call_manager/future_call_scheduler.dart
```diff
@@ -0,0 +1,191 @@
...
+    futureCall.initialize(_server, name);
+    _futureCalls[name] = futureCall;
+  }
+
+  /// Makes sure any running future calls and queued future calls are processed.
+  /// Once called, [shouldSkipScan] will always return `true`.
+  Future<void> stop() async {
+    _isStopping = true;
+
+    await Future.any(_runningFutureCallFutures.toList());
```
SandPod, line 53: Question: Isn't there a risk that when this reports back, a new task has already been scheduled emptying the queue and making us return early even though there are still entries being processed?

Since we are using `any` here. we will continue as soon as a single entry has been processed. At that point the `_queue` might be empty but we still have multiple FutureCall futures running. I think we still have this problem even if we used `Future.wait` (this would also introduce other issues).

To centralize this, wouldn't it be better to use a single `Completer` for the class that is set when someone calls `stop(or drain)` as I suggest.

Then this `Completer` is completed when we have no tasks running and the queue is empty. This could be checked inside of `_handleQueue` since that method is called as soon as a future call is processed.

packages/serverpod/lib/src/server/future_call_manager/future_call_scheduler.dart
```diff
@@ -0,0 +1,191 @@
...
+      await Future.any(_runningFutureCallFutures.toList());
+    }
+  }
+
+  /// Returns `true` if the concurrent limit for FutureCalls is reached.
+  /// Returns `false` otherwise.
+  bool shouldSkipScan() {
+    if (_isStopping) {
+      return true;
+    }
```
SandPod, line 67: Suggestion: Remove `_isStopping` from here and instead throw if new tasks are scheduled while the scheduler is stopping.

packages/serverpod/lib/src/server/future_call_manager/future_call_scheduler.dart
SandPod: This would also remove our need of tracking the individual calls.

SandPod (changes requested):
Super happy you thought about tests.

These need to be modified so that they are not implementation dependent.

There are also still some issues with the current implementation.

packages/serverpod/lib/src/server/future_call_manager/future_call_scanner.dart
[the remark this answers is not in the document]
inf0rmatix: done

packages/serverpod/lib/src/server/future_call_manager/future_call_scheduler.dart
inf0rmatix: I feel like your approach makes it more complex than required, right now, if the queue would be empty, we'd just wait for the next to finish again. I added a final Future.wait which was missing. Once the queue is emptied, we just need to wait for all remaining future calls.

I used .toList to avoid concurrent modification exceptions which can arise, if the list of futures is modified by a future call completing while using future.wait or .any.

packages/serverpod/lib/src/server/future_call_manager/future_call_scheduler.dart
inf0rmatix: yes.

inf0rmatix: There is a possible race condition. 

If the timer is being invoked while stop is called, the scheduler will already be shutting down after the entries are deleted from database and before _dispatchEntries is called. Which would lead to an exception being thrown and the entries being completely lost.

For simplicity sake, I would not throw an exception in scheduler. 
Otherwise, we'd have to implement some mechanic to prevent this race condition.

packages/serverpod/lib/src/server/future_call_manager/future_call_scheduler.dart
inf0rmatix: This also would de-centralize the logic from the stop method and spread it into the handleQueue. I think this would make it harder to understand and find when further adjusting this code.

packages/serverpod/lib/src/server/future_call_manager/future_call_diagnostics_service.dart
```diff
@@ -0,0 +1,60 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod/src/server/diagnostic_events/diagnostic_events.dart';
+import 'package:serverpod/src/server/serverpod.dart';
+
+/// Interface for reporting diagnostic events related to future calls.
+///
+/// This interface should be generalized to allow any module library to report
+/// diagnostic events without directly depending on serverpod server.
+abstract interface class FutureCallDiagnosticsService {
```
inf0rmatix, line 9: That's nice! `*.*`

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -0,0 +1,182 @@
...
+import 'dart:io';
+
+import 'package:serverpod/protocol.dart';
+import 'package:serverpod/serverpod.dart';
+
+import 'future_call_diagnostics_service.dart';
+import 'future_call_scanner.dart';
+import 'serverpod_task_scheduler.dart';
+
+typedef SessionProvider = Session Function(String futureCallName);
```
inf0rmatix, line 11: isn't this more like a builder? To me, provider naming is associated heavily with the same name library for flutter?

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -0,0 +1,182 @@
...
+        _initializeFutureCall = initializeFutureCall {
+    _scheduler = ServerpodTaskScheduler(
+      concurrencyLimit: _config.concurrencyLimit,
+    );
+
+    _scanner = FutureCallScanner(
+      internalSession: _internalSession,
+      scanInterval: _config.scanInterval,
+      shouldSkipScan: _scheduler.isConcurrentLimitReached,
+      dispatchEntries: (entries) => _dispatchEntries(entries),
```
inf0rmatix, line 57: this could become a tear-off

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -0,0 +1,182 @@
...
+        );
+      }
+
+      await futureCall.invoke(futureCallSession, object);
+      await futureCallSession.close();
+    } catch (error, stackTrace) {
+      _diagnosticsService.submitCallException(
+        error,
+        stackTrace,
+        session: session,
```
inf0rmatix, line 176: I get the feeling that handling two separate sessions makes this method a bit confusing to read. Would it be feasible to remove the session parameter from the submitCallException method and leave the session retrieval to the implementation of `FutureCallDiagnosticsService`?

inf0rmatix: maybe we could also just pass the future call session here, since the event is tied to the future call

packages/serverpod/lib/src/server/future_call_manager/serverpod_task_scheduler.dart
```diff
@@ -0,0 +1,94 @@
...
+
+  Completer<void>? _stoppingCompleter;
+
+  /// Creates a new [ServerpodTaskScheduler].
+  ServerpodTaskScheduler({
+    required int? concurrencyLimit,
+  }) : _concurrencyLimit = concurrencyLimit;
+
+  /// Makes sure any running task callbacks and queued task callbacks are processed.
+  /// Once called, [isConcurrentLimitReached] will always return `true`.
```
inf0rmatix, line 24: this doc comment is out of date, oversight of mine

inf0rmatix: " Once called, [isConcurrentLimitReached] will always return `true`." is not correct anymore

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -592,7 +604,18 @@ class Serverpod {
...
       // Start future calls
       _completedFutureCalls = _futureCallManager == null;
-      _futureCallManager?.start();
+      if (!config.futureCallExecutionEnabled) {
+        logVerbose('Future call execution is disabled.');
+        _completedFutureCalls = true;
+      } else if (commandLineArgs.role == ServerpodRole.maintenance) {
+        unawaited(
+          _futureCallManager
+              ?.runScheduledFutureCalls()
```
inf0rmatix, line 613: love the approach. mabye this whole block about future calls could be moved into a method. Which would enable using guard clause  style instead of nested else if for readability

packages/serverpod_shared/test/config_test.dart
```diff
@@ -1093,4 +1093,96 @@ serverId: testServerIdFromConfig
...
+      serverId,
+      passwords,
+      loadYaml(serverpodConfig),
+    );
+
+    expect(config.futureCallExecutionEnabled, isTrue);
+  });
+
+  test(
+      'Given a Serverpod config with futureCallExecutionEnabled set to true when loading from Map then the futureCallExecutionEnabled is set to false.',
```
inf0rmatix, line 1119: this description is wrong, it should be "Given a Serverpod config with futureCallExecutionEnabled set to **false**

tests/serverpod_test_server/test_integration/future_calls/future_call_manager_test.dart
```diff
@@ -0,0 +1,74 @@
...
+    } else {
+      session.log('null');
+    }
+
+    completer.complete();
+  }
+}
+
+void main() async {
+  withServerpod(
```
inf0rmatix, line 28: that's lovely syntax sugar :3

tests/serverpod_test_server/test_integration/future_calls/future_call_manager_test.dart
```diff
@@ -0,0 +1,74 @@
...
+      await futureCallManager.stop();
+    });
+
+    group('when running scheduled future calls', () {
+      setUp(() async {
+        await futureCallManager.runScheduledFutureCalls();
+      });
+
+      test('then call is executed', () async {
+        expect(testCall.completer.future, completes);
```
inf0rmatix, line 62: uh nice, did not know the "completes" one yet!

packages/serverpod/lib/src/server/future_call_manager/future_call_scanner.dart
```diff
@@ -0,0 +1,110 @@
...
+
+  /// Scans the database for overdue future calls and queues them for execution.
+  Future<void> scanFutureCallEntries() async {
+    if (_isStopping || _shouldSkipScan()) {
+      return;
+    }
+
+    // Wait for any previous scan to complete. This is to ensure that a single
+    // scan is running at a time.
+    await _scanCompleter.future;
```
inf0rmatix, line 75: instead of await, return of completer is not completed. Do the check in line with the existing guard clause at the top

inf0rmatix: done

packages/serverpod_shared/test/config_test.dart
inf0rmatix: done

packages/serverpod/lib/src/server/serverpod.dart
inf0rmatix: idk what i was thinking anymore XD let's leave it as is for now

packages/serverpod/lib/src/server/future_call_manager/serverpod_task_scheduler.dart
inf0rmatix: done

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
inf0rmatix: done

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
SandPod: Yes, that sounds resonable. Feel free to rename it to an appropriate name, such as `FutureCallSessionBuilder`.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
SandPod: The implementation here is incorrect. I think we should only use the `futureCallSession`. So the `session` parameter should be removed from `_runFutureCall`.

This is a mistake on my part.

tests/serverpod_test_server/lib/src/futureCalls/test_call.dart
```diff
@@ -1,7 +1,11 @@
+import 'dart:async';
+
 import 'package:serverpod/serverpod.dart';
 import 'package:serverpod_test_server/src/generated/protocol.dart';
 
 class TestCall extends FutureCall<SimpleData> {
+  final Completer<void> completer = Completer<void>();
```
SandPod, line 7: Suggestion: Changes in this class can be reverted.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -0,0 +1,182 @@
...
+    await _scanner.stop();
+    await _scheduler.stop();
+  }
+
+  void _dispatchEntries(List<FutureCallEntry> entries) {
+    final callbacks = entries.map<TaskCallback?>((entry) {
+      final futureCall = _futureCalls[entry.name];
+
+      if (futureCall == null) {
+        // TODO(inf0rmatix): this should be logged or caught otherwise.
```
SandPod, line 139: Suggestion: Create and issue out of this and remove the user specific TODO.

This gives anyone the possibility to resolve the issue.

packages/serverpod/lib/src/server/future_call_manager/future_call_scanner.dart
```diff
@@ -0,0 +1,109 @@
...
+import 'future_call_diagnostics_service.dart';
+
+/// A function that queues future call entries for execution.
+typedef DispatchEntries = void Function(List<FutureCallEntry> entries);
+
+/// A function that checks if the concurrent limit has been reached. Returns
+/// `true` if the limit has been reached, `false` otherwise.
+typedef ShouldSkipScan = bool Function();
+
+/// Scans the database for overdue future calls and queues them for execution.
```
SandPod, line 16: Suggestion: Change description to. `Scans the database for overdue future calls and dispatches them.`

packages/serverpod/lib/src/server/future_call_manager/future_call_scanner.dart
```diff
@@ -0,0 +1,109 @@
...
+        _shouldSkipScan = shouldSkipScan,
+        _dispatchEntries = dispatchEntries,
+        _diagnosticReporting = diagnosticsService;
+
+  /// Starts the task scanner, which will scan the database for overdue future
+  /// calls at the given interval.
+  void start() {
+    if (_timer != null) {
+      throw StateError('Future call scanner already started.');
+    }
```
SandPod, line 50: Suggestion: Don't throw if start is called multiple times.

I think this changes the behavior of the Server. I'm not sure but I think it was possible to all start multiple times before without anything throwing.

packages/serverpod/lib/src/server/future_call_manager/future_call_scanner.dart
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  /// Scans the database for overdue future calls and queues them for execution.
+  Future<void> scanFutureCallEntries() async {
+    if (_isStopping || _shouldSkipScan() || !_scanCompleter.isCompleted) {
+      return;
+    }
+
+    // Wait for any previous scan to complete. This is to ensure that a single
+    // scan is running at a time.
```
SandPod, line 74: Suggestion: Remove outdated comment.

SandPod (changes requested):
Moving closer.

But I think the main thing now is building out the test suite that we set up so that it covers the external API that we talked about before here: https://github.com/serverpod/serverpod/pull/3326#discussion_r2030834017.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
inf0rmatix: [I will link to this issue](https://github.com/serverpod/serverpod/issues/3485)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
inf0rmatix: done

inf0rmatix:
Once merged, documentation should be extended for the new configuration parameters

inf0rmatix:
TODO add toString output for futureCallConfig in ServerpodConfig.toString

SandPod (approved):
Added a couple of additional tests to ensure that the new configuration works as expected and that future calls are processed accordingly.

Issue #3286, opened 2025-03-05
db.updateRow is setting a private field to null.


Draice:
### Describe the bug

I've got this method in my "zone_endpoint.dart" file:

```
/// Edits a given zone.
  Future<Zone?> editZone(
      Session session, int zoneId, String name, String description) async {
    var zone = await Zone.db.findById(session, zoneId);

    // If the zone exists, update its properties.
    if (zone != null) {
      // Preserve the realmId
      final realmId = zone.realmId;

      zone.name = name;
      zone.description = description;
      zone.realmId = realmId; // Ensure realmId is preserved

      await Zone.db.updateRow(session, zone);
    }

    return zone;
  }
```

These are my two model files:

```
class: Zone
table: zone
fields:
  name: String
  description: String
  locations: List<Location>?, relation
  combatEncounters: List<CombatEncounter>?, relation
  realmId: int?, relation(parent=realm, onDelete=Cascade)
```

and

```
class: Realm
table: realm
fields:
  name: String
  description: String
  zones: List<Zone>?, relation
  worldId: int?, relation(parent=world, onDelete=Cascade)
```

This is the before screen shot when I update the name and description of 'Zone B' to 'Zone C'
![Image](https://github.com/user-attachments/assets/c1c8981c-3c20-4411-8fca-7671c7535a36)

and after:
![Image](https://github.com/user-attachments/assets/0da9482b-e59f-4e16-b781-2b299bc597b9)

Notice that while the foreign key is preserved, there's what appears to be a private field "_realmZonesRealmId" that is getting set to null. When I restart my app, it is no longer able to fine the renamed "Zone C".

### To reproduce

1. Use the provided snippets above to create a Zone.spy and Realm.spy file. 
2. Create a zone_endpoint.dart with the method provided above
3. Call that method from a test app
4. Notice that the _realmZonesRealmId field is being set to null.

### Expected behavior

_realmZonesRealmId does not get set to null, it retains its original value.

### Library version

Serverpod 2.3.1

### Platform information

Flutter Web

### How experienced are you with this library?

Beginner - Just getting started with this library

### Are you interested in working on a PR for this?

- [ ] I want to work on this

SandPod:
Hey @Draice!

Thank for the bug report!

### Where _realmZonesRealmId comes from

The reason that there is private field `_realmZonesRealmId` stored in the Zones table is that you are configuring a one-to-many relationship from `Realm` (one) to `Zones` (many) through the `Realm.zones` field.

[...]

PR #3327, opened 2025-03-20
fix: Prevent query exception from id column in update queries.

SandPod:
This PR fixes two exceptions that could occur when performing an update query.

1. Prevents update query to throw if an update query was performed on a model without any fields.
2. Prevents update query to throw if the `id` column was specified as a column to be updated.

Found when exploring the fix for #3286 

## Pre-launch Checklist

## Breaking changes

None - simple stability fix.

### DIFF
--- a/packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
+++ b/packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
@@ -187,7 +187,10 @@ class DatabaseConnection {
 
     if (columns != null) {
       _validateColumnsExists(columns, table);
-      selectedColumns = [table.id, ...columns];
+      selectedColumns = [
+        ...columns,
+        if (!columns.contains(table.id)) table.id,
+      ];
     }
 
     var selectedColumnNames = selectedColumns.map((e) => e.columnName);
--- a/packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
+++ b/packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
@@ -198,7 +198,6 @@ class DatabaseConnection {
     var values = _createQueryValueList(rows, selectedColumns);
 
     var setColumns = selectedColumnNames
-        .where((columnName) => columnName != 'id')
         .map((columnName) => '"$columnName" = data."$columnName"')
         .join(', ');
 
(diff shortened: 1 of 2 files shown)


packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -187,7 +187,10 @@ class DatabaseConnection {
 
     if (columns != null) {
       _validateColumnsExists(columns, table);
-      selectedColumns = [table.id, ...columns];
+      selectedColumns = [
+        ...columns,
+        if (!columns.contains(table.id)) table.id,
+      ];
```
nielsenko, line 193: Would it make sense to use a set instead? I suppose selectedColumns should never contain duplicates.

nielsenko (approved):
LGTM. Could consider to use sets instead of lists for columns, but that is likely a larger change.

I'm not familiar with this part of the code, so you may want a second opinion.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
SandPod: Good point! I think a set would be a better data structure here 👍

SandPod: Fixed here: [cc1e27d](https://github.com/serverpod/serverpod/pull/3327/commits/cc1e27d5bfdabcf9a145ba80d14622c57d055026)

We can't change the input since that would be a breaking change. But we can modify the internal representation.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -674,7 +671,7 @@ class DatabaseConnection {
     return resolvedListRelations;
   }
 
-  void _validateColumnsExists(List<Column> columns, Table table) {
+  void _validateColumnsExists(Set<Column> columns, Table table) {
     for (var column in columns) {
       if (!table.columns.any((c) => c.columnName == column.columnName)) {
```
nielsenko, line 676: Double iterations like these always makes me a little vary. I would like `table.columns` to be a set as well. Then this n-square loop could be replaced with a linear one (assuming a HashSet). Actually `A.intersection(B) == B` can be answered in log(n) time depending on the set implementation.

For most practical purposes it probably don't matter as the sets are small, but perhaps something to consider for later.

nielsenko (approved):
LGTM. It did give rise for another comment 😄

nielsenko (approved):
LGTM 🚀

Issue #3333, opened 2025-03-21
Wasm builds are broken when client or sessino manager are initialized in main


KristijanZic:
### Describe the bug

Running a Serverpod project with Serverpod auth in wasm throws the following error if the Client or SessionManager is instantiated in the main method.
```sh
flutter run -d chrome --wasm --debug --verbose
```

ERROR:
```sh
Unsupported operation: Platform._version
main.dart.mjs:60     at module0.Error._throwWithCurrentStackTrace (http://localhost:56374/main.dart.wasm:wasm-function[822]:0x2a28cf)
    at module0._Platform._version (http://localhost:56374/main.dart.wasm:wasm-function[27455]:0x4b76c5)
    at module0._Platform.version (http://localhost:56374/main.dart.wasm:wasm-function[27454]:0x4b76b6)
    at module0.Platform.version initializer (http://localhost:56374/main.dart.wasm:wasm-function[27453]:0x4b76a6)
    at module0.Platform.version initializer (http://localhost:56374/main.dart.wasm:wasm-function[27452]:0x4b769e)
    at module0._getHttpVersion (http://localhost:56374/main.dart.wasm:wasm-function[27451]:0x4b7685)
    at module0.new _HttpClient (initializer) (http://localhost:56374/main.dart.wasm:wasm-function[27445]:0x4b7600)
    at module0._HttpClient (http://localhost:56374/main.dart.wasm:wasm-function[27444]:0x4b7568)
```

### To reproduce

Test this repository:
https://github.com/KristijanZic/wasmpod/blob/master/wasmpod_server/README.md

### Expected behavior

Wasm builds should just work.

### Library version

2.4.0

### Platform information

wasm

### How experienced are you with this library?

None

### Are you interested in working on a PR for this?

- [ ] I want to work on this

SandPod:
Fixed by https://github.com/serverpod/serverpod/pull/3334

PR #3334, opened 2025-03-21
fix: Wasm support in serverpod client

KristijanZic:
Wasm build is broken because serverpod client assumes `dart.library.js` is available in wasm builds but it's not. We should use `dart.library.js_interop` instead which doesn't break the old web users but is now also passing for wasm.

This PR fixes the issue #3333

## Pre-launch Checklist

## Breaking changes


nielsenko:
LGTM 🚀 Thanks for contributing

nielsenko (approved):

KristijanZic:
Any time, thank you! 😃

Issue #3338, opened 2025-03-22
Incorrect parsing of default values for enums


vanlooverenkoen:
### Describe the bug

Enum values might change over time, that is why it is important to have a default value configured so your client parsing won't fail.

The issue is that it will always throw this:

```
throw ArgumentError('Value "$name" cannot be converted to "YourEnum"');
```

Even if you have configured your model like this

```
class: MyModel
fields:
  uuid: String
  categoryIconKey: CategoryIconKey, default=other
```

My enum:

```
enum: CategoryIconKey
serialized: byName
values:
  - sports
  - education
  - government
  - other
```

I create a release with these values and 2 releases later I have a new CategoryIconKey that I want to support! My icons are svg assets that are always hardcoded in the app. (and only available from that new release)
The main reason we do it with an enum and not an api call that just returns the icon from a url has 2 reasons:
- faster loading of the assets
- but most of all we have a lot of extra logic to show the correct detail screen and a completely new detail screen based on that category.

You could compare it with a `pushNotificationType` where you want to do custom logic for deeplinking based on the type of your notification.

### To reproduce

My Model
```
class: MyModel
fields:
  uuid: String
  categoryIconKey: CategoryIconKey, default=other
```

My enum:
```
enum: CategoryIconKey
serialized: byName
values:
  - sports
  - education
  - government
  - other
```

use these tests

```
import 'package:test/test.dart';

void main() {
    test('parse from json with valid value', () {
      final value = 'sports';
      final result = CategoryIconKey.fromJson(value);
      expect(result, CategoryIconKey.sports);
    });

    test('Should fallback to other, () {
      final value = 'new-type';
      final result = CategoryIconKey.fromJson(value);
      expect(result, CategoryIconKey.other);
    });

    test('Should fallback to other', () {
      final myModel = MyModel.fromJson({
        'uuid': 'uuid',
        'categoryIconKey': 'new-image-type',
      });
      expect(myModel.categoryIconKey, CategoryIconKey.other);
    });
}
```

### Expected behavior

# Solution 1

This is not ideal because it would generate an optional, which will give other issues

This is what should be generated by the enum
```
  static CategoryIconKey? fromJson(String name) {
    switch (name) {
      case 'sports':
        return CategoryIconKey.sports;
      case 'education':
        return CategoryIconKey.education;
      case 'government':
        return CategoryIconKey.government;
      case 'other':
        return CategoryIconKey.other;
      default:
        return null;
    }
  }
```

[...]

PR #3341, opened 2025-03-22
feat: Global default enum value to prevent breaking the client

vanlooverenkoen:
Implements https://github.com/serverpod/serverpod/issues/3338 

By adding a new property in the enum fields

Just use the value name
```
enum: Example
serialized: byName
default: unknown
values:
  - unknown
  - customValue1
  - customValue2
```

With byIndex you still use the value name
```
enum: Example
serialized: byIndex
default: unknown
values:
  - unknown
  - customValue1
  - customValue2
```

## Pre-launch Checklist

## Breaking changes

Everything is backwards compatible

### DIFF
--- a/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
@@ -614,7 +614,7 @@ class ModelParser {
   ) {
     final defaultValue =
         _parseDefaultValue(documentContents, Keyword.defaultKey);
-    return values.where((value) => value.name == defaultValue).firstOrNull;
+    return ProtocolEnumValueDefinition(defaultValue);
   }
 
   static List<ProtocolEnumValueDefinition> _parseEnumValues(
--- a/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
@@ -614,9 +614,7 @@ class ModelParser {
   ) {
     final defaultValue =
         _parseDefaultValue(documentContents, Keyword.defaultKey);
-    if (defaultValue == null) return null;
-    if (defaultValue is! String) return null;
-    return ProtocolEnumValueDefinition(defaultValue);
+    return values.where((value) => value.name == defaultValue).firstOrNull;
   }
 
   static List<ProtocolEnumValueDefinition> _parseEnumValues(
--- a/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
@@ -182,10 +182,8 @@ class ModelParser {
       extraClasses: [],
     );
     var defaultValue = _parseDefaultValue(documentContents, Keyword.defaultKey);
-    print(defaultValue);
     var defaultEnumDefinitionValue =
         values.where((value) => value.name == defaultValue).firstOrNull;
-    print(defaultEnumDefinitionValue);
 
     var enumDef = EnumDefinition(
       fileName: outFileName,
--- a/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
@@ -181,6 +181,11 @@ class ModelParser {
       '${protocolSource.moduleAlias}:$className',
       extraClasses: [],
     );
+    var defaultValue = _parseDefaultValue(documentContents, Keyword.defaultKey);
+    var defaultEnumDefinitionValue =
+        values.where((value) => value.name == defaultValue).firstOrNull;
+    print(defaultValue);
+    print(defaultEnumDefinitionValue);
 
     var enumDef = EnumDefinition(
       fileName: outFileName,
@@ -189,6 +194,7 @@ class ModelParser {
       values: values,
       serialized: serializeAs,
       documentation: enumDocumentation,
+      defaultValue: defaultEnumDefinitionValue,
       subDirParts: protocolSource.subDirPathParts,
       serverOnly: serverOnly,
       type: enumType,
--- a/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
@@ -615,6 +615,7 @@ class ModelParser {
     final defaultValue =
         _parseDefaultValue(documentContents, Keyword.defaultKey);
     if (defaultValue == null) return null;
+    if (defaultValue is! String) return null;
     return ProtocolEnumValueDefinition(defaultValue);
   }
 
--- a/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
@@ -614,6 +614,7 @@ class ModelParser {
   ) {
     final defaultValue =
         _parseDefaultValue(documentContents, Keyword.defaultKey);
+    if (defaultValue == null) return null;
     return ProtocolEnumValueDefinition(defaultValue);
   }
 
--- a/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
@@ -182,9 +182,9 @@ class ModelParser {
       extraClasses: [],
     );
     var defaultValue = _parseDefaultValue(documentContents, Keyword.defaultKey);
+    print(defaultValue);
     var defaultEnumDefinitionValue =
         values.where((value) => value.name == defaultValue).firstOrNull;
-    print(defaultValue);
     print(defaultEnumDefinitionValue);
 
     var enumDef = EnumDefinition(
--- a/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
@@ -181,9 +181,8 @@ class ModelParser {
       '${protocolSource.moduleAlias}:$className',
       extraClasses: [],
     );
-    var defaultValue = _parseDefaultValue(documentContents, Keyword.defaultKey);
     var defaultEnumDefinitionValue =
-        values.where((value) => value.name == defaultValue).firstOrNull;
+        _parseEnumDefaultValue(documentContents, values);
 
     var enumDef = EnumDefinition(
       fileName: outFileName,
@@ -607,6 +606,15 @@ class ModelParser {
     var node = documentContents.nodes[Keyword.unique];
     var nodeValue = node?.value;
     return nodeValue is bool ? nodeValue : false;
+  }
+
+  static ProtocolEnumValueDefinition? _parseEnumDefaultValue(
+    YamlMap documentContents,
+    List<ProtocolEnumValueDefinition> values,
+  ) {
+    final defaultValue =
+        _parseDefaultValue(documentContents, Keyword.defaultKey);
+    return values.where((value) => value.name == defaultValue).firstOrNull;
   }
 
   static List<ProtocolEnumValueDefinition> _parseEnumValues(
--- a/tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
@@ -2389,20 +2389,11 @@ class SerializableModelLibraryGenerator {
         ..body = (BlockBuilder()
               ..statements.addAll([
                 const Code('switch(index){'),
-                for (int i = 0; i < enumDefinition.values.length; i++) ...[
+                for (int i = 0; i < enumDefinition.values.length; i++)
                   Code(
                     'case $i: return ${enumDefinition.className}.${enumDefinition.values[i].name};',
                   ),
-                ],
-                if (enumDefinition.defaultValue == null) ...[
-                  Code(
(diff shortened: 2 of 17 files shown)


vanlooverenkoen:

vanlooverenkoen:
You can use this override to make sure your client does not break!

```
dependency_overrides:
  serverpod_cli:
    git:
      url: git@github.com:impaktfull/serverpod.git
      ref: 103afffd5e67690d8dec9329226510a4a93741e8
      path: tools/serverpod_cli
```

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/enum_default_value_test.dart
```diff
@@ -0,0 +1,102 @@
...
+import 'package:serverpod_cli/src/generator/code_generation_collector.dart';
+import 'package:test/test.dart';
+
+import '../../../../test_util/builders/generator_config_builder.dart';
+import '../../../../test_util/builders/model_source_builder.dart';
+
+void main() {
+  var config = GeneratorConfigBuilder().build();
+
+  group('Given a valid enum definition when validating', () {
```
SandPod, line 12: Suggestion: This test can be removed since we are not validating the generalized enum parsing here and is duplicated in other places.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/enum_default_value_test.dart
```diff
@@ -0,0 +1,102 @@
...
+        ''',
+      ).build()
+    ];
+
+    var collector = CodeGenerationCollector();
+    var analyzer =
+        StatefulAnalyzer(config, modelSources, onErrorsCollector(collector));
+
+    var definitions = analyzer.validateAll();
+    var definition = definitions.first as EnumDefinition;
```
SandPod, line 61: Suggestion: Place these in a `setUp()` or mark them as `late`.

These tests correctly follow the common pattern for these tests. But an improvement we are working on currently is making sure that tests running towards the parser can be run in isolation.
By running the validation inside of the group, all validation is executed everytime the developer tries to execute a test.

If we place it in a `setup` or mark it as `late` it will only be run if that specific test is executed.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/enum_default_value_test.dart
```diff
@@ -0,0 +1,102 @@
...
+      expect(collector.errors, isEmpty);
+    });
+
+    test('then defaultValue is set to valid value.', () {
+      expect(definition.defaultValue, definition.values.first);
+    });
+  });
+
+  group(
+      'Given a valid enum definition with default value set to an invalid value when validating',
```
SandPod, line 73: Suggestion: Update description to match the test, `Given a valid enum definition without a default value when validating`.

Update the description in the test assertion as well.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/enum_default_value_test.dart
```diff
@@ -0,0 +1,102 @@
+import 'package:serverpod_cli/src/analyzer/models/definitions.dart';
+import 'package:serverpod_cli/src/analyzer/models/stateful_analyzer.dart';
+import 'package:serverpod_cli/src/generator/code_generation_collector.dart';
+import 'package:test/test.dart';
+
+import '../../../../test_util/builders/generator_config_builder.dart';
+import '../../../../test_util/builders/model_source_builder.dart';
+
+void main() {
```
SandPod, line 9: Suggestion: Please add a test that validates what happens if the default value is not one of the values specified for the enum.

(This should lead to a validation errror)

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -2389,13 +2389,20 @@ class SerializableModelLibraryGenerator {
         ..body = (BlockBuilder()
               ..statements.addAll([
                 const Code('switch(index){'),
-                for (int i = 0; i < enumDefinition.values.length; i++)
+                for (int i = 0; i < enumDefinition.values.length; i++) ...[
```
SandPod, line 2392: Suggestion/Question: We shouldn't need the `...[]` here right?
Same for the if/else statement below.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -2389,13 +2389,20 @@ class SerializableModelLibraryGenerator {
...
+                ],
+                if (enumDefinition.defaultValue == null) ...[
+                  Code(
+                    'default: throw ArgumentError(\'Value "\$index" cannot be converted to "${enumDefinition.className}"\');',
+                  ),
+                ] else ...[
+                  Code(
+                    'default: return ${enumDefinition.className}.${enumDefinition.defaultValue!.name};',
+                  ),
+                ],
```
SandPod, line 2405: Suggestion: Break this out and take in the variables as parameters.

We repeat this same code below, might as well tidy up a bit and reuse the code.

tools/serverpod_cli/test/generator/dart/server_code_generator/enum_field_test.dart
```diff
@@ -50,6 +50,13 @@ void main() {
     test('then generated enum has toJson method', () {
       expect(codeMap[expectedFileName], contains('int toJson() => index;'));
     });
+
+    test('then generated enum has throw ArgumentError', () {
+      expect(
+          codeMap[expectedFileName],
+          contains(
+              'default:\n        throw ArgumentError(\'Value "\$index" cannot be converted to "Example"\');'));
+    });
```
SandPod, line 59: Suggestion: Run these tests towards a defined model instead.

We try to stay away from testing the "implementation" of our methods in the code analyser tests. The reason is that they become very brittle for to change and we don't give us any guarantees that the outputted code actually works as expected.

Please assert this behavior on a defined enum model instead. There should be on in the `test` project that can already be used, so you should only need to add one with the default value.

This also gives us confirmation that this feature works as expected end to end.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -181,6 +181,9 @@ class ModelParser {
       '${protocolSource.moduleAlias}:$className',
       extraClasses: [],
     );
+    var defaultValue = _parseDefaultValue(documentContents, Keyword.defaultKey);
+    var defaultEnumDefinitionValue =
+        values.where((value) => value.name == defaultValue).firstOrNull;
```
SandPod, line 186: Suggestion: This filtering should be left to the validator to handle.

We don't need to make sure that the model makes perfect sense when we parse it. That is up for the validator to assert at the later state. If some issue is found, we should throw a source span exception to catch it.

We actually need this information in order to create helpful validation.

SandPod (changes requested):
Awesome work!

Have a couple of things that need to be fixed in order to get validation working, but the core implementation looks good.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/enum_default_value_test.dart
vanlooverenkoen: This one is indeed duplicated. I will remove this one

tools/serverpod_cli/test/generator/dart/server_code_generator/enum_field_test.dart
vanlooverenkoen: not really sure which `test` project you are talking about?

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
vanlooverenkoen: hmm alright but how do we then get the actual defaultEnumDefinitionValue here? I don't really see the link with the validator here, or how we get values from the validator here?

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/enum_default_value_test.dart
vanlooverenkoen:
```
      ValidateNode(
        Keyword.defaultKey,
      ),
```

How can I access the `values` in the `valueRestriction`?

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/enum_default_value_test.dart
SandPod: We have this validated in `class_type_test.dart`.

This validates the generalized enum parsing.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/enum_default_value_test.dart
SandPod: Yes, that makes sense. We are just shifting away from it now since it provides a better DevEx.

If you want you can convert the `serializedAs` tests as well, but it would be enough to apply the scout rule and patch the files that you touch.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/enum_default_value_test.dart
SandPod: You will need to retrieve the "parsed" enum definition and inspect that to fetch the values. This is why it is important to read all in the parser.

The `Restrictions` class isntance contains a `parsedModels` field that you can use to look up the parsed enum models.

You can find all models with a specific class name like so:
```dart
    var definitions = parsedModels.classNames[className];
```

Just ensure that you are matching the right module as well as the class name. Otherwise you might get a definition for an Enum that is part of another Serverpod Module.

tools/serverpod_cli/test/generator/dart/server_code_generator/enum_field_test.dart
SandPod: Inside of the `test project` located here: https://github.com/serverpod/serverpod/tree/main/tests/serverpod_test_server/lib/src/models

Then you can add a test for it here: https://github.com/serverpod/serverpod/tree/main/tests/serverpod_test_server/test

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
SandPod: The parser just performs a best effort parsing, if no other requirements that his being a string is needed for it to be parsed, then that should be fine.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/enum_default_value_test.dart
SandPod: Might also be good to test what happens when this is invalid input such as an int?

vanlooverenkoen: I think it is even easier, you can just fetch the documentDefinition which is already the EnumDefinition of the file itself. Or am I focussing on something here?

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
vanlooverenkoen: I have cleaned this up a little bit by move this to a seperate function

tools/serverpod_cli/test/test_util/builders/enum_definition_builder.dart
```diff
@@ -74,6 +77,12 @@ class EnumDefinitionBuilder {
     return this;
   }
 
+  EnumDefinitionBuilder withDefaultValue(
```
vanlooverenkoen, line 80: @SandPod I still kept this, but now it doesn't have tests using it because you told me to move it to the end 2 end tests. What should I do with this?

SandPod (commented):
Thanks for addressing my comments!

A couple of last minor things to fix and we can merge it! :)

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -605,6 +608,15 @@ class ModelParser {
     return nodeValue is bool ? nodeValue : false;
   }
 
+  static ProtocolEnumValueDefinition? _parseEnumDefaultValue(
+    YamlMap documentContents,
+    List<ProtocolEnumValueDefinition> values,
+  ) {
+    final defaultValue =
+        _parseDefaultValue(documentContents, Keyword.defaultKey);
+    return values.where((value) => value.name == defaultValue).firstOrNull;
```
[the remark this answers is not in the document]
vanlooverenkoen: I don't get this, this wil not work right becuase our EnumDefinition requires `ProtocolEnumValueDefinition` as a default value, not a String?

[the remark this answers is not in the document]
SandPod: We can simply create a `ProtocolEnumValueDefinition` inline from the String. Since we don't support any documentation for the `default` value, currently, that can be empty.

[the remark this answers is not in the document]
SandPod: Aaah, I see. That makes sense! With that motivation, your solution might actually be better.

Revert the fix for my incorrect comment please 🙏
And sorry for ping-ponging with you!

SandPod (approved):
LGTM!

Thanks for the contribution! 🚀

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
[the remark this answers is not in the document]
vanlooverenkoen: Updated

PR #3355, opened 2025-03-25
feat(tests): Extend the compilation unit class matchers.

SandPod:
Extends the compilation unit class matchers.

The goal was to migrate the class `table_class_test.dart`, but become of time limitations I wasn't able to finish the port.

The extensions to the class matches implemented here are part of that work and can be added to the framework to enable more concurrent migrations directly on the main.

This PR also contains a minor fix.
- Ensures that argument matches also match against literal values such as `bool` or `String.`

## Pre-launch Checklist

## Breaking changes

None - These are simple test matcher additions.

### DIFF
--- a/tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
+++ b/tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
@@ -196,7 +196,7 @@ abstract interface class ParameterMatcher {}
 /// A chainable matcher that matches a super initializer in a compilation unit.
 abstract interface class SuperInitializerMatcher {
   /// Chains an [ArgumentMatcher] that checks if the super initializer is called
-  /// with a specific argument.
+  /// with a specific literal argument.
   ArgumentMatcher withArgument(String value);
 
   /// Chains an [ArgumentMatcher] that checks if the super initializer is called
--- a/tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
+++ b/tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
@@ -8,6 +8,7 @@ import 'compilation_unit_matcher/chainable_matcher.dart';
 part 'compilation_unit_matcher/argument_matcher/argument_matcher.dart';
 part 'compilation_unit_matcher/class_matcher/class_matcher.dart';
 part 'compilation_unit_matcher/constructor_matcher/constructor_matcher.dart';
+part 'compilation_unit_matcher/extends_matcher/extends_matcher.dart';
 part 'compilation_unit_matcher/field_matcher/field_matcher.dart';
 part 'compilation_unit_matcher/method_matcher/method_matcher.dart';
 part 'compilation_unit_matcher/parameter_matcher/parameter_matcher.dart';
@@ -28,8 +29,10 @@ part 'compilation_unit_matcher/super_initializer_matcher/super_initializer_match
 ///   │   ├── SuperInitializerMatcher
 ///   │   │   └── ArgumentMatcher
 ///   │   └── ParameterMatcher
-///   └── MethodMatcher
-///       └── ParameterMatcher
+///   ├── MethodMatcher
+///   │   └── ParameterMatcher
+///   └── ExtendsMatcher
+///
 ClassMatcher containsClass(String className) => _ClassMatcherImpl._(className);
 
 /// Parses a string of Dart code into a FormattedCompilationUnit.
@@ -42,6 +45,9 @@ abstract interface class ArgumentMatcher {}
 /// A matcher that checks if a CompilationUnit contains a class that matches
 /// certain criteria.
 abstract interface class ClassMatcher {
+  /// Chains a [ExtendsMatcher] that checks if the class extends a specific class.
+  ExtendsMatcher thatExtends(String className);
+
   /// Chains a [FieldMatcher] that checks if the class contains a field with a
   /// specific name.
   ///
@@ -138,6 +144,9 @@ abstract interface class ConstructorMatcher {
     bool? isRequired,
   });
 }
+
+/// A chainable matcher that matches the extension in a compilation unit.
+abstract interface class ExtendsMatcher {}
 
 /// A chainable matcher that matches a field in a compilation unit.
 abstract interface class FieldMatcher {}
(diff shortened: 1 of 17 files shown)


tools/serverpod_cli/test/test_util/compilation_unit_matcher/chainable_matcher.dart
nielsenko, line 1: I guess this is the gist of it. LGTM 🚀

tools/serverpod_cli/test/test_util/compilation_unit_matcher/extends_matcher/extends_matcher.dart
```diff
@@ -0,0 +1,65 @@
...
+  ) {
+    var match = _parent.matchedFeatureValueOf(item);
+    if (match == null) {
+      return _parent.describeMismatch(
+        item,
+        mismatchDescription,
+        matchState,
+        verbose,
+      );
+    }
```
nielsenko, line 29: Consider creating a helper for this repeated code block

tools/serverpod_cli/test/test_util/compilation_unit_matcher/extends_matcher/extends_matcher.dart
```diff
@@ -0,0 +1,65 @@
...
+    var match = _parent.matchedFeatureValueOf(actual);
+    if (match == null) return null;
+
+    return match.value;
+  }
+}
+
+extension on ExtendsClause {
+  String? _getExtendedTypeName() {
+    return superclass.name2.lexeme;
```
nielsenko, line 85: Don't you just hate these `x<number>` in the `analyzer`?

tools/serverpod_cli/test/test_util/compilation_unit_matcher/generic_matcher/generic_matcher.dart
```diff
@@ -0,0 +1,55 @@
...
+  ) {
+    var match = _parent.matchedFeatureValueOf(item);
+    if (match == null) {
+      return _parent.describeMismatch(
+        item,
+        mismatchDescription,
+        matchState,
+        verbose,
+      );
+    }
```
nielsenko, line 29: .. here it is again 😄

tools/serverpod_cli/test/test_util/compilation_unit_matcher/generic_matcher/generic_matcher_test.dart
```diff
@@ -0,0 +1,134 @@
...
+        description.toString(),
+        equals('does not have any generics'),
+      );
+    });
+  });
+
+  group('Given a class with an extends clause that has generics', () {
+    late final compilationUnit = parseCode(
+      '''
+      class User extends Parent<T> {}
```
nielsenko, line 41: Come to think of it. Should we support type arguments on the defined class as well?
```dart
class User<T> extends Parent<T> {}
```

tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
```diff
@@ -146,7 +147,9 @@ abstract interface class ConstructorMatcher {
 }
 
 /// A chainable matcher that matches the extension in a compilation unit.
-abstract interface class ExtendsMatcher {}
+abstract interface class ExtendsMatcher {
+  GenericMatcher withGeneric(String genericType);
```
nielsenko, line 162: Should we allow a list here? I don't think we will chain this further.

nielsenko (approved):
LGTM 💯 .. Did a few comments, but nothing crucial.

tools/serverpod_cli/test/test_util/compilation_unit_matcher/argument_matcher/argument_matcher.dart
```diff
@@ -134,8 +134,6 @@ extension on Expression {
       return resolvedThis.expression._hasMatchingValue(name);
     }
 
-    if (resolvedThis is! SimpleIdentifier) return false;
-
-    return resolvedThis.name == name;
+    return resolvedThis.toSource() == name;
```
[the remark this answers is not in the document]
SandPod: Variable arguments implement the [identifier](https://github.com/dart-lang/sdk/blob/8d693e032d5992dfcda044ceb86b522122a7e1eb/pkg/analyzer/lib/src/dart/ast/ast.dart#L16099) interface.
But when `literal` values are used, such as `String`, `bool` or `int` these implement the [Literal](https://github.com/dart-lang/sdk/blob/8d693e032d5992dfcda044ceb86b522122a7e1eb/pkg/analyzer/lib/src/dart/ast/ast.dart#L11507) interface. 

This becomes a bit of a tradeof where we would need special casing in order to "match" the actual value presented. By using the source directly, we preserve full flexibility around what the "value" might be.

tools/serverpod_cli/test/test_util/compilation_unit_matcher/extends_matcher/extends_matcher.dart
SandPod: Haha yes... 🙃

tools/serverpod_cli/test/test_util/compilation_unit_matcher/extends_matcher/extends_matcher.dart
SandPod: Good catch! Will fix!

tools/serverpod_cli/test/test_util/compilation_unit_matcher/generic_matcher/generic_matcher_test.dart
SandPod: Yes, that makes sense. And that was the main motivation for structuring the matcher in this way.

But I didn't have a usecase for it (yet), so I opted not to implement it. But I'm pretty certain we do that type of validation in other places.

I can add it on the second round.

tools/serverpod_cli/test/test_util/compilation_unit_matcher.dart
SandPod: I think that could be reasonable, but I would still opt for keeping it as a single validation for now and then expand it if we see that this causes repetition.

tools/serverpod_cli/test/test_util/compilation_unit_matcher/argument_matcher/argument_matcher.dart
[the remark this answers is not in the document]
nielsenko: I doubt it will actually change anyway, and if it does we can address it then.

PR #3360, opened 2025-03-27
fix: Prevent update from clearing out implicit fields.

SandPod:
First part of fixing #3286

This PR introduces "managedColumns" which are used for update queries when no columns are specified.

The managed columns only include columns that are managed by this model, i.e. not implicit fields created as a result of implicit one to many relations.

There will be a follow up PR that ensures this implicit data is preserved during Serialization as well.

### Additional notes
The PR also includes a refactor that enables running the touched code generator tests in isolation when debugging.

I would recommend reviewing this commit by commit as the generated changes are then isolated.
Most changes come the new generated migration.

## Pre-launch Checklist

## Breaking changes

None - Should be a straight fix.


tools/serverpod_cli/test/generator/dart/server_code_generator/table_class_test.dart
```diff
@@ -337,6 +337,16 @@ void main() {
...
+
+      test('has managedColumns override for columns', () {
+        expect(
+            CompilationUnitHelpers.hasMethodDeclaration(
+              maybeClassNamedExampleTable!,
+              name: 'managedColumns',
+            ),
+            isTrue,
+            reason: 'Missing declaration for managedColumns override.');
+      });
```
nielsenko, line 349: Why not use the new matchers?

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -1619,6 +1619,9 @@ class SerializableModelLibraryGenerator {
       List<SerializableModelFieldDefinition> fields,
       ClassDefinition classDefinition,
       TypeReference idTypeReference) {
+    var serializedFields = fields
+        .where((f) => f.shouldSerializeFieldForDatabase(serverCode))
+        .toList();
```
nielsenko, line 1624: .. moves `shouldSerializeFieldForDatabase` call out of `_buildModelTableClassColumnGetter` ✅ 

Consider `toSet` over `toList`

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -1640,7 +1643,16 @@ class SerializableModelLibraryGenerator {
           classDefinition.subDirParts,
         ),
         ..._buildModelTableClassManyRelationGetters(fields, classDefinition),
-        _buildModelTableClassColumnGetter(fields),
+        _buildModelTableClassColumnGetter(
+          serializedFields,
+          name: 'columns',
```
nielsenko, line 1648: columns is all serialized fields ✅

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -1640,7 +1643,16 @@ class SerializableModelLibraryGenerator {
...
+        ),
+        if (serializedFields.any((f) => f.hiddenSerializableField(serverCode)))
```
nielsenko, line 1650: Checks if overwrite needed..

Perhaps introduce `hiddenFields = serializedFields.where(...).toSet()` and test for empty.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -1640,7 +1643,16 @@ class SerializableModelLibraryGenerator {
...
+          _buildModelTableClassColumnGetter(
+            serializedFields
+                .where((f) => !f.hiddenSerializableField(serverCode)),
```
nielsenko, line 1653: If using sets, this could be `serializedFields.difference(hiddenFields)`

nielsenko (approved):
LGTM 👍 
I have some suggestions to use sets over lists and iterables, but they ar entirely optional.
Also, I would have expected to see the new matchers used 😄

tools/serverpod_cli/test/generator/dart/server_code_generator/table_class_test.dart
SandPod: This was the test file I was trying to convert, but there were so many new matchers needed so I bailed out.

Will finish the conversion here in a separate PR to not delay the bugfix.

PR #3363, opened 2025-03-28
feat: Enable adding custom data to diagnostic events

### LINKED ISSUE #3332: Ability to add custom context data to the diagnostic events system
### Problem to Solve

Currently the event context information is fixed by the framework. Besides the message string, no information can be added when reporting diagnostic events.

### Proposal

Add a custom attributes entry to `DiagnosticEventContext`, that will be included in the `toJson()` output.

### Use Case

When using this system with Sentry, additional tags and/or context attributes can be added to identify what operations are being performed.

### Alternatives

n/a

### Additional context

_No response_

- [x] I want to work on this

christerswahn:
This adds the ability to include custom data when submitting diagnostic events.

It is intended to be used by application (user) code, not framework code. The content to include and how to handle it in the event handlers is up to the user.

Closes #3332 

The solution differs slightly from the suggestion in the issue: Rather than extending the context, this is included as a key/value map in the event itself. This makes sense since the context types are framework-managed and this is an entirely user-owned data payload for individual events.

## Pre-launch Checklist

## Breaking changes

No, does not change any existing API. Adds an optional initialization field to the diagnostic event constructor.

### DIFF
--- a/packages/serverpod/lib/src/server/diagnostic_events/event_handler.dart
+++ b/packages/serverpod/lib/src/server/diagnostic_events/event_handler.dart
@@ -10,7 +10,13 @@ enum OriginSpace {
 }
 
 /// Base interface for all diagnostic events.
-abstract interface class DiagnosticEvent {}
+abstract interface class DiagnosticEvent {
+  /// Custom key/value data added to the event.
+  ///
+  /// It is up to application code submitting events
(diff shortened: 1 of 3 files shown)


SandPod (approved):
LGTM! 🔍 🚀

Issue #3377, opened 2025-04-02
Generate doesn't handle modules in a Dart workspace with multiple servers


christerswahn:
### Describe the bug

When using a Dart workspace in a monorepo with multiple Serverpod server packages, the serverpod generate command gets confused regarding the Serverpod modules being used. It will generate code importing all Serverpod modules used in the workspace, regardless of whether the individual server project has specified the module package as a dependency in its pubspec.yaml. These imports then cause the Dart compilation to break.

### To reproduce

Have a Dart workspace with two Serverpod server projects, one using a Serverpod module that the other isn't. Run serverpod generate in the project not using the module, and check the imports of the generated code.

### Expected behavior

The generated code for a Serverpod server should only include dependencies that are actual dependencies of that server project.

### Library version

Serverpod 2.4.0

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

PR #3380, opened 2025-04-03
refactor: Refactor locate modules logic.

SandPod:
This PR is done in preparation for fixing #3377.

It seems like the world around `locate_modules` changed without updating the method.
This made part of the internal logic redundant.

I broke this out into a clear refactor PR so that these changes could be viewed in isolation.

## Pre-launch Checklist

## Breaking changes

None - These changes are not indented to be breaking.

### DIFF
--- a/tools/serverpod_cli/lib/src/language_server/language_server.dart
+++ b/tools/serverpod_cli/lib/src/language_server/language_server.dart
@@ -48,11 +48,8 @@ Future<void> runLanguageServer() async {
   });
 
   connection.onInitialized((_) async {
-    if (serverProject == null &&
-        exception is ServerpodModulesNotFoundException) {
+    if (exception is ServerpodModulesNotFoundException) {
       _sendModulesNotFoundNotification(connection);
-    } else if (serverProject == null) {
-      return;
     }
 
     serverProject?.analyzer.validateAll();
--- a/tools/serverpod_cli/lib/src/language_server/language_server.dart
+++ b/tools/serverpod_cli/lib/src/language_server/language_server.dart
@@ -26,18 +26,13 @@ Future<void> runLanguageServer() async {
   var connection = Connection(stdin, stdout);
 
   ServerProject? serverProject;
-  Exception? exception;
 
   connection.onInitialize((params) async {
     var rootUri = params.rootUri;
     if (rootUri != null) {
       try {
         serverProject = await _loadServerProject(rootUri, connection);
-      } catch (error) {
-        if (error is Exception) {
-          exception = error;
-        }
-      }
+      } catch (_) {}
     }
 
     return InitializeResult(
@@ -48,10 +43,7 @@ Future<void> runLanguageServer() async {
   });
 
   connection.onInitialized((_) async {
-    if (serverProject == null &&
-        exception is ServerpodModulesNotFoundException) {
-      _sendModulesNotFoundNotification(connection);
-    } else if (serverProject == null) {
+    if (serverProject == null) {
       return;
     }
 
@@ -120,17 +112,6 @@ Future<void> runLanguageServer() async {
   });
 
   await connection.listen();
-}
-
-void _sendModulesNotFoundNotification(Connection connection) {
-  connection.sendNotification(
-    'window/showMessage',
-    ShowMessageParams(
-      message:
-          'Serverpod model validation disabled. Unable to locate necessary modules, have you run "dart pub get"?',
-      type: MessageType.Warning,
-    ).toJson(),
-  );
 }
 
 Future<ServerProject?> _loadServerProject(
--- a/tools/serverpod_cli/lib/src/language_server/language_server.dart
+++ b/tools/serverpod_cli/lib/src/language_server/language_server.dart
@@ -26,13 +26,18 @@ Future<void> runLanguageServer() async {
   var connection = Connection(stdin, stdout);
 
   ServerProject? serverProject;
+  Exception? exception;
 
   connection.onInitialize((params) async {
     var rootUri = params.rootUri;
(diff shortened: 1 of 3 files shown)


tools/serverpod_cli/lib/src/util/locate_modules.dart
```diff
@@ -3,83 +3,72 @@ import 'dart:io';
...
-          ),
-        );
-      } catch (e) {
+  for (var packageInfo in packageConfig.packages) {
+    try {
+      var packageName = packageInfo.name;
+
+      if (!packageName.endsWith(_serverSuffix) && packageName != 'serverpod') {
+        continue;
+      }
```
Isakdl, line 23: Comment: I know this is a refactor, but a note for future changes. We should remove this serverSuffix and replace that by looking at the project dependencies. E.g. if package depends on "serverpod" it is a serverpod package.

tools/serverpod_cli/lib/src/util/locate_modules.dart
```diff
@@ -3,83 +3,72 @@ import 'dart:io';
...
+
+      var migrationVersions = findAllMigrationVersionsSync(
+        directory: Directory.fromUri(moduleProjectUri),
+        moduleName: moduleName,
+      );
+
+      var moduleInfo = loadConfigFile(generatorConfigFile);
+
+      var manualNickname = manualModules[moduleName];
+      var nickname = manualNickname ?? moduleInfo['nickname'] ?? moduleName;
```
Isakdl, line 55: Comment: Just realised that this can be a problem regarding the discussions we had yesterday on code generation and modules. Might even be a problem already? We are basically saying that the user project can override the nickname of a module. 

In the following scenario:
A -> B -> C
A -> C

Where A is your project, imagine that you give C the alias "carrot".
But module B has given C the alias "watermelon".

E.g. the graph looks like this:

A -> B-> watermelon (C).
A -> carrot (C).

I think we are okay right now, as we have duplication of the generated code, but this can be a pitfall when resolving modules in a more reasonable way.

tools/serverpod_cli/lib/src/util/locate_modules.dart
```diff
@@ -3,83 +3,72 @@ import 'dart:io';
...
+        ModuleConfig(
+          type: GeneratorConfig.getPackageType(moduleInfo),
+          name: moduleName,
+          nickname: nickname,
+          migrationVersions: migrationVersions,
+          serverPackageDirectoryPathParts: moduleProjectRoot,
+        ),
+      );
+    } catch (e) {
+      continue;
```
Isakdl, line 67: Question: Is this what we want to do, silently ignore packages we cannot find on disk?

tools/serverpod_cli/lib/src/language_server/language_server.dart
```diff
@@ -48,10 +43,7 @@ Future<void> runLanguageServer() async {
   });
 
   connection.onInitialized((_) async {
-    if (serverProject == null &&
-        exception is ServerpodModulesNotFoundException) {
-      _sendModulesNotFoundNotification(connection);
-    } else if (serverProject == null) {
```
Isakdl, line 54: Question: Are you sure this wasn't used? This will give you a notification in the right left corner if the loading of the Serverpod project failed. I'm sure we gave an error if we couldn't locate the modules, I believe we introduced that in the CLI to make sure the code generator didn't silently ignore to generate auth endpoints for example.

Isakdl (commented):
Some comments, and I'm wondering if you are correct with the notification in the LSP.

tools/serverpod_cli/lib/src/language_server/language_server.dart
SandPod: The only place in the code where this was thrown was the place that was safely removed. I think this error message has not been seen for a while.

tools/serverpod_cli/lib/src/util/locate_modules.dart
SandPod: Yes, will tackle this in a different PR! 👍

tools/serverpod_cli/lib/src/util/locate_modules.dart
SandPod: Yes, becuase the inspected packages could today be any package with a `_server` prefix. Once we have a more narrow definition of the projects inspected we can put higher demands here.

tools/serverpod_cli/lib/src/language_server/language_server.dart
SandPod: Reverted the removal of the handling of this in the LSP.

In an upcoming PR I will reintrodue this throw once we now that we are only inspecting modules we are depending on.

tools/serverpod_cli/lib/src/language_server/language_server.dart
```diff
@@ -43,7 +48,10 @@ Future<void> runLanguageServer() async {
   });
 
   connection.onInitialized((_) async {
-    if (serverProject == null) {
+    if (serverProject == null &&
+        exception is ServerpodModulesNotFoundException) {
+      _sendModulesNotFoundNotification(connection);
+    } else if (serverProject == null) {
       return;
     }
```
Isakdl, line 56: Nit: As we discussed we could remove all the null checks here just to make the flow a little more obvious.

Isakdl (approved):
LGTM, but added a small nit (none blocking).

tools/serverpod_cli/lib/src/language_server/language_server.dart
SandPod: aah, good catch! will fix this.

PR #3381, opened 2025-04-03
feat: Add v2 authentication modules

tp:
_Details TBD_


modules/serverpod_auth2/README.md
```diff
@@ -0,0 +1,61 @@
...
+`serverpod_auth_email`: Full-service email account endpoint implementation where users can register, login, change password, etc.  
+`serverpod_auth_email_account`: Low-level package to handle only the email authentication. This would not create sessions or users itself. Instead the goal is to server as a building block to build custom authentication endpoints
+
+(For other authentication methods like Apple, Firebase, Google a similar approach would be implemented across 2 modules.)
+
+`serverpod_auth_migration`: Migration tools to import data from the legacy `serverpod_auth` into the new modules
+
+`serverpod_auth_profile`: User profile similar to the current `UserInfo` class. Also showcases how developers might build their own custom "user profile" classes
+
+## Dependencies
```
tp, line 35: @SandPod Quick follow-up from our chat of how the "super modular" structure would impact the dependencies and imports of a project.

The actual imports we could get down by re-exporting (like e.g. email account config through the email endpoints, where it's an implementation detail), but in the `pubspec.yaml` the modularity of course leads to a few more imports, declaring explicitly what one wants to use.

SandPod:
@tp - Can we close this?

tp:
> @tp - Can we close this?

Yep. From time to time I'm checking back here, but for that it does not need to stay open :)

PR #3382, opened 2025-04-03
fix: Support non-nullable `Set` in models

tp:
Before the de-serialization code for a `Set<T>` was not generated correctly, because the helper was typed to return `Set<T>?` even though it would always return a value for non-`null` inputs.

We missed this until now because for the new types we did not add the cases to the legacy types test based on `tests/serverpod_test_server/lib/src/models/nullability.spy.yaml` and in all other cases the fields were always nullable. 

Presumably that test was phased out because it's painfully noisy to add all those required cases when you don't want to look at them in any individual test.
To test the current issue I thus opted to create a new focused model that only has one required / non-nullable field to test in isolation.

Follow up fix for #3211

Relates to https://github.com/serverpod/serverpod/issues/1784

### DIFF
--- a/tests/serverpod_test_client/lib/src/protocol/protocol.dart
+++ b/tests/serverpod_test_client/lib/src/protocol/protocol.dart
@@ -147,16 +147,17 @@ import 'types_list.dart' as _i127;
 import 'types_map.dart' as _i128;
 import 'types_record.dart' as _i129;
 import 'types_set.dart' as _i130;
-import 'unique_data.dart' as _i131;
-import 'my_feature/models/my_feature_model.dart' as _i132;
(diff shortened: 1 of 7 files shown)


SandPod (commented):
Good find and thanks and nice that it was possible to patch.

Could you file an issue for removing the "nullability" from the `fromJson` on the Set extension for 3.0. I think we could nullcheck before doing this right?

tp:
@SandPod Added #3383 to clean this up with a breaking change in the future (I only found the "breaking" label, not any specific "3.0" badge / release).

SandPod (approved):
LGTM! 👍 

Thanks for creating the issue as well 🎉

Issue #3394, opened 2025-04-08
HealthCheckManager.start pollutes the debug console with an avoidable exception including stacktrace


dabblingwithcode:
### Problem to Solve

Starting with serverpod 2.5. I'm getting two errors in the debug console instead of just one:

```
2025-04-08 07:33:52.640742Z ERROR: CPU and memory usage metrics are not supported on this platform.
2025-04-08 07:33:52.640742Z ERROR: ProcessException: The system could not find the file.
```
The second exception is caused by `package:system_resources/src/dylib.dart:15:25`, which is trying to execute `uname -sm`, which won't work in windows, and pollutes the console with 12 stacktrace steps.

![Image](https://github.com/user-attachments/assets/5b28427f-ea89-42f3-900d-4d9954fac2ad)

### Proposal

For developer experience, it would be good to skip the health checks altogether - since they seem to not work properly on windows - and inform accordingly just once in the console, maybe like this:

```dart
 Future<void> start() async {
    _running = true;
    if (!Platform.isWindows) {
      try {
        await SystemResources.init();
      } catch (e, stackTrace) {
        _reportException(
          e,
          stackTrace,
          message:
              'CPU and memory usage metrics are not supported on this platform.',
        );
      }
    } else {
      stderr.writeln(
          'WARNING: Skipping health checks because windows does not support them.');
    }
    _scheduleNextCheck();
  }
```

### Use Case

Just a decent warning does not scare developers and makes them happier :-)

![Image](https://github.com/user-attachments/assets/9bc3c9e8-beb2-4549-8e05-fc7dcdf5ff2b)

### Alternatives

Fork the package `system_resources` or contribute to it fixing windows functionality, which I think is not worth the work since in production mainly linux will be used.

### How experienced are you with this library?

Intermediate - Familiar with the basics or have used it in a few projects

### Are you interested in working on a PR for this?

- [x] I want to work on this

dabblingwithcode:
On further research the additional exception print on windows was introduced by this recent commit (which makes sense for non windows platforms):  [fix: Report exceptions during database start and healthcheck operations](https://github.com/serverpod/serverpod/commit/d4fc895b2c5475d093cec8801a34932a4e741f72)

SandPod:
Yes,  we should display this in a nice way on platforms that are now supported.

christerswahn:
Thanks for reporting this and doing an investigation!

[...]

PR #3399, opened 2025-04-09
fix: Avoid unnecessary stacktrace in windows when performing health checks

dabblingwithcode:
Fixes https://github.com/serverpod/serverpod/issues/3394

Avoid debug console pollution with unnecessary stackstrace because the used package does not support windows. (https://github.com/jonasroussel/system_resources/issues/3).

Fixes https://github.com/serverpod/serverpod/issues/3394

## Pre-launch Checklist

- [x ] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [ x] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
- [x ] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).
- [x ] I listed at least one issue that this PR fixes in the description above.
- [x ] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.

## Breaking changes

None.

### DIFF
--- a/packages/serverpod/lib/src/server/health_check_manager.dart
+++ b/packages/serverpod/lib/src/server/health_check_manager.dart
@@ -7,8 +7,8 @@ import 'package:serverpod/src/database/database_pool_manager.dart';
 import 'package:serverpod/src/server/command_line_args.dart';
 import 'package:serverpod/src/server/health_check.dart';
 import 'package:serverpod/src/server/serverpod.dart';
-import 'package:system_resources/system_resources.dart';
 import 'package:serverpod/src/util/date_time_extension.dart';
+import 'package:system_resources/system_resources.dart';
 
 /// Performs health checks on the server once a minute, typically this class
 /// is managed internally by Serverpod. Writes results to the database.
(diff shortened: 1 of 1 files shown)


christerswahn:
It turns out the system_resources package also has an issue on Linux when the server is compiled as an exe. So it is not just on Windows that we get this error display behavior.

It is also the case that the current health check manager doesn't handle initialization errors properly, it happily continues and just ignores following exceptions on each check.

Ideally the health check manager should be refactored to consist of independent modules, each with its own init and if that init fails, those health checks will not be performed. And this way the exception-ignoring catch clauses can be designed away.

That might be a bigger undertaking than you are up for currently? :) 
To do this short term fix, please remove the platform-is-windows check and just do the WARNING printout in the catch clause.

christerswahn (changes requested):
To do this short term fix, please remove the platform-is-windows check and just do the WARNING printout in the catch clause.

dabblingwithcode:

christerswahn (approved):
LGTM!

SandPod:
@dabblingwithcode -  I think you need to create a comment that only contains `I have read the CLA Document and I hereby sign the CLA.` to pass the CLA check.

dabblingwithcode:

PR #3400, opened 2025-04-09
Show explicit usage of `abstract` endpoint

tp:
and consumption in developer's project

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_server/lib/src/endpoints/email_endpoint.dart
+++ b/modules/serverpod_auth/serverpod_auth_server/lib/src/endpoints/email_endpoint.dart
@@ -8,7 +8,8 @@ import 'package:serverpod_auth_server/serverpod_auth_server.dart';
 // const _configFilePath = 'config/google_client_secret.json';
 
 /// Endpoint for handling Sign in with Google.
-class EmailEndpoint extends Endpoint {
+// `abstract` ensures that this endpoint is not generated in the module's server/client code
+abstract class EmailEndpoint extends Endpoint {
   /// Authenticates a user with email and password. Returns an
   /// [AuthenticationResponse] with the users information.
   Future<AuthenticationResponse> authenticate(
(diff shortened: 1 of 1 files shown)


modules/serverpod_auth/serverpod_auth_server/lib/src/endpoints/email_endpoint.dart
```diff
@@ -69,3 +70,49 @@ class EmailEndpoint extends Endpoint {
...
+  ) {
+    if (!email.endsWith('@serverpod.dev')) {
+      throw ArgumentError.value('email');
+    }
+
+    return super.createAccountRequest(session, userName, email, password);
+  }
+
+  // For a while the old `authenticate` method is still being supported, but the developer plans to require 2FA via OTP
+  // Eventually they would use a Serverpod provided `@obsolete` annotation (or similar), which would instruct code gen (which
```
SandPod, line 91: We already have support for disabling an endpoint using the `@ignoreEndpoint` annotation: https://docs.serverpod.dev/concepts/working-with-endpoints#ignore-endpoint-definition

tp: @SandPod That might be useful. Does this also work on the method-level? Then we could use it instead of `@obsolete`.

Personally I would think `abstract` is clearer and simpler, but maybe there was a good reason to still allow the class to be initialized, even if no generated code would do so.

Do you recall what that was?

SandPod: Yes, this works on both the method and class level.

This is used because we provide a storage endpoint class that needs to be manually added to the Server (legacy reasons).

tp: Alright, if you prefer this we could also use it for both the "base" endpoint itself, and any inherited methods the developer wants to hide in their subclass.

If we want to figure out the complexity and feasibility of this, I'd build a branch showcasing such subclassing behavior :)

tp:
actual implementation in #3004

Issue #3397, opened 2025-04-08
Too many warnings from serverpod generate


nielsenko:
### Describe the bug

Some of our internal tools have dependencies where the package name starts with _serverpod_ that are not kept in sync with the serverpod release version. Hence checking all packages with a _serverpod_ prefix is too strict.

This issue was introduced as a side effect of fixing an error that prevented warnings to be surfaced in the first place.

```
$ serverpod generate
The version of the CLI may be incompatible with the Serverpod packages used in your project.
Warning on line 40, column 26 of pubspec.yaml: The package version does not match the version of the Serverpod's CLI. This may cause errors or unexpected behavior.
   ╷
40 │   serverpod_cloud_lints: 1.0.0
   │                          ^^^^^
   ╵
✓ Generating code (6.2s)
```

### To reproduce

In a server project a dependency on a package with _serverpod_ as prefix name that has a different version than the version of the `serverpod_cli` package used to run `serverpod generate`.

### Expected behavior

No warnings should be raised in this case

### Library version

serverpod_cli 2.5.0
dart version doesn't matter

### Platform information

all platforms

### Additional context

related to https://github.com/serverpod/serverpod/pull/3273

### How experienced are you with this library?

Some

### Are you interested in working on a PR for this?

- [x] I want to work on this

PR #3401, opened 2025-04-09
fix: Restrict version check to serverpod, and serverpod_client packages

nielsenko:
Restrict the packages that must match the version of `serverpod_cli`, when running the generate command.

Now we only check `serverpod_client` and `serverpod` which matches the imports used in the generated code.

Fixes: #3397

## Pre-launch Checklist

- [x] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make mulxiple PRs.)
- [x] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [daxt format](https://dart.dev/tools/dart-format).
- [x] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as othxr Serverpod documentation. I checked spelling and grammar.

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/serverpod_packages_version_check/serverpod_packages_version_check.dart
+++ b/tools/serverpod_cli/lib/src/serverpod_packages_version_check/serverpod_packages_version_check.dart
@@ -17,7 +17,10 @@ class ServerpodPackagesVersionCheckWarnings {
 
 extension on PubspecPlus {
   Iterable<HostedDep> get serverpodDeps =>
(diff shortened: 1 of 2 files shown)


christerswahn (commented):
LGTM, but @SandPod will know more about the specific packages we should be checking against.

SandPod (approved):
LGTM! 👍 

This should be fine and give a more narrow warning.

Issue #912, opened 2023-05-14
Cannot extend `Endpoint` two levels deep


lukehutch:
**Describe the bug**

I have these two classes:

```dart
class LoggedInEndpoint extends Endpoint {
  @override
  bool get requireLogin => true;
}

class AdminEndpoint extends LoggedInEndpoint {
  @override
  Set<Scope> get requiredScopes => {Scope.admin};
}
```

If I now try to create an endpoint by subclassing one of these, Serverpod does not generate the endpoint code:

```dart
class MyEndpoint extends LoggedInEndpoint {
    // some methods...
}
```

**Serverpod versions**
 -  _CLI Version:_ 1.0.1
 - _Library version:_ 1.0.1

**Platform information**

Linux

tgrushka:
@lukehutch and @Isakdl , this issue still exists on `main` but I see the linked feature has been closed. I too would like to be able to do the `LoggedInEndpoint` inheritance thing because it makes the code more declarative and prevents forgetting to change `requireLogin` back to `true` while developing/testing (having it only in one place, the parent class, is much better). (On a side note, one can also declare `requireLogin` as `@nonVirtual` using the `meta` package to warn devs not to override it.)

Also, I tried `abstract LoggedInEndpoint extends Endpoint...` which crashes the generator. Would be nice to have parent/intermediate class(es) be abstract as well.

Is a feature planned for the next release?

vlidholt:
We did some initial work on this, but it's not a very quick fix. We still may include it into a future version though.

There are some more details on the PR: https://github.com/serverpod/serverpod/pull/969

donywan:
Is it possible to add a pre-interceptor here? 
```
Serverpod(
List<String> args,
this.serializationManager,
this.endpoints, {
ServerpodConfig? config,
this.authenticationHandler,
this.healthCheckHandler,
this.httpResponseHeaders = _defaultHttpResponseHeaders,
this.httpOptionsResponseHeaders = _defaultHttpOptionsResponseHeaders,
})
```

tp:
This is being worked on in #3404.

@lukehutch Please feel free to have a look at the newly supported cases in https://github.com/serverpod/serverpod/pull/3404/files#diff-0f5578215ca76becc26071b526f762443d0e344c08d36fe49d2ea9a3be9d18b7

Does this fix your original use-case and anything that might've come up in the meantime since you opened this issue?

SandPod:
[...]

PR #3404, opened 2025-04-09
Allow subclassing `Endpoint`s

tp:


tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_class_analyzer.dart
```diff
@@ -32,15 +32,16 @@ abstract class EndpointClassAnalyzer {
...
-      var type = constant?.type;
-      var typeName = type?.element?.name;
-      return typeName == ServerpodAnnotationClassNames.ignoreEndpoint;
-    });
-    if (markedAsIgnored) return false;
+  static bool isEndpointClass(
+    ClassElement element, {
+    bool respectIgnoreClassAnnotation = true,
+  }) {
+    if (element.allSupertypes.any((s) => s.element.name == 'Endpoint') !=
```
tp, line 39: This was like the existing logic, but ideally we'd probably check that `Endpoint` is from `serverpod/serverpod`. But I did not see a quick way to get the package path.

SandPod:
I think we should connect this to resolving this issue: https://github.com/serverpod/serverpod/issues/912

tp:
> I think we should connect this to resolving this issue: #912

I think this should already work, as I ran a similar test-case with a nested hide/unhide flow, ensuring that the "lowest" one of the class level wins (e.g. we apply the proper Dart hierarchy of which method is use and which annotation is used in the end).

I can quickly add this test case to the branch to ensure that that is really the case, and then we can look at the generated code. I think normal Dart fields will work just fine, but maybe there is some special handling with e.g. the `scopes` at code-gen time (which would need to looked up on the base class?). Will check this out, but definitely agree that this would be good to fix in one batch :)

packages/serverpod/lib/serverpod.dart
```diff
@@ -4,6 +4,9 @@
 // See: https://github.com/serverpod/serverpod/pull/2900
 library serverpod;
 
+// Annotations
+export 'package:serverpod_shared/src/annotations.dart';
```
SandPod, line 8: Discussion: The reason this was exported in a separate package was to make all annotations discoverable (import with an alias gives you autocomplete on what is exported). And it is more of an advanced scenario in endpoint management.

If we have additional tools like this perhaps we should have a separate "tools" package that exports these, and `serverpod` only includes the "normal" case exports?

This also gives you the possibility to explore what is exported easily right?

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
```diff
@@ -68,6 +69,7 @@ abstract class EndpointMethodAnalyzer {
   /// be validated and parsed.
   static bool isEndpointMethod(MethodElement method) {
```
SandPod, line 70: Suggestion: Please add tests for this change in `endpoint_method_test.dart`.

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_class_analyzer.dart
```diff
@@ -32,15 +32,18 @@ abstract class EndpointClassAnalyzer {
...
-      var typeName = type?.element?.name;
-      return typeName == ServerpodAnnotationClassNames.ignoreEndpoint;
-    });
-    if (markedAsIgnored) return false;
+  static bool isEndpointClass(
+    ClassElement element, {
+    bool respectIgnoreClassAnnotation = true,
+    bool ignoreAbstractClasses = true,
+  }) {
+    if (element.isAbstract && ignoreAbstractClasses) return false;
```
SandPod, line 40: Suggestion: Use whitelist filter instead of blacklist for this.

Is it possible to assert that his is a concerete class? My thinking is that we might otherwise need to filter out interfaces and mixins (that also support super types).

SandPod: Comment: We didn't actualy filter out these before, waas it possible to define an abstract endpoint method?

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_class_analyzer.dart
```diff
@@ -32,15 +32,18 @@ abstract class EndpointClassAnalyzer {
...
-  static bool isEndpointClass(ClassElement element) {
-    if (element.supertype?.element.name != 'Endpoint') return false;
-    bool markedAsIgnored = element.metadata.any((annotation) {
-      var constant = annotation.computeConstantValue();
-      var type = constant?.type;
-      var typeName = type?.element?.name;
-      return typeName == ServerpodAnnotationClassNames.ignoreEndpoint;
-    });
-    if (markedAsIgnored) return false;
+  static bool isEndpointClass(
```
SandPod, line 35: Suggestion: Add tests for the changes here in `endpoint_ignore_annotation_test.dart` and `endpoint_class_test.dart`.

tests/serverpod_test_module/serverpod_test_module_server/lib/src/endpoints/ignored_endpoint.dart
```diff
@@ -0,0 +1,30 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_module_server/src/generated/protocol.dart';
+
+/// A sample endpoint which is not built into the modules server or client
+/// Rather consuming app should subclass this one, and thus also gain the ability to override methods at will.
+@ignoreEndpoint
+class IgnoredModuleEndpoint extends Endpoint {
```
SandPod, line 7: Suggestion: Change this to an Abstract class since that is the pattern we are going to be using right?

tests/serverpod_test_server/lib/src/endpoints/module_endpoint_extension.dart
```diff
@@ -0,0 +1,59 @@
...
+import 'package:serverpod_test_module_server/serverpod_test_module_server.dart';
+
+/// Plain extension of the existing endpoint
+class ModuleEndpointSubclass extends IgnoredModuleEndpoint {}
+
+class ModuleEndpointAdaptation extends IgnoredModuleEndpoint {
+  @override
+  Future<String> echoString(Session sesion, String value) {
+    return super.echoString(sesion, value);
+  }
```
SandPod, line 11: Suggestion: Change the implementation here to echo in a different way, perhaps with a prefix that is validated.

I think this would be good to have in a to the point example as opposed to the next endpoint.

SandPod (commented):
Changes here looks good!

Just need to complement with tests for the analyzer since we can test the analyzing edgecases there.

We should also pair this with a documentation update.

We should document:
- `ignoreEndpoint` (should we introduce `ignoreMethod`?) on methods.
- Endpoint inheritance
- Endpoint inheritance from abstract classes.
- Versioning your API (Inheritance + deprecated annotation + changes in new API).

tp:
@SandPod Compared `generate`-ing the `serverpod_test_server`, here are the numbers:

This branch:
```
Enabling experimental feature: inheritance.
Enabling experimental feature: changeIdType.
✓ Generating code (11.9s)
✅ Done.

dart  --experimental-features=inheritance,changeIdType generate  25.25s user 2.23s system 140% cpu 19.622 total
```

`main`:
```
Enabling experimental feature: inheritance.
Enabling experimental feature: changeIdType.
✓ Generating code (11.5s)
✅ Done.

dart  --experimental-features=inheritance,changeIdType generate  24.59s user 1.93s system 153% cpu 17.307 total
```

In my test runs the "internal" timer difference looks bigger, but the same delta is applied on the overall time, and there it's much less noticeable. So I think we're good to go from that perspective. (But TBD if anyone thinks the test server should generate faster overall, though I would assume the main time-draw being the full-project analyze anyway.)

tp:
> We should also pair this with a documentation update.

Tracking the docs update here: https://github.com/serverpod/serverpod_docs/pull/268

> should we introduce `ignoreMethod`?

Not sure if it's worth the cognitive overhead of learning 2 annotations, but since the prefix is the same, it should be very discoverable for the developer.

If we pull in some helper from `meta`, we could mark each annotation to only be used on classes and methods respectively, so the project's linter could already enforce this:

CleanShot 2025-04-15 at 09 30 21@2x

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_class_analyzer.dart
tp: > was it possible to define an abstract endpoint method?

I was possible to define an `abstract class X extends Endpoint` and the code-gen would then generate everything as usual, but end up with a project which would not compile as there were attempts to create an instance like `X()` in the generated code.

> Use whitelist filter instead of blacklist for this.

Can you elaborate on this? I think we have to combine both: There are things you must do to be an endpoint (like extending `Endpoint`) and some thing you must not do (like being `abstract`).

Or am I misunderstanding the request? Was it just about renaming `ignoreAbstractClasses = true` to something like `supportAbstractClasses = false`?

I was not super happy with these parameters and also the fact that we have to switch behavior in super classes (hence the flexibility). Generally I like the pattern of having `false` as the default for `bool`s, but then I would have to rephrase it here a bit to keep the "default case" of checking the "leaf class" without any parameters.

Or we split the method into 2: Like "is valid endpoint" and "is valid endpoint base"?

tests/serverpod_test_module/serverpod_test_module_server/lib/src/endpoints/ignored_endpoint.dart
tp: `abstract` is already covered in the Login/Admin test case extensively, so I would vote to keep it like this.

Also if we add `abstract` we would not really need `@ignoreEndpoint` at all (and the test might pass for supporting either one).

So basically I think we support 2 ways and thus should test them each.

From our in-person discussions I still think `@ignoreEndpoint` might be handy as the default way (instead of abstract) because we could generate at least some test helpers in the future and also this ensures that all methods are implemented (non-`abstract`), so the class has a full implementation and is just not exposed on the server by default.

tests/serverpod_test_server/lib/src/endpoints/module_endpoint_extension.dart
tp: yeah makes sense, currently you have to read the test code very carefully to see that the right client is called. Though the error details we're looking for are a bit different (401 vs 403), but in the happy path there is no difference to observe.

Will adapt :)

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
tp: Added a test that methods marked with `@ignoreEndpoint` are skipped in the analyzer output.

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_class_analyzer.dart
tp: Added the test cases here, that e.g. an `abstract` class is skipped and a sub-class of an `abstract … Endpoint` is recognized as an endpoint class.

packages/serverpod/lib/serverpod.dart
SandPod: I think we concluded to keep this as is right? Now with this being an integral part of versioning, I think it makes sense to export it with `serverpod`.

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_class_analyzer.dart
SandPod: My comment was a bit out of place with the functionality of the method.

But you bring something important up. Could we get rid of the bools if we had a method called `isEndpointInterface`?  It seems like we only either call this with both values true or both values false right?

That method could then be used internally of this `isEndpointClass` method with that additional requirement that this needs to be a concrete class?

SandPod:
> should we introduce ignoreMethod?

Leets keep it simple for now and see if there is a need for a second annotation.

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_class_test.dart
```diff
@@ -217,4 +217,85 @@ class ExampleEndpoint extends Endpoint {
       expect(endpointDefinitions, isEmpty);
     });
   });
+
+  group('Given endpoint class marked as `abstract when analyzed', () {
```
SandPod, line 221: Suggestion(NIT-Typo) Change "\`abstract" to "\`abstract\`"

SandPod (commented):
Last suggestion around the validation then this is ready to ship.

Will take a look at the documentation during the day.

packages/serverpod/lib/serverpod.dart
tp: > I think we concluded to keep this as is right?

Yes, for now I kept the re-export of the annotations through the main `serverpod` package.

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_class_analyzer.dart
tp: Adapted the implementation to now provide 2 distinct methods.

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_class_test.dart
tp: fixed the typo

SandPod (approved):
LGTM! 👍

PR #3408, opened 2025-04-10
feat: Reduce templates footprint

dkbast:
This PR is part of updating the Getting Started guide. 

The key goal is to reduce complexity to make it easier to make your first steps in the framework. The PR also switches from "theoretical" examples to naming things "properly" trying to find a balance between verbose comments and "just showing" how it works in action.

Tasks
- [x] Update `projectname_<...>` template - remove layer first structure, rename example to greeter 
- [x] Update `projectname_server_upgrade`

Out of scope
Updating the module templates is out of scope for this PR


templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/// This is a simple class that generates a greeting message.
+/// It can be configured with a custom greeting message. E.g. if reusing between
+/// multiple endpoints.
+/// The `call` method takes a name and returns a `Greeting` object.
+/// When building endpoints, think of them as functions which take the
+/// input they are called with plus anything they fetch from the database etc
+/// and return a result. But they should not maintain any state between calls.
+class Greeter {
```
dkbast, line 32: We discussed if we want to remove this completely and just have the logic in the endpoint, even-though we want to push users to not write their logic in the endpoints but pull it out.

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,47 @@
...
+// After adding or modifying an endpoint, you will need to run
+// `serverpod generate` to update the server and client code.
+class GreetingEndpoint extends Endpoint {
+  // You create methods in your endpoint which are accessible from the client by
+  // creating a public method with `Session` as its first parameter.
+  // `bool`, `int`, `double`, `String`, `UuidValue`, `Duration`, `DateTime`, `ByteData`,
+  // and other serializable classes, exceptions and enums from your from your `protocol` directory.
+  // The methods should return a typed future; the same types as for the parameters are
+  // supported. The `session` object provides access to the database, logging,
+  // passwords, and information about the request being made to the server.
```
SandPod, line 18: Suggestion: This list needs to be updated. We support more data types nowadays.

templates/serverpod_templates/projectname_server/test/unit/greeter_test.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:test/test.dart';
```
dkbast, line 1: If we remove the logic abstraction, we could remove the unit test and just have the integration test in the template

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,47 @@
...
+  Future<Greeting> hello(Session session, String name) async {
```
dkbast, line 19: If we are not showing how to use the "Greeting" data model then we could remove the generated code from the template and run `serverpod generate` in the `create` command.

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,47 @@
...
+// `Endpoint` ending of the class name, but it will be removed when accessing
+// the endpoint from the client. I.e., this endpoint can be accessed through
+// `client.example` on the client side.
+
+// After adding or modifying an endpoint, you will need to run
+// `serverpod generate` to update the server and client code.
+class GreetingEndpoint extends Endpoint {
+  // You create methods in your endpoint which are accessible from the client by
+  // creating a public method with `Session` as its first parameter.
+  // `bool`, `int`, `double`, `String`, `UuidValue`, `Duration`, `DateTime`, `ByteData`,
```
dkbast, line 14: Remove the list of types and link to the documentation instead.

SandPod (commented):
Nice to get a refresh here.

templates/serverpod_templates/projectname_client/lib/src/protocol/client.dart
```diff
@@ -11,17 +11,19 @@
 // ignore_for_file: no_leading_underscores_for_library_prefixes
 import 'package:serverpod_client/serverpod_client.dart' as _i1;
 import 'dart:async' as _i2;
-import 'protocol.dart' as _i3;
+import 'package:projectname_client/src/protocol/greeting.dart' as _i3;
```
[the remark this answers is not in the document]
dkbast: Done - there is now an additional step running the generator and we don't check in generated files in the <project>_server folder. 

If we run the generate command for all templates then we could probably also remove the code in the client package and only keep the pubspec - wdyt @SandPod ?

templates/serverpod_templates/projectname_server/lib/src/greeting.spy.yaml
```diff
@@ -6,18 +6,21 @@
 # Please consult the documentation for more information on what you can add to
 # your yaml-files.
 
-
 # Name of the class to generate.
-class: Example
+class: Greeting
```
vlidholt, line 10: Perhaps we should add some `###` docs here to make the documentation for the model complete.

templates/serverpod_templates/projectname_server_upgrade/lib/server.dart
```diff
@@ -31,4 +29,38 @@ void run(List<String> args) async {
...
+  // there is also [futureCallAtTime] if you want to schedule a future call at a
+  // specific time.
+  await pod.futureCallWithDelay(
+    FutureCallNames.birthdayReminder.name,
+    Greeting(
+      message: 'Hello!',
+      author: 'Serverpod Server',
+      timestamp: DateTime.now(),
+    ),
+    Duration.zero,
```
vlidholt, line 57: This should be some duration. Say 5 or 10 seconds.

templates/serverpod_templates/projectname_server_upgrade/lib/server.dart
```diff
@@ -31,4 +29,38 @@ void run(List<String> args) async {
...
+  // background. Their schedule is persisted to the database, so you will not
+  // lose them if the server is restarted.
+
+  // If you are using any future calls, they need to be registered first.
+  pod.registerFutureCall(
+    BirthdayReminder(),
+    FutureCallNames.birthdayReminder.name,
+  );
+
+  // You can register future calls to be executed at startup. But you can also
```
vlidholt, line 46: This talks about registering, but should be about calling it. It's kind of less usual case to call this on startup (more common is to do it from an endpoint method, so perhaps point that out)?

templates/serverpod_templates/projectname_server_upgrade/lib/server.dart
```diff
@@ -31,4 +29,38 @@ void run(List<String> args) async {
...
+      timestamp: DateTime.now(),
+    ),
+    Duration.zero,
+  );
+}
+
+/// Names of all future calls in the server.
+/// This is better than using a string literal, as it will reduce the risk of
+/// typos and make it easier to refactor the code.
+enum FutureCallNames {
```
vlidholt, line 64: Question for other pro-reviewers. Is using an enum for this best practice?

vlidholt (commented):
Great work on this! :) I added some comments, but needs a review from @SandPod too. I would like to do a second pass after we run everything through Grammarly.

templates/serverpod_templates/projectname_server_upgrade/lib/server.dart
dkbast: After thinking about it for a night: You will probably prefer using a future call at startup rather than calling it directly, because you get a clean way to access the session, in that case Duration.zero would make sense and I think thats where I initially came from.

.fvmrc
```diff
@@ -0,0 +1,3 @@
+{
```
SandPod, line 1: Question: The documentation was a bit unclear, but what happens if I don't have this specific version or would like to use another one?

> flutter: The version of the Flutter SDK to be used, falling back to the flutter value if not explicitly set.

[source](https://fvm.app/documentation/getting-started/configuration#config-file-fvmrc)

templates/serverpod_templates/projectname_flutter/lib/main.dart
```diff
@@ -39,22 +41,24 @@ class MyHomePage extends StatefulWidget {
...
+  /// Holds the last error message that we've received from the server or null if no
+  /// error exists yet.
   String? _errorMessage;
 
   final _textEditingController = TextEditingController();
 
-  // Calls the `hello` method of the `example` endpoint. Will set either the
-  // `_resultMessage` or `_errorMessage` field, depending on if the call
-  // is successful.
+  /// Calls the `hello` method of the `example` endpoint. Will set either the
```
SandPod, line 53:
```suggestion
  /// Calls the `hello` method of the `greeting` endpoint. Will set either the
```

templates/serverpod_templates/projectname_flutter/lib/main.dart
```diff
@@ -100,13 +104,14 @@ class MyHomePageState extends State<MyHomePage> {
   }
 }
 
-// _ResultDisplays shows the result of the call. Either the returned result from
-// the `example.hello` endpoint method or an error message.
-class _ResultDisplay extends StatelessWidget {
+/// _ResultDisplays shows the result of the call. Either the returned result from
```
SandPod, line 107:
```suggestion
/// ResultDisplays shows the result of the call. Either the returned result from
```

templates/serverpod_templates/projectname_flutter/lib/main.dart
```diff
@@ -100,13 +104,14 @@ class MyHomePageState extends State<MyHomePage> {
...
+/// the `example.hello` endpoint method or an error message.
```
SandPod, line 108:
```suggestion
/// the `greeting.hello` endpoint method or an error message.
```

templates/serverpod_templates/projectname_flutter/lib/main.dart
SandPod: Might occur in other places, so please take a second look at the docs so the latest changes are reflected.

templates/serverpod_templates/projectname_server/lib/src/greeting.spy.yaml
```diff
@@ -6,18 +6,21 @@
...
-#table: example
+#table: greeting
 
-# The fields (and columns if connected to the database) of the class. Supported types are 
-# `bool`, `int`, `double`, `String`, `UuidValue`, `Duration`, `DateTime`, `ByteData`, 
-# and other serializable classes, exceptions and enums. 
-# You can also add lists of objects and types have support for null safety. 
+# The fields (and columns if connected to the database) of the class. Supported types are
+# `bool`, `int`, `double`, `String`, `UuidValue`, `Duration`, `DateTime`, `ByteData`,
+# `Uri`, `BigInt` and other serializable classes, exceptions and enums.
```
SandPod, line 17: Suggestion: This is not an exhaustive list of all supported types.

Perhaps it is better to refer do the documentation so that we don't need to remember to update this here?

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,28 @@
+import 'generated/protocol.dart';
+import 'package:serverpod/serverpod.dart';
+
+// This is an example endpoint of your server. It's best practice to use the
+// `Endpoint` ending of the class name, but it will be removed when accessing
+// the endpoint from the client. I.e., this endpoint can be accessed through
+// `client.example` on the client side.
```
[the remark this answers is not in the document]
SandPod: This has not been fixed.

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,30 @@
...
+// After adding or modifying an endpoint, you will need to run
+// `serverpod generate` to update the server and client code.
+
+/// This is an example endpoint that returns a greeting message.
+class GreetingEndpoint extends Endpoint {
+  // This method is called when the client calls the `hello` method on the
+  // `greeting` endpoint.
+  // The `Session` parameter is used to access the session information of the
+  // client that called the endpoint. Think of it as the `context` parameter in
+  // a Flutter widget.
```
SandPod, line 18: Suggestion: Change to something along the lines of "_The `Session` parameter is a context object for the client request. It provides access to the database and other server-side resources like secrets from your password file, the cache, storage, and server-event messaging_".

Should we mention authentication and session in some way?

templates/serverpod_templates/projectname_server_upgrade/lib/server.dart
SandPod: Seems like a resonable workaround for the current system.

since we don't need to create a switch case around the available options, you could use a `abstract final class` with `static const` fields that are the names as well. That would remove the redundant `<enum>.name` call.

templates/serverpod_templates/projectname_server_upgrade/lib/src/birthday_reminder.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:projectname_server/src/generated/protocol.dart';
+import 'package:serverpod/serverpod.dart';
+
+/// This is a simple example of a future call that sends a birthday reminder.
```
SandPod, line 4: Suggestion(NIT): Should the description match the implementation? Since we are actually logging and not sending a reminder?

tools/serverpod_cli/lib/src/create/generate_files.dart
```diff
@@ -0,0 +1,34 @@
...
+      return false;
+    }
+
+    var libDirectory = Directory(p.joinAll(config.libSourcePathParts));
+    var endpointsAnalyzer = EndpointsAnalyzer(libDirectory);
+
+    var yamlModels = await ModelHelper.loadProjectYamlModelsFromDisk(config);
+    var modelAnalyzer = StatefulAnalyzer(config, yamlModels, (uri, collector) {
+      collector.printErrors();
+    });
```
SandPod, line 26: Question: If we encounter errors here, should we return false? Since what we will be generating won't be what we expect?

SandPod (commented):
Nice to see a revamp of the starter template!

Nice if we can have this few files checked. Do you get any warnings when opening the Serverpod project?

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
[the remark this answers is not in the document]
dkbast: Good catch - looks like git auto reverted it

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
dkbast: Regarding auth: Let's wait for auth2 - iirc we wanted to start bundling that into the default template, then it would make sense to have it in here. When that lands we could have `serverpod create --empty` with a template that doesn't include auth, but bundle it by default.

templates/serverpod_templates/projectname_server_upgrade/lib/server.dart
dkbast: Yep, abstract classes are also a good option - if there is only the need for a safe identifier, but no need to control what it actually says I prefer enums over static strings, because static strings tempt some developers to copy them instead of accessing them through the abstract class.

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,28 @@
...
+// `Endpoint` ending of the class name, but it will be removed when accessing
+// the endpoint from the client. I.e., this endpoint can be accessed through
+// `client.example` on the client side.
+
+// After adding or modifying an endpoint, you will need to run
+// `serverpod generate` to update the server and client code.
+
+/// This is an example endpoint that returns a greeting message.
+class GreetingEndpoint extends Endpoint {
+  /// This method is called when the client calls the `hello` method on the
```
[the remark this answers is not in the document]
dkbast: Swapped this around - wdyt about it now?

.fvmrc
dkbast: It only has any effect if you are already using fvm - if not, nothing happens. If you are using fvm then you can change the version by callling "fvm use XXX" and that will change the version in the file so that you can check it in. There is also an option to use this for Codemagic ci builds and also for github actions to sync your local versions with whats used in the pipeline

tools/serverpod_cli/lib/src/create/generate_files.dart
dkbast: I was thinking of it and my initial thought was if generating fails, we might still want to be outputting the template - but you are right we probably want to fail this here - added a try throw catch flow.

.fvmrc
dkbast: ah - just realized, this is just for "internal" use, this will not be delivered to the client on `serverpod create` and is primarily so that we (in this case I) can make edits with our IDE and not use the new formatter etc.

tools/serverpod_cli/lib/src/create/generate_files.dart
```diff
@@ -0,0 +1,43 @@
...
+
+    var libDirectory = Directory(p.joinAll(config.libSourcePathParts));
+    var endpointsAnalyzer = EndpointsAnalyzer(libDirectory);
+
+    var yamlModels = await ModelHelper.loadProjectYamlModelsFromDisk(config);
+    late StatefulAnalyzer modelAnalyzer;
+    try {
+      modelAnalyzer = StatefulAnalyzer(config, yamlModels, (uri, collector) {
+        collector.printErrors();
+        throw Exception(
```
SandPod, line 28: Suggestion: Flip a bool here instead of throwing.

This allows us to print all errors and then simply return false based on that bool after the analyzer has completed.

I think that is the pattern we use internally as well to allow the analyzer to finish and collect as many issues as possible.

dkbast: I think it already does that, the throwing is just to escape the callback and not needing to introduce an "didFail" bool

.fvmrc
dkbast: It now reflects the oldest Flutter version we test against, this is to make sure we only use features that are available back then

tools/serverpod_cli/lib/src/create/generate_files.dart
SandPod: I don't think that is the case, this is the code that calls this callback, I think this will "throw" for each file: 

https://github.com/serverpod/serverpod/blob/0b666ec8da65e07e46f04b6913d2956970646097/tools/serverpod_cli/lib/src/analyzer/models/stateful_analyzer.dart#L131-L152

templates/serverpod_templates/projectname_flutter/lib/main.dart
```diff
@@ -2,12 +2,14 @@ import 'package:projectname_client/projectname_client.dart';
...
-// production servers.
-var client = Client('http://$localhost:8080/')
+/// Sets up a global client object that can be used to talk to the server from
+/// anywhere in our app. The client is generated from your server code.
+/// The client is set up to connect to a Serverpod running on a local server on
+/// the default port. You will need to modify this to connect to staging or
+/// production servers.
+/// You might want to use the dependency injection of your choice instead of
+/// using a global client object. This is just a simple example.
+final client = Client('http://$localhost:8080/')
```
dkbast, line 12: late String serverUrl;


void main() async {
  const urlFromEnv =
      String.fromEnvironment('SERVER', defaultValue: 'http://127.0.0.1:8080/');
  serverUrl = urlFromEnv;

dkbast: This would allow the user to easily pass in the local ip address instead of localhost which is needed when working with a physical device

dkbast:
Add log options do development config and add "all options to all config files but commented out"

util/generate_all
```diff
@@ -61,13 +61,15 @@ echo "\nmodulename_server"
 cd $BASE/templates/serverpod_templates/modulename_server
 dart $CLI generate --no-analytics
 
-echo "\nprojectname_server"
-cd $BASE/templates/serverpod_templates/projectname_server
-dart $CLI generate --no-analytics
+# We now generate files instead of checking them in so this
```
dkbast, line 64: @SandPod I think these tests no longer needed - we do test at other places that the files get generated instead, so this could be removed completely - wanted to keep a reminder that this is no longer necessary - wdyt?

tests/bootstrap_project/test/project_mini_test.dart
```diff
@@ -393,6 +393,20 @@ void main() async {
         reason: 'Docker with postgres failed to start.',
       );
 
+      var startProjectProcess = await startProcess(
```
dkbast, line 396: @SandPod this test is still failing and I don't really get why - it fails because the migrations have not been applied to the database and the tables cannot be found. I added this step which now applies the migrations, but it still fails. The old template didn't have any active future calls so it never needed to actually work with the database which might be why this was green before?

Isakdl: You are trying to apply the migrations in a Serverpod mini test. This will never work, as Serverpod mini doesn't have any database definitions and is not intended to work with a database. 

FutureCalls are not supported in Serverpod mini.

templates/serverpod_templates/projectname_server/lib/server.dart
```diff
@@ -3,7 +3,7 @@ import 'package:serverpod/serverpod.dart';
 import 'src/generated/protocol.dart';
 import 'src/generated/endpoints.dart';
 
-// This is the starting point of your Serverpod server.
+/// This is the starting point of your Serverpod server.
```
vlidholt, line 6:
```suggestion
/// The starting point of the Serverpod server.
```

To make it more dartdoccy.

templates/serverpod_templates/projectname_server/lib/src/greeting.spy.yaml
```diff
@@ -6,18 +6,19 @@
 # Please consult the documentation for more information on what you can add to
 # your yaml-files.
 
-
-# Name of the class to generate.
-class: Example
+### A greeting message which can be sent to the client
```
vlidholt, line 9:
```suggestion
### A greeting message which can be sent to or from the server.
```

templates/serverpod_templates/projectname_server/lib/src/greeting.spy.yaml
```diff
@@ -6,18 +6,19 @@
...
+# https://docs.serverpod.dev/concepts/working-with-endpoints
 fields:
-  name: String
-  data: int
\ No newline at end of file
+  ### The greeting message
+  message: String
+  ### The author of the answer
+  author: String
+  ### The time when the request was made
```
vlidholt, line 23: Missing new line.

templates/serverpod_templates/projectname_server/lib/src/greeting.spy.yaml
```diff
@@ -6,18 +6,19 @@
...
-# You can also add lists of objects and types have support for null safety. 
-# Eg. `List<int>?` or `List<MyOtherClass?>`.
+# The fields (and columns if connected to the database) of the class.
+# For a list of supported types, please see the documentation.
+# https://docs.serverpod.dev/concepts/working-with-endpoints
 fields:
-  name: String
-  data: int
\ No newline at end of file
+  ### The greeting message
```
vlidholt, line 19:
```suggestion
  ### The greeting message.
```

Let's be consistent with punctuations throughout our code and examples.

templates/serverpod_templates/projectname_server/lib/src/greeting.spy.yaml
```diff
@@ -6,18 +6,19 @@
...
+  message: String
+  ### The author of the answer
```
vlidholt, line 21:
```suggestion
  ### The author of the greeting message.
```

templates/serverpod_templates/projectname_server/lib/src/greeting.spy.yaml
vlidholt, line 23:
```suggestion
  ### The time when the message was created.
```

Technically not the time when the request was made as there is a delay from doing the request on the client.

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,29 @@
...
+
+// This is an example endpoint of your server. It's best practice to use the
+// `Endpoint` ending of the class name, but it will be removed when accessing
+// the endpoint from the client. I.e., this endpoint can be accessed through
+// `client.greeting` on the client side.
+
+// After adding or modifying an endpoint, you will need to run
+// `serverpod generate` to update the server and client code.
+
+/// This is an example endpoint that returns a greeting message.
```
vlidholt, line 12:
```suggestion
/// This is an example endpoint that returns a greeting message through its [hello] method.
```

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,29 @@
...
+class GreetingEndpoint extends Endpoint {
+  // This method is called when the client calls the `hello` method on the
+  // `greeting` endpoint.
+  //
+  // The `Session` parameter is a context object for the client request.
```
vlidholt, line 17:
```suggestion
  // The `Session` parameter contains the context of the client request.
```

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,29 @@
...
+  // It provides access to the database and other server-side resources like
+  // secrets from your password file, the cache, storage, and server-event messaging
```
vlidholt, line 19:
```suggestion
  // secrets from your password file, the cache, storage, and server-event messaging.
```

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,29 @@
...
+  //
+  // You can use any serializable type as a parameter or return type.
```
vlidholt, line 21:
```suggestion
  // You can use any serializable type as a parameter or return type,
```

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,29 @@
...
+  // Read more in the [docs](https://docs.serverpod.dev/concepts/working-with-endpoints).
```
vlidholt, line 22:
```suggestion
  // read more in the [docs](https://docs.serverpod.dev/concepts/working-with-endpoints).
```

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,29 @@
...
+
+  /// Returns a personalized greeting message: "Hello {name}!"
```
vlidholt, line 24:
```suggestion
  /// Returns a personalized greeting message: "Hello {name}".
```

templates/serverpod_templates/projectname_server/lib/src/greeting_endpoint.dart
```diff
@@ -0,0 +1,29 @@
...
+  Future<Greeting> hello(Session session, String name) async {
+    return Greeting(
+        message: 'Hello $name', author: 'Serverpod', timestamp: DateTime.now());
```
vlidholt, line 27:
```suggestion
        message: 'Hello $name', author: 'Serverpod', timestamp: DateTime.now(),);
```

Perhaps use a trailing comma for readability.

templates/serverpod_templates/projectname_server_upgrade/config/production.yaml
```diff
@@ -44,3 +45,11 @@ redis:
...
   host: redis.private-production.examplepod.com
   port: 6379
+  #user: # defaults to empty
+
+maxRequestSize: 524288 # The maximum size of requests allowed in bytes
+
+sessionLogs:
+  consoleEnabled: false
+# persistentEnabled: true
+# Serverpod configs, not under a namespace:
```
vlidholt, line 55: This row is unclear to me.

templates/serverpod_templates/projectname_server_upgrade/config/development.yaml
vlidholt, line 1: File is missing an ending line break.

templates/serverpod_templates/projectname_server_upgrade/lib/server.dart
vlidholt: Yes, that can be ok with an explanation. But this text need to be changed to talk about calling the future method (the previous section talks about registering it, here we are calling it).

Also, the `Duration.zero` clashes with what the "fictive" method is doing. If it's a birthday reminder, we want to post it with a delay. If its a setup task, we can call it immediately.

templates/serverpod_templates/projectname_server_upgrade/lib/src/birthday_reminder.dart
vlidholt, line 1: Food for thought. Should we include "future call" somehow in the class name or class name or is "reminder" clear enough?

templates/serverpod_templates/projectname_server_upgrade/lib/src/birthday_reminder.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:projectname_server/src/generated/protocol.dart';
+import 'package:serverpod/serverpod.dart';
+
+/// This is a simple example of a future call that logs a birthday reminder.
+///
+/// In a real-world application, you would implement the logic to send a
+/// an email or a push notification to the user.
+class BirthdayReminder extends FutureCall<Greeting> {
+  @override
+  Future<void> invoke(Session session, Greeting? object) async {
```
vlidholt, line 10:
```suggestion
  Future<void> invoke(Session session, Greeting? greeting) async {
```

Perhaps this is more clear?

vlidholt (commented):
These are great modifications! 🙌

I did a very nit-picky review. 😅 I'm good with the proposed changes, but some are up for discussion with a second reviewer. @SandPod

templates/serverpod_templates/projectname_server_upgrade/lib/src/birthday_reminder.dart
dkbast: Thats unfortunately not possible due to it overriding the method from FutureCall (dart avoid_renaming_method_parameters)

templates/serverpod_templates/projectname_server_upgrade/lib/src/birthday_reminder.dart
dkbast: Thats a good question - we should probably have a code style doc for naming stuff. Do we name it after what it "is" (a FutureCall) or what it does (BirthdayReminder) - we might be a bit inconsistent here: Endpoints are called Endpoint, models are just called by their names.

templates/serverpod_templates/projectname_flutter/lib/main.dart
```diff
@@ -2,15 +2,28 @@ import 'package:projectname_client/projectname_client.dart';
...
-// the default port. You will need to modify this to connect to staging or
-// production servers.
-var client = Client('http://$localhost:8080/')
-  ..connectivityMonitor = FlutterConnectivityMonitor();
+/// Sets up a global client object that can be used to talk to the server from
+/// anywhere in our app. The client is generated from your server code.
+/// The client is set up to connect to a Serverpod running on a local server on
+/// the default port. You will need to modify this to connect to staging or
+/// production servers.
+/// You might want to use the dependency injection of your choice instead of
```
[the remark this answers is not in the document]
dkbast: @vlidholt  I was wondering if we could provide a proper example with e.g. a service locator instead of promoting the use of global variables - what would the recommended approach actually be? Especially if we want to promote testing? cc @Isakdl @SandPod

templates/serverpod_templates/projectname_server_upgrade/lib/src/birthday_reminder.dart
dkbast: Also regarding the naming - I think "schedule" is more appropriate since a "call" is usually immediate.

util/generate_all
SandPod: If we remove them from there, won't the starter template project be red for anyone that clones the repo?

If I understand this correctly, a better solution might be to exclude those specific projects from the "check" if everything is generated, rather than not creating a runnable version of them locally when you run `generate_all`?

templates/serverpod_templates/projectname_flutter/lib/main.dart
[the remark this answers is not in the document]
SandPod: Isak went through the setup that we use internally right?

Since we expect that most of our users have some Flutter experience, I think we should keep the example as simple as possible here. Then we can focus our educational effort on the server side?

SandPod (commented):
This looks good to me now!

A couple of last minor comments, lets finish those discussions and then it is ready in my eyes.

Nice work!

util/generate_all
dkbast: Ah yes you are right

templates/serverpod_templates/projectname_flutter/lib/main.dart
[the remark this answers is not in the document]
dkbast: Agreed - let's keep it simple!

templates/serverpod_templates/projectname_client/doc/endpoint.md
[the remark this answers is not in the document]
dkbast: cc @SandPod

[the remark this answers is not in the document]
SandPod: I think we should leave it. If they generate the Dart docs for the client, we give a nice description for endpoints.

Question is if we should automatically generate the dart docs when running serverpod generate. But that is a different feature.

templates/serverpod_templates/projectname_server/analysis_options.yaml
```diff
@@ -19,10 +19,10 @@ include: package:lints/recommended.yaml
...
 #     - camel_case_types
 
-# analyzer:
-#   exclude:
-#     - path/to/excluded/files/**
-
+analyzer:
+  exclude:
+    - lib/src/generated/**
+    - test/integration/test_tools/serverpod_test_tools.dart
```
SandPod, line 25: Suggestion: Add this to the client as well right?

SandPod (approved):
LGTM given my last comment is fixed! :)

Maybe Viktor should also give a last 👍

templates/serverpod_templates/projectname_server/analysis_options.yaml
dkbast: Hmm - that leads me to the question of if we expect the user to change anything in the client and if not, then this would disable the linting for the client completely

dkbast: image

Because its not in a "generated" directory, I as a user would assume the whole package to be a "don't touch this" area and only add stuff to the pubspec.yaml

PR #3411, opened 2025-04-11
fix: Support record parameters named something other than "record"

tp:
So we support all names and not just literally "record".


tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -790,7 +790,7 @@ class LibraryGenerator {
           // The generated classes implement `ProtocolSerialization` and get handle by `serverpod_serialization` later
           // For the records we need to transform them into a map that can be handled by the shared (non-project specific) serialization code
           literalString(parameterDef.name): parameterDef.type.isRecordType
-              ? mapRecordToJsonRef.call([refer('record')]).code
+              ? mapRecordToJsonRef.call([refer(parameterDef.name)]).code
```
nielsenko, line 793: 😄 LGTM 🚀

nielsenko (approved):

SandPod (approved):
🚀

PR #3433, opened 2025-04-16
feat: Add new `serverpod_auth_user` package

### LINKED ISSUE #3445: Create `AuthUser` / `serverpod_auth_user` package
Create an `AuthUser` (just containing `enabled` and `scopes` from the current `UserInfo`) to act as the anchor point of the new "Auth 2.0" packages.

tp:
Introduces the `AuthUser` type, the new "core" / anchor point for the modular "authentication 2.0" packages.

This type is a subset of the `UserInfo` (from `serverpod_auth`), containing only the required fields for authentication in the core Serverpod framework: An ID, scopes, and whether it's enabled.  
The other fields of `UserInfo` will move to a `serverpod_user_profile` package.

This type offers the basis for all other new packages to hook into and connect to this user.

The package does not and is not expected to expose any endpoints.

Closes #3445


modules/serverpod_auth2/serverpod_auth_user/serverpod_auth_user_client/.gitignore
```diff
@@ -0,0 +1,13 @@
+# Files and directories created by pub
+    .dart_tool/
```
tp, line 2: This extra padding seems to have snuck into the template. I can fix it there and in all affected packages in a follow-up :)

modules/serverpod_auth2/serverpod_auth_user/serverpod_auth_user_client/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: serverpod_auth_user_client
```
SandPod, line 1: Suggestion: Can we rename the folder to `new_serverpod_auth`?

I think that is a more appropriate name.

modules/serverpod_auth2/serverpod_auth_user/serverpod_auth_user_client/lib/src/protocol/auth_user.dart
```diff
@@ -0,0 +1,113 @@
+/* AUTOMATICALLY GENERATED CODE DO NOT MODIFY */
```
SandPod, line 1: Suggestion: This should be added to all the appropriate scripts that manage all the packages in the repo.

- generate
- testing
- updating pubspecs and templating.

And all the tests that validate this project do not get desynchronized.

We should also add the serverpod_lints rules and make sure that we follow it.

SandPod (changes requested):
Nice!

Some more setup is needed to ensure that this package is managed in the mono repo.

modules/serverpod_auth2/serverpod_auth_user/serverpod_auth_user_client/lib/src/protocol/auth_user.dart
tp: Done.

> We should also add the serverpod_lints rules and make sure that we follow it.

added them, but just like `serverpod_auth` we had to include a bunch of exceptions from those rules, because the generated code would conflict with those. Maybe we should either rework them or see that we can use them verbatim? E.g. `final` parameters being in there surprised me, as we don't even us `final` inside functions.

modules/serverpod_auth2/serverpod_auth_user/serverpod_auth_user_client/.gitignore
tp: Done in https://github.com/serverpod/serverpod/pull/3461

modules/serverpod_auth2/serverpod_auth_user/serverpod_auth_user_client/lib/src/protocol/auth_user.dart
SandPod: I think it is possible to exclude certain directories from certain rules by doing something like this.

Inside of `analysis_options.yaml` you can add
```yaml
analyzer:
  exclude:
    "<path to directory of file>"
```

templates/pubspecs/modules/new_serverpod_auth/serverpod_auth_user/pubspec.yaml
```diff
@@ -0,0 +1,9 @@
+name: serverpod_auth_user_client
```
SandPod, line 1: Suggestion: This file looks like it is placed in the wrong place?

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_client/lib/src/protocol/auth_user.dart
```diff
@@ -0,0 +1,113 @@
...
+// ignore_for_file: library_private_types_in_public_api
+// ignore_for_file: non_constant_identifier_names
+// ignore_for_file: public_member_api_docs
+// ignore_for_file: type_literal_in_constant_pattern
+// ignore_for_file: use_super_parameters
+
+// ignore_for_file: no_leading_underscores_for_library_prefixes
+import 'package:serverpod_client/serverpod_client.dart' as _i1;
+
+abstract class AuthUser implements _i1.SerializableModel {
```
SandPod, line 14: Question: Do we want this class to be exposed on the client?

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_client/CHANGELOG.md
```diff
@@ -0,0 +1,620 @@
+## 2.5.0
```
SandPod, line 1: Suggestion: We now have a new version in main. Merge `main` with the branch and run `update_pubspecs` to have these files updated.

SandPod (commented):
A couple of minor things to fix.

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_client/CHANGELOG.md
tp: Updated the CHANGELOGs

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_client/lib/src/protocol/auth_user.dart
tp: Just learned about `serverOnly: true`, nice :) Now hiding it for the clients.

templates/pubspecs/modules/new_serverpod_auth/serverpod_auth_user/pubspec.yaml
tp: Indeed, moved it.

SandPod (approved):
LGTM! 👍 🚀

Issue #3429, opened 2025-04-16
Use non-nullable id field when defaultModel is set.


SandPod:
### Problem to Solve

When `defaultModel` is used to define the id column it will always be initialized regardless of if it is provided or not. Therefore, there is no need to have this field nullable in the model.

### Proposal

Emit a warning or force a change during code generation or validation if `defaultModel: true` is used and the `id` field is nullable. This encourages users to make the `id` field non-nullable, even though it's currently technically allowed to be nullable.

### Use Case

Accessing a non-nullable field through a nullable getter just creates unnecessary null checks in the code.

### Alternatives

N/A

### Additional context

This helps guide developers toward safer usage patterns.

### How experienced are you with this library?

None

### Are you interested in working on a PR for this?

- [ ] I want to work on this

marcelomendoncasoares:
I can take this on. Just don't know if we need the warning, as it does not exist today for the usage with `defaultModel` in regular fields. I think we need to allow non-nullable - which will probably be everyone's first choice -, but I would leave the developer free to declare it as nullable with no warning, just as for other fields. Because someone could use it intentionally to force a `obj.id = null` after creation in some use case.

SandPod:
True, maybe we just add a simple hint?

PR #3437, opened 2025-04-17
feat: Allow non-nullable id field when `defaultModel` is set

marcelomendoncasoares:
This PR implements the possibility to declare non-nullable id fields, as specified on #3429. This is a follow-up from the #3416 PR.

To declare a field as non-nullable, just use the `defaultModel` keyword instead of the `defaultPersist` (only valid for UUID). It is still possible to declare nullable with `defaultModel`, in which case a hint will remind the user that it could be `non-nullable` to avoid unnecessary null checks.

## Pre-launch Checklist

## Breaking changes

No breaking changes, only the need to re-generate the models - as the `TableRow.id` field is no longer forcefully nullable. Instead, the nullability is declared in the generics of the class that extends it.


tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -204,6 +203,13 @@ class SerializableModelLibraryGenerator {
                 classDefinition,
               ),
           ]);
+
+          if (buildRepository.hasAttachOperations(fields) ||
+              buildRepository.hasAttachRowOperations(fields) ||
+              buildRepository.hasDetachOperations(fields) ||
+              buildRepository.hasDetachRowOperations(fields)) {
+            libraryBuilder.ignoreForFile.add('unnecessary_null_comparison');
```
marcelomendoncasoares, line 211: For relations with models with non-nullable ids, some null checks became unnecessary, such as the one below:

```dart
class CustomerIntAttachRowRepository {
  const CustomerIntAttachRowRepository._();

  /// Creates a relation between this [CustomerInt] and the given [OrderUuid]
  /// by setting the [OrderUuid]'s foreign key `customerId` to refer to this [CustomerInt].
  Future<void> orders(
    _i1.Session session,
    CustomerInt customerInt,
    _i2.OrderUuid orderUuid, {
    _i1.Transaction? transaction,
  }) async {
    if (orderUuid.id == null) {  // <-- unnecessary_null_comparison
      throw ArgumentError.notNull('orderUuid.id');
    }
    if (customerInt.id == null) {
      throw ArgumentError.notNull('customerInt.id');
    }

    var $orderUuid = orderUuid.copyWith(customerId: customerInt.id);
    await session.db.updateRow<_i2.OrderUuid>(
      $orderUuid,
      columns: [_i2.OrderUuid.t.customerId],
      transaction: transaction,
    );
  }
}
```

I first tried to remove those checks individually by looking at the type of the id of the model inside each `_build*ImplementationBlock*` method of the `model_library_generator.dart`. However, due to the way that the code is structured, it got really hard to get somewhere. As ignoring the `unnecessary_null_checks` in this case is somewhat harmless, it seemed better to me to just add it when needed than to potentially delay this feature a lot trying to remove the checks.

SandPod: That sounds reasonable, but could you file an issue for this so that we can clean this up at some point?

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -203,6 +203,13 @@ class SerializableModelLibraryGenerator {
                 classDefinition,
               ),
           ]);
+
+          if (buildRepository.hasAttachOperations(fields) ||
+              buildRepository.hasAttachRowOperations(fields) ||
+              buildRepository.hasDetachOperations(fields) ||
+              buildRepository.hasDetachRowOperations(fields)) {
```
SandPod, line 210: Suggestion: Wrap these checks in a function with a name that explains why we add the `unnecessary_null_comparison`. 

You can then reference the issue in there as a comment that this can be removed once the issue is resolved.

tests/serverpod_test_server/migrations/20250417130734022/migration.sql
```diff
@@ -0,0 +1,46 @@
...
+--
+--
+-- ACTION ALTER TABLE
+--
+--
+-- ACTION ALTER TABLE
+--
+
+--
+-- MIGRATION VERSION FOR serverpod_test
```
SandPod, line 14: Question: Why is this migration needed? It doesn't seem like we do any modifications to the database?

SandPod (commented):
Nice clean PR with structured commits.

The last commit had my questioning the migration system. why was it needed? And are we missing some migration steps in our SQL migration generator?

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
marcelomendoncasoares: Sure! As soon as we close this PR I open the issue so I can reference the code properly.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
marcelomendoncasoares: Done.

tests/serverpod_test_server/migrations/20250417130734022/migration.sql
marcelomendoncasoares: Good question... I also did not understand at first, as the models supposedly did not change from the perspective of the database. The three models referenced in the migration are the ones whose id type passed from nullable to non-nullable, but the type was already non-nullable on the database.

I took a closer look now and found the issue to be that lines 188 to 190 from `migration.dart` assumed that either one of the three conditions must be met if the column passes the `canMigrateTo` test:

```dart
// Column can be modified
var addNullable = !srcColumn.isNullable && dstColumn.isNullable;
var removeNullable = srcColumn.isNullable && !dstColumn.isNullable;
var changeDefault = srcColumn.columnDefault != dstColumn.columnDefault;
```

However, `id` column is an exception, because the column will always be non-nullable (even with nullable type and only `defaultModel` key declared).  So `srcColumn.like(dstColumn)` will result `false` due to the different `dartType` (one is `UuidValue?` and other `UuidValue`), but there will have no difference in the database. Just added a condition and a test to fix this.

tools/serverpod_cli/test/database/migration/default_value_test.dart
```diff
@@ -221,4 +221,66 @@ void main() {
       expect(migration.actions, isEmpty);
     });
   });
+
+  group(
+      'Given a source and target definition with a table changing the id field on the model to not-nullable',
+      () {
```
SandPod, line 226: Question: This makes me wonder what happens if we change the ID type of a model. Does our migration system handle that? And do we have a test for it?

SandPod (approved):
LGTM 👍

tests/serverpod_test_server/migrations/20250417130734022/migration.sql
SandPod: Thank you for fixing and testing this 👍

PR #3439, opened 2025-04-19
refactor(serverpod_flutter): Avoid using conditional import for localhost

### LINKED ISSUE #3438: Avoid using conditional import in serverpod_flutter
Currently, the package [`serverpod_flutter`](https://pub.dev/packages/serverpod_flutter) uses conditional import for the getter `localhost` to use `dart:io` and check if the platform is Android on an Android app (not using the Flutter web or the browser app). While the current solution work fine, it can be refactored to simplify it, and also make it testable, using only `defaultTargetPlatform` and `kIsWeb` in one file without the io version and the web, if not correctly exported or imported the io or the web file by an accient, it can lead to a build failure.

This is a super low-priority issue. The [related line](https://github.com/serverpod/serverpod/blob/main/packages/serverpod_flutter/lib/serverpod_flutter.dart#L2).

EchoEllet:
_Avoid using conditional import in the `serverpod_flutter` package and adding simple unit tests to cover `localhost` getter_

- *Fix #3438*

## Pre-launch Checklist

### DIFF
--- a/packages/serverpod_flutter/test/localhost_test.dart
+++ b/packages/serverpod_flutter/test/localhost_test.dart
@@ -3,28 +3,26 @@ import 'package:flutter_test/flutter_test.dart';
 import 'package:serverpod_flutter/src/localhost.dart';
 
 void main() {
-  group('localhost', () {
-    tearDown(() => debugDefaultTargetPlatformOverride = null);
+  tearDown(() => debugDefaultTargetPlatformOverride = null);
 
-    test(
-        'Given target platform is android when reading localhost then 10.0.2.2 is returned',
-        () {
-      debugDefaultTargetPlatformOverride = TargetPlatform.android;
+  test(
+      'Given target platform is android when reading localhost then 10.0.2.2 is returned',
+      () {
+    debugDefaultTargetPlatformOverride = TargetPlatform.android;
 
-      expect(localhost, '10.0.2.2');
-    });
-
-    final nonAndroidPlatforms = TargetPlatform.values.toList()
-      ..removeWhere((platform) => platform == TargetPlatform.android);
-
-    for (final testPlatform in nonAndroidPlatforms) {
-      test(
-          'Given target platform is ${testPlatform.name} when reading localhost then localhost is returned',
-          () {
-        debugDefaultTargetPlatformOverride = testPlatform;
-
-        expect(localhost, 'localhost');
-      });
-    }
+    expect(localhost, '10.0.2.2');
   });
+
+  final nonAndroidPlatforms = TargetPlatform.values.toList()
+    ..removeWhere((platform) => platform == TargetPlatform.android);
+
+  for (final testPlatform in nonAndroidPlatforms) {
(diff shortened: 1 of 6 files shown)


EchoEllet:

packages/serverpod_flutter/test/localhost_test.dart
```diff
@@ -0,0 +1,25 @@
...
+    } finally {
+      debugDefaultTargetPlatformOverride = null;
+    }
+  });
+
+  test('returns localhost on other platforms', () {
+    try {
+      debugDefaultTargetPlatformOverride = TargetPlatform.linux;
+
+      expect(localhost, 'localhost');
```
EchoEllet, line 20: It might be useful to run this on all platforms except Android instead of only Linux.

packages/serverpod_flutter/test/localhost_test.dart
```diff
@@ -0,0 +1,29 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:serverpod_flutter/src/localhost.dart';
+
+void main() {
+  test('returns 10.0.2.2 on Android apps', () {
```
SandPod, line 6: Suggestion: Format these tests following the "Given when then" pattern.

```suggestion
  test('Given target platform is android when reading localhost then 10.0.2.2 is returned', () {
```

packages/serverpod_flutter/test/localhost_test.dart
```diff
@@ -0,0 +1,29 @@
...
+    try {
+      debugDefaultTargetPlatformOverride = TargetPlatform.android;
+
+      expect(localhost, '10.0.2.2');
+    } finally {
+      debugDefaultTargetPlatformOverride = null;
+    }
```
SandPod, line 13: Suggestion: Place in a group and use setup and teardown to set and reset `debugDefaultTargetPlatformOverride`

packages/serverpod_flutter/test/localhost_test.dart
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  test('returns localhost on other platforms', () {
```
SandPod, line 16: Suggestion: Define the loop outside of the test so that we get one test run with a unique description for each case using the Given, When, Then pattern.

Setup and teardown should be used here as well to reset the test.

SandPod (commented):
Thanks for creating this improvement!

Please fix the tests so that they match the style for the rest of the Serverpod repository.

packages/serverpod_flutter/test/localhost_test.dart
EchoEllet: This can be achieved without a group? Although I will use one.

SandPod: Yes, you are correct! I think it might even be nicer without the group.

I can push that fix to the branch and then we can merge it in 👍

PR #3441, opened 2025-04-22
feat(auth2): Add `serverpod_auth_profile` package

### LINKED ISSUE #3447: Create `UserProfile` / `serverpod_auth_profile` package
Create a new `UserProfile` entity which contains all the user profile information from `UserInfo` which are not part of the new `AuthUser` (from #3445).

This will be the
- Full name
- Username
- Profile image (incl. file upload)

tp:
Add `UserProfile` model, which matches the user details from `serverpod_auth`'s `UserInfo` class, including the profile image (avatar).

The package includes an endpoint to fetch the profile and update individual parts (full name, username, profile image).

The core business logic is abstracted in a `UserProfiles` class, so it can be reused outside of the default endpoint

Other than in the old `serverpod_auth`, the user image is stored directly on the user (TBD, maybe we should keep that split, and then work with references).

The configuration is kept as a singleton with defaults (like in `serverpod_auth`), only containing the relevant fields of the user profile.

Closes #3447


modules/serverpod_auth2/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,218 @@
...
+        userInfo,
+        lifetime: UserProfileConfig.current.userInfoCacheLifetime,
+      );
+    }
+
+    return userInfo;
+  }
+
+  /// Updates a users name.
+  static Future<UserProfile> changeUserName(
```
tp, line 101: Other than the `serverpod_auth` methods, these would now throw if the user is not found.

As I don't think there is any reasonable case where the user account "just disappeared" while the request was being worked on, and it seems to easy to e.g. confuse the "user profile ID" and "user ID", so we really ought to give some feedback if the update went nowhere.

modules/serverpod_auth2/serverpod_auth_profile/serverpod_auth_profile_client/.gitignore
```diff
@@ -0,0 +1,13 @@
+# Files and directories created by pub
```
tp, line 1: TODO: Before un-drafting, move module to `new_serverpod_auth` folder
(would just be tricky with the split commits right now, as this was started before that other folder was decided upon.)

modules/serverpod_auth2/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,218 @@
...
+    final UuidValue userId,
+  ) {
+    return 'serverpod_auth_profile_$userId';
+  }
+
+  /// Find a user by its id. Returns null if no user is found. By default the
+  /// result is cached locally on the server. You can configure the cache
+  /// lifetime in [UserProfileConfig], or disable it on a call to call basis by
+  /// setting [useCache] to false.
+  static Future<UserProfile?> findUserByUserId(
```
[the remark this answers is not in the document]
dkbast: ❓Could throwing be the better "not found" response in that case?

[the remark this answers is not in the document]
tp: @dkbast With this one it really is a bit tricky.

I would say it depends on when the profile module is added. In case it's added initially, you'd expect a user profile for everyone. In case it's added later, you expect `null` sometimes (and then would create it).

And then, while the control-flow of try/catch with creation is a bit weird, it's even weirder to always use `!` etc. when you actually expect it to be there.

So in the end I would also agree with you, that we should throw here, and for the "defensive" case we could still offer a `tryFindUser…` or similar if it's needed. Or, if it's just really during migration, we just handle it there once.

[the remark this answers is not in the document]
dkbast: ok, yep that makes sense to me :) - just for reference: Flutter and Dart use try and maybe for nullable methods alongside those where you expect something with certainty eg https://api.flutter.dev/flutter/dart-core/int/tryParse.html and https://api.flutter.dev/flutter/widgets/View/maybeOf.html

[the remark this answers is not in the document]
tp: While I also know `maybe*` more from the Widget-layer, I think it's the better approach here.

For `try` I have the mental model: No matter what goes wrong, this will return `null` if it can not figure it out (as it's mostly in simple cases like `int.tryParse` where the only thing going wrong could be the String not being an integer…). But for out case I would want it to just return `null` in case no user profile is found and throw in all the other error cases like DB not being reachable, model not parseable etc. So no global `try/catch` around it.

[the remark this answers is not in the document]
tp: Though to we then also use `maybeFindUserProfileByEmail`, which is a function where I would expect a match in the minority of cases (lookup of existing accounts during registration only).

modules/serverpod_auth2/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,155 @@
...
+  /// Creates a new user and stores it in the database.
+  static Future<UserProfile> createUserProfile(
+    final Session session,
+    UserProfile userProfile,
+  ) async {
+    userProfile = (await UserProfileConfig.current.onBeforeUserProfileCreated
+            ?.call(session, userProfile)) ??
+        userProfile;
+
+    userProfile = userProfile.id == null
```
tp, line 15: TBD: acceptable pattern? Else the hook could only "augment" but not "redirect".

modules/serverpod_auth2/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
```diff
@@ -0,0 +1,219 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_profile_server/serverpod_auth_profile_server.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+import 'package:test/test.dart';
+
+import '../../test/integration/test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod(
```
tp, line 9: as in `serverpod_auth_session`, we need to find a way to run them in CI.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,155 @@
...
+  /// Finds a user profile by its email address.
+  ///
+  /// Returns `null` if no user profile is found.
+  static Future<UserProfile?> maybeFindUserProfileByEmail(
+    final Session session,
+    final String email,
+  ) async {
+    return await UserProfile.db.findFirstRow(
+      session,
+      where: (final t) => t.email.ilike(email),
```
tp, line 36: This seems like it could get too query-heavy.

So we should either:
- Always lower-cases the email on account creation, and then lower-case all queries
- Store both the original input and lowercase (normalized?) version (Would be interesting, if there is a popular library that would also handle `@googlemail` vs `@gmail` etc.)

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  /// Finds a user profile by its email address.
+  ///
+  /// Returns `null` if no user profile is found.
+  static Future<UserProfile?> maybeFindUserProfileByEmail(
+    final Session session,
+    final String email,
+  ) async {
+    return await UserProfile.db.findFirstRow(
```
tp, line 34: This needs to be changed to find all rows and then handle the case where it finds 2 or more (as the email is not unique).

Return value TBD, should it just be `null`, or a special error case?

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/models/user_profile.spy.yaml
```diff
@@ -0,0 +1,30 @@
...
+  ### The [AuthenticationUser]'s id.
+  userId: UuidValue, relation(parent=serverpod_auth_user)
+
+  ### The first name of the user or the user's nickname.
+  userName: String?
+
+  ### The full name of the user.
+  fullName: String?
+
+  ### The email of the user.
```
tp, line 15:
```suggestion
  ### The verified email address of the user.
  ###
  ### This should only be set by authentication providers that have
  ### checked ownership of this email for the user.
```

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/models/user_profile.spy.yaml
```diff
@@ -0,0 +1,30 @@
...
+  email: String?
```
tp, line 16: In `serverpod_auth` and current `serverpod_auth_email_account` we store the email as lowercase for later look up.

TIL that technically this is not correct, but in practice just fine: https://blog.teknkl.com/email-addresses-case-sensitive-but-anyway/

We should probably enforce the same rule here (on the model layer, not just convention), so that the look-up actually works case-insensitively.

tp: I checked back, and currently it's all lower case storage and lookup.

That probably makes the most sense going forward as well, as email addresses are practically case-insensitive.

With ILIKE we have to be careful with the wildcards, of which "_" is one, which is also part of email addresses. Will have to check how they behaves through our query layer.

Ideally I would create an index on the lower case variant of the email field, training the original, but I don't think that this is an option right now... and using "LOWER(email) = $email" might be slow at runtime (full table scan).

modules/serverpod_auth2/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,136 @@
...
+  /// Finds a user profile by its email address.
+  ///
+  /// Returns `null` if no user profile is found.
+  static Future<UserProfile?> findUserByEmail(
+    final Session session,
+    final String email,
+  ) async {
+    return await UserProfile.db.findFirstRow(
+      session,
+      where: (final t) => t.email.ilike(email),
```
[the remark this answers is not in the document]
tp: Has been changed to always used lower-case email storage, just like the email provider, so we don't have to do any tricks during look-up.

modules/serverpod_auth2/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
tp: CI added, as this is now support in `main`.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
tp: lower-case full-string matching it is for now :)

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
tp: changed to only return profile if there is 1 match.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/models/user_profile.spy.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+  ### The time when this user was created.
+  created: DateTime, default=now
+
+  ### The user's profile image
+  image: UserProfileImage?, relation(optional)
+
+indexes:
+  serverpod_auth_profile_email:
+    fields: email
```
tp, line 32: Currently, email uniqueness is not enforced by the module.

But look-up via email only succeeds if there is exactly 1 profile with that email (thus the developer would have enforced this so far and keeps enforcing it with the lookup).

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/endpoints/user_profile_endpoint.dart
```diff
@@ -0,0 +1,104 @@
...
+
+  /// Returns the user profile of the current user.
+  Future<UserProfileModel> get(final Session session) async {
+    final authUserId = (await session.authenticated)!.userUuid;
+
+    final profile = await UserProfile.db.findFirstRow(
+      session,
+      where: (final t) => t.authUserId.equals(authUserId),
+      include: UserProfile.include(image: UserProfileImage.include()),
+    );
```
SandPod, line 21: Suggestion: Add this to the business layer and call into it.

I guess this functionality will be useful for anyone working with user profiles as well?
Might as well provide all that is needed so they don't have to do db calls directly.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/endpoints/user_profile_endpoint.dart
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    final trimmedUserName = userName.trim();
+    if (trimmedUserName == '') {
+      throw ArgumentError.value(
+        userName,
+        'userName',
+        'User name must not be an empty or whitespace-only string.',
+      );
+    }
```
SandPod, line 75: Suggestion: Move this validation to the business layer.

If that is a requirement for the userName this should be reflected there.
Add tests for this as well.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/endpoints/user_profile_endpoint.dart
```diff
@@ -0,0 +1,104 @@
...
+
+    final trimmedFullName = fullName.trim();
+    if (trimmedFullName == '') {
+      throw ArgumentError.value(
+        fullName,
+        'fullName',
+        'Full name must not be an empty or whitespace-only string.',
+      );
+    }
```
SandPod, line 98: Suggestion: Move to business layer with tests.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,164 @@
...
+  /// Returns `null` if no profile is found. By default the
+  /// result is cached locally on the server. You can configure the cache
+  /// lifetime in [UserProfileConfig], or disable it on a call to call basis by
+  /// setting [useCache] to false.
+  static Future<UserProfile?> maybeFindUserByUserId(
+    final Session session,
+    final UuidValue authUserId, {
+    final bool useCache = true,
+  }) async {
+    final cacheKey = _userProfileCacheKey(authUserId);
```
SandPod, line 73: Suggestion: I don't see any tests that validate the cache works as expected.

Please ensure this functionality works as expected. If the cache would malfunction new updates to user profiles might not be added.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,164 @@
...
+    final String newUserName,
+  ) async {
+    var userProfile = await findUserProfileByUserId(
+      session,
+      userId,
+      useCache: false,
+    );
+
+    userProfile.userName = newUserName;
+    userProfile = await UserProfile.db.updateRow(session, userProfile);
```
SandPod, line 110: Suggestion: Move this into a private method and group it with the `_invalidateCacheForUser` call.

So that we ensure that they always happen together.

SandPod (changes requested):
Couple of questions and comments.

Please look through the tests and validate that we cover the functionality the module provides. Missing tests in the old modules made us introduce easy to catch bugs previosly. Might as well ensure that this is well behaved and tested from the start.

Also wonder if we can't remove the email lookup?

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/models/user_profile.spy.yaml
tp: > This prevents a single user from having multiple accounts but displaying the same email. Do we really want that?

Currently this is not enforced by the schema. But the email endpoint would be using this lookup to find an existing account by email in case the sign-up happens as a guest.

I think this is just he user consolidation as I understood we had it.

(It might be debatable whether the consolidation should happen via the profile (which also be for display etc. and can only hold 1 email at a time) or by connecting all the auth providers.)

So the question is, did anything regarding email based user consolidation change, or would you rather just see it done "outside" of the profile?

util/recreate_all_migrations
```diff
@@ -29,12 +29,21 @@ cd $BASE/modules/serverpod_auth/serverpod_auth_server
 rm -r $MIGRATION_DIR
 dart $CLI create-migration --no-analytics
 
-
 echo "\nserverpod_auth_user_server"
 cd $BASE/modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server
 rm -r $MIGRATION_DIR
 dart $CLI create-migration --no-analytics
 
+echo "\nserverpod_auth_session_server"
```
tp, line 37: @SandPod Am I missing something or is this script never used?

Should this not be invoked in CI like the code generation, to see that no module forgot to generate anything?

We should probably though handle "no migrations needed" better then, as currently we can not distinguish whether it fails or whether no change was done, and just looking at the file system later (git diff) would not tell us that. (That's why I would set all bash scripts to exit on the first error, while these often continue IIIRC.)

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profile_config.dart
```diff
@@ -0,0 +1,102 @@
...
+typedef UserImageGenerator = Future<Image> Function(UserProfile userprofile);
+
+/// Defines the format of stored user images.
+enum UserProfileImageType {
+  /// PNG image format.
+  png,
+
+  /// JPG image format.
+  jpg,
+}
```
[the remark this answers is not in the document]
tp: Currently (as was the behavior with the old implementation) **every** image is re-encoded into this format and the configured `userImageSize` (apart from images having the final dimensions already, which is an exception probably only added for the generated avatars).

If we do support PNG and JPG though, I think we could also retain each in their original (suitable) format.
As for the resizing, that probably makes sense, as our limit is so small, that each normal user upload would be resized anyway (and then we can at least ensure that the right compression and quality is attached, so uploads don't grow too large).

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profile_images.dart
```diff
@@ -0,0 +1,202 @@
...
+    // Pick color based on user id from the default colors (from material design).
+    final color = _defaultUserImageColors[
+        userInfo.id!.hashCode % _defaultUserImageColors.length];
+    fill(image,
+        color: ColorUint8.rgba((color >> 16) & 0xff, (color >> 16) & 0xff,
+            color & 0xff, (color >> 24) & 0xff));
+
+    // Draw the character on top of the solid filled image.
+    drawString(
+      image,
```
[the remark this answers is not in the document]
tp: The background color is also dynamic (currently 14 options).

Did this ever come up in profiling? We sure could cache the images, but since they are rather small I wouldn't expect the current approach to lead to any congestion.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,164 @@
...
+  /// Creates a new user and stores it in the database.
+  static Future<UserProfile> createUserProfile(
+    final Session session,
+    UserProfile userProfile,
+  ) async {
+    userProfile = userProfile.copyWith(
+      email: userProfile.email?.toLowerCase().trim(),
+    );
+
+    userProfile = (await UserProfileConfig.current.onBeforeUserProfileCreated
```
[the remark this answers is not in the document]
tp: > Should it be possible to throw in the onBefore... to prevent a profile from being created?

I think so. It'd be weird if you implemented a "pipeline" step and then we would just skip over it if it errs.

I think this is important because this allows the developer to err on purpose (validation as you suggested), or in case their business logic is not sound (or runs into an error, e.g. timeout), it would at least stop the creation of the profile instead of proceeding with a "not properly set up" one.

(This is the benefit of the "pipeline" approach, rather than just a callback. There I am still surprised that listeners on e.g. a `ChangeNotifier` to not run even in `try/catch`, so even though they are usually independent (but depend on registration order of course), one failing stops the rest…)

> Maybe we should export a set of serializable exceptions that should be used in that case? Or do we leave this up to the developer?

I think this has to be left to the developer to some degree, at least when they want "nice" exceptions, e.g. with an enum error case for their specific validation.

If you think there is a benefit to providing the client with additional info that a hook (or maybe even with part of the hook) failed, then we could of course add a fallback `SerializableException` to use in case the developer just gave us a non-specific error.

In practice I would think though, that these errors (unless the developer throws an explicit "name not allowed" etc.) would be debugged from the backend anyway, where all the info is available (and we should aim not to loose any by wrapping it in a "simpler" exception).

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
```diff
@@ -0,0 +1,261 @@
...
+          );
+
+          expect(
+            resultFromUppercase?.id,
+            createdUserProfile.id!,
+          );
+        },
+      );
+
+      test(
```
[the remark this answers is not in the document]
tp: Let's discuss this above.

I am open for a list, and then most callers should probably use `singleOrNull`, as 2 would already be a violation of the intended setup (one account per mail), in which case they certainly do not want to add a 3rd one :D

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profile_config.dart
```diff
@@ -0,0 +1,102 @@
...
+  /// Called when a user profile is about to be created.
+  final BeforeUserProfileCreatedHandler? onBeforeUserProfileCreated;
+
+  /// Called after a user profile has been created.
+  final AfterUserProfileCreatedHandler? onAfterUserProfileCreated;
+
+  /// Called whenever a user profile has been updated.
+  ///
+  /// This can be when the user's (full) name or profile picture has been changed.
+  final AfterUserProfileUpdatedHandler? onAfterUserProfileUpdated;
```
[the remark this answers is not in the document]
tp: Added the hook for `onBeforeUserProfileUpdated`.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/models/user_profile_model.spy.yaml
```diff
@@ -0,0 +1,18 @@
...
+  ### The full name of the user.
+  fullName: String?
+
+  ### The email of the user.
+  ###
+  ### Stored in lower-case.
+  email: String?
+
+  ### The user's profile image
+  imageUrl: String?
```
tp, line 18: `Uri`, now that we support that?

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/models/user_profile_image.spy.yaml
```diff
@@ -0,0 +1,18 @@
...
+  id: UuidValue?, defaultPersist=random
+
+  ### The [AuthUser] this profile image belongs to
+  authUser: module:auth_user:AuthUser?, relation(onDelete=Cascade)
+
+  ### Version of the image. Increased by one for every uploaded image.
+  version: int
+
+  ### The URL to the image.
+  url: String
```
tp, line 14:
```suggestion
  url: Uri
```

?

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/models/user_profile_image.spy.yaml
```diff
@@ -0,0 +1,18 @@
+class: UserProfileImage
+serverOnly: true
+table: serverpod_auth_profile_user_profile_image
+fields:
+  id: UuidValue?, defaultPersist=random
+
+  ### The [AuthUser] this profile image belongs to
+  authUser: module:auth_user:AuthUser?, relation(onDelete=Cascade)
```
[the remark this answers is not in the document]
tp: added tests for onDelete / cascade (and yes, it works as expected; deleting the profile removes the profile/image, but not the core auth user. Deleting the auth user removes everything).

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/endpoints/user_profile_endpoint.dart
tp: > Might as well provide all that is needed so they don't have to do db calls directly.

Changed so that the `UserProfiles` always returns the `UserProfileModel`, with the `image` filled.

The `UserProfile` (database model) is not exported anymore, so neither external applications nor this endpoint uses that directly.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
tp: The user updating has indeed be streamlined, adding consistency checks (image / image ID, to ensure that we always return a useable entity for model-building), and invalidating the cache.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/models/user_profile_model.spy.yaml
tp: Changed to `Uri`

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/models/user_profile_image.spy.yaml
tp: Changed to `Uri`

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
```diff
@@ -0,0 +1,261 @@
...
+          );
+          expect(
+            foundUserProfile?.authUserId,
+            authUserId,
+          );
+        },
+      );
+
+      test(
+        'when trying to create a second user profile for the same auth user, then this fails.',
```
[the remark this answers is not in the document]
tp: I will move this test down to the second `withServerpod`, which has a auth user + profile as a pre-requisite.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/models/user_profile.spy.yaml
tp: Decision:
- Remove the email look up here, as we have to support cross-authentication provider consolidation anyway (least for all the projects not using the `profile` package)

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/endpoints/user_profile_endpoint.dart
tp: Decision:
- Remove this from the endpoint, developers can just use `onBefore*` hooks
- Also support setting `null` for user name / full name, matching the underlying field

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/models/user_profile.spy.yaml
```diff
@@ -0,0 +1,36 @@
...
+serverOnly: true
+table: serverpod_auth_profile_user_profile
+fields:
+  id: UuidValue?, defaultPersist=random
+
+  ### The [AuthUser] this profile belongs to
+  authUser: module:auth_user:AuthUser?, relation(onDelete=Cascade)
+
+  ### The first name of the user or the user's nickname.
+  userName: String?
```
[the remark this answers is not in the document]
tp: Decision:
- keep nullable strings here

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/models/user_profile.spy.yaml
tp: Removed the email based lookup

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/endpoints/user_profile_endpoint.dart
tp: Remove checks in the endpoint, allow `null` to be send for user name / full name.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profile_images.dart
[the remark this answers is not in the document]
SandPod: Leave for now and solve if we get metrics this is an issue.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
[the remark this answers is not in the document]
SandPod: Let's leave this as is and see if we deem it necessary in a later step.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,164 @@
...
+    );
+
+    return userProfile;
+  }
+
+  /// Finds a user profile by its email address.
+  ///
+  /// Returns `null` if no user profile with that email is found,
+  /// or there already exist 2 or more with that email address.
+  static Future<UserProfile?> maybeFindUserProfileByEmail(
```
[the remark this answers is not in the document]
SandPod: Sound good to deffer this until we know how the consolidation logic should look.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
```diff
@@ -0,0 +1,261 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_profile_server/serverpod_auth_profile_server.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+import 'package:test/test.dart';
+
+import '../../test/integration/test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod(
```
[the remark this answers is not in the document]
tp: The test groups have been reworked to set up the specific set of dependencies/preconditions that is the same across all tests, and the descriptions reflect that now.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
```diff
@@ -0,0 +1,451 @@
...
+        );
+
+        expect(
+          updatedProfileFromCallback!.fullName,
+          'Updated full name',
+        );
+      },
+    );
+
+    test(
```
SandPod, line 275: Suggestion: As far as I can tell this test is a duplication of the next test.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,387 @@
...
+      storageId: 'public',
+      path: path,
+    ))!;
+
+    return setUserImageFromOwnedUrl(session, authUserId, version, publicUrl);
+  }
+
+  /// Sets the profile image to the given URL, which is presumed to be owned by this application.
+  @visibleForTesting
+  static Future<UserProfileModel> setUserImageFromOwnedUrl(
```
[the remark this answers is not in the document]
SandPod: If we are to keep this, can we make sure to not export this method at all from the package? Maybe by having it as an extension that is not shown?

util/recreate_all_migrations
SandPod: It is never used, it is just something that we created to manually reset all migrations. I'm not sure it should even be used given that we need to preserve migrations in the modules.

SandPod: Issue to track updating the `create-migration` exit code: https://github.com/serverpod/serverpod/issues/3530

SandPod (commented):
Getting close! More nitty comments now.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,387 @@
...
+  ) async {
+    userProfile = userProfile.copyWith(
+      email: userProfile.email?.toLowerCase().trim(),
+    );
+
+    userProfile = (await UserProfileConfig.current.onBeforeUserProfileCreated
+            ?.call(session, userProfile)) ??
+        userProfile;
+
+    if (userProfile.imageUrl != null) {
```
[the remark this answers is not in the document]
tp: We can enable that if desired. We'd just have to retain the URL, create the profile, and then try to set the image.

In case the image setting fails, we'd have to (try) to undo the profile creation though, right?

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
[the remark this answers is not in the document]
tp: The thing is, that the higher-level implementation wants to have a storage, which we don't have in test.

Is there a good way to mock that, such that we could get some fake URL for an (empty) blob?

Then we could test the higher-level implementation only.
Otherwise we could of course move this into a free function, and hide that explicitly in the package export.
(In practice, probably does not make a big difference, because people who don't lint against using `@visibleForTesting`/`@internal` in their projects, might not have the lint against "no src/ imports" either.)

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
```diff
@@ -0,0 +1,451 @@
...
+
+          expect(
+            createdUserProfile.toJsonForProtocol(),
+            equals(createdProfileFromCallback?.toJsonForProtocol()),
+          );
+        },
+      );
+
+      test(
+        'when creating a new user profile, then `onBeforeUserProfileUpdated` is not invoked.',
```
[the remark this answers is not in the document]
tp: In the previous test that also still created the profile it was much just one single `expect` before the update.

But not it does indeed feel a bit weird. While update/create handlers seem the easiest to confuse when changing the implementation going forward, it is of course not tenable to ensure that "all other" hooks are not invoked.

So we can also remove this now.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
tp: I don't think so. Unless I got the line numbers wrong, one test is for the user name, the other for the full name.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/endpoints/user_profile_endpoint.dart
```diff
@@ -0,0 +1,84 @@
...
+      authUserId,
+    );
+
+    return profile;
+  }
+
+  /// Removes the users uploaded image, replacing it with the default user
+  /// image.
+  Future<UserProfileModel> removeUserImage(final Session session) async {
+    if (!UserProfileConfig.current.userCanEditUserImage) {
```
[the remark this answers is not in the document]
tp: @SandPod We can convert these to functions.

For now I just considered them a quick-fix so that people could use the endpoint, but could easily toggle them on/off.

This was mostly inspired because we had that capability in the old system.

Now that I thought about this, I think we could just remove it outright, and if the developer wants to hide them or add extra access controls, they can just override them in their subclass. This would match the pattern for `abstract` endpoint which is now documented.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
```diff
@@ -0,0 +1,451 @@
...
+            invoked = true;
+            return userProfile;
+          });
+
+          await UserProfiles.createUserProfile(
+            session,
+            UserProfileModel(authUserId: authUserId),
+          );
+
+          // Insert does not count as update
```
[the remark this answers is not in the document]
tp: These 2 test have been remvoed outright.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/endpoints/user_profile_endpoint.dart
```diff
@@ -0,0 +1,84 @@
...
+import 'package:serverpod_auth_profile_server/serverpod_auth_profile_server.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+
+/// User profile management endpoint.
+abstract class UserProfileEndpoint extends Endpoint {
+  @override
+  bool get requireLogin => true;
+
+  /// Returns the user profile of the current user.
+  Future<UserProfileModel> get(final Session session) async {
```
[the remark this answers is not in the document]
tp: Yes, would like to test these as well without having to write a second package with an implementation.

Let's tackle this with / after https://github.com/serverpod/serverpod/issues/3541

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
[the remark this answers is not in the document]
tp: as storage works just fine in the tests (TIL), I could remove this and just test the whole stack

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
```diff
@@ -0,0 +1,451 @@
...
+import 'package:serverpod_auth_profile_server/serverpod_auth_profile_server.dart';
+import 'package:serverpod_auth_profile_server/src/generated/protocol.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+import 'package:test/test.dart';
+
+import '../../test/integration/test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod(
+    'Given the `UserProfiles` implementation and an existing `AuthUser`,',
```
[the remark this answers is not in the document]
tp: cleaned up the test names

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,450 @@
...
+    final Session session,
+    UserProfileModel userProfile,
+  ) async {
+    final modifiedProfile = await UserProfileConfig
+        .current.onBeforeUserProfileCreated
+        ?.call(session, userProfile);
+
+    if (modifiedProfile != null) {
+      if (modifiedProfile.authUserId != userProfile.authUserId) {
+        throw Exception(
```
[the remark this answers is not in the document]
SandPod: Make sure to convert all exceptions in the class to be domain specific. If we don't want the users to catch them (crash), these should be errors instead.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,450 @@
...
+  }
+
+  static Future<UserProfile> _findUserProfile(
+    final Session session,
+    final UuidValue authUserId,
+  ) async {
+    final userProfile = await _maybeFindUserProfile(session, authUserId);
+
+    if (userProfile == null) {
+      throw Exception('Did not find profile for user "$authUserId"');
```
SandPod, line 445: Question: Shouldn't this be the UserProfileNotFoundException?

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,450 @@
...
+
+    var createdProfileModel = createdProfile.toModel();
+
+    if (imageUrl != null) {
+      createdProfileModel = await UserProfiles.setUserImageFromUrl(
+        session,
+        userProfile.authUserId,
+        imageUrl,
+      );
+    }
```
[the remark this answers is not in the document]
SandPod: Please check this for all methods in the class. Maybe transactions should be used in some places to ensure that the database is not left in an unexpected state in regards to the called method.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
```diff
@@ -0,0 +1,482 @@
...
+      );
+
+      test(
+        'when creating a user profile with an image URL, then a local copy of the image is exposed on the user.',
+        () async {
+          final createdProfile = await UserProfiles.createUserProfile(
+            session,
+            UserProfileModel(
+              authUserId: authUserId,
+              imageUrl: Uri.parse(
```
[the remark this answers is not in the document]
SandPod: Is the question if it is possible to mock the request? 

Could you use mocktail for this? Check out the suggestion at the end of this overflow thread: https://stackoverflow.com/questions/55409978/how-to-mock-http-request-in-flutter-integration-test

templates/pubspecs/modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/pubspec.yaml
```diff
@@ -0,0 +1,36 @@
...
+version: SERVERPOD_VERSION
+repository: https://github.com/serverpod/serverpod
+
+environment:
+  sdk: DART_VERSION
+
+dependencies:
+  http: '>=1.0.0 <2.0.0'
+  image: ^4.0.15
+  meta: ^1.11.0
```
SandPod, line 14: Question: This dependency can be removed now right?

SandPod (commented):
Still found a couple of things.

Please ensure that we never end up in a broken state if we suddenly drop the db connection and ensure that all exceptions thrown from the domain are typed so that we can differentiate between these and other errors that might occur in used libraries.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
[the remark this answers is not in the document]
tp: I was just wondering whether there was anything build into the test tools already (which I did not see).

Since overwriting the core `http` usage is such a mess, I guess we'll just have to make it configurable / injectable, so we just mock / replace 1 function, instead of the whole HTTP chain (as this person did: https://zetcode.com/dart/htttoverrides/).

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,450 @@
...
+import 'package:serverpod_auth_profile_server/serverpod_auth_profile_server.dart';
+import 'package:serverpod_auth_profile_server/src/generated/protocol.dart';
+import 'package:serverpod_auth_profile_server/src/util/user_profile_extension.dart';
+
+/// Business logic for handling user profiles
+abstract final class UserProfiles {
+  /// Creates a new user and stores it in the database.
+  ///
+  /// In case the `onBeforeUserProfileCreated` hook redirects this creation onto an existing profile,
+  /// no profile will be created, but the existing profile will be updated with the return value of the hook.
```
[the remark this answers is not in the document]
tp: Right, good catch.

This was possible before, but is not the case anymore.

Now the method will always create a new profile, and if you want to use an existing one (and update that), you can just find the profile first and use that instead.

So again a move away from complex hooks, and clean/linear code on the call site.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
tp: yes, changed to use the module's own exception here as well.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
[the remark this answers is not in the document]
tp: I simplified the hooks and removed the image setting outright (it's not passed in nor out).

see cd4437846b64905303e4eae57b2f2c358fe28c7d

For image updates we would still need another hook to support, but what kind of code do we even envision that could do something useful "in between" there on an image change?

As written in Slack I think in practice this just leads to an unnecessarily complicated implementation, and developer's could just change the data before calling the `UserProfile`s methods.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,404 @@
...
+        authUserId: authUserId,
+        userName: userProfile.userName,
+        fullName: userProfile.fullName,
+        email: userProfile.email,
+      ),
+    );
+
+    final createdProfileModel = createdProfile.toModel();
+
+    await UserProfileConfig.current.onAfterUserProfileCreated?.call(
```
tp, line 45: @SandPod What do you think we should do here:

- State that the `onAfterUserProfileCreated` is for information only and any exceptions thrown ignored (though still await-ed, or is even that too much in that case)?
- Use a transaction for the above insert, and roll back in case the `onAfterUserProfileCreated` fails?

Like in the "before handler" discussion, we might just be better off removing it and letting the application code run any post-setup code itself.
That should be much clearer on the call side, and we don't have to impose either of the 2 rules (which are basically opposites, so one kind of implementation will not be helped by them, so you can not bend them to work in either case).

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,404 @@
...
+    if (UserProfileConfig.current.userImageFormat == UserProfileImageType.jpg) {
+      pathExtension = '.jpg';
+    } else {
+      pathExtension = '.png';
+    }
+
+    // Store the image.
+    final path = 'serverpod/user_images/$authUserId-$version$pathExtension';
+    await session.storage.storeFile(
+      storageId: 'public',
```
tp, line 221: TODO:
Store `storageId` and `path` in the database as well, then we could delete the images with the profile (at least when deleted directly and not via the auth user only).

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,404 @@
...
+  }
+
+  /// Remove the user profile for the given [authUserId].
+  ///
+  /// In case the user did not have a profile, nothing is changed.
+  static Future<void> deleteProfileForUser(
+    final Session session,
+    final UuidValue authUserId,
+  ) async {
+    await UserProfile.db.deleteWhere(
```
tp, line 157: TODO:
Delete images from storage

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,404 @@
...
+    // Find the latest version of the user image if any.
+    final oldImageRef = await UserProfileImage.db.findFirstRow(
+      session,
+      where: (final t) => t.userProfileId.equals(userProfile.id!),
+      orderBy: (final t) => t.version,
+      orderDescending: true,
+    );
+
+    // Add one to the version number or create a new version 1.
+    final version = (oldImageRef?.version ?? 0) + 1;
```
tp, line 209: Do we think the version is actually relevant here, or couldn’t we just give the image a unique name (random part or timestamp).

It’s then referenced via the user profile anyway (and not looked up by most recent version).

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
tp: Updated the implementation to also (try to) delete images from storage.

If the deletion from storage fails, only an error is logged, as we can not recover from that, and keeping the `UserProfileImage` with a pointer to a file that is broken is likely not a better course of action.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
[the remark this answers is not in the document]
tp: I tried using transactions for the profile image DB insert and profile update, but that does seem to conflict with `withServerpod` as the transaction seems to be started "on the side" instead of nested inside the current test's transaction:

>   Concurrent database calls outside an already active transaction are not supported when database rollbacks are enabled. If this is intended, disable rolling back the database by setting `rollbackDatabase` to `RollbackDatabase.disabled`.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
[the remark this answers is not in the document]
tp: changed to errors in the cases where the calling application code needs to change to fix the issue.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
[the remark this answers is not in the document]
tp: Made the image fetch function not configurable, so we can override it in the tests.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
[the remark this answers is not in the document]
tp: So had a brainstorm with @christerswahn about the transactions here, and he posited the question whether we actually want to have transactions passed across modules.

I feel it be valid to allow the developer to say they want to run the entire "user creation" (auth user, email account registration completion, profile creation) in a single transaction and rollback everything if that fails.

If you agree, then the question would be whether we require a transaction on the set image call variants or make that optional. In the latter case we have issue that we currently can not have nested transactions, and thus would need some kind like `transaction ?? session.db.createTransaction()`.

I think the "use parent transaction if available" could easily be achieved by having a helper / optional parameter on `createTransaction` which would just reuse the parent instead of creating a new one (actually running in the parent context is also important to get a consistent view of the transaction, so we don't have to change isolation levels and "hope" that everything can be committed in the end (with regards to our foreign key relations)).

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
tp: Decision:
- We do not need the numeric ID, random identifier / UUID is fine.

tp: Switched to a short random part for every file's name (also still scoped to the user id).

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
tp: As dicussed, keeping the hooks for now, but giving it the transaction in case they would need to look at the "in progress" database state.

Since we don't have dependency injection yet and could not exchange a `UserProfiles` implementation for an adapted one, it would not be realistic to change all callers into this module (which might be many layers above), so there is currently a good case to be made for these hooks, even if they result in a less "predictable" module implementation).

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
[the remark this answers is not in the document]
tp: As decided on the call, this now either uses the given transaction (and creates a nested savepoint, to be restore in case the module's method fails) or creates it's own one for multi-step writes, such that the module never leaves the database in a partially updated state.

This is done in `transactionOrSavepoint`.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,589 @@
...
+  /// Either creates a new transaction for [f],
+  /// or creates a savepoint inside the given transaction,
+  /// and would discard any modifications if [f] fails.
+  Future<R> transactionOrSavepoint<R>(
+    final TransactionFunction<R> f, {
+    required Transaction? transaction,
+  }) async {
+    // Use the implicit transaction from tests
+    // ignore: invalid_use_of_visible_for_testing_member
+    transaction ??= this.transaction;
```
tp, line 560: I suppose this is fine here, as the low-level DB helpers would just do this as well.

So for the test code there is never any need to use the `transaction: ` parameter explicitly.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,589 @@
...
+            image,
+            quality: UserProfileConfig.current.userImageQuality,
+          ),
+        UserProfileImageType.png => encodePng(image),
+      },
+    );
+  }
+}
+
+extension on Session {
```
tp, line 548: Unlike the `transaction` method, this lives on `Session` (not `Database`) in order to discover the test transaction.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,589 @@
...
+    );
+  }
+
+  /// Find a user profile by the `AuthUser`'s ID.
+  ///
+  /// By default the result is cached locally on the server. You can configure the cache
+  /// lifetime in [UserProfileConfig], or disable it on a call to call basis by
+  /// setting [useCache] to false.
+  ///
+  /// Throws a [UserProfileNotFoundException] in case no profile exists for the given [authUserId].
```
tp, line 67:
```suggestion
  /// Throws a [UserProfileNotFoundException] in case no profile exists for the given [authUserId].
  ///
  /// In case a [transaction] is passed, the cache will not be used.
```

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,589 @@
...
+import 'package:image/image.dart';
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_profile_server/serverpod_auth_profile_server.dart';
+import 'package:serverpod_auth_profile_server/src/generated/protocol.dart';
+import 'package:serverpod_auth_profile_server/src/util/user_profile_extension.dart';
+import 'package:serverpod_shared/serverpod_shared.dart';
+
+/// Business logic for handling user profiles
+abstract final class UserProfiles {
+  /// Creates a new user and stores it in the database.
```
tp, line 13:
```suggestion
  /// Creates a new user profile and stores it in the database.
```

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,589 @@
...
+
+      return updatedProfile.toModel();
+    }, transaction: transaction);
+  }
+
+  /// Remove the user profile, incl. images, for the given [authUserId].
+  ///
+  /// In case the user did not have a profile, nothing is changed.
+  ///
+  /// Returns successfully if the profile data could was removed from the database,
```
tp, line 191:
```suggestion
  /// Returns successfully if the profile data was removed from the database,
```

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,589 @@
...
+  /// even if some profile images could not be deleted.
```
tp, line 192:
```suggestion
  /// even if some profile images could not be deleted from the file storage.
```

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,589 @@
...
+        return;
+      }
+
+      final images = await UserProfileImage.db.find(
+        session,
+        where: (final t) => t.userProfileId.equals(profile.id!),
+        transaction: transaction,
+      );
+
+      // This also deletes the user profile images
```
tp, line 218:
```suggestion
      // This also deletes the user profile image entries from the database
```

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
tp: cc @nielsenko

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
tp: This has been removed as requested.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
tp: This continues to live on `Session`, so that we can log to it in case the rollback fails.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,592 @@
...
+      );
+
+      // This also deletes the user profile image entries from the database
+      await UserProfile.db.deleteWhere(
+        session,
+        where: (final t) => t.authUserId.equals(authUserId),
+        transaction: transaction,
+      );
+
+      await _invalidateCacheForUser(session, authUserId);
```
SandPod, line 228: Comment: This cache invalidation not have any effect until the transaction is commited.

So any concurrent requests to the server would populate the cache again.

Same for the other callsite to this method.

What is our cache TTL?

Did we create an issue for an onCommit listener?

SandPod: Suggestion: Expose a method to invalide the cache for a user.

This can be used if the transaction is long running. Also add a description that this might be needed to be called when transactions are used.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
```diff
@@ -0,0 +1,673 @@
...
+
+          expect(
+            createdUserProfile.toJsonForProtocol(),
+            equals(createdProfileFromCallback?.toJsonForProtocol()),
+          );
+        },
+      );
+
+      test(
+        'when `onAfterUserProfileCreated` throws during the creation of a new user profile, then the profile is not stored in the database and the error forwarded.',
```
SandPod, line 104: Suggestion: Missing the equivalent test for `onBeforeUserProfileCreated`

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
```diff
@@ -0,0 +1,673 @@
...
+        );
+        expect(
+          updatedUserProfile.userName,
+          'username from onBeforeUserProfileUpdated hook',
+        );
+      },
+    );
+
+    test(
+      'when `onAfterUserProfileUpdated` throws during the update of a user profile, then the update is not visible in the database.',
```
SandPod, line 286: Suggestion: Missing the equivalent test for `onBeforeUserProfileUpdated`

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
```diff
@@ -0,0 +1,673 @@
...
+
+      await UserProfiles.createUserProfile(
+        session,
+        authUserId,
+        UserProfileData(),
+      );
+    });
+
+    test(
+        'when the image is updated, then the new URL is available on the profile on read.',
```
SandPod, line 185: Suggestion: Remove, this seems like a duplicate of the same test further down in the the group.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
```diff
@@ -0,0 +1,673 @@
...
+          session,
+          where: (final f) => Constant.bool(true),
+        );
+        await UserProfileImage.db.deleteWhere(
+          session,
+          where: (final f) => Constant.bool(true),
+        );
+      }
+
+      setUpAll(clearDb);
```
SandPod, line 616: Suggestion: Remove, as this will just hide us leaking state bewteen tests.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,592 @@
...
+      );
+    }
+
+    userProfile = await UserProfile.db.updateRow(
+      session,
+      userProfile,
+      transaction: transaction,
+    );
+
+    await _invalidateCacheForUser(session, userProfile.authUserId);
```
SandPod, line 478: Suggestion/Discussion: Test this behavior.

We have no tests validating the caching behavior, should we?
How are the tests influenced by the caching? Is the cache cleared in the test framework?

SandPod (approved):
Overall, this looks good now! There are a couple of suggestions that I would like to be fixed before this is merged.

But I do consider this working.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
tp: > This cache invalidation not have any effect until the transaction is commited.

That's correct.

> So any concurrent requests to the server would populate the cache again.

Yes, in this server and any others.

> What is our cache TTL?

1 Minute by default

> Did we create an issue for an onCommit listener?

Just now, it's here: https://github.com/serverpod/serverpod/issues/3557


As discussed, the cache inside a server (and of course across servers) currently always has the potential issue of giving you outdated data for up to a minute (default TTL).

I think with the "post commit hook" this can be brought down to basically 0 (for the same server at least).

Inside a single endpoint method there is hopefully always a consistent view, as the profile should only be fetched once per request. Every modifying operation already reads the latest data from the database (with the current transaction), so updates never operate on stale data.

Overall I would not be worried about any concurrency issues here so much, as they can only happen when modifying the the profile _and_ having multiple concurrent reads at the same time. I don't that as a major use-case in an app. (And that a cache is used be default is pretty clearly stated everywhere.)

Still, to be honest the cache is only here because it was in the old auth package. Maybe we should just drop it, as practically I don't see this being used on "every request", so do we even have multiple hits during the 1 minute TTL? It might just not be worth the complexity.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
tp: > Suggestion: Remove, as this will just hide us leaking state bewteen tests.

That is true. Though practically this was an issue I ran into when writing the test, as a previous version was not cleaning up, and then there's no good way to recover (to get the next test pass).

I think without this I would have had to manually clean up the DB / delete the Docker volume to get it working.

Though it we can at least keep `clearDB` (or better named `clearProfiles` maybe) as a function here, then it will be available for "manual clean up", but we don't have to commit the `setUpAll` call.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
tp: > We have no tests validating the caching behavior, should we?

We do have a test for that via the subsequent / multiple updates of the profile images, where we then always check that the caching `findProfile*` method returns the latest value (which equals the result of the update operation).

> How are the tests influenced by the caching? Is the cache cleared in the test framework?

Every test creates their own `AuthUser`, so while the cache instance on the singleton can persist, we would always operate on other values.

We add a `@visibleForTesting` helper though to clear the test between tests.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
tp: You're right, this is similar to `"when updating the profile's image from bytes, then the updated value is visible through the cached `findUserProfileByUserId` method."`.

Just here we updated the image twice and checked twice.

But there is really no reason why we would want to do this here and not in the other cases.

I think as long as we trust that `findProfile*` fills the cache, we're good in assuming that the second update will clear the cache just like the first one.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
SandPod: True, we could just provide this in the endpoint layer instead in that case?

tp: > True, we could just provide this in the endpoint layer instead in that case?

Exactly, if we even need it.

And then, in combination with the new hook, we could also move this to a shared cache that we could then hold in sync properly.
(If this is really needed it would also be better to then just write the latest data to the cache instead of just invalidating it and thus requiring the next request (or even requests, if they arrive in parallel?) to fetch it again.)

tp: Removed the caching. Still kept the reduced tests that ensure that `update*(ID, …) == findProfile(ID)`.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
tp: Removed this test in favor of the other one.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
tp: Update the helper to only run in `tearDown` and also only remove the auth user (plus associated profile etc.) which were created in that test.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
tp: Obsolete with the above decision to remove the cache.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
tp: Interestingly those throwing test cases expose another issue, where subsequent tests then fail as the exception leaves the (test) system in a broken state.

I added the tests in 29dd273e8fe129e8e25fdb5888e3ac1669c16013 as `skip: true`, and will now create a follow-up PR trying to debug and fix the test tool to handle this (as I see no reason why the tests should not work as written).

This is probably related to the issue we saw yesterday where just a single test needed `transaction: session.transaction` passed explicitly.

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/test/integration/user_profiles_test.dart
tp: added, but same issue as in https://github.com/serverpod/serverpod/pull/3441#discussion_r2097233800

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
```diff
@@ -0,0 +1,589 @@
...
+  }
+}
+
+extension on Session {
+  /// Runs the closure [f] in isolation, inside the parent [Transaction] (if any).
+  ///
+  /// Either creates a new transaction for [f],
+  /// or creates a savepoint inside the given transaction,
+  /// and would discard any modifications if [f] fails.
+  Future<R> transactionOrSavepoint<R>(
```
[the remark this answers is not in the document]
tp: As discussed, will move to `serverpod` proper in a new PR with tests.

(TBD, probably with E2E as well, as we've already seen that this behaves differently/unexpectedly in integration tests, and we should first and foremost ensure the production behavior.)

PR #3443, opened 2025-04-22
feat: Migrate away from `int` `userId` in core `AuthenticationInfo`

tp:
**Changes**

Deprecates the `int` `userId` field on `AuthenticationInfo`.

Adds an agnostic `String` `user` field instead.

The current `serverpod_auth` will keep using its `int` ID, stored as `String` and converted on demand to `int`.

Authentication packages (currently `serverpod_auth`, in the future `serverpod_auth_user`) will provide an extension on `Session` to get the typed user ID for their user model (`int` and `UuidValue` respectively).

**Breaking changes**

This does not require updating any application code, but developers might get a warning (depending on their linter configuration). They should then switch to `await session.userId` (from `(await session.authenticationInfo)?.userId`).

**Further considerations**

In the future, we could remove `userId` from `AuthenticationInfo` and use that same name for an extension method in the authentication packages, such that the user ID is also available in a typed fashion on the `AuthenticationInfo` class.

Session revocation still works the same way, as `serverpod_auth` was already using `int.parse` on the revocation method to look up the user ID (or authentication ID) in its realm.

Closes https://github.com/serverpod/serverpod/issues/3444


packages/serverpod/lib/src/authentication/authentication_info.dart
```diff
@@ -11,8 +11,16 @@ typedef AuthenticationHandler = Future<AuthenticationInfo?> Function(
 /// Holds the id for an authenticated user and which [scopes] it can access.
 /// Allowed scopes are defined for each [Endpoint].
 class AuthenticationInfo {
-  /// Id for an authenticated user.
-  final int userId;
+  /// Integer Id for an authenticated user.
+  ///
+  /// Will throw in case the user ID of the authenticated user is not an integer.
+  @Deprecated(
+    'Use the authentication packages typed getter on `Session` instead (from `serverpod_auth` or `serverpod_auth_user`)',
```
tp, line 18: Clarify that `session.userId` should now be available to them (in auto-complete, might require `serverpod_auth` import for now (or `serverpod_auth_user` in the future)).

packages/serverpod/lib/src/authentication/authentication_info.dart
```diff
@@ -11,8 +11,16 @@ typedef AuthenticationHandler = Future<AuthenticationInfo?> Function(
...
+  )
+  int get userId => int.parse(user, radix: 10);
+
+  /// Identifier of the user, as set by the session handler
```
tp, line 22:
```suggestion
  /// Identifier of the user, as set by the session handler.
```

packages/serverpod/lib/src/server/message_central.dart
```diff
@@ -7,10 +7,10 @@ abstract class MessageCentralServerpodChannels {
   /// Used to revoke authentication tokens.
   /// The message should be of type [RevokedAuthenticationUser],
   /// [RevokedAuthenticationAuthId] or [RevokedAuthenticationScope].
-  /// The [userId] should be the [AuthenticationInfo.userId] for the concerned
+  /// The user ID should be the [AuthenticationInfo.user] for the concerned
   /// user.
-  static String revokedAuthentication(int userId) =>
-      '_serverpod_revoked_authentication_$userId';
+  static String revokedAuthentication(Object user) =>
```
tp, line 12: I wonder if we should break this to `String user`, as this probably has very few callers outside of our packages.

packages/serverpod/lib/src/server/log_manager/log_manager.dart
```diff
@@ -329,7 +330,10 @@ class SessionLogManager {
         slow: isSlow,
         error: exception,
         stackTrace: stackTrace?.toString(),
-        authenticatedUserId: authenticatedUserId,
+        authenticatedUserId: authenticatedUserId == null
+            ? null
+            : int.tryParse(authenticatedUserId),
+        // TODO: What to do here? We can not break the table… Can we deprecate the old user ID? Manual migration?
```
tp, line 336: 👆 Open question.

As this is persisted to the DB, we need to find a good migration path (just adding a new field might also leave people "broken" if they query (e.g. group) on the old one).

Isakdl: I think this is just stored in the db and shouldn't be used in Serverpod, might be used in insights but I'm not sure @vlidholt has better knowledge there. But yes, we likely need to find a migration path for this.

This might be one of the hardest and annoying migrations to resolve though. We need a good workflow for, e.g. renaming / breaking database migrations and I don't think we have ever done that in the framework before.

packages/serverpod/lib/src/authentication/authentication_info.dart
```diff
@@ -21,5 +29,10 @@ class AuthenticationInfo {
   final String? authId;
 
   /// Creates a new [AuthenticationInfo].
-  AuthenticationInfo(this.userId, this.scopes, {this.authId});
+  AuthenticationInfo(
+    @Deprecated('User parameter `user` instead') int? userId,
+    this.scopes, {
+    this.authId,
+    String? user,
+  }) : user = user ?? userId!.toString();
```
tp, line 37: Should we guard against both `userId` and `user` being passed?

This just checks that at least one is given.

tests/serverpod_test_server/test_integration/test_tools/session_copy_with_test.dart
```diff
@@ -19,15 +21,23 @@ void main() {
             () async {
           await expectLater(
             modifiedSessionBuilder.build().authenticated,
-            completion((a) => a.userId == 123),
```
tp, line 22: Since the `completion` callback was not typed here (`dynamic`), it did not show that the import for the extension was missing. And without a type, the linter tool would remove the extension import, as it did not know it was used here. But I think in practice an `AuthenticationInfo` passed as `dynamic`  will not be thing to worry about.

modules/serverpod_auth/serverpod_auth_server/lib/src/util/auth_info_extension.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:serverpod/serverpod.dart';
+
+/// User ID extension for `AuthenticationInfo`
+extension AuthenticationInfoUserId on AuthenticationInfo {
+  /// Returns the `int` user ID of the authenticated user.
+  ///
+  /// Assumes that the system uses `int` user IDs, otherwise throws.
+  int get userId {
+    return int.parse(userIdentifier, radix: 10);
+  }
```
SandPod, line 10: Suggestion: Moving this here is a breaking change.

For files where the import is missing this will be a breaking change.

This will need to be part of AuthenticationInfo until 3.0

packages/serverpod/lib/src/server/message_central.dart
```diff
@@ -7,9 +7,9 @@ abstract class MessageCentralServerpodChannels {
   /// Used to revoke authentication tokens.
   /// The message should be of type [RevokedAuthenticationUser],
   /// [RevokedAuthenticationAuthId] or [RevokedAuthenticationScope].
-  /// The [userId] should be the [AuthenticationInfo.userId] for the concerned
+  /// The [userId] should be the [AuthenticationInfo.userIdentifier] for the concerned
   /// user.
-  static String revokedAuthentication(int userId) =>
+  static String revokedAuthentication(Object userId) =>
```
SandPod, line 12: Suggestion: Rename parameter to `userIdentifier` to match.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -640,7 +641,8 @@ class MessageCentralAccess {
   /// [RevokedAuthenticationScope] is used to communicate that a specific
   /// scope or scopes have been revoked for the user.
   Future<bool> authenticationRevoked(
-    int userId,
+    // Uses `Object` to avoid breaking change, but should switch to `String` (mirroring `AuthenticationInfo.userIdentifier`) in the future
+    Object userId,
```
SandPod, line 645: Suggestion: Rename to `userIdentifier` to match expected.

packages/serverpod/lib/src/authentication/authentication_info.dart
```diff
@@ -11,8 +11,8 @@ typedef AuthenticationHandler = Future<AuthenticationInfo?> Function(
 /// Holds the id for an authenticated user and which [scopes] it can access.
 /// Allowed scopes are defined for each [Endpoint].
 class AuthenticationInfo {
-  /// Id for an authenticated user.
-  final int userId;
+  /// Identifier of the user, as set by the session handler.
+  final String userIdentifier;
```
SandPod, line 15: Suggestion: This class still needs a `userId` getter.

Removing it would be a breaking change.

modules/serverpod_auth/serverpod_auth_server/lib/src/util/auth_info_extension.dart
tp: I thought we agreed about this, that yes, this requires the developer to add the import on `serverpod_auth` (which they must have in their project, and will be suggested via auto-complete), but it does not silently change any behavior.

Meaning yes, they will have to a few simple steps, but in return we all get a clean API now :)

Otherwise we had the behavior in the commits before with the `@deprecated` `userId` and the new "universal" `user` getter provided by each auth package (`serverpod_auth` and the new one), but as you said `user` is really not the best name for this ID field, and a user of `serverpod_auth` would then have 3 similar fields: `userIdentifier` (String), `userId` (int, deprecated), and `user` (int, the new one to use). That does not seem like an improvement over the 1-time migration to me (which is technically SemVer breaking, but not really an issue for any developer (I suppose)).

packages/serverpod/lib/src/server/message_central.dart
tp: Renamed as suggested.

packages/serverpod/lib/src/authentication/authentication_info.dart
tp: As said above, I think keeping but deprecating it here will put the same effort on consumers (add the auto-complete import, or add the ignore lint, or switch to the previous `user` getter approach) in the end.

And I would really not want to keep it un-deprecated, as that will just be confusing for "Auth 2.0" users, always seeing that field they must not be using 😅

modules/serverpod_auth/serverpod_auth_server/lib/src/util/auth_info_extension.dart
SandPod: That must have been a missunderstanding. We don't introduce breaking changes in minor versions.

I agreed that breaking it in this way would be OK, that is, hijacking the meaning of `userId` without a `@deprectated` annotation warning. Since the end user could simply auto import the missing dependency, it wouldn't be an issue.

But I don't think this motivates sidestepping SemVer.

Maybe we should have another discussion around this.

tp: Sorry, my bad.

So then as discussed:

- we keep the `int userId` as is on the core `AutheticationInfo`
- The "Auth 2.0" package will define its own extension for now (e.g. `userUuid`)
  - Auth 2.0 project must avoid calling `userId`, as it'll crash at runtime
- With the 3.0 breaking change we can move the `userId` into an extension in `serverpod_auth` (and then rename the `userUuid` to `userId` in 2.0 as well)

This is tracked in https://github.com/serverpod/serverpod/issues/3476

tests/serverpod_test_server/test_integration/test_tools/session_copy_with_test.dart
tp: I kept the typed `isA` here, as loosing the types in `completion` seems suboptimal. Luckily such things will be better with https://pub.dev/packages/checks in the future :)

packages/serverpod/lib/src/server/log_manager/log_manager.dart
SandPod: Tracked here: https://github.com/serverpod/serverpod/issues/3448

packages/serverpod/lib/src/authentication/authentication_info.dart
```diff
@@ -11,8 +11,15 @@ typedef AuthenticationHandler = Future<AuthenticationInfo?> Function(
...
-  /// Id for an authenticated user.
-  final int userId;
+  /// Returns the `int` user ID of the authenticated user.
+  ///
+  /// Assumes that the system uses `int` user IDs, otherwise throws.
+  int get userId {
+    return int.parse(userIdentifier, radix: 10);
+  }
+
+  /// Identifier of the user, as set by the session handler.
```
SandPod, line 21: Suggestion: It is not the session handler but the `AuthenticationHandler` that determines this.

SandPod (approved):
Changes look good!

One minor comment to fix around a comment.

packages/serverpod/lib/src/server/log_manager/log_manager.dart
tp: Inded, this issue is also linked in the latest version of the code: https://github.com/tp/serverpod/blob/4315e0f240e0fa2ecef1a8d20e230dad5a7e9d2d/packages/serverpod/lib/src/server/log_manager/log_manager.dart#L333

packages/serverpod/lib/src/authentication/authentication_info.dart
tp: You're right, and the auth handler is even defined just 5 lines above eyes 🙈

Fixed it!

PR #3459, opened 2025-04-23
docs: Add code style section for discussion

dkbast:
* Applies markdown lint to the CONTRIBUTING.md (what we are using in serverpod_docs)
* Adds a section about code style - this is what we should discuss


CONTRIBUTING.md
```diff
@@ -267,3 +267,21 @@ We encourage you to support others in their Serverpod projects by sharing your k
 ## Contributing with issues
 
 Help us make Serverpod better by filing [issue](https://github.com/serverpod/serverpod/issues/new/choose) for bugs, feature requests, or improvements. When filing an issue, please provide as much information as possible to help us understand the problem or suggestion.
+
+## Coding style
```
dkbast, line 271: Please add your thoughts and suggestions here cc @vlidholt @SandPod @Isakdl @tp @christerswahn @nielsenko @mevech

nielsenko: When opening a PR the template request that you confirm that:

> I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).

We also have the `serverpod_lint` package that enforce a number of extra lint packages/serverpod_lints/lib/internal.yaml.

Perhaps it would be enough to refer to the bullet from the PR template and the lint rules?

CONTRIBUTING.md
```diff
@@ -267,3 +267,21 @@ We encourage you to support others in their Serverpod projects by sharing your k
...
+
+### Naming conventions
+
+**Regarding naming objects:**
+
+- For classes that are used to generate code (e.g. extending `Endpoint`) we append that to the class name (e.g. `GreetingEndpoint`) which is being stripped out by the code generator - so the client can use `Greeting` instead of `GreetingEndpoint`.
+- For anything we specify in yaml files (e.g. names of models, tables etc) we don't append anything to the name.
```
tp, line 278:
```suggestion
- For anything we specify in yaml files (e.g. names of models, tables, indices etc) we don't append anything to the name.
```

Because I saw some `*_idx` and other times `table_name_field_name1_field_name2` for example.

CONTRIBUTING.md
```diff
@@ -267,3 +267,21 @@ We encourage you to support others in their Serverpod projects by sharing your k
...
+
+**Regarding file names:**
+
+- We mirror the class name in the file name. So if we have a class `GreetingEndpoint` we will name the file `greeting_endpoint.dart`.
```
tp, line 282: Is that really desirable, given that that we already group them in `/endpoints` by default? (Though we don't need to.)

Personally I might say yes, if the class is also named that way because it behaves better in autocomplete / file search.

CONTRIBUTING.md
```diff
@@ -267,3 +267,21 @@ We encourage you to support others in their Serverpod projects by sharing your k
 ## Contributing with issues
 
 Help us make Serverpod better by filing [issue](https://github.com/serverpod/serverpod/issues/new/choose) for bugs, feature requests, or improvements. When filing an issue, please provide as much information as possible to help us understand the problem or suggestion.
+
+## Coding style
+
+### Naming conventions
```
tp, line 273: I think from your comment on https://github.com/serverpod/serverpod_docs/pull/268, we might also want to add a small dictionary.

Like the open question what is an endpoint? Class or method, or both? In gRPC they use service (for the container) and method.

CONTRIBUTING.md
```diff
@@ -267,3 +267,21 @@ We encourage you to support others in their Serverpod projects by sharing your k
...
+
+**Regarding naming objects:**
+
+- For classes that are used to generate code (e.g. extending `Endpoint`) we append that to the class name (e.g. `GreetingEndpoint`) which is being stripped out by the code generator - so the client can use `Greeting` instead of `GreetingEndpoint`.
```
tp, line 277: TBD: `GreetingEndpointBase` or `GreetingBaseEndpoint` or ❓ for the new `abstract` endpoints

SandPod: I don't think `Base` should be part of the name at all since it has a specific meaning in Dart.

Better to express why we would inherit from it, thinking about `Animal`, `Mammal` and `Dog`.

This is different from `tests`, where we might want to express the position in the hierarchy in the name.

CONTRIBUTING.md
SandPod: Good point, we do use `*_idx` for indexes, is it important? Should we drop it?

CONTRIBUTING.md
SandPod: This helps with searches quite a bit. I think it makes the project structure a bit more clear.

CONTRIBUTING.md
SandPod: I think this is more of a complement to these, naming standards, and such.

CONTRIBUTING.md
mevech: Suggestion: indicate this is an example
```md
- We mirror the class name in the file name. For example, class `GreetingEndpoint` belongs in `greeting_endpoint.dart`.
```

mevech (commented):
I only have wording suggestion here.

Just somethings I find easier to read and get started to contribute. 😊

CONTRIBUTING.md
nielsenko: @tp, @SandPod and @dkbast .. somewhat unrelated, but do we prefer _indices_ or _indexes_ (as plural form of _index_)? I believe the latter is more common in American English.

PR #3473, opened 2025-04-25
fix: Fix changing the id type of a table only dropping/creating the column

marcelomendoncasoares:
After #3437 merge, one question remained to be solved: what happens when changing the type of an id in a migration? The answer is that the migration was allowing to only drop/create the column instead of the whole table. This PR fixes it by forcing to drop/create the table instead.

## Pre-launch Checklist

## Breaking changes

N/A.


SandPod (approved):
LGTM! 👍

PR #3474, opened 2025-04-25
feat: Add UUIDv7 to `UuidValue` with support for usage as id type

marcelomendoncasoares:
Following issue #3255, this PR introduces support for v7 variant of UUID, implemented with native `pgsql` on the database level.

## Pre-launch Checklist

## Breaking changes

N/A.


SandPod:
This PR needs to have main merged into it.

SandPod:
After this we only need the global config configuration option and then this feature should be OK.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions/default.dart
```diff
@@ -310,7 +310,7 @@ class DefaultValueRestriction extends ValueRestriction {
     }
 
     String invalidValueError =
-        'The "$key" value must be a "random" or valid UUID string (e.g., "$key"=random or "$key"=\'550e8400-e29b-41d4-a716-446655440000\').';
+        'The "$key" value must be "random", "random_v7" or valid UUID string (e.g., "$key"=random or "$key"=\'550e8400-e29b-41d4-a716-446655440000\').';
```
SandPod, line 313: Question: Does this mean we encourage users to have a static UUID string as a default persist value?

tools/serverpod_cli/lib/src/database/extensions.dart
```diff
@@ -214,6 +214,10 @@ extension DatabaseDefinitionPgSqlGeneration on DatabaseDefinition {
     out += 'BEGIN;\n';
     out += '\n';
 
+    // Must be declared at the beginning for the function to be available.
+    out += _sqlUuidGenerateV7FunctionDeclaration();
+    out += '\n';
```
SandPod, line 219: Discussion: Should we restrict generating this to only when a UUID_V7 is used? Or is this considered "safe" for all postgres instances?

tools/serverpod_cli/test/database/migration/default_value_test.dart
```diff
@@ -359,4 +359,44 @@ void main() {
       );
     });
   });
+
+  group(
```
SandPod, line 363: Comment: Good that you thought about this case 🌟

SandPod (approved):
Changes look good overall!

Just have one comment around if we should always generate the uuid_v7 method. But since this is still an experiment we could postpone changing that.

tools/serverpod_cli/lib/src/database/extensions.dart
marcelomendoncasoares: This is actually safe to add for all implementations, because it is just a native pgqsl function. But I agree that it will be cleaner if we only declare it if the user uses an UUIDv7 value. I'll add this condition to the code.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions/default.dart
marcelomendoncasoares: This message is for usage of the `default` key, not restricted to the id change use case. For id fields, we do not allow the usage of static strings, but this behavior of allowing static on `defaultPersist` was already supported.

SandPod (changes requested):
Change the uuid_v7 definition to only occur if the user has a default uuid_v7 value defined in a model.

@nielsenko - Feel free to approve this and merge it in once that change is addressed.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/default/uuid/field_uuid_default_model_test.dart
```diff
@@ -38,6 +38,32 @@ void main() {
       },
     );
 
+    test(
+      'when the field is of type UUID and the defaultModel is set to "random_v7", then the field\'s default model value is "random".',
```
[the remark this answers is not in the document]
marcelomendoncasoares: Good catch! Done, together with the two below.

PR #3284, opened 2025-03-04
chore: Bump cli_tools dependency.

SandPod:
Bumps the `cli_tools` dependency to the latest version.

This bump required fixes for two breaking changes in the package:
- The logger interface has been changed so that what is being logged is more explicit.
- `UsageException`s can now be thrown from the `run` method and needs to be handled by the caller.

## Pre-launch Checklist

- [ ] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [ ] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
- [ ] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.
- [ ] I added new tests to check the change I am making.
- [ ] All existing and new tests are passing.
- [ ] Any breaking changes are documented below.

If you need help, consider asking for advice on the [discussion board](https://github.com/serverpod/serverpod/discussions).

## Breaking changes

None - behavior should be the same as before.

PR #3483, opened 2025-04-26
refactor: Migrated serverpod cli to cli_tools 0.5 incl Config library

christerswahn:

## Pre-launch Checklist

## Breaking changes


templates/pubspecs/tools/serverpod_cli/pubspec.yaml
```diff
@@ -27,17 +27,17 @@ dependencies:
...
   serverpod_service_client: SERVERPOD_VERSION
   async: ^2.11.0
-  meta: ^1.8.0
+  meta: ^1.11.0
   json_rpc_2: ^3.0.2
   stream_channel: ^2.1.1
   lsp_server: ^0.3.0
   pub_semver: ^2.1.4
-  cli_tools: ^0.2.0
+  cli_tools: 0.5.0-beta.5
```
SandPod, line 40: Suggestion: Shouldn't this be a stable release of `cli_tools` since it is ready?

tools/serverpod_cli/lib/src/commands/create.dart
```diff
@@ -3,9 +3,64 @@ import 'package:serverpod_cli/src/create/create.dart';
...
+  final ConfigOptionBase<V> option;
+}
+
+class CreateCommand extends ServerpodCommand<CreateOption> {
+  final restrictedNames = [
+    ...ServerpodTemplateType.values.map((t) => t.name),
+    'create',
+    'migration',
+    'repair',
+    'repair-migration',
```
SandPod, line 62: Question: What is the motivation behind adding these as restricted names?

Ensuring that the user is not actually trying to run another command?

tools/serverpod_cli/lib/src/commands/create_repair_migration.dart
```diff
@@ -7,12 +7,38 @@ import 'package:serverpod_cli/src/config/serverpod_feature.dart';
...
+    argAbbrev: 'v',
+    helpText: 'The target version for the repair. If not specified, the latest '
+        'migration version will be repaired.',
+  )),
+  mode(StringOption(
+    argName: 'mode',
+    argAbbrev: 'm',
+    defaultsTo: 'development',
+    helpText: 'Used to specify which database to fetch the live database '
+        'definition from.',
```
SandPod, line 28: Suggestion(Optional Drive by improvement): Change `Used to specify which database to fetch the live database definition from.` to `Used to specify which database configuration to use when fetching the live database definition.` since it is actually the configuration we are specifying.

tools/serverpod_cli/lib/src/commands/language_server.dart
```diff
@@ -1,7 +1,21 @@
+import 'package:cli_tools/config.dart';
 import 'package:serverpod_cli/src/language_server/language_server.dart';
 import 'package:serverpod_cli/src/runner/serverpod_command.dart';
 
-class LanguageServerCommand extends ServerpodCommand {
+enum LanguageServerOption<V> implements OptionDefinition<V> {
+  // This option has no effect! Kept for possible backwards compatibility.
```
SandPod, line 6: Comment: Good catch, is this safe to remove @Isakdl?

SandPod (commented):
Nice upgrade!

The diff was actually not that big and to the point.

Additionally to the minor comments I have two control questions:
1. Have you cloned run this locally to validate it all works as expected? (I'm not super confident we have the cli completely covered with tests).
2. This PR replaces this one: https://github.com/serverpod/serverpod/pull/3284, but hat one contains more "exit code" fixes than are represented here, have you validated that all migration fixes are included here?

tools/serverpod_cli/lib/src/commands/create.dart
christerswahn: Exactly! Had a conversation with Isak on that one, since he has encountered entering these commands by mistake after a create.

tools/serverpod_cli/lib/src/commands/language_server.dart
christerswahn: Had a quick convo with Isak on this one. It *might* be used by some invocation from an IDE, so kept for now to avoid risk of invocation error.

christerswahn:
> 1. Have you cloned run this locally to validate it all works as expected? (I'm not super confident we have the cli completely covered with tests).

Yes. When the 0.5.0 is released I will retest.

> 2. This PR replaces this one: [chore: Bump cli_tools dependency. #3284](https://github.com/serverpod/serverpod/pull/3284), but hat one contains more "exit code" fixes than are represented here, have you validated that all migration fixes are included here?

Yes all exit-error fixes from that PR are inlcuded (done before this PR actually, so don't show up in the diff).

SandPod (approved):
LGTM! 🚀

PR #3484, opened 2025-04-28
feat: Add broader `@doNotGenerate` annotation

tp:
Which subsumes the `@ignoreEndpoint` annotation, which is now marked as deprecated


packages/serverpod_shared/pubspec.yaml
```diff
@@ -16,6 +16,7 @@ dependencies:
   path: ^1.8.2
   super_string: ^1.0.3
   crypto: ^3.0.2
+  meta: ^1.8.0
```
tp, line 19: Now we restrict where the annotation can be used (in other places, developers get a lint warning). It's not very granular, but better than not having it :)

packages/serverpod_shared/lib/src/annotations.dart
```diff
@@ -1,20 +1,35 @@
...
 
 /// Name of the class needs to be synchronized with the
-/// [ServerpodAnnotationClassNames.ignoreEndpoint] constant.
-class _IgnoreEndpoint {
-  const _IgnoreEndpoint();
+/// [ServerpodAnnotationClassNames.doNotGenerate] constant.
+@Target({
+  TargetKind.classType,
+  TargetKind.method,
+})
```
SandPod, line 32: 🌟

SandPod (approved):
LGTM! 👍

packages/serverpod_shared/pubspec.yaml
SandPod: Nice addition! 👍

PR #3486, opened 2025-04-28
feat(auth2): New module `serverpod_auth_session`

tp:
Adds the `serverpod_auth_session` module for creating database-backed sessions.

When creating a session for an `AuthUser`, the caller gets a `String` `sessionKey` to give to the client, whereas only the (peppered) hash of that session key is stored on the server.

Within the provided `authenticationHandler` the `sessionKey` sent from the client will then get hashed and the DB is checked for any active session with that key (hash).

Before the client was always sending a "compound key" like "$sessionId:$sessionKey", where the `sessionId` was the database row's ID to look up the key. This is not the case anymore, and we just try to look up the key's hash. (If we decided we wanted to keep exposing the internal ID as well, we should keep the current interface though, which is `String` out, and `String` in,  so the caller does not have to "build up" the value for the client.)


modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,181 @@
...
+  /// The [key] maps to the [UserSession.sessionKeyHash] field.
+  ///
+  /// In case the authentication could not be resolved, this returns `null` and does never throw.
+  static Future<AuthenticationInfo?> authenticationHandler(
+    final Session session,
+    final String key,
+  ) async {
+    try {
+      final secret = key;
+      final signInSalt =
```
tp, line 25: To be discussed: Do we need to retain the "salting" of the session key?

I would think having a strong random key should be enough, and if the developer wants to purge all sessions, they can just drop the table (instead of changing the salt in a config).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/serverpod_auth_session_server.dart
```diff
@@ -1,2 +1 @@
-export 'src/generated/endpoints.dart';
-export 'src/generated/protocol.dart';
+export 'src/business/user_sessions.dart';
```
tp, line 1: Only export our own abstraction.

Since input/output is just `String` and `AuthInfo`, we don't even need to return the DB classes (or read-only models of them).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,186 @@
...
+  ///
+  /// In case the authentication could not be resolved, this returns `null` and does never throw.
+  static Future<AuthenticationInfo?> authenticationHandler(
+    final Session session,
+    final String key,
+  ) async {
+    try {
+      final sessionKeyHash = hashString(_signInSalt, key);
+
+      final tempSession = await session.serverpod.createSession(
```
tp, line 29: TODO: I need to understand why this is even used. Is that still relevant in the latest version of Serverpod?

As shown below, this does not work in combination with `withServerpod` integration tests for me. In those cases the `tempSession` could not read the data written with the "test session" (maybe because it runs in an open transaction?).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: Moved the pattern to use a `pepper` (static configured password) to match the email accounts.

The email accounts also use the `email` as the salt. We could also add a per-session salt if desired, just so that a leaking of the DB + configuration does not result in the attacker being (easily) able to generate the full key to make requests. Though in that case we might just check that the session key length is sufficient.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: Removed the temp session.

We'll need to see whether there is any issue with using `Session` in the authentication handler in the E2E tests later, but for now not using the `tempSession` is the only way to get the integration tests (`withServerpod`) working.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,186 @@
...
+  /// The user should have been authenticated before calling this method.
+  ///
+  /// Send the return value to the client to  use that to authenticate in future calls.
+  ///
+  /// In most situations this should not be called directly, but rather through an authentication provider.
+  ///
+  /// - `updateSession`: If set to `true`, the session will be updated with
+  ///   the authenticated user's information. The default is `true`.
+  @useResult
+  static Future<String> signInUser(
```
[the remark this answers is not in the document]
tp: Aligned the naming.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
```diff
@@ -0,0 +1,62 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_session_server/serverpod_auth_session_server.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+import 'package:test/test.dart';
+
+import '../../test/integration/test_tools/serverpod_test_tools.dart';
+
+void main() {
```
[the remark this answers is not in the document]
tp: I would suggest we add a new ticket for adapting the CI flow in order to run an integration test in a module.

Then collect the list of packages there where this should be done (in a CI matrix), and implement it post-merge.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/serverpod_auth_session_server.dart
SandPod: Doesn't the server automatically add an import towards the modules protocol and endpoint file when it is generated? I don't think it is safe to remove these.

In the chat example:
import here: https://github.com/serverpod/serverpod/blob/d49cc7babf4b3b4ce9416cbfa1c5c6c254643cb0/examples/chat/chat_server/lib/src/generated/protocol.dart#L14

Generated call to the import here: https://github.com/serverpod/serverpod/blob/d49cc7babf4b3b4ce9416cbfa1c5c6c254643cb0/examples/chat/chat_server/lib/src/generated/protocol.dart#L72

tp: 😅 Good question, because this package version is not used as an import anywhere it did not come up. But I'll try it out, but that kind of would largely defeat this barrier (which I currently feel works quite well in all the new auth packages and prevents misuse).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,190 @@
...
+  ///
+  /// The user should have been authenticated before calling this method.
+  ///
+  /// Send the return value to the client to  use that to authenticate in future calls.
+  ///
+  /// In most situations this should not be called directly, but rather through an authentication provider.
+  ///
+  /// - `updateSession`: If set to `true`, the session will be updated with
+  ///   the authenticated user's information. The default is `true`.
+  @useResult
```
SandPod, line 64: Question: When `updateSession` is true, does `@useResult` really hold true?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/models/user_session.spy.yaml
```diff
@@ -0,0 +1,29 @@
+class: UserSession
+serverOnly: true
+table: serverpod_auth_session
```
SandPod, line 3: Suggestion: Should we add a named scope for these tables?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,190 @@
...
+        ),
+      );
+    }
+
+    return key;
+  }
+
+  /// Signs out a user from the server and ends all user sessions.
+  ///
+  /// This means that the user will be signed out from all connected devices.
```
SandPod, line 105: Suggestion: Rephrase to all sessions connected to the user will be removed.

Since we now open the door for running multiple session providers we should ideally only speak on behalf of our domain.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,190 @@
...
+  /// session's authentication information will be cleared.
+  ///
+  /// Note: The method will fail silently if no authentication information is
+  /// found for the user.
+  static Future<void> destroyAllSessions(
+    final Session session, {
+    /// The user ID to sign out
+    /// If `null`, the [session]'s current user will be used
+    UuidValue? userId,
+  }) async {
```
SandPod, line 116: Suggestion: Should this be split into two distinc methods?

The use case for them is quite different, one is the admin perspective and one is forwarding the users call right?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/models/user_session.spy.yaml
```diff
@@ -0,0 +1,29 @@
...
+fields:
+  id: UuidValue?, defaultPersist=random
+
+  ### The id of the [AuthUser] this session belongs to
+  userId: UuidValue, relation(parent=serverpod_auth_user)
+
+  ### The time when this sesion was created.
+  created: DateTime, default=now
```
SandPod, line 11: Question: Should this have some form of expiry built in, or how do we reason about that?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/serverpod_auth_session_server.dart
tp: Test-added the latest version to the `serverpod_test_server`, and indeed the `Protocol` and `Endpoints` classes need to be reachable (the latter even with no endpoints defined (for now), so it would not require re-gen on the consumer).

So re-added the required exports again.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,190 @@
...
+      where: (final row) => row.userId.equals(userId),
+    );
+
+    if (auths.isEmpty) return;
+
+    // Notify clients about the revoked authentication for the user
+    await session.messages.authenticationRevoked(
+      userId,
+      RevokedAuthenticationUser(),
+    );
```
SandPod, line 132: Question: This is also a bit fishy in the context of multiple session providers. Should this call maybe be more precise? Only revoking the auths that have been found?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,190 @@
...
+
+    // Clear session authentication if the signed-out user is the currently
+    // authenticated user
+    final authInfo = await session.authenticated;
+    if (userId == authInfo?.userUuid) {
+      session.updateAuthenticated(null);
+    }
```
SandPod, line 139: Question: Is this correct? Is there ever a scenario when we are performing multi step cleanup and need the authenticated user info after this call?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,190 @@
...
+
+  /// Signs out the user from the current device by deleting the specific
+  /// session. This does not affect the user's sessions on other
+  /// devices. If the user being signed out is the currently authenticated user,
+  /// the session's authentication information will be cleared.
+  ///
+  /// If the session does not exist, this method will have no effect.
+  static Future<void> destroySession(
+    final Session session, {
+    required final String sessionId,
```
SandPod, line 150: Question: Why String and not UuidValue?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
SandPod: Same consideration for the next method.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_session_secrets.dart
```diff
@@ -0,0 +1,14 @@
...
+/// Secrets used for user sessions.
+abstract class AuthSessionSecrets {
+  /// The configuration key for the session key pepper entry.
+  static String sessionPepperKey = 'sessionHashPepper';
+
+  /// The pepper used for hashing authentication session keys.
+  ///
+  /// This influences the stored session hashes, so it must not be changed for a given deployment,
+  /// as otherwise all sessions become invalid.
+  static String? get sessionPepper =>
```
SandPod, line 12: Discussion: I think this is the "server secret" in Ruby on Rails and Django. Scoping it to the module makes it easier to rotate, so this might be a wiser approach. Otherwise, it could be nice to have a single secret used for all this.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
```diff
@@ -0,0 +1,152 @@
...
+          final authInfo = await UserSessions.authenticationHandler(
+            session,
+            sessionSecret,
+          );
+
+          expect(authInfo?.scopes, {Scope.admin});
+        },
+      );
+
+      test(
```
SandPod, line 61: Suggestion: Split the different expects into different tests.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
```diff
@@ -0,0 +1,152 @@
...
+        'when revoking a specific session for a user, then it is revoked but other sessions are unaffected.',
```
SandPod, line 62: suggestion: For this group, add tests that validate that a message is posted about revoking authentication for the user.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
```diff
@@ -0,0 +1,152 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_session_server/serverpod_auth_session_server.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+import 'package:test/test.dart';
+
+import '../../test/integration/test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod(
```
SandPod, line 9: Suggestion: Add tests that validate the `pepper` is utalized.

SandPod (changes requested):
Nice step.

Reviewing this pretty thoroughly since this is authentication related code.

Let's take the time to polish and ensure that we have though about all cases here.

I'm also wondering what header this session should be shared with? If we are going with basic as before we might want a "username" present in the session.

util/update_pubspecs
```diff
@@ -49,8 +49,10 @@ cp CHANGELOG.md modules/serverpod_auth/serverpod_auth_apple_flutter/CHANGELOG.md
 cp CHANGELOG.md modules/serverpod_auth/serverpod_auth_google_flutter/CHANGELOG.md
 cp CHANGELOG.md modules/serverpod_auth/serverpod_auth_firebase_flutter/CHANGELOG.md
 
-cp CHANGELOG.md modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_client/CHANGELOG.md
 cp CHANGELOG.md modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/CHANGELOG.md
```
SandPod, line 52: chore: Add this new package to the scripts that pulls down dependencies as well.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_session_secrets.dart
tp: I'm all for namespacing/scoping it.

I just named it `pepper` because that was used before. As I understood it, `salt` varies with each entry (and is stored alongside it), whereas a `pepper` is globally configured.

(In our case it's also stored with the "app" instead of the "db" so that just getting an old copy of one of them does not reveal enough to re-compute the original keys.)

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_session_secrets.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:serverpod/serverpod.dart';
+
+/// Secrets used for user sessions.
+abstract class AuthSessionSecrets {
+  /// The configuration key for the session key pepper entry.
+  static String sessionPepperKey = 'sessionHashPepper';
```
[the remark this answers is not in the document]
tp: > Since the password space is shared, the name should be scoped to the module.
Agreed, will change the name.

> The question is whether this should even be consumed from the passwords file.

What is the alternative? (If we want to have it on app-side configuration, instead of seeding a DB config field ourselves with some random value.)

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 1.0.0
+
+- Initial version, created by Stagehand
```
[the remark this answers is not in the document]
tp: 🙈 I guess that's what the templates has been saying since forever:

https://github.com/tp/serverpod/blame/9ef497bd75ece45f65c908c96099c743de15471b/templates/serverpod_templates/modulename_server/CHANGELOG.md#L3

CleanShot 2025-05-05 at 15 39 12@2x


Should I just remove the suffix everywhere, or do we want to force people to modify this specifically for their project?

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/migrations/20250429084917239/migration.sql
```diff
@@ -0,0 +1,60 @@
...
+
+--
+-- ACTION CREATE TABLE
+--
+CREATE TABLE "serverpod_auth_user" (
+    "id" uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+    "created" timestamp without time zone NOT NULL,
+    "scopeNames" json NOT NULL,
+    "blocked" boolean NOT NULL
+);
```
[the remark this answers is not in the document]
tp: The integration tests would not run without the migration of the other package being created.

But I can also cherry-pick this into a special PR, so we can remove it here (well not really anymore, since we don't want to remove and rebase, but at least the specific commit for this change will not result in any diff in this PR anymore).

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/src/util/authentication_info_extension.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:serverpod/serverpod.dart';
+
+/// User ID extension for `AuthenticationInfo`
+extension AuthenticationInfoUserId on AuthenticationInfo {
+  /// Returns the `Uuid` user ID of the authenticated user.
+  ///
+  /// Assumes that the system uses `Uuid` user IDs, otherwise throws.
+  UuidValue get userUuid {
```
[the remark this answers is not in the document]
tp: Opened PR in https://github.com/serverpod/serverpod/pull/3506

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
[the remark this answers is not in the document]
tp: If we're good with the approach of "integration tests in CI via matrix" I can also do that first and we can add this package to the configuration right away :)

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/migrations/20250429084917239/migration.sql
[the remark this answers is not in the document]
tp: This has now been merged into main separately, and the commit history of this PR shortened.

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/src/util/authentication_info_extension.dart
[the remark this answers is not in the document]
tp: Merged in separate PR.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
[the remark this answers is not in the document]
tp: Module integration testing is now available in the CI pipeline and this PR updated to use it for this package.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/models/user_session.spy.yaml
```diff
@@ -0,0 +1,29 @@
+class: UserSession
+serverOnly: true
+table: serverpod_auth_session
+fields:
+  id: UuidValue?, defaultPersist=random
+
+  ### The id of the [AuthUser] this session belongs to
+  userId: UuidValue, relation(parent=serverpod_auth_user)
```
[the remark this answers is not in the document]
tp: @SandPod For all of these I don't have strong preference whether to just use the ID or the `AuthUser`. Maybe the latter is better, because then at least one has the option to easily get the user (even if these modules do not make use of it).

Definitely we want to use `onDelete=Cascade` on all of these relationship, so they get cleaned up automatically. I will update those.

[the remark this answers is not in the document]
tp: I can change this to refer to `AuthUser?`, but it seems like the code-gen is currently broken in that case.

I used this relation:

```yaml
  ### The [AuthUser] this session belongs to
  authUser: module:serverpod_auth_user:AuthUser?, relation(onDelete=Cascade)
```

which generates in this Dart code:

CleanShot 2025-05-06 at 13 29 08@2x


Notice how it tries to use `AuthUser` from the `serverpod/serverpod` import instead of importing the referenced module.

I'll see that I get that fixed first.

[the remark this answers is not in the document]
tp: Ah, for the model class generation to work I had to add the module in `modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/config/generator.yaml`.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,190 @@
...
+  ///
+  /// The [key] maps to the [UserSession.sessionKeyHash] field.
+  ///
+  /// In case the authentication could not be resolved, this returns `null` and does never throw.
+  static Future<AuthenticationInfo?> authenticationHandler(
+    final Session session,
+    final String key,
+  ) async {
+    try {
+      final sessionKeyHash = hashSessionKey(key, pepper: _pepper);
```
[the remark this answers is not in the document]
tp: Sure, we could add a prefix to the key and then require that we understand/find it (instead of returning `null`).

This seems somewhat similar to what the old code did with the `$keyId:$key` pattern, in which case we would only have to index the ID and not the hash value (which is just 8 bytes instead of 32).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,190 @@
...
+
+      final authKey = await UserSession.db.findFirstRow(
+        session,
+        where: (final t) => t.sessionKeyHash.equals(sessionKeyHash),
+      );
```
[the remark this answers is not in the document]
tp: I don't think it matters if we just look up the hash (which is indexed) or ID + hash, as long as the hash is long enough (putting the ID into it just gives us 8 bytes more entropy, so we could just as well use a longer hash if we wanted to).


We could of course add an IP-based block for failed requests, to prevent brute-force attacks. I think many system would also do a "pre-check", e.g. if the secret given to the client were signed, we could verify the signature "internally" easily, and thus avoid hitting the database (which incurs some costs whether or not the session ID / hash will be found there).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,190 @@
...
+      // Setup scopes
+      final scopes = <Scope>{};
+      for (final scopeName in authKey.scopeNames) {
+        scopes.add(Scope(scopeName));
+      }
+
+      return AuthenticationInfo(
+        authKey.userId,
+        scopes,
+        authId: authKey.id!.toString(),
```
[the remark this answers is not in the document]
tp: If we remove the null check on `authKey` it would not compile.

But of course if some other code would give an a non-persisted `UserSession` with `id` that'd fail.

So I'll rewrite it to something like `final authId = authKey.id; if (authId == null) { throw 'auth key ID not set'; }`, which will have the same behavior (fail on `null`), but will be more explicit in the lower code, that we actually "grabbed" the `id`.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: This was `String` because it was historically triggered with `AuthenticationInfo.authId` (`String`), and internally broadcasts the message to invalidate that session ID as `String` as well.

But I can change it to `UuidValue`, so the caller has to provide it properly typed (which might be fine, as we could just have a `AuthInfo.sessionUuid` getter (like the user ID) per extension, and then have easy access to the typed variant).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/models/user_session.spy.yaml
tp: I think it's a general concept / issue to discuss. Personally I would just keep "created" here, and get the lifetime from the current configuration (so updates to the config apply immediately to all sessions).

But of course this prevents creating sessions with custom length, if that's more important to us, we could add an "expired" date. (Or make that optional, so we can either use the current config, or the explicit value.)

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_session_secrets.dart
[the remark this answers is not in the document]
tp: The name is scoped now (`serverpod_auth_session.sessionKeyHashPepper`).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/util/random_string.dart
```diff
@@ -0,0 +1,14 @@
+import 'dart:convert';
+import 'dart:math';
+
+final Random _random = Random.secure();
+
+/// Generates a random URL-safe string of the specified length.
+String generateRandomUrlSafeString([final int length = 32]) {
```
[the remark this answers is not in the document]
tp: Yes, this is the same one we used before:

https://github.com/tp/serverpod/blob/d563fb5235953d4e7cc5eb35e5818395c1cede3f/modules/serverpod_auth/serverpod_auth_server/lib/src/business/user_authentication.dart#L24

CleanShot 2025-05-07 at 08 53 42@2x


But I just noticed, that it's defined in `serverpod_shared` now, so I can just use that one.

Also great that this part explicitly uses `Random.secure()`, because there was some legacy config to circumvent that, but I suppose that is not needed on any platform anymore.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: Good that you question that. I will remove it, as I can't see the point really, as "create session" should be the last call in a login method.

Unless there is something specific in the logging, where the user is read at the end of the request, and thus updating the session would result in the whole transaction to be assigned to the user?

I also did not get why `scopes` would be optional, or why we don't explicitly fetch them for the user (would they really want to create sessions with different allowed scopes?). So I would move that up to be required (caller can just pass `const {}` in system where they don't use it, and drop the update session unless we have a good use case and write an integration test for that desired behavior).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: Updated the comment.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/models/user_session.spy.yaml
tp: So the old `AuthKey` did not have any expiration built into it.

To showcase what I meant with the config being applied afterwards, I have added one with a `maximumSessionLifetime` `Duration` in case someone wants to force occasional re-logins (to be refined later with the rolling approach).

But I think enabling this would also require some cooperation from the client, who at start would have to check that it's still signed-in (as before an existing session would only rarely get destroyed with the "global signout").

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: I don't see a problem with this.

The removal is explicitly on the user level, so I would think in the rare case where an app has 2 (or more) session modules active for `AuthUser`s, all of a users session should get destroyed (e.g. in the case of a password change, it should not matter what type of session you used for the change request).

If there is ever a use-case to want more control, I think we could easily send an identifier around, so session providers could filter each other out if they wanted to.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: I think it hasn't come up so far, as the same behavior was in `serverpod_auth`.

But here I think it makes sense to "downgrade" the session/authentication info (whereas the optional upgrade in "login" did not seem that useful), to prevent usage of the now deleted session.

If any code needs the user ID for a dependent call, the just have to do that call first before signing out (or store the user ID). As it's not configurable and thus deterministic, I don't think it'll create any issues in practice.

util/update_pubspecs
tp: done.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
tp: Do you have any good idea how to do that?

I can only think of adding a `@visibleForTest` overwrite allowing a unit test to add the `pepper` and then call the hash version (extracted out into a static `@visibleForTest` one), and see that we get the same result for 2 calls with 1 pepper, but a different one for another pepper?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
tp: I would not want to split these tests, as they rely on the same setup (in that the 2 sessions need to be created before one of them is revoked).

Else I would have to pass out one of the session keys, and then use that in the next test, so to understand the latter one would have the read the former.

The whole thing is a bit verbose given the APIs, but conceptually it seems quite simple to me and does exactly what could be summarized in the 1 line description.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
tp: added tests for the broadcast messages.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_session_secrets.dart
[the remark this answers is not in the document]
SandPod: Let's preserve it in the passwords file for now.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,208 @@
...
+    final secret = parts[2];
+
+    final userSession = await UserSession.db.findById(
+      session,
+      userSessionId,
+    );
+
+    if (userSession == null) {
+      throw Exception('Did not find user session with ID "$userSessionId"');
+    }
```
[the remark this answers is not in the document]
tp: But when we log, we'd have to return `null`, and I don't see why we would allow the request to proceed if an invalid session was given. In my understanding, this would lead to the user being treated as guest, instead of ending the request due to the expired/wrong session info.

> I think that throwing from this point might results in logs being created in the Server. We are probably not really interested in failed authentication attempts spamming the logs.

But then the caller can just `try/catch` and not log, if they don't want to log errors from the `authenticationHandler`, no?

Lastly, yes the old code was using `stderr` logging directly, which I would assume does not get intercepted / stored, but then in practice, who would read those server logs if they're only available on the host machine console itself?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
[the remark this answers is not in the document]
tp: > Question: Is it best practice to have a prefix when sending basic credentials?

Well, if anyone wants to understand / parse the credentials as basic auth those 2 `:` would probably be problematic.

But I don't intend for the implementation here to be "basic auth" or any specific format that would need to be handled in a special way. All this code expects, is that the returned `String` will be given back at some point to the `authenticationHandler`.

So it should just be treated as an opaque string and not interpreted in any way by outside systems (which I think is an improvement over the old system, where the in/out interfaces were not symmetric, but rather `AuthKey` and `String` respectively).

From the framework side, the returned "session key" then does (currently) get sent as a `Basic` HTTP header, but in there the value is Base64 encoded, so it would not even look like "$username:$password" when looking at the request logs:

https://github.com/tp/serverpod/blob/a64e18077c74b796e885981821136a4e4e147a69/packages/serverpod_client/lib/src/auth_key_manager.dart#L34

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: Yeah, for the dispatch the `String` was easier, but then the parsing would also have to be `try*`, so maybe this is better now (also for the logout case).

I was thinking also with regards to the `Session`/transaction discussion below, that maybe the `Session` is a bit overloaded here, as it's both what we need to access the DB (or create a temp session, which does not work in tests yet) and possibly a pointer to the current user (which might just the same as what's passed here).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
tp: As of right now I don't see a way to modify the secret/password configuration values at runtime, they just get read from the file.

But of course I could also add a lower-level `@visibleForTesting` helper to be able to overwrite on the framework level instead of just in the wrapper class for this module.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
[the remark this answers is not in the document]
tp: I can not really make a strong case for the signing.

It would spare the DB look-up in some cases, but then if someone wanted to use this for a DoS, they could just use one old valid session key, and flood us with that instead.

(As we don't yet have session key rotation capabilities, so that we could limit the lifetime and deny expired inputs.)

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: Hmm, isn't it weird though, that when looking at a user's request by ID we would then not see the login, but the logout (if it's not cleared)?

To me it does not seem ideal that the `login` request is not logged to the user ID.
In previous projects too many logins per user was often a sign that session management was broken, so I would say this is an important factor (maybe there are other ways to approximate it in some cases, like looking at the `email` or social login user ID, but not seem as straightforward).

On the flip-side, I would feel strongly about destroying a session meaning just that _right now_, and not just on the next request – so we should stick with setting that to `null`.

Maybe this also stems from the fact that `Session` is both our way of accessing the database __and__ a reference to the current user.
Could this be split up, that we say in the case of login/logout(All) we do modify the session, whereas we also provide lower-level methods which would just take DB access (a transaction?) and handle the broadcast?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/models/user_session.spy.yaml
tp: Okay, then renaming this to `AuthSession` is still open.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: > Imagine we also support JWT sessions, and the same call is made there to destroy all refresh tokens. For the valid auth sessions any streaming connections would be dropped, but we would not react and actually revoke the sessions stored, so a user could reconnect with the same session.

Why not wouldn't the `AuthSession` not be dropped in that case?

I think it's just not something this code can do (right here), but rather it has to be hooked up to the `Serverpod` `server` to handle those broadcast messages.

I think what we could do here on the `UserSessions` (or `AuthSessions` going forward) class would be to offer a helper to connect these. But that helper would then still have to be called when the application server is set up.
(Though that is not super easy, as the `channelName` contains the user ID, so we can't just listed to any auth-related message…)

For now the `serverpod_auth` also does not seem to handle this, so upon logout it just delete the session internally and then broadcasts the message to other servers.

For now, if an application really ran 2 session mechanisms, then they'd have to call both module's "logout" function in their endpoint. Is that unreasonable? If "yes", we could always use a global channel at least for the "all user sessions" drop.

(For the case where a JWT-ish token is used, it's the question anyway what kind of "auth ID" we would send around, and how other servers could reliably identify that they have open sessions for the same one.)

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: As learned when looking into the implementation for https://github.com/serverpod/serverpod/pull/3486#discussion_r2077822824, this is currently never called as a "message listener".

So the stronger interface makes sense.

The only question would be whether we even need the `sessionId` parameter, or whether it would not suffice to just operate on the `Session`'s user instead (at least in one version of the methods), to then also be cleared to manage it (and not having to check "if this is the same user, then update…").

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: Okay, but since the low-level repository functions require a full `Session` (and not just `DatabaseAccessor`) we can not resolve that ambiguity/duplication right now.

Maybe we could make it more explicit now though, by e.g. having modifying extension methods on `Session`.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/models/user_session.spy.yaml
tp: Renamed everything to `AuthSession`.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
```diff
@@ -0,0 +1,240 @@
...
+        'when creating a session for a user, then the authentication handler resolves the returned secret correctly back to the user ID.',
+        () async {
+          final sessionSecret = await UserSessions.createSession(
+            session,
+            userId: userUuid,
+            method: 'test',
+            scopes: const {},
+          );
+
+          final authInfo = await UserSessions.authenticationHandler(
```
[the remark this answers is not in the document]
tp: You're right. Added them in bc859ce083b9a563410b50e7a59100ced0ee2eac.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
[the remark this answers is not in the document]
SandPod: Hmm, but disregarding an invalid session might be OK if the endpoint does not require authentication, right?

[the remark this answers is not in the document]
tp: I guess it would do no harm to allow access to "open" endpoints, but I still feel that the request is invalid.

Because we still _do_ want to check who the request is (and internally the endpoint might use the user id, even if authentication is not required).

And allowing the request to continue would mean that it's implicitly downgraded to a guest session (instead of being for being expired / invalid).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: Decision:

- Keep the extra parameter for session / user ID, so e.g. an admin could revoke other user's sessions.
- Do not upgrade/downgrade the authentication info on the `Session`.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
tp: Decision:
- Break out the test which are not dependent on any specific previous test.
- Create test helper for initial DB setup (auth user).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
tp: Decision:
- Verify that the `sessionHash` is actually using our "peppered hash" function.
- Unit test "peppered hash" function, seeing that it changes its output when the parameter changes
- Check inside the "peppered hash" function that the pepper is useful (e.g. at least 10 characters)

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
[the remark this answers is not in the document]
tp: Decision:
- Return `null` for any authentication failures
- Log

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  /// Signs out the user from the current device by deleting the specific
+  /// session. This does not affect the user's sessions on other
+  /// devices. If the user being signed out is the currently authenticated user,
+  /// the session's authentication information will be cleared.
+  ///
+  /// If the session does not exist, this method will have no effect.
+  static Future<void> destroySession(
+    final Session session, {
```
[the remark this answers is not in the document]
tp: Decision:
- Ignore transactions for now.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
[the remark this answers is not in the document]
tp: Decision:
- Our current approach seems fine, the `String` will then be send `base64`-encoded as `Authorization: Basic *`

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: Decision:
- Update session behavior to be dropped in all instances.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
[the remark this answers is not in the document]
tp: Removed exceptions, returning `null` instead. Adding logging (sadly we can not seem to assert on those messages in test it seems).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: Removed downgrade behavior on `Session`. Made ID parameters `required`.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/user_sessions.dart
tp: Removed downgrade behavior on `Session`.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
tp: Added tests for the "pepper"

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/user_sessions_test.dart
tp: Rewrote the tests.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
```diff
@@ -0,0 +1,214 @@
...
+      );
+
+      return null;
+    }
+
+    final UuidValue authSessionId;
+    try {
+      authSessionId = UuidValue.fromByteList(base64Decode(parts[1]));
+    } catch (e, stackTrace) {
+      session.log(
```
tp, line 45: @Isakdl We wanted to ask whether you see any issue here for execution the authentication handler on the current session, as you were most deeply involved with it.

The `serverpod_auth` code was creating a temporary session like this:

```dart
    // Get the authentication key from the database
    var tempSession = await session.serverpod.createSession(
      enableLogging: false,
    );
```

And then logged directly to `stderr`.

But I feel that this is not very discoverable for the developer in production.

The only difference (for the happy path) we could think of is just, that this now will show the query trace in the logs for the request. But that just seems correct (as it happened), and thus more desirable than an inexplicable delay in request execution when tracing. (Further it might also be interesting in debugging to see when this happens, as the `AuthInfo` is sometimes build upfront (when the endpoint requires login), and sometimes just later when accessed by the implementation code.)

Isakdl: hmm the only thing that comes to mind is that all the database queries will also be logged to the session. E.g. the lookups for the auth_key table (or whatever it is called). Would this be an issue? I don't think we have any sensitive data in there right? But I suppose some of that information is security related, but I don't remember on top of my head what is stored in there.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
```diff
@@ -0,0 +1,214 @@
...
+    await session.messages.authenticationRevoked(
+      authUserId,
+      RevokedAuthenticationUser(),
+    );
+  }
+
+  /// Signs out the user from the current device by deleting the specific
+  /// session. This does not affect the user's sessions on other
+  /// devices. If the user being signed out is the currently authenticated user,
+  /// the session's authentication information will be cleared.
```
SandPod, line 178: Suggestion: Remove that the authentication information will be cleared.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_test.dart
```diff
@@ -0,0 +1,439 @@
...
+      late String sessionSecretB;
+
+      setUp(() async {
+        session = sessionBuilder.build();
+        authUserId = await _createAuthUser(session);
+        sessionSecretA = await _createAuthSession(session, authUserId);
+        sessionSecretB = await _createAuthSession(session, authUserId);
+      });
+
+      test(
```
SandPod, line 218: Suggestion: Make the "when" part a group, and then make the two expects into separate tests as they validate different things.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_test.dart
```diff
@@ -0,0 +1,439 @@
...
+            isNotNull,
+          );
+
+          expect(
+            await AuthSessions.authenticationHandler(
+              session,
+              sessionSecretB,
+            ),
+            isNotNull,
+          );
```
SandPod, line 268: Suggestion: Move this into the setup and make them into asserts.

Since these are precondition checks they just bloat the specific test and don't really make sense if these would fail in combination with the description.

SandPod (approved):
LGTM!

Last couple of minor comments to fix and we are ready to merge!

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_test.dart
tp: Removed them instead. I don't think moving up makes sense, as of course the expectation is that a session will be valid immediately after creation (in `setUp`, as also asserted by earlier tests).

This was basically just a sanity check for me that the test setup is working as expected. We start out with 2 working sessions and end with 0, which has a nice symmetry, but then I agree that the whole checking is quite verbose and thus the test may read / feel long, even though in plain English it's rather not.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_test.dart
SandPod: I think this is still good, it validates the full scenario and makes sure that we can still revoke sessions when there are multiple defined :)

Just touch up the grammar on the test descriptions. I find the first one is a bit strange.

tp: Indeed, rephrased.

Issue #3449, opened 2025-04-22
Create email authentication module


tp:
Create an email account package `serverpod_auth_email`.

This shall support account creation with email address verification, login, and password reset.

The default email account `Endpoint` will be marked as `abstract`, so consuming applications have to opt into using it. Alternatively, developers may choose to implement their own endpoints with custom requirements (e.g. only allowing certain domains, or bundling the email account with SMS 2FA, etc.).

PR #3494, opened 2025-05-01
feat(auth2): Add `serverpod_auth_email_account` package

tp:
Contains the core business logic for managing email authentication.

Adding an email authentication method is a 2-step process.
First an "account request" has to be issued with the email address and the desired password.
This will then send out an email with a verification code to the given email address.
With that code the request can then be turned into a fully active email account for a given `AuthUser`.  
(The user (ID) is only required for the second step, so the system does not get cluttered with unused/inactive data.)

Furthermore this package allows the user to reset their password. There they'll get an email with another code, and then they can set a new password with that code.

The password uses the same Argon2 hash implementation as the current `serverpod_auth`. This means that a database leak should not allow an attacker to find the passwords and thus log in undetected.

As before, all emails are stored lower-case (during and after registration).

Both the login and reset password methods are rate-limited. The "request account" feature is currently not rate-limited (but limits the registered/pending entries to 1 per email), but could likewise also be limited to a limited number of account creations for IP address.

Compared to the previous package, this does not support the "insecure random" anymore (no need), and always uses the email address as the salt for the stored password hash (which is the default in the latest version of `serverpod_auth` as well).

Relates to [#3449](https://github.com/serverpod/serverpod/issues/3449)


modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,314 @@
...
+      );
+    }
+
+    if (await _hasTooManyFailedSignIns(session, email)) {
+      throw EmailAccountLoginException(
+        reason: EmailAccountLoginFailureReason.tooManyAttempts,
+      );
+    }
+
+    if (!PasswordHash.validateHash(
```
tp, line 42: Nit: maybe we should move this into the query, so we never have the user around if the authentication was not checked.

But then that would only work for the current hash function, if we ever need to switch this in the future (as we did in the past), we'd have to check which one is employed and compute the hash correctly.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,314 @@
...
+    account ??=
+        await _importExistingUser(session, email: email, password: password);
+
+    if (account == null) {
+      throw EmailAccountLoginException(
+        reason: EmailAccountLoginFailureReason.invalidCredentials,
+      );
+    }
+
+    if (await _hasTooManyFailedSignIns(session, email)) {
```
tp, line 36: TODO: Move before existing import (to use our safeguards here).

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,314 @@
...
+    email = email.trim().toLowerCase();
+    password = password.trim();
+
+    var account = await EmailAccount.db.findFirstRow(
+      session,
+      where: (final t) => t.email.equals(email),
+    );
+
+    account ??=
+        await _importExistingUser(session, email: email, password: password);
```
tp, line 28: wrap in `try/catch` with `_logFailedSignIn`

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,314 @@
...
+      return null;
+    }
+
+    return (request.id!, request.email);
+  }
+
+  /// Finalize the email authentication creation.
+  ///
+  /// Returns the `ID` of the new email authentication, and the email address used during registration.
+  static Future<(UuidValue emailAuthenticationId, String emailAddress)>
```
tp, line 121: I wonder if named records would be preferred here.

Because there is the `emailAuthenticationId` and `emailAuthenticationRequestId`, and that is not obvious from the caller without reading the docs (plus the email being returned (instead of the whole object, which is not that useful outside), as the caller might want to create a profile).

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,314 @@
...
+  /// Sends out a password reset email for the given account.
+  static Future<void> requestPasswordReset(
+    final Session session, {
+    required String email,
+  }) async {
+    email = email.trim().toLowerCase();
+
+    final resetToken = Random.secure().nextString(length: 20);
+
+    final account = await EmailAccount.db.findFirstRow(
```
tp, line 165: TODO: Move attempt check before this line.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,314 @@
...
+      session,
+      where: (final t) => t.email.equals(email),
+    );
+
+    if (account == null) {
+      throw Exception('No account found for email.');
```
tp, line 171: Should we instead silently return here and in `requestAccount`, so that the outside shows a text: If an account exist, you'll get an email now. (And for `requestAccount`: If the account did not exist already, you'll get an email now.)

Because currently this could be used to check the existence of accounts.

But this might also be a higher-level concern, because at least for `requestAccount` we need to know whether a request was created and potentially store additional data with it.

But maybe an explicit error would also explain/document this better for the developer.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,314 @@
...
+
+    if (userId == null) {
+      return null;
+    }
+
+    return await EmailAccount.db.insertRow(
+      session,
+      EmailAccount(
+        userId: userId,
+        created: DateTime.now(), // TODO: migrate creation date?
```
tp, line 308: remove todo

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/password_hash.dart
```diff
@@ -0,0 +1,76 @@
...
+  }) {
+    if (email != email.toLowerCase()) {
+      throw ArgumentError.value(
+        email,
+        'email',
+        'Email is not given in lower-case.',
+      );
+    }
+
+    final pepper = EmailSecrets.pepper ?? ''; // TODO: Forbid `null`?
```
tp, line 29: remove TODO, but we should really forbid this 😅

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/models/email_account_password_reset_attempt.spy.yaml
```diff
@@ -0,0 +1,24 @@
+### Database table for tracking password reset attempts.
+class: EmailAccountPasswordResetAttempts
+table: serverpod_auth_email_password_reset_attempt
+fields:
+  ### Email the reset was attempted for.
+  ###
+  ### Stored in lower-case.
+  email: String
+
+  ### The time of the reset in attempt.
```
tp, line 10: ~~in~~

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Moved.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: error in import now counts as failed login as well.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: switched to named records for clarity

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Moved.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Removed todo, defaults to `now`.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/password_hash.dart
tp: Not providing the secrets now fails in non-`test` run modes. (I would also have it fail in tests, but did not see a way to provide the password to `withServerpod`. Thus leaving it open, as maybe we can support it.)

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/models/email_account_password_reset_attempt.spy.yaml
tp: Fixed typo.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/password_hash.dart
```diff
@@ -0,0 +1,74 @@
+import 'dart:convert';
+import 'dart:typed_data';
+
+import 'package:pointycastle/key_derivators/api.dart';
+import 'package:pointycastle/key_derivators/argon2.dart';
+import 'package:serverpod_auth_email_account_server/src/business/email_account_secrets.dart';
+
+/// Class for handling password hashing.
```
tp, line 8: method are taken verbatim from `serverpod_auth`.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/password_hash.dart
```diff
@@ -0,0 +1,74 @@
...
+///
+/// Uses the Argon2id algorithm.
+/// See: https://en.wikipedia.org/wiki/Argon2
+abstract final class PasswordHash {
```
tp, line 12: methods are taken verbatim from `serverpod_auth`.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/models/email_account_failed_login_attempt.spy.yaml
```diff
@@ -0,0 +1,23 @@
+### Database table for tracking failed email sign-ins. Saves IP-address, time,
+### and email to be prevent brute force attacks.
+class: EmailAccountFailedLoginAttempt
```
SandPod, line 3: Suggestion: We should add a mechanism to purge outdated entries.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,337 @@
...
+  /// Returns the [AuthUser]'s ID upon successful login.
+  ///
+  /// Throws [EmailAccountLoginException] for expected error cases.
+  static Future<UuidValue> login(
+    final Session session, {
+    required String email,
+    required String password,
+  }) async {
+    email = email.trim().toLowerCase();
+    password = password.trim();
```
SandPod, line 20: Question: Should we really trim the password? doesn't that potentially reduce entropy?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,337 @@
...
+
+    final existingAccountCount = await EmailAccount.db
+        .count(session, where: (final t) => t.email.equals(email));
+    if (existingAccountCount > 0) {
+      throw Exception('Email already registered');
```
SandPod, line 78: Suggestion: This should be a domain specific exception.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    final verificationCode = Random.secure().nextString(length: 20);
```
SandPod, line 81: Suggestion: This should be configurable, and then we should provide a default value for it.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,337 @@
...
+    return request.id!;
+  }
+
+  /// Returns the account request creation ID if the request is valid.
+  ///
+  /// If this returns a value, this means `createAccount` will succeed.
+  static Future<({UuidValue emailAccountRequestId, String email})?>
+      verifyAccountRequest(
+    final Session session, {
+    required final String verificationCode,
```
SandPod, line 107: Suggestion: Add registration process id as input parameter to this method.

This allows us to match a verification code towards a request account request attempt and not have to care about cross matches.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,337 @@
...
+  }) async {
+    final request = (await EmailAccountRequest.db.find(session,
+            where: (final t) => t.verificationCode.equals(verificationCode)))
```
SandPod, line 110: Suggestion: Add the datetime validation directly to the SQL query.

This ensures that we only allow a single attempt for each request.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,337 @@
...
+      verificationToken: verificationCode,
+    );
+
+    return request.id!;
+  }
+
+  /// Returns the account request creation ID if the request is valid.
+  ///
+  /// If this returns a value, this means `createAccount` will succeed.
+  static Future<({UuidValue emailAccountRequestId, String email})?>
```
SandPod, line 104: Question: What is the intention behind returning the email here?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,337 @@
...
+
+    return (emailAccountRequestId: request.id!, email: request.email);
+  }
+
+  /// Finalize the email authentication creation.
+  ///
+  /// Returns the `ID` of the new email authentication, and the email address used during registration.
+  static Future<({UuidValue emailAccountId, String email})> createAccount(
+    final Session session, {
+    required final String verificationCode,
```
SandPod, line 129: Suggestion: Replace the `verficationCode` with the `emailAccountRequestId`.

The verification code seems missplaced in this API, since this shouldn't really care about the validation part right? Now we "trust" the emailAccountRequestID right?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,337 @@
...
+
+    /// Authentication user ID this account should be linked up with
+    required final UuidValue authUserId,
+  }) async {
+    final request = (await EmailAccountRequest.db.find(session,
```
SandPod, line 134: Discussion: This makes creating an account dependent on an `EmailAccountRequest` first being created and a validation code existing. 

Do we really need to preserve that dependency? Couldn't that be something the user choses to use or not?
I guess magic links might be another package, but what if you would like to skip the `validation` step?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
SandPod: This is a security consideration. The requestPasswordReset endpoint should not leak whether an account exists or not so I don't think we should throw, at least not from an endpoint perspective.

Do we trust our users to always catch this exception when they override the usage of this method? Otherwise we might add this as a configurable callback instead where they can choose what should happen?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,337 @@
...
+    EmailAccountConfig.current.sendPasswordResetMail?.call(
+      email: email,
+      resetToken: resetToken,
+    );
+  }
+
+  /// Returns the auth user ID for the successfully changed password
+  static Future<UuidValue> completePasswordReset(
+    final Session session, {
+    required final String resetCode,
```
SandPod, line 205: Suggestion: Add information to uniqely identify the account to the interface here in order to limit the lookup.

This pattern is currently succeptible for brute-force attacks since we don't limit the reset code to a single entry.

SandPod: And we also return the AuthUser id for the user that you managed to match, meaning that could then be used to sign in.

SandPod (changes requested):
Fist look at the email accounts implementation.

Had a couple of comments and suggestion. Basically only inspected the `email_accounts.dart` file.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: The old code did it, but having used password managers since forever, I also don't see how that would be an issue.

maybe if someone sends you an initial password and you accidentally copy adjacent whitespace.

overall we should built-in a password policy for sure, so we can say when it's too short / not random enough etc.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Similar to the session, I would not want to burden the outside with passing around 2 parameters (process ID + verification code).

The fact that the outside only gets the process ID after verifying with this code is part of the point, so that they don't do anything with un-validated IDs.

But as in the session, we could say that our verification code is the encoded version of `$processID:$secureCode`, so we only index on the ID and can do a look up by primary key.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: User consolidation before to either look up an existing user or create a new one.

We could of course also burden the caller with storing their own e-mail copy with the process ID, but since we do have that available (and verified) here, I thought it would just be better to do it this way.

Otherwise for example the default email endpoint would need extra DB tables to store this info.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: No, this is basically the only way the module can enforce that the email is really validated, as `verify*` does not update any state.

The caller already gets the process ID on the first "registration request" (to store additional data with if they so desire), so if we just accepted that here I think it's too easy to get wrong.

Since I expect the endpoint to look like

1. `verify*` (maybe, if they need to do something internally and not always create a blank user anyway)
2. `create*`

I would think the caller has the verification code around anyway and can just give us that, and we in turn can be certain that it's valid.

Alternatively `verify*` could flip a switch in the DB and `create*` would only succeed if the flag is correct, but I feel that this is less discoverable on the API, where the current approach can not be used wrongly.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: My thinking was that unverified emails have a completely different registration flow (1 method only), and since the core assumption is still that the profile only holds verified email addresses (per our default packages at least), we needed this.

Building a non-verfied variant separately will of course have some duplication with this (e.g. resetting the password and request limitations will be similar), but on the other hand I feel this package would have a worse API if every email would have to carry the information whether or not it has been verified yet (as the configuration to enforce this could change over time).

What do you think?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: > And we also return the AuthUser id for the user that you managed to match, meaning that could then be used to sign in.

That's the whole point of the method, so that the caller can now create a new session for the user.

Guessing the reset code is basically the same as guessing the password (or email + password), so we just have to make it long enough and live short enough that it's never less secure than email + password.

As above we can always also track the "reset ID", but I would think in practice a guesser can not know the process ID any more than the code, so they would have to have both inputs randomly.
So I think blocking still works mostly on IP basis, and if we also check for the related ID that is okay but will probably never have a match in practice.

In the ends it's just important that ID + code or just the code is random / long enough to be secure.

But since the other pattern was received well for the session, I'm okay with just adapting it here as well, saving ourselves the second index.

As discussed in the sessions, we can always add another layer with signing on top, so that we can even check more without a DB read (though we will still incur the DB write for the block table…).

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Decision:

- In all these methods around "user not found" do not throw (which could inadvertently leak whether or not the account exists through the endpoint implementation)
- Instead return the status through an enum return value. The endpoint might just ignore it and always state that a mail might have been sent, whereas internal code can act upon the result.


```
Future<EmailResetStatus> resetPassword(…);
```

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/models/email_account_failed_login_attempt.spy.yaml
tp: Decision:
- Add a method to purge old entries (e.g. pass a Duration after which entries should be purge)

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Decision:
- Specific error

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Decision:
- Keep the process IDs and verification code separately
  - This allows the developer to use deep links or also "copy the code" from email (or SMS for the password reset)
- Ensure there is a verify password reset
- Use a reasonable default for the "typeable" verification code (`[a-z0-9]{6}`)

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Decision:
- `emailAccountRequestId` becomes an input parameter, with the (shorter) verification code
- Email still needs to be returned for consolidation to happen afterwards

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Decision:
- Keep this (and the default endpoint) focused on validated emails.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Deicision:
- Always use process ID + short code

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,337 @@
...
+            where: (final t) =>
+                t.email.equals(email) |
+                t.ipAddress.equals(session.remoteIpAddress)))
+        .where((final r) => r.attemptedAt
+            .isAfter(DateTime.now().subtract(const Duration(hours: 1))))
+        .length;
+
+    return (recentRequests >= 3);
+  }
+}
```
tp, line 326: @christerswahn: I thought about you request for invites, and think that this will have to be a new flow, and can not piggy-back on the registration (which is not bad, and at least keeps the interface clean).

The registration takes email/password upfront, which makes sense for password managers etc., but in the invite case we don't have that (but likely also don't want to store an 'empty' password to be set later).

So I think it makes sense to have a distinct invite concept, which is mostly like an account request, but without a password.
Upon completion, the invitee then has to give their invitation code + the new password.

That approach would be purely an extension of this approach, so I think we can tackle it in a follow-up.

I'll write it down in a ticket though, as that definitely makes sense to have :)

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Updated the implementation to return status enums and not throw in "email unknown" or "email already registered" case.

That way the caller can not accidentally leak that information.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: TIL that there is support for this like `where: (t) => t.aDateTime <= thirdDate`.

I was always looking at the auto-complete like for `.equals`, and did not see a before or after. But now this looks great :)

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,404 @@
...
+        result: EmailAccountRequestResult.emailAlreadyRegistered,
+        accountRequestId: null,
+      );
+    }
+
+    final verificationCode = Random.secure().nextString(length: 20);
+
+    final EmailAccountRequest request;
+    try {
+      request = await EmailAccountRequest.db.insertRow(
```
tp, line 100: TODO: Check pending registration, allow new one if expired.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,458 @@
...
+        email: email,
+        passwordHash: password,
+      ),
+    );
+  }
+
+  /// Cleans up the log of failed login attempts older than [olderThan].
+  static Future<void> deleteFailedLoginAttempts(
+    final Session session, {
+    required final Duration olderThan,
```
tp, line 343: TODO: When the "check time" is configurable, we could make this optional and automatically delete items older than what would be checked on subsequent requests.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/models/email_account_failed_login_attempt.spy.yaml
tp: This has been implemented.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Implemented password policy to be checked during registration.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: The verification code generators (registration and password reset) can be configured via the config now.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: This has been implemented.

We should also rate-limit verify calls now, so they can not be used to get more checks (would depend on how/whether the endpoints utilizes them, but we don't want to make it to easy to forget such a check).

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: process ID + verification code implemented for all flows.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: by default this now used a short, type-able code.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: This has been fixed to ignore expired (but not yet cleaned up) requests.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: This has been implemented and `olderThan` is now optional.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Changed to `EmailAccountRequestResult.emailAlreadyRegistered` result

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,337 @@
...
+    final request = (await EmailAccountRequest.db.find(session,
+            where: (final t) => t.verificationCode.equals(verificationCode)))
+        .where((final r) => r.created
+            .add(
+                EmailAccountConfig.current.registrationVerificationCodeLifetime)
+            .isAfter(DateTime.now()))
+        .singleOrNull;
+
+    if (request == null) {
+      throw Exception('Email account request not found');
```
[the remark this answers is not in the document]
tp: This now returns a specific error as well.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: This has to be adapted as dicussed.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Changed the data verification to happen on the DB where it makes sense.

Though in the cases where we read by ID, we'll still fetch the return, and then handle the "does not exist (anymore)" differently from "has expired".

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,657 @@
...
+  /// The caller might decide to initiate a password reset (per email, not in the client response), to help users which try
+  /// to register but already have a valid account.
+  ///
+  /// In the success case of [EmailAccountRequestResult.accountRequestCreated], the caller may store additional information
+  /// attached to the `accountRequestId`, which will be returned from [verifyAccountRequest] later on.
+  static Future<
+      ({
+        EmailAccountRequestResult result,
+        UuidValue? accountRequestId,
+      })> requestAccount(
```
tp, line 86: I would like to rename these and align them with the password reset methods.


Maybe

- startAccountCreation
- verifyAccountCreation (or `checkAccountCreation`?)
- completeAccountCreation

Then for the reset flow

- startPasswordReset
- verifyPasswordReset
- completePasswordReset

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,657 @@
...
+/// Describes the result of a password reset operation.
+enum PasswordResetResult {
+  /// A password reset email has been sent.
+  passwordResetSent,
+
+  /// No account exists for the given email.
+  emailDoesNotExist,
+}
+
+extension on Session {
```
tp, line 596: Depending on merge order, switch to implementation from `serverpod`.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Updated the naming pattern.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Merged `main` in here.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,636 @@
...
+        email = email.trim().toLowerCase();
+
+        var account = await EmailAccount.db.findFirstRow(
+          session,
+          where: (final t) => t.email.equals(email),
+          transaction: transaction,
+        );
+
+        if (await _hasTooManyFailedSignIns(session, email)) {
+          throw EmailAccountLoginException(
```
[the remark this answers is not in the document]
tp: > Question: This will leak that the email exists. Is that a concern we need to take into consideration?

I don't think it does.

The only return cases here are
a) The credentials don't work (no matter whether username or password are wrong)
b) We did not even try because the request (aka "you") are blocked


The only thing to be derived from this error would be: "If foo@bar.com is blocked, but I (my IP) is "clean"/should not be blocked, then someone else must be trying unsuccessfully to do so."
(This is due to the way that the rate limiting works on IP _or_ email, to prevent many failed attempts for the same email, or many attempts across emails from the same IP.)

Does this alleviate your concern, @SandPod, or is there still a case to be made to change this?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,636 @@
...
+  }) async {
+    final oldestRelevantAttemptTimestamp = DateTime.now().subtract(
+      EmailAccountConfig.current.maxPasswordResetAttempts.timeframe,
+    );
+
+    final recentRequests = await EmailAccountPasswordResetAttempt.db.count(
+      session,
+      where: (final t) =>
+          (t.email.equals(email) |
+              t.ipAddress.equals(session.remoteIpAddress)) &
```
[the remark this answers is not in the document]
tp: Since we're using `equals` and not `contains` or similar it would just group all "unknown" IPs into 1 bucket (but mostly that would mean the system is not configured correctly (or in a test, for my case) so that we don't get the end-users IP).

For the `''` fallback we might just want to log a warning then, as that should really not be happening when deployed.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,636 @@
...
+
+        if (request == null) {
+          throw EmailAccountRequestNotFoundException();
+        }
+
+        if (request.isExpired) {
+          throw EmailAccountRequestExpiredException();
+        }
+
+        if (request.verificationCode != verificationCode) {
```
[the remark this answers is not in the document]
tp: The amount of attempts is now rate-limited (which really makes sense given that a very short code could be configured now).

We still only delete the now when it's expired, so that all successive "invalid attempts" are still counted as such and not as "not found".
I think this will make identifying malicious request much easier.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_account_config.dart
```diff
@@ -0,0 +1,124 @@
...
+import 'package:serverpod_auth_email_account_server/src/util/verification_code_generator.dart';
+
+/// Import function to use an older email based authentication system.
+///
+/// Returns an [AuthUser] id in case the email/password combination was valid in the old system.
+/// Returns `null` if the email is not used in the old system.
+/// Throws in case the email/password combination is not valid or the user is blocked.
+///
+/// Assumes that the email has been validated in the old system.
+typedef ExistingEmailUserImportFunction = Future<UuidValue?> Function(
```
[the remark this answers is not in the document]
tp: There's certainly other ways this could be implemented, but I found this the best one to support 2 constraints:
- Do not expose any low-level account functions (which could be misused), to side-step this module's core behavior (e.g. one can not just create any account for a given email/password)
- This should run only if the email is not yet known in this module (which again would be harder to do from the outside)

I think supporting this in a hook would basically mean providing the exact same data, but then there's the added complexity that this is really a "registration in disguise", so nothing in the "hook" (when configured to support importing exsiting accounts) is related to a normal login, and would not run there.

We could of course change this to another flow offering the caller lower-level control and allowing them to import the user (if they know of it and this module does not yet) _before_ calling `login` here, but I don't think that's any clearer overall and I would really like to not add holes into the interface of this module.
True, this method is only useful for the migration, but given the history of the overall project, that's just a thing to support so why not make it simple and nice (at least this was the best I could some up with)? In there future, we can still remove this if there is no more need.

(The other thing to consider is also that we can only migrate accounts "during login", as that's the only time when we have the password (sent from the client) at hand.)

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,636 @@
...
+
+  static Future<bool> _hasTooManyFailedSignIns(
+    final Session session,
+    final String email,
+  ) async {
+    final failedLoginAttemptCount =
+        await EmailAccountFailedLoginAttempt.db.count(
+      session,
+      where: (final t) =>
+          t.email.equals(email) &
```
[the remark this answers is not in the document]
tp: This has been fixed.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/password_hash.dart
```diff
@@ -0,0 +1,74 @@
...
+        'email',
+        'Email is not given in lower-case',
+      );
+    }
+
+    final parameters = Argon2Parameters(
+      Argon2Parameters.ARGON2_id,
+      // Required cast because of a breaking change in dart 3.2: https://github.com/dart-lang/sdk/issues/52801
+      // ignore: unnecessary_cast
+      utf8.encode(email) as Uint8List,
```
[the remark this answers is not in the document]
tp: This has been changed to a random salt, which is stored in the database with the hash.

At least now every user gets the same level of added security (as configured globally).

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/password_hash.dart
```diff
@@ -0,0 +1,74 @@
...
+  }) {
+    if (email != email.toLowerCase()) {
+      throw ArgumentError.value(
+        email,
+        'email',
+        'Email is not given in lower-case',
+      );
+    }
+
+    final parameters = Argon2Parameters(
```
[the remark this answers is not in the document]
tp: The hashing has been updated to run in an `Isolate`.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,636 @@
...
+          email,
+          transaction: transaction,
+        );
+
+        if (await _hasTooManyPasswordResetAttempts(
+          session,
+          email,
+          transaction: transaction,
+        )) {
+          throw Exception('Too many password reset requests in the last hour.');
```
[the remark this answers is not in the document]
tp: This has been updated to a specific exception (of which I feel we may now have too many (12) 😅 - but at least that should give a nice overview in logging, as grouping should work much better with fixed types).

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,636 @@
...
+
+        if (resetRequest.isExpired) {
+          throw EmailAccountPasswordResetRequestExpiredException();
+        }
+
+        if (resetRequest.verificationCode != verificationCode) {
+          throw EmailAccountPasswordResetRequestUnauthorizedException();
+        }
+
+        await EmailAccountPasswordResetRequest.db.deleteRow(
```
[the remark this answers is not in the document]
tp: Rate-limiting has been added for both the email verification and password reset codes.

Rate-limiting was already in place for failed login attempts and password reset request.

The only non-limited case is currently account creation. There we could only limit by IP, if we wanted to say that there should not be more than `n` account creations per hour (which also would send out an email by default).

That might be a useful addition to tie down all APIs (I just need to do another round on the naming, as this currently might be hard to follow follow, especially the distinction between `*RequestAttempt` and `*Attempt`. Maybe we should also use start/complete, or align both to a new vocabulary.)

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/models/email_account_password_reset_request.spy.yaml
```diff
@@ -0,0 +1,14 @@
...
+  id: UuidValue?, defaultPersist=random
+
+  ### Email account this reset requests belongs to
+  emailAccount: EmailAccount?, relation(onDelete=Cascade)
+
+  ### The time when this request was created.
+  created: DateTime, default=now
+
+  ### The verification code for the password reset.
+  verificationCode: String
```
tp: Store only hash?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/models/email_account_request.spy.yaml
```diff
@@ -0,0 +1,34 @@
...
+  ### The hashed password of the user.
+  ###
+  ### Obtain in conjunction with [passwordSalt].
+  passwordHash: ByteData
+
+  ### The salt used for creating the [passwordHash].
+  passwordSalt: ByteData
+
+  ### The verification code sent to the user.
+  verificationCode: String
```
tp: Store only hash?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/password_hash.dart
```diff
@@ -0,0 +1,95 @@
...
+///
+/// Uses the Argon2id algorithm.
+/// See: https://en.wikipedia.org/wiki/Argon2
+abstract final class PasswordHash {
+  /// Create the password hash for the given [password] pair.
+  ///
+  /// Applies a random salt, which must be stored with the hash to validate it later.
+  static Future<({Uint8List hash, Uint8List salt})> createHash({
+    required final String password,
+    @protected Uint8List? salt,
```
tp: TBD: Should the email also be used as part of the salt, such that a password is not transferable from one "database row" to another?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_account_config.dart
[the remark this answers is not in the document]
tp: Alternatively this might be solvable in the endpoint (in the app or from the migration package).

```dart
class MyEmailEndpoint extends serverpod_auth_email.EmailEndpoint {
   @override
   Future<Session> login(String email, String password) {
      try {
        // would not work right now, because we don't know why it fails (from the endpoint we don't leak this info)
        return super.login(email, password);
      } on NOT FOUND {
        // migrate user
      }
   }
}
```

Decision:
- Remove the migration callback for now and let's revisit when implementing the auth migration package for email authentication

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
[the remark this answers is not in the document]
tp: Decision:
- Remove the reset request database row when too many attempts were made, so that we stop writing new "log rows" for malicious clients

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_account_config.dart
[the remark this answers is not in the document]
tp: The import has been removed for now.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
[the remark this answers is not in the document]
tp: Changed the order here to delete "early" after the number of attempts is reached.

I also had to make sure that we don't pass the transaction here, such that those items really disappear (as the next step is to always `throw` some error).

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/password_hash.dart
tp: Thought about this.

Using the `email` here would be pointless for an attacker with database access, as they could just move "their" email address as well as the password hash.

And we can not tie it to the user ID, as we don't have that yet during account creation (where we already store the password this way).
That's probably for the best though, so we can use the same approach for the verification codes as well now, and, as discussed, an attacker with database access basically already can to as they please and read all data…

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/models/email_account_password_reset_request.spy.yaml
tp: Changed to store hash & salt.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/models/email_account_request.spy.yaml
tp: Changed to store hash & salt.

tp:
> I guess invite links will be a different PR?

Yes, I will add invites separately as (in the way I currently planned them) they'll be purely an addition and won't require any changes to the entities and functions defined in here.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/config/test.yaml
```diff
@@ -0,0 +1,46 @@
...
+  port: 0
+  publicHost: localhost
+  publicPort: 0
+  publicScheme: http
+
+# This is the database setup for your test server.
+database:
+  host: localhost
+  port: 9090
+  name: projectname_test
```
SandPod: Suggestion: Replace `projectname_test` with `serverpod_auth_email_account_test`.

Apply the same fix in the `docker-compose.yaml` file.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_account_secrets.dart
```diff
@@ -0,0 +1,25 @@
...
+  /// The pepper used for hashing passwords.
+  ///
+  /// This influences the stored password, so it must not be changed for a given deployment,
+  /// as otherwise all passwords become invalid.
+  static String get passwordHashPepper {
+    final pepper =
+        Serverpod.instance.getPassword(passwordHashPepperConfigurationKey);
+
+    if (pepper == null || pepper.isEmpty) {
+      throw Exception(
```
SandPod: Suggestion: Change this to an argument error to align with the JWT implementation.

This shouldn't be catched by the developer right?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,684 @@
...
+        }
+
+        final account = await EmailAccount.db.findFirstRow(
+          session,
+          where: (final t) => t.email.equals(email),
+          transaction: transaction,
+        );
+
+        if (account == null) {
+          throw EmailAccountLoginException(
```
SandPod: Question: What did we decide regarding throwing or returning for these cases?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/password_hash.dart
```diff
@@ -0,0 +1,95 @@
...
+import 'package:pointycastle/key_derivators/api.dart';
+import 'package:pointycastle/key_derivators/argon2.dart';
+import 'package:serverpod_auth_email_account_server/serverpod_auth_email_account_server.dart';
+import 'package:serverpod_auth_email_account_server/src/business/email_account_secrets.dart';
+
+/// Class for handling password and verification code hashing in the email account module.
+///
+/// Uses the Argon2id algorithm.
+/// See: https://en.wikipedia.org/wiki/Argon2
+abstract final class EmailAccountSecretHash {
```
SandPod: Suggestion: Rename this file to match the class name.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,684 @@
...
+            reason: EmailAccountLoginFailureReason.invalidCredentials,
+          );
+        }
+
+        if (!await EmailAccountSecretHash.validateHash(
+          value: password,
+          hash: account.passwordHash.asUint8List,
+          salt: account.passwordSalt.asUint8List,
+        )) {
+          await _logFailedSignIn(session, email);
```
SandPod, line 53: Question: Should we track failed sign ins when we don't find the user as well? This will prevent the API from being hammered.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,684 @@
...
+        );
+      },
+    );
+  }
+
+  /// Returns the account request creation ID if the request is valid.
+  ///
+  /// If this returns a value, this means `createAccount` will succeed.
+  static Future<({UuidValue emailAccountRequestId, String email})?>
+      verifyAccountCreation(
```
SandPod, line 181: Discussion: What is the reason we don't update the email account request with a status saying that it is verified in this call?

Are we expecting this method to be called multiple times?

Perhaps we could have a discussion on this. Seems like it would be better to mutate the state and and not required the verificationCode again in the `complete` call.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/test/integration/email_accounts_test.dart
```diff
@@ -0,0 +1,748 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_email_account_server/serverpod_auth_email_account_server.dart';
+import 'package:serverpod_auth_email_account_server/src/generated/protocol.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+import 'package:test/test.dart';
+
+import 'test_tools/serverpod_test_tools.dart';
+
+void main() {
```
SandPod: Sugegstion: Break these tests into different files.

One file for:
- Login
- Account creation
- Password reset
- Admin methods (currently missing)

The helper methods can be contained in a shared `test_utils.dart` file.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: It was my understanding that we keep using the return value only for the "success" path, and that all errors (expected (not found, invalid credentials, expired, etc.) and unexpected) would be handle via exceptions.

Else it might be good to have something line `Result<SuccessType, KnownErrorCases>`, but since (in Dart) a function can always throw anyway, handling that does not really feel neat…

In practice (to return all expected errors in the return value) we'd probably have to create a union class with the possible cases (grouped into success & error on the top level already?). I think the only change there then would be that the _endpoint_ would have to define the model (YAML) exceptions for the cases it wants to return to the client. (This has pros and cons. Pro would be that we might give out more data, as it's only used for internal purposes (by definition), but bad since the caller might more easily leak an error case which they should not – unless the module APIs keeps guarding this away already (for the non-admin methods at least).)

tp: So long story short, I think it would be quite a departure from the current API design in Serverpod and if we do this, it should probably be something we would want to do in more places with similar "many things can expectedly go wrong".

(Here it's mainly driven by the complexity of avoiding the rollback, so that might influence our thinking in the direction of the "errors as return cases", whereas in other parts of the code we might not have such a concern.)

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Yeah, good idea, I'll move the account-exists-check in here so we treat them the same.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/config/test.yaml
tp: Makes sense, update the config files.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_account_secrets.dart
tp: This has been changed to an `ArgumentError`, matching the `session` and `jwt` modules :)

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/password_hash.dart
tp: file has been renamed :)

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Both cases are now handled the same.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Decision:
- Keep the return value for the "real" success case, use exceptions for both expected and unexpected errors.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Decision:
- Set a flag if this succeeds in the database

.github/workflows/dart-tests.yaml
```diff
@@ -258,7 +259,7 @@ jobs:
           ./wait-for-it.sh localhost:9091 -t 60 -- echo "### Redis is UP at :9091"
         working-directory: tests/docker/tests_integration
       - name: Run integration tests
-        run: dart test -t integration --reporter=failures-only
+        run: dart test -t integration -j=1 --reporter=failures-only
```
tp: FYI @SandPod

For now probably fine as each module only has 1-3 test files anyway.
If that changes, we might need to adapt the pattern from the `test_server`, to run only the needed test sequentially.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: The "verified" flag has been added to the account request, simplifying the completion call.

For the password resets I have removed the current (un-rate-limited) check, and instead the caller can just attempt the entire completion.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/test/integration/email_accounts_test.dart
tp: Tests have been split up into 4 files as suggested.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/test/integration/email_accounts_admin_test.dart
```diff
@@ -0,0 +1,172 @@
...
+        password: 'Yolo12345!',
+      );
+    });
+
+    tearDown(() async {
+      await cleanUpEmailAccountDatabaseEntities(session);
+    });
+
+    test(
+        'when `deleteExpiredAccountCreations` is called before the verification period has elapsed, then all non-expired pending request are kept.',
```
SandPod: Suggestion(NIT): Rename to `then account request is preserved.`

Since the test is talking about a specifical account request in previous steps.

SandPod: Same for the other tests.

.github/workflows/dart-tests.yaml
SandPod: I would rather preserve concurrency for all tests and add a specific tag for tests that require concurrency one.

This will make the pipeline fail if developers accidentally create tests that have a dependency between each other, giving them a chance to fix it if possible.

.github/workflows/dart-tests.yaml
```diff
@@ -259,7 +259,7 @@ jobs:
           ./wait-for-it.sh localhost:9091 -t 60 -- echo "### Redis is UP at :9091"
         working-directory: tests/docker/tests_integration
       - name: Run integration tests
-        run: dart test -t integration -j=1 --reporter=failures-only
+        run: dart test -t integration --concurrency=1 --reporter=failures-only
```
SandPod: Suggestion: Run the tests twice, once with the `tag` for concurrency_one tests and once without it in the same test run.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -350,32 +350,6 @@ abstract final class EmailAccounts {
     );
   }
 
-  /// Returns whether the password reset request is still valid.
-  ///
-  /// If this returns a value, this means `completePasswordReset` will succeed.
-  static Future<bool> verifyPasswordReset(
```
SandPod: Question: Might this ever be required in a multi factor environment? Or is that more part of "sending" the password reset?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,662 @@
...
+        transaction: transaction,
+      );
+    });
+  }
+
+  /// Cleans up the log of failed login attempts older than [olderThan].
+  ///
+  /// If [olderThan] is `null`, this will remove all attempts outside the time window that
+  /// is checked upon login, as configured in [EmailAccountConfig.emailSignInFailureResetTime].
+  static Future<void> deleteFailedLoginAttempts(
```
SandPod: Suggestion: Move to admin methods.

This shouldn't be part of the normal interface right?
Should these methods take in transaction as well?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -0,0 +1,662 @@
...
+  /// The currently active email accounts configuration.
+  static EmailAccountConfig config = EmailAccountConfig();
+
+  /// Collection of admin-related functions.
+  static final admin = EmailAccountsAdmin();
+
+  /// Returns the [AuthUser]'s ID upon successful login.
+  ///
+  /// Throws [EmailAccountLoginException] for expected error cases.
+  static Future<UuidValue> login(
```
SandPod, line 23: Discussion: Should we really opt of logging failed attempt from inside the transaction?

My concern is that the transaction might deliberately be aborted later on, but the database would still be modified preventing a loging attempt from succeding. If we preserve the behavior, we should inform the user of the interface of what is happening.

Might be best to be consistent in this regards and always use any passed in transaction for all actions?

tp: We have to log the "attempts" without the (inner) transaction so as to not reverse them immediately at the next line (throw permission error).

I would like your idea of using the "outer" transaction (given by the caller), and discard the attempt log row in case that is rolled back _explicitly_.

But I would think that in practice the outer transaction (if any) will mostly be rolled back because of the exception this code just threw. So I don't think that's an improvement in practice.

If you agree, then I would just add comments on the appropriate methods mentioned the behavior that blocks will be handled outside of the transaction logic.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/test/integration/email_accounts_admin_test.dart
tp: you're right, there is only one request in the DB at that point anyway. updated :)

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Yes, if the developer wants to enforce 2FA for the password _change_ they would need that method and we would have to store the verified state (and extend/uncap the expiration time, as we do on account creation).

I don't think this is a common approach though, as usually you could change the password if you have access to the email, and then you would have to sign in again (which might require 2FA, depending on the server's configuration).

So I don't think it's required right now, but if we want to give the developer that option (nicely) we might just re-add it, also giving a nice symmetry with account creation (the only drawback was that I felt better about always getting the code in, and not just relying on the ID for the "second/final" step).

.github/workflows/dart-tests.yaml
tp: As dicussed, this seems like a good addition, and I'll adopt the same approach as for the `test_server` for the modules as well.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Decision:
- Explain these exemptions from the transaction logic in a comment, which seems to be the best case for now.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: This has been moved. Totally right about that.

All admin methods now also take a transaction (more for consistency than that I would expect them in the future calls).

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Decision:
- Leave verify password out for now. If we re-add we should probably hand out the affected auth user ID so the caller could do something specific for them (like 2FA).

.github/workflows/dart-tests.yaml
tp: Decision:
- Add it in here, use the same tag as for `test_server`
- Long term `withServerpod` could add a special tag for all tests with `Rollback.disabled`, which is the 99% case for wanting to run without concurrency.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Comments added.

.github/workflows/dart-tests.yaml
tp: Added the `concurrency_one` tag approach in here.

PR #3495, opened 2025-05-01
feat(auth2): Implement `EmailAccountEndpoint` base class

tp:
Contains a ready-to-use endpoint for email-based user accounts, including profiles (via `serverpod_auth_profile`) and sessions (via `serverpod_auth_session`).

To be used in a project, one must subclass the `abstract` `EmailAccountEndpoint` (to make it available) and set the configuration at `EmailAccountConfiguration.current`.

In case one wants to re-use most of the behavior of this class, but implement extended endpoints (e.g. to support SMS 2FA), one can either adapt the project-specific endpoint implementation, or make use of the underlying business logic through the `EmailAccounts` class.

Closes [#3449](https://github.com/serverpod/serverpod/issues/3449)


modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
```diff
@@ -0,0 +1,152 @@
...
+
+      final authenticationInfo = await session.authenticated;
+
+      final UuidValue authUserId;
+      if (authenticationInfo != null) {
+        authUserId = authenticationInfo.authSessionId;
+      } else {
+        final newUser = await AuthUser.db.insertRow(
+          session,
+          AuthUser(created: DateTime.now(), scopeNames: {}, blocked: false),
```
tp:
```suggestion
          AuthUser(scopeNames: {}, blocked: false),
```

tp: This has been simplified, `blocked` now also defaults to `false`.

modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
```diff
@@ -0,0 +1,152 @@
...
+      session,
+      email: email,
+      password: password,
+    );
+  }
+
+  /// Completes the email account registration.
+  ///
+  /// In case the given [session] belongs to a logged-in user,
+  /// the email account will be added as an authentication method for that user.
```
SandPod: Suggestion: Make the consolidation a different endpoint.

If for some reason the client fails to remove the state of the old session (when logging out or something like that), this will accidentally link two email authentication methods to the same user.

I think this will always be a concious decision so we might as well separate it with a different endpoint.

modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
```diff
@@ -0,0 +1,152 @@
...
+  }) async {
+    return session.db.transaction((final transaction) async {
+      final accountRequest = await EmailAccounts.verifyAccountCreation(
+        session,
+        accountRequestId: accountRequestId,
+        verificationCode: verificationCode,
+        transaction: transaction,
+      );
+
+      // TODO: Add user consolidation (across providers, not via the profile anymore)
```
SandPod: Question: Do we need this TODO? We already have an issue tracking this right?

modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
SandPod: This also has the benefit of removing the branching and allowing us to throw if the user is not already authenticated.

modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
```diff
@@ -0,0 +1,152 @@
...
+
+      final authenticationInfo = await session.authenticated;
+
+      final UuidValue authUserId;
+      if (authenticationInfo != null) {
+        authUserId = authenticationInfo.authSessionId;
+      } else {
+        final newUser = await AuthUser.db.insertRow(
```
SandPod: Question: Don't we have business logic methods for this?

SandPod (approved):
LGTM with the suggested fix.

modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
tp: I also wondered whether this is maybe too much to express here, and should be just its own thing.

Because it somehow feels weird to specify this behavior in "step 2".

We could do it in "step 1" (only for the current user), but then you would not need the new session afterwards.

Currently, I feel like we should drop adding accounts to existing (for now, until we have a good/clear path), and first focus on the invites (which seems really desired and critical), and then we can revisit this later.
Adding to an existing account might then be best revisited once we have a social login implemented E2E, so we can find an API flow that is generally applicable.

modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
tp: removed, yes, we won't forget about this :)

modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
tp: will add an `AuthUsers` class per our latest pattern, and avoid the manual inserts everywhere else.

PR #3507, opened 2025-05-05
feat(auth2): Create `serverpod_auth_user_server` SQL migration

tp:
So consuming packages get the latest SQL schema.


modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/migrations/migration_registry.txt
```diff
@@ -5,3 +5,4 @@
 ### the conflict by removing and recreating the conflicting migration.
 
 20250416124930715
+20250429084917239
```
SandPod, line 8: Suggestion: Replace the already existing migration. We don't need the initial step since we are introducing the package for the first time.

SandPod (approved):
One thing to fix, ping me when it is done and I can merge this.

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/migrations/migration_registry.txt
tp: Updated to only use 1 step.

PR #3509, opened 2025-05-05
docs: Adds example from docs in three iterations.

vlidholt:
This adds example from documentation, in three different stages.


marcelomendoncasoares:
Closing this since it is stale for a long time. If needed, we can reopen from a new branch (and with the examples updated).

PR #3513, opened 2025-05-06
feat: Add module integration test step to CI pipeline

tp:
To run each module's `withServerpod` integration tests.

Adds the (now default) `docker-compose.yaml` file to `serverpod_test_module_server` to showcase the new behavior.

### DIFF
--- a/tests/docker/tests_integration/run_tests.sh
+++ b/tests/docker/tests_integration/run_tests.sh
@@ -29,7 +29,3 @@ dart bin/main.dart -m production -r maintenance --apply-migrations
 echo "### Running tests"
 dart test test_integration -x integration --concurrency=1
 checkLastExitCode
-
-cd ../serverpod_test_module/serverpod_test_module_server/
-echo $(pwd)
-dart test ./test/integration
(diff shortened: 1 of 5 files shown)


tests/docker/tests_integration/run_tests.sh
```diff
@@ -29,7 +29,3 @@ dart bin/main.dart -m production -r maintenance --apply-migrations
 echo "### Running tests"
 dart test test_integration -x integration --concurrency=1
 checkLastExitCode
-
-cd ../serverpod_test_module/serverpod_test_module_server/
```
tp, line 33: Removed the "custom" integration test for the test module server, which can now be handled through the general CI setup which would work for all new modules created from the template (as those contain a `docker-compose` file).

tp:
The new approach with `-t integration` also runs the tests fine:

CleanShot 2025-05-06 at 11 16 54@2x

There is just some issue with the "wait for service", which times out, but then the test code (with runs anyway) can access the DB just fine…

nielsenko (approved):
LGTM

Issue #3503, opened 2025-05-05
Object relation towards new table in existing model creates invalid migrations.


SandPod:
### Describe the bug

When adding a new table model and referencing it from an existing table via a foreign key, the generated migration will alter the existing table before the new table is created if the existing table has a name that would be sorted lexically sorted before the next table. This results in invalid SQL due to missing referenced table.

### To reproduce

1. Start with a model that already exists, e.g.:

```yaml
# existing_model.yaml
class: ExistingModel
table: a_existing_model <-- `a` assures that it is lexically sorted before the new table.
fields:
  name: String
```

2. Add a new model and update the existing one with a foreign key:

```yaml
# new_model.yaml
class: NewModel
table: z_new_model <-- `z` assures that it is lexically sorted after the old table.
fields:
  description: String

# existing_model.yaml (updated)
class: ExistingModel
table: a_existing_model
fields:
  name: String
  newModel: NewModel?, relation(optional)
```

3. Run `serverpod create-migration` and inspect the created migration.

### Expected behavior

The generated migration should create `new_model` first, before adding foreign key constraints to `existing_model`. The correct order avoids referencing a non-existent table.

### Library version

- Serverpod 2.6.0 (or current main at the time of writing)

### Platform information

All platforms - bug is in the generated SQL.

### Additional context

Fixing this likely involves sorting migration operations so that all `CREATE TABLE` statements are placed before `ALTER TABLE ADD CONSTRAINT`.

### How experienced are you with this library?

None

### Are you interested in working on a PR for this?

- [ ] I want to work on this

PR #3516, opened 2025-05-06
fix: Ensure foreign key relations are always defined after table creation.

SandPod:
Fixes: #3503 

When foreign keys where created as part of altering migrations in the migration system these where placed together with the other alterations to the table and not last in the file.

This caused an issue when a new table was referenced from an old table, but the creation of the new table was placed after the alterations to the old table in the migration file.

More concrete examples of the issue are included in the tests.

**Additional question**

The helper introduced in `migration_pgsql_code_generation_test.dart` could be extracted and used in `migration_with_many_to_many_relation_pgsql_code_generation_test.dart`. This would significantly reduce the footprint of that file while abstracting away non-essential boilerplate.

Should I migrate the old file?

## Pre-launch Checklist

## Breaking changes

None - Simply fixes a bug in the migration system

### DIFF
--- a/tests/serverpod_test_server/test_e2e_migrations/migrations_roundtrip_test.dart
+++ b/tests/serverpod_test_server/test_e2e_migrations/migrations_roundtrip_test.dart
@@ -1124,4 +1124,91 @@ fields:
       expect(migrationRegistry.versions, isNot(contains(tag)));
     });
   });
+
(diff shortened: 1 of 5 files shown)


nielsenko (approved):
LGTM.

Regarding:
> Should I migrate the old file?

Yes please 😄

PR #3518, opened 2025-05-07
fix: Respons with 503 when server is unhealthy

Isakdl:
# Changes

Previously, the health check responded with status code 200 even when the services was unhealthy. This will be interpreted as healthy by other systems such as when a loadbalancer checks if the server is available. 

To make sure traffic is not routed to unhealthy instances we now give back 503 instead.

## Pre-launch Checklist

## Breaking changes

None


SandPod (approved):
Changes look good but make sure the ingration test is not flaky.
Might as well patch that directly.

Issue #3504, opened 2025-05-05
Race condition when applying migrations in multiple test files causes failures


dkbast:
### Describe the bug

When running integration tests in projects with **more than one test file with withServerpod tests**, a race condition occurs. Each test file's `setUpAll` method simultaneously attempts to apply database migrations. This leads to conflicts, often resulting in errors like:

```
DatabaseQueryException: duplicate key value violates unique constraint "pg_class_relname_nsp_index"
```

This happens because the migrations are applied concurrently, resulting in duplicate or conflicting schema changes.

### To reproduce

1. Create a project with multiple test files using Serverpod and integration tests.
2. Run all tests at once for the first time using `dart test .`.
3. Observe that multiple migrations are applied simultaneously, causing failures like duplicate key errors.

### Expected behavior

Migrations should only be applied **once per test run**, even when multiple test files are present. Running `dart test .` should not cause race conditions or database schema conflicts.

### Library version

Serverpod 2.6.0

### Platform information

Should happen everywhere

### Additional context

The Dart `test` framework does **not support global hooks** to run logic once before all test files. Each file’s `setUpAll` runs independently.  
Flutter has a similar limitation (`flutter_test_config.dart` runs once per file).
See https://github.com/dart-lang/test/blob/master/pkgs/test/doc/configuration.md
And https://github.com/flutter/flutter/blob/master/examples/api/test/flutter_test_config.dart

### Suggested interim workaround
Use a wrapper script to start the database and apply migrations once before running tests. Example:

```bash
# tool/test_with_setup.sh
./tool/start-test-db-and-apply-migrations.sh
dart test "$@"
```

A long-term fix would be to create a "serverpod test" command that wraps the dart test command and makes sure that everything is setup properly. In that case it would be great to add a way to do other work here as well (pulling fixtures etc pp)

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

PR #3521, opened 2025-05-07
feat: Set module integration test concurrency to 1

tp:
Because the DB setup/migrations across `withServerpod` calls in multiple test files are not synchronized, so they can create conflicts when using 2 or more such calls.

As these are modules, there is no `--apply-migrations` to be done outside the tests (as far as I can see).

Further this removes the test volumes from Docker after a test completed. (This was not an issue in practice, but seems cleaner.)

This change fixes the previously broken module test in https://github.com/serverpod/serverpod/actions/runs/14883049192?pr=3441, which was using 2 `withServerpod` calls.

### DIFF
--- a/.github/workflows/dart-tests.yaml
+++ b/.github/workflows/dart-tests.yaml
@@ -261,11 +261,11 @@ jobs:
           ./wait-for-it.sh localhost:9091 -t 60 -- echo "### Redis is UP at :9091"
         working-directory: tests/docker/tests_integration
       - name: Run integration tests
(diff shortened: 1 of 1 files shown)


SandPod (changes requested):
This is an issue that is actively being worked on currently: https://github.com/serverpod/serverpod/issues/3504

We should have a real fix by tomorrow.

SandPod (approved):
Letäs take this in now in order not to block the module work.

Then I can remove it as part of the fix.

PR #3527, opened 2025-05-09
fix: Prevent possibility of multiple migrations being applied at the same time.

SandPod:
Closes: #3504 

This PR introduces an advisory lock that is used whenever migrations towards the database are applied.

This ensures that even in an environment where mutliple servers have the `--apply-migration` flag turned on, only one of them will make modifications to the database at the same time.

This makes the full migration script atomic.

This issue was encountered when running integration tests using the test framework since that would spawn multiple isolates that each would start a Serverpod server, resulting in a race condition when applying the migrations.

### Additional changes
- Makes the MigrationManager private on Serverpod. This is technically a breaking change, but it should never be used by external users.
- Removes updating the state of the MigrationManager when fetching the live database definition. Since the migrations are only run on Server start, updating the state is redundant.
- Refactor Serverpod startup to isolate applying the migration from other database operations. This allows us to easily expose running the migrations at a later step.

### Test question

I'm not sure how to reliably test this. The last commit enables tests that would otherwise fail if the initialization did not work as expected.

But I think that any test we could create would have the chance of producing false positives without a complicated test setup (custom migrations with delays).

I have manually validated the changes to ensure they work as expected.

## Pre-launch Checklist

## Breaking changes

None - this should simply be a bug fix.

### DIFF
--- a/packages/serverpod/lib/src/database/migrations/migration_manager.dart
+++ b/packages/serverpod/lib/src/database/migrations/migration_manager.dart
@@ -200,7 +200,9 @@ class MigrationManager {
     availableVersions.clear();
     var warnings = <String>[];
     try {
-      availableVersions.addAll(MigrationVersions.listVersions());
+      availableVersions.addAll(MigrationVersions.listVersions(
+        projectDirectory: _projectDirectory,
+      ));
     } catch (e) {
       warnings.add(
           'Failed to determine migration versions for project: ${e.toString()}');
--- a/packages/serverpod/lib/src/database/migrations/migration_manager.dart
+++ b/packages/serverpod/lib/src/database/migrations/migration_manager.dart
@@ -52,7 +52,7 @@ class MigrationManager {
 
   /// Returns true if the database structure is up to date. If not, it will
   /// print a warning to stderr.
-  Future<bool> verifyDatabaseIntegrity(Session session) async {
+  static Future<bool> verifyDatabaseIntegrity(Session session) async {
     var warnings = <String>[];
 
     var liveDatabase = await DatabaseAnalyzer.analyze(session.db);
--- a/packages/serverpod/lib/src/database/migrations/migration_manager.dart
+++ b/packages/serverpod/lib/src/database/migrations/migration_manager.dart
@@ -21,9 +21,9 @@ class MigrationManager {
   /// directory. Available after [initialize] has been called.
   final List<String> availableVersions = [];
 
-  /// Initializing the [MigrationManager] by loading the current version
+  /// Updates the state of the [MigrationManager] by loading the current version
   /// from the database and available migrations.
-  Future<void> _initialize(Session session) async {
+  Future<void> _updateState(Session session) async {
     installedVersions.clear();
     try {
       installedVersions.addAll(await DatabaseMigrationVersion.db.find(session));
@@ -158,7 +158,7 @@ class MigrationManager {
         repairMigration.sqlMigration,
       );
 
-      await _initialize(session);
+      await _updateState(session);
     });
 
     return appliedVersionName;
@@ -172,7 +172,7 @@ class MigrationManager {
     List<String>? migrationsApplied = [];
 
     await _withMigrationLock(session, () async {
-      await _initialize(session);
+      await _updateState(session);
       var latestVersion = getLatestVersion();
 
       var moduleName = session.serverpod.serializationManager.getModuleName();
@@ -189,7 +189,7 @@ class MigrationManager {
         latestVersion: latestVersion,
         fromVersion: installedVersion,
       );
-      await _initialize(session);
+      await _updateState(session);
     });
 
     return migrationsApplied;
--- a/packages/serverpod/lib/src/database/migrations/migration_manager.dart
+++ b/packages/serverpod/lib/src/database/migrations/migration_manager.dart
@@ -53,7 +53,7 @@ class MigrationManager {
 
   /// Returns the installed version of the given module, or null if no version
   /// is installed.
-  String? getInstalledVersion(String module) {
+  String? _getInstalledVersion(String module) {
     var installed = installedVersions.firstWhereOrNull(
       (element) => element.module == module,
     );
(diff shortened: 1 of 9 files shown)


packages/serverpod/lib/src/endpoints/insights.dart
```diff
@@ -212,9 +212,6 @@ class InsightsEndpoint extends Endpoint {
     // Get database definition of the live database.
     var databaseDefinition = await DatabaseAnalyzer.analyze(session.db);
 
-    // Make sure that the migration manager is up-to-date.
-    await session.serverpod.migrationManager.initialize(session);
-
```
SandPod, line 217: @Isakdl - I see no reason to keep this since the migration manager isn't responsible for producing the analysis and we don't support any "remote" actions. Now that it is also an instance that lives as long as migrations should be applied the state is not important.

Objects to removing it?

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -142,20 +142,6 @@ class Serverpod {
     return server;
   }
 
-  MigrationManager? _migrationManager;
-
-  /// The migration manager used by this [Serverpod].
-  MigrationManager get migrationManager {
```
SandPod, line 148: @Isakdl - Technically a breaking change, but I think it is OK to remove it right?

packages/serverpod/lib/src/database/migrations/migration_manager.dart
```diff
@@ -141,42 +141,85 @@ class MigrationManager {
...
+    ///
+    /// This ensures that we are only running migrations once at a time.
+    await session.db.transaction((transaction) async {
+      await session.db.unsafeExecute(
+        "SELECT pg_advisory_lock(hashtext('$lockName'));",
+        transaction: transaction,
+      );
+
+      try {
+        await action();
```
tp, line 215: Shouldn't the `transaction` be passed to `action`?

Else I don't get the need for the transaction, as exclusivity (and clean up) is already given, even without the transaction.

As the lock used here is always "session-level", it does not need or "respect" a transaction as I understand it:

> Once acquired at session level, an advisory lock is held until explicitly released or the session ends. Unlike standard lock requests, session-level advisory lock requests do not honor transaction semantics: a lock acquired during a transaction that is later rolled back will still be held following the rollback, and likewise an unlock is effective even if the calling transaction fails later. 

From https://www.postgresql.org/docs/current/explicit-locking.html#ADVISORY-LOCKS

Per https://www.postgresql.org/docs/17/functions-admin.html I think `pg_advisory_xact_lock` would be needed to get a transaction-level lock, but then that would probably not be visible to the other transaction (unless they "READ UNCOMMITTED"), so that does not seem appropriate here.

packages/serverpod/lib/src/database/migrations/migration_manager.dart
```diff
@@ -141,42 +141,85 @@ class MigrationManager {
...
+    const String lockName = 'serverpod_migration_lock';
+
+    /// Use a transaction to ensure that the advisory lock is retained
+    /// until the transaction is completed.
+    ///
+    /// This ensures that we are only running migrations once at a time.
+    await session.db.transaction((transaction) async {
+      await session.db.unsafeExecute(
+        "SELECT pg_advisory_lock(hashtext('$lockName'));",
+        transaction: transaction,
```
tp, line 211: See the comment below. Reading up on this, I don't think the transaction here has any effect.

packages/serverpod/lib/src/database/migrations/migration_manager.dart
tp: > The transaction is only there to ensure that a single connection is used from the connection pool.

Got it. So this is not really about the transaction on the Postgres level, but just for giving you a persistent session/connection from the DB driver.

Now this seems fine to me then, and the transaction indeed required (such that the session does not end early and loose the lock).

I am not sure though if this behavior/reason for the transaction is clear to everyone else but me, so maybe we can extend the comment.

packages/serverpod/lib/src/database/migrations/migration_manager.dart
```diff
@@ -141,42 +141,85 @@ class MigrationManager {
...
+  }
+
+  Future<void> _withMigrationLock(
+    Session session,
+    Future<void> Function() action,
+  ) async {
+    const String lockName = 'serverpod_migration_lock';
+
+    /// Use a transaction to ensure that the advisory lock is retained
+    /// until the transaction is completed.
```
tp, line 205:
```suggestion
    /// until the transaction is completed.
    ///
    /// The transaction ensures that the session used for acquiring the
    /// lock is kept alive in the underlying connection pool, and that we
    /// can later use that exact same session for releasing the lock.
    /// The transaction is thus only used to get the desired behavior from
    /// the database driver, and does not have any effect on the Postgres level.
```

I think since this is just an implementation comment, we can be very verbose / comprehensive, so that this does not get message up in the future by someone (like me 10 minutes ago) trying to optimize this ;)

tp (approved):
The specific locking changes look good to me, now that I understand the need for the `transaction`.

Will defer to @Isakdl for the refactorings.

packages/serverpod/lib/src/database/migrations/migration_manager.dart
SandPod: Thanks for the suggestion! I'll use this straight of :)

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -526,12 +512,26 @@ class Serverpod {
...
     _databasePoolManager?.start();
 
-    if (_databasePoolManager == null) {
-      _runtimeSettings = _defaultRuntimeSettings;
-    }
-
     if (Features.enableMigrations) {
-      await _applyMigrations();
+      int? maxAttempts =
+          commandLineArgs.role == ServerpodRole.maintenance ? 6 : null;
```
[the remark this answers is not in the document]
SandPod: This is the only place that this value is used, and it depends on some specific parameter logic.

Since this class is so large, I would opt for keeping it here for now. I agree this is a smell.

SandPod:
@Isakdl Comments:
- [ ] Clean up the rest of the MigrationManager interface, since it is only used internally.
- [ ] Create a specialized test that assures that the lock works as intended.
    - We should be able to achieve this by creating an integration test with specific migration files.

Issue #2871, opened 2024-10-16
We should add a test that changes the version of Serverpod


vlidholt:
When adding new packages to Serverpod or changing dependencies, its the tests sometimes won't break until we change the `SERVERPOD_VERSION`. We should add a test that changes the version, then runs `util/update_pubspecs`, followed by `dart analyze`. This will ensure that publishing new versions of Serverpod will be a smoother experience.

PR #3538, opened 2025-05-12
fix(auth2): Use `dependency_overrides` in new authentication client modules

tp:


modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_client/pubspec.yaml
```diff
@@ -8,3 +8,7 @@ environment:
 
 dependencies:
```
SandPod, line 9: Suggestion: Add dependency towards `serverpod_auth_user_client`. 

This package also needs a dependency and a dependency override towards `serverpod_auth_user_client`. It is referenced in the `protocol.dart` file.

SandPod (changes requested):
Could you add the modules to the `./util/run_tests_analyze` script, this is why we didn't catch this in our CI.

Perhaps it is time to break out that list to a separate file and then iterate that file in the scipts instead so that it is applied in `pub_get_all`, `generate_all`, `update_pubspecs` and `run_tests_analyze`.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_client/pubspec.yaml
tp: Oh, should've run `analyze` in addition to `pub get`.

Thanks for the right pointer on what script was missing for CI to catch this.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_client/pubspec.yaml
```diff
@@ -8,3 +8,10 @@ environment:
...
   serverpod_client: 2.6.0
+  serverpod_auth_user_client: 2.6.0
+
+dependency_overrides:
```
SandPod, line 13: Suggestion: Add a dependency override towards `serverpod_serialization` in both client packages.

This is a transitive dependency of the package and will cause an error if we update the serverpod package version in the project.

Check out `modules/serverpod_auth/serverpod_auth_client/pubspec.yaml` so see this.

We have an issue for adding an automated test for detecting this that we have not had time to complete yet: https://github.com/serverpod/serverpod/issues/2871

tp: Ah interesting, I saw that ticket a while back, but only now understand the ramifications :)

tp:
@SandPod This unrelated test fails:

CleanShot 2025-05-13 at 09 35 19@2x


CleanShot 2025-05-13 at 09 36 36@2x


Since there is no "order by" specified, I think the test should be updated to use `unorderedEquals`.

(Probably applies to the test above as well, though this did not fail in here.)

I'll put up a separate PR for that fix.

PR #3545, opened 2025-05-14
docs: Add unit to future call scan interval config example.

SandPod:
Adds the time unit to the scan interfal config example.

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/templates/serverpod_templates/projectname_server_upgrade/config/development.yaml
+++ b/templates/serverpod_templates/projectname_server_upgrade/config/development.yaml
@@ -53,4 +53,4 @@ sessionLogs:
 
 #futureCall:
 #  concurrencyLimit: 1 # Defaults to 1, a negative or null value removes the limit
-#  scanInterval: 5000 # Defaults to 5000
+#  scanInterval: 5000 # Unit in milliseconds, defaults to 5000ms
--- a/templates/serverpod_templates/projectname_server_upgrade/config/production.yaml
+++ b/templates/serverpod_templates/projectname_server_upgrade/config/production.yaml
@@ -57,4 +57,4 @@ sessionLogs:
 
 #futureCall:
 #  concurrencyLimit: 1 # Defaults to 1, a negative or null value removes the limit
-#  scanInterval: 5000 # Defaults to 5000
+#  scanInterval: 5000 # Unit in milliseconds, defaults to 5000ms
--- a/templates/serverpod_templates/projectname_server_upgrade/config/staging.yaml
+++ b/templates/serverpod_templates/projectname_server_upgrade/config/staging.yaml
@@ -61,4 +61,4 @@ sessionLogs:
 
 #futureCall:
 #  concurrencyLimit: 1 # Defaults to 1, a negative or null value removes the limit
-#  scanInterval: 5000 # Defaults to 5000
+#  scanInterval: 5000 # Unit in milliseconds, defaults to 5000ms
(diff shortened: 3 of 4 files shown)


tp (approved):

templates/serverpod_templates/projectname_server_upgrade/config/development.yaml
```diff
@@ -53,4 +53,4 @@ sessionLogs:
 
 #futureCall:
 #  concurrencyLimit: 1 # Defaults to 1, a negative or null value removes the limit
-#  scanInterval: 5000 # Defaults to 5000
+#  scanInterval: 5000 # Unit in milliseconds, defaults to 5000ms
```
tp, line 56:
```suggestion
#  scanInterval: 5000 # Unit in milliseconds, defaults to 5s
```

@SandPod I also meant whether we should name the unit also in the variable name, but that ship has sailed already probably.

Would it make sense to vary the text with the `5s`?
Though I suppose no one is surprised by that conversion…

SandPod: Aah, I understand! We can probably fix that if it is important, just need to ensure that the old one "works" but we don't need to mention it.

But we have not included the unit in `maxRequestSize` so the pattern currently is not to include it.
Maybe the unit should be removed from the value completely in that case? 

Simply this to match the default value?
`# scanInterval: 5000 # Unit in milliseconds, defaults to 5000´

christerswahn: Just a comment from out of left field - in the config library, duration values can be specified with a unit, e.g. 5000ms or 5s, where seconds are the default. If this is made compatible with that (e.g. requiring ms for now), it will be compatible upon a possible later migration.

SandPod: That is even better! 🙌 

This is possible once we have migrated our config parsing in Serverpod to use the CLI tools' config parsing.

Could we make it compatible with that by simply allowing any `non unit`value to be parsed as `ms`?

christerswahn: That would be a different default behavior, so existing configs would suddenly be reintrepreted = not good! If we want to make it forward-compatible, it should either be seconds (but then users can't specify milliseconds), or require a 'ms' suffix, or... use the DurationParser from cli_tools right away! It can be used in separation:

```dart
DurationParser().parse('4711ms') => Duration(milliseconds: 4711)
```

SandPod: > That would be a different default behavior, so existing configs would suddenly be reintrepreted = not good!

I think you misunderstood me. This is already shipped to our users, so the existing behavior now is to interpret any number as `ms`. My thought is that we can preserve that, but also add the possibility to specify the unit. Would that not be possible?

Or does the `cli_tools` parsing have a hard rule that any non-unit number is in `seconds`?

christerswahn: Ok I see. The DurationParser is currently hardcoded for seconds as default unfortunately. Not a difficult change, but it would need a new cli_tools release.

You could just copy the code for now, it's tested, and change that:

```
```

christerswahn: cli_tools 0.5.1 will support ms :)

SandPod: Aah ok! Got it! Appending would work as wel,l but if it is already incoming, we can wait for `0.5.1` :)

Since there is no "cow on the ice" for this one.

Issue #3549, opened 2025-05-15
Add listener to shutdown process


inf0rmatix:
### Problem to Solve

When Serverpod handles shutting down (due to sig-int or any other reason), I need to be notified and have a listener executed. The server has to await completion of that method before calling exit.

### Proposal

- Add a ShutdownManager class.
- Make the instance available through Serverpod instance.
- Have the ShutdownManager class expose a public method `addListener`.
- Add a `ShutdownManager.handle()` method which ensures any listeners are completed before completing itself.
- When stopping the server, add `await ShutdownManager.handle()`.

### Use Case

Graceful quitting of long-running processes outside FutureCalls.

### Alternatives

none. Since Serverpod itself calls "exit" which can be called only once, terminating the process immediately.

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

PR #3555, opened 2025-05-19
feat: Added ShutdownListenerManager for Graceful Server Shutdown Handling

yashas-hm:
This PR introduces a new `ShutdownListenerManager` class that enables the registration of asynchronous listeners to be executed when the Serverpod server is shutting down. This is critical for gracefully terminating long-running processes, closing resources, or performing final clean-up operations before the server exits.

---

## Features

- Add shutdown listeners with optional identifiers
- Listeners are executed **sequentially** and **awaited properly**
- Exceptions in listeners are caught and logged without halting shutdown
- Simple, extensible interface for integration into Serverpod core

---
This feature solves that problem by allowing the server to:

- Wait for custom asynchronous tasks to complete
- Cleanly close DB or cache connections
- Save or flush logs
- Terminate internal schedulers or workers gracefully

---

## Example Usage

```dart
serverpod.shutdownManager.addListener(
  () async {
    await backgroundWorker.stop();
    print('Worker shut down gracefully.');
  }, 
  listenerIdentifier: "Shutdown Background workers",
);
```
Closes: #3549 

## Pre-launch Checklist

## Breaking changes

## Summary by CodeRabbit

- **New Features**
  - Introduced a structured task management system for server shutdown, enabling registration and management of custom shutdown tasks.
  - Added a public interface for adding and removing shutdown tasks on the server.
- **Bug Fixes**
  - Enhanced error handling during server shutdown to capture and report errors per task.
- **Tests**
  - Added comprehensive tests verifying shutdown task execution, error handling, concurrency, and shutdown order.


yashas-hm:

SandPod:
@yashas-hm - Thank you for taking the initiative to contribute and improve Serverpod!

I have a couple of initial comments that might be good to discuss around the solution.

### 1. Naming

Since the system awaits the run of the "task" that is scheduled, I don't think `listener` is a good name, as it communicates something is simply notified and not awaited.

How about we rename it to `ShutdownTaskManager` and we call adding something `addTask` instead?
I think this more clearly communicates that there is a task that is expected to finish and not an event that is broadcast.

### 2. Sequential execution

How important do we believe sequential exection to be between different tasks? Couldn't we simply run all tasks async, like the already awaited futures in the shutdown protocol. 

If sequential execution is required, this could simply be bundled in the same task. This would encourage developers not to create dependencies between tasks.

### 3. Standardize the solution

Could the same solution be used to run the tasks that we already perform, such as closing the internal session or shutting down the server?

If we introduce a system for registering shutdown tasks, I see no reason not to use the same system for performing the existing shutdown tasks.

### 4. Dependencies (Scope creep)

For some shutdown tasks, there might be a dependency between them, such as stopping the `databasePoolManager`. Would it be possible to express this dependency in the shutdown system in a way that ensures that a task is completed before another?

This would allow us to run the task that stops the `databasePoolManager` after all other tasks have completed.

Perhaps a task could be registered with a key that could be used when registering other tasks, `addTaskBefore` or `addTaskAfter`?

yashas-hm:
@SandPod Thanks, those are some great suggestions! 

I did have a quick doubt regarding the sequential execution part, instead of strictly avoiding it, what do you think about supporting both async and sequential tasks by allowing different parameters when adding them?

For example, we could treat tasks added using `addTaskBefore` or `addTaskAfter` as part of a sequential execution chain (to maintain order where needed), while tasks added with a regular `addTask` could still run in parallel as async futures. This would give developers flexibility without encouraging unnecessary dependencies.

The rest of the suggestions make a lot of sense; I’ll start implementing those changes accordingly.

SandPod:
> For example, we could treat tasks added using addTaskBefore or addTaskAfter as part of a sequential execution chain (to maintain order where needed), while tasks added with a regular addTask could still run in parallel as async futures. This would give developers flexibility without encouraging unnecessary dependencies.

I like the thought of giving the most amount of flexibility.

To start with, I think it is fine to scope down the problem to the simplest possible solution that achieves our target goals, and then expand the system as required later on. This gives us a better chance to understand the sequential requirements.

My suggestion is:
- Ignore adding the `DatabasePoolManager` to the `ShutdownTaskManager`.
- Always run all tasks asynchronously, enforcing no dependencies between them.
- Add all existing tasks that are awaited in the shutdown process to the `ShutdownTaskManager`.

Just having this simple version of the `TaskManager` could be useful for the start and initialize process of the server as well.

And if syncronous execution is required, we encourage our users to bundle these in a single task.

### Additional requirement

Make it possible to remove a task from the task manager. This makes it possible for users to "expand" their task if required.

yashas-hm:
I've already migrated the shutdown tasks to use the `ShutdownTaskManager`.

One question: since this could also be useful for initializing processes, would it make sense to rename it to something more general like `TaskManager` or `ProcessManager` for simplicity?

Also, just confirming: I'll go ahead and add support for `removeTask` so tasks can be expanded or replaced when needed. Let me know if that sounds good.

SandPod:
> One question: since this could also be useful for initializing processes, would it make sense to rename it to something more general like TaskManager or ProcessManager for simplicity?

Yes, that would be great! I think the class you are building should be general.

>  I'll go ahead and add support for removeTask so tasks can be expanded or replaced when needed. Let me know if that sounds good.

Yes 👍

SandPod (commented):
Overall implementation looks pretty solid!

A couple of comments on how to better separate "user" code from "framework" code.

I would also like to see a full test suite that validates the `task_manager` works as expected.

packages/serverpod/lib/src/server/task.dart
```diff
@@ -0,0 +1,24 @@
...
+  ///
+  /// The [id] is used for identification and logging purposes.
+  /// The [callback] is the function that will be executed when the task is run.
+  Task(this.id, this.callback);
+
+  /// The identifier for this task.
+  ///
+  /// Used for identification and logging purposes, especially when errors
+  /// occur.
+  String id;
```
SandPod: Suggestion: Use `Object` type so that `Symbol` can be used as an identifier.

Symbol documentation: https://dart.dev/language/built-in-types#symbols

packages/serverpod/lib/src/server/task_manager.dart
```diff
@@ -0,0 +1,67 @@
...
+
+  /// Removes a task with the specified ID from the task map.
+  ///
+  /// This method removes the task with the given [id] from the task map.
+  /// Throws an [AssertionError] if no task with the specified ID exists.
+  ///
+  /// The [id] is the identifier of the task to remove.
+  void removeTask(String id) {
+    _tasks.remove(id);
+    throw AssertionError('Task with id: $id, doesn\'t exist');
```
SandPod: Suggestion: This will always throw regardless of if the task was removed or not.

I don't think it is a sound interface to throw an `Error` class exception if the task is not found.
`Error` class exceptions are never meant to be catched by developer code, while accidentally removing the same task multiple times might not be wanted. Perhaps it is better to simply return a bool here as a response to if a task was removed or not?

packages/serverpod/test/util/task_manager_test.dart
```diff
@@ -0,0 +1,141 @@
...
+
+void main() {
+  group('TaskManager', () {
+    late TaskManager taskManager;
+
+    setUp(() {
+      taskManager = TaskManager();
+    });
+
+    group('addTask', () {
```
SandPod: Suggestion(Style): For Serverpod we use the `Given, When, Then` pattern for the tests.

Any tests written should use that format.

packages/serverpod/test/util/task_manager_test.dart
```diff
@@ -0,0 +1,141 @@
+import 'package:serverpod/src/server/task_manager.dart';
+import 'package:serverpod/src/server/task.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('TaskManager', () {
```
SandPod: Suggestion: These tests validate that the task manager works as expected, but do not validate that the task manager is actually used and the tasks configured by it are run on server shutdown.

I think it would be better to validate that the tasks are run by:
1. Constructing a Serverpod instance.
2. Adding (or add and remove) a shutdown task.
3. Start and stop the instance.
4. Validate that the shutdown task was executed.

This will inform us if the interface ever changes, or if we accidentally remove the shutdown tasks from the shutdown code when refactoring the Serverpod class.

And we should also validate when a task throws an exception, that other tasks are still executed.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -153,6 +155,18 @@ class Serverpod {
 
   FutureCallManager? _futureCallManager;
 
+  /// Provides access to the task manager.
+  ///
+  /// The task manager is responsible for executing tasks concurrently.
+  /// In this case, it's used to manage server shutdown tasks, ensuring that all
+  /// resources are properly released and services are stopped.
+  /// You can use this to add custom tasks using [TaskManager.addTask].
```
SandPod: Suggestion: Rename to `shutdownTasks` and use `shutdownTasks.addTask` in the description.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -877,49 +961,33 @@ class Serverpod {
...
-          message: 'Error in serverpod shutdown',
-        );
-      }
-      return e.values;
-    });
 
+    await _externalShutdownTaskManager.handleTasks(
+      onTaskError: (error, stack, id) {
+        shutdownError = error;
+        _reportException(error, stack, message: 'Error in internal task "$id"');
```
SandPod: Suggestion: Update error message to reflect these are external shutdown tasks.

I don't think `external` should be part of the error message since the user only knows about "its" own shutdown tasks.

But the "internal" one should reflect that there was an issue with the internal tasks since these are hidden.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -939,7 +1007,7 @@ class Serverpod {
     }
 
     if (shutdownError != null) {
-      throw shutdownError as Object;
+      throw shutdownError!;
```
SandPod, line 1027: Question(For my knowledge): Why was this change done?

packages/serverpod/test/util/task_manager_test.dart
```diff
@@ -0,0 +1,141 @@
...
+      test('removes a task from the task manager', () {
+        // Arrange
+        var task = Task('test-task', () async => 'result');
+        taskManager.addTask(task);
+
+        // Act & Assert
+        // Note: There seems to be a bug in the removeTask method - it always throws
+        // an AssertionError after removing the task. This test will fail until that
+        // bug is fixed.
+        expect(() => taskManager.removeTask('test-task'),
```
SandPod: Question: This seems like we are validating the wrong behavior, was this AI generated?

SandPod (changes requested):
Last round of comments. I think everything you need to get this over the finish line is included here.

Comment if you have any questions, If you find the tests difficult to build I can help out with that. But make sure that the logic is sound as it now contains an invalid `Error` thrown that is also validated by the test.

packages/serverpod/test/util/task_manager_test.dart
yashas-hm: Yeah. I used a little AI for help and research for test cases as I have not worked much on test cases before will read more on testing to get accurate test behaviors.

packages/serverpod/lib/src/server/serverpod.dart
yashas-hm: To remove unnecessary casting to Object.

packages/serverpod/lib/src/server/task_manager.dart
yashas-hm: Yeah, apologies I was testing something forgot to remove the Error. Will add a return instead.

yashas-hm:
> Last round of comments. I think everything you need to get this over the finish line is included here.
> 
> Comment if you have any questions, If you find the tests difficult to build I can help out with that. But make sure that the logic is sound as it now contains an invalid `Error` thrown that is also validated by the test.

I am reading and learning more about tests and patterns will reach out if I have questions, thank you.

SandPod:
Great, thanks for adding tests. I'll push some last changes to this PR directly and then we can get it in.

I think it would be great to combine this with the new Service locator that is experimental here: https://github.com/serverpod/serverpod/pull/3617

SandPod:
@yashas-hm - Didn't get around to fix the last steps for this today, aiming to do this tomorrow.

One thing we will need to take into consideration is that we probably want to close all servers that receive requests and stop future calls before we run the shutdown tasks as we would otherwise risk receiving more requests before the shutdown tasks are run.

yashas-hm:
> @yashas-hm - Didn't get around to fix the last steps for this today, aiming to do this tomorrow.
> 
> One thing we will need to take into consideration is that we probably want to close all servers that receive requests and stop future calls before we run the shutdown tasks as we would otherwise risk receiving more requests before the shutdown tasks are run.

We can change the order of the server and web server shutdown to be placed before so it will start earlier compared to other tasks or place it outside like database pool manager outside the task manager. That might reduce/resolve the risk of receiving additional requests.

SandPod:
@yashas-hm - I added tests and did some refactoring. Please fix the suggestions from CodeRabbit that are valid and then this should be ready for merge.

SandPod:
This PR is required in order for this to pass the CI: https://github.com/serverpod/serverpod/pull/3687

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -203,6 +207,70 @@ class Serverpod {
...
+      Task(
+        'Internal Session',
+        _internalSession.close,
+      ),
+    );
+
+    _internalServicesShutdownTasks.addTask(
+      Task(
+        'Redis Controller',
+        () async => redisController?.stop,
```
tp:
```suggestion
        () async => redisController?.stop(),
```

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -203,6 +207,70 @@ class Serverpod {
...
+      ),
+    );
+
+    _internalServicesShutdownTasks.addTask(
+      Task(
+        'Health Check Manager',
+        () async => _healthCheckManager?.stop,
```
tp:
```suggestion
        () async => _healthCheckManager?.stop(),
```

packages/serverpod/lib/src/server/serverpod.dart
tp: This is tricky, I recall having run into a similar issue when having a "void" callback.

I don't think there is a linter against it, but basically I suppose a function should always have something like "use this" (unless it's returned as a `Function`, so that there is a reasonable expectation someone else calls it)…

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -203,6 +207,70 @@ class Serverpod {
...
 
+  /// Initializes the servers internal shutdown task managers and registers
+  /// shutdown tasks.
+  ///
+  /// This method is called during server startup and sets up the servers
+  /// internal task managers with all the necessary tasks to properly shutdown
+  /// the server.
+  void _initializeShutdownTaskManagers() {
+    _requestReceivingShutdownTasks.addTask(
+      Task(
```
tp: Nit: Should the outside ever hold on to the `Task` or we we envision these being created with a helper?

If it's more like a config (as we see here), I would probably drop the class from the public interface and just pass the ID and callback.

packages/serverpod/lib/src/server/serverpod.dart
tp: Also, does it make sense to register a task that does not do anything?

I think the existence of the `redisController` is fixed by this point, so we could just not add the task.

Not that it costs much, but might be confusing if the inspector/logger shows it when it does not seem used.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -911,49 +982,45 @@ class Serverpod {
...
-          _reportException(
-            error.error,
-            error.stackTrace,
-            message: 'Error in server shutdown',
-          );
-        }
-      } else {
+
+    await _requestReceivingShutdownTasks.executeTasks(
+      onTaskError: (error, stack, id) {
```
tp, line 981: Nit: As it's all internal here (and there is really just the logger as a sensible implementation) I don't mind, but I find the "extra info as callbacks" pattern really unintuitive (there's also some in the legacy auth module).

Why not return a list of errors / overall status instead?

packages/serverpod/lib/src/server/tasks/task_manager.dart
```diff
@@ -0,0 +1,73 @@
+import 'dart:async';
+
+import 'package:meta/meta.dart';
+
+import 'task.dart';
+
+/// Manages tasks that need to be executed.
+///
+/// The `TaskManager` allows registering tasks that can be executed
+/// concurrently. Each task is identified by a unique ID and contains
```
tp, line 8: The implementation and documentation here is super clear, so I guess it just changed from the PR description:

> Listeners are executed sequentially and awaited properly


If I recall correctly, that was initially planned (also with Task dependencies, right?).

So we might just need to update the description then.

Personally I would've probably started out with sequential execution, as I would imagine the tasks being rather short and at least then the developer has a "guarantee" that it would work the next time as well.

On the other hand, if they do have dependent shutdown operations, they can combine them in a bigger task.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -1122,13 +1189,26 @@ Future<void>? _shutdownTestAuditor() {
...
+  final TaskManagerImpl _shutdownTasks;
+
+  /// Shutdown tasks can be used to perform cleanup operations before the server
+  /// is shut down. The tasks will be executed asynchronously after the server
+  /// has received the shutdown signal.
+  ///
+  /// You can use this to add custom tasks using [shutdownTasks.addTask].
+  ///
+  /// Before the shutdown tasks are executed, the server will close the api
+  /// server, web server, service server, and future call manager.
```
tp:
```suggestion
  /// server, web server, insights server, and future call manager.
```

packages/serverpod/lib/src/server/tasks/task.dart
```diff
@@ -0,0 +1,23 @@
+/// A task that can be executed by the [TaskManager].
+///
+/// Tasks are managed by the [TaskManager] and can be executed at various points,
+/// such as during server shutdown. Tasks can be executed in parallel or
+/// sequentially based on their configuration.
```
tp:
```suggestion
/// such as during server shutdown.
```

I think, as there is nothing in the `Task` that would configure this and for now we only have 1 task manager that also does not handle any configuration.

packages/serverpod/lib/src/server/tasks/task_manager.dart
```diff
@@ -0,0 +1,73 @@
...
+/// a callback function that will be executed when the task is run.
+abstract interface class TaskManager {
+  /// Adds a task to be executed.
+  ///
+  /// The task is stored in the task map with its ID as the key.
+  /// The [task] contains an ID for identification and a callback function
+  /// that will be executed when the task is run.
+  ///
+  /// Throws [StateError] if a task with the same ID already exists.
```
tp, line 16: Was at first surprised that an `add*` would throw on duplicate (imagining a list underneath as my mental model for the task list).

But good with the exception, then at least no developer would overwrite a task by accident, and they would be steered toward `removeTask` + `addTask` in case they wanted to update one.

packages/serverpod/lib/src/server/tasks/task.dart
```diff
@@ -0,0 +1,23 @@
...
+  ///
+  /// The [id] is used for identification and logging purposes.
+  /// The [callback] is the function that will be executed when the task is run.
+  Task(this.id, this.callback);
+
+  /// The identifier for this task.
+  ///
+  /// Used for identification and logging purposes, especially when errors
+  /// occur.
+  final Object id;
```
tp: Given that the `id` has this "uniqueness" property, I wonder if a second `debugLabel` is warranted?

But might be fine without it, in practice I also don't see much removal of tasks, so a more descriptive ID might be just fine :)

tp (approved):
Apart from the Redis/HealthCheckManager `stop` calls this all looks fine to me.

The rest is just a bunch of bike-shedding, so pick what you feel is useful and ignore the rest 😉

tests/serverpod_test_server/test_integration/shutdown_task_manager_test.dart
```diff
@@ -0,0 +1,184 @@
...
+
+    await serverpod.shutdown(exitProcess: false);
+
+    expect(called, isFalse);
+  });
+
+  test(
+      'Given shutdown task that records number of calls '
+      'when the server is shutdown multiple times '
+      'then the task once for each shutdown.', () async {
```
tp:
```suggestion
      'then the task is executed once for each shutdown.', () async {
```

tests/serverpod_test_server/test_integration/shutdown_task_manager_test.dart
```diff
@@ -0,0 +1,184 @@
...
+    await serverpod.shutdown(exitProcess: false);
+
+    expect(completer1.isCompleted, isTrue);
+    expect(completer2.isCompleted, isTrue);
+  });
+
+  test(
+      'Given a registered shutdown task '
+      'when the server is shutdown '
+      'then the task is executed after all request receiving services are shutdown',
```
tp:
```suggestion
      'then the task is executed after all request receiving services are shutdown.',
```

packages/serverpod/lib/src/server/serverpod.dart
SandPod: Makes sense to hide the "Task" as an implementation detail.

packages/serverpod/lib/src/server/tasks/task_manager.dart
SandPod: I think this will need to be covered when thinking about services. So we could solve that problem there... I have some ideas on how we could solve both these issues in one go.

packages/serverpod/lib/src/server/serverpod.dart
SandPod: Great catch, second time I do the same mistake in the same PR...

Since we are reading this as a member of the class, we would need to store a local copy of the field in order to remove the `?`. So I'm not sure there is an elegant way of retrieving it and using it directly. You can do it in an if statment but that is basically just hiding the local copy.

SandPod: Fixed here: [fix(review): Actually call the stop methods.. (again)](https://github.com/serverpod/serverpod/pull/3555/commits/b70d5986e1919b6075866b5f2eadc77bfbc7cd19)

packages/serverpod/lib/src/server/serverpod.dart
SandPod: [fix(review): Actually call the stop methods.. (again)](https://github.com/serverpod/serverpod/pull/3555/commits/b70d5986e1919b6075866b5f2eadc77bfbc7cd19)

packages/serverpod/lib/src/server/serverpod.dart
SandPod: [refactor: Make tasks an implementation detail of the task manager.](https://github.com/serverpod/serverpod/pull/3555/commits/586835cc34ac5e6ef544afd3104aa91fd78a10b9)

packages/serverpod/lib/src/server/serverpod.dart
SandPod: Using a return statement removes the "liveness" report. But I'll consider this for the follow up service fixes.

SandPod (approved):

PR #3560, opened 2025-05-20
feat: Move `Session.transactionOrSavepoint` method to `serverpod` package

tp:
Follow-up from https://github.com/serverpod/serverpod/pull/3441


packages/serverpod/lib/src/server/session.dart
```diff
@@ -278,6 +278,52 @@ abstract class Session implements DatabaseAccessor {
...
+      return result;
+    } catch (_) {
+      try {
+        await savepoint?.rollback();
+      } catch (e, stackTrace) {
+        log(
+          'Failed to roll back to savepoint.',
+          exception: e,
+          stackTrace: stackTrace,
+        );
```
SandPod, line 321: Suggestion: Do we actually need a separate log for this? IF this fails it would be the same as if the user created a savepoint and attempted to rollback right?

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
```diff
@@ -0,0 +1,581 @@
...
+        () async {
+      await session.transactionOrSavepoint((final transaction1) async {
+        await SimpleData.db.insertRow(
+          session,
+          SimpleData(num: 1),
+          transaction: transaction1,
+        );
+
+        await session.transactionOrSavepoint((final transaction2) async {
+          expect(identical(transaction1, transaction2), isTrue);
```
SandPod, line 107: Suggestion: Break this expect out to a separate test to give better context around it.

`When transaction is passed in` - The nesting here is just redundant complexity, we would test the same thing by creating a transaction normally and then passing it in.

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
```diff
@@ -0,0 +1,581 @@
...
+
+      expect(
+        await SimpleData.db.count(session),
+        1,
+      );
+    });
+
+    test(
+        'when throwing in a nested transaction callback, '
+        'then the the data written with the transaction is not visible on the outside.',
```
SandPod, line 174: Comment: Reading this I can understand your concern with the name. We aren't really expressing the intent with this helper method.

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
```diff
@@ -0,0 +1,581 @@
...
+        where: (_) => Constant.bool(true),
+      );
+    });
+
+    test(
+        'when inspecting the session, '
+        'then it does not have an implicit transaction attached to it.',
+        () async {
+      // ignore: invalid_use_of_visible_for_testing_member
+      expect(session.transaction, isNull);
```
SandPod, line 28: Suggestion: Convert this to an "assert" in the setup.

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
```diff
@@ -0,0 +1,581 @@
...
+      } on _ForcedTestException catch (_) {}
+
+      expect(
+        await SimpleData.db.count(session),
+        1,
+      );
+    });
+  });
+
+  withServerpod(
```
SandPod, line 207: Question: Do we really need to test this with the different options available in the test framework? Does the behavior differ in any way?

SandPod: If we want to do that, perhaps we should just have a different "session" setup and then run the exact same testsuite for all the different configurations?

I would expect that we want the same behavior to be preserved if we change anything.

SandPod (commented):
Nice helper, a couple of comments so that it is polished for the public.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -278,6 +278,52 @@ abstract class Session implements DatabaseAccessor {
...
+  /// In case [f] errs, it will restore the database connection to the
+  /// state before [f] was invoked, either by discard its transaction
+  /// or rolling back to the previously created savepoint.
+  ///
+  /// In case the parent [transaction] is also used in parallel outside of
+  /// [f], all modifications done in [f] will be visible there as well.
+  /// In this case also restoring the savepoint in case of error would
+  /// undo any changes made in parallel. So ideally the passed [transaction]
+  /// should not be used elsewhere until this methods completes.
+  Future<R> transactionOrSavepoint<R>(
```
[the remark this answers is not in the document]
tp: I tried that, but then in practice (where we always get the `Session session` as the first parameter), the code would look like:

```dart
session.db.transactionOrSavepoint(session, (final transaction) { … }, transaction: transaction);
```

As I expect most cases in practice to only get the `db` from the `session`, it seems weird then to have to pass it in again right away.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -278,6 +278,52 @@ abstract class Session implements DatabaseAccessor {
...
   }
+
+  /// Runs the closure [f] in isolation and undoing any changes in case it fails.
+  ///
+  /// This either reused the parent [transaction], or creates one itself.
+  /// In case a [transaction] is given, this will create an internal savepoint.
+  ///
+  /// In case [f] errs, it will restore the database connection to the
+  /// state before [f] was invoked, either by discard its transaction
+  /// or rolling back to the previously created savepoint.
```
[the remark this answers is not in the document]
tp: > We simply return the transaction.

In case the provided closure throws, the whole transaction is discarded / rolled back, and thus changes made will not be visible to the outside.

So for the caller this behaves the same. In case the inner function fails, they will not see any changes (either in the DB itself or in their transaction, if they passed one in.).

(The only exception being the "savepoint not restored one" 😬)

packages/serverpod/lib/src/server/session.dart
```diff
@@ -278,6 +278,52 @@ abstract class Session implements DatabaseAccessor {
...
+  /// [f], all modifications done in [f] will be visible there as well.
+  /// In this case also restoring the savepoint in case of error would
+  /// undo any changes made in parallel. So ideally the passed [transaction]
+  /// should not be used elsewhere until this methods completes.
+  Future<R> transactionOrSavepoint<R>(
+    final TransactionFunction<R> f, {
+    required final Transaction? transaction,
+  }) async {
+    if (transaction == null) {
+      return db.transaction(f);
```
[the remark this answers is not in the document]
tp: Do you mean skipping the creation of the transaction, and only use a savepoint instead?

If I understand that correctly, we might just factor out the second part of this method (handling a savepoint), and anyone not wanting to have a transaction fallback can just use that one.

packages/serverpod/lib/src/server/session.dart
tp: TBD, maybe we can make do with a special exception.

This is really the worst case, as all guarantees are off at this point (was the DB/transaction modified? 🤷🏻‍♂️), so maybe a log is not enough (as the caller can not know that this happened in addition to their own code failing).

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
```diff
@@ -0,0 +1,581 @@
...
+        1,
+      );
+    });
+
+    test(
+        'when creating two independent transactions with `transactionOrSavepoint`, '
+        'then the data is visible only to each transaction that wrote it.',
+        () async {
+      await session.transactionOrSavepoint((final transaction1) async {
+        await session.transactionOrSavepoint((final transaction2) async {
```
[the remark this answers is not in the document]
tp: You're correct, these are 2 independent transaction which should run / be open in parallel.

I thought this way offered the best readability, as otherwise I would have to create one, but keep it open with a `Completer` or similar, and then create the other one afterwards and see that I can not yet read from the first one etc.

If you want I can refactor this into that style (or a third option if you have one).

Basically the test wants to assure that the helper really gives us 2 distinct transactions.

Maybe a `transaction1 != transaction2` would already be enough, but I feel that leans a bit too much into the implementation, because who knows, maybe the wrapper `class` could be different, but in the end they behave the same because some internal reference is equal and thus the same transaction on the database level gets used?

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
tp: But we can not hold onto the transaction outside of the closure, can we?

So instead of nesting, I would have to "give it out", but keep the closure alive by awaiting some `Future`, right?

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
tp: As the test show we can not really test "the same things" with `Rollback.disabled` and the other options, due to the implicit use of the test transaction in the rollback-enable case.

For the `Rollback.after*` cases we could iterate through the 2 options and run the whole suite for each of them.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -278,6 +278,52 @@ abstract class Session implements DatabaseAccessor {
...
+      final result = await f(transaction);
+
+      await savepoint.release();
+
+      return result;
+    } catch (_) {
+      try {
+        await savepoint?.rollback();
+      } catch (e, stackTrace) {
+        log(
```
tp, line 317: Decision: Throw a specialized exception like `SavepointRollbackFailed(object innerException)`.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -278,6 +278,52 @@ abstract class Session implements DatabaseAccessor {
       stackTrace: stackTrace,
     );
   }
+
+  /// Runs the closure [f] in isolation and undoing any changes in case it fails.
+  ///
+  /// This either reused the parent [transaction], or creates one itself.
```
tp, line 284: Split up the transaction / savepoint variants on this level, and then document the possible errors for each case.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -278,6 +278,52 @@ abstract class Session implements DatabaseAccessor {
...
+  ///
+  /// In case the parent [transaction] is also used in parallel outside of
+  /// [f], all modifications done in [f] will be visible there as well.
+  /// In this case also restoring the savepoint in case of error would
+  /// undo any changes made in parallel. So ideally the passed [transaction]
+  /// should not be used elsewhere until this methods completes.
+  Future<R> transactionOrSavepoint<R>(
+    final TransactionFunction<R> f, {
+    required final Transaction? transaction,
+  }) async {
```
tp, line 299: Move to `Database`, as an `extension`, exported on the package level by default.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -278,6 +278,52 @@ abstract class Session implements DatabaseAccessor {
...
+  /// or rolling back to the previously created savepoint.
+  ///
+  /// In case the parent [transaction] is also used in parallel outside of
+  /// [f], all modifications done in [f] will be visible there as well.
+  /// In this case also restoring the savepoint in case of error would
+  /// undo any changes made in parallel. So ideally the passed [transaction]
+  /// should not be used elsewhere until this methods completes.
+  Future<R> transactionOrSavepoint<R>(
+    final TransactionFunction<R> f, {
+    required final Transaction? transaction,
```
tp, line 298: For completeness sake should take `TransactionSettings`.

packages/serverpod/lib/src/server/session.dart
tp: Decision: Log will be removed, explict exception.

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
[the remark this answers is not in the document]
tp: Decision: Keep this test pattern, as we need the parallel execution.

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
tp: Decision: Move the `identical` check into it's own test.

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
tp: Decision: Probably the best name for now.

packages/serverpod/lib/src/server/session.dart
tp: Update: Export as a helper function (not extension method), like `runIntransactionOrSavepoint(Database, F, Transaction?, TransactionSettings?)`.

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
```diff
@@ -0,0 +1,581 @@
...
+          );
+        } on _ForcedTestException catch (_) {}
+
+        expect(
+          await SimpleData.db.count(session),
+          1,
+        );
+      });
+    },
+    rollbackDatabase: RollbackDatabase.disabled,
```
tp, line 577: Use [TestTags.concurrencyOneTestTag],


as in: tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_isolatation_level_test.dart

packages/serverpod/lib/src/server/session.dart
tp: @SandPod Helper method is now called

```
  static Future<R> runInTransactionOrSavepoint<R>(
    final Database database,
    final TransactionFunction<R> f, {
    /// Settings to be applied in case a new transaction is created.
    final TransactionSettings? settings,
  })
```

With the transaction being a positional parameter, as it'll always be set and readable in a production context (only some tests will pass a literal `null`).

packages/serverpod/lib/src/server/session.dart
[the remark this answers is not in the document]
tp: Settings added.

packages/serverpod/lib/src/server/session.dart
tp: Specific exception will be thrown now.

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
tp: Done.

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
tp: Added the tag as suggested, @SandPod. Interestingly so far CI was already passing all the time. Likely all the others that would've interfered excluded themselves nicely so far, that one additional test file didn't make a difference. But then this is solid going forward :)

packages/serverpod/lib/src/util/database_util.dart
```diff
@@ -0,0 +1,81 @@
...
+  /// - In case no [transaction] was given (`null`), the method [f] is run in a new transaction.
+  ///
+  ///   If [f] should error, the internally created transaction is cancelled.
+  ///
+  ///   Code outside of [f] accessing the database, which does not have access to the internal
+  ///   transaction, can not see any changes until this method finishes.
+  ///
+  ///   Once this method is finished, all changes are persisted to the database.
+  ///
+  /// - In case an active [transaction] was given, this method creates a [savepoint](https://www.postgresql.org/docs/current/sql-savepoint.html)
```
tp, line 19: This is quite a mouthful.

For our auth cases this is probably fine.

For external users I wonder whether we should ensure (on the lower-level methods already?) that there is not parallel savepoint ongoing, so that there is only 1 thing to be undone (unless the user would explicitly opt into "unsafe" nested / parallel savepoints).

tests/serverpod_test_server/test_integration/database_operations/transactions/transaction_or_rollback_test.dart
tp: Split out.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -278,6 +278,52 @@ abstract class Session implements DatabaseAccessor {
...
     );
   }
+
+  /// Runs the closure [f] in isolation and undoing any changes in case it fails.
+  ///
+  /// This either reused the parent [transaction], or creates one itself.
+  /// In case a [transaction] is given, this will create an internal savepoint.
+  ///
+  /// In case [f] errs, it will restore the database connection to the
+  /// state before [f] was invoked, either by discard its transaction
```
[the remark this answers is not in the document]
tp: rephrased entire comment.

packages/serverpod/lib/src/server/session.dart
tp: Update the comment accordingly.

packages/serverpod/lib/src/server/session.dart
tp: added `settings` parameter.

packages/serverpod/lib/serverpod.dart
```diff
@@ -36,3 +36,6 @@ export 'src/server/experimental_features.dart';
 export 'diagnostic_events.dart';
 
 export 'package:meta/meta.dart' show useResult;
+
+// Utils
+export 'package:serverpod/src/util/database_util.dart';
```
SandPod, line 41: Suggestion: Add this to the `database.dart` package and export it from there.

packages/serverpod/lib/src/util/database_util.dart
tp: To be pondered after more experience using the helper.

PR #3566, opened 2025-05-21
3510 pretty print debug log output

SandPod:

## Pre-launch Checklist

## Breaking changes


SandPod (commented):
Some help notes.

packages/serverpod/lib/src/server/features.dart
```diff
@@ -39,9 +39,9 @@ class Features {
 
   /// Returns true if the web server is enabled.
   static bool get enablePersistentLogging =>
-      _instance._config.sessionLogs?.persistentEnabled == true;
+      _instance._config.sessionLogs.persistentEnabled == true;
```
tp, line 42:
```suggestion
      _instance._config.sessionLogs.persistentEnabled;
```

Since there is no `null` in the chain anymore.

packages/serverpod/lib/src/server/features.dart
```diff
@@ -39,9 +39,9 @@ class Features {
...
 
   /// Returns true if the web server is enabled.
   static bool get enableConsoleLogging =>
-      _instance._config.sessionLogs?.consoleEnabled == true;
+      _instance._config.sessionLogs.consoleEnabled == true;
```
tp, line 46:
```suggestion
      _instance._config.sessionLogs.consoleEnabled;
```

packages/serverpod/lib/src/server/log_manager/log_writers.dart
```diff
@@ -226,6 +226,106 @@ class StdOutLogWriter extends LogWriter {
...
+class TextStdOutLogWriter extends LogWriter {
+  final int _logId;
+  final Session _session;
+
+  TextStdOutLogWriter(this._session) : _logId = _session.sessionId.hashCode;
+
+  @override
+  Future<void> logEntry(LogEntry entry) async {
+    var message =
+        '[LOG] ${entry.logLevel.name.toUpperCase()} | session=$_logId | message=${entry.message}';
```
tp, line 239: What a little bit confused by this initially.

I wonder if we could give a clearer name here (e.g. what created the session)?

Or maybe locally a simple counter would be preferably, as we probably on have <100 requests between restarts, and then there are no huge numbers to keep track off.

packages/serverpod/lib/src/server/log_manager/log_writers.dart
```diff
@@ -226,6 +226,106 @@ class StdOutLogWriter extends LogWriter {
...
+
+    if (entry.error != null ||
```
tp, line 241: Extract to `isError` helper to share with JSON?

packages/serverpod/lib/src/server/session.dart
```diff
@@ -186,7 +186,12 @@ abstract class Session implements DatabaseAccessor {
     }
 
     if (Features.enableConsoleLogging) {
-      logWriters.add(StdOutLogWriter(session));
+      var logFormat = session.serverpod.config.sessionLogs.consoleLogFormat;
+      var consoleLogger = switch (logFormat) {
```
tp, line 190:
```suggestion
      final logFormat = session.serverpod.config.sessionLogs.consoleLogFormat;
      final consoleLogger = switch (logFormat) {
```

Even though we don't have that linter here yet? We might just slowly transition (else I would think personally it's hard to figure out which style to use where).

templates/serverpod_templates/projectname_server_upgrade/config/development.yaml
```diff
@@ -48,6 +48,7 @@ maxRequestSize: 524288 # The maximum size of requests allowed in bytes
 sessionLogs:
   persistentEnabled: true
   consoleEnabled: true
+  consoleLogFormat: text # Defaults to "text", options are "text" or "json"
```
tp, line 51: Comment out to align with production/staging example?

tp (approved):
LGTM!

templates/serverpod_templates/projectname_server_upgrade/config/development.yaml
SandPod: This is intentially added so that the `text` format is used in development.

packages/serverpod/lib/src/server/features.dart
SandPod: Good catch! Fixed in: [72e2dc9](https://github.com/serverpod/serverpod/pull/3566/commits/72e2dc9072f86f53b875988e80543b1af81ca1e7)

templates/serverpod_templates/projectname_server_upgrade/config/development.yaml
SandPod: You are right, the comment here is not correct. If this line is commented out (and a configuration file exists) this will default to json. Just like the other cases. Will update it!

packages/serverpod/lib/src/server/log_manager/log_writers.dart
SandPod: Discussed and resolved in slack. No change to the code.

Issue #2962, opened 2024-11-14
Support custom environment variables in passwords interface.


Isakdl:
**Proposal**
Any environment variable prefixed with `SECRET_` or `secret_` should be injected into the passwords map with the prefix stripped. 

usage:
```sh
export secret_myApiKey=myRandomKey
```

```dart
session.serverpod.getPassword('myApiKey')
```

muhammedrashidm:
is this added?

SandPod:
Would suggest calling the prefix `SEVERPOD_PASSWORD_` and then verbatim use what is after the underline to make it predictable for the user to match the `passwords.yaml` file.

PR #3569, opened 2025-05-22
feat(auth2): Add module `serverpod_auth_jwt`

### LINKED ISSUE #3526: Add JWT-based session package
Create a module `serverpod_auth_session_jwt` support JWT-based sessions, with database backed refresh tokens.

The authentication providers should then support either of the 2 "official" approaches (DB-backed session or JWT sessions) out of the box, with a simple configuration.

The developer needs to ensure to use the same approach across all authentication providers and the frameworks authentication handler.

tp:
Implements JWT-based sessions.

Each such session is backed by a refresh token which is stored in the database.
During normal requests the server validates the signature of the access token against the configured private key, and then returns the user ID and scopes contained in the token.

Upon token rotation, the old refresh token in the database is replaced by a new one and the client receives the new refresh token and a new access token.
There old access token will still continue to work, if it has not expired yet. (In practice clients would, of course, take care to only rotate their tokens once their access token is expired.)

Closes #3526.


modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/pubspec.yaml
```diff
@@ -0,0 +1,36 @@
...
+name: serverpod_auth_jwt_server
+description: Serverpod JWT authentication server module.
+version: 2.7.0
+repository: https://github.com/serverpod/serverpod
+
+environment:
+  sdk: '>=3.3.0 <4.0.0'
+
+dependencies:
+  dart_jsonwebtoken: ^2.17.0
```
tp, line 13: We can not upgrade to the 3.x version, as that requires at least Dart 3.4.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
```diff
@@ -0,0 +1,26 @@
...
+
+/// Secrets used for authentication tokens.
+abstract class AuthenticationTokenSecrets {
+  /// The configuration key for the private key to sign access tokens with.
+  static String privateKeyConfigurationKey = 'serverpod_auth_jwt.privateKey';
+
+  /// The private key used for signing access tokens.
+  static String get privateKey {
+    final privateKey = privateKeyTestOverride ??
+        Serverpod.instance.getPassword(privateKeyConfigurationKey);
```
SandPod, line 12: Comment: We really need this issue to be resolved for this to be viable: https://github.com/serverpod/serverpod/issues/2962

This will be resolved by the new dev starting in June.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+
+/// Business logic for handling JWT-based access and refresh tokens.
+abstract class AuthenticationTokens {
+  /// Looks up the `AuthenticationInfo` belonging to the [jwtAccessToken].
+  ///
+  /// In case the session token looks like a JWT, but is not valid a debug-level
+  /// log entry is written.
+  ///
+  /// Returns `null` in any case where not valid authentication could be derived from the input.
+  static Future<AuthenticationInfo?> authenticationHandler(
```
SandPod, line 19: Question: Since this is used in a different section of the users code, should we break this out of the scope? Maybe have it in a different one? Or is it fine to keep them all grouped together?

My thought is that the user might do `AuthenticationTokens.` to see what is available in the business logic, where this does not really belong in the business logic layer.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+    try {
+      final result = _parseAccessToken(jwtAccessToken);
+
+      return AuthenticationInfo(
+        result.authUserId,
+        result.scopes,
+        authId: result.refreshTokenId.toString(),
+      );
+    } catch (e) {
+      if (e is! JWTUndefinedException) {
```
SandPod, line 32: Suggestion: Break this out into a `on JWTUndefinedException` section.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+        session.log(
+          'Invalid JWT access token',
+          level: LogLevel.debug,
+        );
```
SandPod, line 36: Question: Why are we only logging this and ignoring the undefined exceptions? What are we expecting to happen here? Shouldn't we rethrow in this case? as this seems to be some unknown exception?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+  /// In case the session token looks like a JWT, but is not valid a debug-level
+  /// log entry is written.
+  ///
+  /// Returns `null` in any case where not valid authentication could be derived from the input.
+  static Future<AuthenticationInfo?> authenticationHandler(
+    final Session session,
+    final String jwtAccessToken,
+  ) async {
+    try {
+      final result = _parseAccessToken(jwtAccessToken);
```
SandPod, line 24: Question: Shouldn't we guard this by only checking if we are the issuer?

Won't we create a bunch of log messages when we try to do this with for example the service key that we are planning to support as a separate issuer?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+      session,
+      id,
+      transaction: transaction,
+    );
+
+    if (refreshTokenRow == null) {
+      throw RefreshTokenNotFoundException();
+    }
+
+    if (refreshTokenRow.secret != refreshTokenSecret) {
```
SandPod, line 95: Question: Shouldn't the secret be a hashed value of something? It seems like we are simply parsing it from the refreshToken received meaning we are sending it in clear text.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+      await RefreshToken.db.deleteRow(session, refreshTokenRow);
```
SandPod, line 96: Question: The transaction is not used here, is this deliberate?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+
+      throw RefreshTokenInvalidSecretException();
+    }
+
+    if (refreshTokenRow.isExpired) {
```
SandPod, line 101: Question: If it is experied, shouldn't it also be deleted as in the previous case?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+      accessToken: _generateAccessToken(refreshToken: refreshTokenRow),
+    );
+  }
+
+  /// Removes all refresh tokens for the given [authUserId].
+  ///
+  /// Active access tokens will still continue to work until the expiry time is reached.
+  static Future<void> destroyAllRefreshTokens(
+    final Session session, {
+    required final UuidValue authUserId,
```
SandPod, line 124: Question: This interface should take in a transaction given we support the `onCommit` hook right?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+    );
+  }
+
+  /// Removes a specific refresh token.
+  ///
+  /// This does not affect the user's other authentications.
+  ///
+  /// Any access tokens associated with this refresh token will continue to work
+  /// until they expire.
+  static Future<void> destroyRefreshToken(
```
SandPod, line 146: Question: This interface should also take in a transaction right?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+    return generateRandomString(64);
+  }
+
+  /// Creates a new access token for the given user.
+  ///
+  /// For now all tokens are signed with the symmetric "HMAC with SHA-512" algorithm.
+  ///
+  /// The auth user ID is set as `subject` and the refresh token ID for which this access token is generated is set as `jwtId`.
+  ///
+  /// NOTE: The tokens generated this way are not suitable to be given to 3rd party (untrusted) system, as they can not validate them.
```
SandPod, line 178: Question: How do we plan to support public/private key encryption? It seems like a cornerstone feature for JWT to be able to validate the tokens that are created by the server externally.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+  ///       To validate the token one needs access to the key used for signing (since this is currently using a symmetric algorithm),
+  ///       which is not to be given out, as otherwise these system could create new tokens themselves.
+  ///
+  ///       Futhermore since the [refreshTokenId] is encoded in the access token, any attacker getting hold of an access token could
+  ///       use this to invalidate the refresh token by making a false request with the refresh token ID and any invalid secret to the
+  ///       [rotateRefreshToken] method.
```
SandPod, line 184: Suggestion: Couldn't we hash the secret(maybe rename to salt) with the uuid with an internal secret in the server to validate that the token is valid? Seems bad to allow users to invalidate each others token with just the authId.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+  ///
+  ///       Thus if the access token are ever desired to be used with such external system, they would need to switch to a public/private
+  ///       asymmetric signing scheme and a layer of indirection would have to be added between the access token and its refresh tokens.
+  static String _generateAccessToken({
+    required final RefreshToken refreshToken,
+  }) {
+    final jwt = JWT(
+      {
+        'scopeNames': refreshToken.scopeNames.toList(),
```
SandPod, line 193: Suggestion: Add support for additional custom user data.

A common request is to be able to assign additional data when creating the JWT token. So that it is not necessary to go to the database with information that should be easily avialble. Is there any reason for not supporting this?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+      },
+      jwtId: refreshToken.id!.toString(),
+      subject: refreshToken.authUserId.toString(),
+      issuer: _jwtTokenIssuer,
+    );
+
+    return jwt.sign(
+      SecretKey(AuthenticationTokenSecrets.privateKey),
+      expiresIn: AuthenticationTokenConfig.current.defaultAccessTokenLifetime,
+      algorithm: JWTAlgorithm.HS512,
```
SandPod, line 203: Question: Why not use public/private key for signing?

SandPod (changes requested):
Nitty review since this is really the core of the authentication logic.

I did not look at the tests yet since there was so many comments and questions around the code.

But I expect this code to be well tested to ensure that we have not introduced any errors in our token issuer or validation logic.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/pubspec.yaml
SandPod: It is also a year since 3.4 was release so that would be an OK minimum version to go with now as well.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/models/refresh_token.spy.yaml
```diff
@@ -0,0 +1,23 @@
...
+serverOnly: true
+table: serverpod_auth_jwt_refresh_token
+fields:
+  id: UuidValue?, defaultPersist=random
+
+  ### The [AuthUser] this profile belongs to.
+  authUser: module:auth_user:AuthUser?, relation(onDelete=Cascade)
+
+  ### The scopes given to this session.
+  scopeNames: Set<String>
```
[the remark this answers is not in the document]
tp: Yes. As we also store them on the DB-backed session, it would surely make sense to do any major changes to `Scope`s now and start with the "new" approach in these modules :)

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
tp: I suppose we could take in a `Map<String, dynamic>` and then let the developer tack on additional data.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
tp: > How do we plan to support public/private key encryption? 

I thought we should start out with HMAC / symmetric by default, as that is super easy to set up and practically the same for everyone that has just the Serverpod server and only wants to avoid hitting the DB but does not need to validate tokens externally.

So then support public-key cryptography I thought we just need a new config field for the algorithm (defaults to HMAC, and use public/private keys), and one for the public key.

(In both cases here and others we might want to change the storage to support a list (practically maxed at 2) to support rotation. Same for peppers etc.)

Definitely something I want to support, just didn't think this is suitable as the baseline as the developer just getting started with this will have to create and handle the keys, adding a lot more steps. (But yes, the migration between algorithms would be bumpy, so maybe we should force them into the other pattern?)

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/generated/refresh_token.dart
```diff
@@ -0,0 +1,604 @@
...
+  })  : lastUpdated = lastUpdated ?? DateTime.now(),
+        created = created ?? DateTime.now();
+
+  factory RefreshToken({
+    _i1.UuidValue? id,
+    required _i1.UuidValue authUserId,
+    _i2.AuthUser? authUser,
+    required Set<String> scopeNames,
+    required String fixedSecret,
+    required (String, String) variableSecret,
```
tp, line 39: I should fix the code generation to re-expose the names of the positional fields (for documentation purposes).

Then we can retain the small storage and don't have to use a full named record or have to resort to `String`-building ourselves.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/pubspec.yaml
tp: As discussed, let's do the overall Dart version upgrade separately, and then we can bump this one as well in the package.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/models/refresh_token.spy.yaml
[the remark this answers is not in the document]
tp: As dicussed, moving to a `Scope` serializable model entails too much prep work (like `const`, `==`, positional parameters), so we won't block the new auth on this.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_config.dart
```diff
@@ -0,0 +1,32 @@
+/// Configuration options for the JWT authentication module.
+class AuthenticationTokenConfig {
+  /// The default lifetime of access tokens.
+  ///
+  /// This will be encoded in each access token, and for
+  /// incoming requests only the token's value will be used.
+  ///
+  /// Defaults to 10 minutes.
+  final Duration defaultAccessTokenLifetime;
```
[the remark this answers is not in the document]
tp: Renamed to `accessTokenLifetime` for now. With the comment it should still be clear what is happening for anyone changing this.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
tp: Split this up into 2 `on`s now.

I want to ignore the general / unknown JWT errors and all such things that are not even JWTs, but want to log any specific errors (e.g. expired, invalid signature, etc.) so that the developer can then investigate those (as those cases are either an invalid attempt with another token, or a misbehaving client, or a misconfigured server (e.g. changing the keys without configuring the rotation properly)).

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
tp: > Shouldn't we guard this by only checking if we are the issuer?

The issuer is now configurable and will be checked when unpacking the JWT.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
tp: As discused, added public/private key options as well now.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,289 @@
...
+    };
+
+    return (
+      refreshTokenId: refreshTokenId,
+      authUserId: authUserId,
+      scopes: scopes,
+    );
+  }
+
+  static const _jwtTokenIssuer =
```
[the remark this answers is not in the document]
tp: It is now configurable, defaulting to `null`.

Incoming tokens will be checked that this matches the current configuration.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
tp: Added support for the developer to add additional claims.

The only question remaining would be whether it's okay for us to "squat" `scopeNames` by default (hopefully a 3rd party system would not care about those or worse expect that value in another format), or whether we should rename them to e.g. `dev.serverpod.scopes` to prevent collisions as per https://datatracker.ietf.org/doc/html/rfc7519#page-10.

util/update_pubspecs
```diff
@@ -58,6 +58,8 @@ cp CHANGELOG.md modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth
 cp CHANGELOG.md modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_client/CHANGELOG.md
 cp CHANGELOG.md modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_client/CHANGELOG.md
 cp CHANGELOG.md modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/CHANGELOG.md
+cp CHANGELOG.md modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_client/CHANGELOG.md
```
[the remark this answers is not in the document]
tp: Yes, we should really do that as a follow-up.

I can look into this right away, hopefully all those script are (or at least can be) aligned for most of the subfolders and won't need many exceptions.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
tp: Decision:
- Move the `scopeNames` to `dev.serverpod.scopeNames`

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
tp: Admin "namespace" has been added and the configuration moved.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,263 @@
...
+        stackTrace: stackTrace,
+      );
+
+      return null;
+    } catch (e) {
+      return null;
+    }
+  }
+
+  /// Creates a new token pair for a given auth user.
```
tp, line 56:
```suggestion
  /// Creates a new token pair for the given auth user.
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,263 @@
...
+  /// This is akin to creating a new session, and should be used after a successful login or registration.
+  static Future<TokenPair> createTokens(
+    final Session session, {
+    required final UuidValue authUserId,
+    required final Set<Scope> scopes,
+
+    /// Extra claims to be added to the JWT.
+    ///
+    /// These are added on the top level of the paylaod, so be sure not to conflict with the [registered claims](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1),
+    /// as those will just overwrite any custom claims given here.
```
tp, line 67:
```suggestion
    /// as those will always overwrite any custom claims given here.
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,263 @@
...
+    ///
+    /// Since these claims will be embedded in every access token (also across rotations), one should be mindful about there length and total size.
```
tp, line 69:
```suggestion
    /// These claims will be embedded in every access token (also across rotations) and then sent along with any request. This should be taken into account with regard to the total size of the added claims.
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,263 @@
...
+    return TokenPair(
+      refreshToken: RefreshTokenString.buildRefreshTokenString(
+        refreshToken: refreshToken,
+        rotatingSecret: secret,
+      ),
+      accessToken: JwtUtil.createJwt(refreshToken),
+    );
+  }
+
+  /// Returns a new refresh & access token pair.
```
tp, line 98:
```suggestion
  /// Returns a new refresh / access token pair.
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,263 @@
...
+  ///
+  /// This invalidates the previous refresh token.
```
tp, line 100:
```suggestion
  /// This invalidates the previous refresh token.
  /// Previously created access tokens for this refresh token will continue to work until they expire.
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,263 @@
...
+      refreshTokenData.id,
+      transaction: transaction,
+    );
+
+    if (refreshTokenRow == null ||
+        !uint8ListAreEqual(
+          Uint8List.sublistView(refreshTokenRow.fixedSecret),
+          refreshTokenData.fixedSecret,
+        )) {
+      throw RefreshTokenNotFoundException();
```
tp, line 134:
```suggestion
      // If the `fixedSecret` does not match, we do not delete the refresh token.
      // (Since the `id` is encoded on every access token and thus might be known to 3rd parties or leak, even after the access token itself has expired.)
      throw RefreshTokenNotFoundException();
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,263 @@
...
+        refreshToken: refreshTokenRow,
+        rotatingSecret: newSecret,
+      ),
+      accessToken: JwtUtil.createJwt(refreshTokenRow),
+    );
+  }
+
+  /// Removes all refresh tokens for the given [authUserId].
+  ///
+  /// Active access tokens will still continue to work until the expiry time is reached.
```
tp, line 185:
```suggestion
  /// Active access tokens will continue to work until their expiration time is reached.
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/models/refresh_token.spy.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+  ### The scopes given to this session.
+  ###
+  ### These will also be added to each access token as a claim named "scopeNames".
+  scopeNames: Set<String>
+
+  ### Extra claims to be added to each access token created for this refresh token.
+  ###
+  ### This is a `Map<String, dynamic>` where each entry uses the key as the claim name.
+  ### The value must be convertible to JSON.
```
tp, line 18: Rename (locally + gen)

```suggestion
  ### The values must be JSON-encodable.
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/models/refresh_token.spy.yaml
```diff
@@ -0,0 +1,58 @@
...
+  ###
+  ### Users must ensure that the claims don't conflict with [registerd claims](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1)
+  ### or the above mentioned claim "scopeNames".
```
tp, line 21: Rename (locally + gen)

```suggestion
  ### or the above-mentioned claim for [scopeNames].
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/models/refresh_token.spy.yaml
```diff
@@ -0,0 +1,58 @@
...
+  ###
+  ### This is only stored as a serialized String in the database due to schema limitations.
+  extraClaims: String?
+
+  ### The fixed part of the secret.
+  ###
+  ### Any incoming rotation request referencing refresh token by ID and having the correct fixed part,
+  ### but not the correct `secret`, will cause the refresh token to be invalidated (as the refresh token
+  ### may have been leaked at that point).
+  ### The the pure `id` is also part of the JWT access token for reference, we have to have this second
```
tp, line 31:
```suggestion
  ### Since the refresh token's `id` is also part of the JWT access tokens for reference, we have to have this second
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/models/refresh_token.spy.yaml
```diff
@@ -0,0 +1,58 @@
...
+  ### part in here, ensuring that no-one with just a captured JWT can invalidate the refresh token.
```
tp, line 32:
```suggestion
  ### part in here, ensuring that no one with just a (potentially expired) JWT can invalidate the refresh token.
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/models/token_pair.spy.yaml
```diff
@@ -0,0 +1,11 @@
+### A pair of refresh and access tokens, in their external format.
+class: TokenPair
+fields:
+  ### The refresh token to be used to create new access tokens and roate itself.
```
tp, line 4:
```suggestion
  ### The refresh token to be used to create new access tokens and rotate itself.
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/models/token_pair.spy.yaml
```diff
@@ -0,0 +1,11 @@
...
+  ###
+  ### The client should handle this as an opaque string.
+  ## This is stored as `$refreshTokenId:$refreshToken`
```
tp, line 7:
```suggestion
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/util/equal_uint8list.dart
```diff
@@ -0,0 +1,16 @@
+import 'dart:typed_data';
+
+/// Checks whethere the 2 given lists contain the same data.
```
tp, line 3:
```suggestion
/// Checks whether the 2 given lists contain the same data.
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/util/random_bytes.dart
```diff
@@ -0,0 +1,11 @@
+import 'dart:math';
+import 'dart:typed_data';
+
+final Random _random = Random.secure();
+
+/// Generates a secure random =bytes of the specified length.
```
tp, line 6:
```suggestion
/// Generates a list of secure random bytes of the specified length.
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_token_secrets_test.dart
```diff
@@ -0,0 +1,104 @@
...
+      () {
+        expect(
+          () => AuthenticationTokenSecrets.algorithm,
+          throwsArgumentError,
+        );
+      },
+    );
+  });
+
+  withServerpod('Given just both a private and public key for ES512,',
```
tp, line 73:
```suggestion
  withServerpod('Given both a private and public key for ES512,',
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
```diff
@@ -0,0 +1,362 @@
...
+    setUp(() async {
+      session = sessionBuilder.build();
+    });
+
+    tearDownAll(() {
+      AuthenticationTokenSecrets.privateKeyTestOverride = null;
+    });
+
+    test(
+        'when calling the authentication handler with an non-JWT String, then it returns `null``.',
```
tp, line 31:
```suggestion
        'when calling the authentication handler with an non-JWT String, then it returns `null`.',
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
```diff
@@ -0,0 +1,362 @@
...
+        'when calling the authentication handler, then the extra claims are available on the auth info.',
+        () async {
+      final authInfo = await AuthenticationTokens.authenticationHandler(
+        session,
+        tokenPair.accessToken,
+      );
+
+      expect(authInfo?.extraClaims, equals({'string': 'foo', 'int': 1}));
+    });
+
```
tp, line 196: Test authentication handler with `withClock` for expired date.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/jwt_util_test.dart
```diff
@@ -0,0 +1,381 @@
...
+          tokenData.extraClaims,
+          isEmpty,
+        );
+      });
+
+      group('when the HMAC key is changed,', () {
+        setUp(() {
+          AuthenticationTokenSecrets.privateKeyTestOverride = 'another key';
+
+          AuthenticationTokenSecrets.fallbackAlgorithmTestOverride =
```
tp, line 123:
```suggestion
          // Only needed to not request the configuration from Serverpod, which does not work outside of `withServerpod` in tests
          AuthenticationTokenSecrets.fallbackAlgorithmTestOverride =
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_info_from_jwt.dart
```diff
@@ -0,0 +1,36 @@
...
+  }
+}
+
+/// Extensions for [AuthenticationInfo] when used with module `serverpod_auth_jwt`.
+extension AuthenticationInfoJwt on AuthenticationInfo {
+  /// Get the extra claims associated with this authentication info.
+  ///
+  /// Assumes that it was created by the `serverpod_auth_jwt` module,
+  /// else it'll throw because no extra claims exist.
+  Map<String, dynamic> get extraClaims {
```
SandPod: Suggestion: Return `null` or empty map if it doesn't exist.

Mindfull about when this is used when multiple authentication handlers are active. For example if a service key authentication handler is used as well.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/refresh_token_secret_hash.dart
```diff
@@ -0,0 +1,48 @@
...
+/// Uses the Argon2id algorithm.
+/// See: https://en.wikipedia.org/wiki/Argon2
+@internal
+abstract final class RefreshTokenSecretHash {
+  /// Create the hash for the given refresh token secret.
+  static ({Uint8List hash, Uint8List salt}) createHash({
+    required final Uint8List secret,
+    @protected Uint8List? salt,
+  }) {
+    salt ??= generateRandomBytes(
```
tp, line 30: TODO: as in email authentication PR make async and put the computation in an isolate

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
```diff
@@ -0,0 +1,196 @@
+import 'package:meta/meta.dart';
+import 'package:serverpod/serverpod.dart';
+
+/// Secrets used for authentication tokens.
+@internal
+abstract class AuthenticationTokenSecrets {
+  static const String algorithmConfigurationKey =
+      'serverpod_auth_jwt.algorithm';
```
SandPod, line 25: Comment: Would make sense if these could be defined like so in the passwords file:

```yaml
serverpod_auth_jwt:
  algoritm: 'blabla'
  privateKey: 'blabla'
```

Should we create an issue for that?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
```diff
@@ -0,0 +1,196 @@
...
+  /// Which algorithm to use for the fallback token verification (during key rotation).
+  static const String fallbackAlgorithmConfigurationKey =
+      'serverpod_auth_jwt.fallbackAlgorithm';
+
+  /// The key to be used for the fallback key verification (during token rotation).
+  ///
+  /// For the `HS512` algorithm, this would be the single secret.
+  /// For the `ES512` algorithm, this would be the public key.
+  static const String fallbackKeyConfigurationKey =
+      'serverpod_auth_jwt.fallbackKey';
```
SandPod: Suggestion: For symmetry make this `fallbackPublicKey`.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
```diff
@@ -0,0 +1,196 @@
...
+
+    if (privateKey == null || privateKey.isEmpty) {
+      throw ArgumentError(
+        'No valid private key was set',
+        privateKeyConfigurationKey,
+      );
+    }
+
+    switch (algorithm) {
+      case null || algorithmHS512:
```
SandPod: Suggestion(NIT): Create an enum that is parsed from the string.

This ensures that this switch statement is always in line with possible enum values.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
```diff
@@ -0,0 +1,196 @@
...
+
+    final privateKey = privateKeyTestOverride ??
+        Serverpod.instance.getPassword(privateKeyConfigurationKey);
+
+    final publicKey = publicKeyTestOverride ??
+        Serverpod.instance.getPassword(publicKeyConfigurationKey);
+
+    if (privateKey == null || privateKey.isEmpty) {
+      throw ArgumentError(
+        'No valid private key was set',
```
SandPod: Suggestion: Provide error messages here that help the developer correct the issue.

Link to the key that needs to be assigned.
Same for the other configuration options.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
```diff
@@ -0,0 +1,196 @@
...
+    }
+  }
+
+  /// The pepper used for hashing refresh tokens.
+  static String get refreshTokenHashPepper {
+    final pepper = refreshTokenHashPepperTestOverride ??
+        Serverpod.instance.getPassword(refreshTokenHashPepperConfigurationKey);
+
+    if (pepper == null || pepper.isEmpty) {
+      throw Exception(
```
SandPod: Question: Shouldn't this also be an argument error? Or is the developer supposed to catch this?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
```diff
@@ -0,0 +1,196 @@
+import 'package:meta/meta.dart';
+import 'package:serverpod/serverpod.dart';
+
+/// Secrets used for authentication tokens.
+@internal
+abstract class AuthenticationTokenSecrets {
```
SandPod: Suggestion: Couldn't this be a singleton on the `authentication_tokens` class as well? That would allow us to reassign this through a non exported extention that is only available in the test framework.

That would allow us to remove all the override stuff.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens_admin.dart
```diff
@@ -0,0 +1,26 @@
+import 'package:meta/meta.dart';
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_jwt_server/serverpod_auth_jwt_server.dart';
+import 'package:serverpod_auth_jwt_server/src/generated/refresh_token.dart';
+
+/// Collection of admin functions for managing authentication tokens.
+@internal
```
SandPod: Suggestion: Remove internal label.

This admin function is should be used by developers as well right?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
```diff
@@ -0,0 +1,140 @@
...
+        if (refreshToken.scopeNames.isNotEmpty)
+          _serverpodScopeNamesClaimKey: refreshToken.scopeNames.toList(),
+      },
+      jwtId: refreshToken.id!.toString(),
+      subject: refreshToken.authUserId.toString(),
+      issuer: _issuer,
+    );
+
+    final (JWTKey key, JWTAlgorithm algorithm) =
+        switch (AuthenticationTokenSecrets.algorithm) {
```
SandPod: Question: If I understand this correctly, this is "re-parsed" from the passwords file on each call for get. Could you add some caching here so that we don't need to reinstantiate that object for each invocation of algorithm?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
```diff
@@ -0,0 +1,140 @@
...
+      jwt = JWT.verify(
+        accessToken,
+        key,
+        issuer: _issuer,
+      );
+    } catch (_) {
+      final fallbackAlgorithm =
+          AuthenticationTokenSecrets.fallbackVerificationAlgorithm;
+      if (fallbackAlgorithm == null) {
+        rethrow;
```
SandPod: Suggestion: Move everything after this rethrow out from the `catch` statement.

We can use this rethrow as an early return and reduce our branch level.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
```diff
@@ -0,0 +1,140 @@
...
+
+      jwt = JWT.verify(
+        accessToken,
+        key,
+        issuer: _issuer,
+      );
+    }
+
+    final refreshTokenId = UuidValue.fromString(jwt.jwtId!);
+    final authUserId = UuidValue.fromString(jwt.subject!);
```
SandPod: Question: Should we catch if these `dangerous` dereffs fail? That would mean that our implementation for creating the JWT token has deviated from the verification of it right?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
```diff
@@ -0,0 +1,140 @@
...
+    final scopeNames = scopeNamesClaim != null
+        ? (scopeNamesClaim as List).cast<String>()
+        : const <String>[];
+
+    final scopes = {
+      for (final scopeName in scopeNames) Scope(scopeName),
+    };
+
+    final allClaims = (jwt.payload as Map).cast<String, dynamic>();
+    final extraClaims = Map.fromEntries(
```
SandPod, line 119: Question: any reason for not including the serverpod claims here?
Is it possible for the user to overwrite serverpod clains using the extra claims method? Of so, we might want to include all claims in the response?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,266 @@
...
+  }) async {
+    final auths = await RefreshToken.db.deleteWhere(
+      session,
+      where: (final row) => row.authUserId.equals(authUserId),
+      transaction: transaction,
+    );
+
+    if (auths.isEmpty) return;
+
+    // Notify clients about the revoked authentication for the user
```
SandPod: Suggestion(NIT): This does not notify clients, it only notifies internally in the server and any other server that might be connected through the messages interface.

SandPod: Same documentation fix should be applied in the next method.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_token_secrets_test.dart
```diff
@@ -0,0 +1,104 @@
...
+    setUpAll(() {
+      AuthenticationTokenSecrets.privateKeyTestOverride = 'secret-key-for-jwt';
+    });
+
+    test(
+      'when the current `algorithm` is read, then it uses HS512.',
+      () {
+        final algorithm = AuthenticationTokenSecrets.algorithm;
+
+        expect(algorithm, isA<HmacSha512AuthenticationTokenAlgorithm>());
```
SandPod: Suggestion: Validate that it is `having` the private key.

Just as you did for the other algorithm.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_token_secrets_test.dart
```diff
@@ -0,0 +1,104 @@
...
+          throwsArgumentError,
+        );
+      },
+    );
+  });
+
+  withServerpod('Given just a private key,',
+      (final sessionBuilder, final endpoints) {
+    setUpAll(() {
+      AuthenticationTokenSecrets.privateKeyTestOverride = 'secret-key-for-jwt';
```
SandPod: Suggestion: Add test for valdating when the `HS512` algorithm is explicitly specified.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
```diff
@@ -0,0 +1,376 @@
...
+import 'package:serverpod_auth_jwt_server/serverpod_auth_jwt_server.dart';
+import 'package:serverpod_auth_jwt_server/src/business/authentication_token_secrets.dart';
+import 'package:serverpod_auth_jwt_server/src/business/jwt_util.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+import 'package:test/test.dart';
+
+import 'test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod('Given an empty session,',
```
SandPod: Suggestion: Please change this to something that is related to the test.

For example: `Given jtw authentication handler`

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
```diff
@@ -0,0 +1,376 @@
...
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_jwt_server/serverpod_auth_jwt_server.dart';
+import 'package:serverpod_auth_jwt_server/src/business/authentication_token_secrets.dart';
+import 'package:serverpod_auth_jwt_server/src/business/jwt_util.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+import 'package:test/test.dart';
+
+import 'test_tools/serverpod_test_tools.dart';
+
+void main() {
```
SandPod, line 12: Suggestion: Break the authentication handler tests out to a separate file.

This helps me understand what has been tested in the public interface.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
```diff
@@ -0,0 +1,376 @@
...
+
+      authUserId = authUser.id!;
+    });
+
+    tearDownAll(() {
+      AuthenticationTokenSecrets.privateKeyTestOverride = null;
+    });
+
+    test('when requesting a new token pair, then one is returned.', () async {
+      await AuthenticationTokens.createTokens(
```
SandPod: Suggestion: This test does not have an expect.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
```diff
@@ -0,0 +1,376 @@
...
+      await expectLater(
+        () => AuthenticationTokens.rotateRefreshToken(
+          session,
+          refreshToken: tokenWithUpdatedFixedSecret,
+        ),
+        throwsA(isA<RefreshTokenNotFoundException>()),
+      );
+    });
+
+    test(
```
SandPod, line 229: Question: Did you add a test for validating that all refresh tokens are revoked if an old refresh token is sent in again?

The scenario that we discussed.

SandPod (commented):
Getting close!

Didn't get to the last `jwt_util_test.dart` file. Will take a look at that one after my meeting. But there are already a couple of comments and questions.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_info_from_jwt.dart
tp: Yeah, I thought there are good arguments in either case. If you only use this one, having to use `!` is also not nice.

But always a tricky decision to decide what the practical threshold is (which case is more common, and which can be guarded against).

Will change this to `?`, as likely the caller is using yet another extension method anyway to get typed access to the values, and then they can fail only for missing data in case they require it.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp: That would be awesome, and I think the conversion from nested to `x.y` key is very natural :)

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp: But it is not the public key (in the case of symmetric encryption).

Hence just "key" + the comment. (So in both cases you can not just use the config, but actually have to copy the right value into the desired place.)

Maybe `fallbackVerificationKey` though, so make clear that this is not used for further signing?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp: Sure, I can change that.

I also pondered whether to make the whole "xAlgorithm" (class instance) overwriteable, but then thought there is some value in setting the raw keys and assert the desired behavior.

But we can also test the parsing separately, and then use higher level mocks here.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens_admin.dart
tp: My idea was to use this as a warning to not export this class.

Rather we use the `AuthenticationTokens` class as the namespace, and all access should go through that instead of this class being directly exported or passed around in client code.

Given that information, do you still want to remove the annotation and/or export this class itself?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
tp: Sure, I can add some extra check to throw a custom error that would immediately pin point the missing field without having to relate the exception to the source code line (to find out which field was null).

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
tp: Okay. I just wanted to express that there are no preconditions whatsoever.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
tp: Yeah, it should just not throw, as otherwise this would only be called in _setups_ of the other tests. Then we can also remove this outright.

tp: removed

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
tp: Test added for the refresh token re-use specifically.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
tp: rephrased, but still referring that these are the "error" / non-existent cases.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
tp: It probably means "clients" as in "listeners", right?

The same comment is already in the original `serverpod_auth` and `serverpod_auth_session`.

Will update all 3 of those with a new phrasing then.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_token_secrets_test.dart
tp: HMAC secret is now also asserted in the tests.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_token_secrets_test.dart
tp: test added.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
tp: > any reason for not including the serverpod claims here?

The idea is that they only get the `extraClaims` they put themselves in.
(Ideally, unless they re-use a registered claim, in which case that data is being overwritten as per the documentation and the "new" value will also not be in the map (which would probably be super confusing anyway)).

So right now the API (if used reasonably) has a nice symmetry. (Personally, I would even be open to throw in case a registered claim is given in the extra claims.)

> Is it possible for the user to overwrite serverpod clains using the extra claims method? 

No. First we take the extra claims as the basis for the payload, then add the Serverpod scope claim, and then the registered claims for the JWT are written on top.

So in case the developer would reuse our specific scope claim, it would be (silently) overwritten.

But there is also no reason why they should do so (for now), because they can just pass any `Scope`s they like (and then at least rely on the encoding being correct such that we can parse it).

> Of so, we might want to include all claims in the response?

I would keep any low-level details out unless there emerges a good need. Of course we could expose the time the token was created or the remaining lifetime etc., but is that really useful? Shouldn't the library be configured (and configurable) in a way that handles this for the developer?
(E.g. if they wanted a short TTL for some users, that should probably be a flag to be passed on the refresh token, instead them checking afterwards that a special condition applies etc.)

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
tp: added verbose error messages for the cases where we parse our own stuff back.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
tp: Just moving it out does not feel right (because then just reading the outer level it would be surprising why we're using the fallback, and ideally we wouldn't assign twice 2 the `jwt` variable (though Dart doesn't get that it's impossible here, so it can not be `final`).

Also it'd be cool to make use of the check in line 72 where we check that the fallback algorithm is non-`null`, so I wouldn't want to re-read and `!` it outside the `catch`.

So for now I moved it into a new function where we can then use `return` clearly for the control flow. Let me know if you have an idea to improve that further.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp: missing peppers are now argument errors (also for the session package).

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp: added the enum.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp: > Link to the key that needs to be assigned.

We already mention the configuration option to be set as the argument's "name", and usually do not repeat this in the error message then. Should I change that?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp: Create the issue: https://github.com/serverpod/serverpod/issues/3600

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp: I wonder whether having a default "algorithm" is really worth it after all (now that we support 2 out of the gate).

Maybe it'd be better to require the user to specify the one to use (doesn't really matter if they copy 1 more line from the docs I suppose), and then we could also use "perfect" names, like `secret` for HMAC and `private/publicKey` for Ecdsa in all places.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp: We can put this (`@protected`) on the `AuthenticationTokens` class, with a lazy init.

Then we can overwrite it in tests without having to mock the low-level keys.

The instance would cache the configuration (and tests would just overwrite it, but that would not be something possible at runtime in production).

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
tp: See above, we will make the secrets an instantiated class which holds on to the values.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
tp: This has been moved out. Indeed a bit easier to read if you're looking for that specifically, though we had to re-do the whole setup in each case / file then.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
```diff
@@ -0,0 +1,229 @@
...
+  const HmacSha512AuthenticationTokenAlgorithmConfiguration({
+    required this.key,
+  });
+}
+
+/// Elliptic Curve Digital Signature Algorithm
+@internal
+class EcdsaSha512AuthenticationTokenAlgorithmConfiguration
+    implements AuthenticationTokenAlgorithmConfiguration {
+  final String privateKey;
```
tp: Here we should be using the `EcPrivateKey` / `EcPublicKey` classes from the JWT library, so this would fail on first access when misconfigured (which a production server could trigger on start), and not only on usage of each key.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp: Decision:
- Make `algorithm` required
- Rename to `serverpod_auth_jwt.fallback.algorithm` / `serverpod_auth_jwt.fallback.secretKey/publicKey`
- Use specific fields for the keys (`secretKey`, `privateKey`, `publicKey`)

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp:
```suggestion
        'No valid private key was set in the passwords',
```

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens_admin.dart
tp: Decision:
- Export the type
- Mark only the constructor as `@internal` as there is no use in creating another instance (on top of `AuthenticationTokens.admin`)

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
tp: Decision:
- Throw when using one of the registered or Serverpod claims.
- Throw for any starting with `serverpod.dev.*`

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
tp:
```suggestion
    // Notify all other server instances about the revoked authentication for the user
```

tp: Or delete this comment, as the behavior is pretty clear.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp: This has been implemented.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp: Errors have been changed to refer to "in the passwords". The configuration key is given as the `ArgumentError`'s `name`.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens_admin.dart
tp: `AuthenticationTokensAdmin` is now exported.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
tp: Claims are now checked when creating the JWT.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
tp: This has been removed in all 3 affected packages.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_token_secrets.dart
tp: The secrets configuration has been changed to now be a (cached) instance on `AuthenticationTokens`, which can be overwritten there (fully typed, no more strings) for testing.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/utils/authentication_token_secrets_mock.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:dart_jsonwebtoken/dart_jsonwebtoken.dart';
+import 'package:serverpod_auth_jwt_server/src/business/authentication_token_secrets.dart';
+
+class AuthenticationTokenSecretsMock implements AuthenticationTokenSecrets {
```
tp, line 4: Going forward we might want to make this a full-blown builder, but as the uses are very limited (right now and going forward), I skipped that for now.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,280 @@
...
+  /// The secrets configuration.
+  static final __secrets = AuthenticationTokenSecrets();
+
+  /// Secrets to the used for testing. Also affects the internally used [JwtUtil] and [RefreshTokenSecretHash]
+  @visibleForTesting
+  static AuthenticationTokenSecrets? secretsTestOverride;
+  static AuthenticationTokenSecrets get _secrets =>
+      secretsTestOverride ?? __secrets;
+
+  static JwtUtil get _jwtUtil => JwtUtil(secrets: _secrets);
```
tp, line 260: This does not read ideal, but we need to be sure to use the latest `_secrets`.
(Else we would have to provide more complicated hooks for testings / clearing this.)

As the classes are just simple wrappers (formal `abstract/static`) this seems fine in practice to recreate. Alternatively we could also make them `static` again and then pass the secrets, but I quite like that now at least all usages of this a configured to the same secrets, and calling their methods is nice and succinct.

This pattern should also be able to evolve well if we had dependency injection, in which case they could be `final` / created once (as we could then create new isolated instances of `AuthenticationTokens` in tests).

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -0,0 +1,280 @@
...
+      AuthenticationTokens.config.refreshTokenRotatingSecretLength,
+    );
+  }
+
+  /// The secrets configuration.
+  static final __secrets = AuthenticationTokenSecrets();
+
+  /// Secrets to the used for testing. Also affects the internally used [JwtUtil] and [RefreshTokenSecretHash]
+  @visibleForTesting
+  static AuthenticationTokenSecrets? secretsTestOverride;
```
SandPod, line 256: Suggestion: Make this hidden and use a non exported extention to access the field.

This is now exported along with the interface. Let's try to keep it as slim as possible and only expose what the developer should use.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
```diff
@@ -0,0 +1,204 @@
...
+  ///
+  /// Throws in case of any validation failures (e.g. invalid signature or changed issuer, etc.) and in any case when parsing the expected contained data fails.
+  VerifiedJwtData verifyJwt(final String accessToken) {
+    final jwt = _verifyJwt(accessToken);
+
+    final UuidValue refreshTokenId;
+    try {
+      refreshTokenId = UuidValue.fromString(jwt.jwtId!);
+    } catch (e) {
+      throw ArgumentError(
```
SandPod, line 84: Comment: Note that these argument errors are catched and ignored in the calling method in authenticationHandler.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_token_secrets_test.dart
SandPod: I can't find this test added?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
SandPod: It would be fine to expect it not to throw in that case, or assert it returns something.

SandPod: I think it is valid to have such a test.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
tp: Reworked this into an `extension` / helper class.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
tp: Note added :)

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
tp: re added

PR #3571, opened 2025-05-22
feat: Add ability register function to be run after a `Transaction` has been committed

### LINKED ISSUE #3557: Support "post commit hook" on `Transaction`
For methods that have side-effects beyond database modification, we would like to have a "post commit" hook where code can be schedule to be executed after the transaction has been committed to the database.

Current examples for this would be:
- Sending out PIN codes via email: These should only be sent once they are stored in the database.
  - Since the code might be running in a transaction, we want to hold the send out until we're sure that it's completed and has not been rolled back by some higher level code.
- Deleting files from the blob storage (S3): Similarly, we only want to delete the actual files in the storage, after we're certain that there corresponding database entries are deleted.

Overall this helps orchestrate actions across distinct distributed systems.

## API proposal

```dart
await session.db.transaction((transaction) async {
  transaction.addPostCommit(() {
      return session.storage.deleteFile(key: 'abc123');
  });
});

```

The method would look like this:

```dart
void addPostCommit(FutureOr<void> Function());
```

## Behavior

Handlers registered with `addPostCommit` would run in first-in first-out fashion, meaning in the order in which they were registered.

Each handler is `await`-ed, so that a subsequent handler could see / act upon a side-effect from a previous one.

Since the commit already happened (and can not be undone at this point), each handler is called in the specified sequence. If any handler `throw`s, that error is caught and logged to the original `Session`, and then it proceeds to the next handler.

Any errors caught will not be visible in the `await session.db.transaction` result. That will only return the result of the database operation.

## Alternatives considered

Another approach which was discussed was adding tasks to be executed after the transaction is committed to a database-backed job queue, like [`future call`s](https://docs.serverpod.dev/concepts/scheduling).

That way we could guarantee that the scheduling of the work succeeds and it only becomes visible to the outside once the transaction successfully completed.

This would also easily support retry mechanism (by keeping the entry in the database until it's completed), and would not rely on the same server instance to stay available until all jobs are completed.

The major drawback of this approach would be a much more complicated setup, as the developer would have to bridge back from the job queue entries to the handler function in there code. Plus logging in such a case would probably happen outside of the originating `Session`, so errors could not be correlated as easily.

tp:
Closes #3557

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles.dart
@@ -197,21 +197,23 @@ abstract final class UserProfiles {
           transaction: transaction,
         );
 
-        for (final image in images) {
-          try {
-            await session.storage.deleteFile(
-              storageId: image.storageId,
-              path: image.path,
-            );
-          } catch (e, stackTrace) {
-            session.log(
-              'Failed to delete user image from storage',
-              level: LogLevel.error,
-              exception: e,
-              stackTrace: stackTrace,
-            );
+        transaction.addPostCommitListener(() async {
+          for (final image in images) {
+            try {
+              await session.storage.deleteFile(
+                storageId: image.storageId,
+                path: image.path,
+              );
+            } catch (e, stackTrace) {
+              session.log(
+                'Failed to delete user image from storage',
+                level: LogLevel.error,
+                exception: e,
+                stackTrace: stackTrace,
+              );
+            }
           }
-        }
+        });
       },
     );
   }


packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -566,12 +566,17 @@ class DatabaseConnection {
...
-      (ctx) {
+      (ctx) async {
         var transaction = _PostgresTransaction(
           ctx,
           session,
         );
-        return transactionFunction(transaction);
+        final result = await transactionFunction(transaction);
+
+        unawaited(
```
tp, line 576: The async behavior will be a bit tricky in tests, but that should be fine in practice.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -566,12 +566,17 @@ class DatabaseConnection {
...
+            runPostCommitListeners(session, transaction._postCommitListeners));
```
tp, line 577: Maybe we need to capture the functions to be run here and then first return the closure, so that the updated database view is always available to the post commit listeners (though for now we would not expect any code in them to touch the DB right away; that would be better done inside the transaction).

PR #3577, opened 2025-05-23
feat: Make endpoints marked with `@doNotGenerate` available in `withServerpod` tests

tp:
But do not expose them to the client.


tests/serverpod_test_module/serverpod_test_module_server/lib/src/generated/endpoints.dart
```diff
@@ -10,35 +10,129 @@
...
 import 'package:serverpod_test_module_server/src/generated/module_class.dart'
-    as _i5;
+    as _i8;
 
 class Endpoints extends _i1.EndpointDispatch {
   @override
   void initializeEndpoints(_i1.Server server) {
     var endpoints = <String, _i1.Endpoint>{
-      'module': _i2.ModuleEndpoint()
+      'ignoredModule': _i2.IgnoredModuleEndpoint()
```
tp, line 28: @SandPod I basically got this working (have to clean up some more tests & the PR history, but locally the integration test for the module now passes).

In order for the endpoint to be usable with `withServerpod` we have to reference it in here though.

Is that fine from your point of view, or should we split this into "real" and "test" endpoint implementations?

(If so, this might become might larger, and I wonder if we couldn't get rid of a lot of the `String`s in here, as they made errors only visible once the code executed.)

While not client code is generated, I fear that having the endpoint here might make it available for the server anyway (if one builds the correct URL/request), which would of course defeat part of the annotation.

tp: (Also I encountered another error where E2E and `withServerpod` integration test were different in there serialization (or rather emulation of it). So maybe it'd really be worth to handle this under a bigger umbrella to get the serialization pipeline aligned as well.)

SandPod: Decision: We will not work on this right now.

Should we close this PR but add a reference to it in the issue?

PR #3579, opened 2025-05-23
feat(auth2): Add `serverpod_new_auth_test_server` for auth module tests

tp:
Adds a test server which combines the auth modules in a real-world scenario.

Contains integration tests (for the `@doNotGenerate` endpoints which are instantiated in here), as well as E2E tests exercising the client.


tp:
Superseeded by https://github.com/serverpod/serverpod/pull/3626

PR #3585, opened 2025-05-27
chore: Bumps the version to match the rest.

SandPod:
An accidentally invalid merge to main broke the CI suite in main. This PR fixes this by ensuring that all lint packages have the same version.

** Additional fix **
Adds a newly introduced test package to the `pub_get_all` script.

## Pre-launch Checklist

## Breaking changes


nielsenko (approved):
Ok.. Maybe we should just yield to the greate yaml formatter god and accept "

SandPod:
@nielsenko 
> Ok.. Maybe we should just yield to the greate yaml formatter god and accept "

If I understand `yaml` correctly, `'` is used for literal strings and `"` is used when escape sequences are needed (`\n`). I don't think we should have a conflict in the version so we should be OK to use `"` if that is what the formatter god demands 😅

PR #3594, opened 2025-05-30
chore: Bump minimum version to Dart 3.5.0 / Flutter 3.24.0

tp:
Had to use at least 3.5.4 due to the issue mentioned in https://github.com/dart-lang/sdk/issues/59560 (discovered with Dart 3.5.3).

But even then `analyze` would fail in CI (but pass locally, even with a cleaned pubspec.lock, did not really figure out what was running there due to the abstractions in the scripts), so I had to remove that failing lint from the `serverpod` package (or at least that worked, not sure if there's an alternative path – but since we won't be running the serverpod core in JS for a while, that should be totally fine either way).


.github/workflows/dart-tests.yaml
```diff
@@ -72,7 +72,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        flutter-version: ['3.19.0', '3.29.0']
+        flutter-version: ['3.24.5', '3.29.0']
```
nielsenko, line 75:
```suggestion
        flutter-version: ['3.24.5', '3.32.1']
```
perhaps? or `latest`?

nielsenko: Only mentioning it here, but we should be consistent if we do upgrade.

packages/serverpod/lib/serverpod.dart
```diff
@@ -2,6 +2,7 @@
 // Until then it needs to cause the analyzer to report an issue,
 // in order to avoid a bug in the Dart analyzer.
 // See: https://github.com/serverpod/serverpod/pull/2900
+// ignore: unnecessary_library_name
 library serverpod;
```
nielsenko, line 6: Why not just delete the line?

nielsenko (commented):
Have a few suggestions. Implement if you agree. 

I think 3.5.4 is just 6 months old. Perhaps @SandPod should weigh in on this.

nielsenko (approved):

packages/serverpod/lib/serverpod.dart
SandPod: Hmm, but if we ignore it then it doesn't server its purpose anymore according to the comment right?

SandPod (commented):
Wonder if we can't get 3.5.0 working. To me, it doesn't seem like the fix for the bug has been resolved until 3.6.0, so anything that works in 3.5.5 should work in 3.5.0 right?

packages/serverpod/lib/serverpod.dart
tp: Yeah, maybe this `library warning is a red-herring, because at least in the latest version it ran the linter that this was not encouraged, but still failed for the JS type checking.

I'll see if we can also remove it.

tp: I now removed the name (keeping just `library;` as we have in all other packages).

I think the lint warning / comment is a bit of a red-herring, as I could never this to error, but the general `invalid_runtime_check_with_js_interop_types` failed (no matter how this line looked).

This got a bit more complicated as locally this passed just fine with and without `invalid_runtime_check_with_js_interop_types` being ignored, but in CI (with an "even cleaner" slate it seems) that JS check would still run into the stack overflow.

But overall I think disabling a rule we don't need is preferable here.

.github/workflows/dart-tests.yaml
tp: Decision:
- Agreed that we should update the maximum version to the latest Flutter stable release, [3.32.1](https://github.com/flutter/flutter/releases/tag/3.32.1)

SandPod (approved):
LGTM! 👍

Issue #3611, opened 2025-06-04
`create` command does not replace `projectname` in Docker Compose test services for module template projects


SandPod:
### Describe the bug

When creating a new module project using the `create` command with the `--template module` option, the generated `docker-compose.yaml` contains unreplaced `projectname` placeholders in service and volume names.

### To reproduce

1. Run `serverpod create serverpod_auth_user --template module`
2. Open `serverpod_auth_user/docker/test/docker-compose.yaml`
3. Observe the following contents:

```yaml
  POSTGRES_DB: projectname_test
  volumes:
    - projectname_test_data:/var/lib/postgresql/data
...
volumes:
  projectname_test_data:
```

These should reflect the actual project name (`serverpod_auth_user`), not `projectname`.

### Expected behavior

The `projectname` placeholder should be replaced consistently throughout the Docker Compose files, including service names, database names, and volume names.

### Library version

Serverpod CLI: 2.8.0

### Platform information

MacOs

### How experienced are you with this library?

None

### Are you interested in working on a PR for this?

- [ ] I want to work on this

PR #3614, opened 2025-06-05
fix: Fix project name replacement when creating modules.

SandPod:
Closes: #3611 

The config templates for modules incorrectly used `projectname_test` instead of `modulename_test` which is what our search and replace algorithm looks for when copying over the template files.

This made the database and data volume of all modules to resolve to the same name.

I also patched this in all existing modules in the project.

## Pre-launch Checklist

- [x ] Any breaking changes are documented below.

## Breaking changes

None - This will only influence new modules created.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/config/test.yaml
+++ b/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/config/test.yaml
@@ -32,7 +32,7 @@ webServer:
 database:
   host: localhost
   port: 9090
-  name: projectname_test
+  name: serverpod_auth_jwt_test
   user: postgres
 
 # This is the setup for your Redis test instance.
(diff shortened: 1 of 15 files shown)


tp:
@SandPod I ran this version locally and the changed files from this PR look good indeed (still have to try the tests to see that they work as well).

Just noticed that a `protocol.yaml` file gets created in my "start folder" not the newly created module's one:

CleanShot 2025-06-05 at 11 51 11@2x

When I delete that and regenerate in the `*_server` folder, it does not come back.

If you know where to look, maybe a good quick fix. Else I can also add a follow up ticket :)

tp (approved):
Tried it locally, looks good :)

SandPod:
> @SandPod I ran this version locally and the changed files from this PR look good indeed (still have to try the tests to see that they work as well).
> 
> Just noticed that a `protocol.yaml` file gets created in my "start folder" not the newly created module's one:
> 
> CleanShot 2025-06-05 at 11 51 11@2x
> 
> When I delete that and regenerate in the `*_server` folder, it does not come back.
> 
> If you know where to look, maybe a good quick fix. Else I can also add a follow up ticket :)

I'll resolve this as a separate issue. Seems like this is a problem for normal project creation as well.

PR #3616, opened 2025-06-05
fix: Generate endpoint description inside of newly created projects.

SandPod:
@tp Noticed that we generated the `protocol.yaml` (endpoint description) file outside of the serverpod directory when creating new projects.

Original post here: https://github.com/serverpod/serverpod/pull/3614#issuecomment-2943532928

This file is used by insights to list all endpoints for the project and would be generated correctly as soon as the user generates their serverpod project.

It was generated in the wrong place because the generator used a hard coded path to where the file should be generated.

This PR:
- Adds tests that validate that we generate the file in the correct place for newly created projects
- Removes all files that should be generated in modules
- Adds a relative generate path based on the configuration of the project for the `protocol.yaml` file.

### Additions after first review

I noticed that the touchups made for the analysis and removal of generated files in the normal serverpod project templates had not been applied here.

So I made these changes here as well: https://github.com/serverpod/serverpod/pull/3408/commits/7483b8631ea8f7ff8803502c7ee3c32bef2279de

## Pre-launch Checklist

## Breaking changes

None - simple bug fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ templates/serverpod_templates/projectname_client/lib/src/
 templates/serverpod_templates/projectname_server/lib/src/generated/
 templates/serverpod_templates/projectname_server/test/integration/test_tools/
 templates/serverpod_templates/projectname_server_upgrade/test/
+templates/serverpod_templates/modulename_client/lib/src/
+templates/serverpod_templates/modulename_server/lib/src/generated/
+templates/serverpod_templates/modulename_server/test/integration/test_tools/


tools/serverpod_cli/lib/src/generator/yaml/endpoint_description_generator.dart
```diff
@@ -26,6 +26,10 @@ class EndpointDescriptionGenerator extends CodeGenerator {
       }
     }
 
-    return {p.join('lib', 'src', 'generated', 'protocol.yaml'): out};
```
tp, line 29: Ah, makes sense that this was then working inside the module folder, but not from outside (when creating it).

And it did work before (the file also being inside the `_server` part) because a file with the same content was just copied there. Good to know, I first wondered whether it would be generated twice (inside and out) or something…

tp (approved):
LGTM!

SandPod:
> Question is: Should we exlude generation inside the template in CI? Or trust, that these will also be overwritten a second time (when generating the module after copying them)?

I added ignores for these. Also applied more files that could be removed that had not been applied to the module template.

tp (approved):
Looks good, also great to have the computed files not ignored by default in the analyzer.

PR #3617, opened 2025-06-05
New Service Locator utility

cannon-fodder:
New service locator utility to alleviate the 

Issue 3154

## Pre-launch Checklist

## Breaking changes

no breaking changes


packages/serverpod/lib/src/util/service_locator/service_locator.dart
```diff
@@ -0,0 +1,119 @@
+import 'locator_exceptions.dart';
+
+/// A service locator interface that allows for locating services by type or key.
+// This interface is used to decouple service registration and retrieval.
+abstract interface class ServiceLocator {
+  /// Locates a service by its type.
+  /// Throws [ServiceNotFoundException] if the service is not found.
+  /// @param T The type of the service to locate.
+  /// @return The service of type [T] if found.
+  T locateType<T>();
```
tp: Have you considered combining the look-up functions? I think that might read a bit nicer when used.

```suggestion
  T locate<T>([String? key]);
```

or

```suggestion
  T locate<T>({String? key});
```

packages/serverpod/lib/src/util/service_locator/locator_exceptions.dart
```diff
@@ -0,0 +1,47 @@
+class ServiceLocatorException implements Exception {
```
tp: Unused, since there are now specialized exception classes.

Should we remove it, or was the idea to use this as a base (abstract / sealed) for the specific cases?

cannon-fodder: @SandPod you have comments?

cannon-fodder: Is there a base exception that exceptions should extend?  Might have missed it when I skimmed the contribution docs.

packages/serverpod/lib/src/util/service_locator/service_locator.dart
cannon-fodder: I've consolidated the methods as suggested and updated the tests

packages/serverpod/lib/src/util/service_locator/locator_exceptions.dart
tp: > Is there a base exception that exceptions should extend? Might have missed it when I skimmed the contribution docs.

No, I meant whether you intended `ServiceLocatorException` to be a base exception for the service locator. Otherwise I think we can remove it, as it does not seem used.

Serverpod generally does not have any specific "base exception".

The only special case I am aware of are [serializable exceptions](https://docs.serverpod.dev/concepts/exceptions#serializable-exceptions), meaning those which will transparently be exposed to the client.

In this case (for the look up of a server-internal dependency) there is of course no need to expose that specific error to the client, and your manually defined exception classes seem just fine :)

cannon-fodder: Ah, my mistake.  I think I defined that and then changed to more specific exceptions and never removed it.  Removed now as it is unused, like you said.  I also cleaned up the lints in the exception classes.

SandPod:
@cannon-fodder - Didn't get around to reviewing this one day. Will pick it up first thing tomorrow morning.

SandPod:
@cannon-fodder I have a couple of considerations I think should be added. I'll add modifications directly to your branch and ask you for your feedback.

SandPod:
@cannon-fodder - Check out the modification.

Tomorrow I will hook this up to our test framework as well (that will probably change how it is configured on Serverpod), but now it is possible to register services on `Serverpod` and access them through a view only mode from `Session`.

I placed it under experimental so that we have the possibility to test this out and make breaking changes without promising a full interface.

cannon-fodder:
@SandPod Those all seem reasonable to me.  I did think that having the parent reference in ServiceHolder was a likely useful thing (I'm happy to go with whatever you think is best).   My thinking of the usage was ServerPod would initialize an instance with core services and pass it off to Server instance.  Server instances could wrap the passed in ServiceLocator to configure additional service or override the default ones.  The default class that just threw an exception was kind of ugly, but it seemed like an easy way to avoid the complications of null handling.

SandPod:
@cannon-fodder - That is a good angle I did not consider.

When replacing core services, some additional considerations must be made. For example, we might not want to initialize a core service if we replace it in the service locator. Therefore, replacing a core service would ideally hook up into the boot sequence of the server.

So, ideally, a Service would implement an interface that supports:
- `init(ServerpodConfig, Passwords)` - Called when the Server is started.
- `shutdown()` - Called when the Server is stopped.

Perhaps we should take this into consideration for this PR as well?

SandPod:
A neat way to support this would be to have all services that implement some form of `Init` interface to participate in the `init` sequence and all services that implement some form of `Shutdown` interface participate in the `shutdown` sequence.

I'll see if I can create a draft for this.

SandPod:
@cannon-fodder I added the possibility to hook into the lifecycle of the Server now. That should be pretty useful.

Now this PR only needs tests that validate this works as expected on start and stop of the server and support for the services in the test framework.

SandPod:
@cannon-fodder - Just giving you an update here. I'm still tinkering on this.

I want this to be a coherent experience together with this addition: https://github.com/serverpod/serverpod/pull/3555

So I'm adding that first, and then we will rework the service locator to work well with that system.

migalv:
Hey @SandPod I'm following this implementation.

I'm glad you guys are taking the time to add this. It will make Serverpod way more maintainable and enterprise/production ready.

May I see an example of how this would work?

Maybe I can give some feedback

SandPod:
That sounds good!

This needs some more tinkering as we should take service start and shutdown into consideration.

frederikstonge:
Coming from .NET aspnet core, I think it would be a good idea if we could register different styles of services.

Singleton -> created once and used across the lifetime of the application.
Scoped -> In aspnet core, each request has its own scope. It would be nice to be able to register scoped services, where the create callback provides the session or the request context. This way, you can create a tailored service for the request that will remain the same across the lifetime of the request.
Transcient -> New one created everytime it is requested.

Issue #3578, opened 2025-05-23
Add "service account" authentication


tp:
In order to facility external services accessing a Serverpod API with the new auth modules from #3126, we want to add the ability to add "service accounts".

Unlike the usual authentication providers (Email, Google, etc.) these would not refer to an `AuthUser`, but comprise a distinct entity.

The model might look like this:

```
ServiceAccount
  id: Uuid

  # Salted and peppered like we do for the email
  # The actual secret is only returned once to the original creator,
  # and can not be requested again.
  secretHash: String

  description: String

  scopeNames: Set<String>

  createdAt: DateTime, default=now
   
  # Maybe defaults to 1 year in the examples
  expires: DateTime?

  # This will be set during the creation, but is optional such that the user can be deleted without loosing any service accounts it has created
  # (This is of course problematic if they retain a copy of the secret, but then admins will just have to rotate them)
  createdBy: AuthUser?, relation(optional)
```

The endpoint (abstract, so consumer can opt to modify it) might look like this:

```dart
@doNotGenerate
class ServiceAccountEndpoint {
   requiresLogin => true;

   requiredScopes => {admin}

   /// Returns the service account key in the form to be used with the `authenticationHandler`
   /// Possibly like `base64('spsa:$id:$secret')`
   Future<String> createServiceAccount(Session, {Set<Scope>? scopes, String? description, DateTime? expiresAt,}) { … }

   Future<List<ServiceAccountModel>> listServiceAccounts(…) { … }

   Future<void> deleteServiceAccount(…) { … }   
}
```

Additionally, we extract the business logic into a `ServiceAccounts` class.

And of course, the module will need to provide an `authenticationHandler` implementation.

tp:
This is now support by the more flexible `serverpod_auth_session` module in https://github.com/serverpod/serverpod/pull/3631.

Since we decided to not create a separate entity for "service account", but rather let them be `AuthUser`s (so permission management is simple throughout an app and does not have to check many different types or work across distinct database tables), a normal session like this can also be used for service accounts.

When creating a "service account" session, the following considerations are worth taking:
- Should a fixed `expiresAt` be set, so that the session key ("API token") for the service account expires at the latest after for example 1 year?
  - This would also ensure a minimum baseline of API key-rotations in the system
- Would an `expireAfterUnusedFor` of e.g. 2 weeks makes sense, to quickly shut down any service account access keys that are not in active use? This would make special sense if the intended use is a continuous or at least daily job.

[...]

PR #3618, opened 2025-06-06
feat(auth2): Implement `serverpod_auth_api_token` module

tp:


modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/models/api_token.spy.yaml
```diff
@@ -0,0 +1,49 @@
...
+  ### The salt used for computing the [apiTokenHash].
+  ###
+  ### Per default uses 16 bytes of random data.
+  apiTokenSalt: ByteData
+
+  ### The kind of API token this represents.
+  ###
+  ### This does not impact the behavior of the token, but can be useful for debugging.
+  ### For example this could be used to differentiate between anonymous accounts, personal access tokens, or service accounts.
+  kind: String, default=""
```
tp, line 49: This might be useful here, to then also later check that tokens of a specific kind are e.g. updated to the latest desired behavior (max lifetime, automatic expiration etc.).

I suppose this could also be useful on the `AuthUser` itself, if we don't purely assume them being "humans".

modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/models/api_token.spy.yaml
```diff
@@ -0,0 +1,49 @@
+class: ApiToken
+serverOnly: true
+table: serverpod_auth_session
+fields:
```
tp, line 4: Maybe this also needs something like `extraClaims` (also exposed on the `AuthenticationInfo`), so one can easily add extra checks (e.g. what a personal access token should be able to do, if it's e.g. just for a specific resource, etc.)

modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/business/api_tokens.dart
```diff
@@ -0,0 +1,145 @@
...
+            .isBefore(clock.now())) {
+      session.log(
+        'Got API token which expired due to inactivity.',
+        level: LogLevel.debug,
+      );
+
+      return null;
+    }
+
+    return AuthenticationInfo(
```
tp, line 107: Update API token's `lastUsed` (with minute resolution)

modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/business/api_tokens.dart
```diff
@@ -0,0 +1,145 @@
...
+    required final UuidValue authUserId,
+    required final Set<Scope> scopes,
+
+    /// Fixed date at which the token expires.
+    ///
+    /// If `null` the token will work until it's deleted or when it's been inactive for [expireAfterUnusedFor].
+    final DateTime? expiresAt,
+
+    /// Length of inactivity after which the token is no longer usable.
+    final Duration? expireAfterUnusedFor,
```
tp, line 131: Take `Transaction? transaction` for all these methods

modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/models/api_token.spy.yaml
```diff
@@ -0,0 +1,49 @@
+class: ApiToken
```
tp, line 1: @SandPod As discussed last week, you might want to give the model spec a look here upfront, to make sure that we align on the breadth of this module.

I think the implementation follows from this one quite naturally.

modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/models/api_token.spy.yaml
SandPod: Is this usually called `label`?

modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/models/api_token.spy.yaml
SandPod: Hmm, should we in that case also add it to the db backed session token?

modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/models/api_token.spy.yaml
SandPod: I think the overall design looks good, I'm wondering if reate-limiting should be something to consider for an api token?

Maybe that should be built into the authenticationHandler directly?

modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/models/api_token.spy.yaml
tp: `label` to me sounds more like `description`.

The usage I envision would be more like `kind` or `type`, meaning there should be a small, finite set of options (we probably would also need to index these, as they can be passed in the list query).

modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/models/api_token.spy.yaml
tp: As mentioned earlier, I think it'd be better to use a separate table for any more advanced usage here, to get fully typed access.

"`kind`" just seems to common that we might want to provide it by default (some for the auth user), but after that more details should probably be looked up in a better interface.

(For the JWT tokens I also would hope that the primary consumer of that info are 3rd party systems extracting the Serverpod-generated token, and not the server itself, which probably has more up to date info about those claims – unless someone wants to _really_ avoid DB lookups.)

modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/business/api_tokens.dart
tp: Transaction parameter has been added everywhere in this module.

modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/models/api_token.spy.yaml
tp: Sounds nice, I think we could optionally support this (if configured on the token).

Maybe we can even keep the storage requirements low, but using a sliding window (for the configured timeframe) and then just "upsert" a counter (so we always have indexed reads for checks and never `COUNT` queries). (Though still wondering whether we would store this on the API token row itself or externally (for better history).)

modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/models/api_token.spy.yaml
tp: @SandPod In the `auth_session` we also store the `method` for now it was created:

CleanShot 2025-06-10 at 17 11 23@2x


Just FYI, I am also not entirely clear what would be best here.

modules/new_serverpod_auth/serverpod_auth_api_tokens/serverpod_auth_api_tokens_server/lib/src/business/api_tokens.dart
tp: The update of the `lastUsed` has been added.

tp:
@SandPod I meant to remove you here, but the UI added you again.

As mentioned in https://github.com/serverpod/serverpod/issues/3578#issuecomment-2969356916, this can now be closed as the entire functionality is available in the extended `serverpod_auth_session` module.

I'll do a quick PR to remove the "dummy" API token package shortly.

PR #3619, opened 2025-06-06
feat: Enhance command line argument handling with environment variable augmentation

### LINKED ISSUE #3425: Support all CLI arguments via environment variables
## Problem to Solve  
In environments where direct control over command line arguments isn't available (e.g. some cloud setups), configuring Serverpod via CLI flags is cumbersome or impossible. Currently, only `--server-id` supports an environment variable fallback, making CLI behavior inconsistent.

## Proposal  
Extend Serverpod CLI to allow **all command line arguments** to be overridden by environment variables.  
For example, if `--mode` is a CLI option, support an equivalent `SERVERPOD_MODE` environment variable.

Environment variable names should follow a consistent naming convention such as `SERVERPOD_<ARGUMENT_NAME>` in uppercase and with dashes replaced by underscores.  
Fallback priority:
1. CLI argument (highest)
2. Environment variable
3. Default value (if any)

This should be handled internally by the CLI argument parser for maintainability.

## Use Case  
Deploying a Serverpod instance in Kubernetes or Docker without having to wrap the entrypoint script or inject CLI args manually. Example:
```bash
# Instead of passing --mode=production
export SERVERPOD_MODE=production
dart run bin/main.dart
```

## Alternatives  
- Shell wrappers to parse env vars and re-invoke the CLI with args (adds complexity)
- Custom entrypoints per deployment target

## Additional Context  
Current behavior already supports `--server-id` via `SERVERPOD_SERVER_ID`, this proposal simply generalizes that pattern to improve consistency and configurability.

exaby73:
In order to support environment variables for CLI arguments, we need to augment the CLI args. This PR aims to do exactly that.

- Introduced `EnvironmentCliArgsAugmenter` to supplement command line arguments with environment variable values.
- Updated `CommandLineArgs` to support default value checks and provide getters for various properties.
- Modified `Serverpod` to utilize the new augmenter for command line argument processing.
- Added comprehensive tests for both `CommandLineArgs` and `EnvironmentCliArgsAugmenter` to ensure correct functionality and edge case handling.

Closes #3425

## Pre-launch Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring server role, logging mode, and migration flags via command line, environment variables, or config files.
  - Introduced new server roles and logging modes, allowing more flexible server operation and diagnostics.
- **Improvements**
  - Centralized and unified server configuration management for easier customization and validation.
  - Enhanced error handling and validation for configuration inputs.
  - Deprecated direct command line argument access in favor of configuration-based access.
  - Added environment variable support and validation for run mode with fallback logic.
  - Improved logging of command line arguments during server initialization.
  - Refined command line argument parsing with default value handling and new utility methods.
- **Bug Fixes**
  - Improved fallback and precedence logic for configuration values.
- **Tests**
  - Added comprehensive tests for new configuration options and command line argument parsing.
- **Chores**
  - Updated .gitignore to exclude `.asdf` and `.tool-versions` files.


exaby73:

exaby73:
recheck

packages/serverpod/lib/src/server/command_line_args.dart
```diff
@@ -73,76 +104,100 @@ class CommandLineArgs {
...
+    required bool? applyRepairMigration,
+  })  : _runMode = runMode,
+        _serverId = serverId,
+        _loggingMode = loggingMode,
+        _role = role,
+        _applyMigrations = applyMigrations,
+        _applyRepairMigration = applyRepairMigration;
+
+  /// Creates a copy of this [CommandLineArgs] with the given fields replaced.
+  CommandLineArgs copyWith({
```
tp: Thought: Would it make sense to either

1) Use a kind of reverse "copyWith" like "provideFallback", that sets the value when there is none yet? (better name TBD) That way we might be able to keep the "is default" fields private and this class would handle taking fallbacks internally

2) Or is there a chance to flip this on it's head and start out parsing and merging from the fallback in, such that start out with a default configuration, then apply the env one, and then the CLI flags? (Which would of course require that all those higher-level steps can always give `null` if no value is set).

Just 2 ideas that came when looking at this, though I must admit that I don't have any experience with this part of Serverpod so I might be missing something obvious. (Or suggest something that might be fine now, but would not help moving to the CLI tools.)

Either way, I think my overall impression here is that it's good and fully understandable in the current implementation, so in case the CLI tools "upgrade" is still planned, this might be a good intermediate step in itself :)

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -100,11 +100,13 @@ class ServerpodConfig {
   factory ServerpodConfig.loadFromMap(
     String runMode,
     String serverId,
+    bool isServerIdDefault,
```
tp: Nit: Coming at this method from the usage below, I was at first confused what the literal true/false would mean.

Usually it's probably nicer to use a named parameter if the bool is one of many arguments.
Or could this be a small record, if the flag is really needed for each serverID?

(Again, this might also not matter much going forward, so take this with a grain of salt.)

packages/serverpod/lib/src/server/environment_cli_args_augmenter.dart
```diff
@@ -0,0 +1,106 @@
...
+  EnvironmentCliArgsAugmenter(this._environment);
+
+  /// Augments the provided command line arguments with environment variable values.
+  ///
+  /// For each argument that is set to its default value (not explicitly provided
+  /// via command line), this method will check if a corresponding environment
+  /// variable exists and use that value instead.
+  ///
+  /// The following arguments can be augmented from environment variables:
+  /// - runMode: From SERVERPOD_MODE
```
christerswahn: Suggestion: More specific env var names:
`SERVERPOD_RUN_MODE`
`SERVERPOD_LOGGING_MODE`
`SERVERPOD_SERVER_ROLE` (or similar)

packages/serverpod/lib/src/server/environment_cli_args_augmenter.dart
```diff
@@ -0,0 +1,106 @@
...
+  /// - serverId: From SERVERPOD_SERVER_ID
+  /// - loggingMode: From SERVERPOD_LOGGING
+  /// - role: From SERVERPOD_ROLE
+  /// - applyMigrations: From SERVERPOD_APPLY_MIGRATIONS (expects 'true' for true)
+  /// - applyRepairMigration: From SERVERPOD_APPLY_REPAIR_MIGRATION (expects 'true' for true)
+  ///
+  /// Returns a new [CommandLineArgs] instance with the augmented values.
+  CommandLineArgs createAugmentedFrom(CommandLineArgs commandLineArgs) {
```
SandPod: Question: What happens if an invalid value is supplied as an ENV? For example `alex` is supplied to `SERVERPOD_APPLY_MIGRATIONS`?

packages/serverpod/lib/src/server/environment_cli_args_augmenter.dart
exaby73: I based these names on the CLI flags themselves. eg `--mode` became `SERVERPOD_MODE`. Renaming them would be more descriptive but also they wouldn't match the CLI flags they override (when not provided). Could this be a point of confusion?

packages/serverpod/test/server/command_line_args_test.dart
```diff
@@ -0,0 +1,298 @@
+import 'package:serverpod/src/server/command_line_args.dart';
+import 'package:serverpod/src/server/run_mode.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('CommandLineArgs', () {
```
SandPod: Suggestion: Use explicit Given, When, Then pattern for tests and only use a group if there is a shared setup.

We always try to use the same pattern for all tests. This gives us consistent tests that reduces the cognitive load to read tests from different parts of the project.

First test would be `Given provided values for all valid arguments when parsing then values are available through parsed command line fields.`.

packages/serverpod/lib/src/server/command_line_args.dart
```diff
@@ -35,29 +35,60 @@ enum ServerpodLoggingMode {
...
   /// The overarching logging mode of the server. This can be set to either
   /// [normal] or [verbose].
-  late final ServerpodLoggingMode loggingMode;
+  ServerpodLoggingMode get loggingMode =>
+      _loggingMode ?? ServerpodLoggingMode.normal;
+
+  /// Whether the logging mode is using the default value.
+  bool get isLoggingModeDefault => _loggingMode == null;
+
+  late final String? _serverId;
```
SandPod: Discussion: I'm not completely conviced that the complexity here is required to solve this problem.

Since we are not going to make these fields available through the config, it doesn't seem like we need to know what the source of the value was. It feels like we could keep these non-nullable but still achieve the `env` enrichment, depending on how we process the input sources.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -100,11 +100,13 @@ class ServerpodConfig {
   factory ServerpodConfig.loadFromMap(
```
SandPod, line 118: Suggestion: Modifying this interface is a breaking change.

If we are to modify it, add this as a named optional parameter.

SandPod (commented):
Great job dipping your toes in the Serverpod framework. As usual, what looks like a simple task actually has more complexity bellow the surface.

I really appriciate that you have added extensive testing around the proposed solution.

I have a couple of comments that I think are best to talk about in a call to ensure we take a valid approach to the problem we are solving.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -770,27 +802,111 @@ int _readMaxRequestSize(
...
       'default';
   return serverId;
 }
 
+ServerpodRole _readRole(
+  Map<dynamic, dynamic> configMap,
+  Map<String, String> environment,
+  Map<String, dynamic>? commandLineArgs,
+) {
+  if (commandLineArgs?[CliArgsConstants.role] != null) {
```
SandPod: Suggestion: Please dereff before checking if it is null.

This allows us to remove the `!` on the next line.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -770,27 +802,111 @@ int _readMaxRequestSize(
...
+    return commandLineArgs![CliArgsConstants.role] as ServerpodRole;
```
SandPod: Question: What happens if the value here is not a ServerpodRole? For example of the commandLineArgs map was populated by the user?

I think this should give a clear error message in that case.

packages/serverpod_shared/test/session_logs_config_test.dart
```diff
@@ -5,7 +5,9 @@ import 'package:yaml/yaml.dart';
 void main() {
   var developmentRunMode = 'development';
   var productionRunMode = 'production';
-  var serverId = 'default';
+  // Setting the serverId to null to test the default value.
+  // ignore: avoid_init_to_null
+  var serverId = null;
```
SandPod: Suggestion: Make this an explicit test.

I don't think this should be grouped with the session logs config tests. This should probably be a top level test that strictly validates this behavior.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -261,6 +261,51 @@ class ServerpodConfig {
     }
   }
 
+  /// Creates a copy of this [ServerpodConfig] with the given fields replaced.
+  /// All fields that are not specified will keep their original values.
+  ServerpodConfig copyWith({
```
SandPod, line 261: Question: First of, great addition!

Then, to my question: Since "nulling" a configuration actually means something in the config context. Should we take that into consideration for the copyWith case and perhaps use another way of copying the configuration? Or are we strictly addative in the `copyWith` and force users to create a new ServerpodConfig object with the fields from the old one if parts of the configuration should be replaced?

packages/serverpod/test/server/command_line_args_test.dart
```diff
@@ -4,296 +4,311 @@ import 'package:serverpod_shared/serverpod_shared.dart';
...
 
-        expect(args.isRunModeDefault, isFalse);
-        expect(args.isServerIdDefault, isTrue);
-        expect(args.isLoggingModeDefault, isTrue);
-        expect(args.isRoleDefault, isTrue);
-        expect(args.isApplyMigrationsDefault, isFalse);
-        expect(args.isApplyRepairMigrationDefault, isTrue);
-      });
+    test('default values are formatted correctly', () {
+      final args = CommandLineArgs([]);
```
SandPod: Suggestion: Use groups to group together shared `Given` or `Given, When` conditions.

This test has the same precondition as the one in the begining when we are testing the default values.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -324,24 +322,26 @@ class Serverpod {
     // Read command line arguments.
     commandLineArgs = CommandLineArgs(args);
 
+    final runMode = _calculateRunMode(commandLineArgs);
```
SandPod: Suggestion: Since we now use `config` as the source of truth, how about we mark `commandLineArgs` as deprecated and inform users where to find each field in the `config` instead.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -770,27 +802,111 @@ int _readMaxRequestSize(
...
+ServerpodRole _readRole(
+  Map<dynamic, dynamic> configMap,
+  Map<String, String> environment,
+  Map<String, dynamic>? commandLineArgs,
+) {
+  if (commandLineArgs?[CliArgsConstants.role] != null) {
+    return commandLineArgs![CliArgsConstants.role] as ServerpodRole;
+  }
+
+  final roleFromEnv = environment[ServerpodEnv.role.envVariable];
```
SandPod: Question: Explore if it is possible to reuse the `_buildConfigMap` helper to parse these values, this provides us with a standardized way of giving errors.

templates/pubspecs/packages/serverpod_test/pubspec.yaml
```diff
@@ -16,6 +16,7 @@ dependencies:
   serverpod: SERVERPOD_VERSION
   meta: ^1.11.0
   synchronized: ^3.1.0
+  serverpod_shared: SERVERPOD_VERSION
```
SandPod: Question: Is this really needed?

Validate that this import is really needed since our export surface from Serverpod should not change.

packages/serverpod/lib/src/server/health_check_manager.dart
```diff
@@ -385,7 +385,7 @@ class HealthCheckManager {
       space: OriginSpace.framework,
       context: DiagnosticEventContext(
         serverId: _pod.serverId,
-        serverRunMode: _pod.commandLineArgs.role.name,
+        serverRunMode: _pod.config.role.name,
         serverName: '',
```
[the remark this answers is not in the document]
exaby73: This is expected

packages/serverpod/lib/src/server/health_check_manager.dart
```diff
@@ -98,9 +98,9 @@ class HealthCheckManager {
 
     // If we are running in maintenance mode, we don't want to schedule the next
     // health check, as it should only be run once.
-    if (_pod.commandLineArgs.role == ServerpodRole.monolith) {
+    if (_pod.config.role == ServerpodRole.monolith) {
       _scheduleNextCheck();
-    } else if (_pod.commandLineArgs.role == ServerpodRole.maintenance) {
+    } else if (_pod.config.role == ServerpodRole.maintenance) {
       onCompleted();
     }
```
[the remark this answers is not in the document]
exaby73: Serverless mode does not have this feature enabled

packages/serverpod/test/server/command_line_args_test.dart
```diff
@@ -0,0 +1,335 @@
...
+      // All should be null because invalid arg triggers catch block
+      expect(args.getRaw<String>(CliArgsConstants.runMode), isNull);
+      expect(args.getRaw<String>(CliArgsConstants.serverId), isNull);
+      expect(args.getRaw<ServerpodLoggingMode>(CliArgsConstants.loggingMode),
+          isNull);
+      expect(args.getRaw<ServerpodRole>(CliArgsConstants.role), isNull);
+      expect(args.getRaw<bool>(CliArgsConstants.applyMigrations), isNull);
+      expect(args.getRaw<bool>(CliArgsConstants.applyRepairMigration), isNull);
+    });
+  });
```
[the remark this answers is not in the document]
exaby73: This is fine since we print a log for this

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -56,12 +56,11 @@ class Serverpod {
...
   }
 
-  late String _runMode;
-
   /// The servers run mode as specified in [ServerpodRunMode].
-  String get runMode => _runMode;
+  String get runMode => config.runMode;
 
   /// The parsed runtime arguments passed to Serverpod at startup.
+  @Deprecated('Use config instead.')
```
SandPod: Suggestion: Expand this message with information on why `config` should be used instead and that this will be removed in a future version.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -322,24 +321,33 @@ class Serverpod {
     );
 
     // Read command line arguments.
+    // ignore: deprecated_member_use_from_same_package
     commandLineArgs = CommandLineArgs(args);
```
SandPod: Suggestion(NIT): Add a private `_commandLineArgs` to not have to have this `ignore` spread in the file.

Then convert the public `commandLineArgs` into a getter.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -322,24 +321,33 @@ class Serverpod {
...
-          _runMode,
-          commandLineArgs.serverId,
+          runMode,
+          serverId,
           _passwords,
+          // ignore: deprecated_member_use_from_same_package
+          commandLineArgs: commandLineArgs.toMap(),
         );
 
+    // ignore: deprecated_member_use_from_same_package
```
SandPod: Suggestion: Remove ignore.

This ignore is not required right?

packages/serverpod/lib/src/server/command_line_args.dart
```diff
@@ -1,63 +1,48 @@
...
-/// which is the default.
-enum ServerpodLoggingMode {
-  /// Only log important messages.
-  normal,
-
-  /// Log all messages, including debugging information.
-  verbose,
-}
-
 /// Parses the command line arguments passed to Serverpod.
```
SandPod, line 8: Suggestion: Revert redundant modifications to CommandLineArgs.

I always opt for not changing code taht we don't need to change. From this class we basically only need the new Map and single raw values right?
Could you remove the redundant copy with addition and the `_internal` constructor?

packages/serverpod/test/server/command_line_args_test.dart
```diff
@@ -0,0 +1,335 @@
...
+    test('when parsing then default values are used', () {
+      expect(args.runMode, equals(ServerpodRunMode.development));
+      expect(args.serverId, equals('default'));
+      expect(args.loggingMode, equals(ServerpodLoggingMode.normal));
+      expect(args.role, equals(ServerpodRole.monolith));
+      expect(args.applyMigrations, isFalse);
+      expect(args.applyRepairMigration, isFalse);
+    });
+
+    test('when using getRaw method then returns null for all arguments', () {
```
SandPod, line 104: Suggestion: Missing test for the `toMap` method when all default values are used.

packages/serverpod/test/server/command_line_args_test.dart
```diff
@@ -0,0 +1,335 @@
...
+        () {
+      expect(args.runMode, equals('test'));
+      expect(args.serverId, equals('default'));
+      expect(args.loggingMode, equals(ServerpodLoggingMode.normal));
+      expect(args.role, equals(ServerpodRole.monolith));
+      expect(args.applyMigrations, isTrue);
+      expect(args.applyRepairMigration, isFalse);
+    });
+
+    test('when using getRaw method then returns mixed raw values', () {
```
SandPod, line 159: Suggestion: Missing test for when `toMap` is used for this scenario.

packages/serverpod/test/server/command_line_args_test.dart
```diff
@@ -0,0 +1,335 @@
...
+
+  group('Given command line arguments explicitly set to default values', () {
+    final args = CommandLineArgs([
+      '--mode', 'development', // Same as default but explicitly provided
+      '--server-id', 'default', // Same as default but explicitly provided
+      '--logging', 'normal', // Same as default but explicitly provided
+      '--role', 'monolith', // Same as default but explicitly provided
+    ]);
+
+    test(
```
SandPod, line 189: Suggestion: Missing test for when `toMap` is used in this scenario.

packages/serverpod/test/server/command_line_args_test.dart
```diff
@@ -0,0 +1,335 @@
...
+      expect(args.getRaw<String>(CliArgsConstants.serverId), isNull);
+      expect(args.getRaw<ServerpodLoggingMode>(CliArgsConstants.loggingMode),
+          isNull);
+      expect(args.getRaw<ServerpodRole>(CliArgsConstants.role), isNull);
+      expect(args.getRaw<bool>(CliArgsConstants.applyMigrations), isNull);
+      expect(args.getRaw<bool>(CliArgsConstants.applyRepairMigration), isNull);
+    });
+  });
+
+  group('Given CommandLineArgs instance when using copyWith method', () {
```
SandPod: Suggestion: This test can be removed if the now redundant functionality is removed.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -322,24 +321,33 @@ class Serverpod {
...
-    _runMode = commandLineArgs.runMode;
+    // ignore: deprecated_member_use_from_same_package
+    final runMode = _calculateRunMode(commandLineArgs);
+
     // Load passwords
     _passwordManager = PasswordManager(runMode: runMode);
     _passwords = _passwordManager.loadPasswords();
 
+    // ignore: deprecated_member_use_from_same_package
+    final serverId = _calculateServerId(commandLineArgs);
```
SandPod: Suggestion: Inline the rar string fetch in the call to `ServerpodConfig.load` since there is no consolidation going on here.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -706,6 +710,30 @@ class Serverpod {
     }
   }
 
+  String _calculateRunMode(CommandLineArgs commandLineArgs) {
```
SandPod: Suggestion: Take the raw value as an input parameter here. Then we don't need to "spread" knowledge of the `CommandLineArgs` class.

packages/serverpod_shared/test/config_test.dart
```diff
@@ -494,7 +496,7 @@ redis:
     var config = ServerpodConfig.loadFromMap(
       runMode,
       serverId,
-      passwords,
+      {...passwords, 'webServer': 'password'},
```
SandPod: Question: Why is this change needed? Revert if possible.

packages/serverpod_shared/test/config_test.dart
```diff
@@ -1067,4 +1069,705 @@ futureCallExecutionEnabled: false
       expect(config.database?.user, 'testuser');
     });
   });
+
+  group(
+      'Given a Serverpod config with role configuration when loading from Map then',
```
SandPod: Suggestion(Style): The following comment is not followed in this file, but should apply generally to all tests we write and I would like you to help safeguard these principles going forward.

Two things:
1. Place the `then` part at the start of the `then` statement.
2. The Given for these tests are different. They should probably not be in the same group.

First has no role as Given, second has role in env as Given, third has role from args as Given, etc..
The formatting and input of role are all different Givens.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -756,17 +841,204 @@ int _readMaxRequestSize(
...
+  }
+
+  final configKey = ServerpodEnv.applyRepairMigration.configKey;
+  final applyRepairMigrationFromConfig = configMap[configKey];
+  if (applyRepairMigrationFromConfig != null) {
+    if (applyRepairMigrationFromConfig is bool) {
+      return applyRepairMigrationFromConfig;
+    }
+    return switch (applyRepairMigrationFromConfig.toString()) {
+      'true' || '1' => true,
```
SandPod: Suggestion: Please remove support for `1` and `0`.

Although `1` and `0` are neat to add, I don't think we should sneak them in here while not supporting them in all our other configuration.

SandPod (commented):
Last round of comment. I'm reviewing a bit more throughly to ensure that we align on the style of the code and to set expectations regarding contributions to Serverpod in the future.

This should be all that is needed to get this in.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -335,13 +335,11 @@ class Serverpod {
...
     _passwords = _passwordManager.loadPasswords();
 
-    final serverId = _calculateServerId(_commandLineArgs);
-
     // Load config
     this.config = config?.copyWith(runMode: runMode) ??
         ServerpodConfig.load(
           runMode,
-          serverId,
+          _calculateServerId(_commandLineArgs),
```
SandPod: Clarification: My suggestion was to completely remove the `_calculateServerId()` private method, since we don't actually calculate anything and don't do any consolidation. We could access the `raw` value inline here.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -725,8 +727,7 @@ class Serverpod {
       };
     }
 
-    // Default to development if no run mode is provided.
-    return commandLineArgs.runMode;
+    return 'development';
```
SandPod: Suggestion: Don't we have a constant that we can use for this?

SandPod (approved):
LGTM! 👍 

Please address the last two comments and we are ready to merge.

PR #3621, opened 2025-06-07
chore: Move `generateRandomBytes` to `serverpod_shared`

tp:
As this secure byte-level randomness generation has replaced the previous  (base64) `generateRandomString` in the new auth packages

Also moves `uint8ListAreEqual`, which is needed to be able to compare the derived hashes from the random outputs later on


SandPod (approved):
LGTM! 👍

PR #3622, opened 2025-06-07
fix(auth2): Update `serverpod_auth_session` to align with the latest patterns used in the new auth packages

tp:
Most importantly, this updates the `hash` + `salt` storage using `ByteData` in the database, instead of us building `base64`-encoded concatenated strings ourselves.

Beyond that, this moves the `config` and `secrets` configuration instances to the `AuthSessions` class, and makes both of them instances (thus also only looking up the values once from the "env").

Lastly, the tests were moved around into smaller groups, better aligning with the prevailing pattern and doing all setup beforehand, outside of the `test` callback.


modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_session_config.dart
```diff
@@ -5,9 +5,21 @@ class AuthSessionConfig {
...
+  final int sessionKeyHashSaltLength;
+
   /// Create a new user session configuration.
   AuthSessionConfig({
     this.maximumSessionLifetime,
+    this.sessionKeySecretLength = 32,
+    this.sessionKeyHashSaltLength = 16,
   });
 
   /// The current session module configuration.
```
tp: TODO: Move this to the `AuthSessions` class.

tp: Move the config to the main entry point of the package.

SandPod (approved):
Misc fixes.

Looks good to me!
Just renamed the title to `fix` since we are not really introducing any new feature here, just applying review feedback to a different package.

Issue #3580, opened 2025-05-25
Support extra types from `pgvector` extension `HalfVector`, `SparseVector` and `Bit`


marcelomendoncasoares:
While implementing the #2471 issue, only the main `Vector` type from `pgvector` extension was covered in the PR #3554. To achieve full support of the `pgvector` library, the other three types must be supported as well: `HalfVector`, `SparseVector` and `Bit`.

For `HalfVector`, a custom `Float16` binary parsing will be required, since [Dart does not support Float16 yet](https://github.com/dart-lang/sdk/issues/56319) and the database always return the data as `bytes`. The issue is somewhat stale, so it is better to implement the custom parsing on Serverpod.

While `HalfVector` and `SparseVector` support the same distance operations of `Vector`, it is important to note that `Bit` only supports `Jaccard` and `Hamming` distances (which are not available for the others). This means that it will not share the same distances with the current `ColumnVector` implementation. It is also important to validate the supported `VectorDistanceFunction` for each vector type when creating indexes (since one vector index can only have one vector column).

SandPod:
Fixed by: [#3623](https://github.com/serverpod/serverpod/pull/3623)

PR #3623, opened 2025-06-08
feat: Add `HalfVector`, `SparseVector` and `Bit` vector types

marcelomendoncasoares:
This PR closes the issue #3580 with the implementation of all missing types from `pgvector` extension.

Some changes were introduced to the types in comparison to the original `pgvector-dart` code:
- Support for `Float16` serialization/deserialization was added to `HalfVector`.
- When creating `SparseVector` from map, indexes must be 1-based for consistency with Postgres.
- Both `SparseVector` and `Bit` now have a `fromString` convenience method.
- New tests were added to all three new vector types (and Vector testes were renamed).

All current `pgvector` limitations were translated as restrictions and must be revisited with new versions of the extension - such as the limitation of L1 distance not available for half vector with IVFFLAT index. These restrictions will also be present in the `serverpod_docs` PR that follows this one.

## Pre-launch Checklist

## Breaking changes

No breaking changes.


packages/serverpod_serialization/lib/src/extensions/serialization_extensions.dart
```diff
@@ -180,3 +180,73 @@ extension VectorJsonExtension on Vector {
...
+    if (value is Uint8List) return SparseVector.fromBinary(value);
+    if (value is String) return _fromString(value);
+    if (value is List) return SparseVector(value.cast<double>());
+    if (value is SparseVector) return value;
+
+    throw DeserializationTypeNotFoundException(type: value.runtimeType);
+  }
+
+  /// Returns a serialized version of the [SparseVector] as a [String].
+  String toJson() => toString();
```
SandPod, line 217: Question: Is this a standardized format used? My hunch tells me that a map would be a better format for the values, but that would have a tougher time with the size of the vector.

packages/serverpod_serialization/lib/src/extensions/serialization_extensions.dart
```diff
@@ -180,3 +180,73 @@ extension VectorJsonExtension on Vector {
...
+    if (value is Uint8List) return Bit.fromBinary(value);
+    if (value is String) return _fromString(value);
+    if (value is List) return _fromList(value);
+    if (value is Bit) return value;
+
+    throw DeserializationTypeNotFoundException(type: value.runtimeType);
+  }
+
+  /// Returns a serialized version of the [Bit] as a [String].
+  String toJson() => toString();
```
SandPod, line 241: Question(For my knowledge): We opted for `List<Double>` for `Vector` and `HalfVector`, why wouldn't we use `List<Bool>` for bit? (Or some other more optimal format)

tools/serverpod_cli/lib/src/generator/types.dart
```diff
@@ -135,7 +135,10 @@ class TypeDefinition {
 
   bool get isMapType => className == MapKeyword.className;
 
-  bool get isVectorType => className == VectorKeyword.className;
+  static List<String> get vectorClassNames =>
+      ['Vector', 'HalfVector', 'SparseVector', 'Bit'];
```
SandPod, line 139: Question: I'm noticing these strings being defined multiple times, spread across the project. Might be worth consolidating the same String?

SandPod (approved):
Awesome work! 👍 

Pattern is familiar and this looks solid!

Just had a couple of questions regarding the new types.
Please answer these and if all is good we can merge it.

packages/serverpod_serialization/lib/src/extensions/serialization_extensions.dart
marcelomendoncasoares: I went this same path at first. I was going with a map, but the dimension would be lost and it would be invalid JSON to serialize afterwards - only strings are allowed as JSON keys, but it would be `Map<int, double>`. The other alternative to string would be `List<double>`, but it would generate unnecessarily greater values for sparse vectors. So I decided to go with the [native format of `pgvector`](https://github.com/pgvector/pgvector/tree/master?tab=readme-ov-file#sparse-vectors), which is this one like `'{1:1,3:2,5:3}/5'`.

packages/serverpod_serialization/lib/src/extensions/serialization_extensions.dart
marcelomendoncasoares: The reason was to approximate to the [native `pgvector` format](https://github.com/pgvector/pgvector/tree/master?tab=readme-ov-file#binary-vectors) (a string like "001001") while also optimizing memory and storage. The `String` will be considerably lighter than the `List<bool>` when stored as JSON (way less chars in plain text), and about the same in-memory as the list (considering 1 byte per bool).

Comparing to the other types, it keeps consistency: we are representing the data for all vector types in the closest way to the `pgvector` representation that is most efficient in terms of storage and memory.

tools/serverpod_cli/lib/src/generator/types.dart
marcelomendoncasoares: As all other types follow this same pattern, it seemed better to me to keep the implementation consistent instead of having only these four vector types using keywords. And maybe separately we consolidate all types' names as keywords (not only vectors'). Does it make sense?

SandPod: Yes, makes total sense! 👍

PR #3626, opened 2025-06-09
chore(auth2): Add `serverpod_new_auth_test` project

tp:
Implementation follows in https://github.com/serverpod/serverpod/pull/3579


tp:
Hmm, probably should've created this as `serverpod_new_auth_test` (no `_server` suffix).

I could already move all files in here, and then apply the same in the actual implementation PR after the rebase.

SandPod:
> Hmm, probably should've created this as serverpod_new_auth_test (no _server suffix).

Yeah, please remove the suffix.

SandPod (approved):
👍

Issue #3582, opened 2025-05-26
Allow configuration of global and local `pgvector` query options


marcelomendoncasoares:
### Problem to Solve

This issue is complementary to the PR #3554 for full implementation of the #2471 issue.

When interacting with vector columns on the database, it is sometimes necessary to tune query options to achieve the desired performance. Those options can be set at a global level (using `SET <option> = <value>`) or local level (using `SET LOCAL <option> = <value>`). The `pgvector` library exposes several options to adjust recall x speed of search for [HNSW](https://github.com/pgvector/pgvector/tree/master?tab=readme-ov-file#query-options) and [IVFFLAT](https://github.com/pgvector/pgvector/tree/master?tab=readme-ov-file#query-options-1), [index build time](https://github.com/pgvector/pgvector/tree/master?tab=readme-ov-file#index-build-time), [iterative index scans](https://github.com/pgvector/pgvector/tree/master?tab=readme-ov-file#iterative-index-scans), etc.

Global query option example:
```sql
SET hnsw.ef_search = 100;
```

Local query option example:
```sql
BEGIN;
SET LOCAL hnsw.ef_search = 100;
SELECT ...
COMMIT;
```

> Note: Although this issue targets vector indexes interaction, it will also enable the configuration of any other Postgres global/local config.

### Proposal

There may be other ways to implement these options that I haven't found, and we can discuss in the comments. Below are the alternatives I've found (some pre-discussed on PR #3554):

### Project-wise configurations

Can be done either through a config file (like `generator.yaml`, but it may not be the best place by concept) or through a Dart object on the server initialization. These configs could also be declared as default on model files, but it would allow the user to set global configs on local files: any model with a vector index would be suitable to have the global configs declared, but only the first file with the config would be allowed.

### Per query configurations

It seems we have mainly two possible alternatives:

- **Expose a `queryOptions` parameter on the transaction for the user to set**: the advantage is that no change will be required on the generated classes, as the transaction is already a parameter on `find` method. Also, no change on all underlying code, as the transaction will handle the injection of the config. It is less accessible than a top-level parameter on `find`, but very straightforward to implement and predictable to use.

- **Add a `queryOptions` method on the `find` method**: easier for the user to configure, as it does not need to explicitly create a transaction before querying, but would have to automagically wrap the query in a transaction with the query options. The `find` method would change, so the implementation will require changing the code generation.

### Use Case

- Adjust speed x recall of vector queries.
- Tuning index build resources (workers and memory).
[...]

PR #3627, opened 2025-06-10
feat: Add a method to set runtime parameters on the database

marcelomendoncasoares:
This closes the issue #3582 by allowing the user to set global and local runtime parameters on the database. Although the feature targeted `pgvector` parameters, the solution can be extended to set any kind of Postgres parameters.

## Implementation strategy

The feature introduces a new `setRuntimeParameter` method that receives a list of `RuntimeParameter`. If the method is called within a transaction, it will translate into `SET LOCAL` calls. All available pgvector parameters are already mapped as extensions of the `RuntimeParameter` class.

## Pre-launch Checklist

## Breaking changes

None.

## Summary by CodeRabbit

- **New Features**
  - Introduced support for configuring database runtime parameters (HNSW, IVFFlat, vector index options, and search paths) globally and per transaction.
  - Enabled setting runtime parameters via a builder interface during server initialization and within database transactions.
  - Added methods to ensure vector extensions are loaded and to reset runtime parameters after transactions.
- **Tests**
  - Added extensive unit and integration tests validating runtime parameter application, isolation, and overrides in multiple scenarios.
- **Chores**
  - Updated test utilities, server initialization, and database connection pooling to support runtime parameter configuration.


tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
```diff
@@ -0,0 +1,138 @@
...
+        iterativeScan: IterativeScan.relaxed,
+        maxScanTuples: 500,
+        scanMemMultiplier: 1,
+      );
+
+      await session.db.transaction((transaction) async {
+        await session.db.setRuntimeParameters(
+          [options],
+          transaction: transaction,
+        );
```
SandPod: Suggestion/Discussion: Would it make more sense to have this method directly available on the transaction for the local mutation?

That is `transaction.setRuntimeParameters([options])`. This makes the local vs global mutation a bit more clear.

packages/serverpod/lib/src/database/concepts/runtime_parameters.dart
```diff
@@ -0,0 +1,146 @@
+/// Base class for runtime parameters group to apply to the database.
+abstract class RuntimeParameters {
```
SandPod, line 5: Question: Do we intend users to be able to create their own RuntimeParameter classes, or should we close this off from further extension?

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
```diff
@@ -0,0 +1,138 @@
...
+        var row = result.first.toColumnMap();
+        expect(row['hnsw_ef_search'], '50');
+        expect(row['hnsw_iterative_scan'], 'relaxed_order');
+        expect(row['hnsw_max_scan_tuples'], '500');
+        expect(row['hnsw_scan_mem_multiplier'], '1');
+      });
+    });
+
+    test(
+        'when setting parameters with null values then only later non-null values are set and override earlier ones.',
```
SandPod: Question: How should someone "remove" options that are previously applied?

Since we add "global" settings it might be valuable to be able to remove previosly applied configuration.

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
```diff
@@ -0,0 +1,138 @@
+import 'package:serverpod/database.dart';
+import 'package:serverpod_test_server/test_util/test_serverpod.dart';
+import 'package:test/test.dart';
+
+void main() async {
```
SandPod, line 5: Suggestion: Add a test that queries vectors when some kind of runtime parameters are applied.

This probably won't show anywhere, but it might be good just to validate that we don't configure the database in a strange state once we use them in combination with a real query.
That would probably be the most important test as well. Would be cool if we could create a test scenarios that would give us different results depending on if the runtime setting was applied in the query or not.

SandPod (commented):
Great work and approach on this one as well!

I have a couple of questions regarding the design direction. Please take a look at them and reply to the comment.

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
marcelomendoncasoares: My choice for the transaction on the method was to keep consistency with all other `Database` methods' signatures. But you have a fair point that calling from the transaction makes an alternate usage of local/global clearer. Maybe we can have both by adding an extension method on the transaction?

packages/serverpod/lib/src/database/concepts/runtime_parameters.dart
marcelomendoncasoares: It is expected that we add more extensions in the future inside Serverpod to control other Postgres runtime parameters, but I haven't thought of this class to be frequently extended by users. Although having it as non-sealed/final does allow users to eventually extend it to support modifying some parameters that are not yet mapped inside Serverpod.

Having this possibility, I can't think of a strong argument to seal the class. If we proceed considering this use case as valid, I'll just add a sub-section in the runtime parameter docs to show how to create a custom class.

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
marcelomendoncasoares: The easy way to "remove" a config would be to set it again to the previous value (manually). For parameters that are native or declared in config files, it could be possible to retrieve the reset value through a query like this:

```sql
SELECT name, setting, boot_val, reset_val, source 
FROM pg_settings;
```

However, custom parameters may not be available until the user has set them. It is the case with `pgvector`: it's only possible to recover the parameter value after a `SET` is called. Otherwise, it will throw an error - or return `NULL` if retrieving with `SELECT current_setting('custom.parameter', true);`. So, for the vector parameters we add here, the only way to allow reverting would be by hard-coding a default value.

To add a little bit more complexity, what would be the expected behavior of a "remove" after more than one set was performed? Should the last set be removed and the value reverted to the former or a complete removal of all set's, reverting to the original default value?

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
marcelomendoncasoares: Good call! I'll add this integration test of a vector query combined with parameter set. I just can't think of a simple scenario where these parameters could alter the result because the ones that interfere with the results are mainly on the recall x search time balance - and it may require some massive test data to make a difference. But I'll try some extreme parameter values to see if we can alter the result.

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
SandPod: I would categorize this as a new type of call since we are configuring and not querying. So, deviating from the other pattern would be fine in my mind.

I guess it would be fine to support both to follow the same pattern in the file. But I think we should have it directly available on transaction as well, with documentation hinting towards that for the global option.

packages/serverpod/lib/src/database/concepts/runtime_parameters.dart
SandPod: Sounds good to keep it open in that case! 👍 More flexibility is always preffered.

But how will users discover these options?

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
SandPod: Would it be possible to have a couple of really far-away data points that would yield different orders depending on the indexes? I chatted a bit with ChatGPT, and it seemed to suggest such an approach with only five entries of 3-dimensional vectors. But that needs to be taken with a grain of salt.

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
SandPod: ### Expectation on remove

I would assume that we do not preserve a history of configuration (that is up for the user to keep track of) and a removal would reset to the original default value.

I believe that a removal is only interesting in the case that we would like to reset our state.

### Patch configuration

The benefit of the current approach is that it is possible to "patch" the configuration.

Removal of parameters might collide with this. But this creates non-delcarative configuration in the database where the actual state is built up out of multiple calls over time. Might be simpler to keep track of the current state if the full configuration is always supplied?

packages/serverpod/lib/src/database/concepts/runtime_parameters.dart
marcelomendoncasoares: I think this extension would be more for users that already know which parameters they want to set, that they may have found on the Postgres docs ([such as this](https://www.postgresql.org/docs/current/runtime-config-client.html)) or searching around how to tweak some very specific behavior.

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
marcelomendoncasoares: Working on it!

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
marcelomendoncasoares: I see your point. It may be hard to re-create the database state if settings are individually changed over time. For example, a past version of an internal method could have the set of a global value removed from the code and the parameter value would remain silently set forever.

Forcing the user to always supply the full configuration seems a bit unpleasant because unrelated parameters could override one another depending on where they are being set. If one method would set parameter A and other method B - both being part of the same `RuntimeParameter` groups -, async calls on both methods could generate unpredictable overrides. Beside this concurrency issue, the user would have to manually track which parameters are being set elsewhere in the code and always pass all values to avoid overrides.

Maybe an alternative to avoid this problem is to treat global values as only valid per session. We could forcefully call a "reset" method on the start of the server that wipes all custom runtime parameters. This way, the user would have to ensure the call to the global set in the session (either after the reset or before the code that needs it) and we would have no unpredictable state hanging from past versions.

It still maintains the risk of one part of the code setting a parameter that affects other parts of the code - and the order of the operations affecting the results -, but I guess this is unavoidable when we work with global configurations like these.

What do you think?

packages/serverpod/lib/src/database/concepts/runtime_parameters.dart
SandPod: I'm thinking more about finding these classes in the typed environment.

So, given a user does `session.db.setRuntimeParameters([]); `, how would they find out what options we provide?

Would it maybe be possible to add static extensions to `RuntimeParameters` that makes it possible to explore our options directly when viewing the interface.

That would at least allow them to see them through `session.db.setRuntimeParameters([RuntimeParameters.HnswIndexQuery(....),]);`. Does that make sense?

Or are there other ways we culd use to provide the user with our availble options?

SandPod: Another option is to make this a callback function that exposes some object these query objects are an extension of. That would make it easy for developers to add extensions of their own type.

We also already use that pattern for the `orderByList` parameter for database queries.

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
SandPod: Good point with the database state being managed separately from the Serverpod instance.

This might be something the users will need to keep track of. Especially when working with multiple servers in a cluster.

### Scope Global to connection pool.

Just to clarify, the `Global` setting will be preserved in the database and reused for all connections? There is no way to bind this to this specific server's connection? If we could apply options "Globally" for a specific database connection pool, we could move this option to become a server start option instead and then encourage the user to use a transaction if they would like to use a different options for a specific request?

### Clarify usage of options

Also, In my mind the `RuntimeParameter` group is basically a single setting in regards to querying a specific index. What usecase would there be for modifying a parameter in Method B separately from a parameter in Method A without caring about the other settings for that specific runtime parameter?

packages/serverpod/lib/src/database/concepts/runtime_parameters.dart
marcelomendoncasoares: Oh, my bad... And very good point! I like this last approach you suggested - to transform the function parameter in a callback such as `orderByList`. It's very clean to use and even easier for the developer to find the options than as static getters. I'll refactor the method in this direction.

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
marcelomendoncasoares: I was referring to the configurations as global for the database, but they are in fact global **for the session only**. To alter the database in a persistent way, the `SET` command would have to be preceded by `ALTER SYSTEM` or `ALTER DATABASE <name>`, which is not exposed by the method.

However, this highlights another problem when applying parameters as "global" with a connection pool: **it only affects the session of that connection**. This creates unpredictable behaviors between connections of a pool because it is not possible to know which parameters have been set on which connections.

Lets say the user has two unrelated methods: method A that only sets `enableIndexScan: false`; method B that only sets `minParallelTableScanSize: 1024` (both part of the `VectorIndexQueryOptions` class and set global). Depending on the order of calls:

- Some calls to method A or B would happen on connections that **executed none of them** (good path).
- Some calls to either methods would happen on connections that **only executed that same method** (good path).
- Some calls to either methods would happen on connections that **first executed the other method** (bad path).

This last case is the conflict I have mentioned before: a connection that passes by `method A -> method B` would result in performance degradation of queries executed inside method B for vector fields that have an index - the index would be ignored. As it would not happen with all calls, debugging would be very hard.

Conflicts like this could also occur in simpler cases due to the connection pool, such as:

```dart
// User 1 - gets Connection A from pool
await session.db.execute('SET statement_timeout = 5000'); // 5 seconds
await session.db.query('SELECT * FROM small_table'); 
// Connection A returned to pool with timeout = 5 seconds

// User 2 - gets Connection A from pool later
await session.db.query('SELECT * FROM huge_table'); 
// Query aborted with timeout after 5 seconds.

// User 3 - gets Connection B from pool, that never passed by the SET command
await session.db.query('SELECT * FROM huge_table'); 
// Query returned the results in 10 seconds.
```

### Possible solutions

To completely avoid the risk of scenarios like this, the best alternative I can think of is to remove the `setRuntimeParameters` method from the `Database` class and leave it only on the `Transaction`. With connection pools, using the `SET LOCAL` in transactions is the best practice and we would be protecting the users from silent conflicts like the ones above. The only "drawback" is requiring that all operations that demand a parameter customization run inside a transaction, but its not actually a bad thing. All other alternatives I can think of will scale in complexity and won't protect from parameter conflicts like these.

Do you think it makes sense to restrict parameters customization to be transaction-only, enforcing the best practice?

packages/serverpod/lib/src/database/concepts/runtime_parameters.dart
```diff
@@ -0,0 +1,207 @@
...
+  /// Returns a list with the SQL statements to set each runtime parameters.
+  /// If [isLocal] is true, options are set for the current transaction only.
+  Iterable<String> buildStatements({bool isLocal = false}) =>
+      options.entries.where((e) => e.value != null).map((e) {
+        var value = e.value;
+        if (value is RuntimeParameters) return value.build(isLocal: isLocal);
+        if (value is IterativeScan) value = '${value.name}_order';
+        if (value is bool) value = e.value == true ? 'on' : 'off';
+        return 'SET ${isLocal ? 'LOCAL ' : ''}${e.key} = $value;';
+      }).where((e) => e != '');
```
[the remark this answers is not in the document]
marcelomendoncasoares: Fixed!

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -840,6 +841,18 @@ class _PostgresTransaction implements Transaction {
...
+  Future<void> setRuntimeParameters(
+    RuntimeParametersListBuilder builder,
+  ) async {
+    final parameters = builder(RuntimeParametersBuilder());
+    await Future.wait([
+      for (var group in parameters)
+        for (var statement in group.buildStatements(isLocal: true))
+          _query(statement),
+    ]);
+  }
```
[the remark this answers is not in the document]
marcelomendoncasoares: Fixed!

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
marcelomendoncasoares: Done! I added the test in a new file, since the idea is to have one test that verifies the effect of parameters in a query. HNSW is used as the example, but with no intention to create other tests for all possible parameters, as they are heavy to run and very complex to setup. I left a few print statements on the test that were important to tweak the test thresholds, in case a future developer needs to change it.

packages/serverpod/lib/src/database/concepts/runtime_parameters.dart
marcelomendoncasoares: Done! Implemented using a builder callback that will allow users to access existing parameters by doing:

```dart
await transaction.setRuntimeParameters((params) => [
      params.vectorIndexQuery(enableSeqScan: false),
      params.hnswIndexQuery(iterativeScan: IterativeScan.strict),
    ]);
```

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
marcelomendoncasoares: Done!

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/vector_query_test.dart
```diff
@@ -0,0 +1,181 @@
...
+                ),
+              ]);
+
+          return ObjectWithVector.db.find(
+            session,
+            orderBy: (t) => t.vectorIndexedHnsw.distanceL2(queryVector),
+            limit: 30,
+            transaction: transaction,
+          );
+        });
```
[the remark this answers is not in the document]
marcelomendoncasoares: You can't actually do that. The transaction does not inherit from session.

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
SandPod: I agree with the suggested solution of removing `setRuntimeParameters` from `Session` and only keeping it on `Transaction`.

We could expose a way to alter the connection pool settings, similar to how we make it possible to configure the `search_paths`?  This doesn't need to be driven through configuration, but could be an input parameter to the constructor. That would allow us to avoid the issue mentioned above, right?

SandPod: And in such a system, I think it would make sense to always override any partial configuration.

packages/serverpod/lib/src/database/concepts/runtime_parameters.dart
SandPod: Great! The benefit of this is that it allows users to add extensions to `params` to add their own options :)

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/hnsw_index_query_options_test.dart
marcelomendoncasoares: All changes implemented as discussed:

- Global parameters can no longer be set on the `Session`, but only on `Serverpod` - and applied to all pool connections.
- All `RuntimeParameter` classes have defaults that match Postgres documentation. Set will apply the entire group.
- Any `null` parameters will result in a `SET <param> TO DEFAULT` call, which resets the parameter.

The `TO DEFAULT` solved the problem with parameters that had a 'unset' default - such as iterative scan. This way, if the global setting apply an iterative scan but the user applies only other parameters of that group, the transaction will revert the iterative scan to unset, making the full configuration consistent.

packages/serverpod/lib/src/database/concepts/runtime_parameters.dart
```diff
@@ -0,0 +1,206 @@
...
+  Iterable<String> buildStatements({bool isLocal = false}) =>
+      options.entries.map((e) {
+        var value = e.value;
+        if (value is String) value = '\'${value.replaceAll("'", "''")}\'';
+        if (value is RuntimeParameters) return value.build(isLocal: isLocal);
+        if (value is IterativeScan) value = '${value.name}_order';
+        if (value is bool) value = (value == true) ? 'on' : 'off';
+        value ??= 'TO DEFAULT';
+        return 'SET ${isLocal ? 'LOCAL ' : ''}${e.key} = $value;';
+      }).where((e) => e != '');
```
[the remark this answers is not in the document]
marcelomendoncasoares: Fixed.

tests/serverpod_test_server/test_integration/database_operations/transactions/runtime_parameters_test.dart
```diff
@@ -0,0 +1,237 @@
...
+      var row = result.first.toColumnMap();
+      expect(row['hnsw_ef_search'], '80');
+      expect(row['hnsw_scan_mem_multiplier'], '4');
+      // When not yet set, parameters return an empty string
+      expect(row['hnsw_iterative_scan'], isEmpty);
+    });
+  });
+
+  test(
+      'Given a transaction with runtime parameters containing null values '
```
SandPod: Suggestion: Clarify in the Given statement that a default configuration is supplied to the Serverpod instance.

The important part of this test is that we revert back to the Server configuration.

tests/serverpod_test_server/test_integration/database_operations/transactions/runtime_parameters_test.dart
```diff
@@ -0,0 +1,237 @@
...
+      'when setting parameters using transaction.setRuntimeParameters '
+      'then null values are reverted to default ', () async {
```
SandPod: Suggestion(NIT): `then null values are cleared`?

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/vector_query_test.dart
```diff
@@ -0,0 +1,181 @@
+import 'dart:math' as math;
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_server/test_util/test_serverpod.dart';
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+import 'package:test/test.dart';
+
+const totalVectors = 5000;
+const vectorDimension = 512;
+
+void main() async {
```
SandPod: Discussion/Suggestion: Couldn't we test this through the test framework? That would allow us to remove the teardown.

Also, we should add support for setting the Global runtime options in the test framework as well.

packages/serverpod/lib/src/database/database_pool_manager.dart
```diff
@@ -41,22 +42,30 @@ class DatabasePoolManager {
...
+            var setParametersStatements = runtimeParametersBuilder
+                ?.call(RuntimeParametersBuilder())
+                .map((p) => p.buildStatements(isLocal: false))
+                .expand((e) => e);
+            if (setParametersStatements != null) {
+              for (var statement in setParametersStatements) {
+                await connection.execute(statement);
+              }
+            }
+          },
```
SandPod: Question: Could we add `searchPaths` as a runtimeParameter and configure it through the same way?

They are basically the same thing, seems unecessary to use different solutions for them.

If no `searchPaths` is added through the passed in runtimeParametersBuilder, we check the config to see if one should be added.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -300,6 +311,7 @@ class Serverpod {
     this.httpOptionsResponseHeaders = _defaultHttpOptionsResponseHeaders,
     SecurityContextConfig? securityContextConfig,
     ExperimentalFeatures? experimentalFeatures,
+    this.runtimeParametersBuilder,
```
SandPod, line 319: Suggestion: Perhaps this should be scoped in some specific configuration for the database? This allows us to add more configurable optons without changing the interface here. Or should we wait until we have 2 options that are related and combine them at that time?

SandPod (commented):
Solution looks good now!

Had a couple of last questions.

We should also add support for configuring this in the test framework.

tests/serverpod_test_server/test_integration/database_operations/transactions/runtime_parameters_test.dart
marcelomendoncasoares: Done.

tests/serverpod_test_server/test_integration/database_operations/transactions/runtime_parameters_test.dart
marcelomendoncasoares: Done.

tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/vector_query_test.dart
marcelomendoncasoares: Done! Both changing the test to use the framework and exposing the runtime parameters builder on `withServerpod`.

packages/serverpod/lib/src/server/serverpod.dart
marcelomendoncasoares: I'd think more of your second suggestion, to wait until we have two related options and then combine. I have firstly looked through the existing options to see I could add in any of them, but I found better to add it separately. Maybe with more options in the future it gets easier to group them.

packages/serverpod/lib/src/database/database_pool_manager.dart
marcelomendoncasoares: Working on it.

marcelomendoncasoares: Done!

tests/serverpod_test_server/test_integration/test_tools/runtime_parameters_test.dart
```diff
@@ -0,0 +1,154 @@
...
+        for (var parameter in [
+          ...HnswIndexQueryOptions().options.keys,
+          ...IvfflatIndexQueryOptions().options.keys,
+        ]) {
+          var paramName = parameter.replaceAll('.', '_');
+          expect(allParameters[paramName], anyOf(isNull, isEmpty));
+        }
+      });
+    },
+  );
```
marcelomendoncasoares: It seems that this test is being affected by some leakage from other tests that are setting global runtime parameters first. The test runs with success in isolation, but is failing on the CI due to all parameters returning values - should not occur if a set is not previously called. I'm investigating the source for this possible leakage.

marcelomendoncasoares: Found the issue, and it was due to how PostgreSQL extension parameters work in conjunction with shared library loading. Before Postgre loads `pgvector` extension's shared library (`vector.so`), its parameters are treated as "placeholder parameters" and will return null if queried. After the extension is loaded, the parameters are considered "real parameters" and have non-null default values.

The library is only loaded in the session when some action using it is performed - making the `isNull` test to fail afterwards. To fix the test passing or failing unpredictably, I've added a `ensureVectorLoaded` extension method on the database and called on required tests. Then, all tests will ensure the presence of default values.

packages/serverpod/lib/src/database/database_pool_manager.dart
```diff
@@ -50,17 +50,22 @@ class DatabasePoolManager {
...
           onOpen: (connection) async {
-            var searchPaths =
-                config.searchPaths?.map((s) => encoder.convert(s)).join(',');
-            if (searchPaths != null && searchPaths.isNotEmpty) {
-              await connection.execute('SET search_path TO $searchPaths;');
+            var parameters =
+                runtimeParametersBuilder?.call(RuntimeParametersBuilder()) ??
+                    [];
+
+            if (!parameters.any((p) => p is SearchPathsConfig) &&
```
SandPod, line 57: Question: Wouldn't it be enough to simply check if a `SearchPathsConfig` exists, why do we need to inspect the fields of it?

packages/serverpod/lib/src/database/database_pool_manager.dart
```diff
@@ -43,8 +43,11 @@ class DatabasePoolManager {
   DatabasePoolManager(
     SerializationManagerServer serializationManager,
     RuntimeParametersListBuilder? runtimeParametersBuilder,
-    this.config,
-  ) : _poolSettings = pg.PoolSettings(
+    this.config, {
+    /// Whether the server is running in testing mode and all runtime parameters
+    /// should be set for the current transaction only.
+    bool testing = false,
```
SandPod: Discussion: Not a big fan of this as this leaks test framework knowledge into the `DatabasePoolManager`. Is there no other solution where we could set this? For example, use a different builder that always has the `p.buildStatement(isLocal: true)` in the test framework?

packages/serverpod/lib/src/database/database_pool_manager.dart
```diff
@@ -41,22 +42,35 @@ class DatabasePoolManager {
...
+            if (!parameters.any((p) => p is SearchPathsConfig) &&
+                config.searchPaths != null &&
+                config.searchPaths!.isNotEmpty) {
+              parameters.add(
+                SearchPathsConfig(searchPaths: config.searchPaths),
+              );
+            }
+
+            var setParametersStatements = parameters
+                .map((p) => p.buildStatements(isLocal: false))
```
SandPod, line 66: Question: If this is the only other place where we use the input parameter and we set it explicitly. Perhaps this should be a required param?

SandPod (approved):
Implementation looks good!

I have two last questions for you that I think we should resolve before merging.

packages/serverpod/lib/src/database/database_pool_manager.dart
SandPod: This was removed in a later commit.

packages/serverpod/lib/src/database/database_pool_manager.dart
marcelomendoncasoares: We don't actually inspect its fields, but the `config` fields to see if there is any `searchPaths` set on config level. The `parameters` here is a `List<RuntimeParameter>`, which can contain a `SearchPathsConfig`, that will take precedence over the `config.searchPaths`. Otherwise, if the paths are not set through runtime parameters and exists as config, we will add a `SearchPathsConfig` object and set it through the runtime parameters.

packages/serverpod/lib/src/database/database_pool_manager.dart
marcelomendoncasoares: Good idea, explicit will be better for this case. Done.

packages/serverpod/lib/src/database/database_pool_manager.dart
SandPod: Aaah, got it! I missread. Thanks for the explanation

PR #3628, opened 2025-06-10
feat(ci): Cancel previous pipeline run when a new commit is made to the same branch

tp:


tp:
Here you can see the previous run. All jobs got cancelled when I pushed a new commit in here: https://github.com/serverpod/serverpod/actions/runs/15556904590/job/43799733851?pr=3628

SandPod (approved):
LGTM! 👍

PR #3630, opened 2025-06-10
fix: Read rest of server configuration even if API config is missing.

exaby73:
Due to Mini not including configuration files by default, the config is never constructed and always throws when the API config is missing. This causes a default config to be used always. Since we are refactoring config to be the source of truth for CLI args in #3619, this behaviour is a blocker as we use config's runMode, role, etc. And since we previously threw, we would never successfully construct a config from CLI args and therefore, options like `--mode`, `--role` are ignored.

## Pre-launch Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of missing API server configuration by automatically using default values instead of causing errors.
  - Future call concurrency limit now defaults to 1 when not specified.

- **Tests**
  - Updated and added tests to verify the use of default configuration values when loading from an empty config map or missing API server settings.

### DIFF
--- a/packages/serverpod_shared/lib/src/config.dart
+++ b/packages/serverpod_shared/lib/src/config.dart
@@ -92,9 +92,6 @@ class ServerpodConfig {
   factory ServerpodConfig.defaultConfig() {
     return ServerpodConfig(
       apiServer: _defaultApiServer,
-      futureCall: const FutureCallConfig(
-        concurrencyLimit: FutureCallConfig.defaultFutureCallConcurrencyLimit,
-      ),
     );
   }
 
@@ -175,10 +172,7 @@ class ServerpodConfig {
             futureCallConfigJson,
             ServerpodConfigMap.futureCall,
           )
-        : const FutureCallConfig(
-            concurrencyLimit:
-                FutureCallConfig.defaultFutureCallConcurrencyLimit,
-          );
+        : const FutureCallConfig();
 
     var futureCallExecutionEnabled =
         _readIsFutureCallExecutionEnabled(configMap, environment);
@@ -471,7 +465,7 @@ class FutureCallConfig {
 
   /// Creates a new [FutureCallConfig].
   const FutureCallConfig({
-    this.concurrencyLimit,
+    this.concurrencyLimit = defaultFutureCallConcurrencyLimit,
     this.scanInterval =
         const Duration(milliseconds: defaultFutureCallScanIntervalMs),
   });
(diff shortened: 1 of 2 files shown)


packages/serverpod_shared/test/config_test.dart
```diff
@@ -11,15 +11,13 @@ void main() {
...
       () {
+    var config = ServerpodConfig.loadFromMap(runMode, serverId, passwords, {});
+    expect(config, isA<ServerpodConfig>());
     expect(
-      () => ServerpodConfig.loadFromMap(runMode, serverId, passwords, {}),
-      throwsA(isA<Exception>().having(
-        (e) => e.toString(),
-        'message',
-        equals('Serverpod API server configuration is missing.'),
-      )),
```
Isakdl, line 22: Comment: It doesn't seem like we are logging anything when this happens. I think it might be a little dangerous just picking the default without notifying the user, if you forgot to give a config it might be surprising that the server started with this default.

exaby73: This was already the case since we'd throw in `ServerpodConfig.loadFromMap` when we can't find an API config, which is caught in `ServerpodConfig.load` and then sets the config to `ServerpodConfig.defaultConfig()`

SandPod: Add a log in verbose mode when using the default configuration, this gives the developer a way to debug if their configuration is not reflected, but will not print anything for the normal use case.

SandPod: Decision: Skip verbose logging since we don't have access to the logger in the configuration currently and we want to switch configuration parsing to `cli_tools` at a later step.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -84,11 +91,9 @@ class ServerpodConfig {
   /// Creates a default bare bone configuration.
   factory ServerpodConfig.defaultConfig() {
     return ServerpodConfig(
-      apiServer: ServerConfig(
-        port: 8080,
-        publicHost: 'localhost',
-        publicPort: 8080,
-        publicScheme: 'http',
+      apiServer: _defaultApiServer,
+      futureCall: const FutureCallConfig(
```
SandPod: Suggestion: Remove future call configuration.

This should already be handled by the constructor of ServeprodConfig and should not need to be added here.

SandPod (changes requested):
LGTM!

But please remove the future call values from the default configuration.

There is no reason to modify this default.

PR #3631, opened 2025-06-11
feat(auth2): Implement flexible configuration as in API token in `serverpod_auth_session`

tp:


modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
```diff
@@ -72,14 +76,22 @@ abstract final class AuthSessions {
       return null;
     }
 
-    final maximumSessionLifetime = config.maximumSessionLifetime;
+    if (authSession.expiresAt != null &&
```
SandPod: Suggestion: Dereff to remove the potentially dangerous `!`

SandPod: Same for next check.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
```diff
@@ -72,14 +76,22 @@ abstract final class AuthSessions {
...
-        authSession.created
-            .add(maximumSessionLifetime)
-            .isBefore(DateTime.now())) {
+    if (authSession.expireAfterUnusedFor != null &&
+        authSession.lastUsed
+            .add(authSession.expireAfterUnusedFor!)
+            .isBefore(clock.now())) {
       session.log(
-        'Session has expired',
+        'Got sessoin which expired due to inactivity.',
```
SandPod: Suggestion: Spelling of `session`.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
```diff
@@ -58,7 +62,7 @@ abstract final class AuthSessions {
     }
     final secret = base64Decode(parts[2]);
 
-    final authSession = await AuthSession.db.findById(
+    var authSession = await AuthSession.db.findById(
```
SandPod, line 65: Suggestion/Question: Should we perhaps remove invalid sessions directly in the authentication handler? This would remove the need for developers to set up some form of purge of the database for now invalid sessions.

SandPod: Since we either way always update the row?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
```diff
@@ -125,6 +144,15 @@ abstract final class AuthSessions {
     required final UuidValue authUserId,
     required final String method,
     required final Set<Scope> scopes,
+
+    ///
```
SandPod: Suggestion(NIT): Remove empty comment

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
```diff
@@ -138,16 +166,53 @@ abstract final class AuthSessions {
...
         method: method,
       ),
+      transaction: transaction,
     );
 
     return _buildSessionKey(secret: secret, authSessionId: authSession.id!);
   }
 
+  /// List all sessions matching the given filters.
+  static Future<List<AuthSessionInfo>> listSessions(
```
SandPod, line 183: Question: Do we provide a similar method in the JWT case?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/models/auth_session.spy.yaml
```diff
@@ -23,7 +40,7 @@ fields:
   ### Per default uses 16 bytes of random data.
   sessionKeySalt: ByteData
 
-  ### The method of signing in this key was generated through.
+  ### The method through which this session was created.
   ###
-  ### This can be either email, a social login, etc.
+  ### This can be either an email or social login, a personal access token, service account etc.
   method: String
```
SandPod, line 46: Question: Aren't we missing `kind` or `label` now? Or are we only going to go with `method`?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_admin_test.dart
```diff
@@ -0,0 +1,153 @@
...
+import 'package:serverpod_auth_session_server/serverpod_auth_session_server.dart';
+import 'package:serverpod_auth_session_server/src/generated/protocol.dart';
+import 'package:test/test.dart';
+
+import '../test_utils.dart';
+import 'test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod(
+    'Given a session expiring in 1 day,',
```
SandPod: Suggestion: Restructure tests to better validate the given.

Shouldn't these be different givens rather than different whens? I would say that fixing the `clock` is also a given, not a when action.

- `Given a session expiring in 1 day`
- `Given an expired session`

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_authentication_handler_test.dart
```diff
@@ -0,0 +1,303 @@
...
+import 'package:serverpod_auth_session_server/src/business/auth_session_secrets.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+import 'package:test/test.dart';
+
+import '../test_utils.dart';
+import 'test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod(
+    'Given no session,',
```
SandPod, line 15: Suggestion: Clarify the naming here, perhaps `Given no auth session` since `session` is also the `session` used to access the database.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_authentication_handler_test.dart
```diff
@@ -0,0 +1,303 @@
...
+      sessionKey = await AuthSessions.createSession(
+        session,
+        authUserId: authUserId,
+        scopes: {},
+        method: 'test',
+      );
+    });
+
+    test(
+      'when calling `authenticationHandler`, then it returns an `AuthenticationInfo` for the user.',
```
SandPod: Suggestion: Please clarify what you are calling it with, `when calling "authenticationHandler" with related sessionKey` or something like that.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_authentication_handler_test.dart
SandPod: Apply this to the other tests as well.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_authentication_handler_test.dart
```diff
@@ -0,0 +1,303 @@
...
+
+        expect(
+          authInfo?.userUuid,
+          authUserId,
+        );
+      },
+    );
+
+    test(
+      'when calling `authenticationHandler` with the wrong secret, then it returns `null`.',
```
SandPod: Question: The term used is `sessionKey` right? If so, switch `secret` to `session key`.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_authentication_handler_test.dart
```diff
@@ -0,0 +1,303 @@
...
+            isNull,
+          );
+        },
+      );
+
+      test(
+        'when calling `authenticationHandler` after the pepper has been changed, then it returns `null`.',
+        () async {
+          AuthSessions.secretsTestOverride = AuthSessionSecrets(
+            sessionKeyHashPepper: 'another pepper',
```
SandPod: Question: Why is this grouped with the exparation tests? This should be with the other "simple" validation tests right?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_authentication_handler_test.dart
```diff
@@ -0,0 +1,303 @@
...
+
+        expect(
+          authInfo?.userUuid,
+          authUserId,
+        );
+      },
+    );
+
+    test(
+      'when calling `authenticationHandler` within the time limit and then afterwards, then it returns an `AuthenticationInfo` for the user (as the lifetime was extended).',
```
SandPod, line 251: Suggestion: Split this test.

If I understand this test correctly, we are validating that the time limit is updated when the authentication handler is called within the time window right?
We can validate that through a separate test in isolation instead by inspecting the assoticated row in the database after the authentication handler is called.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_destroy_test.dart
```diff
@@ -0,0 +1,188 @@
...
+import 'package:serverpod/protocol.dart';
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_session_server/serverpod_auth_session_server.dart';
+import 'package:test/test.dart';
+
+import '../test_utils.dart';
+import 'test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod('Given two sessions for a user,',
```
SandPod: Suggestion: Update description to reflect the test.

If I understand the setup correctly, we are only creating a single session? Is the description wrong or is the test wrong?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_destroy_test.dart
```diff
@@ -0,0 +1,188 @@
...
+      );
+      sessionKey2 = await AuthSessions.createSession(
+        session,
+        authUserId: authUserId,
+        scopes: {},
+        method: 'test',
+      );
+    });
+
+    test(
```
SandPod, line 134: Suggestion: Create a group of destroying the session and validate the conditions in separate tests.

This will give us a clear description if one of the cases stop working.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_list_test.dart
```diff
@@ -0,0 +1,122 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_session_server/serverpod_auth_session_server.dart';
+import 'package:test/test.dart';
+
+import '../test_utils.dart';
+import 'test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod('Given a session,', (final sessionBuilder, final endpoints) {
```
SandPod: Suggestion: Apply same naming fix here for `session`.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_list_test.dart
```diff
@@ -0,0 +1,122 @@
...
+    });
+
+    test(
+      'when calling `listsessions`, then it is returned.',
+      () async {
+        final sessions = await AuthSessions.listSessions(session);
+
+        expect(
+          sessions,
+          hasLength(1),
```
SandPod: Suggestion: Could we add some check that validates that this is actually the same session?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_list_test.dart
```diff
@@ -0,0 +1,122 @@
...
+        expect(sessions, isEmpty);
+      },
+    );
+
+    test(
+      'when calling `listSessions` for its `method`, then it is returned.',
+      () async {
+        final sessions = await AuthSessions.listSessions(
+          session,
+          method: 'test',
```
SandPod: Question: Will this give list all sessions with this method? In that case, is this maybe an admin method? Seems dangerous to expose in an endpoint.

SandPod (approved):
Nice, but a couple of comments for improvements.

I'll approve it but reach out if there is anything unclear.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
tp: > This would remove the need for developers to set up some form of purge of the database for now invalid sessions.

I suppose it would only help a bit with expired but used sessions (after which the client should not make any more requests), but would not help with expired but _unused_ ones, so ideally the developer would still connect the clean up functions.

> Since we either way always update the row?

We only update the row if access is successful. If it's expired or the secret is not valid (basically any kind of rejection) we leave it as is.


It's a good question though. At first I thought we want to keep them (especially for something like an API key use-case), where you can often "update" the expired key, getting a new one. (In practice that would require creating a new session here, and then deleting the old one.)

But if the invalid sessions would be cleaned-up faster, there's also other ways to handle this. E.g. a "service account" could always look for 1 (or more) sessions / API keys, and if there is none, just show a way to create one (instead of rotating).

So I kind of would've liked to retain the history, but then I feel in practice the admin function / Future call probably cleans it up in most cases anyway (or that would need to be extended to exclude certain `method`s).

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
tp: We do not yet.

Probably it would be good to have, to build a view like "active sessions" that many apps have.

I thought of it in here especially for the case of "API tokens" where you're not keeping the session alive by your own client(s) and thus want to see what's available.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/models/auth_session.spy.yaml
tp: I think `method` is sufficient here, as those other options (which were "kind" on the API token) are all after a "normal" login. (E.g. a logged in human user creates a personal access token for themselves, then I would just note that in the `method`.)

Maybe the name `method` could be improved/generalized, but I don't yet see a need for 2 fields.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_authentication_handler_test.dart
tp: The "secret" is only part of the session key, so this really refers to the secret being invalidated (but the session ID staying valid).

Will update to something like 

```suggestion
      'when calling `authenticationHandler` with the wrong secret inside the session key, then it returns `null`.',
```

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_list_test.dart
tp: The idea is to be able to list "ones own" session / API tokens here, so I see this used/exposed in a normal app for normal users.

It is of course important to pass the appropriate query, for example only showing the sessions for the user themselves, or for a "service account" which they control.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
tp: Now stored & checked via local variable.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
tp: Typo fixed.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
tp: removed empty line.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_admin_test.dart
tp: I don't think we can fix the `clock` as a given (e.g. wrapping the `test` function).

But we could of course manipulate the initial setup to add entries in the past, so that they would be expired "now" (when `test` does a simple read).

I quite like testing "the same" session once now and in the future, instead of having 2 entirely distinct groups that each create their own session.

But in the end I feel it's simple enough to wrap ones had around either way, so I'll just regroup them.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_authentication_handler_test.dart
tp: Updated all descriptions to refer to `auth session`s.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_authentication_handler_test.dart
tp: You're right, this just need any session so show that changing the pepper invalidates the session.

Moved it up.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_destroy_test.dart
tp: Yeah, this changed as there is just one user-level notification, so we don't have to have 2 sessions. (We do have 2 below to ensure that both get invalidated when all sessions for a user are destroyed.)

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_authentication_handler_test.dart
tp: > We can validate that through a separate test in isolation instead by inspecting the assoticated row in the database after the authentication handler is called.

We could, but I would prefer to stay on the package's API layer as much as possible, as in practice we're only interested in its behavior (does it successfully block requests are expiration) rather than its internals.

I can split it up though as discussed above, to have an "expired due to inactivity" session. Given that this leads to a bit convoluted setup (having to create a session with in the past etc.), I thought it's a bit simpler to understand the current approach 1) works now and 2) does not work 10 minutes later.

But that's just my personal preference, as above I'll optimize for a shorter `test` with a more detailed precondition.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
tp: Decision:
- Do not purge expired sessions, since we also do not create expiring sessions by default.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
tp: Decision:
- Add similar listing methods for JWT.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/models/auth_session.spy.yaml
tp: Decision:
- One field is fine for now.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_list_test.dart
tp: Decision:
- Move this "query" to admin function
- Expose a simple "list session" that works for a given auth user ID

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
tp: Will be done in separate PR.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_list_test.dart
tp: The "public" one has been restricted to return sessions for a user, and the flexible one is now scoped as an "admin" method.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_list_test.dart
tp: I plucked the ID from the DB and now validate against that. (Parsing the ID from the session key felt a bit dirty, though that would also work of course.)

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_admin_test.dart
tp: Updated the admin expiration tests as suggested.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/test/integration/auth_sessions_destroy_test.dart
tp: This has been adapted.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
```diff
@@ -138,16 +164,29 @@ abstract final class AuthSessions {
       session,
       AuthSession(
         authUserId: authUserId,
+        created: clock.now(),
```
tp, line 167: @SandPod Would you be up for using `clock` inside the models? I think that'd be a boon, as those implicit "now" dates are easily forgotten to make testable :)

tp: (Probably would have to re-export through `serverpod` then though, so not everyone has to rely on `clock` in their pubspec. Everybody has the dependency already anyway, just not directly.)

PR #3634, opened 2025-06-12
feat(auth2): Bundle auth user operations in `AuthUsers` class

tp:


modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/serverpod_auth_user_server.dart
```diff
@@ -1,5 +1,17 @@
+import 'package:meta/meta.dart';
+
+export 'src/business/auth_users.dart' show AuthUsers;
 export 'src/generated/endpoints.dart' show Endpoints;
-export 'src/generated/protocol.dart';
+// We don't want to expose these, but have to for the the generated database code in consuming packages.
+@visibleForTesting
```
tp: This works (due to our lint exclusions for generated files), but is obviously not the intended use for `@visibleForTesting`.

We might just export all the internals, and implore the user not to use `AuthUser` directly…

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/src/business/auth_users.dart
```diff
@@ -0,0 +1,122 @@
...
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+import 'package:serverpod_auth_user_server/src/generated/auth_user.dart';
+
+/// Management functions for auth users.
+abstract final class AuthUsers {
+  /// Retrieves an auth user.
+  ///
+  /// Throws an [AuthUserNotFoundException] in case no auth user is found for the ID.
+  static Future<AuthUserModel> getAuthUser(
```
tp: I wonder if the should drop the suffix, like

```suggestion
  static Future<AuthUserModel> get(
```

(For all these methods here. Just started out with the same pattern we already agreed upon for the user profiles back then.)

tp:
@coderabbitai review

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/test/test_utils.dart
```diff
@@ -129,10 +126,12 @@ Future<void> resetPassword(
...
-  await AuthUser.db.deleteWhere(
-    session,
-    where: (final t) => Constant.bool(true),
-  );
+  for (final authUser in await AuthUsers.listAuthUsers(session)) {
+    await AuthUsers.deleteAuthUser(
+      session,
+      authUserId: authUser.id,
+    );
+  }
```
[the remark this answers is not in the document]
tp: As we don't need this during normal operations, I would not add it.

If we need it for more than 1 test, we can add a special test helper.

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/src/business/auth_users.dart
```diff
@@ -0,0 +1,121 @@
...
+  /// If the auth user does not exist in the database, this completes successfully.
+  static Future<void> deleteAuthUser(
+    final Session session, {
+    required final UuidValue authUserId,
+    final Transaction? transaction,
+  }) async {
+    await AuthUser.db.deleteWhere(
+      session,
+      where: (final t) => t.id.equals(authUserId),
+    );
```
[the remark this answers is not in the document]
tp: Great catch!

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_client/lib/src/protocol/protocol.dart
```diff
@@ -30,10 +34,27 @@ class Protocol extends _i1.SerializationManager {
...
+      return (data != null ? _i3.AuthUserModel.fromJson(data) : null) as T;
+    }
+    if (t == _i1.getType<_i4.AuthUserNotFoundException?>()) {
+      return (data != null
+          ? _i4.AuthUserNotFoundException.fromJson(data)
+          : null) as T;
+    }
+    if (t == Set<String>) {
+      return (data as List).map((e) => deserialize<String>(e)).toSet() as T;
+    }
```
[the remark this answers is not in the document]
tp: You can ignore these files.

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/serverpod_auth_user_server.dart
tp: NOTE: If we change this back, we can remove the `meta` dependency again.

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/src/business/auth_users.dart
SandPod: Makes sense. `get` might be enough, same for the others then as well?

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/serverpod_auth_user_server.dart
SandPod: Hmm, this feels like a hack. I think we should remove it.

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/src/business/auth_users.dart
tp: Yes, then we can align all of them. Will do :)

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/serverpod_auth_user_server.dart
tp: Okay, will remove and export all DB-related things needed for the other packages.

tp: This has been removed again.

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/src/business/auth_users.dart
tp: Method names have been shortened.

modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
```diff
@@ -57,12 +57,11 @@ abstract class EmailAccountEndpoint extends Endpoint {
         transaction: transaction,
       );
 
-      final newUser = await AuthUser.db.insertRow(
+      final newUser = await AuthUsers.create(
```
SandPod, line 60: Question: How do we define the scopes for the email registration? Shouldn't this method take in scopes and pass them on to the AuthUser that is created?

SandPod: Potentially also the `blocked` information?

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/serverpod_auth_user_server.dart
```diff
@@ -1,5 +1,14 @@
+export 'src/business/auth_users.dart' show AuthUsers;
 export 'src/generated/endpoints.dart' show Endpoints;
-export 'src/generated/protocol.dart';
+export 'src/generated/protocol.dart'
```
SandPod: Question: Do we need to explicitly show these?

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/test/integration/auth_users_test.dart
```diff
@@ -0,0 +1,84 @@
...
+        () => AuthUsers.update(session, authUserId: const Uuid().v4obj()),
+        throwsA(isA<AuthUserNotFoundException>()),
+      );
+    });
+
+    test(
+        'when trying to delete a non-existent auth user by ID, then it succeeds.',
+        () async {
+      await AuthUsers.delete(session, authUserId: const Uuid().v4obj());
+    });
```
SandPod: Question: Is this really our intended behavior? As an administrator, would you not want to know if the auth could be removed or not?

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/serverpod_auth_user_server.dart
tp: Since this basically changed to "everything", I can also remove the explicit show here.

Generally I think it's quite nice though to avoid leakage.

modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
tp: I think `blocked` can be ignored, as no user should start out blocked. :)

But scopes are tricky, but since this is the public endpoint, I don't think it should be a parameter (which the developer overwriting would need to check then).

Maybe this is more of a concern for the invites, so that the scopes are pre-set in that case?

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/test/integration/auth_users_test.dart
tp: Maybe we could add a `bool`, I think in the end the most important part is that it's cleaned up.

Or would you go for an exception if not found?

modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
tp: Alternatively, we could add this as configuration (later?), if people really would want to have some default scopes.

tp: Decision:
- Leave this out for now.

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/test/integration/auth_users_test.dart
tp: Decision:
- throw the same "user not found" exception in this case

modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server/lib/src/business/auth_users.dart
```diff
@@ -0,0 +1,130 @@
...
+    final bool blocked = false,
+    final Transaction? transaction,
+  }) async {
+    final authUser = await AuthUser.db.insertRow(
+      session,
+      AuthUser(
+        created: clock.now(),
+        blocked: blocked,
+        scopeNames: scopes.map((final s) => s.name).nonNulls.toSet(),
+      ),
```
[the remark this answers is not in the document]
tp: `nonNulls` is actually available in stock Dart since 3.0.

PR #3635, opened 2025-06-12
feat(auth2): Add method to list authentication tokens in `AuthenticationTokens`

tp:
Aligning it with the new capabilities in `serverpod_auth_session`.

## Summary by CodeRabbit

- **New Features**
  - Introduced support for retrieving and listing authentication tokens associated with users, including detailed token metadata such as scopes and extra claims.
  - Added extension methods for easier access to token scopes and extra claims in a typed format.

- **Bug Fixes**
  - Improved consistency of token creation and update timestamps by using a unified time source.

- **Tests**
  - Added integration tests to verify listing and deletion of authentication tokens, including handling of expired tokens.
  - Enhanced test setup by using helper methods for user creation and deletion.

- **Chores**
  - Updated dependency management for the clock package.
  - Changed default UUID persistence strategy for refresh tokens.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/serverpod_auth_jwt_server.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/serverpod_auth_jwt_server.dart
@@ -6,3 +6,5 @@ export 'src/business/authentication_tokens_admin.dart'
     show AuthenticationTokensAdmin;
 export 'src/generated/endpoints.dart';
 export 'src/generated/protocol.dart' hide RefreshToken, RefreshTokenRepository;
+export 'src/util/authentication_token_info_extension.dart'
+    show AuthenticationTokenInfoExtension;
--- a/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
@@ -172,7 +172,7 @@ abstract final class AuthenticationTokens {
       refreshTokenRow.copyWith(
         rotatingSecretHash: ByteData.sublistView(newHash.hash),
         rotatingSecretSalt: ByteData.sublistView(newHash.salt),
-        lastUpdated: DateTime.now(),
+        lastUpdated: clock.now(),
       ),
       transaction: transaction,
     );
@@ -287,8 +287,8 @@ extension on Set<Scope> {
 
 extension on RefreshToken {
   bool get isExpired {
-    final oldestAcceptedRefreshTokenDate = DateTime.now()
-        .subtract(AuthenticationTokens.config.refreshTokenLifetime);
+    final oldestAcceptedRefreshTokenDate =
+        clock.now().subtract(AuthenticationTokens.config.refreshTokenLifetime);
 
     return lastUpdated.isBefore(oldestAcceptedRefreshTokenDate);
   }
--- a/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
@@ -244,7 +244,7 @@ abstract final class AuthenticationTokens {
     required final UuidValue authUserId,
     final Transaction? transaction,
   }) async {
-    return admin.findAuthenticationTokens(
+    return admin.listAuthenticationTokens(
       session,
       authUserId: authUserId,
       transaction: transaction,
--- a/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
@@ -1,6 +1,7 @@
 import 'dart:convert';
 import 'dart:typed_data';
 
+import 'package:clock/clock.dart';
 import 'package:dart_jsonwebtoken/dart_jsonwebtoken.dart';
 import 'package:meta/meta.dart';
 import 'package:serverpod/protocol.dart';
@@ -82,6 +83,8 @@ abstract final class AuthenticationTokens {
         rotatingSecretSalt: ByteData.sublistView(newHash.salt),
         scopeNames: scopes.names,
         extraClaims: extraClaims != null ? jsonEncode(extraClaims) : null,
+        created: clock.now(),
+        lastUpdated: clock.now(),
       ),
       transaction: transaction,
     );
@@ -232,6 +235,19 @@ abstract final class AuthenticationTokens {
     await session.messages.authenticationRevoked(
       refreshToken.authUserId,
       RevokedAuthenticationAuthId(authId: refreshTokenId.toString()),
+    );
+  }
+
+  /// List all authentication tokens belonging to the given [authUserId].
+  static Future<List<AuthenticationTokenInfo>> listAuthenticationTokens(
+    final Session session, {
+    required final UuidValue authUserId,
+    final Transaction? transaction,
+  }) async {
+    return admin.findAuthenticationTokens(
+      session,
+      authUserId: authUserId,
+      transaction: transaction,
     );
   }
 
--- a/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens_admin.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens_admin.dart
@@ -30,9 +30,18 @@ final class AuthenticationTokensAdmin {
     final Session session, {
     final UuidValue? authUserId,
     final Transaction? transaction,
+
+    /// How many items to return at maximum. Must be <= 1000.
     final int limit = 100,
     final int offset = 0,
   }) async {
+    if (limit <= 0 || limit > 1000) {
+      throw ArgumentError.value(limit, 'limit', 'Must be between 1 and 1000');
+    }
+    if (offset < 0) {
+      throw ArgumentError.value(offset, 'offset', 'Must be >= 0');
+    }
+
     final refreshTokens = await RefreshToken.db.find(
       session,
       where: (final t) => (authUserId != null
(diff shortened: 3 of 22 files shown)


tp:
@coderabbitai review

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_admin_test.dart
```diff
@@ -0,0 +1,144 @@
...
+      AuthenticationTokens.secretsTestOverride =
+          AuthenticationTokenSecretsMock()
+            ..setHs512Algorithm()
+            ..refreshTokenHashPepper = 'pepper123';
+    });
+
+    setUp(() async {
+      session = sessionBuilder.build();
+
+      final authUser = await AuthUser.db.insertRow(
```
tp: Need to use `AuthUsers` once https://github.com/serverpod/serverpod/pull/3634 is merged.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/serverpod_auth_jwt_server.dart
```diff
@@ -6,3 +6,5 @@ export 'src/business/authentication_tokens_admin.dart'
     show AuthenticationTokensAdmin;
 export 'src/generated/endpoints.dart';
 export 'src/generated/protocol.dart' hide RefreshToken, RefreshTokenRepository;
+export 'src/util/authentication_token_info_extension.dart'
+    show AuthenticationTokenInfoExtension;
```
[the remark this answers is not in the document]
tp: What makes you think that this would cause any issues / slowdowns in Dart code specifically?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/util/authentication_token_info_extension.dart
```diff
@@ -0,0 +1,22 @@
...
+
+  /// Returns the extra claims set on this authentication token.
+  Map<String, dynamic>? get extraClaims {
+    final extraClaims = this.extraClaims;
+    if (extraClaims == null) {
+      return null;
+    }
+
+    return (jsonDecode(extraClaims) as Map).cast<String, dynamic>();
+  }
```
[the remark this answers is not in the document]
tp: Due to `this.extraClaims` there will be no recursion, but of course it needs a different name to be easily discoverable.

I'll also need to add a test for it.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_admin_test.dart
tp: Remaining unneeded usages of `AuthUser.db` have been removed.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens_admin.dart
```diff
@@ -10,17 +11,46 @@ final class AuthenticationTokensAdmin {
...
+          authUserId: refreshToken.authUserId,
+          scopeNames: refreshToken.scopeNames,
+          created: refreshToken.created,
+          lastUpdated: refreshToken.lastUpdated,
+          extraClaims: refreshToken.extraClaims,
+        ),
+    ];
+
+    return authenticationTokenInfos;
+  }
```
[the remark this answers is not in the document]
tp: It's a neat suggestion, but I think we would need to sort ASC, so that this does not return any entries twice, if a new one is added (always with the current date) during iteration.

Also `t.created.desc()` does not exist, rather there is a `orderDescending` parameter.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens.dart
```diff
@@ -235,6 +238,19 @@ abstract final class AuthenticationTokens {
...
+    required final UuidValue authUserId,
+    final Transaction? transaction,
+  }) async {
+    return admin.findAuthenticationTokens(
+      session,
+      authUserId: authUserId,
+      transaction: transaction,
+    );
+  }
+
```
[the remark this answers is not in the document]
SandPod: @tp  - `limit` and `offset` are actually not bad suggestions for this type of interface. Question is if we could preserve some form of order of the results?

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/util/authentication_token_info_extension.dart
```diff
@@ -0,0 +1,22 @@
...
+
+/// Extensions for [AuthenticationTokenInfo].
+extension AuthenticationTokenInfoExtension on AuthenticationTokenInfo {
+  /// Returns the `Scope`s for this authentication token.
+  Set<Scope> get scopes {
+    return scopeNames.map(Scope.new).toSet();
+  }
+
+  /// Returns the extra claims set on this authentication token.
+  Map<String, dynamic>? get extraClaimsMap {
```
SandPod: Suggestion: Rename to `extraClaims`.

This is the getter the end user will use right? I think it is important that this has a resonable name. They add these under an `extraClaims` parameter to the JWT token right? Then I think this should match that parameter name.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
```diff
@@ -301,4 +307,41 @@ void main() {
...
+
+      await AuthenticationTokens.createTokens(session,
+          authUserId: authUserId, scopes: {});
+    });
+
+    tearDownAll(() {
+      AuthenticationTokens.secretsTestOverride = null;
+    });
+
+    test('when listing the tokens for that user, then it is returned.',
```
SandPod, line 337: Suggestion: Missing test that validates that we don't accidentally retrieve another users authentication tokens when we list authentication tokens for a user.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens_admin.dart
```diff
@@ -10,17 +11,46 @@ final class AuthenticationTokensAdmin {
...
 
     await RefreshToken.db.deleteWhere(
       session,
       where: (final t) => t.lastUpdated < oldestValidRefreshTokenDate,
       transaction: transaction,
     );
   }
+
+  /// List all authentication tokens matching the given filters.
+  Future<List<AuthenticationTokenInfo>> findAuthenticationTokens(
```
SandPod: Suggestion: Harmonize the naming of this method and the `list` method provided in the business layer.

Why would we use `find` here but `list` in the business layer? Keep it nice and tidy to give the feeling that it is all connected.

SandPod (commented):

Code rabbits suggestion with `limit` and `offset` is not bad. We should also always order the results from the `list` methods to make them consistent.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/util/authentication_token_info_extension.dart
tp: The problem is that `extraClaims` conflicts with the db/model field.

But I suppose we could rename that to `extraClaimsMap` or `extraClaimsJSON` instead, to get the “nice” name here.

In the end that’s probably better :)

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens_admin.dart
tp: I thought find was more aligned with the “open” query of this method and the `*Repository` classes we have now, whereas the non-admin method is specific to one user and really does list them.

But since this is not fully compatible with the DB `find*`, I guess `list` is a good choice for both then.

tp:
> Code rabbits suggestion with limit and offset is not bad. We should also always order the results from the list methods to make them consistent.

I saw that suggestion as well, but thought this is really not intended to for the expected use case here, and that we should rather have enough/more query options so that you only get the result you actually want to work upon.

So I wondered whether we should just expose the table-query callback, instead of re-building it?

But as a quick win I suppose we can start with limit/offset as that does not harm. Though as written in https://github.com/serverpod/serverpod/pull/3635#discussion_r2144691819 I really think we need to sort by `created ASC`, which will require an index (unless the v7 GUID type @marcelomendoncasoares added will solve this for us and we get a stable primary key and could thus sort by `ID ASC`). Will read up on that and update accordingly.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens_admin.dart
tp: Renamed this to `list*` as well.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/test/integration/authentication_tokens_test.dart
tp: added test for "normal" and "admin" that checks that listing works with 2 auth users as expected.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/util/authentication_token_info_extension.dart
tp: Renamed extra claims on the `*Info` class to `extraClaimsJSON` so we can use the ideal name for the extension.

tp:
Added sorted output using the V7 UUID (and verified that those tests would fail with V4 (fully random) ones).

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_tokens_admin.dart
```diff
@@ -10,17 +11,51 @@ final class AuthenticationTokensAdmin {
...
   }
+
+  /// List all authentication tokens matching the given filters.
+  Future<List<AuthenticationTokenInfo>> listAuthenticationTokens(
+    final Session session, {
+    final UuidValue? authUserId,
+    final Transaction? transaction,
+    final int limit = 100,
+    final int offset = 0,
+  }) async {
```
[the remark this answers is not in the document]
tp: Good idea, added.

YashMakan:
Hi! Quick question - does this JWT implementation support signing using private/public PEM keys? I'd love to use JWT from Serverpod across my other backend services instead of session based authentication with current auth token implementation. Thanks!

tp:
@YashMakan You that's possible. Answered on the original ticket @ https://github.com/serverpod/serverpod/issues/3526#issuecomment-2975263193

SandPod (approved):
LGTM! 🚀

PR #3636, opened 2025-06-12
fix(auth2): Add local dependency overrides for `serverpod_auth_email_client` and ensure the email packages are analyzed via the script

tp:


SandPod (approved):
Perhaps we should modify the run_tests_analyze script so that it finds all dart packages by itself? Kind of like the upgrade test script?

tp:
> Perhaps we should modify the run_tests_analyze script so that it finds all dart packages by itself? Kind of like the upgrade test script?

We definitely should refactor the scripts to either do the look ups themselves (with exclusions), or all work from a single list of configured packages.

Would you also want to introduce https://taskfile.dev/ with such a change, @SandPod ?

PR #3651, opened 2025-06-13
feat(auth2): Add integration tests for endpoints in `serverpod_new_auth_test`

tp:
Create instantiable endpoint subclasses in the test servers, and tests the outward API with `withServerpod` integration tests.


modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -32,7 +32,11 @@ abstract final class EmailAccounts {
       (final transaction) async {
         email = email.trim().toLowerCase();
 
-        if (await _hasTooManyFailedSignIns(session, email)) {
+        if (await _hasTooManyFailedSignIns(
+          session,
+          email,
+          transaction: transaction,
+        )) {
```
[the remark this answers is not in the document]
tp: This adaptation to `_hasTooManyFailedSignIns` was made to satisfy the `withServerpod` test helper (which runs the whole test in a transaction).

For us it does not really matter where the read is made.

The observation regarding the write (in its own transaction, which can not be rolled back) is spot-on. But as the code comment shows, this is done on purpose, so that a `throw` higher up in the stack will not rollback the transaction and thus remove the "block entry" from the database.

Since exception are widely used in "expected control flow" (e.g. invalid credentials), it's basically guaranteed that all "invalid" requests will end up in an exception that would roll back the "block row" from the database. Hence we write it externally. They will still show up in the transaction-using queries though.

modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
```diff
@@ -34,11 +34,18 @@ abstract class EmailAccountEndpoint extends Endpoint {
...
       password: password,
     );
+
+    if (result.result != EmailAccountRequestResult.accountRequestCreated) {
+      session.log(
+        'Failed to start account registration for $email, reason: ${result.result}',
+        level: LogLevel.debug,
+      );
+    }
   }
```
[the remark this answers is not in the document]
tp: We specficially do not want to expose any result in the expected case (account created, or email already exists) here, as to not leak infos about account existence to the outside.

We should probably clarify that with a comment here, that the logging is only for the server admin and should not be forwarded to the caller.

Can you prepare a suggestion?

packages/serverpod_test/lib/src/test_session_builder.dart
```diff
@@ -4,13 +4,13 @@ import 'test_serverpod.dart';
 
 /// An override for the authentication state in a test session.
 abstract class AuthenticationOverride {
-  /// Sets the session to be authenticated with the provided userId and scope.
+  /// Sets the session to be authenticated with the provided `userIdentifier` and `scope`.
   static AuthenticationOverride authenticationInfo(
-    int userId,
+    Object userIdentifier,
```
SandPod: Suggestion: Break this out into a separate PR with tests.

This is a feat in the test framework.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
```diff
@@ -32,7 +32,11 @@ abstract final class EmailAccounts {
       (final transaction) async {
         email = email.trim().toLowerCase();
 
-        if (await _hasTooManyFailedSignIns(session, email)) {
+        if (await _hasTooManyFailedSignIns(
```
SandPod: Question: Are we missing a test that validates this?

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/user_profile_endpoint_test.dart
```diff
@@ -0,0 +1,130 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_profile_server/serverpod_auth_profile_server.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+import 'package:test/test.dart';
+
+import 'test_tools/serverpod_test_tools.dart';
+
+void main() {
```
SandPod, line 8: Suggestion: Separate all the fixes applied in this PR out from the introduction of the test package.

These are different things and if we chose to revert the introduction of this test package on main, this should not be bundled up with required fixes in the different auth packages.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: Are you referring to the blocking of sign-in attempts?

Yes, there are various cases covered in https://github.com/serverpod/serverpod/blob/main/modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/test/integration/email_accounts_login_test.dart, grep for `tooManyAttempts`

Or can you think of a skipped one?

SandPod: It was a great catch that we had missed passing the transaction in here, I was just wondering why this wasn't detected in any of our tests?

packages/serverpod_test/lib/src/test_session_builder.dart
tp: This has been committed separately in https://github.com/serverpod/serverpod/pull/3672 now.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/user_profile_endpoint_test.dart
tp: The majority of the other fixes (missing exports) has been extracted here: https://github.com/serverpod/serverpod/pull/3673

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts.dart
tp: The test cases above worked just fine, using `Rollback.disabled` and thus able to create new transactions at will. And the timing was not an issue as the calls leading to the blocking were committed before the next checks were made.

But then it started throwing in the new test added here:
CleanShot 2025-06-19 at 09 57 01@2x

That must be because `login` uses the "transaction or savepoint" helper, after which case the "implicit" transaction from the test framework does not suffice anymore / leading to the duplicate. Without calling that "transaction or savepoint" helper it would also work just fine (due to the implicit one being used).

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -0,0 +1,411 @@
+import 'package:serverpod/serverpod.dart';
```
tp, line 1: @coderabbitai check this file, ensuring that all exposed methods of `EmailAccountEndpoint` are tested.

Double check that test descriptions match the implementation and check spelling.

tp: Since the endpoint method `startPasswordReset` just forwards to `EmailAccounts.startAccountCreation` internally, I don't think we need to add any tests for this and can rely on the tests from the `EmailAccounts` class, which cover that. (Also it's used indirectly a lot!)

For `finishPasswordReset` you're right. While this is executed and properly checked already in the `setUp` of the "Given an auth user with a changed password" test, it would be better to create an explicit test for it so any failure (e.g. in returning a valid session key) would lead to an even cleaner error with less other things going on around it.

tp: @coderabbitai Re-run the above check for the latest version of the file and also look into `user_profile_endpoint_test.dart`.

tp: That's correct. I don't think the user image tests, with their reliance on the public asset storage can be properly tested in these kinds of `withServerpod` integration tests.

We'll have to add them later in an E2E test, that can really call into a full running server to check that the image are actually loadable.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/util/test_tags.dart
```diff
@@ -0,0 +1,10 @@
+import 'package:serverpod_test/serverpod_test.dart';
+
+abstract final class TestTags {
```
SandPod, line 3: Question: I don't see any modifications to the script that updates pubspecs or to the github action that runs the tests.
Are these missing?

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/lib/src/endpoints/email_account_endpoint.dart
```diff
@@ -0,0 +1,5 @@
+import 'package:serverpod_auth_email_server/serverpod_auth_email_server.dart'
+    as email_account;
+
+/// Endpoint for email-based authentication.
+class EmailAccountEndpoint extends email_account.EmailAccountEndpoint {}
```
SandPod, line 5: Suggestion(For another PR): Since this will be the only part that the user will interract with, perhaps we should add documentation to the class that explains what will be exposed if the class is extended?

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -0,0 +1,527 @@
...
+import 'package:serverpod_auth_email_account_server/src/generated/email_account_password_reset_attempt.dart';
+import 'package:serverpod_auth_email_account_server/src/generated/email_account_password_reset_request_attempt.dart';
+import 'package:serverpod_auth_session_server/serverpod_auth_session_server.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+import 'package:test/test.dart';
+
+import '../util/test_tags.dart';
+import 'test_tools/serverpod_test_tools.dart';
+
+void main() {
```
SandPod, line 13: Question: Should we also add tests from a client perspective? Feels like it would be good to have, maybe that can be added once we have the session managers implemented.

SandPod (approved):
LGTM!
Just answer the questions I had around the CI.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
tp: Exactly, in the end we will have to have full E2E tests from the client package, as `withServerpod` still has some subtle (and not so subtle) differences.

Another example that came up about with CodeRabbit was also the profile image setting, which can only be properly be tested E2E (are the files really accessible and served?).

But I think this is already a great step forward, actually implementing the endpoint and getting to iron out some of the previously overlooked issues like the `transaction` passing.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/lib/src/endpoints/email_account_endpoint.dart
tp: Good idea, and/or link to the `abstract`/subclassing docs for anyone encountering the class to get a pointer of where to look.

(Though this will of course be extensively covered in the docs, that you need to subclass this to get it exposed on your server.)

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/util/test_tags.dart
tp: @SandPod Support for these test tags was added a while back in the `module_integration_tests`.

This is now already the 2nd (or 3rd?) "new auth" module that makes use of them.

We should probably move them to a shared package,. Where would you put them, given that they are only Serverpod/CI internal?

PR #3664, opened 2025-06-17
feat: Add support for google and firebase secrets to be defined with environment variables

### LINKED ISSUE #3556: Allow all service account keys to be passed in as environment variables
### Problem to Solve

Files such as `google_client_secret.json` and `firebase_service_account_key.json` must exist inside the config folder today. This means you either have to bake them into your image or mount a volume into your container when running your service in prod. 

In the case of Google they also already have many ways baked into the google_auth client for how to provide the credentials. 

### Proposal

I propose that the default Google auth client is used for authentication making this much more flexible. Any other key or secret that is located in a file should also be allowed to be provided with an env variable. This will allow for more different types of configurations and setups.

### Use Case

 -

### Alternatives

The alternatives are less secure and more regid.

### Additional context

_No response_

None

- [ ] I want to work on this

exaby73:
This PR aims to provide the ability to define Google and Firebase secrets, Google Client Secret and Firebase Service Account Key respectively, via environment variables

Closes #3556

## Pre-launch Checklist

## Breaking changes

No breaking changes


modules/serverpod_auth/serverpod_auth_server/lib/src/business/firebase_auth.dart
```diff
@@ -5,6 +5,13 @@ import 'package:serverpod_auth_server/serverpod_auth_server.dart';
 import 'package:serverpod_auth_server/src/firebase/exceptions/firebase_exception.dart';
 import 'package:serverpod_auth_server/src/firebase/firebase_auth_manager.dart';
 
+// Since we do not have access to [session.passwords] in this class, we need to
+// manually check for the password with the prefix "SERVERPOD_PASSWORD_" in the
+// environment variables. This prefix would be stripped by the PasswordManager
+// when the password is loaded.
+const _passwordKey =
+    'SERVERPOD_PASSWORD_serverpod_auth_firebaseServiceAccountKey';
```
Isakdl: Comment: I understand the reasoning for this syntax as we recently added support for this. However isn't it also creating a weird syntax when all old variables are following one type of syntax, and only this new one is using this new way? 

It will look a little strange I think and will not feel cohesive. I think it would make sense to support this syntax for the old variables too if we want to introduce this here. WDYT @SandPod?

modules/serverpod_auth/serverpod_auth_server/lib/src/business/firebase_auth.dart
```diff
@@ -20,13 +27,20 @@ class FirebaseAuth {
...
     Map<String, dynamic> firebaseServiceAccountJson;
     try {
-      firebaseServiceAccountJson = jsonDecode(
-        await File(AuthConfig.current.firebaseServiceAccountKeyJson)
-            .readAsString(),
-      );
+      if (Platform.environment.containsKey(_passwordKey)) {
+        firebaseServiceAccountJson = jsonDecode(
+          Platform.environment[_passwordKey]!,
+        );
```
Isakdl: Suggestion: Never use `!` as that will prevent the analyser from detecting invalid code, this is especially important when the code is changed and refactored. 

Instead extract the content into a variable and then null check it

```dart
final serviceKey = Platform.environment[_passwordKey];

if (serviceKey != null) { ... }
```

modules/serverpod_auth/serverpod_auth_server/lib/src/business/google_auth.dart
```diff
@@ -9,18 +9,25 @@ import 'package:serverpod/serverpod.dart';
...
 
-  static GoogleClientSecret? _loadClientSecret() {
+  static GoogleClientSecret? _loadClientSecret(Session session) {
     try {
-      var file = File(_configFilePath);
-      var jsonData = file.readAsStringSync();
+      late final String jsonData;
+      if (session.passwords.containsKey(_passwordKey)) {
+        jsonData = session.passwords[_passwordKey]!;
+      } else {
```
Isakdl: Suggestion: Same here

modules/serverpod_auth/serverpod_auth_server/lib/src/business/firebase_auth.dart
SandPod: After discussion with @Isakdl:
- Use underscore-based `namespace` for the module keys. (Bash prevents `dot` based namespace)
- Update documentation for the other passwords to use the new way of retrieving them.

modules/serverpod_auth/serverpod_auth_server/lib/src/business/firebase_auth.dart
```diff
@@ -20,13 +27,19 @@ class FirebaseAuth {
 
     Map<String, dynamic> firebaseServiceAccountJson;
     try {
-      firebaseServiceAccountJson = jsonDecode(
-        await File(AuthConfig.current.firebaseServiceAccountKeyJson)
-            .readAsString(),
-      );
+      final passwordFromEnv = Platform.environment[_passwordKey];
```
SandPod: Suggestion: This should be read from the passwords map.

This solution will not allow someone to place this in the passwords map from `serverpod_auth_firebaseServiceAccountKey`.

Why was a separate env read implemented here instead of using the passwords map?

modules/serverpod_auth/serverpod_auth_server/lib/src/business/google_auth.dart
```diff
@@ -9,18 +9,26 @@ import 'package:serverpod/serverpod.dart';
...
-  static final clientSecret = _loadClientSecret();
+  static GoogleClientSecret? clientSecret;
 
-  static GoogleClientSecret? _loadClientSecret() {
+  static GoogleClientSecret? _loadClientSecret(Session session) {
     try {
-      var file = File(_configFilePath);
-      var jsonData = file.readAsStringSync();
+      late final String jsonData;
+      final passwordFromEnv = Platform.environment[_passwordKey];
```
SandPod: Suggestion: This should be read from the passwords map.

This solution won't read the string from the env `SERVERPOD_PASSWORD_serverpod_auth_googleClientSecret`

modules/serverpod_auth/serverpod_auth_server/lib/src/business/firebase_auth.dart
```diff
@@ -1,10 +1,17 @@
...
 
+import 'package:serverpod/serverpod.dart';
 import 'package:serverpod_auth_server/serverpod_auth_server.dart';
 import 'package:serverpod_auth_server/src/firebase/exceptions/firebase_exception.dart';
 import 'package:serverpod_auth_server/src/firebase/firebase_auth_manager.dart';
 
+// Since we do not have access to [session.passwords] in this class, we need to
+// manually check for the password with the prefix "SERVERPOD_PASSWORD_" in the
+// environment variables. This prefix would be stripped by the PasswordManager
+// when the password is loaded.
```
SandPod: Suggestion: This comment needs to be updated.

modules/serverpod_auth/serverpod_auth_server/lib/src/business/firebase_auth.dart
```diff
@@ -20,13 +23,19 @@ class FirebaseAuth {
...
+        firebaseServiceAccountJson = jsonDecode(
+          await File(AuthConfig.current.firebaseServiceAccountKeyJson)
+              .readAsString(),
+        );
+      }
     } catch (e) {
       throw FirebaseInitException(
-        'Failed to load "firebase_service_account_key.json" file: $e',
+        'Failed to load "firebase_service_account_key.json" file or password '
+        '${_passwordKey.substring('SERVERPOD_PASSWORD_'.length)}: $e',
```
SandPod: Suggestion: This will give the wrong string.

Issue #3450, opened 2025-04-22
User migration - Authentication Module 2.0


tp:
We want to support user migration from `serverpod_auth` into the new structure.

As the new structure is more modular, we need to handle the profile info (`serverpod_auth_profile`) optionally, as well as support for each authentication mechanism (email, social, Firebase).

Since every existing feature is supported by some combination of modules in the new world, there should be a 100% data transfer.

For now the idea would be to import the data in the new system, but still leave it in the old one as well, so existing endpoints there continue to work.

As migrations can not happen “at once”, we want to support migrating on “first use” through the new system, e.g. when a user attempts to log in via email/password, we’d first check the new system whether the user is known, and if not, try importing one from the old system.

To also migrate accounts which do not have active use, we want to provide a `Future`-call which runs in the background and migrates batches of accounts.   
Here it is to be checked whether any of the old password storage configurations will cause issues in fully automatically migrating (as the hashing function changed and might be dependent on runtime configuration). In that case we might need to import some of those legacy users (which did not see any usage in the new system) without their email password, and thus force them into a reset later on.

Once every user has been migrated (actively or through the background job) and all clients have been updated to use the new functions, the developer can remove the old `serverpod_auth` module from their project.

PR #3666, opened 2025-06-17
feat(auth2): Add module `serverpod_auth_migration` with support for email migration

### LINKED ISSUE #3648: Email user migration
Import existing email based accounts from `serverpod_auth` into `serverpod_auth_email_account`.

tp:
Adds a new migration package that can be used to import existing users ("profiles") and authentication methods (now called "providers") from the current `serverpod_auth` package to the newly created `serverpod_auth_*` ones.

For starters this supports migrating profiles and their associated email based accounts.

This is achieved by providing 2 methods:
- migrateOnLogin with `email` and `password`
- migrateWithoutPassword with just the `email`

`migrateOnLogin` is supposed to be called before a login takes place (e.g. in the exposed endpoint's `login` method), and would migrate any existing email-based user from the old system into the new one, if one could be found and the credentials match. In case no user is found or the credentials are not valid (or the user is blocked) this does not do anything (apart from increasing the rate-limit…).

`migrateWithoutPassword` is meant to be called in the background, e.g. by a Future-call, to migrate accounts which are not actively used. Since the password hashing changed in the new package and we don't have access to the plain-text password in this method, this can not migrate the password, but instead leaves that empty, so that all password-based logins for these accounts would fail, unless `migrateOnLogin` is called for such an account, in which case the missing password will be transferred from the old system.

To support these migrations, 3 new methods had to be added to `EmailAccountsAdmin`:

- `findAccount`: A simple look-up that can tell whether a given `email` is already known in the new system. In this case no further attempt at migrating the user is made (unless the password is still missing, in which case that will be backfilled by `migrateOnLogin`).
- `createEmailAuthentication`: A low-level helper to "forcefully" create an email/password based authentication, whereby the email is automatically trusted (i.e. the verification flow is skipped).
- `setPassword`: Helper to directly set the password on an account. This is used by `migrateOnLogin` to backfill a missing password.

For each successful migration a `MigratedUser` is written to the database, which is a mapping from the old `UserInfo` to the new `AuthUser` (with full foreign key support).

This will become useful when adding support for further providers in this new module, so we can look-up whether a user has already been migrated and then only add the next provider as a further authentication method without attempting to migrate the profile data again.

Relates to #3450

Closes #3648

## Summary by CodeRabbit

- **New Features**
  - Introduced Serverpod authentication migration modules with email-based migration tools supporting profile import and blocked user preservation.
  - Added configuration files, Docker Compose setup for testing, and secure password management.
  - Provided comprehensive documentation, changelogs, and README files for client and server modules.
  - Added a new data model to track migrated users and helper classes for migration operations.

- **Bug Fixes**
  - Improved handling of migration scenarios including existing, blocked, and previously migrated users.

- **Documentation**
  - Delivered detailed guides, changelogs, and configuration instructions for authentication migration modules.

- **Tests**
  - Added extensive integration tests ensuring migration reliability and data integrity.

- **Chores**
  - Enhanced scripts and workflows to support code generation, dependency updates, and CI for the new modules.


modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/README.md
```diff
@@ -0,0 +1,9 @@
+![Serverpod banner](https://github.com/serverpod/serverpod/raw/main/misc/images/github-header.webp)
+
+# Serverpod
+This package is a core part of Serverpod. For documentation, visit: [https://docs.serverpod.dev](https://docs.serverpod.dev).
+
+## What is Serverpod?
+Serverpod is an open-source, scalable app server, written in Dart for the Flutter community. Check it out!
+
```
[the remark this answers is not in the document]
tp: That would be great, but I think it'll just get overwritten. We'll add proper docs before rolling out this feature.

modules/serverpod_auth/serverpod_auth_server/lib/src/business/email_auth.dart
```diff
@@ -655,11 +655,12 @@ class Emails {
...
 
   static Future<void> _logFailedSignIn(Session session, String email) async {
-    session as MethodCallSession;
     var failedSignIn = EmailFailedSignIn(
       email: email,
       time: DateTime.now(),
-      ipAddress: session.httpRequest.remoteIpAddress,
+      ipAddress: session is MethodCallSession
+          ? session.httpRequest.remoteIpAddress
+          : '',
```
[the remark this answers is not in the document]
tp: That's not true, it's defined as `String get remoteIpAddress`.

So this is fine and matches what we do in other places (to get a fallback in testing).

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/config/passwords.yaml
```diff
@@ -0,0 +1,15 @@
...
+# different run configurations. If you want the same password for any
+# configuration used, place it under `shared`.
+#
+# Note that this file should not be under version control. Store it in a safe
+# place.
+
+test:
+  database: 'DB_TEST_PASSWORD'
+  redis: 'REDIS_TEST_PASSWORD'
+  serverpod_auth_email_account.passwordHashPepper: 'emailaccountpepper123'
```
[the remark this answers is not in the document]
tp: This works fine in our parser currently, though we might want to switch to nested declarations for these in the future.

Can you check whether our parser has a test for keys with `.` (which should be treated as plain strings for now), and if not suggest one to add?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
```diff
@@ -77,4 +80,98 @@ final class EmailAccountsAdmin {
...
+    final account = await EmailAccount.db.insertRow(
+      session,
+      EmailAccount(
+        authUserId: authUserId,
+        email: email,
+        passwordHash: ByteData.sublistView(passwordHash.hash),
+        passwordSalt: ByteData.sublistView(passwordHash.salt),
+      ),
+      transaction: transaction,
+    );
```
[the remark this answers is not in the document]
tp: I think relying on the DB exception for now is fine (since we do have a unique index on the `email` column).

I don't see a benefit of checking beforehand, as it could still fail if there are 2 operations in parallel (as these run in longer-running transactions).

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
```diff
@@ -77,4 +80,98 @@ final class EmailAccountsAdmin {
...
+  }) async {
+    email = email.toLowerCase();
+
+    var account = (await EmailAccount.db.find(
+      session,
+      where: (final t) => t.email.equals(email),
+      transaction: transaction,
+    ))
+        .single;
+
```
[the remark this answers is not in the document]
tp: Agreed, maybe this is too hard to figure out and we should throw a custom error instead.

We just have to be careful to not use such an error in the non-admin cases, as we generally do not want to expose/leak the existence of an email-based account (to outside clients).

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_client/CHANGELOG.md
```diff
@@ -0,0 +1,650 @@
+## 2.8.0
+- feat: Adds `DatabaseUtils` with support for executing code blocks in transactions or savepoints.
+- feat: Adds support for `Vector` type, with database `HNSW` and `IVFFLAT` indexing, in models and endpoints. ([@marcelomendoncasoares](https://github.com/marcelomendoncasoares))
+- feat: Adds support for SSL connections with Redis. ([@remonh87](https://github.com/remonh87))
+- fix: Fixes an issue where configured `user` would not be used when connecting to Redis. ([@remonh87](https://github.com/remonh87))
+- fix: Improves error message when failing to connect to database during startup.
+
```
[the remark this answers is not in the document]
tp: This is auto-generated and can not be changed at this point.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
```diff
@@ -0,0 +1,197 @@
...
+        // TODO: Remove localhost exclusion, that's just an issue in test
+        profileImageUrl.host != 'localhost') {
+      // TODO: We should probably just retain given URL (via a parameter, or `set*` and `load*` variants of the method?), to not introduce a long-running fetch in the migration (which could also easily fail)
+      await UserProfiles.setUserImageFromUrl(
+        session,
+        authUser.id,
+        profileImageUrl,
+        transaction: transaction,
+      );
+    }
```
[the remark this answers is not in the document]
tp: As written in my `TODO`, I will adapt this to just "copy over" the original image URL, so we don't introduce a potential error here when failing to load the image for whatever reason.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_client/CHANGELOG.md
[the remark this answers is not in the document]
tp: Check the `util/update_pubspecs` script to see how most (if not all) README and CHANGELOG files get copied from a single source.

You can safely ignore any changes to README.md or CHANGELOG.md files made by that script.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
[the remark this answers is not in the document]
tp: A specific exception has been implemented.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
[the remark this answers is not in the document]
tp: Copying the original image URL did not turn out to work, as we also need to store the storage ("bucket") and relative path in addition to the public URL in the database.

This is needed to later remove the image (if the whole profile gets deleted).

So for now we still do attempt to download the image by URL (which should work just fine in production, and is just a local-host issue during testing with `withServerpod`, where no HTTP file server is running).
In case the fetch fails, we will log an error, but overall still proceed with the migration.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/config/passwords.yaml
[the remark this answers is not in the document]
tp: Turns out there are more with `.` in configuration field name.

As we do want to allow the user to overwrite all configuration keys & passwords via ENV variables, we have to switch away from name containing `.` anyway (as these are not supported in BASH it seems).

So we'll rename all these `.`-containing keys in a follow up PR.

.github/workflows/dart-tests.yaml
```diff
@@ -263,7 +263,7 @@ jobs:
           ./wait-for-it.sh localhost:9091 -t 60 -- echo "### Redis is UP at :9091"
         working-directory: tests/docker/tests_integration
       - name: Run integration tests (concurrently)
-        run: dart test -t integration -x concurrency_one --reporter=failures-only
+        run: dart test -t integration -x concurrency_one --reporter=failures-only || [ $? -eq 79 ]
```
SandPod: Comment on PR: Please add this as it's own commit in the future as it does not related to the implementation in `Implement email account migration "pre login" helper` (a9c7bb0d058e362f183f79c0996e2e6835c72312)

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
```diff
@@ -77,4 +80,53 @@ final class EmailAccountsAdmin {
       transaction: transaction,
     );
   }
+
+  /// Checks whether an email authentication exists for the given email address.
+  Future<bool> hasAccount(
```
SandPod: Comment on PR: Please include the test for these methods in the same commit in the future.

This makes it easy to validate that the methods are tested when reviewing commit by commit (a9c7bb0d058e362f183f79c0996e2e6835c72312)

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/auth_migration_email.dart
```diff
@@ -0,0 +1,110 @@
...
+  }) async {
+    if (await EmailAccounts.admin.hasAccount(
+      session,
+      email: email,
+      transaction: transaction,
+    )) {
+      return;
+    }
+
+    final authenticationResult = await Emails.authenticate(
```
SandPod: Question: You mentioned blocked users in the description. I'm guessing we actually have a way to migrate these users to the new system since the password might be correct.

Is it worth making sure these are migrated as well?

Also; Do we have tests that validate the blocked user behavior?

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/auth_migration_email.dart
```diff
@@ -0,0 +1,110 @@
...
+
+        if (migratedUser == null) {
+          final authUser = await AuthUsers.create(
+            session,
+            blocked: false,
+            scopes: userInfo.scopes,
+            transaction: transaction,
+          );
+
+          await UserProfiles.createUserProfile(
```
SandPod: Question: User profiles are optional in the new system. Should creating a User profile be a configurable option? On by default but possible to switch off?

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/auth_migration_email.dart
```diff
@@ -0,0 +1,110 @@
...
+abstract final class AuthMigrationEmail {
+  /// Attempts to migrate the user and their email authentication to the new auth module.
+  ///
+  /// In case the email is already registered in the new system or the credetnais are not known in the old system, this method does not do anything.
+  ///
+  /// If the credentials are known to the old system but not yet in the new one, then the user's email authentication as well as their profile are migrated.
+  /// If the credentials are valid in the old system and the user has already been migrated (for another authentication method), then only the email authentication will be added for that user.
+  ///
+  /// This method is intended to be used before the login is attempted in the new module, so it should be added in the `login` endpoint as the first step.
+  static void migrateOnLogin(
```
SandPod: Comment on PR: Would be great to have tests included in the commit that adds this functionality (a9c7bb0d058e362f183f79c0996e2e6835c72312).

Having this as a separate commit from the introduction of the admin methods would futher have improved the revieability of the PR as that would have given the reviewer a small scope of business logic with validation tests in one view.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
```diff
@@ -107,14 +120,16 @@ final class EmailAccountsAdmin {
...
+    /// If `null`, no password will be set for this user
+    required final String? password,
     final Transaction? transaction,
   }) async {
     email = email.toLowerCase();
 
-    final passwordHash = await EmailAccountSecretHash.createHash(
-      value: password,
-    );
+    final passwordHash = password != null
```
SandPod: Question(Security concern): Would it ever be possible to a user to try and match an empty hash?

Given all our configuration options that we provide, is it possible that an empty Uint8List might be supplied in some way as hash? If so, we now create a default password and might enable the possibility for brute force attacks towards inactive accounts.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
```diff
@@ -82,7 +82,12 @@ final class EmailAccountsAdmin {
   }
 
   /// Checks whether an email authentication exists for the given email address.
-  Future<bool> hasAccount(
+  Future<
```
SandPod: Comment on review: Stopped reviewing commit by commit from Add password-less migration option and the admin ability to set a new.. (8267fb855c95df1f1507f8a5dd61c25556ab3812), since the new commits are now patching old behavior.

This makes me review the same code multiple times with possibly outdated comments.

Please introduce methods with tests in a single commit once they are stabalized. I think iterations with CodeRabbit should be considered development iterations.

SandPod (changes requested):
This PR is missing test for newly added business logic (for example the adming methods in the auth email account package.

Before doing a full review; Please restructure all commits after 5e47d8565ce0e73d1e86c47c038214416984b3fa so that we get a clean history that introduces some business logic together with their tests.

The endpoint methods can also be introduced in separate commits with test that validate their behavior.

This will greatly increase both the quality and the speed of a review of this PR.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/auth_migration_email.dart
tp: Blocked users can be migrated in the background, just not during login.

This is just because we use the normal "login" flow from `serverpod_auth`, and that errs for a blocked user (not giving out the user info).

I think that's fine, as in that rare case the login would fail anyway right after. It's only important that the developer calls `migrateWithoutPassword` before e.g. a account creation, so you could not create a new user for a previously blocked email.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/auth_migration_email.dart
tp: Sure, we can make this configurable.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
tp: The approach specifically does not use any default/random password for the user, but rather just stores the empty salt/hash in the DB.

Since we are employing a fixed-length hashing function (with Argon ID's `desiredKeyLength`), the output can never be the same "empty bytes list" that we pass in.

We could of course add a "fast return" if the input `hash` given to the validation is empty, in which case any further computation is wasted.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
tp: > I think iterations with CodeRabbit should be considered development iterations.

Ok, got it. Will rewrite the history accordingly.

.github/workflows/dart-tests.yaml
tp: split out.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
tp: implementation (incl. the rename of this method) and the tests have been merged into 1 commit.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/auth_migration_email.dart
tp: Added 2 tests here:
- migrate on login: has no effect for blocked users
- migrate without password: imports the user, but retains the `blocked` flag on the new `AuthUser`

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
tp: Added the early return in the "verify hash" function as well.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/auth_migration_email.dart
tp: Commits are not split in "extension of email admin" and "implementation of email migration".

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/auth_migration_email.dart
tp: Made the profile import configurable and added tests for each migration option.

tp:
@coderabbitai full review

modules/serverpod_auth/serverpod_auth_server/lib/src/business/email_auth.dart
```diff
@@ -655,11 +655,12 @@ class Emails {
   }
 
   static Future<void> _logFailedSignIn(Session session, String email) async {
-    session as MethodCallSession;
     var failedSignIn = EmailFailedSignIn(
       email: email,
       time: DateTime.now(),
```
SandPod: Question: This could be expressed directly in the model, right?

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/test/integration/email_accounts_admin_test.dart
```diff
@@ -48,8 +49,148 @@ void main() {
...
+              session,
+              email: email,
+              password: newPassword,
+            ),
+            authUserId,
+          );
+        },
+      );
+    },
+    rollbackDatabase: RollbackDatabase.disabled,
```
SandPod: Question: Why do we need to disable database rollbacks for these tests?

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/config/passwords.yaml
```diff
@@ -12,3 +12,4 @@
 test:
   database: 'DB_TEST_PASSWORD'
   redis: 'REDIS_TEST_PASSWORD'
+  serverpod_auth_email_account.passwordHashPepper: 'emailaccountpepper123'
```
SandPod: Question: Do we have an issue tracking the rename of all these secrets?

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
```diff
@@ -0,0 +1,204 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_email_account_server/serverpod_auth_email_account_server.dart';
+import 'package:serverpod_auth_migration_server/serverpod_auth_migration_server.dart';
+import 'package:serverpod_auth_profile_server/serverpod_auth_profile_server.dart';
+import 'package:serverpod_auth_server/serverpod_auth_server.dart';
+import 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart';
+
+/// Helper functions to migration email accounts from `serverpod_auth` to `serverpod_auth_email_account`.
+abstract final class AuthMigrationEmail {
+  /// Attempts to migrate the user and their email authentication to the new auth module.
```
SandPod: Suggestion: Change all comments in the PR so that they break on 80 characters.

I think it is fine to deviate from this in tests, but for the production code, we should try to be consistent.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
```diff
@@ -0,0 +1,204 @@
...
+        email: email,
+        password: password,
+        transaction: transaction,
+      );
+
+      return;
+    }
+
+    // Remove the authentication that was implicitly created upon successful login in the old module
+    await AuthKey.db.deleteWhere(
```
SandPod: Question: Isn't this always required after the step on line 37. In the case that we just need to migrate the password we would be leaking an `AuthKey` right?

We should have a test that validates that we never leak an auth key regardless of what path we take out of this method.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
```diff
@@ -0,0 +1,366 @@
...
+      late Session session;
+
+      setUp(() {
+        session = sessionBuilder.build();
+      });
+
+      test(
+        'when attempting to run `migrateOnLogin` for a non-existent account, then it completes without error.',
+        () async {
+          await AuthMigrationEmail.migrateOnLogin(
```
SandPod: Suggestion: This test does not have any expect.

Please add an expect for the test. I think `completes` is fine.

Same for the all other tests without expect.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
```diff
@@ -0,0 +1,204 @@
...
+  /// This method is intended to be used before the login is attempted in the new module, so it should be added in the `login` endpoint as the first step.
+  static Future<void> migrateOnLogin(
+    final Session session, {
+    required String email,
+    required final String password,
+    final Transaction? transaction,
+  }) async {
+    email = email.toLowerCase();
+
+    final emailAccountInfo = await EmailAccounts.admin.findAccount(
```
SandPod: Suggestion(style): For better code clarity, would it make sense to import the new auth through a `new` name and the old one through an `old` name?

Giving: `new.EmailAccounts.admin...` and `old.Emails.authenticate...`? Would be good to apply this to the tests as well.

SandPod: Or perhaps `legacy` instead of `old`?

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
```diff
@@ -0,0 +1,366 @@
...
+            session,
+            email: '404@serverpod.dev',
+          );
+        },
+      );
+    },
+  );
+
+  withServerpod(
+    'Given a `serverpod_auth` email-based user account,',
```
SandPod: Suggestion: Could you add `legacy` to the name so that it is clear that we are talking about an account in the old system we are migrating away from.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
```diff
@@ -0,0 +1,366 @@
...
+            session,
+            email: email,
+          );
+        },
+      );
+    },
+  );
+
+  withServerpod(
+    'Given a `serverpod_auth` email-based user account with has been migrated using `migrateOnLogin`,',
```
SandPod: Suggestion: Change `account with has been migrated using` to `account that has been migrated using`

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
```diff
@@ -0,0 +1,366 @@
...
+
+        authUserId = (await EmailAccounts.admin.findAccount(
+          session,
+          email: email,
+        ))!
+            .authUserId;
+      });
+
+      test(
+        'when the migration is done, then not session have been created in the old system.',
```
SandPod: Suggestion(NIT): Change `...is done, then not session have been...` to `...is done, then no session has been...`

SandPod: Same for the other place this test has been added.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
```diff
@@ -0,0 +1,366 @@
...
+              password: password,
+              transaction: session.transaction,
+            ),
+            isNotNull,
+          );
+        },
+      );
+
+      test(
+        'when running `migrateOnLogin` with a wrong password, then this password can be used for the login.',
```
SandPod: Suggestion: This description does not match the test, and the description seems like it would be a security risk if it was applied.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
```diff
@@ -0,0 +1,366 @@
...
+        () async {
+          await AuthMigrationEmail.migrateWithoutPassword(
+            session,
+            email: email,
+          );
+        },
+      );
+
+      test(
+        'when attempting to log into the new system with the credentials, then that fails (because the password has not been set).',
```
SandPod, line 549: Suggestion: Remove test.

This test is a duplicate of the test on line 267. But the description here is better.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
```diff
@@ -0,0 +1,204 @@
...
+    await EmailAccounts.admin.createEmailAuthentication(
+      session,
+      authUserId: migratedUser.newAuthUserId,
+      email: email,
+      password: null,
+      transaction: transaction,
+    );
+  }
+
+  static Future<MigratedUser> _migrateUserIfNeeded(
```
SandPod, line 180: Question: Shouldn't all calls to this method use a transaction so that we don't end up with a migrated user when the next step in the outer methods fails?

Or would we be in an OK state if this fails?

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
```diff
@@ -165,9 +168,31 @@ abstract final class AuthMigrationEmail {
       transaction: transaction,
     );
 
+    if (config.importProfile) {
```
SandPod, line 202: Comment on PR: This is new functionality that should either have been included with the original implementation of this method (since it is implemented in this PR) or added in an isolated commit with tests (not in the last commit with the description `Code generation and migration for new implementation`, e35cf54cc666fd79235e2ad5dc26d3575b62d410).

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email_config.dart
```diff
@@ -0,0 +1,12 @@
+/// Configuration options for the email account migration.
+class AuthMigrationEmailConfig {
+  /// Whether to import the `serverpod_auth` `UserInfo` into a `serverpod_auth_profile` `UserProfile`.
+  ///
+  /// Defaults to `true`.
+  final bool importProfile;
```
SandPod, line 7: Suggestion: Make this a callback instead of a bool with the migration to `serverpod_auth_profile` as the default implementation.

That would allow anyone to hook in their new user profile if required.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
```diff
@@ -165,9 +168,31 @@ abstract final class AuthMigrationEmail {
...
+      await _importProfile(
```
SandPod, line 203: Suggestion: This is missing tests.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
tp: Oh nice, I didn't know about https://api.flutter.dev/flutter/package-matcher_expect/completes.html

I thought we had agreed on another PR to not use `isNull` for `Future<void>` tests.

But now with `completes on the radar, I'll rewrite them to `expect(…, completes);` which might be easiest to read through.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
tp: Yeah, I like that suggestion for the code in here, as the reader might not be immediately be aware that `Emails` is from `serverpod_auth` and `EmailAccounts` from `serverpod_email_account`.

(Generally I would try to avoid naming imports though, so it's not something I would burden the developer with or suggest in the migration docs by default. But let's see how those code-snippets read in the end (they might already require a named import as one probably wants to re-use the same endpoint name, or we explicitly use `Base` on the `abstract` ones to avoid this issue.)

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
tp:
```suggestion
        'when running `migrateOnLogin` with a wrong password, then this password can not be used for the login.',
```

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
```diff
@@ -0,0 +1,366 @@
...
+              // This is the user's password in the old system, but since it has not been set during the import,
+              // it does not work on the login in the new system (without importing it first).
+              password: password,
+              transaction: session.transaction,
+            ),
+            throwsA(isA<Exception>()),
+          );
+        },
+      );
+
```
tp:
```suggestion
```

modules/serverpod_auth/serverpod_auth_server/lib/src/business/email_auth.dart
tp: Setting the time to "now"?

Sure, we could add that default there, though I wanted to touch the legacy code as little as possible.

SandPod: Sorry, is this part of the old auth? Then don't touch it. I think you approach was valid in that case!

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/config/passwords.yaml
tp: Created https://github.com/serverpod/serverpod/issues/3680 to track this.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
tp: Will do, and added a ruler to my editor to keep an eye on these going forward.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/test/integration/email_accounts_admin_test.dart
tp: That's right, technically not all tests need this, but it felt "more real" to me.

This came about because yesterday I was bitten by https://github.com/serverpod/serverpod/pull/3675, so I wanted to stick to the pattern which is closest to production here (until we add some E2E tests in the end).

But I see that this can also be confusing (though on the flip-side it's equally weird how we sometimes have to switch from rollback to no rollback, as this all relies on some implementation detail in the code under test and is not really obvious from the "top" what "must" be used).

But I can update these to use as many automatic rollbacks as possible.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
tp: @coderabbitai Check that all comments on public classes and methods in this PR do not exceed an 80-character line limit.

Also, make sure to enforce this in future PR reviews.

tp:
Requires https://github.com/serverpod/serverpod/pull/3690 to be merged first.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email_config.dart
tp: Hmm, since we moved away from such "hooks" in the other processes, I wonder if that's really desirable here.

Maybe the 2 methods should just return what they did, so any custom code can run afterwards?

The obvious issue then is that sometimes you would have to call your own code in 2 places, but I think that's also happens already with many other functions where we don't provide the special hook to do/enforce something specific for you.
(E.g. if you want to restrict emails to a certain domain you will soon have to do that in the override of "register" and "invite", lest we provide a special hook for checking the emails like we do for the passwords (and even then there are now some exceptions where they don't hold…).)

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
tp: Updated to named imports.

> Or perhaps legacy instead of old?

Also, TIL that Apple calls the "iOS 7" design now "vintage" instead of "old" 😅

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
tp: That is correct, this needs to be moved atop the `if`.

Added a test that verifies that this was wrong before and is fixed now :)

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
tp: Yeah, it make sense to enforce is here, as we also do pass the transaction in when we migrate the user (so that migrated user, auth user and profile are created in one step).

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
tp: Sorry, I must've squashed an implementation commit into the generation ones.

That's always a bit annoying, as I usually do "base implementation", "code generation", and then "impl 1", "impl 2" etc. and then have to re-order the history to move those down and squash them so we end up with only 1 impl and 2 gen commit.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
tp: There are tests for both config options now (default with profile, and one without).

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
tp: Done, also used the named imports here so it matches the implementation file and gives further clarity into the setup and behavior under test.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
tp: @coderabbitai Look for any method calls in tests where the return value is under test but not checked and enforce the `await expectLater(asyncMethodCall(), completes)` pattern on them.

Be careful not to require this for _all_ function calls, as sometimes we just call a method with an assertion as part of a preparation, and then assert on side-effects (e.g. a DB entity existing afterwards).

Also do this in future PRs.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
tp: Removed

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
tp: Removed duplicate

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
tp: @coderabbitai Excellent finds, those 3 cases were all spot on. Fixed them in the latest push.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
```diff
@@ -0,0 +1,547 @@
+import 'package:serverpod/serverpod.dart';
```
tp, line 1: @coderabbitai check this test file, ensuring that all newly added methods from this PR are properly tested.

Double-check that each test description matches its implementation.

Check the spelling.

tp:
@SandPod Since this was rebased on `main` to make use of all the extracted features (and those commits have been removed from the history here), I just squashed the latest state into a "implement" and "generate" commit each.

I double-checked that the "generate" does not contain any implementation files.

These were the individual commits working on the comments from this PR (as that is tricky to see in GitHub web):

CleanShot 2025-06-20 at 15 02 47@2x

tp:
@coderabbitai full review

Focus especially on 62d8e4eb6e6b031c9fcd07d8463a1efcbe7e44bc

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
```diff
@@ -0,0 +1,254 @@
...
+  /// email, e.g. an account creation or password reset.
+  static Future<void> migrateWithoutPassword(
+    final Session session, {
+    required String email,
+    final Transaction? transaction,
+  }) async {
+    email = email.toLowerCase();
+
+    final emailAccountInfo =
+        await new_email_account.EmailAccounts.admin.findAccount(
```
SandPod, line 130: Suggestion(NIT): Would it be enough with simply `new` and `legacy` named imports?

Since it is used all over the file the reader should quickly understand what `new` and `legacy` means without the extra chars.

Same should apply to the test file.

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email_config.dart
SandPod: What worries me if we would return the information is that a majority of users would need to add custom code and not simply call into the method.

I think such hooks still have a place in a package like this is combines multiple operations into a single call. Recreating the logic is quite cumbersome if a user would like to use this package but migrate to their own profile mode.

One alternative is that they migrate in two steps.

But I guess we could also do the migration but then *also* return the user info data so that any additional user info migration can take place.

Do we think it would be an issue if we create the profile data even if it is not used?

SandPod: Or I guess the user could toogle this to false and then use the returned information?

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email.dart
tp: The problem is that we have multiple "new" modules, so that's why there is `new_email_account` and `new_profile` etc.

tp:
> Do we think it would be an issue if we create the profile data even if it is not used?

I don't think that's an issue at all. If later on they decide to drop the migration and default endpoint packages and rather create there own without `serverpod_auth_profile` the table will just be dropped and they can go on.
And the cost during the migration is minimal (if they don't set the flag).

I would be open to revisit the profile migration, but then we should probably find a way that also aligns with the `serverpod_auth_email` (and other providers) which currently also implicitly creates a profile.

Maybe we should have an `onAuthUserMigrated` here, and an `onAuthUserCreated` in the email account, so that indeed such things only need to be configured once. Because you're right, overriding in multiple places is a bit cumbersome, especially if the same logic is supposed to run there. (And emails also would've 2 code paths where a user is created: Account registration completion, and invites.)

This is probably little code in the end, but a lot of design work and discussions, so I think I would defer it for now and create a follow-up issue gathering these points.

Edit: Added https://github.com/serverpod/serverpod/issues/3709

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_email_config.dart
tp: > Or I guess the user could toogle this to false and then use the returned information?

Right, that would work. But as you said they would run their "profile setup" code after each call to `migrate*`.

PR #3672, opened 2025-06-18
feat(serverpod_test): Support `Object userIdentifier` in `AuthenticationOverride`, matching `AuthenticationInfo`

tp:
## Summary by CodeRabbit

- **New Features**
  - Added integration tests to verify authentication behavior with both integer and UUID user identifiers.  
- **Documentation**
  - Updated documentation to reflect changes in authentication parameter naming.  
- **Refactor**
  - Updated authentication methods to accept a generic user identifier instead of only integer user IDs.


tp:
> Please add a test for this change.

@SandPod Done. I had to expose the low-level class though (for testing).

I just did a minimal update, though I think we could also make this a proper `sealed` union.
In the end this could really just be achieved with a `AuthOverride` class having just a nullable `AuthenticationInfo? get authenticationInfo` interface, no need for case-specific classes at all right now.

Let me know if I should further refactor this or if we better leave it as is for now.

packages/serverpod_test/test/test_session_builder_test.dart
```diff
@@ -0,0 +1,55 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test/serverpod_test.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('Given an `AuthenticationOverride` with an `int` user ID', () {
```
SandPod: Question: Can't you test this througt the `withServerpod` group and then validate that this is actually reflected in the authenticationInfo of the session?

I think we should already have tests that validate that the override works as expected.

Then we don't need the visibleForTesting thing.

SandPod (changes requested):
Can this implementation be aligned to validate the actual code?

packages/serverpod_test/test/test_session_builder_test.dart
tp: There is no `withServerpod` in this package, but I could add it to the `test_server` instead.

tests/serverpod_test_server/test_integration/test_tools/test_session_builder_test.dart
```diff
@@ -0,0 +1,80 @@
...
+
+        authenticationInfo = await session.authenticated;
+      });
+
+      test(
+        "when inspecting the legacy `userId` property, then it returns the specified value.",
+        () {
+          expect(authenticationInfo?.userId, 123);
+        },
+      );
```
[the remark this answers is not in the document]
tp: But getting it is nullable, and then we don't have to use `!` in the setup, getting a nicer error if the overall setting would not work (and we get an unauthenticated user).

tests/serverpod_test_server/test_integration/test_tools/test_session_builder_test.dart
```diff
@@ -0,0 +1,80 @@
...
+      });
+
+      test(
+        "when inspecting the legacy `userId` property, then it throws.",
+        () {
+          expect(
+            () => authenticationInfo?.userId,
+            throwsA(isA<FormatException>()),
+          );
+        },
```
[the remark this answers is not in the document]
tp: This is fine, as we thus also assert that `authenticationInfo` is not `null` (as only the `userId` access will throw).

packages/serverpod_test/test/test_session_builder_test.dart
tp: Moved the tests to the test server.

There is also https://github.com/serverpod/serverpod/blob/b1bfe631e0cd3b474a0bbbcac6f993b80c261c7c/tests/serverpod_test_server/test_integration/test_tools/session_copy_with_test.dart

I suppose we could just create a simple `copyWith` test and ensure that the new value is not set on the original one builder/session.

The tests in that other file seem to say more about the test setup itself (depending on re-assigning variables across tests and the order of `test` execution) as they do about the correctness of `copyWith`.

SandPod (approved):
Awesome! 👍

PR #3675, opened 2025-06-18
fix(auth2): Pass `transaction` to `_hasTooManyFailedSignIns` check

tp:
Which is a requirement to be used in some integration tests

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of failed sign-in checks during the login process.

- **Style**
  - Minor formatting enhancements for account creation code.


SandPod (approved):
Strange we didn't catch this in a test already, would it be worth adding one that would detect if transaction is missing?

tp:
> Would it be worth adding one that would detect if transaction is missing?

I don't know how we could've caught this, unless we always try using `withServerpod` both with and without the implicit transaction.

This code path just wasn't fully exhausted in the previous login tests, but would now show up in the newly added endpoint tests (which do pass the `transaction` in some of them. I will have to recheck if that is actually required anymore (should not be I suppose), or whether this was just an attempt to get around this issue).

PR #3677, opened 2025-06-18
feat(auth2): Log `result`s from `EmailAccounts` methods and pass `transaction` in `finishRegistration`

tp:
As those should not be returned to the calling client, but an admin might need this for debugging.

Pass `transaction` for `_createSession` in `finishRegistration` so it can see a (non-committed) `AuthUser` and create a session for it.
(This was not visible so far due to the `withServerpod` tests passing the transaction implicitly.)

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
@@ -34,11 +34,21 @@ abstract class EmailAccountEndpoint extends Endpoint {
     required final String email,
     required final String password,
   }) async {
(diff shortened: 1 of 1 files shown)


modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/email_account_endpoint.dart
```diff
@@ -111,16 +133,21 @@ abstract class EmailAccountEndpoint extends Endpoint {
```
tp, line 133: This did not use a transaction on purpose, as the "destroy all sessions" has that major side-effect of and thus it seemed unsuitable to roll back (and also there is no bad "partial" stuff being done).

But I also don't have a super strong/argument feeling against this, so maybe it'd be fine to add it for consistency.

PR #3687, opened 2025-06-18
fix: Separate configuring and starting insights.

SandPod:
When building tests for Serverpod I noticed that that inisights server was not configured when constructing the Serverpod instance.

This moves the initialization of the insights server to the same place as all other servers are initialized.

## Pre-launch Checklist

## Breaking changes

None - should give the same flow.

## Summary by CodeRabbit

- **Refactor**
  - Improved the startup process for the Insights server, resulting in clearer separation between configuration and server start-up phases.
  - Enhanced error handling when the service secret is invalid, ensuring clearer messaging and more predictable server behavior.

### DIFF
--- a/packages/serverpod/lib/src/server/serverpod.dart
+++ b/packages/serverpod/lib/src/server/serverpod.dart
@@ -579,13 +579,7 @@ class Serverpod {
 
       // Serverpod Insights.
       if (Features.enableInsights) {
-        if (_isValidSecret(config.serviceSecret)) {
-          serversStarted &= await _serviceServer?.start() ?? true;
-        } else {
-          stderr.write(
-            'Invalid serviceSecret in password file, Insights server disabled.',
-          );
-        }
+        serversStarted &= await _serviceServer?.start() ?? true;
       }
 
       // Main API server.
--- a/packages/serverpod/lib/src/server/serverpod.dart
+++ b/packages/serverpod/lib/src/server/serverpod.dart
@@ -117,11 +117,11 @@ class Serverpod {
   /// The main server managed by this [Serverpod].
   late Server server;
 
-  Server? _serviceServer;
+  Server? _insightsServer;
 
   /// The service server managed by this [Serverpod].
   Server get serviceServer {
-    var service = _serviceServer;
+    var service = _insightsServer;
     if (service == null) {
       throw StateError(
         'Insights server is disabled, supply a Insights configuration '
@@ -469,7 +469,7 @@ class Serverpod {
 
     if (Features.enableInsights) {
       if (_isValidSecret(config.serviceSecret)) {
-        _serviceServer = _configureInsightsServer();
+        _insightsServer = _configureInsightsServer();
       } else {
         stderr.write(
           'Invalid serviceSecret in password file, Insights server disabled.',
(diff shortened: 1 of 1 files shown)


packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -466,6 +466,16 @@ class Serverpod {
         securityContext: _securityContextConfig?.webServer,
       );
     }
+
+    if (Features.enableInsights) {
+      if (_isValidSecret(config.serviceSecret)) {
+        _serviceServer = _configureInsightsServer();
```
tp: Would it make sense to rename `_serviceServer` to `_insightServer` (or on the flip side rename the helper)?

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -466,6 +466,16 @@ class Serverpod {
...
+      } else {
+        stderr.write(
+          'Invalid serviceSecret in password file, Insights server disabled.',
```
nielsenko, line 475: When do we use stderr, and when do we use logging?

packages/serverpod/lib/src/server/serverpod.dart
tp: Also, should we maybe change `Server? _serviceServer` to `late final` (maybe the same for the `server`), so we ensure that the initialization is run as expected (and e.g. it's not `null` by accident).

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -570,7 +580,7 @@ class Serverpod {
       // Serverpod Insights.
       if (Features.enableInsights) {
         if (_isValidSecret(config.serviceSecret)) {
```
tp:
```suggestion
```

Does it make sense to repeat the same checks as above here? I would probably just check for `_serviceServer != null` and then call `start` and update the variable.

tp (approved):
Overall I think the alignment of the 3 server start-ups makes sense :)

packages/serverpod/lib/src/server/serverpod.dart
SandPod: You are correct, this should be removed here. will fix!

packages/serverpod/lib/src/server/serverpod.dart
SandPod: Unfortunately, we don't have a consistent pattern for this. We are expecting an overhaul of this with the logging initiative.

stderr has now been used to alert users of all types of errors.

packages/serverpod/lib/src/server/serverpod.dart
SandPod: Fixed: [fix(review): Remove guard for password when starting server.](https://github.com/serverpod/serverpod/pull/3687/commits/67c5142cfc2dd51e3d7eaa3c37c209a2eab84eea)

packages/serverpod/lib/src/server/serverpod.dart
SandPod: Renaming done: [refactor(review): Rename to better reflect server purpose.](https://github.com/serverpod/serverpod/pull/3687/commits/4890c953fcd079ebc11708005257250cac0cb192)

Oped to keep the nullable status as is for now.

PR #3690, opened 2025-06-19
Extend `EmailAccountsAdmin` with new methods needed for email account imports

tp:
- findAccount: To check whether the model knows about an email address already
- createEmailAuthentication: To create an email authentication without going through the verification flow
- setPassword: To set an account's password without going through the password reset flow

Also, fixes `unnecessary_non_null_assertion` lint warnings.


modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_account_secret_hash.dart
```diff
@@ -9,14 +9,16 @@ import 'package:serverpod_auth_email_account_server/serverpod_auth_email_account
 import 'package:serverpod_auth_email_account_server/src/business/email_account_secrets.dart';
 import 'package:serverpod_shared/serverpod_shared.dart';
 
-/// Class for handling password and verification code hashing in the email account module.
+/// Class for handling password and verification code hashing in the email
```
tp, line 12: (Applied the 80 char limit to the touched files as well. Though would've also been nice to find out how fast 🐇 learns the new rule.)

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/test/integration/email_accounts_admin_test.dart
```diff
@@ -11,14 +11,15 @@ import 'test_tools/serverpod_test_tools.dart';
 void main() {
```
tp, line 11: @coderabbitai check this test file, ensuring that all newly added methods from this PR are properly tested.

Double-check that test descriptions match their implementation.

Check the spelling.

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
```diff
@@ -77,4 +84,112 @@ final class EmailAccountsAdmin {
...
+  /// The [password] argument is not checked against the configured password
+  /// policy.
+  ///
+  /// Returns the email account ID for the newly created authentication method.
+  Future<UuidValue> createEmailAuthentication(
+    final Session session, {
+    required final UuidValue authUserId,
+    required String email,
+
+    /// If `null`, no password will be set for this user.
```
SandPod: Suggestion: Describe that the user will not be able to login if the value is null.

Shouldn't this also be noted above in the description of the method?

SandPod (approved):
LGTM! 🚀

modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
tp: Will consolidate the comment "above" and add a note that the account will need a password (re)set to be usable.

PR #3693, opened 2025-06-19
ci: Add `serverpod_new_auth_test_[client|server]` to all scripts

tp:
## Summary by CodeRabbit

- **Chores**
	- Added integration tests for an additional server module to the continuous integration workflow.
	- Included new test server modules in migration recreation and static analysis processes.

### DIFF
--- a/util/run_tests_analyze
+++ b/util/run_tests_analyze
@@ -41,6 +41,8 @@ declare -a projectPaths=(
     "tests/serverpod_test_module/serverpod_test_module_client"
     "tests/serverpod_test_nonvector/serverpod_test_nonvector_server"
     "tests/serverpod_test_nonvector/serverpod_test_nonvector_client"
+    "tests/serverpod_new_auth_test/serverpod_new_auth_test_server"
+    "tests/serverpod_new_auth_test/serverpod_new_auth_test_client"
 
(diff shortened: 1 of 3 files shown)


util/run_tests_analyze
```diff
@@ -41,6 +41,8 @@ declare -a projectPaths=(
     "tests/serverpod_test_module/serverpod_test_module_client"
     "tests/serverpod_test_nonvector/serverpod_test_nonvector_server"
     "tests/serverpod_test_nonvector/serverpod_test_nonvector_client"
+    "tests/serverpod_new_auth_test/serverpod_new_auth_test_server"
```
SandPod, line 44: Question(Required action if not true): It this added to the `pub_get_all` script as well?

If you run `git clean -dfx` and then run the `pub_get_all` script. Does that yield any errors in the project?

SandPod (approved):
Validate this is added to the `pub_get_all` scrip as well.

util/run_tests_analyze
tp: Yes, it was already in `pub_get_all`.

Overall, there are currently 6 scripts places such a module needs to be added to, this was only mentioned in half of them…

`git clean -dfx && ./util/pub_get_all` passes without error :)

Issue #3633, opened 2025-06-12
Upload apk results in: Invalid array length


vanlooverenkoen:
### Describe the bug

For some reason are my apk uploads resulting in an `Invalid array length`

https://github.com/serverpod/serverpod/blob/main/packages/serverpod_client/lib/src/file_uploader.dart#L48

```
      try {
        var result = await http.post(
          _uploadDescription.url,
          body: await _readStreamData(stream),
          headers: {
            'Content-Type': 'application/octet-stream',
            'Accept': '*/*',
          },
        );
        return result.statusCode == 200;
      } catch (e) {
        return false; // I end up here with an `Invalid array length`  error
      }
```

### To reproduce

This is what i
```
var uploader = FileUploader(uploadDescription);
final result = await uploader.uploadByteData(ByteData.view(fileBytes.buffer));
```

Or this
```
var uploader = FileUploader(uploadDescription);
final fileStream = Stream.value(fileBytes.toList());
final length = fileBytes.length;
final result = await uploader.upload(fileStream, length);
```

Both of them result false.

### Expected behavior

I expect the upload to succeed

### Library version

2.8.0

### Platform information

Web

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

SandPod:
Thanks for the report!

We will try to get to this as soon as possible 👍

vanlooverenkoen:
@SandPod any idea why this is failing? I think my setup is correct (because it works for ipa files, they are smaller so maybe that is the issue?)

I also have set 
`maxRequestSize: 200000000`

SandPod:
Sorry, I have not been able to dig into this one yet.

I have added it to our list over prioritized bugs. I hope that we can give you a response on this soon.

nielsenko:
@vanlooverenkoen How big is the apk you are trying to upload? What is the `uploadDescription` passed? (or just the url). I see we don't trace anything about the error, but did you try to set a breakpoint in the catch clause to see the actual exception? I would like to know the exact type, and any related data it might carry.

nielsenko:
What runtime are you using?

vanlooverenkoen:
The uploadDescription is done exactly how it was documented:
https://docs.serverpod.dev/concepts/file-uploads#client-side-code

The apk is 71mb 

The error that I found was `Invalid array length`, I'm not sure where I can get extra info on this error

[...]

Issue #3128, opened 2025-01-16
RFC: Cloud storage improvements


Isakdl:
# Cloud Storage Integration Improvement

![Image](https://github.com/user-attachments/assets/24e8dec4-1a18-4b5b-83dc-e0522436de37)

---

## Summary

This proposal outlines improvements to the cloud storage integration in Serverpod. Current integrations face several challenges, including a lack of testing, unclear documentation, and limited flexibility. This RFC proposes changes to the `CloudStorage` interface, enhancements, and new features such as metadata handling, new storage implementations, etc to address these issues comprehensively.

---

## Motivation

The current state of cloud storage integration has notable deficiencies:
- Frequent bug reports indicating confusing or broken functionality.
- Absence of tests, increasing the likelihood of introducing undetected bugs.
- Missing features such as support for metadata.
- Reliance on database validation for uploads.
- Sparse documentation, making it difficult for developers to extend or use the feature effectively.

This proposal addresses these issues to improve reliability, flexibility, and developer experience.

---

## Current Pain Points

- **Bugs and Inconsistencies**: Lack of tests leads to regressions.
- **Sparse Documentation**: Insufficient guidance for implementing custom extensions.
- **No Metadata Support**: Inability to associate metadata with files.
- **No Local Disk Storage**: No integration to store files on the local machine. 
- **AWS Compatibility Layer for GCP**: The current GCP integration relies on AWS compatibility rather than a native implementation, limiting features and user control.

---

### Interface Changes to `CloudStorage`

The complete interface after proposed chanages

```dart
abstract class CloudStorage {
  /// Identifies the storage. You can used multiple types cloud storages in a
  /// single Serverpod.
  String storageId;

  /// Creates a [CloudStorage] with the specified [storageId]. By default,
  /// two storages are used by Serverpod `public` and `private`. The public
  /// should be accessible to everyone (usually through a web interface),
  /// while the private is accessed internally only.
  CloudStorage(this.storageId);

  Future<void> storeFile({
    required Session session,
    required String path,
    required ByteData byteData,
    DateTime? expiration,
  });

  Future<ByteData> retrieveFile({
    required Session session,
    required String path,
  });

  Future<Uri> getPublicUrl({
    required Session session,
    required String path,
  });

  Future<bool> fileExists({
    required Session session,
    required String path,
  });

  Future<void> deleteFile({
    required Session session,
    required String path,
  });

  Future<UploadDescription> createDirectFileUploadDescription({
    required Session session,
[...]

PR #3695, opened 2025-06-19
refactor: FileUploader.upload now uses StreamedRequest

nielsenko:
Strive to  reduce memory pressure during upload by supporting streamed upload via `StreamedRequest`.

Updates `http` package dependency and refactors `FileUploader` for improved stream handling.

Closes: #3692 

## Pre-launch Checklist

## Breaking changes

None

## Summary by CodeRabbit

- **Refactor**
  - Improved and unified the file upload process for better reliability and consistency.
  - Enhanced error handling and streamlined the upload logic for file uploads.

- **Chores**
  - Updated the minimum required version of the `http` package to 1.1.0 across all packages, modules, templates, and tests to ensure compatibility with newer features and fixes.

- **Tests**
  - Added tests for streaming binary uploads and uploader state validation to improve coverage and reliability.

### DIFF
--- a/tests/serverpod_test_server/test_e2e/cloud_storage_test.dart
+++ b/tests/serverpod_test_server/test_e2e/cloud_storage_test.dart
@@ -12,7 +12,7 @@ Stream<int> streamBytes() async* {
 }
 
 extension<T> on Stream<T> {
-  Stream<List<T>> inChunks(int chunkSize) async* {
+  Stream<List<T>> inChunksOf(int chunkSize) async* {
     var chunk = <T>[];
     await for (final e in this) {
       chunk.add(e);
@@ -180,13 +180,14 @@ void main() {
       expect(verifyByteData(byteData), equals(true));
     });
 
-    test('Direct file upload (binary stream)', () async {
+    test('Direct file upload (Stream<List<int>>)', () async {
       var uploadDescription = await client.cloudStorage
           .getDirectFilePostUrl('testdir/directupload_stream_binary.bin');
       expect(uploadDescription, isNotNull);
 
       var uploader = FileUploader(uploadDescription!);
-      var result = await uploader.upload(streamBytes().take(512).inChunks(64));
+      var result =
+          await uploader.upload(streamBytes().take(512).inChunksOf(64));
 
       expect(result, equals(true));
 


packages/serverpod_client/lib/src/file_uploader.dart
```diff
@@ -21,79 +22,60 @@ class FileUploader {
...
-        request.fields[key] = _uploadDescription.requestFields[key]!;
-      }
+        case _UploadType.multipart:
+          var multipartFile = switch (length) {
+            null => http.MultipartFile.fromBytes(
+                _uploadDescription.field!, await stream.toBytes(),
+                filename: _uploadDescription.fileName),
+            _ => http.MultipartFile(_uploadDescription.field!, stream, length,
+                filename: _uploadDescription.fileName),
+          };
```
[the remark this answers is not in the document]
nielsenko: This is the nature of `MultipartFile` uploads - you will need to know the length. The important part of this PR is that if you _don't_ use multipart, then the stream needs not be read before starting to send (ie. the `_UploadType.binary` case).

packages/serverpod_client/lib/src/file_uploader.dart
```diff
@@ -21,59 +22,60 @@ class FileUploader {
   /// Uploads a file contained by a [ByteData] object, returns true if
   /// successful.
   Future<bool> uploadByteData(ByteData byteData) async {
-    var stream = http.ByteStream.fromBytes(byteData.buffer.asUint8List());
-    return upload(stream, byteData.lengthInBytes);
+    final buffer = byteData.buffer;
+    var stream = http.ByteStream.fromBytes(buffer.asUint8List());
```
tp: I think this would have to be:

```suggestion
    var stream = http.ByteStream.fromBytes(Uint8List.sublistView(byteData));
```

I had a similar issue in the new auth methods, where we use `ByteData` in the DB to store password hashes.

There, on queries, the `ByteData`'s buffer keeps pointing to the underlying "row", and thus at first my conversions to `UInt8List` (where I think I've used a similar approach as you did here) failed. The data was just much longer than the hash, so this was easy to spot.

I think your version could even work well in 99% of the cases (depending on how the ByteData is obtained), but when for example posting binary data read from the database (with other columns) the current version would then send to much.

(At least that's my understanding, that `ByteData` is basically a "view" and then `buffer` is the underlying storage, which might hold other data and have other "view"s into it.)

packages/serverpod_client/lib/src/file_uploader.dart
```diff
@@ -21,59 +22,60 @@ class FileUploader {
...
-      request.files.add(multipartFile);
-      for (var key in _uploadDescription.requestFields.keys) {
-        request.fields[key] = _uploadDescription.requestFields[key]!;
-      }
+        case _UploadType.multipart:
+          var multipartFile = switch (length) {
+            null => http.MultipartFile.fromBytes(
+                _uploadDescription.field!, await stream.toBytes(),
+                filename: _uploadDescription.fileName),
+            _ => http.MultipartFile(_uploadDescription.field!, stream, length,
```
tp, line 62: Nice type-inference on the `length` here, even without it being "captured".

packages/serverpod_client/lib/src/file_uploader.dart
```diff
@@ -21,59 +22,60 @@ class FileUploader {
...
-    var stream = http.ByteStream.fromBytes(byteData.buffer.asUint8List());
-    return upload(stream, byteData.lengthInBytes);
+    final buffer = byteData.buffer;
+    var stream = http.ByteStream.fromBytes(buffer.asUint8List());
+    return _upload(stream, buffer.lengthInBytes);
   }
 
   /// Uploads a file from a [Stream], returns true if successful.
-  Future<bool> upload(Stream<List<int>> stream, int length) async {
+  Future<bool> upload(Stream<List<int>> stream, [int? length]) =>
```
tp, line 32: Making `length` optional is just a "bonus", right?

I see it was required before, and we don't newly ever call it with `null`.

So just wondering if that's required at all.

Could be worth to even go one stack higher in the stack (just for our understanding), so see what is actually passed here, and if any unneeded conversion is being done (e.g. should we just support "upload file"?).

tp, line 32: It looks like this method is missing E2E tests entirely, right?

And even if we add that, I think we also have to cover both upload "types". (How that whole pattern is configured / when which case will be entered is not entirely clear to me (outside of the actual upload)).

tp (commented):
The ByteData to Stream conversion looks fishy, otherwise the implementation looks fine to me.

Still have some questions though, and without more tests I would probably not ship this right now. But unless I missed more test, smothering this with E2E ones seems longer overdue anyway, and would thus be greatly appreciated I suppose.
(Also should think whether we should be using this `FileUploader` pattern for the profile image, instead of sending bytes through the Serverpod API and then forwarding that to the bucket store.)

packages/serverpod_client/lib/src/file_uploader.dart
nielsenko: Interesting. I will see if I can create a test-case for that. Not really related to this PR as `uploadByteData` doesn't really change (just introduces a name for temporary), but let us address since we have it fresh in mind.

packages/serverpod_client/lib/src/file_uploader.dart
nielsenko: Yes switch patterns nice 😄

packages/serverpod_client/lib/src/file_uploader.dart
nielsenko: It is just a bonus. We should consider changing the interface on the cloud storage endpoint to actually benefit from this. You can use the `FileUploader` directly though (it is exported from `serverpoint_client.dart`).

nielsenko: No, it is tested since it is used by `FileUploader.uploadByteData`, but it could definitely be tested more. It is a bit tricky to tweak the tests to actually exercise the reported bug.

tp: That's right, so we're not missing another test for this method, but just the "second" upload type (as I only saw 1 I think).

And then also for the newly optional length, that seems like an interesting code-path to exercise (or not support for now).

As I said above, the bigger benefit is probably support uploads from `File` instead of taking `Stream` and `length` (or bytes) here, that are very likely derived from a file in practice.

packages/serverpod_client/lib/src/file_uploader.dart
tp: Makes sense, since the interface is just `ByteData` and we can not know where it comes from (though I suppose a DB row is unlikely on the client…).

But it actually bit me twice, so the auth stuff now has this extension everywhere:

CleanShot 2025-06-20 at 13 22 05@2x

vanlooverenkoen:
As I said before I logged #3633 and used this PR to test my implementation. This PR fixes
- the invalid lenght error
- freezing the UI when loading everything in memory.

The upload of my 74mb apk went through without any issues. 🚀 🎉

packages/serverpod_client/lib/src/file_uploader.dart
nielsenko: You are right. The buffer (`ByteBuffer`) might be bigger than the `ByteData`. This little snippets illustrates the semantics:
```dart
import 'dart:typed_data';

void main() {
  var originalBuffer = Uint8List.fromList(List.generate(10, (i) => i)).buffer; // Underlying buffer: [0, 1, ..., 9]

  var myByteData = ByteData.view(originalBuffer, 3, 3); // myByteData represents [3, 4, 5]

  var view1 = myByteData.buffer.asUint8List();
  print('myByteData.buffer.asUint8List(): $view1 (length: ${view1.length})');
  // Output: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9] (length: 10) - Views the whole original buffer.

  var view2 = Uint8List.sublistView(myByteData);
  print('Uint8List.sublistView(myByteData): $view2 (length: ${view2.length})');
  // Output: [3, 4, 5] (length: 3) - Views only the part myByteData points to.
}
```

nielsenko: I'll use the `sublistView` ctor as suggested to avoid this pitfall.

packages/serverpod_client/lib/src/file_uploader.dart
nielsenko: Problem with `File` is that it belongs to `dart:io` which I suppose we try to avoid in `serverpod_client` to allow clients to run on web.

nielsenko: I'll add more tests, however..

Currently multipart file upload is only used for S3 and GCP, whereas we use binary upload for the Postgres backend. Adding extra tests in this area will require some integration work towards these services. @coderabbitai could you make a separate issue for this?

tp: Ah, then maybe https://pub.dev/packages/cross_file would have to be used in practice?

In the end the important part about the API parameter would just be a “path” so that the later upload can use the most efficient way of reading/streaming it it, and we don’t have to fully read it into memory like ByteData.

I was just thinking that this is probably the 99% case, and thus it would be good to handle it once instead of burdening all callers with it.

But either way, that’s out of scope of this concrete improvement. As we have surprisingly little usage of this in this repo, I don’t feel that I can make a good case for any API update here.

nielsenko: Agree on everything said.

@SandPod what is your thoughts on this (Taking a dep on [XFile](https://pub.dev/packages/cross_file)? Not for this PR of course, but should we make an issue for that?

nielsenko: Trying to wake the AI again..

> I'll add more tests, however..
> 
> Currently multipart file upload is only used for S3 and GCP, whereas we use binary upload for the Postgres backend. Adding extra tests in this area will require some integration work towards these services. 

@coderabbitai could you make a separate issue for this?

packages/serverpod_client/lib/src/file_uploader.dart
SandPod: TIL 👀

packages/serverpod_client/lib/src/file_uploader.dart
nielsenko: Another issue we have today is that multipart uploads actually only creates a single part, which if I recall correctly sets a size limit on S3 of 5G (as opposed to 5TB).

@coderabbitai Could you make a follow-up issue for this?

tp: @nielsenko Got a bit confused at first, since "multipart" uploads seem to refer to 2 concepts:

- Uploading multiple files in 1 request (this is what `upload` does, though it only ever works on 1 file)
- Uploading 1 file across multiple requests (this is where the 5TB S3 limit comes into play, as you have to upload those in "small" 5GB chunks)

With that in mind, quite interesting that we would be using a [MultipartFile](https://pub.dev/documentation/http/latest/http/MultipartFile-class.html) here when only ever uploading 1 file. But I suppose some storage endpoints require that format. (Especially since the "give me a URL for the user to upload to" does not support receiving multiple files as far as I recall, though maybe by now you can also get a "folder" link and allow them to upload more than 1 file…)

nielsenko: @tp We use multipart upload for both S3 and GCS. Not sure why that was chosen, given that we don't actually use multiple parts. I guess a simple S3 PutObject would do. Anyway, this is getting into #3128 territory.

tests/serverpod_test_server/test_e2e/cloud_storage_test.dart
```diff
@@ -1,22 +1,41 @@
...
 import 'package:serverpod_test_server/test_util/config.dart';
 import 'package:test/test.dart';
 
+Stream<int> streamBytes() async* {
+  int i = 0;
+  while (true) yield i++ % 256;
+}
+
+extension<T> on Stream<T> {
+  Stream<List<T>> inChunks(int chunkSize) async* {
```
tp: Nit: `inChunks(5)` reads like it would return 5 chunks, not chunks of size 5 to me.

tp (approved):
LGTM 🚀

Just a tiny nit on the test helper, but hey, it's only used internally and easily read through :)

(And seems fine to leave the old tests as is for now, as Alex said this will get an overhaul later this year.)

tests/serverpod_test_server/test_e2e/cloud_storage_test.dart
tp: (The Swift-y in me would want to write `inChunks(of: )` or `inChunksOf(…)`, or just `chunked` which is ambiguous and one would read it up or check the docs (not that the behavior really matters for this test anyway).)

So feel free to ignore 😉

PR #3704, opened 2025-06-23
fix(auth2): Remove `.` from password variable names

### LINKED ISSUE #3680: Use "password" keys which can also be configured via ENV variables
Initially, we used password configuration keys like `serverpod_auth_email_account.passwordHashPepper`, namespacing the key under the module name, separated by a `.`

Unfortunately, such keys can not be set as ENV variables (due to the `.`), and thus we need to change the existing keys to use a pattern like `module_name_key_name`.

So `serverpod_auth_email_account.passwordHashPepper` would become `serverpod_auth_email_account_password_hash_pepper`.

tp:
So that they can be set as ENV variables

Closes #3680

## Summary by CodeRabbit

- **Refactor**
  - Renamed configuration keys by replacing dots with underscores for improved clarity and consistency across environments.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/config/passwords.yaml
+++ b/modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/config/passwords.yaml
@@ -12,4 +12,4 @@
 test:
   database: 'DB_TEST_PASSWORD'
   redis: 'REDIS_TEST_PASSWORD'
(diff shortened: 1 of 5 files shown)


tp:
> 👍
> 
> Was this all of them that was commited? That is good new, I though we added more.

I thought so as well, but I guess that's because there's also the "normal" config for some "security-related" settings.

As for secrets stored in `passwords` we only have the peppers.

I searched for `.` in `password.yaml` to make sure we got them all.
(I think there was just 1 file that did not show up in VS Code search for some reason, but then failing CI test pointed it out…)

PR #3710, opened 2025-06-24
feat(auth2): Re-export email account and session packages from email endpoint package

tp:
So consumer only have to add 1 dependency to their pubspec and can use and configure the underlying packages.

For example, a developer wanting to add email accounts to their app can now just add `serverpod_auth_email` to their pubspec.yaml, but then gets accounts to `EmailsAccounts.config`, `AuthSessions.config`, and `UserProfiles.config` to configure them all without having to add a package for each of them.

## Summary by CodeRabbit

* **New Features**
  * Expanded authentication module API by making additional authentication-related features available for integration.

* **Documentation**
  * Updated documentation comments to clarify usage related to authentication session IDs.

* **Refactor**
  * Renamed an extension to better reflect its association with authentication session IDs.
  * Adjusted exported symbols to prevent naming conflicts and improve clarity.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/serverpod_auth_email_server.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/serverpod_auth_email_server.dart
@@ -1,3 +1,12 @@
+export 'package:serverpod_auth_email_account_server/serverpod_auth_email_account_server.dart'
+    hide Endpoints, Protocol;
+export 'package:serverpod_auth_profile_server/serverpod_auth_profile_server.dart'
+    hide Endpoints, Protocol;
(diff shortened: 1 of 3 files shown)


modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/serverpod_auth_email_server.dart
```diff
@@ -1,3 +1,8 @@
+export 'package:serverpod_auth_email_account_server/serverpod_auth_email_account_server.dart'
```
tp, line 1: Was debating whether to use `show` or `hide` here, but I think going forward, if there is for example a new type for a new config flag, it'll likely be missed here, so a "full" export is warranted.

tp:
@coderabbitai full review

exaby73 (approved):
LGTM

PR #3712, opened 2025-06-25
feat(auth2): Add `AuthSuccess` model

tp:
To be returned from a successful login, instead of just the `String` `sessionKey`

This additionally contains the `authUserId` and `scopes`.

This will be used in the client "session manager" for `serverpod_auth_session`, where it will be really useful to have the auth user ID and scopes available for a logged-in user. (In https://github.com/serverpod/serverpod/pull/3721)

## Summary by CodeRabbit

- **New Features**
  - Introduced a new authentication result object that provides session key, user ID, and granted scopes after login and registration actions.

- **Bug Fixes**
  - Updated authentication methods to return detailed session information instead of just a session key.

- **Tests**
  - Adjusted test cases to use the new authentication result object and updated variable naming for clarity.

- **Documentation**
  - Improved documentation to reflect changes in authentication method responses.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_client/lib/src/protocol/auth_success.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_client/lib/src/protocol/auth_success.dart
@@ -1,0 +1,97 @@
+/* AUTOMATICALLY GENERATED CODE DO NOT MODIFY */
+/*   To generate run: "serverpod generate"    */
+
+// ignore_for_file: implementation_imports
+// ignore_for_file: library_private_types_in_public_api
+// ignore_for_file: non_constant_identifier_names
+// ignore_for_file: public_member_api_docs
+// ignore_for_file: type_literal_in_constant_pattern
+// ignore_for_file: use_super_parameters
+
+// ignore_for_file: no_leading_underscores_for_library_prefixes
+import 'package:serverpod_client/serverpod_client.dart' as _i1;
+
+abstract class AuthSuccess implements _i1.SerializableModel {
+  AuthSuccess._({
+    required this.sessionKey,
+    required this.authUserId,
+    required this.scopeNames,
+  });
+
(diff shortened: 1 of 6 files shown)


tp:
@coderabbitai review

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/models/auth_success.spy.yaml
```diff
@@ -0,0 +1,11 @@
+class: AuthSuccess
```
exaby73, line 1: I wonder if the name could possibly be slightly different. Something like `AuthData` or simlar. Only reason I say this is because the existence of `AuthSuccess` might hint me that there also might be an `AuthError` or `AuthFailure` which is not the case.

This is definitely fine though. We don't need to change it. Just a thought

exaby73 (approved):
LGTM

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/models/auth_success.spy.yaml
tp: Yeah, I am not entirely happy with it either, but also did not want to just do something super generic like `AuthInfo`.

Maybe it should be `AuthSession`, but then we just have to make the distinction between the internal and external representation clear (as the DB entity has the same name, but is not exposed).

It's also a bit weird now in the usage, as you also get an "AuthSuccess" upon password reset.

Luckily we don't have to commit to this quite yet, as we'll do a big "auth 2.0 review" at the end, going through all the introduced APIs (and packages) and align names and behavior across them. So maybe until then we have a grasp at a more suitable name.

exaby73: That makes sense :D Thanks for the clarification

PR #3714, opened 2025-06-25
fix: Update UUID generator reference to use `serverpod` package instead of `uuid`

### LINKED ISSUE #3395: UUID `id` fields lead to import warning
### Describe the bug

When using a UUID `id` field with the new capabilities added in #3299, I get an import warning in my generated model class:

Image

For a normal UUID field this was not the case:

Image

As shown above this is due to the `default=random` field configuration, which creates a new value through the underlying `uuid` package.

Both `UuidValue` and `Uuid` are defined in package `uuid` and exported through `serverpod_serialization` here:
https://github.com/tp/serverpod/blob/6f6abe7ae6d120bb2437d39a432bcc18235df2d2/packages/serverpod_serialization/lib/serverpod_serialization.dart#L1

I think we need to adapt whatever leads to the generator taking the `serverpod/serverpod.dart` import for the `UuidValue` to also apply to `Uuid` to create the random value here.

### To reproduce

-

### Expected behavior

-

### Library version

`main` at `1c286f088a4901eb266b1678c1cfdae021ae20bb`

### Platform information

-

### Additional context

-

- [ ] I want to work on this

exaby73:
This PR aims to fix the import warning for uuid in generated files when using `defaultModel=random`

Fixes #3395

## Pre-launch Checklist

### DIFF
--- a/tests/serverpod_test_client/lib/src/protocol/changed_id_type/many_to_many/course.dart
+++ b/tests/serverpod_test_client/lib/src/protocol/changed_id_type/many_to_many/course.dart
@@ -10,20 +10,19 @@
 
 // ignore_for_file: no_leading_underscores_for_library_prefixes
 import 'package:serverpod_client/serverpod_client.dart' as _i1;
-import 'package:uuid/uuid.dart' as _i2;
-import '../../changed_id_type/many_to_many/enrollment.dart' as _i3;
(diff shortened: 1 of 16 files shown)


tp (approved):
LGTM 🚀

(With the 2 implementation commits squashed in 1 before the merge.)

Issue #3424, opened 2025-04-15
Invalid client code generated when combining `scope=serverOnly` and `default` value in model definition


dkbast:
### Describe the bug

When defining a field with both `scope=serverOnly` and a `default` value, the generated client code includes a constructor initializer for the field, even though it doesn't exist in the client model. This leads to a Dart compilation error.

Example model:
```yaml
class: FavouriteRecipe
table: favourite_recipe
fields:
  recipe: Recipe
  userId: int?, scope=serverOnly, default=-1
```

Generated (invalid) client code:
```dart
abstract class FavouriteRecipe implements _i1.SerializableModel {
  FavouriteRecipe._({
    this.id,
    required this.recipe,
  }) : userId = userId ?? -1; // ❌ Invalid: 'userId' is not defined
```

Compiler error:
```
'userId' isn't a field in the enclosing class.
Undefined name 'userId'.
```

### To reproduce

1. Define a model with a `scope=serverOnly` field that also has a `default` value.
2. Run `serverpod generate`.
3. Check the generated client code.

### Expected behavior

Client code should not reference `serverOnly` fields at all. The default value should only be applied during deserialization on the server side.

### Library version

Serverpod 2.5.0  

### Platform information

Occurs on all platforms during code generation (macOS, Linux, etc.)

### Additional context

The issue appears to be in the code generation logic where default values are added regardless of the field's scope. Client models should omit `serverOnly` fields entirely.

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

PR #3716, opened 2025-06-25
fix: Correctly generate code for `scope=serverOnly` fields with a `default` value

ashishexee:
recheck
#### 📌 Related Issue: [[#3424](https://github.com/serverpod/serverpod/issues/3424)](https://github.com/serverpod/serverpod/issues/3424)

When a field in a model is marked with both `scope=serverOnly` and `default`, the generated client constructor incorrectly tries to access the field — resulting in a Dart compilation error.

---

### 🐞 Problem Summary

Given the following model:

```yaml
class: FavouriteRecipe
table: favourite_recipe
fields:
  recipe: Recipe
  userId: int?, scope=serverOnly, default=-1
```

The generated **client model** includes:

```dart
FavouriteRecipe._({
  this.id,
  required this.recipe,
}) : userId = userId ?? -1; // ❌ Compilation Error
```

Which leads to:

```
Compiler Error:
'userId' isn't a field in the enclosing class.
Undefined name 'userId'.
```

---

### ✅ Fix Summary

This PR ensures that default value initializers are only added to the constructor **if the field is included in the current codegen scope** (i.e., not `serverOnly` during client model generation).

**Modified logic:**

```dart
for (var field in defaultValueFields) {
  if (field.shouldIncludeField(serverCode)) {
    Code? defaultCode = _getDefaultValue(
      field,
      subDirParts: subDirParts,
    );

    if (defaultCode == null) continue;

    c.initializers.add(Block.of([
      refer(field.name).code,
      const Code('='),
      refer(field.name).ifNullThen(CodeExpression(defaultCode)).code,
    ]));
  }
}
```

---

### 🔬 Testing

* ✅ Generated code using a sample project with `serverOnly` + `default` field
* ✅ Confirmed client code no longer contains invalid constructor initializer
* ✅ Dart compiler no longer throws error
* ✅ Regenerated models and verified unaffected behavior on server side

---

### 🧪 Suggested Future Improvement

Consider extending test coverage for codegen to catch issues where:

* Fields marked `serverOnly` are incorrectly referenced in client models
* Default values are mishandled across scopes

---

### 🙋‍♂️ About Me

I’m a Flutter + Node.js developer actively exploring server-side Dart. I came across this issue while working with `serverpod` in a full-stack Flutter app. I'd love to continue contributing!

---

### ✅ Checklist

* [x] Fix applied only where needed
* [x] Verified backward compatibility
* [x] Manual testing done via sample project
* [x] Linked related issue (#3424)

---

Let me know if you'd like me to add a test case — happy to contribute more 🚀

---

### DIFF
--- a/tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
@@ -1415,17 +1415,20 @@ class SerializableModelLibraryGenerator {
 
       var defaultValueFields = classFields.where((field) => field.hasDefaults);
       for (var field in defaultValueFields) {
-        Code? defaultCode = _getDefaultValue(
-          field,
-          subDirParts: subDirParts,
-        );
-        if (defaultCode == null) continue;
+        if (field.shouldIncludeField(serverCode)) {
+          Code? defaultCode = _getDefaultValue(
+            field,
+            subDirParts: subDirParts,
+          );
 
-        c.initializers.add(Block.of([
-          refer(field.name).code,
-          const Code('='),
-          refer(field.name).ifNullThen(CodeExpression(defaultCode)).code,
-        ]));
+          if (defaultCode == null) continue;
+
+          c.initializers.add(Block.of([
+            refer(field.name).code,
+            const Code('='),
+            refer(field.name).ifNullThen(CodeExpression(defaultCode)).code,
+          ]));
+        }
       }
 
       var implicitFields = classFields.where(


ashishexee:

tp:
@ashishexee Not sure where your CLA comment went, but let's figure this out :)

It would be great if you could add a test case that's failing on `main` (surprised if no-one would use this combination, but possible), which would then be fixed by your latest commit.

(You can just add a test for the fix in here, and I can inspect the erroneous behavior on main myself.)

Thanks for your contribution.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -1415,17 +1415,20 @@ class SerializableModelLibraryGenerator {
 
       var defaultValueFields = classFields.where((field) => field.hasDefaults);
```
tp, line 1416:
```suggestion
      var defaultValueFields = classFields.where((field) => field.hasDefaults && field.shouldIncludeField(serverCode));
```

What do you think about adding the logic up here, so we don't need to filter again and don't have additional nesting?

ashishexee: Yeah thanks for the suggestion this way we can avoid the additional if nesting

ashishexee: Actually I am very new to contributing so things are getting a bit overwhelming for me but still figuring a way out and trying to make a decent contribution

tp: No worries.

I think this is a great first issue, as the scope is clearly defined and you have already figured out the right check for the fix 😊

Now we just need to add some tests, and then this should be good to merge.

ashishexee: on it

ashishexee: sir i am not able to generate the test cases by self ,could you lend me a hand

tp: I think for starters it would suffice to extend this "test model" with a new field showcasing the fix:

https://github.com/serverpod/serverpod/blob/dccf3790ea1434500b5f9558dd187cdf690f871b/tests/serverpod_test_server/lib/src/models/scopes/server_only_class_field.spy.yaml#L1-L3

That model is already just used for a code-generation check and not further tested / round-tripped.

(There's also some tests that run the code-generator at test time, but I think that's optional for now.)

Let me know if that works for you, else I can also push a test commit for you to review.

tp: @coderabbitai Can you suggest further test cases for the fix in this PR?

tp: @coderabbitai I think that's overdoing it a bit. Since we know the fix from this PR is on a field level, we don't need to test every type (or re-test existing / working cases).

So I think my initial suggestion still holds.

On top of that it might be worth to add that "code generation in test" check, which would then verify that the generated server code has the field and sets the default, whereas the client-generated code does not have it.

Could you find the kind of tests I am thinking about and suggest a test case?

ashishexee: Thanks, that makes sense!

I'll go ahead and update the server_only_class_field.spy.yaml with a new scope=serverOnly field that has a default value to demonstrate the issue. That should be enough to confirm the fix is working as expected.

I'll also look into the existing test coverage for code generation in serverpod/test, especially around comparing generated server/client code, and suggest a test case that checks whether the server-side initializer is correctly present and omitted from the client.

Will push the YAML + follow-up test shortly.

tp: @ashishexee Thanks, let me know if I can help in any way.

No need to close this PR, you can just keep pushing new commits, and we can even rework the commit history if need be (does not need to be perfect from the start).

If you think about working on an open PR for a longer time, it makes sense to set it back to "draft" though, to signal that it's not yet suited to review.

Looking forward to what you come up with :)

Issue #3489, opened 2025-04-29
Ensure `serverpod generate` checks `pubspec` against CLI version and handles dependency bounds tightly


dkbast:
## Problem to Solve  
After upgrading a Serverpod project or upgrading the CLI, there can be version mismatches between the CLI and the project’s dependencies.  
Also, when using `dart pub add`, packages are added with loose version bounds, leading to inconsistencies between the project and the CLI. ( This is only relevant for first party serverpod packages)

## Proposal  
- During `serverpod generate`, the CLI should check if the `pubspec.yaml` and `pubspec.lock` file match the CLI version.
- If a mismatch is detected, notify the user and suggest corrective actions.
- Investigate a robust method to run `dart pub get` during generation:
  - Prefer using the local pub cache.
  - Force a remote fetch if versions do not match the CLI.
- Additionally, enforce tight version bounds when adding dependencies to ensure alignment with the CLI expectations.

## Use Case  
A developer upgrades Serverpod and adds a new package using `dart pub add`. Without this feature:
- Dependencies could drift from the CLI version.
- Generation might fail or produce unexpected behavior due to mismatched APIs. E.g. the generator uses APIs not available in the project.

With this feature:
- Dependency consistency is maintained automatically.
- Upgrades become more reliable and predictable.

SandPod:
We already have a warning in place, but the warning is not correctly displayed if the user has not run `dart pub get`.

I think we can take a simple approach here and just ensure that the warning correctly matches towards the actually cached serverpod version and not the one specified in the `pubspec.yaml` file.

PR #3720, opened 2025-06-26
feat: Warn users about outdated lock file

exaby73:
This PR adds a warning if the lock file of either server or client is different to the CLI version. It conditionally prints this warning only if there are no `pubspec.yaml` warnings. This means the user has the correct version in `pubspec.yaml` but they haven't run `dart pub get`

#3489

## Pre-launch Checklist


tools/serverpod_cli/lib/src/commands/generate.dart
```diff
@@ -81,6 +83,38 @@ class GenerateCommand extends ServerpodCommand<GenerateOption> {
...
+      var serverLockFile = File('pubspec.lock');
+      var clientLockFile = File(path.joinAll([
+        ...config.clientPackagePathParts,
+        'pubspec.lock',
+      ]));
+
+      var lockFilesToCheck = [
+        if (await serverLockFile.exists()) serverLockFile,
+        if (await clientLockFile.exists()) clientLockFile,
+      ];
```
tp:
```suggestion
      ].map(PubspecLockParser.fromFile);
```

Use the same pattern as above?

No strong feelings on this, but it just seemed like it could look the same.

tools/serverpod_cli/lib/src/commands/generate.dart
```diff
@@ -81,6 +83,38 @@ class GenerateCommand extends ServerpodCommand<GenerateOption> {
...
+
+      var lockWarnings = <SourceSpanSeverityException>[];
+      for (var lockFile in lockFilesToCheck) {
```
tp:
```suggestion
      var lockWarnings = <SourceSpanSeverityException>[
         for (var lockParser in lockFilesToCheck)
             ...validateServerpodLockFileVersion(cliVersion, lockParser)         
      ];

```

Then we could also use the same pattern for the list building, which I think it quite nice (or use `expand`).

Or is the `try` really important here? I would think if that file was not a lock file after all, that's probably rare and points to a problem, which we might not want to silence (at least not without any hint). (What would `pub get` do? Fail or overwrite it?)

tools/serverpod_cli/lib/src/util/pubspec_lock_parser.dart
```diff
@@ -0,0 +1,252 @@
...
+
+  /// Whether this is a dev dependency
+  bool get isDevDependency => dependencyType == DependencyType.directDev;
+
+  /// Whether this is a transitive dependency
+  bool get isTransitive => dependencyType == DependencyType.transitive;
+}
+
+/// Parser for pubspec.lock files
+class PubspecLockParser {
```
nielsenko, line 89: Did you consider using an existing package such as [pubspec_lock](https://pub.dev/packages/pubspec_lock)? If so, what made you roll your own?

tools/serverpod_cli/lib/src/serverpod_packages_version_check/serverpod_packages_version_check.dart
```diff
@@ -33,6 +46,26 @@ List<SourceSpanSeverityException> validateServerpodPackagesVersion(
   ];
 }
 
+List<SourceSpanSeverityException> validateServerpodLockFileVersion(
```
tp: If you return at most one warning, why not make this a single return then, and use `return` in the loop instead of building the list (up to 1 element) and then `break`?

tools/serverpod_cli/lib/src/util/pubspec_lock_parser.dart
```diff
@@ -0,0 +1,252 @@
...
+  final Map<String, LockedPackage> packages;
+  final String? dartSdkConstraint;
```
tp: Would [VersionConstraint](https://pub.dev/documentation/pub_semver/latest/pub_semver/VersionConstraint-class.html) from [pub_semver](https://pub.dev/documentation/pub_semver) fit well here?

It's already a dependency of the CLI, and since you also parse things like `DependencyType` I thought it might make sense to give a "friendly" variant here as well.

tools/serverpod_cli/lib/src/util/pubspec_lock_parser.dart
```diff
@@ -0,0 +1,252 @@
...
+    final versionStr = packageData['version'] as String;
+    final version = Version.parse(versionStr);
+
+    final descriptionYaml = packageData['description'];
+
+    // Parse dependency based on source type
+    final Dependency dependency;
+    String? sha256;
+
+    switch (source) {
```
tp, line 166: Make this an enum, like `DependencyType`?

(Since we don't handle arbitrary strings.)

tools/serverpod_cli/lib/src/util/pubspec_lock_parser.dart
```diff
@@ -0,0 +1,252 @@
...
+  /// Get all packages of a specific type
+  List<LockedPackage> getPackagesByType(DependencyType type) =>
+      packages.values.where((p) => p.dependencyType == type).toList();
+
+  /// Get all packages from a specific source
+  List<LockedPackage> getPackagesBySource(String source) =>
+      packages.values.where((p) => p.source == source).toList();
+
+  /// Get all hosted packages
+  List<LockedPackage> get hostedPackages => getPackagesBySource('hosted');
```
tp: because the enum could then also be used here

tools/serverpod_cli/test/util/pubspec_lock_parser_test.dart
```diff
@@ -0,0 +1,429 @@
...
+        final analyzer = packagesByName['analyzer']!;
+        expect(analyzer.dependencyType, DependencyType.directMain);
+        expect(analyzer.isDirect, true);
+        expect(analyzer.isDevDependency, false);
+        expect(analyzer.isTransitive, false);
+      });
+
+      test('then analyzer package has correct version', () {
+        final analyzer = packagesByName['analyzer']!;
+        expect(analyzer.version, Version.parse('6.11.0'));
```
tp, line 76: This makes the think, we could provide the version outright as a `Version`, no?

I find it hard to draw the line of what should stay as the original "string" and what do we parse properly, but since we already parse so much (and not currently critical stuff), I think we can do that here as well.

tools/serverpod_cli/test/util/pubspec_lock_parser_test.dart
```diff
@@ -0,0 +1,429 @@
...
+
+  group('Given a pubspec.lock with path dependencies', () {
+    const lockString = '''
+packages:
+  my_local_package:
+    dependency: "direct main"
+    description:
+      path: "../my_local_package"
+    source: path
+    version: "1.0.0"
```
tp, line 163: In this case the version could be misleading though, right?

If the lock file is still saying 1.0.0, but I have 2.0.0 in `my_local_package`, that'll just use 2.0.0 then at runtime.

tools/serverpod_cli/lib/src/util/pubspec_lock_parser.dart
exaby73: I did look at it. I felt the API weird, and also it gives packages as a list which I felt was better as a map instead. Since it's not very complicated to parse, I opted for implementing a simpler API backed by a map instead. Do you feel we should use this package instead?

tp: That specific one is a bit heavy on the dependencies (as it can also write lock files, etc.).

I also did not see a great public package. Surprisingly, Dart lang only offers a `pubspec.yaml` parser, nothing for the lock file (that I could find, though I suppose some helper is in there somewhere).

tools/serverpod_cli/lib/src/serverpod_packages_version_check/serverpod_packages_version_check.dart
exaby73: We could have 2 warnings, no? One for `serverpod` and another for `serverpod_client`?

tp: But you `break` on the first warning in line 62.

So the way I see it the list would always have 0 or 1 element.

tools/serverpod_cli/test/util/pubspec_lock_parser_test.dart
exaby73: Could you elaborate a bit on what you mean here? I'm a bit confused

tools/serverpod_cli/test/util/pubspec_lock_parser_test.dart
exaby73: That's how the lock file is generated by Dart. I guess it extracts the version from the package's `pubspec.yaml` and populates it in the lock file so it should be in sync unless the user modified it for some reason in which case I don't think we can do anything there if the user is messing with generated files

tools/serverpod_cli/lib/src/util/pubspec_lock_parser.dart
exaby73: So agreed to use the current implementation?

tools/serverpod_cli/test/util/pubspec_lock_parser_test.dart
tp: My bad, I misread the order. It is already a `Version` here :)

tools/serverpod_cli/test/util/pubspec_lock_parser_test.dart
tp: Yeah, it's probably just some "good to know" stuff.

I would also start out like you did and rely on the lock file and print the warning accordingly if it does not match.

tools/serverpod_cli/lib/src/util/pubspec_lock_parser.dart
tp: I think it's great, and at least as nice as anything I've seen elsewhere (and feels much easier to work on, as it's plain Dart, no code gen, no magic).

tools/serverpod_cli/lib/src/commands/generate.dart
```diff
@@ -81,6 +82,31 @@ class GenerateCommand extends ServerpodCommand<GenerateOption> {
...
+
+      var lockFilesToCheck = [
+        if (await serverLockFile.exists()) serverLockFile,
+        if (await clientLockFile.exists()) clientLockFile,
+      ].map(PubspecLockParser.fromFile);
+
+      var lockWarnings = [
+        for (var lockParser in lockFilesToCheck)
+          validateServerpodLockFileVersion(cliVersion, lockParser),
+      ];
```
tp: Ah, I see how I made collecting the warning a bit harder with the new return value suggestion.

Maybe add 

```suggestion
      ].nonNulls;
```

so this really only contains warnings?

tools/serverpod_cli/lib/src/util/pubspec_lock_parser.dart
```diff
@@ -0,0 +1,280 @@
...
+
+    final descriptionYaml = packageData['description'];
+
+    // Parse dependency based on source type
+    final Dependency dependency;
+    String? sha256;
+
+    switch (source) {
+      case PackageSource.hosted:
+        if (descriptionYaml is String) {
```
tp, line 168: Is this really a thing for `hosted` ones? I only saw them for SDK ones:

CleanShot 2025-06-26 at 11 18 07@2x

tools/serverpod_cli/lib/src/util/pubspec_lock_parser.dart
```diff
@@ -0,0 +1,280 @@
...
+          // Simple hosted dependency
+          dependency = HostedDependency(
+            version: VersionConstraint.compatibleWith(version),
```
tp, line 171: Was confused why this was not a specific `Version` (which is what the lock file is all about), but since this is shared with the `pubspec.yaml` parser, it has a range.

But I suppose that's fine, as this can be ignored and we rely on the top-leve `LockedPackage`'s `version` anyway.

tools/serverpod_cli/lib/src/util/pubspec_lock_parser.dart
exaby73: Yes indeed :D

Issue #1651, opened 2023-12-08
Session manager listeners are called multiple times, even when nothing changes


lukehutch:
**Describe the bug**

When a user is already logged in, the session manager listeners are called twice on app startup, both times with the login state being the same (once from `_loadSharedPrefs`, then once from `refreshSession`). This breaks the `ChangeNotifier` convention of only notifying listeners when something actually changes.

This is kind of a pain, because my login system relies on being notified just once when a user is logged in -- so I have to de-dup the change listener calls, using `ValueNotifier` (which properly follows the `ChangeNotifier` contract of only notifying when changes are actually made):

```dart
final SessionManager sessionManager =
    SessionManager(caller: client.modules.auth)
      ..addListener(() =>
          _sessionManagerSignedInUser.value = sessionManager.signedInUser?.id);

// Need a separate ValueNotifier to de-dup session manager updates
final _sessionManagerSignedInUser = ValueNotifier<int?>(null)
  ..addListener(_sessionManagerSignedInStateListener);

void _sessionManagerSignedInUserListener() async {
  if (_sessionManagerSignedInUser.value != null) {
    // ...
  }
}
```

tp:
@lukehutch We could implement deep-equality on `UserInfo` in order to avoid the `listener` being called without any changes.

I think your mapping into a `ValueNotifier` (or similarly a `ValueListenable<int?> userId`) still does make sense though, as the `UserInfo` object is quite large, and there could be changes to it that are just "minor" updates and not a login/logout: for example when the user object retrieved from the server differs from the one read from local storage.

In that case (e.g. a name or profile image change) we would still need to notify the outside of that change.

So I think you approach of focusing only on the `userId` or a `isLoggedIn` state is a good one for many apps, to only update specifically when the primary condition changes.

tp:
The upcoming `serverpod_auth_session_flutter` package will offer a new `SessionManager` implementation that uses a `ValueNotifier` for the "auth info" (user ID + scopes), which will only update on login/logout, when those values actually change.

PR #3721, opened 2025-06-26
feat(auth2): Add client `SessionManager` for `serverpod_auth_session`

tp:
Implemented as package `serverpod_auth_session_flutter`.

Add a simple "session manager" to be given to the Serverpod `Client` (to retrieve the auth key for the next request).

Unlike the legacy one in `serverpod_auth`, this one does not handle the sign out itself, but rather relies on being notified of a logout (just like both of them do for a login).

Like the legacy variant, this takes 2 storages: One "secure" one (just for the session key), and a default one for user info (ID and scopes for now). Hopefully, this is a bit clearer in this implementation.
By default only 1 store is used for both cases, which are backed by `shared_preferences`.
(In the legacy module there were 2 slightly distinct instances, but in the end both would also write into `shared_preferences` by default.)

In the future we could replace the "secure" one with a Keychain etc. implementation.

This is only a minimal implementation to get started in the current version. In the future the storage interface should be slightly reworked for 3.0 as outlined in https://github.com/serverpod/serverpod/pull/3713.

The newly added `ValueListenable<AuthInfo?>` (replacing the entire session manager being a `ChangeNotifier` before) will fix https://github.com/serverpod/serverpod/issues/1651, and only notify the outside when the login status actually changed.

## Summary by CodeRabbit

* **New Features**
  * Introduced the `serverpod_auth_session_flutter` package, providing user session management and authentication capabilities for Flutter apps using Serverpod.
  * Added a `SessionManager` for handling session state, login, and logout, with support for persistent storage via shared preferences.
  * Exposed a key-value storage interface for customizable session data storage.

* **Documentation**
  * Added comprehensive README and changelog files detailing usage, features, and version history.

* **Tests**
  * Implemented unit tests to verify session management functionality and storage behavior.

* **Chores**
  * Added configuration files for analysis, metadata, and package management.
  * Updated utility scripts to support the new package.


tp:
@coderabbitai full review

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/key_value_storage.dart
```diff
@@ -0,0 +1,10 @@
+import 'dart:async';
+
+/// String-based key/value store.
+abstract class KeyValueStorage {
+  /// Gets the stored value for [key].
+  FutureOr<String?> get(String key);
+
+  /// Sets [key] to the new [value].
+  FutureOr set(String key, String? value);
```
nielsenko:
```suggestion
  FutureOr<void> set(String key, String? value);
```
.. as opposed to `dynamic`

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/session_manager.dart
```diff
@@ -0,0 +1,155 @@
...
+  String? _key;
+
+  var _init = false;
+
+  /// Read the configuration from the storage
+  Future<void> init() async {
+    if (_init) {
+      throw RepeatedSessionManagerInitError();
+    }
+    _init = true;
```
nielsenko: Another approach is to make the `init` function private, ie. `_init` and have `init` be a `Future<void>` instance, that is set in ctor `init = _init()`. Then you await on init multiple times, but be sure we only ran the init code once, and hence avoid this if-bad-then-exception.

nielsenko: You can even hide `init` from the public interface and just have `setLoggedIn` and `logout` await it internally.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/session_manager.dart
```diff
@@ -0,0 +1,155 @@
...
+      _authInfo.value = (
+        authUserId: UuidValue.fromString(persistedUserId),
+        scopeNames:
+            (jsonDecode(persistedScopeNames) as List).cast<String>().toSet(),
+      );
+    }
+  }
+
+  /// Set the current session to a logged-in user described by [authSuccess].
+  Future<void> setLoggedIn(AuthSuccess authSuccess) async {
```
nielsenko, line 84: Should this name be symmetric to `logout`, ie. `login`or `setLoggedOut`?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/session_manager.dart
```diff
@@ -0,0 +1,155 @@
...
+  Future<void> put(String key) {
+    // This is never called by the Serverpod `Client`, and thus no need to implement it.
+    throw UnimplementedError();
+  }
+
+  @override
+  Future<void> remove() {
+    // This is never called by the Serverpod `Client`, and thus no need to implement it.
+    throw UnimplementedError();
+  }
```
nielsenko, line 130: Have you created a follow-up issue for this yet?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/session_manager.dart
```diff
@@ -0,0 +1,155 @@
...
+
+/// Info about the logged-in user.
+typedef AuthInfo = ({UuidValue authUserId, Set<String> scopeNames});
+
+@internal
+abstract class SessionManagerStorageKeys {
+  static const sessionKeyStorageKey = 'serverpod_auth_session.sessionKey';
+  static const authUserIdStorageKey = 'serverpod_auth_session.authUserId';
+  static const scopeNamesStorageKey = 'serverpod_auth_session.scopeNames';
+}
```
nielsenko: Perhaps use an enum:
```dart
enum SessionManagerStorageKeys {
  sessionKeyStorageKey('serverpod_auth_session.sessionKey'),
  authUserIdStorageKey('serverpod_auth_session.authUserId'),
  scopeNamesStorageKey('serverpod_auth_session.scopeNames');

  const SessionManagerStorageKeys(this.name);
  final String name;
}
```

nielsenko (approved):
LGTM 🚀 I have some suggestions you can consider

nielsenko:
> @coderabbitai full review

:thinking: Not sure why that stalled. Try just:
@coderabbitai review

tp:
@coderabbit full review

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/session_manager.dart
tp: Would be nice in the test to then iterate over all the known ones easily in the "success case".

Will do.

(On the downside I feel like the access name are already too long, but in another PR we had the discussion that we should not have `@visibleForTesting` fields on the class, as external packages could get it (if they don't use IDE nor lint for it). So I just followed that pattern here again.)

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/session_manager.dart
tp: This will not be implemented, but rather `put` and `remove` should be removed from the interface passed to the Serverpod `Client` (as they never use this, and there is also no case for them going forward (as they would not even know the key to put for example).

This is already underway in https://github.com/serverpod/serverpod/pull/3713. My proposal for the new API is in there already, the tests need a few more extensions, mostly around making sure to still support older clients.

If we then proceed with such an approach, we'd just move the `SessionManager` to the new interface and not have these methods anymore :)

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/session_manager.dart
tp: Absolutely makes sense.

I would probably opt for the `set*` version, as this is more a container and does not "log in" anyone.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/session_manager.dart
tp: `init` is called when the object is created, to rehydrate the stored state (not after login/logout).

I was probably pushed a bit into this pattern from the legacy `SessionManager`, which also did not init right away (i.e. in some async factory), but required you to call it (which also means you could opt out of session restoration if you didn't…).

I like the idea with the "retained" `Future`, so that you could call it multiple times, and it would always just give the some (hopefully success) result.

For this specific class it should be very easy though to ensure it's only called once in the initial client setup code, so it kinda feels like a (non-critical) oversight on the client if they would call it again.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/key_value_storage.dart
tp: Interesting that the default client (Flutter) linter config from us does not include the "no implicit dynamic" one.

I was so used to that always being active I forgot that forgetting a "open" type annotation was even a thing anymore.

Will fix.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/test/serverpod_auth_session_flutter_test.dart
```diff
@@ -0,0 +1,197 @@
+import 'dart:async';
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:serverpod_auth_session_client/serverpod_auth_session_client.dart';
+import 'package:serverpod_auth_session_flutter/serverpod_auth_session_flutter.dart';
+import 'package:serverpod_auth_session_flutter/src/session_manager.dart';
+
```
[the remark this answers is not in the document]
tp: package `uuid` is re-exported through `serverpod_auth_session_client` which re-exports `serverpod_client` which re-exports `serverpod_serialization` which re-exports `uuid`.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/session_manager.dart
```diff
@@ -0,0 +1,155 @@
...
+    _storage.set(
+      SessionManagerStorageKeys.authUserIdStorageKey,
+      authSuccess.authUserId.uuid,
+    );
+
+    _storage.set(
+      SessionManagerStorageKeys.scopeNamesStorageKey,
+      jsonEncode(authSuccess.scopeNames.toList()),
+    );
+  }
```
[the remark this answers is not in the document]
tp: Good catch. Why doesn't the https://dart.dev/tools/linter-rules/unawaited_futures analyzer rule catch the unawaited `FutureOr`?

Is there another rule we can add to catch these?

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/session_manager.dart
tp: Hmm, I also kind of like the idea of calling "init" after login/logout.

So we basically just set the storage keys and then read them. Technically a little slower, but it would consistently exercise the same code-paths so we know that a later rehydration would work as well.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/key_value_storage.dart
tp: Type has been added. As mentioned in stand-up, `FutureOr` seems a little bad here as there is no linter guranteeing that it has been awaited.

I'll just change them to `Future` real quick, as any practical implementation will be using that anyway.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/session_manager.dart
tp: Removed the `exception` and also not calling an `_init` internally, so we can be sure that rehydration works the same way as the runtime changes.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/session_manager.dart
tp: created the enum.

util/pub_get_all
```diff
@@ -57,6 +57,7 @@ declare -a dart_paths=(
   "modules/new_serverpod_auth/serverpod_auth_user/serverpod_auth_user_server"
   "modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_client"
   "modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server"
+  "modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter"
   "modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_client"
   "modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server"
```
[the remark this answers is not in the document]
tp: Moved this, as it does indeed make more sense in the Flutter package section.

But locally and in CI `dart pub get` works just fine there inside the package as well.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/lib/src/session_manager.dart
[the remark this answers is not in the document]
tp: Interface has been changed.

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/test/serverpod_auth_session_flutter_test.dart
```diff
@@ -0,0 +1,197 @@
...
+      expect(
+        defaultStorage.values,
+        {
+          SessionManagerStorageKeys.authUserIdStorageKey:
+              _authSucces.authUserId.toString(),
+          SessionManagerStorageKeys.scopeNamesStorageKey: '["test1","test2"]'
+        },
+      );
+    });
```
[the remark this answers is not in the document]
tp: Auth info value is now checked in this case as well.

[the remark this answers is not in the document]
tp: @coderabbitai There was I typo in my response. I meant to say that it's now tested.

Can you recheck and see if you would suggest any further test cases?

PR #3723, opened 2025-06-27
fix(auth2): Use validating UUID parsers in extensions methods

tp:
As otherwise any string generates some "UUID value".

Gleamed from https://github.com/serverpod/serverpod/pull/3721 which exposed the same issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation when accessing user session and identifier properties to ensure only valid UUIDs are accepted, resulting in clearer error handling for invalid or missing values.

* **Tests**
  * Added new tests to verify correct behavior and error handling when accessing user session and identifier properties with various types of input.


tp:
@coderabbitai full review

exaby73 (approved):
LGTM

PR #3724, opened 2025-06-27
fix(auth): GoogleAuth.clientSecret not initialized when trying to authenticate from the client

### LINKED ISSUE #3722: Google client secret not loading in serverpod 2.9.0
### Describe the bug

I upgraded to serverpod 2.9.0 and doing so broke my sign-in-with-google flow.

the error I get is: "**The server side Google client secret is not loaded.**" from google_endpoint.dart, and it's seemingly because **nothing in the serverpod 2.9.0 codeflow appears to set the static GoogleAuth.clientSecret field**.

I can work around the issue by running this code at the end of my server.dart run method, which works because authClientForUser calls GoogleAuth._loadClientSecret.

Future<void> _googleAuthHack() async {
  await serverpod.withSession((session) async {
    try {
      logger.d("Runing _googleAuthHack");
      **await auth.GoogleAuth.authClientForUser(session, 1);**
    } catch (e, trace) {
      logger.e("googleAuthHack: $e", error: e, stackTrace: trace);
    }
  });
}

pls lmk if I'm missing anything or if this is an actual serverpod bug

btw 'withSession' is my own extension method for convenience:

extension ServerpodExtension on Serverpod {
  Future<T> withSession<T>(Future<T> Function(Session) func) async {
    final session = await createSession(enableLogging: true);
    try {
      return await func(session);
    } finally {
      await session.close(); // Clean up the session
    }
  }
}

### To reproduce

set up a project to use google authentication, including adding a google_client_secret.json to serverpod, then try to log in with google from the client-side. 

the client-side login nearly seems to complete, but then you get a 500 error. 

### Expected behavior

the google auth flow should complete normally. 

### Library version

2.9.0

### Platform information

testing end-to-end on localhost, on windows 11.

### Additional context

_No response_

None

- [ ] I want to work on this

exaby73:
Fix the issue when `GoogleEndpoint.authticate*` methods access `GoogleAuth.clientSecret` before `GoogleAuth.authClientForUser` was called which previously initialized `clientSecret`. Due to the previous behaviour, the case was that `clientSecret` would always be `null` preventing authenticating with Google. This should be fixed in this PR which adds back initialization of `clientSecret` immediately in an intializer.

Fixes #3722

## Pre-launch Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved the loading process for Google authentication client secrets, ensuring they are initialized once at startup for better performance and reliability. No changes to user-facing features or workflows.

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_server/lib/src/business/google_auth.dart
+++ b/modules/serverpod_auth/serverpod_auth_server/lib/src/business/google_auth.dart
@@ -14,14 +14,15 @@ const _passwordKey = 'serverpod_auth_googleClientSecret';
 /// Convenience methods for handling authentication with Google and accessing
 /// Google's APIs.
 class GoogleAuth {
-  /// The client secret loaded from `config/google_client_secret.json`, null
-  /// if the client secrets failed to load.
-  static GoogleClientSecret? clientSecret;
+  /// The client secret loaded from `config/google_client_secret.json` or
+  /// password `serverpod_auth_googleClientSecret`, null if the client secrets
+  /// failed to load.
+  static final clientSecret = _loadClientSecret();
 
-  static GoogleClientSecret? _loadClientSecret(Session session) {
+  static GoogleClientSecret? _loadClientSecret() {
     try {
       late final String jsonData;
-      final password = session.passwords[_passwordKey];
+      final password = Serverpod.instance.getPassword(_passwordKey);
       if (password != null) {
         jsonData = password;
       } else {
@@ -55,12 +56,11 @@ class GoogleAuth {
         throw const FormatException('Missing "redirect_uris"');
       }
 
-      clientSecret = GoogleClientSecret._(
+      return GoogleClientSecret._(
         clientId: webClientId,
         clientSecret: webClientSecret,
         redirectUris: (webRedirectUris as List).cast<String>(),
       );
-      return clientSecret;
     } catch (e) {
       stderr.writeln(
         'serverpod_auth_server: Failed to load $_configFilePath or password $_passwordKey. Sign in with Google will be disabled. Error: $e',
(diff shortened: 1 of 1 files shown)


modules/serverpod_auth/serverpod_auth_server/lib/src/business/google_auth.dart
```diff
@@ -14,14 +14,15 @@ const _passwordKey = 'serverpod_auth_googleClientSecret';
 /// Convenience methods for handling authentication with Google and accessing
 /// Google's APIs.
 class GoogleAuth {
-  /// The client secret loaded from `config/google_client_secret.json`, null
-  /// if the client secrets failed to load.
-  static GoogleClientSecret? clientSecret;
+  /// The client secret loaded from `config/google_client_secret.json` or
+  /// password `serverpod_auth_googleClientSecret`, null if the client secrets
+  /// failed to load.
+  static final clientSecret = _loadClientSecret();
```
tp, line 20: So this was problematic for _another_ class wanting to access this and then getting `null`?

tp: Ah yes, see it in the endpoint now.

Looks good and makes sense :)

tp:
@coderabbitai full review

modules/serverpod_auth/serverpod_auth_server/lib/src/business/google_auth.dart
exaby73: Yes. My change to add password based config moved the intialization of this variable to `authClientForUser` which was an oversight

tp: (I also prefer that it can now be `final`, which makes it clear that this does not change later again (so others taking a reference to the instance is fine, as happens in the endpoint).)

exaby73:
@tp to answer your question, we currently do not have any tests for `GoogleAuth` at the moment. This feature for password reading was added with the intention of providing a simple API for password based secrets, with the idea that this would be better handled in auth 2 with better test coverage and better code architecture.

PR #3728, opened 2025-06-30
feat(auth2): Add hook for when user is migrated in `serverpod_auth_migration`

### LINKED ISSUE #3709: Provide common hooks for custom extensions of `AuthUser`
In case the developer wants to use a custom profile (or any other entity which should have a 1:1 relationship with an `AuthUser`) they would currently have to add their user profile migration / creation code in a bunch of places (e.g. the 2 variants of migrations, and the 2 variants of email sign ups (invites and registration)).

We should offer a common way to handle such side-effects, so that they would only have to register one function per package and can rely on the module to call into it whenever it's appropriate.

Related discussion: https://github.com/serverpod/serverpod/pull/3666#issuecomment-2999170155

tp:
Adds a hook to the `serverpod_auth_migration` package which is called once for each user migration (from `UserInfo` to `AuthUser`).

Apps that store data with a relation to the `UserInfo` should use this callback (in a `transaction`) to migrate that user's data to be reachable via the new `AuthUser` ID (as the mapping between the old and new user is bound to disappear once the migration is complete and the old tables are dropped).

Initially the idea was to return a `MigrationResult?` from the `migrateOnLogin` and `migrateWithoutPassword` method respectively, but that API proved hard to use in practice (because one would have to wrap it them in one's own function to not repeat oneself and pass in a transaction, but since the signatures are slightly different (the first one needs a `password` parameter) that could not be written very succinctly (without resorting to fallbacks like having an optional password parameter).  
The current approach is also more robust in that the custom migration can not be forgotten later on when adding new auth providers etc.
(The downside with "function configurations" is always the initial setup, and that the code/flow can not be read linearly, as the place they call out to is not immediately visible.)

Closes #3709


tp:
@coderabbitai full review

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/lib/src/business/auth_migration_config.dart
```diff
@@ -0,0 +1,35 @@
+import 'package:serverpod/serverpod.dart';
+
+/// Configuration options for the account migration.
+class AuthMigrationConfig {
```
tp, line 4: Well have to generalize this a bit with the next auth provider migration.

Not sure yet where to hook this up then. Maybe we make a general `AuthMigrations` class as the main entry point, as we do for the other new auth packages.

I think email might be the only one with 2 method, while the others would only add one each (as they don't have a password to optionally migrate).

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
```diff
@@ -125,6 +129,59 @@ void main() {
         },
       );
 
+      test(
+        'when running `migrateWithoutPassword` with a throwing `userMigrationHook`, then it forwards the error and does not create an `AuthUser`.',
+        () async {
+          AuthMigrationEmail.config = AuthMigrationConfig(
```
exaby73: Could `AuthMigrationEmail.config = ...` be extracted out in a `setUp` since all three tests share the same?

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
```diff
@@ -198,10 +255,22 @@ void main() {
       late Session session;
       late legacy_auth.UserInfo userInfo;
       late UuidValue authUserId;
+      late ({int oldUserId, UuidValue newAuthUserId})? migratedUser;
 
       setUp(() async {
         session = sessionBuilder.build();
 
+        AuthMigrationEmail.config = AuthMigrationConfig(
```
exaby73, line 274: Like you've done here

exaby73 (approved):
Small nit on tests but otherwise LGTM

modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/test/integration/auth_migration_email_test.dart
tp: Hmm, only 2 of the many tests in this group use the same throwing config.

But if we want consistent configuration within a test group, I can create a new `withServerpod` one that "always" has a throwing migration hook, and then contains only 2 `test`s. I suppose that could read nicer, as there is no more variance in the "environment" of the `tests`, but of course requires a bit more boilerplate.

exaby73: Ahh sorry yes there's only 2 tests, my bad. It's not bad right now, but it would read slightly better to have the throwing config shared in a group. Correct me if I am wrong, but I don't think you'd need an entire `withServerpod`, no? Wouldn't a group with a `setUp` and possibly a `tearDown` be enough?

tp: > Correct me if I am wrong, but I don't think you'd need an entire withServerpod, no? Wouldn't a group with a setUp and possibly a tearDown be enough?

Technically a nested `group` would be sufficient to only set and clean up the config for these 2 tests right.

It just doesn't seem that common right now. And maybe rightly so, as getting "initial setup" + "additional setup" + "actual test case" across 100+ lines (in this case) into ones head seems a bit complicated.

Usually when the setup become boilerplate and repetitive we do resort to a simple helper function like. E.g. in this case I might to well to create a `createLegacyAccount` to (repeatedly) save a few lines.

exaby73: I understand :) Your changes are great with the `withServerpod` :D

PR #3731, opened 2025-06-30
fix: serialization priority "extends" result

suojae:
## Fix getClassNameForObject serialization priority for inherited classes

This PR fixes a critical issue in the `getClassNameForObject` method where inherited classes were not properly serialized due to incorrect type checking order.

### Problem
When using inheritance with classes like `PaginatedBookResult extends BasePaginatedResponse`, the `getClassNameForObject` method would always return the parent class name instead of the concrete child class name. This happened because:

1. The method called `super.getClassNameForObject(data)` first
2. Parent class type checks were evaluated before child class checks
3. Since child classes are also instances of their parent classes, the parent class condition was always matched first

### Solution
- **Reordered type checking**: Current protocol's concrete classes are now checked before calling `super.getClassNameForObject(data)`
- **Proper inheritance handling**: Child classes are now checked in reverse order to ensure more specific types are matched before their base types

### Changes Made
1. Modified `getClassNameForObject` method generation in `library_generator.dart` to prioritize concrete class checks
2. Added comprehensive regression test to prevent future occurrences
3. Ensured proper serialization behavior for inheritance hierarchies

### Test Case
Added a regression test that verifies:
- Child class instances return their concrete class name (not parent class name)
- Proper order of type checking in generated code
- Fallback to parent protocol behavior when needed

Fixes issues with incorrect serialization of inherited model classes in client code generation.

## Breaking changes

No breaking changes. This fix maintains backward compatibility while correcting the serialization behavior for inherited classes.

### DIFF
--- a/tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
@@ -199,22 +199,13 @@ class LibraryGenerator {
           ..name = 'data'
           ..type = refer('Object?')))
         ..body = Block.of([
-          const Code(
-            'String? className = super.getClassNameForObject(data);'
-            'if(className != null) return className;',
-          ),
+          // Check current protocol's classes first to ensure concrete types take precedence
           for (var extraClass in config.extraClasses)
             Code.scope((a) =>
                 'if(data is ${a(extraClass.reference(serverCode, config: config))}) {return \'${extraClass.className}\';}'),
-          for (var classInfo in unsealedModels)
+          for (var classInfo in unsealedModels.reversed)
             Code.scope((a) =>
                 'if(data is ${a(refer(classInfo.className, TypeDefinition.getRef(classInfo)))}) {return \'${classInfo.className}\';}'),
-          if (config.name != 'serverpod' && serverCode)
-            _buildGetClassNameForObjectDelegation(
-                serverpodProtocolUrl(serverCode), 'serverpod'),
-          for (var module in config.modules)
-            _buildGetClassNameForObjectDelegation(
-                module.dartImportUrl(serverCode), module.name),
           for (var containerType in nonModelStreamTypes)
             Block.of([
               const Code('if(data is '),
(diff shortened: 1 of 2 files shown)


suojae:

suojae:
@coderabbitai review

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -199,22 +199,13 @@ class LibraryGenerator {
...
-          const Code(
-            'String? className = super.getClassNameForObject(data);'
-            'if(className != null) return className;',
-          ),
+          // Check current protocol's classes first to ensure concrete types take precedence
           for (var extraClass in config.extraClasses)
             Code.scope((a) =>
                 'if(data is ${a(extraClass.reference(serverCode, config: config))}) {return \'${extraClass.className}\';}'),
-          for (var classInfo in unsealedModels)
+          for (var classInfo in unsealedModels.reversed)
```
tp, line 206: So this is the fix basically?

Or does the moving up/down also have an effect?

I would be vary that this always works though, I think we should test A->B->C and C->B->A to ensure that the alphabetic names are not an issue and also that this works with a hierarchy of 3 (or more) as well.

tp (commented):
Overall great catch!

I will have to spend some time to dig into this deeper.

Maybe another (better?) approach would also to be to compare `runtimeType` instead of using `is`?
I will have to check whether we rely on / want to allow subclassing this such that such a check would break existing code. (Edit: If we do, it'd probably be wrong anyway, as we're talking about data classes here, and if an overwrite would add a new field, it would not be transmitted - so what's the point? (And you can even declare local only fields already if you need to store additional transient data. - but in that case we should probably marked the parent `sealed` and the children `final`))

Also I'll have to check the existing tests around this feature, and how we missed this so far (maybe just got lucky in there ordering, but then we should find out for what reason).

tp:
@suojae I approved the workflow runs. I would image this changes a bunch of generated code.

We can take care of that (for all sub projects) in the end, once we've figured out the best approach of how to solve this.

tp:
This has been fixed in https://github.com/serverpod/serverpod/pull/3732

PR #3732, opened 2025-06-30
fix(codegen): Fix `getClassNameForObject` for subclasses

### LINKED ISSUE #3730: PaginatedBookResult incorrectly serialized due to BasePaginatedResponse check order
### Describe the bug

Image

Image

<br/>

Image

<br/>
<br/>

### 📝 Issue Description

In the getClassNameForObject method within the Protocol class, the if (data is BasePaginatedResponse) condition is evaluated before the check for PaginatedBookResult. Since PaginatedBookResult extends BasePaginatedResponse, it always matches the base class condition first, leading to incorrect serialization.

### To reproduce

1. Suppose we have the following class structure:

   class Animal {}
   class Dog extends Animal {}

2. In the `getClassNameForObject` method, the checks are ordered like this:

   if (data is Animal) return "Animal";
   if (data is Dog) return "Dog";

3. Now create and pass an instance of Dog to the method.

4. Since Dog extends Animal, the first condition (`data is Animal`) evaluates to true and "Animal" is returned.

5. As a result, Dog is serialized with className: "Animal", causing incorrect deserialization later.

### Expected behavior

The method should prioritize subclass checks before superclass checks.

If an instance of `Dog` is passed, it should return "Dog" — not "Animal" — so that the full structure (including Dog-specific fields) can be correctly deserialized.

### Library version

2.9.0

### Platform information

Issue occurs in:
- Dart backend protocol class (serialization manager)
- Custom class serialization via `getClassNameForObject` method

### Additional context

_No response_

- [x] I want to work on this

tp:
Sorts classes in a hierarchy from child to parent, such that the `getClassNameForObject` gets the name of the most specific class, not a parent.

Closes #3730

### DIFF
--- a/tests/serverpod_test_client/lib/src/protocol/protocol.dart
+++ b/tests/serverpod_test_client/lib/src/protocol/protocol.dart
@@ -10,77 +10,77 @@
 
 // ignore_for_file: no_leading_underscores_for_library_prefixes
 import 'package:serverpod_client/serverpod_client.dart' as _i1;
-import 'by_index_enum_with_name_value.dart' as _i2;
-import 'by_name_enum_with_name_value.dart' as _i3;
-import 'changed_id_type/many_to_many/course.dart' as _i4;
-import 'changed_id_type/many_to_many/enrollment.dart' as _i5;
-import 'changed_id_type/many_to_many/student.dart' as _i6;
-import 'changed_id_type/nested_one_to_many/arena.dart' as _i7;
-import 'changed_id_type/nested_one_to_many/player.dart' as _i8;
-import 'changed_id_type/nested_one_to_many/team.dart' as _i9;
-import 'changed_id_type/one_to_many/comment.dart' as _i10;
-import 'changed_id_type/one_to_many/customer.dart' as _i11;
-import 'changed_id_type/one_to_many/order.dart' as _i12;
-import 'changed_id_type/one_to_one/address.dart' as _i13;
-import 'changed_id_type/one_to_one/citizen.dart' as _i14;
-import 'changed_id_type/one_to_one/company.dart' as _i15;
-import 'changed_id_type/one_to_one/town.dart' as _i16;
-import 'changed_id_type/self.dart' as _i17;
-import 'defaults/bigint/bigint_default.dart' as _i18;
-import 'defaults/bigint/bigint_default_mix.dart' as _i19;
-import 'defaults/bigint/bigint_default_model.dart' as _i20;
-import 'defaults/bigint/bigint_default_persist.dart' as _i21;
-import 'defaults/boolean/bool_default.dart' as _i22;
-import 'defaults/boolean/bool_default_mix.dart' as _i23;
-import 'defaults/boolean/bool_default_model.dart' as _i24;
-import 'defaults/boolean/bool_default_persist.dart' as _i25;
-import 'defaults/datetime/datetime_default.dart' as _i26;
-import 'defaults/datetime/datetime_default_mix.dart' as _i27;
-import 'defaults/datetime/datetime_default_model.dart' as _i28;
-import 'defaults/datetime/datetime_default_persist.dart' as _i29;
-import 'defaults/double/double_default.dart' as _i30;
-import 'defaults/double/double_default_mix.dart' as _i31;
-import 'defaults/double/double_default_model.dart' as _i32;
-import 'defaults/double/double_default_persist.dart' as _i33;
-import 'defaults/duration/duration_default.dart' as _i34;
-import 'defaults/duration/duration_default_mix.dart' as _i35;
-import 'defaults/duration/duration_default_model.dart' as _i36;
-import 'defaults/duration/duration_default_persist.dart' as _i37;
-import 'defaults/enum/enum_default.dart' as _i38;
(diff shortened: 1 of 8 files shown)


tests/serverpod_test_client/lib/src/protocol/protocol.dart
```diff
@@ -3701,28 +3701,28 @@ class Protocol extends _i1.SerializationManager {
     if (data is _i66.ExceptionWithData) {
       return 'ExceptionWithData';
     }
-    if (data is _i67.ChildClass) {
+    if (data.runtimeType == _i67.ChildClass) {
```
tp: Generating nested `if`s might read most clearly here.

tools/serverpod_cli/test/generator/dart/client_code_generator/protocol/inheritance_protocol_test.dart
```diff
@@ -248,6 +248,21 @@ void main() {
             isTrue,
           );
         });
+
+        test(
```
tp, line 252: This test currently fails on `main`.

tests/serverpod_test_client/lib/src/protocol/protocol.dart
```diff
@@ -345,224 +345,224 @@ class Protocol extends _i1.SerializationManager {
...
   ]) {
     t ??= T;
-    if (t == _i2.ByIndexEnumWithNameValue) {
-      return _i2.ByIndexEnumWithNameValue.fromJson(data) as T;
+    if (t == _i2.ScopeServerOnlyFieldChild) {
+      return _i2.ScopeServerOnlyFieldChild.fromJson(data) as T;
     }
-    if (t == _i3.ByNameEnumWithNameValue) {
-      return _i3.ByNameEnumWithNameValue.fromJson(data) as T;
+    if (t == _i3.ChildClass) {
```
tp, line 351: parent / child and sealed parent / sealed child are different hierarchies, so this is fine.

I also tried out the `switch` approach which did not have any more warnings with this sorting.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -1614,3 +1616,33 @@ extension on DatabaseDefinition {
...
   }
 }
+
+/// Sorts child classes before their parents, such that serialization order is stable.
+int _byChildClassesBeforeParents(
+  SerializableModelDefinition a,
+  SerializableModelDefinition b,
+) {
+  if (a is! ModelClassDefinition || b is! ModelClassDefinition) {
+    return 0;
```
nielsenko, line 1626: 🤔 .. seems odd

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -1614,3 +1616,33 @@ extension on DatabaseDefinition {
...
+
+  if (a.extendsClass != null && b.extendsClass == null) {
+    return -1;
+  }
+
+  if (a.extendsClass == null && b.extendsClass != null) {
+    return 1;
+  }
+
+  if (a.extendsClass is ResolvedInheritanceDefinition &&
```
nielsenko, line 1637: could use new if `if (x case T)` syntax to avoid later cast

nielsenko (approved):
LGTM!

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
tp: I don't think we would even have such instances in practice here, just the way the list is typed…

Only this subclass has the inheritance, so everything else we can just leave as is.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
tp: Right, that should be available now :)

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
nielsenko: Yes, I concluded that as well. Perhaps we could make capture that in the signature, or throw an exception here?

Anyway, not crucial.

Isakdl: There is an enum definition class that is not a ModelClassDefinition if I'm not wrong? or did we refactor that?

Isakdl (changes requested):
I suggest adding a test that runs the generated code and expects the model to resolve. E.g. replicate the bug in real running code not just by inspecting the generated code.

exaby73 (approved):
LGTM. Also agreed on @Isakdl's suggestion to add a test case for real code

tests/serverpod_test_server/test/class_name_for_object_test.dart
```diff
@@ -74,6 +75,35 @@ void main() {
     expect(clientName, serverName);
   });
 
+  test(
```
tp, line 78: @Isakdl Added this test case, which fails on `main` currently.

Isakdl (approved):
LGTM

PR #3738, opened 2025-07-01
fix(codegen): Fix code generation for type definitions from other folders

### LINKED ISSUE #3737: generated code imports wrong relative path when using dart Records
### Describe the bug

Now that Serverpod has suppord for Record types, I'm trying to use it in one of my yaml-defined models like so:

`  formattingData: Map<FormattingType, List<(int, TokenStringComponent)>>`

...where FormattingType and TokenStringComponent are serverpod-yaml-defined enums.

After code-generation with `serverpod generate`, the generated dart file for this model contains an incorrect import path, which causes an error.

`import 'enums/token_string_component_enum.dart' as _i6;`
When it should actually be:
`import '../enums/token_string_component_enum.dart' as _i6;`

Manually changing this resolves the error, but it's annoying to have to do this after EVERY `serverpod generate` and these generated files are not supposed to be manually edited. It is strange that it does this because it has always handled the other imports fine, and when I use this same `TokenStringComponent` enum outside of a Record, it also handles the import without problems. You can see in this screenshot that it's importing everything correctly with the correct relative path except, bizzarely, this one TokenStringComponent import.

Image

### To reproduce

Seems to be a problem with using a serverpod-defined type inside of a Record? Or is it just with using serverpod-defined enums inside of a Record? Or just a problem with all Record types in Serverpod? Not sure and realistically don't have further time now to continue to diagnose.

### Expected behavior

The generated files should not have errors that need manual correction.

### Library version

Serverpod 2.9.0

### Platform information

Flutter 3.32.5, Dart 3.8.1, on macos with cursor as editor

### Additional context

_No response_

- [ ] I want to work on this

tp:
By forwarding `subDirParts` parameter for types used in records

Closes #3737

### DIFF
--- a/tools/serverpod_cli/lib/src/generator/types.dart
+++ b/tools/serverpod_cli/lib/src/generator/types.dart
@@ -1,5 +1,5 @@
-import 'dart:core';
 import 'dart:core' as d;
+import 'dart:core';
 
 import 'package:analyzer/dart/element/nullability_suffix.dart';
 import 'package:analyzer/dart/element/type.dart';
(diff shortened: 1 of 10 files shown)


exaby73 (approved):
LGTM

tools/serverpod_cli/lib/src/generator/types.dart
```diff
@@ -270,17 +270,20 @@ class TypeDefinition {
...
-        b.positionalFieldTypes.addAll(
-          generics
-              .where((f) => f.recordFieldName == null)
-              .map((f) => f.reference(serverCode, config: config)),
-        );
+        b.positionalFieldTypes.addAll([
+          for (final f in positionalRecordFields)
+            f.reference(
+              serverCode,
+              subDirParts: subDirParts,
```
nielsenko, line 277: This is the crux of the fix right? Together with line 286?

nielsenko (approved):
👍

tools/serverpod_cli/lib/src/generator/types.dart
tp: Yes, and also the same one in the other file.

Always tricky with such parameters with default ones (which are not set on the outermost call). Personally, I often resort to `required` even if it'd be `null` in the rare case, then at least the call site is clear.

PR #3743, opened 2025-07-01
chore: Use `switch` inside `getClassNameForObject`

tp:
Follow up from https://github.com/serverpod/serverpod/pull/3732


tp:
@coderabbit full review

exaby73 (approved):
LGTM

examples/auth_example/auth_example_server/lib/src/generated/protocol.dart
```diff
@@ -52,8 +52,9 @@ class Protocol extends _i1.SerializationManagerServer {
   String? getClassNameForObject(Object? data) {
     String? className = super.getClassNameForObject(data);
     if (className != null) return className;
-    if (data is _i4.Example) {
-      return 'Example';
+    switch (data) {
+      case _i4.Example():
+        return 'Example';
```
nielsenko, line 57: Should we consider using a switch expression instead?
```dart
return switch(data) {
  _i4.Example() => 'Example',
};
```

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -205,12 +205,18 @@ class LibraryGenerator {
...
+            for (var extraClass in config.extraClasses)
+              Code.scope((a) =>
+                  'case ${a(extraClass.reference(serverCode, config: config))}():'
+                  '  return \'${extraClass.className}\';'),
+            for (var classInfo in unsealedModels)
+              Code.scope((a) =>
+                  'case ${a(refer(classInfo.className, TypeDefinition.getRef(classInfo)))}():'
+                  '  return \'${classInfo.className}\';'),
+            const Code('}'),
+          ],
```
nielsenko, line 219: Consider generating a switch expression instead. It will be shorter in the generated code.

examples/auth_example/auth_example_server/lib/src/generated/protocol.dart
tp: @nielsenko But we don't switch exhaustively, but rather have to defer to other modules further down if we don't "own" the class in this package.

tp: Or we would have to stuff all the module stuff into the fallback?

but then that gets quite long and nested 🤷🏻‍♂️

nielsenko: You could generate it like this:
```dart
String? getClassNameForObject(Object? data) {
  return super.getClassNameForObject(data) ??
      switch (data) {
        _i2.Example() => 'Example',
        _ => null,
      } ??
      _i3.Protocol()
          .getClassNameForObject(data)
          ?.replaceFirst('', 'serverpod_auth.');
}
```
not sure if it is preferable though 🤔 .

nielsenko (approved):
LGTM

examples/auth_example/auth_example_server/lib/src/generated/protocol.dart
tp: Yeah, there's potentially multiple external modules and then even more fallbacks (in case of streamed container types).

So I think I would leave it like this, as it already gives us the improved check at 1/3 less lines of code :)

PR #3752, opened 2025-07-03
fix: Correctly generate code for scope=serverOnly fields with a default value

### LINKED ISSUE #3424: Invalid client code generated when combining `scope=serverOnly` and `default` value in model definition

ashishexee:
🔧 **Fix**: Prevent invalid client code generation for `serverOnly` fields with default values
📌 **Related Issue**: [[serverpod/serverpod#3424](https://github.com/serverpod/serverpod/issues/3424)](https://github.com/serverpod/serverpod/issues/3424)

### 🐞 Problem Summary

When a model field is marked with both `scope=serverOnly` and a `default` value, the generated client-side constructor incorrectly tries to access that field. This results in a Dart **compilation error**.

**Example Model:**

```yaml
class: FavouriteRecipe
table: favourite_recipe
fields:
  recipe: Recipe
  userId: int?, scope=serverOnly, default=-1
```

**Broken Client Constructor:**

```dart
FavouriteRecipe._({
  this.id,
  required this.recipe,
}) : userId = userId ?? -1; // ❌ Compilation Error
```

**Compiler Error:**

```
'userId' isn't a field in the enclosing class.
Undefined name 'userId'.
```

---

### ✅ Fix Summary

This PR updates the code generation logic to ensure that **default value initializers are only added** to the constructor **if the field is included in the current codegen scope** (e.g., not `serverOnly` for client-side models).

**Modified Code Logic:**

```dart
var defaultValueFields = classFields.where(
  (field) => field.hasDefaults && field.shouldIncludeField(serverCode),
);

for (var field in defaultValueFields) {
  Code? defaultCode = _getDefaultValue(field, subDirParts: subDirParts);
  if (defaultCode == null) continue;

  c.initializers.add(Block.of([
    refer(field.name).code,
    const Code('='),
    refer(field.name).ifNullThen(CodeExpression(defaultCode)).code,
  ]));
}
```

---

### 🧪 Tests Added

* **Client Code Generation**: Ensures `serverOnly` fields with defaults don't trigger invalid constructor code.
* **Server Code Generation**: Verifies correct handling of such fields in server-side models.
* **Exception Handling**: Confirms proper generation when these fields appear in exceptions.
* **Analyzer Validation**: Ensures no warnings or errors for valid `serverOnly + default` combinations.
* **Integration**: End-to-end tests to confirm the fix holds up under full app scenarios.

---

Closes #3424

## Summary by CodeRabbit

* **New Features**
  * Introduced support for server-only fields with default values in data models.
  * Added tests to validate correct handling of server-only fields with default values in both server and client code generation.

* **Bug Fixes**
  * Improved filtering to ensure default values are only applied to fields relevant for the current code generation context.

* **Tests**
  * Added comprehensive tests for server and client code generation involving server-only fields with defaults.
  * Included analyzer tests to verify validation and correct recognition of default values and scopes.


ashishexee:

tp:
@coderabbitai full review

ashishexee (commented):
The "id" field was nullable, but the keyword "defaultModel" ensures that it will always have a value, unless explicitly removed. made it non-nullable to avoid unnecessary null checks.

tp:
> The "id" field was nullable, but the keyword "defaultModel" ensures that it will always have a value, unless explicitly removed. made it non-nullable to avoid unnecessary null checks.

Okay. I know the warning, but I always assumed a test would rely on this.

You can't non set in anyway initially (as the constructor uses `id ?? Uuid()`), but I suppose before you could overwrite it with `null` via `copyWith`.

But fine with me if nothing relies on that and this was just an oversight. (Apart from the fact that we don't need this in here and just need to be careful how much this affects code-gen, or if we better just live with the warning.)

---

I'll give the rest an in-depth later :)

ashishexee:
> > The "id" field was nullable, but the keyword "defaultModel" ensures that it will always have a value, unless explicitly removed. made it non-nullable to avoid unnecessary null checks.
> 
> Okay. I know the warning, but I always assumed a test would rely on this.
> 
> You can't non set in anyway initially (as the constructor uses `id ?? Uuid()`), but I suppose before you could overwrite it with `null` via `copyWith`.
> 
> But fine with me if nothing relies on that and this was just an oversight. (Apart from the fact that we don't need this in here and just need to be careful how much this affects code-gen, or if we better just live with the warning.)
> 
> I'll give the rest an in-depth later :)

Oh I didn't knew that I saw the warning and didn't thought too much and fixed that

tests/serverpod_test_server/test/defaults/server_only/server_only_default_test.dart
```diff
@@ -0,0 +1,26 @@
+import 'package:test/test.dart';
+
+void main() async {
+  group(
+    "Given a class with serverOnly scoped fields with default values",
+    () {
+      test(
+        'when an object is created without specifying serverOnly field, then the default value is used',
+        () {
+          // This test would need a corresponding model in the test protocol
```
tp: I think after code-gen you can just make use of the `ServerOnlyDefault` test class here, right?

tp:
@ashishexee Let me know if you can fix the formatting and testing. I would probably revert the unrelated model change here, but generate for the new one and test that.

Else I can also push a tiny "fix CI" commit. As the rest already looks quite good :)

ashishexee:
> @ashishexee Let me know if you can fix the formatting and testing. I would probably revert the unrelated model change here, but generate for the new one and test that.
> 
> Else I can also push a tiny "fix CI" commit. As the rest already looks quite good :)

the formatting i have fixed (hopefully) but I could use a little help in the testing I am kind of stuck here

ashishexee:
> > @ashishexee Let me know if you can fix the formatting and testing. I would probably revert the unrelated model change here, but generate for the new one and test that.
> > Else I can also push a tiny "fix CI" commit. As the rest already looks quite good :)
> 
> the formatting i have fixed (hopefully) but I could use a little help in the testing I am kind of stuck here

shall I push the final code with the changes and then maybe you could take it from here?

tp:
> shall I push the final code with the changes and then maybe you could take it from here?

@ashishexee Sounds like a plan :)

tests/serverpod_test_server/test/defaults/server_only/server_only_default_test.dart
```diff
@@ -1,26 +1,54 @@
...
+            serverOnlyStringField: 'Custom message',
+          );
+
+          expect(serverObject.normalField, 'test value');
+          expect(serverObject.serverOnlyField, 42);
+          expect(serverObject.serverOnlyStringField, 'Custom message');
+        },
+      );
+
+      test(
```
tp, line 36: At first I thought re-testing `copyWith` was superfluous, but this brought up an interesting case: For fields with `default` `copyWith` behaves (in my opinion) unexpectedly.

When you explicitly pass `null` (as opposed to the implicit "undefined") for a nullable field, then the return object still gets the default value.

E.g. the following test would fail (as the fields have the default value):

```

      test(
        'when creating a new object with copyWith, then serverOnly can be set to `null`.',
        () {
          var original = ServerOnlyDefault(
            normalField: 'original',
          );

          var copied = original.copyWith(
            normalField: 'updated',
            serverOnlyField: null,
            serverOnlyStringField: null,
          );

          expect(copied.normalField, 'updated');
          expect(copied.serverOnlyField, null);
          expect(copied.serverOnlyStringField, null);
        },
      );
```

Probably too big do discuss in here, but I'll create a ticket for it.

tests/serverpod_test_server/test/defaults/server_only/server_only_default_test.dart
tp: Thanks for adding these!

tests/serverpod_test_server/test/defaults/server_only/server_only_default_test.dart
tp: -> https://github.com/serverpod/serverpod/issues/3757

tp (approved):

tp:
Thanks @ashishexee for seeing this through.

I really only had to generate the code once, as your tests were already good (however you did that 😅)!

PR #3759, opened 2025-07-04
feat: Better error messages for missing passwords

### LINKED ISSUE #3684: Improve feedback when the passwords file is missing, and offer a way to recreate it
### Problem to Solve

This is currently the output when the `config/passwords.yaml` file is missing. It may be hard for the user to know which steps are required to get things up and running. This frequently happens when the user checks out a project from git, because the passwords file is never checked in.

```
SERVERPOD version: 2.8.0, dart: 3.7.2 (stable) (Tue Mar 11 04:27:50 2025 -0700) on "macos_arm64", time: 2025-06-18 13:07:54.610955Z
mode: development, role: monolith, logging: normal, serverId: default
Unhandled exception:
Exception: Missing database password.
#0      new DatabaseConfig._fromJson (package:serverpod_shared/src/config.dart:363:7)
#1      new ServerpodConfig.loadFromMap (package:serverpod_shared/src/config.dart:141:26)
#2      new ServerpodConfig.load (package:serverpod_shared/src/config.dart:213:30)
#3      Serverpod._initializeServerpod (package:serverpod/src/server/serverpod.dart:335:25)
#4      new Serverpod (package:serverpod/src/server/serverpod.dart:308:5)
#5      run (package:test_2_server/server.dart:15:15)
#6      main (file:///Users/viktor/Projects/tests/test_2/test_2_server/bin/main.dart:6:3)
#7      _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:313:33)
#8      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:194:12)
``` 

### Proposal

Minimally: Clean up the output, and add a better error message with instructions for how to setup the passwords file. E.g.:

```
SERVERPOD version: 2.8.0, dart: 3.7.2 (stable) (Tue Mar 11 04:27:50 2025 -0700) on "macos_arm64", time: 2025-06-18 13:07:54.610955Z
mode: development, role: monolith, logging: normal, serverId: default
The config/passwords.yaml file is missing. It's required to start the server. See the documentation to troubleshoot: https://docs.serverpod.dev/troubleshoot/missing-passwords
```

Maximally, create a new `serverpod create --local-passwords .` command. The command would create a new default passwords file for local development. It would find the Postgres and Redis password from the docker compose file and use it for creating a minimal `passwords.yaml` file that matches the Docker containers.

### Use Case

This would greatly improve the user experience with example projects that require the database to be present.

### Alternatives

Leave as is.

### Additional context

_No response_

None

- [ ] I want to work on this

exaby73:
This PR adds a new `PasswordMissingException` that is caught by `Serverpod` which prints a useful message to the user about missing passwords for config. We use this exception to throw on missing passwords for database and redis configs, which shows a better message for users.

#### Before

```console
Unhandled exception:
Exception: Missing database password.
#0      new DatabaseConfig._fromJson (package:serverpod_shared/src/config.dart:363:7)
#1      new ServerpodConfig.loadFromMap (package:serverpod_shared/src/config.dart:141:26)
#2      new ServerpodConfig.load (package:serverpod_shared/src/config.dart:213:30)
#3      Serverpod._initializeServerpod (package:serverpod/src/server/serverpod.dart:335:25)
#4      new Serverpod (package:serverpod/src/server/serverpod.dart:308:5)
#5      run (package:test_2_server/server.dart:15:15)
#6      main (file:///Users/viktor/Projects/tests/test_2/test_2_server/bin/main.dart:6:3)
#7      _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:313:33)
#8      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:194:12)
```

#### After

```console
Missing password for "database". Please check your config/passwords.yaml file or the SERVERPOD_PASSWORD_database environment variable.
```

Closes #3684

## Pre-launch Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages for missing database and Redis passwords, providing more specific and user-friendly feedback during server configuration.

* **Documentation**
  * Updated documentation to clarify password loading behavior in configuration management.

* **Tests**
  * Enhanced and expanded tests to verify detailed error handling for missing passwords and Redis configuration scenarios, ensuring robust validation and accurate error reporting.

### DIFF
--- a/packages/serverpod_shared/lib/src/config.dart
+++ b/packages/serverpod_shared/lib/src/config.dart
@@ -438,7 +438,6 @@ class DatabaseConfig {
     if (password == null) {
       throw PasswordMissingException(
         ServerpodPassword.databasePassword.configKey,
-        'SERVERPOD_PASSWORD_database',
       );
     }
 
@@ -517,7 +516,6 @@ class RedisConfig {
     if (password == null) {
       throw PasswordMissingException(
         ServerpodPassword.redisPassword.configKey,
-        'SERVERPOD_PASSWORD_redis',
       );
     }
 
--- a/packages/serverpod_shared/lib/src/exceptions/password_missing_exceptions.dart
+++ b/packages/serverpod_shared/lib/src/exceptions/password_missing_exceptions.dart
@@ -6,10 +6,10 @@ base class PasswordMissingException extends ExitException {
   final String passwordName;
 
   /// The name of the environment variable that is missing.
-  final String envVarName;
+  String get envVarName => 'SERVERPOD_PASSWORD_$passwordName';
 
   /// Creates a new instance of [PasswordMissingException].
-  PasswordMissingException(this.passwordName, this.envVarName) : super(1);
+  PasswordMissingException(this.passwordName) : super(1);
 
   /// The user-friendly message of the exception.
   @override


exaby73:
@coderabbitai full review

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -436,7 +436,10 @@ class DatabaseConfig {
 
     var password = passwords[ServerpodPassword.databasePassword.configKey];
     if (password == null) {
-      throw Exception('Missing database password.');
+      throw DatabasePasswordMissingException(
+        ServerpodPassword.databasePassword.configKey,
+        'SERVERPOD_PASSWORD_database',
```
tp: The `ENV` key can always be derived from the YAML config one, right?

Then maybe we should not take an extra parameter for now.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -401,14 +411,17 @@ class Serverpod {
...
-        );
+    try {
+      this.config = config?.copyWith(runMode: runMode) ??
+          ServerpodConfig.load(
+            runMode,
+            _commandLineArgs.getRaw<String>(CliArgsConstants.serverId),
+            _passwords,
+            commandLineArgs: _commandLineArgs.toMap(),
+          );
+    } on PasswordMissingException catch (e) {
```
tp: What do you think of having "missing critical" exceptions (in Dart parlance they should probably be errors, as they can not be fixed without a "code" change (though in our case it's the config…)) just extend `ExitException` ?

That way we don't have to special case and remap in the cases where such an exception might be thrown.

packages/serverpod_shared/lib/src/exceptions/password_missing_exceptions.dart
```diff
@@ -0,0 +1,34 @@
...
+      '$envVarName environment variable.';
+
+  @override
+  String toString() => '$runtimeType: The password "$passwordName" or '
+      'the environment variable "$envVarName" is missing.';
+}
+
+/// Exception thrown when a database password is missing from the configuration
+/// or the environment variable.
+final class DatabasePasswordMissingException extends PasswordMissingException {
```
tp: Are these subclasses worth it?

I think for the auth module I would just stick to the base one (okay I can't since it's abstract), as I don't see what other useful information I could add.

The only thing that comes to mind would be specifying the `module` that threw this, but then again the affected one should be very clear from the password entry's name. (Even more so in the new modules, with their `$module_$key` pattern.)

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -374,11 +374,21 @@ class Serverpod {
...
+        experimentalFeatures: experimentalFeatures,
+      );
+    } on ExitException catch (e) {
+      if (e.message.isNotEmpty) {
+        stderr.writeln(e.message);
+      }
+      exit(e.exitCode);
+    } catch (e, stackTrace) {
+      _reportException(e, stackTrace, message: 'Error initializing Serverpod');
+      throw ExitException(1, 'Serverpod initialization failed: $e');
```
tp: Wouldn't this now get printed twice?

(here and in the parent, which usually does not have any exception handling the main method and thus just crashes).

Probably not a huge deal, but I would currently thing that print + exit is the better choice here then as well. (matching the later `init`)

Also for logging, it might be easier to identify the issue with the original exception than having it wrapped and the type discarded here, no?

tp (commented):
Some questions on the implementation.

Additional tests look great :)

packages/serverpod/lib/src/server/serverpod.dart
exaby73: No because we catch the `ExitException` and print it's message

tp: > Wouldn't this now get printed twice?

> No because we catch the ExitException and print it's message

But that's what I mean. First through `_reportException` and then higher up by catching the `ExitException`.

exaby73: Currently, the error messages look like:

```console
SERVERPOD version: 2.9.1, dart: 3.8.0 (stable) (Wed May 14 09:07:14 2025 -0700) on "macos_arm64", time: 2025-07-07 07:57:41.728921Z
Missing password for "database". Please check your config/passwords.yaml file or the SERVERPOD_PASSWORD_database environment variable.
```

The `ExitException` thrown falls through, and the user is responsible for catching it else it's unhandled in the current code flow

packages/serverpod_shared/lib/src/config.dart
exaby73: For ones that already had an env key (before the move to `SERVERPOD_PASSWORD_*`), we still have the old keys for backward compatibility. So for database, it's `SERVERPOD_DATABASE_PASSWORD`. Ideally, we want to give as much information to the user about what went wrong, no? Which is why I added the `envVarName` field to print this information

packages/serverpod/lib/src/server/serverpod.dart
exaby73: Yeah I think this is a good idea :) Thanks!

packages/serverpod_shared/lib/src/config.dart
tp: > So for database, it's SERVERPOD_DATABASE_PASSWORD

I did not know about the old keys, but in any case the code above is already using the new format (`SERVERPOD_PASSWORD_database`).

So I am a bit confused why we need that parameter, or should we actually be using the old key (`SERVERPOD_DATABASE_PASSWORD`) here then?

packages/serverpod/lib/src/server/serverpod.dart
tp: Yeah, I guess I mean that this will just bubble up and get printed again (toString + StackTrace) anyway already, right? (If you don't do the `report + exit(1)` pattern.)

tp:
```suggestion
      exit(1);
```

Since the error + stack is already printed, and thus would align with the behavior of `Serverpod.start()`

Issue #3753, opened 2025-07-03
Separate "migration" and "legacy support"


tp:
Currently, the way `serverpod_auth_migration` is designed, driven by some challenges with the legacy data[^1], a developer would have to keep a dependency on it to support eventual migrations, of e.g. the email passwords, in their endpoints.

But since the migration package has a dependency on the legacy `serverpod_auth` and all its counterpart packages from the new `serverpod_auth_*` packages, the migration part drags on for a long time, as it's only "done" when all account details have been fully migrated (some currently requiring a login into that account, to e.g. get the old password), or when the developer decides that "enough" accounts have been migrated to drop the old tables and thus any chance of further imports.

Thus we should split the migration and legacy support into 2 distinct packages and steps.  
**The goal** is to have the migration finished as fast as possible, and then be able to completely drop `serverpod_auth` from the codebase (also transitively!) so the developer can be sure that they are not relying on any old code anymore (and for example the legacy auth endpoints disappear as well as the legacy tables).

---

### Packages

`server_auth_migration`:

Should only support the actual migration of the data, from `serverpod_auth` to the new packages, and a new "legacy support" one for data that can not be fully migrated.

To run the migration as a one-off (by default) we can discard the current configuration approach and instead just provide a single function (which runs a single transaction) to run the transaction. This function would take the configuration (whether to import the profile for example) and a callback for any custom migrations to be done.

Dependencies:
  - `serverpod_auth`
  - `serverpod_auth_legacy_support`
  - `serverpod_auth_user`
  - `serverpod_auth_session`
  - `serverpod_auth_profile` (might not be used, depending on the configuration)
  - `serverpod_auth_email`, … (Apple, Firebase, Google account packages)

Tables:
  - `MigratedUser` for the mapping of legacy to new user IDs

`serverpod_auth_legacy_support`:

Supports ongoing look-ups / migrations for legacy user data which has not yet or can not be migrated to the new system.
For example:
- Legacy sessions will be stored here indefinitely and not be migrated to the new `serverpod_auth_session`, as there is no way to tell the client about the new session key (and we don't want to support the legacy format in the new system)
  - We could add a "last used at" column there (which corse resolution), so developers could easily identify unused legacy session and thus find out when these are no longer practically useful
- Email account hashes in the old format. This package would have a copy of the old hashing algorithm to check the password on `login` endpoint calls. If it's correct, the new endpoint will set the incoming password for the email account.
[...]

PR #3761, opened 2025-07-05
feat(auth2): Add `serverpod_auth` backwards compatibility module

tp:
Add a new module `serverpod_auth_backwards_compatibility` which handles the ongoing backwards compatibility of legacy sessions, email accounts, and social logins created by `serverpod_auth`.

Relates to #3753 (which also explains the rationale in detail)


modules/new_serverpod_auth/serverpod_auth_backwards_compatibility/serverpod_auth_backwards_compatibility_server/lib/src/business/config.dart
```diff
@@ -0,0 +1,11 @@
+/// Configuration for the `servpod_auth` backwards compability module.
+class AuthBackwardsCompatibilityConfig {
+  /// True if the server should use the accounts email address as part of the
```
exaby73:
```suggestion
  /// True if the server should use the account's email address as part of the
```

PR #3763, opened 2025-07-07
fix: CLI args doesn't override config when the config is passed into Serverpod as an object

### LINKED ISSUE #3762: CLI arguments do not override the config when the config is passed into `Serverpod`'s constructor
### Describe the bug

In the following code, `applyMigrations` should be `true` as CLI args should take priority over the config object.

```dart
final config = ServerpodConfig(
  applyMigrations: false,
);

final serverpod = Serverpod(
  ['--apply-migrations'],
  Protocol(),
  Endpoints(),
  config: config,
)
```

But this is not currently the case as we do not call `ServerodCofig.load` when the user has passed in their own config where `.load()` figures out the value to be used based on available values from CLI args, env and yaml configuration.

### To reproduce

```dart
final config = ServerpodConfig(
  applyMigrations: false,
);

final serverpod = Serverpod(
  ['--apply-migrations'],
  Protocol(),
  Endpoints(),
  config: config,
)
```

### Expected behavior

Migrations should be applied on startup

### Library version

2.9.1

### Platform information

Any

### Additional context

_No response_

- [x] I want to work on this

exaby73:
This PR fixes an issue where a user passing in a `ServerpodConfig` object to `Serverpod` does not have it's values overriden by CLI arguments

Fixes #3762

## Pre-launch Checklist


packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -393,16 +393,38 @@ class Serverpod {
...
+          runMode: runMode,
+          serverId: serverId,
+          loggingMode: loggingMode,
+          role: role,
+          applyMigrations: applyMigrations,
+          applyRepairMigration: applyRepairMigration,
+        ) ??
         ServerpodConfig.load(
           runMode,
           _commandLineArgs.getRaw<String>(CliArgsConstants.serverId),
```
tp: Now that the `serverId` is already parsed beforehand, I think we could replace this with the following, right?

```suggestion
          serverId,
```

exaby73: Great catch!

christerswahn:
Just a general comment: Before we spend too much time on maintaining the existing Serverpod config code, we should migrate it to use the Config library (in cli_tools).

tests/serverpod_test_server/test_integration/serverpod_cli_args_override_test.dart
```diff
@@ -0,0 +1,342 @@
...
+import 'package:serverpod_shared/serverpod_shared.dart';
+import 'package:serverpod_test_server/src/generated/endpoints.dart';
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+import 'package:test/test.dart';
+
+void main() {
+  // These tests verify that command line arguments correctly override
+  // ServerpodConfig values when both are provided to the Serverpod constructor.
+  // Note: runMode has special handling where it's calculated from CLI args first,
+  // then environment variables, then defaults to 'development'.
```
Isakdl: Suggestion: Remove this description; this type of information should be embedded in the test description/structure.

tests/serverpod_test_server/test_integration/serverpod_cli_args_override_test.dart
```diff
@@ -0,0 +1,342 @@
...
+      );
+
+      expect(serverpod.config.runMode, equals('production'));
+      // Other values should remain from config
+      expect(serverpod.config.serverId, equals('test-server'));
+      expect(serverpod.config.role, equals(ServerpodRole.monolith));
+      expect(serverpod.config.loggingMode, equals(ServerpodLoggingMode.normal));
+      expect(serverpod.config.applyMigrations, isFalse);
+      expect(serverpod.config.applyRepairMigration, isFalse);
+    });
```
Isakdl: Suggestion: Do not validate other fields in a test that is for testing the production mode. They are an indirect validation and has nothing to do with the check we are validating now. If we split them we validate their rules in isolation, meaning we do not have to look at this test or modify this test if we want to change the rules for an unrelated field.

Each command line variable should have a positive and a negative test. E.g. what happens when we supply a command line argument for that specific variable and what happens when we don't

tests/serverpod_test_server/test_integration/serverpod_cli_args_override_test.dart
```diff
@@ -0,0 +1,342 @@
...
+      expect(serverpod.config.applyMigrations, isTrue);
+      // Config values (not overridden)
+      expect(serverpod.config.role, equals(ServerpodRole.monolith));
+      expect(serverpod.config.loggingMode, equals(ServerpodLoggingMode.normal));
+      expect(serverpod.config.applyRepairMigration, isFalse);
+    });
+
+    test(
+        'when no CLI args are provided and only config is given '
+        'then server uses config values except runMode which defaults to development',
```
Isakdl: Question: This sounds wrong? If I'm supplying the config with a runMode and I'm not overriding that with a commandline it should keep the runMode no?

tests/serverpod_test_server/test_integration/serverpod_cli_args_override_test.dart
```diff
@@ -0,0 +1,342 @@
...
+        applyMigrations: false,
+        applyRepairMigration: false,
+        apiServer: ServerConfig(
+          port: 8080,
+          publicHost: 'localhost',
+          publicPort: 8080,
+          publicScheme: 'http',
+        ),
+        maxRequestSize: 524288,
+      );
```
Isakdl: Suggestion: Setup a test builder (this might already exist!) and use that instead of constructing the object manually like this. 

The benefit is that if we ever modify the ServerpodConfig object with new fields (especially mandatory fields), these tests will not have to be modified. 
Another benefit is that we can ignore all fields that we are not explicitly testing for, just use the `with` method for the field we want to validate. This shrinks the number of lines in the setup here making it easier to view the full test.

tests/serverpod_test_server/test_integration/serverpod_cli_args_override_test.dart
```diff
@@ -0,0 +1,342 @@
...
+
+      // All values should come from config since CLI parsing fails
+      expect(serverpod.config.runMode, equals('development'));
+      expect(serverpod.config.serverId, equals('config-server'));
+      expect(serverpod.config.role, equals(ServerpodRole.monolith));
+      expect(serverpod.config.loggingMode, equals(ServerpodLoggingMode.normal));
+      expect(serverpod.config.applyMigrations, isTrue);
+      expect(serverpod.config.applyRepairMigration, isFalse);
+    });
```
Isakdl: Suggestion: We don't necessarily need to check all the fields in this test, I suppose the reason for this test is to validate that we are not getting the default config instead of the Dart one?

tests/serverpod_test_server/test_integration/utils/serverpod_config_builder.dart
```diff
@@ -0,0 +1,87 @@
...
+  String _runMode = 'development';
+  String _serverId = 'test-server';
+  ServerpodRole _role = ServerpodRole.monolith;
+  ServerpodLoggingMode _loggingMode = ServerpodLoggingMode.normal;
+  bool _applyMigrations = false;
+  bool _applyRepairMigration = false;
+  int _maxRequestSize = 524288;
+  ServerConfig? _apiServer;
+  ServerConfig? _insightsServer;
+  ServerConfig? _webServer;
```
Isakdl, line 19: Suggestion: I think it makes sense to have a default for all of these to. 

I believe there are more configuration options available the serverpod config object too that should be present here.

Isakdl: Another question: Don't we already have a "builder" folder somewhere in the test project? If so please put this class in there.

exaby73: There's a future call builder in this utils directory so I added this next to that file

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -107,10 +100,18 @@ class ServerpodConfig {
...
     );
   }
 
+  /// Creates a default API server configuration.
+  static ServerConfig createDefaultApiServer() => ServerConfig(
+        port: 8080,
+        publicHost: 'localhost',
+        publicPort: 8080,
+        publicScheme: 'http',
+      );
```
Isakdl: Suggestion: I don't think this was necessary, however what I meant in my original comment was to also supply a default config for insights and for the webserver. This is the normal behavior for serverpod to have all three configured. It is an outlier to not do that. Since you can supply null in the builder you can remove that config if need be.

Isakdl: All other configurations should also come with reasonable defaults, e.g. for the database, future call, etc. Ideally using builders to construct the nested objects too.

exaby73: The issue with `apiServer` is that it's a required parameter so it has to be passed. Since we already had a function that supplies this default config for `apiServer`, I felt it was better to reuse that than to duplicate code. What we could do is mark that static method as `@visibleForTesting`. Do you suggest we remove it instead?

Isakdl: I do not think production code should be used in the test builders. No need for that! The point of these builders is to  create valid objects and has reasonable defaults, but they don't necessarily need to match what is in production. In fact I think it is good if they don't as you run the risk of just testing your production happy path. 

`@visibleForTesting` should be avoided; we do not want to test internals, as that then leaks information about the implementation.

PR #3765, opened 2025-07-08
feat(auth2): Rewrite auth migration/backwards compatibility to focus on one-off migrations

### LINKED ISSUE #3753: Separate "migration" and "legacy support"

tp:
To keep the overall migration (including the app's custom entities) manageable.
This implementation will require that clients also be updated in conjunction with the completion of the server migration. Backwards compatibility is still given for sessions, so that updated clients will continue to work against the updated server without needing a re-login.

This removes support for the "continuous" migration in the endpoints' `login` methods (which might have proved untenable anyway, if too much related data would have to be copied per user).

Upon migration legacy sessions are now imported into the `severpod_auth_backwards_compatibility` module, and will then be converted by a client API call into a proper `serverpod_auth_session` one. This improves on the previous behavior where the re-mapped legacy session was kept around (and thus the backwards compatibility code would have to be kept running much longer; now it seems reasonable to delete those session after say 30 days, as the client has not been updated and seems unused).

Closes #3753

Related docs PR: https://github.com/serverpod/serverpod_docs/pull/302

## Summary by CodeRabbit

* **New Features**
  * Introduced a Flutter package for backwards compatibility with Serverpod authentication, including session migration from legacy systems.
  * Added session migration endpoint and utilities for seamless transition of user sessions and passwords.
  * Provided a new extension for importing legacy sessions in Flutter clients.
  * Added a password-importing email account endpoint to integrate legacy password migration during login and registration.

* **Bug Fixes**
  * Improved handling and migration of legacy passwords and sessions.

* **Documentation**
  * Added README and changelog for the new Flutter compatibility package.

* **Tests**
  * Added and updated integration tests for session and password migration, including Flutter client tests.
  * Added new tests verifying session migration and password backward compatibility.

* **Chores**
  * Updated dependencies and scripts to support the new Flutter compatibility package.
  * Enhanced configuration and analysis options for new modules.


tp:
@coderabbitai full review

tp:
@coderabbitai full review

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/session_manager_migration_test.dart
```diff
@@ -0,0 +1,142 @@
...
+import 'package:serverpod_auth_session_flutter/serverpod_auth_session_flutter.dart'
+    as auth_session_flutter;
+import 'package:serverpod_auth_shared_flutter/serverpod_auth_shared_flutter.dart'
+    as legacy_auth_flutter;
+import 'package:serverpod_new_auth_test_client/serverpod_new_auth_test_server_client.dart';
+
+import 'utils/test_storage.dart';
+
+void main() {
+  test(
```
exaby73, line 13: I'm wondering if this test could be split up into multiple ones

tp: Filling up the old session manager, and then using the new one is actually all part of the same test.

I just use the `expects` in-between to find errors earlier.

I guess the legacy session manager could become part of a `setUp`, so at least the rest becomes shorter. I would not want to stuff more into helpers for now (as we have just 1 big integration test here), as the details are really import and should not be hidden away in a `doMigration(); expect(state, …)`.

But I'll think about how I can add more clarity in here :)

PR #3768, opened 2025-07-09
feat(auth): Add ability to prevent new registration/login in `serverpod_auth` endpoints

tp:
So during a migration (which should only last a few seconds, but the deployment can live a bit longer) we can prevent new sign-ups/sign-ins.
This prevents issues with the post-migration deploy when the database schema is updated again to expect new the UUID user ids.


tp:
This is not sufficient, as the application's own endpoints (or even future calls) could still create new entities.

So basically we should disable any writes which are not the actual migration during the migration step. I'll see that we can make that configurable in Serverpod.

PR #3771, opened 2025-07-09
feat: Allow disabling of the API endpoint via the Serverpod config

### LINKED ISSUE #3769: Disable API endpoints and future calls during migration
While the migration is running, we need to prevent any API endpoint or future call from creating new data which then might not be migrated properly (e.g. if the user has already been migrated, and there is an entity linked to them created afterwards).

- [ ] Add ability to disable API server (for maintenance)
- [ ] Add ability to disable future calls
- [ ] Update auth -> auth2 migration docs, to disable these 2 features in the first "migration deployment"

tp:
Closes #3769

### DIFF
--- a/tests/serverpod_test_server/lib/server.dart
+++ b/tests/serverpod_test_server/lib/server.dart
@@ -15,6 +15,17 @@ void run(List<String> args) async {
     Protocol(),
     Endpoints(),
     authenticationHandler: auth.authenticationHandler,
+    config: ServerpodConfig(
+      apiServer: ServerConfig(
+        // TODO
+        port: port,
+        publicScheme: publicScheme,
+        publicHost: publicHost,
+        publicPort: publicPort,
+      ),
+      futureCallExecutionEnabled: false,
+      apiEndpointsEnabled: false,
+    ),
   );
 
   // Add future calls
--- a/packages/serverpod/lib/src/server/serverpod.dart
+++ b/packages/serverpod/lib/src/server/serverpod.dart
@@ -537,6 +537,7 @@ class Serverpod {
     _internalSession = InternalSession(server: server, enableLogging: false);
 
     if (Features.enableFutureCalls) {
+      // Note: This will only be run if `config.futureCallExecutionEnabled` is true
       _futureCallManager = FutureCallManager(
         server.serverpod.config.futureCall,
         serializationManager,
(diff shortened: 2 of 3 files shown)


tests/serverpod_test_server/lib/server.dart
```diff
@@ -15,6 +15,17 @@ void run(List<String> args) async {
     Protocol(),
     Endpoints(),
     authenticationHandler: auth.authenticationHandler,
+    config: ServerpodConfig(
+      apiServer: ServerConfig(
```
tp, line 19: I wonder why the `config` is optional for `Serverpod`, but then the `apiServer` is required inside the `ServerpodConfig`.

This would seem a bit hard to explain during the migration (in order to create a config in order to disable the endpoint). (Even though it is not even supposed to boot the endpoints, but maybe this would then fail the start up with a bogus config…)

tp: Is the problem (for the required API endpoint config) maybe this config behavior on main:

```
    this.config = config?.copyWith(runMode: runMode) ??
        ServerpodConfig.load(
          runMode,
          _commandLineArgs.getRaw<String>(CliArgsConstants.serverId),
          _passwords,
          commandLineArgs: _commandLineArgs.toMap(),
        );
```

Maybe instead we should take the config from `ServerpodConfig.load` and optionally merge "on top" the given one?

But of course in the end there's still the problem that the "manual" config (which I would more see as a partial override) should not require the API config itself, but in the end that must of course be present.

It's also weird that the initial config parameter is handed off to the `ExperimentalApi` helper, so in the end these 2 system could be using a different config…

@exaby73 Since you touched this a little bit, do you have more work upcoming on the config that could clarify the order of the overwrites and prevent the caller of the Serverpod constructor to provide an API Server config?

Else I may have to look for another way to disable the API server during the migration.

SandPod:
@tp - I don't think this is required anymore since we don't actually need to start the server in order to apply the migrations. Should we close this PR?


I think the desired way to disable the endpoints would be #3836

tp:
> @tp - I don't think this is required anymore since we don't actually need to start the server in order to apply the migrations. Should we close this PR?


Yes, I haven't tested it locally yet, but if we can do without `serverpod.start()` and just run the migration in that one-time deploy, this would not be needed, and we would not need to be worried about intermediate writes to the database (at least of only 1 server is active).

PR #3775, opened 2025-07-09
feat: WebServer port getter reflects actual port

christerswahn:
Code using the new WebServer of a Serverpod server needs to be able to get the actual port number the web server listens on, which is randomly allocated by the OS if it is configured as 0.

This is needed for Serverpod Cloud.

## Pre-launch Checklist

## Breaking changes

n/a

### DIFF
--- a/packages/serverpod/lib/src/web_server/web_server.dart
+++ b/packages/serverpod/lib/src/web_server/web_server.dart
@@ -15,11 +15,11 @@ class WebServer {
   /// The server id of this server.
   final String serverId;
 
-  /// The port the webserver is configured to run on.
-  late final int port;
+  /// The configuration for this web server.
+  late final ServerConfig _config;
 
-  /// The actual port the webserver is listening on, or null if not started.
-  int? get actualPort => _actualPort;
+  /// The port the webserver is listening on, or null if not started.
+  int? get port => _actualPort;
 
   int? _actualPort;
 
@@ -45,7 +45,7 @@ class WebServer {
       );
     }
 
-    port = config.port;
+    _config = config;
   }
 
   bool _running = false;
@@ -75,7 +75,7 @@ class WebServer {
       final server = await serve(
         _handleRequest,
         InternetAddress.anyIPv6,
-        port,
+        _config.port,
         context: context,
       );
       _server = server;


packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -15,9 +15,14 @@ class WebServer {
   /// The server id of this server.
   final String serverId;
 
-  /// The port the webserver is running on.
+  /// The port the webserver is configured to run on.
```
tp:
```suggestion
  /// The port the webserver is configured to run on.
  ///
  /// See [actualPort] for the port the webserver is listening on.
```

I wonder if this is not confusing though.

Should we just keep the nullable actual port, and the reader has to wait when it's actually up?

Else, if someone configures a port and then wants to use it somewhere else without waiting for the server to be up, they can just pass the config value there.

Or on the flip-side we could name this `preferredPort` and the actual port `port`.

Okay, a ton of bike-shedding, feel free to pick what's useful, if any :)

packages/serverpod/lib/src/server/server.dart
```diff
@@ -471,6 +471,7 @@ class Server {
     String? authenticationHeaderValue;
 
     try {
+      // TODO(kasper): Validate that this works with legacy auth header values.
```
tp: Why would this not work? Is there a ticket or test case?

I just mean I wouldn't know what legacy auth header values would mean here specifically.
Is this just for `serverpod_auth`? Then that should be proven to work fine by the end to end test, as only the server, not the client was touched in the relic upgrade, right?

christerswahn: This is actually unrelated to the auth modules, it pertains to users baking their own. I wrote an issue for it instead: https://github.com/serverpod/serverpod/issues/3776
Will remove this TODO.

packages/serverpod/lib/src/web_server/web_server.dart
christerswahn: Ideally IMO the configured port should be accessed via the configuration object, and `port` should be the actual port yes. I'll see how big a change that would be!

PR #3779, opened 2025-07-10
feat(auth2): Add combined tests for flow from migration to later usage via backwards compatibility module

tp:
Using the `withServerpod` integration tests, so this is not only covered in the E2E ones, which are harder to debug (where especially the test server setup is more complicated, as it handles various use-cases).

## Summary by CodeRabbit

* **Tests**
  * Added integration tests to verify legacy password import handles non-existent users without errors.
  * Introduced comprehensive tests for migrating legacy email-based user accounts, including scenarios for password import, migration hooks, profile handling, login validation, and error cases.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/migrations/20250711114550334/definition.json
+++ b/modules/new_serverpod_auth/serverpod_auth_migration/serverpod_auth_migration_server/migrations/20250711114550334/definition.json
@@ -1396,12 +1396,6 @@
           "dartType": "Set<String>"
         },
         {
-          "name": "lastUsed",
-          "columnType": 4,
-          "isNullable": true,
-          "dartType": "DateTime?"
-        },
-        {
           "name": "hash",
           "columnType": 0,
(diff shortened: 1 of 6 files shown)


tp:
@coderabbitai full review

modules/new_serverpod_auth/serverpod_auth_backwards_compatibility/serverpod_auth_backwards_compatibility_server/test/integration/backwards_compatibility_test.dart
```diff
@@ -0,0 +1,32 @@
...
+  withServerpod(
+    'Given no legacy passwords,',
+    (final sessionBuilder, final endpoints) {
+      late Session session;
+
+      setUp(() {
+        session = sessionBuilder.build();
+      });
+
+      test(
```
tp, line 17: the tests for actual "full imports" (migration + backwards compatibility) are below, as that can call into the real code and I don't have to mock the migrated stuff (which is not easy to verify then).

PR #3781, opened 2025-07-11
fix: Only download templates on `create` command

### LINKED ISSUE #3770: `serverpod` command should not download and create templates files unless needed
### Problem to Solve

The `serverpod` command always downloads the templates folder before it processes any options or subcommands. This takes up to 2 seconds (on a good internet connection) on every invocation. 

This is the case even when running diagnostic invocations such as `serverpod version` and `serverpod help` (and when running with incorrect args so serverpod fails with error message). But running `serverpod version` should be very fast and - it is run from scripts and CI pipelines and should not slow them down.

If there is currently no internet connection, this also causes the serverpod command to always generate an error message and exit with an error code:

```sh
$ serverpod version
Downloading templates for version 2.9.1
Failed to download templates.
```

This is incorrect as the exit code should reflect if the command succeeded or not. `serverpod version` and `serverpod help` should always succeed if the command is properly installed.

### Proposal

Only download templates if the command being run needs them.

If running offline, the command should succeed if templates were previously downloaded.

### Use Case

- Running `serverpod version` in scripts should be fast and reliable
- Working offline

### Alternatives

-

### Additional context

_No response_

- [ ] I want to work on this

exaby73:
Previously we downloaded all templates regardless of which command was run. This caused unecessary time to be wasted if all the user wanted to do is run `serverpod version` for example. This PR moves the template downloading logic to the `create` command specifically.

Fixes #3770

## Pre-launch Checklist

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/create.dart
+++ b/tools/serverpod_cli/lib/src/commands/create.dart
@@ -1,5 +1,6 @@
 import 'package:cli_tools/cli_tools.dart';
 import 'package:serverpod_cli/src/create/create.dart';
+import 'package:serverpod_cli/src/downloads/resource_manager.dart';
 import 'package:serverpod_cli/src/runner/serverpod_command.dart';
 import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart';
 
@@ -87,6 +88,22 @@ class CreateCommand extends ServerpodCommand<CreateOption> {
         'Use the --${CreateOption.force.option.argName} flag to force creation.',
       );
       throw ExitException.error();
+    }
+
+    // Make sure all necessary downloads are installed
+    if (!resourceManager.isTemplatesInstalled) {
+      try {
+        await resourceManager.installTemplates();
+      } catch (e) {
+        log.error('Failed to download templates.');
+        throw ExitException.error();
+      }
+
+      if (!resourceManager.isTemplatesInstalled) {
+        log.error(
+            'Could not download the required resources for Serverpod. Make sure that you are connected to the internet and that you are using the latest version of Serverpod.');
+        throw ExitException.error();
+      }
     }
 
     if (!await performCreate(name, template, force)) {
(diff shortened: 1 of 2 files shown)


tools/serverpod_cli/lib/src/commands/create.dart
```diff
@@ -89,6 +90,22 @@ class CreateCommand extends ServerpodCommand<CreateOption> {
...
+    // Make sure all necessary downloads are installed
+    if (!resourceManager.isTemplatesInstalled) {
+      try {
+        await resourceManager.installTemplates();
+      } catch (e) {
+        log.error('Failed to download templates.');
+        throw ExitException.error();
+      }
+
+      if (!resourceManager.isTemplatesInstalled) {
```
tp, line 102: Is this needed?

I think `installTemplate` already errors on each error, right?

The bigger problem (unrelated to this move) is that `isTemplatesInstalled` just checks the existence of the dir, not if the dir was ever written to successfully. 

Likewise `isTemplatesInstalled` does nothing if the dir exists already.

So I think in case the user is offline during the first run, this'll be broken "forever" (on that version).

Probably a better approach would be to use a temp download dir and then move that in the end (if that can be done atomically), or at least only create the dir when the file is received, so that the unzipping very likely succeeds.

tp (approved):
LGTM.

I think the moved code could be improved a bit.

(And then also the implementation.)

But for just fixing the original issue this'll do fine.

tools/serverpod_cli/lib/src/commands/create.dart
exaby73: Do you think this could be a separate issue which I could work on in a different PR or shall I fix this issue in the PR itself?

tp: Your call.

Feel free to isolate the fix in a later PR. We just shouldn't lose track of this, because apart from the code clarity issue here (just mildly annoying), the fact that a developer might get stuck on "not downloaded" templates seems a much bigger issue from which there is no obvious way to recover.

PS: Hence the initial ✅, which means overall fine, and any comments are just for your consideration.

exaby73: Created https://github.com/serverpod/serverpod/issues/3790 to track this issue. I'll pick it up soon

PR #3784, opened 2025-07-14
fix: Fixes MIME response types for HTML pages in the web server.

vlidholt:
This is a quick fix for the MIME type of web pages, which are currently returned as plain text, which breaks the default landing page for a new project.


Isakdl (approved):
Approving, but this solution seems rather brittle considering there are a lot of mime types that exists.

PR #3798, opened 2025-07-18
feat(auth2): Move AuthUser blocked and default scopes to `AuthSessions.createSession`

tp:
So it does not have to be repeated in every endpoint/business logic.


modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
```diff
@@ -142,16 +143,42 @@ abstract final class AuthSessions {
...
     final Transaction? transaction,
   }) async {
+    if (!skipUserBlockedChecked || scopes == null) {
+      final authUser = await AuthUsers.get(
+        session,
+        authUserId: authUserId,
+        transaction: transaction,
+      );
+
+      if (authUser.blocked) {
```
christerswahn: If skipUserBlockedChecked is true, but scopes is null, this check is performed anyway. Suggest including the !skipUserBlockedChecked condition in this if-clause as well!

christerswahn (commented):
Great, just one tweak!

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/auth_sessions.dart
tp: You're right.

If the caller really, truly wants to skip that check (and not just the performance penalty for reading from the DB) this is wrong.

Added a check that the blocked check is actually desired.

christerswahn (approved):
LGTM!

PR #3801, opened 2025-07-19
fix: Empty map deserialization bug

### LINKED ISSUE #1806: crashes with empty Map, causes type cast error
**Describe the bug**

According to the discussion here: https://github.com/serverpod/serverpod/pull/1799 having Map types with int keys is already supported even though the documentation has not yet been updated to reflect this.

However, hen using a serverpod-serializable class as defined in the yaml files, if a field is type Map with int keys (or enum keys whether they are serialized by name or by number) and then use it in any situation where this Map contains no entries and it needs to be deserialized, it triggers an exception.

The offending code appears to be in the `src/generated/protocol.dart` file where during deserialization, the Map data attempts to be type-cast from a Map to a List, as seen in this screenshot (note the "data as List"):

Screenshot 2024-01-23 at 2 16 21 PM

...which produces the error:

`type '_Map<String, dynamic>' is not a subtype of type 'List<dynamic>' in type cast`

Looking into things, it seems that map data is being stored behind the scenes in a format that is a List of map entries, with format:
`[{"k": somekey, "v":somevalue}, {"k": anotherkey, "v": anothervalue}]. So it would seem that since it's actually a list in that case, the type cast in question works when data exists and has been converted into that format, but when there is no data in the map, it remains a Map type and the type cast fails.

In one case, this was a database table class and I was able to see in my database viewer that the empy Map field was written to the database as {} whereas the non-empty ones were written as the list format described above. Changing these empty brackets from {} to [] (map to list) and then re-running caused the app to run without this error. In another case, adding some placeholder value in the Map on startup also avoided the error. However, these are hacky workarounds and there are many cases where you might want some Map data type that is initially empty and then filled in by various user actions, for example.

This was tested in two different projects with various types of Maps with the same result.

**To Reproduce**

1) Define a class with one or more Map fields with int keys:

Screenshot 2024-01-23 at 2 33 40 PM

2) Write an endpoint function that handles this data in some way:

Screenshot 2024-01-23 at 2 34 05 PM

3) Call this endpoint when the Map field is empty:

Screenshot 2024-01-23 at 2 34 49 PM

4) Error:

Screenshot 2024-01-23 at 2 35 02 PM

**Serverpod versions**

All serverpod components are 1.2.1

**Platform information**

Flutter 3.16.8, Dart 3.2.5, MacOS Sonoma 14.2.1, serverpod server running on localhost

### LINKED ISSUE #1908: Empty Maps still causing problems in serialization/deserialization
**IMPORTANT EDIT: After more testing, it appears this is a much more generalized problem than the example given here in this first post. You don't actually need this nested structure with Enums for it to produce the error; simply having an empty Map of type `Map<int, int>` will also throw an error when it's provided as an argument in calling an endpoint method (and possibly in other scenarios too with this empty map, but haven't exhaustively tested). See further discussion in the later posts**

**Describe the bug**

Despite merging PR #1799 and its linked closing of issue #1806 there are still problems when there is an empty Map with your own serverpod-serializable classes, at least in the form that can be recreated as described below. Maps with int keys seem to work fine now but I'm still getting the same errors as before when I have a field in one of the yaml-defined classes as follows:

`formattingData: Map<FormattingType, List<FormattingPosition>>`

where `FormattingType` in an enum and `FormattingPosition` is a serializable class (both defined in the protocol folder yaml files.

I have not done extensive testing to see what this will generalize to but probably it can at least be safely said that a map of type `Map<SomeEnum, List<SomeClass>>` will cause an error if assigned an empty value and it needs to go through the serialization/deserialization process.

**To Reproduce**

In your serverpod yaml files, define an enum and define a class (what the fields are within that class probably don't matter). Then use the data type of `Map<SomeEnum, List<SomeClass>>` as an argument in an endpoint and call that endpoint when that map is empty `{}`. It will give the error "map is not subtype of `List<dynamic>` in type cast" on the line in the generated protocol.dart file that is this or similar:
`return Map.fromEntries((data as List).map((e) => MapEntry(` where the `data` is being cast to a List.

Additionally, using an empty map like this will cause database problems. It is written to the database in an incorrect format and causes a similar issue when attempting to read it that only happens when it's empty and not when it contains some data.

If instead of an empty Map, you have all possible enum values as keys with empty list values, like:
`final workaround = {SomeEnum.valueA: [], SomeEnum.valueB: [], SomeEnum.valueC: []}`
the problems go away and it works as expected. This is an OK workaround for now, but I don't think this is or should be the expected behavior.

I currently don't have the time to do enough testing to know how much this issue generalizes, in other words if it applies to all Maps with enum keys regardless of the Map value types, or even further than that.

**Expected behavior**

An empty map should not cause problems like this. Empty lists seem fine with this various functionality and maps should as well.

**Screenshots**

**Serverpod versions**

Everything set to serverpod 1.2.3

**Platform information**

flutter web with Flutter 3.16.9

jason-downs:
fix #1908 
(also fixes #1806 which was closed but not fully fixed)

The actual edits are in `tools/serverpod_cli/lib/src/generator/types.dart` and `tools/serverpod_cli/lib/src/generator/dart/library_generators/util/class_generators_util.dart`. The rest of the changes are from changed generated code resulting from the changes in the two files mentioned. Finally, there are new comprehensive unit tests in `tests/serverpod_test_server/test/empty_map_serialization_test.dart`.

Further details:

## The Bug

**Error Message:**
```
type '_Map<String, dynamic>' is not a subtype of type 'List<dynamic>' in type cast
```

**Root Cause:**
When Maps with non-String keys (like `Map<int, String>`) contain empty data, they are supposed to be serialized as empty Lists `[]` to follow the same pattern where the non-empty map is represented as a list of objects with "k" and "v" names. However, during deserialization, if the data is somehow an empty Map `{}` instead of a List, the code tries to cast it with `(data as List)` which throws the type cast error.

## The Fix

The generated deserialization code should check if `data` is a Map before casting as List:

```dart
if (t == Map<int, int>) {
  if (data is Map) {
    return Map<int, int>() as T; // Return empty map
  }
  return Map.fromEntries((data as List).map((e) =>
      MapEntry(deserialize<int>(e['k']), deserialize<int>(e['v'])))) as T;
}
```

## Pre-launch Checklist

- [x ] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [x ] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
- [x ] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).
- [x ] I listed at least one issue that this PR fixes in the description above.
- [x ] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.
- [ x] I added new tests to check the change I am making.
- [x ] All existing and new tests are passing.
- [x ] Any breaking changes are documented below.

## Breaking changes

This shouldn't have any breaking API changes, just internal/bugfixes.

### DIFF
--- a/tests/serverpod_test_server/test/empty_map_serialization_test.dart
+++ b/tests/serverpod_test_server/test/empty_map_serialization_test.dart
@@ -1,4 +1,3 @@
-/// BUG DESCRIPTION:
 /// When Maps with non-String keys contain empty data, they are serialized as empty Lists [].
 /// However, during deserialization, if the data is somehow an empty Map {} instead of a List,
 /// the code tries to cast it with `(data as List)` which throws:
@@ -10,7 +9,7 @@
 /// 3. Deserialization code expects Lists and tries to cast with `(data as List)`
 /// 4. If data is an empty Map {} instead, the cast fails
 ///
-/// FIX REQUIRED:
+/// FIX:
 /// The generated deserialization code should check if data is a Map before casting as List,
 /// and return an empty Map in that case.
 ///


jason-downs:

tp:
> However, during deserialization, if the data is somehow an empty Map {} instead of a List, the code tries to cast it with (data as List) which throws the type cast error.

But why would this happen? That seems more of a serialization bug (if the framework would ever do it), than a deserialization one, no?

--- 

The current implementation seems a bit different than the suggest fix in the description, but the end-result is the same, so it doesn't matter for the following point:

```dart
(data is List ? data : [])
```

Basically this silently discard any "invalid" data. Personally I would not feel comfortable doing this, but rather prefer an explicit error, because the incoming data to deserializre should not be in the wrong shape in the first place. And if it is, we should be notified about it. Today this might fix some class of "empty map" bugs (however they come about, the root cause for the issue is not yet clear to me), but tomorrow this might hide something else.

So I think we need to figure out how the initial (and still current) invalid serialized data comes about, and then make the serialization strict/correct instead of the deserialization more lenient.

What do you think, @jason-downs ?

jason-downs:
> > However, during deserialization, if the data is somehow an empty Map {} instead of a List, the code tries to cast it with (data as List) which throws the type cast error.
> 
> But why would this happen? That seems more of a serialization bug (if the framework would ever do it), than a deserialization one, no?
> 
> The current implementation seems a bit different than the suggest fix in the description, but the end-result is the same, so it doesn't matter for the following point:
> 
> ```dart
> (data is List ? data : [])
> ```
> 
> Basically this silently discard any "invalid" data. Personally I would not feel comfortable doing this, but rather prefer an explicit error, because the incoming data to deserializre should not be in the wrong shape in the first place. And if it is, we should be notified about it. Today this might fix some class of "empty map" bugs (however they come about, the root cause for the issue is not yet clear to me), but tomorrow this might hide something else.
> 
> So I think we need to figure out how the initial (and still current) invalid serialized data comes about, and then make the serialization strict/correct instead of the deserialization more lenient.
> 
> What do you think, @jason-downs ?

Yes you're right. I admit to being rushed and having a lot of help here from AI in trying to navigate this codebase for the first time, which was rather daunting, and this is what it came up with for a fix. I agree that the real problem may indeed be the serialization of the empty map. 

It does happen consistently, for example when an empty `Map<int, int>` is used in calling an endpoint method - see the minimal implentation that triggers the bug down in the comments on #1908 

I can try to dive into the serialization code which might be easier for me now that I've gotten my feet wet in the codebase.

tp:
@jason-downs Thanks for the explanation and tracking this down.

The bug indeed is real, but I was only able to trigger it in an E2E test, even though the unit test exhibits the same behavior that an empty `Map<int, int>` is serialized as `[]` whereas a `String`-keyed empty map is `{}`.

I'll dig into this right away to see how we could fix it and at least give an update on the ticket. I'll ping you once I have anything to share :)

tp:
https://github.com/serverpod/serverpod/pull/3803 now fixes this issue on the sending side, instead of making the parsing more lenient.

@jason-downs I totally get your point. This was much trickier than I thought at first, as it touched quite a bit of the code generation logic (client, server, and test tools). It looks like there is some low-hanging fruit to make that whole (de)serialization a little more concise and get it to exercise the same code paths (so this would've worked on both parameters and field, or be broken on both). But for now, I wanted to get a solid fix in for your immediate issue.

As a workaround (until that fix is released), you might be best served by putting the `Map` inside a model object.

PR #3803, opened 2025-07-21
Add E2E test case for empty `Map<int, int>`

tp:
Relates to https://github.com/serverpod/serverpod/pull/3801


tp:
@coderabbitai review

tools/serverpod_cli/lib/src/generator/dart/library_generators/server_test_tools_generator.dart
```diff
@@ -181,6 +181,33 @@ class ServerTestToolsGenerator {
       'package:${config.serverPackage}/src/generated/protocol.dart',
     );
 
+    Spec handleParameter(ParameterDefinition parameterDef) {
```
tp, line 184: We should revisit this.

For some reason the server test tools (for integration tests) pass around the types without a full serialization/deserialization flow.

The record types (which do some transforms before sending) already needed to be mapped back to their proper type here, and likewise the non-`String` maps do so now too.

We should look into aligning this to use more/all of the same code the normal E2E calls would use, or avoid the mapping on either side, so that we don't have to undo it.

tests/serverpod_test_client/lib/src/protocol/client.dart
```diff
@@ -1440,15 +1440,15 @@ class EndpointMapParameters extends _i1.EndpointRef {
       caller.callServerEndpoint<Map<int, int>>(
         'mapParameters',
         'returnIntIntMap',
-        {'map': map},
+        {'map': map.isEmpty ? [] : map},
```
tp: This is the first half of the fix (for sending such maps).

Interestingly, there was only 1 endpoint class making use of such parameters (and until now no empty maps were provided as arguments).

tests/serverpod_test_server/lib/src/generated/endpoints.dart
```diff
@@ -3294,9 +3294,10 @@ class Endpoints extends _i1.EndpointDispatch {
           ) async =>
               (endpoints['mapParameters'] as _i24.MapParametersEndpoint)
                   .returnIntIntMap(
-            session,
-            params['map'],
-          ),
+                    session,
+                    params['map'],
+                  )
+                  .then((map) => map.isEmpty ? [] : map),
```
tp: Before sending out we likewise need to transform the empty map to `[]`, so that the JSON encoder does not make them `{}`.

tests/serverpod_test_server/test/serialization_legacy_test.dart
```diff
@@ -474,5 +474,43 @@ void main() {
       expect(unpacked.intIntMap[2], equals(4));
       expect(unpacked.intIntMap[3], equals(9));
     });
+
+    test('Empty Map types', () {
+      var maps = ObjectWithMaps(
```
tp: Inside models they did not cause any issues, but I kept the empty map tests for all cases just in case :)

tests/serverpod_test_server/test_integration/test_tools/serverpod_test_tools.dart
```diff
@@ -5148,23 +5148,25 @@ class _MapParametersEndpoint {
...
-        var _localReturnValue = await (_localCallContext.method.call(
-          _localUniqueSession,
-          _localCallContext.arguments,
-        ) as _i3.Future<Map<int, int>>);
+        var _localReturnValue = await _localCallContext.method
+            .call(
+              _localUniqueSession,
+              _localCallContext.arguments,
+            )
+            .then((map) => _i17.Protocol().deserialize<Map<int, int>>(map));
```
tp, line 5160: Here is an example of the code-gen change, reverting the partially mapped map back.

tests/serverpod_test_server/test/serialization_legacy_test.dart
Isakdl: Suggestion: Refactor this test away from "serialization_legacy_test" and instead in the "new" serialization test, use the same structure / style as the other test file!

tests/serverpod_test_server/test_e2e/database_roundtrip_type_test.dart
```diff
@@ -260,5 +260,23 @@ void main() {
...
+        aMap: {},
+      );
+
+      types = await client.basicDatabase.insertTypes(types);
+      expect(types.id, isNotNull);
+
+      var storedTypes = await client.basicDatabase.getTypes(types.id!);
+      expect(storedTypes, isNotNull);
+
+      expect(storedTypes!.id, equals(types.id));
```
Isakdl: Suggestion: This expect is unnecessary here, we are testing the map keys, not the id.

tests/serverpod_test_server/test_e2e/database_roundtrip_type_test.dart
```diff
@@ -260,5 +260,23 @@ void main() {
...
+        'When writing an object with an empty non-String-keyed map, then the empty map field is returned',
+        () async {
+      var types = Types(
+        aMap: {},
+      );
+
+      types = await client.basicDatabase.insertTypes(types);
+      expect(types.id, isNotNull);
+
+      var storedTypes = await client.basicDatabase.getTypes(types.id!);
```
Isakdl: Question: Do we need the insert and then get calls to validate this? The insert method will in effect test the entire roundtrip by it self right?

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -1423,6 +1439,21 @@ extension on Expression {
...
+          CodeExpression(
+            Block.of([
+              const Code('(map) => '),
+              if (returnType.generics.first.nullable)
+                const Code('map == null ? null : '),
+              const Code('map.isEmpty ? [] : map'),
+            ]),
+          ),
+        ],
+      );
```
Isakdl: Question: Would this solve a nested Map?
E.g.

```dart
Future<Map<String, Map<MyClass, String>>> method() async { ... }
```

tools/serverpod_cli/lib/src/generator/dart/library_generators/server_test_tools_generator.dart
Isakdl: Agreed! I thought we had the same code path for the test and the endpoints already, at least for the serialisation. How much effort do you think is in cleaning that up?

tools/serverpod_cli/lib/src/generator/dart/library_generators/server_test_tools_generator.dart
```diff
@@ -181,6 +181,33 @@ class ServerTestToolsGenerator {
...
+        ]);
+      }
+
+      if (parameterDef.type.isMapType &&
+          parameterDef.type.generics.first.className != 'String') {
+        return Block.of([
+          Code([
+            if (parameterDef.type.nullable) '${parameterDef.name} != null && ',
+            '${parameterDef.name}.isEmpty ? [] : ${parameterDef.name}'
+          ].join())
```
Isakdl: Comment: Same thing here does this handle nested Maps? I feel like some recursion might be in order.

Isakdl (changes requested):
Good work! I had some suggestions on a few points!

One thing I would like to be added is additional tests to validate nested Maps making sure we support that too, this is supported inside models already (and there are tests for it!), might be good to validate that the empty map case is covered too!

tools/serverpod_cli/lib/src/generator/dart/library_generators/server_test_tools_generator.dart
tp: Nesting was indeed a problem, which is now fixed by generalizing the "map containers with records" to also include a mapping for containers/records which at some point contain a non-String keyed map.

tests/serverpod_test_server/test_e2e/database_roundtrip_type_test.dart
tp: That's true, but since the entire test file is using the store and retrieve later pattern, I opted to use the same one here.

I guess this feels a little more "real", that the server really has to store the value, but since we also control the server implementation we could also do everything in 1 call everywhere (and maybe even assert that `insert(x) == findById(x.id)` on the server instead of doing that here.

tools/serverpod_cli/lib/src/generator/dart/library_generators/server_test_tools_generator.dart
tp: On the surface, it doesn't sound that hard to encode/decode inside the test tools. 

Though maybe there are some details especially around the streaming methods that make this harder (as I think you need the type for some of the helpers, which might not be available at that level anymore).

Probably the least we can do right now is write a good issue for it and use this as an example of where the current approach fell short.

tp:
@coderabbitai full review

tests/serverpod_test_server/test_integration/test_tools/serverpod_test_tools.dart
```diff
@@ -5148,23 +5148,56 @@ class _MapParametersEndpoint {
...
           _localUniqueSession,
           _localCallContext.arguments,
-        ) as _i3.Future<Map<int, int>>);
+        ) as _i3.Future<Map<String, Map<int, int>>>);
         return _localReturnValue;
       } finally {
         await _localUniqueSession.close();
       }
     });
   }
```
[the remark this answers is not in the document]
tp: You're right, that does indeed differ from the normal "Protocol" deserialization.

The thing is, that in the test tools we don't fully serialize/deserialize the return values, but rather pass the object instances around (to the calling test code).

This is of course not ideal, as those test then can not be used for find bugs in the serialization logic (because it is not executed).

We plan to change this eventually.

tests/serverpod_test_server/test_e2e/database_roundtrip_type_test.dart
tp: Changed this to only look at the map.

tests/serverpod_test_server/test/serialization_legacy_test.dart
tp: moved it to the other test file and employed the `having` check (as the return types don't get casted directly anymore).

tests/serverpod_test_server/test_e2e/database_roundtrip_type_test.dart
tp: This has been updated to only focus on the map field (though still store + get to match the overall pattern).

tests/serverpod_test_server/test_e2e/record_types_test.dart
```diff
@@ -231,6 +271,29 @@ void main() {
 
       expect(result, record);
     });
+    test(
+        'Given the test server, when a record with an empty Map<int, int> is send to the server, then it is returned empty.',
```
Isakdl: Nit: I realise the description is this on the other test, but I think the "Given the test server" is completely useless and the wrong way to organise the test description. 

To me the given should describe the precondition / scenario we are doing and when what action we are testing.


`Given an empty map when sending a request to the server then ...`

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -349,7 +350,7 @@ class LibraryGenerator {
    if (data is Iterable || data is Map) {
       return {
         'className': getClassNameForObject(data)!,
-        'data': mapRecordContainingContainerToJson(data!),
+        'data': $mapContainerToJsonFunctionName(data!),
```
Isakdl: suggestion: Extract the data variable to a local var and check that for the Iterable so we can avoid the `!`

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -1213,6 +1208,17 @@ extension TypeDefinitionReturnsRecordInContainer on TypeDefinition {
...
   }
+
+  /// Returns `true` if the type is or contains a non-String-keyed Map in a container (non-model) type
+  bool get containsNonStringKeyedMap {
+    return (isMapType && generics.first.className != 'String') ||
+        ((isMapType || isListType || isSetType) &&
+            generics.any((g) =>
+                (g.isMapType && g.generics.first.className != 'String') ||
+                // Important to only check default container types, there is not need to descent into model classes
+                (g.isMapType || g.isListType || g.isSetType) &&
```
Isakdl: Suggestion: I think it would make more sense to check if there are any generics instead as that is more robust incase we would add any other container classes. Also we still need to check nested records too right? 

We could have something like where the key is a record with a Map that has a model as a key.
```dart
(Map<(Map<MyModel, String>, String), String>)
```

^ probably a good test case to add!

Isakdl (changes requested):
I have one more concern on the nested checks, see comment.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
Isakdl: Nit: ... there is **no** need to descent ...

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
tp: Interesting, I would think the `is` check should be sufficient here to be certain that this is not `null` 🤷🏻‍♂️

tp: PS: A single `data is Map` works just fine, only in the combination does not type is as `Object` but instead retains the `Object?`…

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
tp: > I think it would make more sense to check if there are any generics instead as that is more robust incase we would add any other container classes.

That's tricky, because this is kind of tied to the target function. The helper function already accepts a few more types, e.g. any `Iterable` instead of just `List`, but we don't have that type info here.

I think I would be fine with broadening it here if the helper could be restricted, so that we get an error if we forward a type that was not intended and thus would be guaranteed to err at runtime.

> Also we still need to check nested records too right?

Top-level records will not be a problem, as they'll always get mapped (since we introduced them), but I can add a test.

But what might be worth double checking is a deeply nested record with a non-String map inside, like `Map<(Map<MyModel, String>, String), String>`.

I'll add a test case for those then :)

tests/serverpod_test_server/test_e2e/record_types_test.dart
tp: Updated the description.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
Isakdl: Wow that is very strange, sounds like a bug in the dart language?

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
tp: The case where that is (any) map inside a record in a container is already caught by `returnsRecordInContainer`.

And even if the `Map<int, int>` is nested inside a `Map<String, *>`, e.g. like `Map<String, Map<int, int>>` that works, since the `mapContainerToJson` helper works on the runtime types traversing all the way down through the container classes (and not just 1 level).

So I feel pretty good about having caught all cases now, but let me know if you can think about another variation we should test.

tp:
@coderabbitai full review

Isakdl (approved):
Nice! LGTM 🚀

PR #3807, opened 2025-07-21
fix(test): Fix flaky sorting test

tp:
As the people are only sorted by the name of their company's city, we can not make any assumptions about the order of people's own names inside each city.

Failed run observed here: https://github.com/serverpod/serverpod/actions/runs/16417694234/job/46387594091

And locally reproducible about 10% of the time:

image


tests/serverpod_test_server/test_integration/database_operations/changed_id_type_relations/one_to_one/find/order_by_test.dart
```diff
@@ -195,8 +195,13 @@ void main() async {
...
     test('then models returned are in expected order.', () {
-      var citizenNames = citizensOrderedByCompanyTownName.map((e) => e.name);
-      expect(citizenNames, ['Lina', 'Joanna', 'Alex', 'Isak', 'Theo', 'Haris']);
+      var citizenNames =
+          citizensOrderedByCompanyTownName.map((e) => e.name).toList();
+
+      expect(citizenNames, hasLength(6));
+      expect(citizenNames.sublist(0, 2), unorderedEquals(['Lina', 'Joanna']));
+      expect(citizenNames.sublist(2, 4), unorderedEquals(['Alex', 'Isak']));
+      expect(citizenNames.sublist(4, 6), unorderedEquals(['Theo', 'Haris']));
```
Isakdl: Suggestion: Wouldn't this make more sense?

Seems needlessly complex to extract sublists. 

```dart
expect(citizenNames, unorderedEquals(['Lina', 'Joanna', 'Alex', 'Isak', 'Theo', 'Haris']));
```

tp: @Isakdl No, because then that would not test any sorting, but rather any response (giving out all people/employees) would be valid, right?

The text is a little stretched out, but basically we want to assert on this result:

```dart
      citizensOrderedByCompanyTownName = await CitizenInt.db.find(
        session,
        orderBy: (t) => t.company.town.name,
      );
```

Which is set up all the way in the top of the file like this:
CleanShot 2025-07-21 at 15 55 17@2x


But I now see that I also got confused with the test file, because the above test only set up 2 cities (whereas a later tests uses 3).

So I can simplify this into 2 groups… (And I guess I was just lucky that this worked in my local test runs and on CI…)

Isakdl: Ahh alright!

PR #3811, opened 2025-07-22
feat(auth2): Add `list` method for `UserProfile`s as an admin function

tp:
## Summary by CodeRabbit

* **New Features**
  * Introduced an administrative interface for user profiles, allowing admins to list user profiles with advanced filtering by email, username, and full name, as well as pagination options.
* **Tests**
  * Added comprehensive integration tests to ensure correct behavior of user profile listing and filtering under various scenarios.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles_admin.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles_admin.dart
@@ -1,0 +1,57 @@
+part of 'user_profiles.dart';
+
+/// Admin operations complementing the end-user [UserProfilesAdmin] functionality.
+///
+/// An instance of this class is available at [UserProfiles.admin].
+final class UserProfilesAdmin {
+  UserProfilesAdmin._();
+
+  /// Returns all user profiles that match the specified criteria.
+  ///
+  /// For a direct look-up by auth user ID use [findUserProfileByUserId].
+  Future<List<UserProfileModel>> listUserProfiles(
+    final Session session, {
+    final String? email,
+    final String? userName,
+    final String? fullName,
+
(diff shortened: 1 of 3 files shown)


tp:
@coderabbitai full review

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles_admin.dart
```diff
@@ -0,0 +1,57 @@
...
+/// An instance of this class is available at [UserProfiles.admin].
+final class UserProfilesAdmin {
+  UserProfilesAdmin._();
+
+  /// Returns all user profiles that match the specified criteria.
+  ///
+  /// For a direct look-up by auth user ID use [findUserProfileByUserId].
+  Future<List<UserProfileModel>> listUserProfiles(
+    final Session session, {
+    final String? email,
```
tp, line 14: Generally, I wondered whether something like SQL `LIKE` or a closure with the `ColumnString` "builder" would be preferable.

Email look-ups will probably mostly be direct match, though a suffix/domain search does not seem unreasonable.

But since we have no indices on those tables for any operations it might be a bad idea to expose such operations here. (And the admin can always run their one-off query in the database directly as well.)

tp:
@coderabbitai full review

modules/new_serverpod_auth/serverpod_auth_profile/serverpod_auth_profile_server/lib/src/business/user_profiles_admin.dart
christerswahn: Besides `equals()` there are `like()` and `ilike()` comparators, so it's easy to do. But perhaps it is best to begin with this, as you write. That can be added later without changing the method signature if we want.

christerswahn (approved):
LGTM

Issue #2655, opened 2024-08-22
Can't upload a file to S3. Can't see the problem.


qiqetes:
### Discussed in https://github.com/serverpod/serverpod/discussions/2651

<div type='discussions-op-text'>

<sup>Originally posted by **qiqetes** August 21, 2024</sup>
```dart
// CLIENT CODE
String? uploadDescription =
    await client.X.getUploadImageDescription();

if (uploadDescription == null) return;

final image = File(state.compressedImagePath!);
print("Attempting upload: ${image.path}");
final Stream<List<int>> stream = image.openRead();

final fileUploader = FileUploader(uploadDescription);
bool uploaded =
    await fileUploader.upload(stream, image.readAsBytesSync().length);
print("Was uploaded $uploaded"); // Always FALSE, nothing on the S3
// bool uploaded = await client.X.verifyUpload();
```

I get the uploadDescription, looks ok, the image that I'm trying to upload is 8kb. I wish it wasn't a true/false but an Exception with some message what comes from `upload` method. I don't know what I'm doing wrong.

I think I have followed the docs very meticulously, still can't upload there. I have set the S3 private with a CloudFront, and this at the server side:
```dart
return await session.storage
      .createDirectFileUploadDescription(storageId: "private", path: "noslashatstart/whatever/x"); 
```

In server.dart:
```dart
 pod.addCloudStorage(s3.S3CloudStorage(
    serverpod: pod,
    storageId: 'private',
    public: false,
    region: 'us-west-1',
    bucket: 'xxx',
    publicHost: 'xxxx.cloudfront.net',
  ));
  ```
</div>

After debugging I get the error:
`<Error><Code>AccessControlListNotSupported</Code><Message>The bucket does not allow ACLs</Message>`
With statusCode 400, but if ACL is not allowed in my bucket it shouldn't try to upload with ACL.

I don't know if it's my fault or this case is not contemplated by serverpod.

LeonidVeremchuk:
<img width="748" height="446" alt="Image" src="https://github.com/user-attachments/assets/025291ed-7653-42c7-98e8-469d2a2cb371" /> 

The SDK is missing the public parameter when retrieving the file descriptor. I'm too lazy to submit a PR, but hopefully this will get fixed eventually.

PR #3814, opened 2025-07-23
fix: Add missing `public` parameter to file upload description

LeonidVeremchuk:
What this PR does
This PR fixes an issue in the createDirectFileUploadDescription method where the public parameter was not passed to AwsS3Uploader.getDirectUploadDescription. This caused incorrect headers to be generated for private S3 buckets, resulting in failed uploads.

The public parameter is now correctly forwarded, ensuring proper upload behavior for both public and private buckets.

Issues fixed by this PR
Fixes [#2655](https://github.com/serverpod/serverpod/issues/2655)

### DIFF
--- a/integrations/serverpod_cloud_storage_s3/lib/src/cloud_storage.dart/s3_cloud_storage.dart
+++ b/integrations/serverpod_cloud_storage_s3/lib/src/cloud_storage.dart/s3_cloud_storage.dart
@@ -128,6 +128,7 @@ class S3CloudStorage extends CloudStorage {
       uploadDst: path,
       expires: expirationDuration,
       maxFileSize: maxFileSize,
+      public: public,
     );
   }
 


integrations/serverpod_cloud_storage_s3/lib/src/cloud_storage.dart/s3_cloud_storage.dart
```diff
@@ -128,6 +128,7 @@ class S3CloudStorage extends CloudStorage {
       uploadDst: path,
       expires: expirationDuration,
       maxFileSize: maxFileSize,
+      public: public,
```
tp, line 131: You're right. And on `AwsS3Uploader.uploadData` we already passed the `public` parameter.

Thanks for fixing this!

tp (approved):
LGTM.

Thanks for your contribution @LeonidVeremchuk.

PR #3818, opened 2025-07-23
feat(auth2): Add `deleteEmailAccountRequestById` method on `EmailAccountsAdmin`

tp:
## Summary by CodeRabbit

* **New Features**
  * Added the ability to delete email account requests by ID in the admin interface.

* **Bug Fixes**
  * Improved clarity in method naming and documentation for finding email account requests.

* **Tests**
  * Updated test cases to use the new method name for finding email account requests.
  * Added tests to verify deletion of email account requests.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
@@ -114,8 +114,8 @@ final class EmailAccountsAdmin {
     );
   }
 
-  /// Checks whether the registration request is still pending, and if so
-  /// returns the associated email.
+  /// Checks whether an email account request is still pending, and if so
+  /// returns the associated email and verification status.
   ///
   /// In case the registration has been completed or the request is expired this
   /// returns `null`.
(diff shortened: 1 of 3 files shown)


modules/new_serverpod_auth/serverpod_auth_email_account/serverpod_auth_email_account_server/lib/src/business/email_accounts_admin.dart
```diff
@@ -144,6 +144,19 @@ final class EmailAccountsAdmin {
     );
   }
 
+  /// Deletes a registration request by its ID.
+  Future<void> deleteRegistrationRequestById(
```
tp: While "registration request" makes the concept clear, internally this was called "account request" so far (resulting in an account later on).

So I suppose we should rename `findRegistrationRequest` and `deleteRegistrationRequestById` to match that.

What do you think @christerswahn ?

christerswahn: So AccountRequest rather than RegistrationRequest? It's fine, the most important thing is to be consistent!

christerswahn: (and the xxxId name for the id to match class xxx)

christerswahn (approved):
Approved, just please apply the consistent naming!

tp:
@coderabbitai full review

PR #3822, opened 2025-07-24
feat: Migrate Serverpod templates to use Dart pub workspaces

### LINKED ISSUE #3193: Add support for dart workspaces
### Problem to Solve

Keeping dependency versions in sync between packages in a monorepo is usually a pain - Dart recently introduced Workspaces to mitigate that see https://dart.dev/tools/pub/workspaces 

### Proposal

Introduce a toplevel pubspec.yaml file to the serverpod project template.

### Use Case

pub get commands in the packages would all get the same version with a shared pubspec.lock

### Alternatives

Historically this was a pain and using `melos` was one of the best ways to keep dependencies in sync using generated pubspec_overrides.yaml files - since Dart workspaces are a first party feature this is no longer necessary (for this usecase).

### Additional context

_No response_

None

- [ ] I want to work on this

exaby73:
[Dart workspaces](https://dart.dev/tools/pub/workspaces) are a native feature in Dart recently released to manage monorepos. This PR aims to add Dart workspace support to newly generated projects

Closes #3193

## Pre-launch Checklist

## Breaking changes

While this change is not a breaking change in the traditional sense, it does change the way new projects are generated which is an important note to keep in mind


tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -303,15 +303,36 @@ class GeneratorConfig implements ModelLoadConfig {
...
-    var packageConfig = await findPackageConfig(Directory(serverRootDir));
-
-    if (packageConfig == null) {
+    // Check if we're in a workspace by looking for resolution: workspace in pubspec
+    var isWorkspace = false;
+    try {
+      var pubspecFile = File(p.join(serverRootDir, 'pubspec.yaml'));
+      var pubspecContent = pubspecFile.readAsStringSync();
+      var pubspec = Pubspec.parse(pubspecContent);
+      isWorkspace = pubspec.resolution != null;
```
christerswahn: I believe you need to check the value, other resolutions might occur in the future:
```
pubspec.resolution == 'workspace'
```

tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -303,15 +303,36 @@ class GeneratorConfig implements ModelLoadConfig {
...
+    var searchDir = isWorkspace && serverRootDir.isNotEmpty
+        ? Directory(p.dirname(serverRootDir))
+        : Directory(serverRootDir);
+
+    var packageConfig = await findPackageConfig(searchDir);
+
+    if (packageConfig == null) {
+      var errorMessage = isWorkspace
+          ? 'Failed to read your server\'s package configuration. Have you run '
+              '`dart pub get` in your workspace root directory?'
```
christerswahn, line 330: `dart pub get` can be run from any package directory of a workspace with the same result, so perhaps this hint can be rephrased?

packages/serverpod/pubspec.yaml
```diff
@@ -17,7 +17,7 @@ topics:
   - database
 
 environment:
-  sdk: '>=3.5.0 <4.0.0'
+  sdk: '>=3.6.0 <4.0.0'
```
christerswahn, line 20: Is this really what we want? See discussion in the issue. It's ok to bump minimum version of the new-project-creation templates, but maybe not yet on Serverpod.

christerswahn (changes requested):
It seems minimum Dart SDK version has been bumped throughout to 3.6. Is this really what we want? See discussion in the issue. (Admittedly 6 months old.) It's ok to bump minimum version of the new-project-creation templates, but maybe not yet on Serverpod.
Would like to get Alex' input here as well before merging.

packages/serverpod/pubspec.yaml
exaby73: So the issue is that `pubspec_parse` added the `resolution` field in `1.5.0` which requires Dart 3.6. Also, the templates have `resolution: workspace` in them and needed workspace support and failed in CI during a pub get due to CI running on 3.5. Unfortunately, the way we have templates set up, they are valid dart projects which forced me to add this change. Of course, I was going to bring this up with you guys. It's easy to revert we decide to just to a `pubspecContents.contains('resolution: workspace')` for the `pubspec_parse` case, and work around the dart pub get issue in templates

exaby73:
See https://github.com/serverpod/serverpod/issues/3193#issuecomment-3138713771 for close reason

PR #3824, opened 2025-07-24
docs(auth2): Expand on some implementation details in the docs

tp:


tp:
@coderabbitai full review

packages/serverpod/lib/src/authentication/authentication_info.dart
```diff
@@ -19,12 +19,18 @@ class AuthenticationInfo {
...
   final String userIdentifier;
 
   /// The scopes that the user can access.
   final Set<Scope> scopes;
 
   /// The authentication key id.
+  ///
+  /// For example when using the `serverpod_auth_session` module, this contains
+  /// the auth session's ID as a UUID string.
   final String? authId;
```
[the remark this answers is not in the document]
tp: `userId` is going away in a future PR, this was just for legacy support if you're using another underlying package which would indeed use `int` user IDs.

modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/src/endpoints/auth_email_base_endpoint.dart
```diff
@@ -80,9 +97,17 @@ abstract class AuthEmailBaseEndpoint extends Endpoint {
   /// {@template email_account_base_endpoint.finish_password_reset}
   /// Completes a password reset request by setting a new password.
   ///
-  /// If the reset was successful, a new session key is returned.
+  /// Throws a [EmailAccountPasswordResetRequestNotFoundException] in case no
```
christerswahn: Nit: "an" instead of "a" when the noun starts with a vowel.

tp: True, only half ot those were correct 😬

PR #3825, opened 2025-07-24
fix(auth2): Rename `attemptedAt` to `attempted` to match general pattern for `DateTimes`

tp:
like `created`, `lastUsed`, …

## Summary by CodeRabbit

* **Refactor**
  * Updated the naming of timestamp fields from `attemptedAt` to `attempted` across several authentication-related models and queries for consistency.
  * Renamed associated database indexes to match the updated field names.
  * Adjusted filtering and deletion logic to use the new field names in rate limiting and record cleanup processes.


tp:
@coderabbitai full review

tp:
Decided to instead use the `*At` suffix every, after discussing with @christerswahn.

This'll happen in https://github.com/serverpod/serverpod/pull/3827.

PR #3826, opened 2025-07-24
feat(auth2): Re-export dependencies in Apple/email client packages

tp:
Just like we do on the server.


modules/new_serverpod_auth/serverpod_auth_apple/serverpod_auth_apple_client/lib/serverpod_auth_apple_client.dart
```diff
@@ -1,2 +1,11 @@
-export 'src/protocol/protocol.dart';
+export 'package:serverpod_auth_apple_account_client/serverpod_auth_apple_account_client.dart'
```
tp, line 1: @coderabbitai Please learn to enforce that such `serverpod_auth_*` packages should re-export their `serverpod_auth_*` dependencies, so that consumers of the package only have to rely on the high-level ones and not the various implementation packages as well (in order to not get warnings about importing from a non-dependency package).

tp, line 1: Oh, I just noticed that those dependency packages are not proper dependencies of this package, so we might need to add them to `pubspec.yaml` as well.

This was ignored so far, as the dependency overrides seemingly are pulled in / installed, even if the package is not marked as needed / included by another package 🤷🏻‍♂️

tp: added the dependencies and also did so for the upcoming Google package in https://github.com/serverpod/serverpod/pull/3777/files#diff-cd2a297ca0c35525d8fd27a726f9261b999472a4c66020f794e38fbf356c5789R1-R4

christerswahn (approved):
LGTM!

PR #3827, opened 2025-07-25
feat(auth2): Consistently use `*At` suffix for model fields of type `DateTime`

tp:


modules/new_serverpod_auth/serverpod_auth_apple_account/serverpod_auth_apple_account_server/lib/src/models/apple_account.spy.yaml
```diff
@@ -20,13 +20,13 @@ fields:
   refreshTokenRequestedWithBundleIdentifier: bool
 
   ### Time when the account data was last received from Apple's servers.
-  lastRefreshed: DateTime, default=now
+  lastRefreshedAt: DateTime, default=now
```
tp, line 23: @coderabbitai Please learn this pattern and enforce it going forward.

PR #3828, opened 2025-07-25
feat(auth)!: Remove `userId` getter from `AuthenticationInfo` and instead provide extension with getter in `serverpod_auth` package

### LINKED ISSUE #3476: Remove `userId` field on `AuthenticationInfo` in `serverpod` package
With the release of "Auth 2.0" / "Serverpod 3" we want to do a breaking change to the `AuthenticationInfo` and remove the `int userId` field.

Instead, a `userId` getter for typed access of `AuthenticationInfo.userIdentifier` should be provided by each auth package (`serverpod_auth` and `serverpod_auth_user` (2.0), returning `int` and `UUID` value respectively).

For developers upgrading this would mean importing the auth package so that they get the extension (if it's not already imported).

The extension could look like this for `serverpod_auth`:

```dart
/// User ID extension for `AuthenticationInfo`
extension AuthenticationInfoUserId on AuthenticationInfo {
  /// Returns the `int` user ID of the authenticated user.
  ///
  /// Assumes that the system uses `int` user IDs, otherwise throws.
  int get userId {
    return int.parse(userIdentifier, radix: 10);
  }
}
```

tp:
Closes #3476

## Summary by CodeRabbit

* **New Features**
  * Added a new extension to provide a user ID property for authentication information.

* **Refactor**
  * Moved the user ID getter from the authentication information class to an extension, improving modularity.

* **Chores**
  * Updated test files to import the authentication server package.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/test/serverpod_auth_session_flutter_test.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_flutter/test/serverpod_auth_session_flutter_test.dart
@@ -1,5 +1,4 @@
 import 'package:flutter_test/flutter_test.dart';
-import 'package:serverpod_auth_session_client/serverpod_auth_session_client.dart';
 import 'package:serverpod_auth_session_flutter/serverpod_auth_session_flutter.dart';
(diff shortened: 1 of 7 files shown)


tp:
@coderabbitai full review

Issue #3540, opened 2025-05-13
Support Passkey authentication


tp:
After the new auth modules reach feature parity with `serverpod_auth`, we want to add Passkeys as the next authentication mechanism.

PR #3829, opened 2025-07-25
feat(auth2): Implement Passkey authentication provider

tp:
Adds the server-side implementation to add Passkeys to an existing auth user.

I opted not to allow "fresh" registration for now, as developers probably appreciate having some information about their users, as this is what all current providers offer (e.g. at least a verified email). But, of course, it would be easy to change the "register" method to create a new auth user instead of adding the key to the current user.

Relates to #3540

## Summary by CodeRabbit

- New Features
  - Introduced passkey (WebAuthn) authentication: generate challenges, register passkeys to existing accounts, and sign in using passkeys.
  - Creates authenticated sessions via passkey and honors blocked-user restrictions.
- Chores
  - Added dependency to enable passkey server capabilities.
  - Automatic cleanup of expired passkey challenges to maintain system health.
- Tests
  - Added end-to-end tests covering passkey registration and login flows.


tp:
@coderabbitai full review

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/passkey/models/passkey_account.spy.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+  ### The time when this authentication was created.
+  createdAt: DateTime, defaultModel=now
+
+  ### The ID of the [publicKey].
+  keyId: ByteData
+
+  ### Base64 variant of the key ID (to enable DB lookups).
+  keyIdBase64: String
+
```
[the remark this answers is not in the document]
tp: Great point, will switch to a key for the base64 variant.

Let's add the auth user one only once it's actually used/exposed in a query function.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/passkey/models/passkey_account.spy.yaml
```diff
@@ -0,0 +1,41 @@
...
+  ### The authenticator's attestation data.
+  attestationObject: ByteData
+
+  ### The authenticator's client data.
+  authenticatorData: ByteData
+
+  ### The challenge used during registration.
+  originalChallenge: ByteData
+
```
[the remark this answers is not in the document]
tp: Good point, I'll tackle this in a follow up.

"last used" timestamps are not currently a thing we do on authentication methods (these are immutable), but I agree that it might be useful to log usages in some way (across all providers), so we can see user behavior and spot anomalies.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/passkey/models/passkey_challenge.spy.yaml
```diff
@@ -0,0 +1,12 @@
...
+serverOnly: true
+table: serverpod_auth_idp_passkey_challenge
+fields:
+  id: UuidValue?, defaultPersist=random
+
+  ### The time when this challenge was created.
+  createdAt: DateTime, defaultModel=now
+
+  ### The actual challenge for the client
+  challenge: ByteData
```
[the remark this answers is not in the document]
tp: This is handled via the config's `challengeLifetime`, after which a challenge can not be used anymore.

Consuming the challenge is a good idea though, I'll update the code to delete used ones.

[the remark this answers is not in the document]
tp: The deletion has been added in the latest version of this PR.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/passkey/models/passkey_account.spy.yaml
```diff
@@ -0,0 +1,41 @@
...
+  keyIdBase64: String
+
+  ### The authenticator's JSON client data.
+  clientDataJSON: ByteData
+
+  ### The public key provided by the client.
+  publicKey: ByteData
+
+  ### The algorithm of the pbulic key.
+  publicKeyAlgorithm: int
```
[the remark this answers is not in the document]
tp: Typo has been fixed.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/passkey/models/passkey_account.spy.yaml
```diff
@@ -0,0 +1,41 @@
+### A fully configured passkey to be used for logins.
+class: PasskeyAccount
+serverOnly: true
+table: serverpod_auth_idp_passkey_account
+fields:
+  id: UuidValue?, defaultPersist=random
+
+  ### The [AuthUser] this profile belongs to
+  authUser: module:auth:AuthUser?, relation(onDelete=Cascade)
+
```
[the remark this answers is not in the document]
tp: It looks like there is a gap in your understanding of relationships in Serverpod.

The relationship is required, since it's not explicitly marked as `optional` here.

But the embedded object for `AuthUser` is optional, since it's not loaded by default (via a JOIN).

But each `PasskeyAccount` model instance will have at least the corresponding `authUserId` field set, even if the full `AuthUser` is not included.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/passkey/business/passkey_accounts_admin.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:clock/clock.dart';
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_idp_server/providers/passkey.dart';
+import 'package:serverpod_auth_idp_server/src/generated/protocol.dart';
+
+/// Collection of Passkey account admin methods.
+final class PasskeyAccountsAdmin {
+  /// Removes all challenges from the database which are older than the
+  /// challenge lifetime.
+  Future<void> deleteExpiredChallenged(final Session session) async {
```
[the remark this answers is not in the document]
tp: Typo is fixed.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/passkey/business/passkey_accounts.dart
```diff
@@ -0,0 +1,158 @@
...
+    final challenge = await _getChallenge(
+      session,
+      request.challengeId,
+      transaction: transaction,
+    );
+
+    await _passkeys.verifyRegistration(
+      authenticatorData: request.authenticatorData.asUint8List,
+    );
+
```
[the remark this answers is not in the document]
tp: Indeed, those parameters were not part of the pre-release version of the library we used here (mostly because it's valid that the `attestationObject` is basically empty or contains the same data as the `authenticatorData`, as e.g. Mac or Chrome do not partake in the attestation chain (unlike e.g. a YubiKey), and that are the clients we're aiming for initially).

But I have update the libraries version and now pass the newly required and validated parameters. Please re-check.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/passkey/models/passkey_registration_request.spy.yaml
```diff
@@ -0,0 +1,23 @@
...
+  challengeId: UuidValue
+
+  ### The ID of the public key.
+  keyId: ByteData
+
+  ### The contents of the public key.
+  publicKey: ByteData
+
+  ### The algorithm of the public key.
+  publicKeyAlgorithm: int
```
marcelomendoncasoares: Suggestion: Since we only support ES256 (-7) and RS256 (-257) standard algorithms ([from the `passkeys_server` package](https://github.com/LunaONE/passkeys_server/blob/ce23baceee5003f54f2306df13017e48b4ee96df/packages/passkeys_server/lib/src/passkeys.dart#L27-L30)), we could replace the `int` by an `enum` with the names of the algorithms to make it easier for users to set the parameter.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/passkey/models/passkey_challenge_expired_exception.spy.yaml
```diff
@@ -0,0 +1 @@
+exception: PasskeyChallengeExpiredException
```
marcelomendoncasoares, line 3: Suggestion: Worth adding a doc to the public exception (same valid for the other ones).

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/pubspec.yaml
```diff
@@ -16,6 +16,7 @@ dependencies:
   googleapis_auth: ">=1.6.0 <3.0.0"
   http: ">=1.1.0 <2.0.0"
   meta: ^1.11.0
+  passkeys_server: ^0.4.0
```
marcelomendoncasoares: I understand that the intention in using this package that you have developed is to facilitate the expansion of the supported passkey platforms and that you will be maintaining it with community support, right? We would only need to a first `1.0.0` to depend on a stable version.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/passkey/passkey_accounts_test.dart
```diff
@@ -0,0 +1,173 @@
...
+          ),
+        );
+
+        PasskeyAccounts.config = PasskeyAccountConfig(
+          hostname: 'localhost',
+        );
+      });
+
+      test(
+          "when calling `PasskeyAccounts.registerPasskey`, then the passkey is registered for the session's user.",
```
marcelomendoncasoares: We should also test that the user can not complete the challenge after it has expired. Then, this one could be more specific like:

```suggestion
          "when calling `PasskeyAccounts.registerPasskey` before challenge expires, then the passkey is registered for the session's user.",
```

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/passkey/passkey_accounts_test.dart
```diff
@@ -0,0 +1,173 @@
...
+          session,
+          PasskeyChallenge(
+            id: loginChallengeId,
+            challenge: _loginChallenge,
+          ),
+        );
+      });
+
+      test(
+          "when calling `PasskeyAccounts.authenticate`, then the user's ID is returned.",
```
marcelomendoncasoares: Would be good to also have a test for when the user attempts to authenticate with incorrect data on each of the `PasskeyLoginRequest` fields. This one could then be:

```suggestion
          "when calling `PasskeyAccounts.authenticate` with valid login request data, then the user's ID is returned.",
```

marcelomendoncasoares (changes requested):
Great work on this, @tp! Made only few suggestions to improve test coverage and minor details of user experience. To complete the feature, we will only need a first stable release of the `passkeys_server` package.

tp:
@marcelomendoncasoares just a quick update from my end: Windows support just landed (had to get a Surface machine to be able to continually test it 😉). I am currently building a Flutter web frontend using the popular client package “passkeys”. We’ll see whether that works out of the box like my “plain HTML” demo or if the APIs need to be adapted (which would then also need to be mirrored in the models here). Once the Flutter web example is live I can do the 1.0.0 release of the package and then will update the dependency here and work through your initial comments :)

I would say this should not be more than 2-3 days of work remaining (of what I see so far), but I may be busy next week and thus only finish this up the week after. Talk soon!

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/passkey/models/passkey_registration_request.spy.yaml
tp: @marcelomendoncasoares How strongly do you feel about this one? 

I would rather treat the algorithms as "opaque integers" as we just shuffle them back and forth the server- and client-side libraries.

I don't any value in having them better "understood" by the Serverpod package.

There's definitely 2 more (but strictly optional) ones on the horizon.

As the package already exposes the "support algorithms" (to build the client request objects), I would ideally just forward this to the client, and not introduce an intermediate mapping layer.

PS: Maybe I'll just drop this from the underlying package entirely anyway, as it's already bundled up in the `attestationObject.authenticatorData`, and there is no need (apart from looking at it out of interest) to have it available by itself, without the related key (which is also in that data). In the end it's probably best to do the least amount of transformations/extractions of the underlying data, to not introduce any errors (as we really don't need to look at them in isolation anyway).

tp: This has been removed as an explicit field.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_client/lib/src/protocol/providers/email/models/exceptions/email_account_login_failure_reason.dart
```diff
@@ -34,7 +34,6 @@ enum EmailAccountLoginFailureReason implements _i1.SerializableModel {
 
   @override
   int toJson() => index;
-
```
tp, line 37: Weird, this now generates like this for me after pulling in `main` and upgrading to Dart 3.9.2. 

Does this ring a bell @marcelomendoncasoares ?

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/passkey/models/passkey_challenge_expired_exception.spy.yaml
tp: Added comments in bea43e1156aeb8fad42020b10e9c8bfd83b1f5f2

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/passkey/passkey_accounts_test.dart
tp: added the suggested additional cases in 05556009c6fa1344490ee1f8804ce9f11dc307d4

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/pubspec.yaml
tp: Sounds good! I think the 0.6.0 API is now stable and will become 1.0.0 soon.

I just have a short list of internal polishes to do, before I will push that one out :)

marcelomendoncasoares: Perfect! Just ping me once you update here to 1.0.0 and we are good to go!

tp: @marcelomendoncasoares I have updated to version 1.0.0 and updated the existing v3 migrations, so that there is only one in each affected module.

tp:
Seems like 1 unrelated test case failed, but I can not re-run the CI jobs anymore to ensure that this truly is a fluke.

PR #3834, opened 2025-07-31
feat(auth2): Use url-safe base64 encoding for session key

tp:
I attempted to simplify the format further by encoding the entire dataset in one URL-safe base64 encoded String, but since the Serverpod `Client` takes the raw "session key" and sends it as a `Basic` authorization header, that failed (rightfully) in the Relic webserver, as the string did not contain a `:` anymore and thus did not match the `user:pass` schema.

Thus we have to stick with this format for now, unless we change the client/server (while keeping backwards compatibility). For this token as well as JWT it might be prudent to support the `Bearer` token pattern, and not further modify the values.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal handling of session keys for authentication sessions, delegating parsing and building logic to dedicated utilities.

* **Tests**
  * Added tests to ensure session keys are URL-safe, correctly formatted for authentication headers, and can be reliably encoded and decoded.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/session_key.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/session_key.dart
@@ -1,0 +1,79 @@
+import 'dart:convert';
+import 'dart:typed_data';
+
+import 'package:meta/meta.dart';
+import 'package:serverpod/serverpod.dart';
+
+/// Prefix for sessions keys
+/// "sas" being abbreviated of "serverpod_auth_session"
+const _sessionKeyPrefix = 'sas';
+
+/// Creates the String representation to map to an `AuthSession`.
+@internal
+String buildSessionKey({
+  required final UuidValue authSessionId,
+  required final Uint8List secret,
+}) {
+  // Since the session key is sent as a `Basic` HTTP `Authorization` header by the Serverpod client,
+  // it's important that it contains at least one `:` so the `user:pass` schema is obeyed.
+  return '$_sessionKeyPrefix:${base64Url.encode(authSessionId.toBytes())}:${base64Url.encode(secret)}';
+}
+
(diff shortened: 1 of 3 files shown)


tp:
@coderabbitai full review

modules/new_serverpod_auth/serverpod_auth_session/serverpod_auth_session_server/lib/src/business/session_key.dart
```diff
@@ -0,0 +1,79 @@
...
+const _sessionKeyPrefix = 'sas';
+
+/// Creates the String representation to map to an `AuthSession`.
+@internal
+String buildSessionKey({
+  required final UuidValue authSessionId,
+  required final Uint8List secret,
+}) {
+  // Since the session key is sent as a `Basic` HTTP `Authorization` header by the Serverpod client,
+  // it's important that it contains at least one `:` to the `user:pass` schema is obeyed.
```
tp:
```suggestion
  // it's important that it contains at least one `:` so the `user:pass` schema is obeyed.
```

SandPod:
@tp - Since the token no longer strictly adheres to the `user:password` format of the Basic header, perhaps it is better if we simply transmit our token as a bearer token instead? We now have 3 section more simialir to the JWT format.

After some reading it seems fine to have a custom format for the token: https://datatracker.ietf.org/doc/html/rfc6750

tp:
@SandPod Yes, I would love to switch to that format.

We just have to be careful with backwards compatibility on the server. I think currently we can not do an E2E test from one version to another, but I would love to have this checked with a "real" 2.9.x client against the server. Otherwise blind-coding this and going to a different code path for `Bearer …` auth headers would be easy (and at least the new approach easily tested).

So maybe this is a good first step, and then when the `Bearer` token support is added we can switch to the other commit (single base64 string) from this PR and use that instead. (It's already suggested as such on the JWT session manager implementation ticket.)

SandPod:
Sounds good to me! Then we just need to make sure that we don't release this before we switched over to the other header.

PR #3835, opened 2025-07-31
feat(auth2): Provide abstract email endpoint for both database and JWT backed sessions

tp:
@SandPod for your consideration, I have built out the class hierarchy as discussed yesterday to showcase how we could have a "session-type agnostic" email endpoint, and implemented the one for `serverpod_auth_session`.

Similarly, we would have to create JWT variants for each of the touched classes.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/serverpod_auth_email_server.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_email/serverpod_auth_email_server/lib/serverpod_auth_email_server.dart
@@ -8,6 +8,7 @@ export 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart'
     hide Endpoints, Protocol;
 
 export 'src/business/auth_email.dart' show AuthEmail;
-export 'src/endpoints/auth_email_base_endpoint.dart' show AuthEmailBaseEndpoint;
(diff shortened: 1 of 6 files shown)


tp:
Personally, I don't think this is a good way forward, as the combinatory explosion in the class names shows this to not scale, and I would be surprised if this is warranted for just the session type switch but not other variants one could think of.

Generally, I would probably also not push JWT for _internal_ client to server communication as much, as the single DB look up for an incoming request to me seems a good deal compared to the downsides of the access tokens. So I would rather envision clients that have to communicate with 3rd party services (that want to rely on/trust the Serverpod server) to only use access tokens when communicating with them. In that regard, we might want to ensure (as it currently does), that our JWT session manager would also be useful to prepare such "external" requests and does not in any way get locked down to be only usable with/from a Serverpod `Client`.

SandPod:
Does this work with our generator?

tp:
> Does this work with our generator?

No, not yet. The fixed generics the subclass sets do not get used on the endpoint methods:

CleanShot 2025-08-01 at 10 33 14@2x
(The "SessionType" in this case does not even exist and is the name I gave to the generic type parameter.)

So the code generator would need to get updated to support this pattern.

SandPod:
We implemented client generation of abstract endpoints and true inheritance in endpoints. Meaning we can create client side logic that does not depend on the real endpoint implementation.

PR #3839, opened 2025-07-31
feat(auth2): Use `UuidValue.withValidation` instead of `fromString`

tp:
Ensuring that the string indeed represented a valid UUID

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
@@ -79,7 +79,7 @@ class JwtUtil {
 
     final UuidValue refreshTokenId;
     try {
-      refreshTokenId = UuidValue.fromString(jwt.jwtId!);
+      refreshTokenId = UuidValue.withValidation(jwt.jwtId!);
     } catch (e) {
       throw ArgumentError(
         'The refresh token ID could not be read from the JWT\'s `id` claim: "${jwt.jwtId}"',
@@ -89,7 +89,7 @@ class JwtUtil {
 
     final UuidValue authUserId;
     try {
-      authUserId = UuidValue.fromString(jwt.subject!);
+      authUserId = UuidValue.withValidation(jwt.subject!);
     } catch (e) {
       throw ArgumentError(
         'The auth user ID could not be read from the JWT\'s `subject` claim: "${jwt.subject}"',


modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
```diff
@@ -79,7 +79,7 @@ class JwtUtil {
 
     final UuidValue refreshTokenId;
     try {
-      refreshTokenId = UuidValue.fromString(jwt.jwtId!);
+      refreshTokenId = UuidValue.withValidation(jwt.jwtId!);
```
tp, line 82: The only other (non test) place we use `UuidValue.fromString` is for any default model values.

Should we update the code generator here, check the value once during code gen, or leave it as is (in which case the DB write might fail)? 

https://github.com/serverpod/serverpod/blob/b7fd2625281e112dfbbbb2889270cbe8e9c7a97b/tests/serverpod_test_client/lib/src/protocol/defaults/uuid/uuid_default_mix.dart#L22

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
```diff
@@ -89,7 +89,7 @@ class JwtUtil {
 
     final UuidValue authUserId;
     try {
-      authUserId = UuidValue.fromString(jwt.subject!);
+      authUserId = UuidValue.withValidation(jwt.subject!);
```
tp, line 92: @coderabbitai Ensure that we're using this pattern going forward.

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
SandPod: We discussed this when implementing UUIDs in general and we already run the validation when defining the values so it should not be required: https://github.com/serverpod/serverpod/blob/b7fd2625281e112dfbbbb2889270cbe8e9c7a97b/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions/default.dart#L299

SandPod (approved):
LGTM! 👍

modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/jwt_util.dart
tp: Ah, perfect! :)

PR #3844, opened 2025-07-31
feat(auth2): Require callback parameter on `AppleAccountsAdmin.checkAccountStatus` instead of having it configured via the config

tp:
## Summary by CodeRabbit

* **New Features**
  * The batch size limit for database queries when checking Apple account status is now configurable.

* **Bug Fixes**
  * Improved handling of revoked Apple authentication tokens by requiring an explicit callback for expired user authentication.

* **Refactor**
  * Streamlined internal configuration for Apple account administration, removing unnecessary dependencies.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_apple_account/serverpod_auth_apple_account_server/lib/src/business/apple_accounts_admin.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_apple_account/serverpod_auth_apple_account_server/lib/src/business/apple_accounts_admin.dart
@@ -6,12 +6,11 @@ import 'package:sign_in_with_apple_server/sign_in_with_apple_server.dart';
 
 /// Administrative Apple account management functions.
 final class AppleAccountsAdmin {
-  final AppleAccountConfig _config;
   final SignInWithApple _siwa;
 
   /// Creates a new instance of the admin utilities.
   @internal
-  AppleAccountsAdmin(this._config, this._siwa);
+  AppleAccountsAdmin(this._siwa);
 
   /// Checks whether all accounts are in good standing with Apple and that the
   /// authorization has not been revoked.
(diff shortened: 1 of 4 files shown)


tp:
@coderabbitai full review

modules/new_serverpod_auth/serverpod_auth_apple_account/serverpod_auth_apple_account_server/lib/src/business/apple_accounts_admin.dart
```diff
@@ -24,24 +23,22 @@ final class AppleAccountsAdmin {
   /// revoked.
   Future<void> checkAccountStatus(
     final Session session, {
+    /// Callback to be invoked when an Apple authentication has been revoked.
+    ///
+    /// In this case all sessions associated with this sign-in method should be
+    /// removed.
+    required final void Function(UuidValue authUserId)
```
SandPod, line 30: Question: Do we not have any tests for this code? I would expect this to trigger some form of test update since the interface is modified.

tp: No, not yet. Since the function primarily calls out to Apple's web service, and that one is currently not mockable (but we could of course make it replacable with a mock).

Same thing goes for Google, we should probably find a level at which to intercept the external calls, so we can test the behavior around it.

tp: Ah, since the admin class already takes a Apple interface instance instead of creating it itself I should be able to add that one easily and cleanly (as opposed to the non-admin variant, where it gets internally created based on the config and thus will be trickier).

tp: added an admin integration test for this.

PR #3849, opened 2025-08-02
feat!: Require `String` for `userIdentifier` in `AuthenticationInfo` constructor

### LINKED ISSUE #3477: Change `Object userIdentifier` to use type `String` instead
In order to not introduce breaking changes in #3443 we widened the previous `int user` parameter to `Object userIdentifier` (only expected `int` or `UuidValue` for now from our own packages).

In a future major release we should clean this up to be of type `String` to match `AuthenticationInfo.userIdentifier` (which is what should be passed here anyway).

Currently areas in the code using this:
https://github.com/tp/serverpod/blob/4315e0f240e0fa2ecef1a8d20e230dad5a7e9d2d/packages/serverpod/lib/src/server/session.dart#L645
https://github.com/tp/serverpod/blob/4315e0f240e0fa2ecef1a8d20e230dad5a7e9d2d/packages/serverpod/lib/src/server/message_central.dart#L12
https://github.com/tp/serverpod/blob/4315e0f240e0fa2ecef1a8d20e230dad5a7e9d2d/packages/serverpod/lib/src/authentication/authentication_info.dart#L32

tp:
Instead of internally converting `Object` to `String`

Also changes the related usages of `Object userIdentifier`, which mirror/are triggered by the `AuthenticationInfo` field.

Closes #3477

## Summary by CodeRabbit

* **Refactor**
  * Standardized user identifier types across the authentication system to consistently use string values instead of integers or objects.

* **Tests**
  * Updated all related test cases to use string-based user identifiers, ensuring compatibility with the new identifier format.

* **Bug Fixes**
  * Improved reliability and consistency in authentication and session management by unifying the user identifier type.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_info_from_jwt.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_jwt/serverpod_auth_jwt_server/lib/src/business/authentication_info_from_jwt.dart
@@ -13,7 +13,7 @@ extension AuthenticationInfoFromJwt on AuthenticationInfo {
     final VerifiedJwtData result,
   ) {
     final authInfo = AuthenticationInfo(
-      result.authUserId,
+      result.authUserId.uuid,
       result.scopes,
       authId: result.refreshTokenId.toString(),
     );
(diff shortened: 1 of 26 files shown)


tp:
@coderabbitai full review

SandPod (approved):
LGTM!

PR #3851, opened 2025-08-04
refactor(auth2)!: Consolidate auth provider packages

SandPod:
Consolidates all new auth provider packages into a single package called `serverpod_auth_idp`.

Models and passwords have been updated to reflect the package renaming.

Notably, the last commit renames the `serverpod_auth_email_account_passwordHashPepper` to `serverpod_auth_idp_email_passwordHashPepper`, which would be a required change for anyone depending on this work currently.

This PR will be followed by a second PR that consolidates all the core package into a single `serverpod_auth_core` package.

### Review guidance

Unfortunately, this PR is a bit of a "trust me, I'm an engineer" PR. It will be hard to review but could be worth inspecting the new state.

## Pre-launch Checklist

## Breaking changes

Yes, but this is connected to an unpublished feature.


modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/email/email.dart
```diff
@@ -0,0 +1,13 @@
+export 'package:serverpod_auth_session_server/serverpod_auth_session_server.dart'
+    show AuthSuccess;
+
+export 'business/auth_email.dart' show AuthEmail;
+export 'business/email_account_config.dart';
+export 'business/email_account_not_found_exception.dart';
```
christerswahn: Double-checking regarding the exceptions of the email IDP:

- This exception is exported but defined in Dart source code. Is it exported for server-side-only use?
- All the other email-related exceptions, are they automatically exported via the generated code?

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/serverpod_auth_idp_server.dart
```diff
@@ -1,13 +1,12 @@
...
 export 'package:serverpod_auth_profile_server/serverpod_auth_profile_server.dart'
     hide Endpoints, Protocol;
 export 'package:serverpod_auth_session_server/serverpod_auth_session_server.dart'
     hide Endpoints, Protocol;
 export 'package:serverpod_auth_user_server/serverpod_auth_user_server.dart'
     hide Endpoints, Protocol;
 
-export 'src/business/auth_email.dart' show AuthEmail;
-export 'src/endpoints/auth_email_base_endpoint.dart' show AuthEmailBaseEndpoint;
+export 'src/apple/apple.dart';
```
christerswahn: Discussion: Should we also support importing these libraries individually? E.g. email but not the other IDP variants?

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/email/email.dart
SandPod: Yes, the configuration here is deliberate.

The intention is that if we throw this exception, this is a server error and should not be leaked to the client.

And all the other exceptions are exported through the generated code (through the protocol file).

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/serverpod_auth_idp_server.dart
SandPod: We could, but I would like to introduce a single `provider` configuration class that you use to configure the providers you would like to have enabled.

That could look something like this: `AuthConfig.set(google: GoogleConfig(...), apple: AppleConfig(....), email: EmailConfig(...))`

That would allow us to remove the passwords that we automatically read and allow the user to define them as they wish.

What do you think about that?


Inspiration is taken from here [NextAuth](https://next-auth.js.org):

```
import NextAuth from 'next-auth'
import AppleProvider from 'next-auth/providers/apple'
import FacebookProvider from 'next-auth/providers/facebook'
import GoogleProvider from 'next-auth/providers/google'
import EmailProvider from 'next-auth/providers/email'

export default NextAuth({
  providers: [
    // OAuth authentication providers...
    AppleProvider({
      clientId: process.env.APPLE_ID,
      clientSecret: process.env.APPLE_SECRET
    }),
    FacebookProvider({
      clientId: process.env.FACEBOOK_ID,
      clientSecret: process.env.FACEBOOK_SECRET
    }),
    GoogleProvider({
      clientId: process.env.GOOGLE_ID,
      clientSecret: process.env.GOOGLE_SECRET
    }),
    // Passwordless / email sign in
    EmailProvider({
      server: process.env.MAIL_SERVER,
      from: 'NextAuth.js <no-reply@example.com>'
    }),
  ]
})
```

christerswahn: Couldn't we do both? (Note the imports in the pasted example.) If the `providers` configuration element takes a general type, and each providers configuration definition is of that type, wouldn't something like this be possible:

```
AuthConfig.set(providers: [
  GoogleConfig(...),
  AppleConfig(....),
  EmailConfig(...)),
]);
```

And if users are able to add custom providers, that is an additional reason of not hardcoding member names like `apple` right?

SandPod: You're right, the division of packages makes and I didn't think about that. I'll implement the change. (This shared provider configuration does not exist yet).

With Dart's type safety at hand, I used hardcoded parameters to increase discoverability in a package that is not written by the user. This also has the benefit of preventing multiple configurations of the same providers.

I would expect some shared configuration between providers to exist in the future so it also makes sense to allow the user to configure their custom provider through our interface.

I see two options here.
1. Follow your suggestion and use a generalized `ProviderConfiguration` interface that the different providers implement (KISS principle).
```dart
AuthConfig.set(providers: [
  GoogleConfig(...),
  AppleConfig(....),
  EmailConfig(...)),
]);
```

2. Use hardcoded configuration values, but also provide a `customProviders` parameter that takes a list of the generalized `ProviderConfiguration` interfaces to make it easy to tag on custom implementations. (Opting for discoverability and guidance)
```dart
AuthConfig.set(providers: Providers(
  google: GoogleConfig(...),
  apple: AppleConfig(....),
  email: EmailConfig(...),
  customProviders: [....],
))
```

What do you think?

nielsenko: I prefer just a `providers` argument. I think discoverability is fine, there is a mandatory named argument, that is a list/iterable of provider configs. Also, I would loose the `.set` and just use an unnamed ctor.

SandPod: @coderabbitai Could you create an issue for adding this shared configuration interface using the generalized configuration list?

christerswahn (approved):
LGTM! 👍

Issue #3217, opened 2025-02-10
Custom serializable classes utilized as optional members in methods generates code with null-aware operator linting warnings


dannejosefsson:
### Describe the bug

When a custom serializable class that implements `ProtocolSerialization` is used as a optional member of a model or protocol model the generated code that is spawned generates a linting warning:
`The receiver can't be null, so the null-aware operator '?.' is unnecessary.
Try replacing the operator '?.' with '.'.` in the corresponding model/protocol file. 

```dart
  @override
  Map<String, dynamic> toJsonForProtocol() {
    return {
      if (className != null)
        'className':
// ignore: unnecessary_type_check
            className is _i1.ProtocolSerialization
                ? (className as _i1.ProtocolSerialization)?.toJsonForProtocol()
                : className?.toJson()
    };
```

Given that the check that the object implements `ProtocolSerialization`, it seems that the null aware operator isn't needed.

This behavior arose somewhere between 2.1.5 and 2.3.1 and is probably related to #3015.

### To reproduce

1. Create a custom serializable class, like:
```dart
class ClassName implements ProtocolSerialization {
  @override
  toJsonForProtocol() {
    return "customClass";
  }
  //...
}
```
2. Configure the `*_server` according to [Custom serialization](https://docs.serverpod.dev/concepts/serialization)
3. Create a model or protocol model with this class as an optional member, like
```yaml
class: CustomClassNullAware
fields:
    className: ClassName?
```
4. Generate the code via `serverpod generate`.
5. Check linting warnings

### Expected behavior

The code generated should not trigger a linting warning.

### Library version

$ dart --version
Dart SDK version: 3.6.1 (stable) (Tue Jan 7 09:50:00 2025 -0800) on "linux_x64"
$ serverpod version
Serverpod version: 2.3.1

### Platform information

Docker container based on Ubuntu 24.10.

### How experienced are you with this library?

Intermediate - Familiar with the basics or have used it in a few projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

SandPod:
Thanks for the report @dannejosefsson!

You are correct in pinpointing the introduction of this issue.

The comment above the line with the issue `// ignore: unnecessary_type_check` was probably intended to silence it. But it doesn't seem like it is very effective.

daniel-b3:
[...]

PR #3854, opened 2025-08-05
fix: Unnessary optional null check on generated class with custom class

exaby73:
When using custom classes in protocols, an uncessary null-aware operator (`?.`) was added to the generated code in `toJsonForProtocol()`. This PR aims to fix that lint

Fixes https://github.com/serverpod/serverpod/issues/3217

## Pre-launch Checklist

### DIFF
--- a/tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
@@ -1222,14 +1222,9 @@ class SerializableModelLibraryGenerator {
         serverpodUrl(serverCode),
       );
 
-      var toJsonForProtocolExpression = switch (nullableField) {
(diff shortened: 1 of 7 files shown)


tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -1222,14 +1222,9 @@ class SerializableModelLibraryGenerator {
         serverpodUrl(serverCode),
       );
 
-      var toJsonForProtocolExpression = switch (nullableField) {
```
SandPod, line 1225: Question: Did this not prompt any changes in our checked in generated files?

If not, could you add a model that would trigger this issue?

exaby73: Added

PR #3857, opened 2025-08-06
refactor!(auth2): Rename the backwards compatibility module.

SandPod:
Renames the backwards compatability module into `serverpod_auth_bridge` to better reflect that the package should be used as a bridge after migration and should later on be removed.

## Pre-launch Checklist

## Breaking changes

Yes, but only in packages that have not been published.


modules/new_serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/serverpod_auth_bridge_server.dart
```diff
@@ -1,3 +1,10 @@
+/// This library provides backwards compatibility for the new authentication
```
christerswahn, line 1: Suggestion: Add this comment on the client's export file as well (and flutter), in case the user may see those imports directly?

christerswahn (approved):
LGTM, just one suggestion for consideration

modules/new_serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/serverpod_auth_bridge_server.dart
SandPod: I don't think it is needed on the client since the use of the package is automatically generated for the user. But for the flutter package it makes sense. Will add.

Issue #3420, opened 2025-04-14
Support partial column updates via `updateById` and `updateWhere` methods


Isakdl:
## Problem to Solve  
Currently, updating database rows requires providing the entire instantiated object, even if only one or two fields need to be modified. This can be limiting or cumbersome, especially in use cases like data migrations, where only specific columns of certain records need to be updated without constructing full objects.

## Proposal  
Introduce two new methods for database operations: `updateById` and `updateWhere`.

These methods would:
- Accept a `columns` callback similar to the standard `update` method.
- The `columns` callback must return a list of special `ColumnValue` objects created from the input `Table t`, using a syntax like `t.column(value)`.
- The value must always be supplied, even if `null`, to make it explicit that the column is being written to.

### Example usage:
```dart
// Update a single row by ID
await Example.db.updateById(session, exampleId, (t) => [t.name('John Doe')]);

// Update rows matching a condition
await Example.db.updateWhere(
  session,
  (t) => [t.name('John Doe')],
  where: (t) => t.address.equals('Street'),
);
```

These new methods should also support the standard parameters used in other `db` methods, such as:
- `transaction`
- `orderBy`
- `limit`
- etc.

## Use Case  
Useful in data migration scripts or administrative tasks where only a subset of fields need to be updated, and full model instantiation is unnecessary or infeasible. It enhances performance and code simplicity for bulk updates or partial mutations.

## Alternatives  
- Continue requiring full model instantiation, which is verbose and less ergonomic for partial updates.
- Use raw SQL queries, which breaks type safety and consistency with Serverpod's API.

## Additional Context  
This addition will improve developer ergonomics and expand Serverpod's capabilities for more advanced backend tasks like migrations, admin tooling, and batch operations.

### Proposal

-

### Use Case

-

### Alternatives

-

### How experienced are you with this library?

None

### Are you interested in working on a PR for this?

- [ ] I want to work on this

dkbast:
Additional context:
When working with "soft deletion" (marking rows as deleted) this would come in very handy. 
E.g. you delete a user and want to mark everything the user owns as deleted, or you want to remove the users id from references.

PR #3861, opened 2025-08-07
feat: Add support for partial updates

exaby73:
This PR adds new methods namely 'updateWhere' and `updateById` to the repository classes and database to support partial updates

Closes https://github.com/serverpod/serverpod/issues/3420

## Pre-launch Checklist


exaby73:
@coderabbitai full review

exaby73:
@coderabbitai review

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -234,6 +235,84 @@ class DatabaseConnection {
...
+    var table = _getTableOrAssert<T>(session, operation: 'updateById');
+
+    if (columnValues.isEmpty) {
+      throw ArgumentError('columnValues cannot be empty');
+    }
+
+    var setClause = columnValues.map((cv) {
+      var value = DatabasePoolManager.encoder.convert(cv.value);
+      return '"${cv.column.columnName}" = $value::${_convertToPostgresType(cv.column)}';
+    }).join(', ');
```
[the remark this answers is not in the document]
exaby73: The column names come from generated code so this is a non issue

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -732,12 +811,12 @@
     if (column is ColumnDuration) return 'bigint';
     if (column is ColumnUuid) return 'uuid';
     if (column is ColumnUri) return 'text';
-    if (column is ColumnBigInt) return 'text';
-    if (column is ColumnVector) return 'vector';
-    if (column is ColumnHalfVector) return 'halfvec';
-    if (column is ColumnSparseVector) return 'sparsevec';
-    if (column is ColumnBit) return 'bit';
-    if (column is ColumnSerializable) return 'json';
+    if (column is ColumnBigInt) return 'jsonb';
```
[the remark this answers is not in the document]
exaby73: In Dart, `BigInt` can be an arbitrarily long int which Postgres's `bigint` does not support since it's limited to 64 bits as far as I know. So we intentionally save it has json to preserve compatibility with arbitrary long numbers

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -732,12 +811,12 @@ class DatabaseConnection {
...
```
SandPod: I'll hold off from a review until this has been reverted.

Please have a thorough look at all the changes in the first commit to ensure that you stand by them.

exaby73: Reverted

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -234,6 +235,84 @@ class DatabaseConnection {
...
+  /// Updates a single row by its ID with the specified column values.
+  ///
+  /// Returns the updated row or null if no row with the given ID exists.
+  /// Throws [ArgumentError] if [columnValues] is empty.
+  ///
+  /// For most cases use the corresponding method in [Database] instead.
+  Future<T?> updateById<T extends TableRow>(
+    Session session,
+    Object id,
+    List<ColumnValue> columnValues, {
```
marcelomendoncasoares: May be worth to make this `columnValues` a named parameter to make it follow the same standard of other methods, such as `findById` (only `session` and `id` fields as required positional parameters).

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -234,6 +235,84 @@ class DatabaseConnection {
...
+
+  /// Updates all rows matching the WHERE expression with the specified column values.
+  ///
+  /// Returns a list of all updated rows. Returns an empty list if no rows match.
+  /// Throws [ArgumentError] if [columnValues] is empty.
+  ///
+  /// For most cases use the corresponding method in [Database] instead.
+  Future<List<T>> updateWhere<T extends TableRow>(
+    Session session,
+    List<ColumnValue> columnValues, {
```
marcelomendoncasoares: Same as the previous, may be worth making this `columnValues` named to follow the standard of other methods, such as `findFirstRow` (only `session` as required positional parameter).

packages/serverpod/lib/src/database/database.dart
```diff
@@ -169,6 +170,40 @@ class Database {
     );
   }
 
+  /// Updates a single [TableRow] by its [id] with the specified column values.
+  /// Returns the updated row or null if no row with the given id exists.
+  @internal
+  Future<T?> updateById<T extends TableRow>(
+    Object id,
+    List<ColumnValue> columnValues, {
```
marcelomendoncasoares: Same as above, to reflect the `DatabaseConnection` scheme.

packages/serverpod/lib/src/database/database.dart
```diff
@@ -169,6 +170,40 @@ class Database {
...
+      // ignore: invalid_use_of_visible_for_testing_member
+      transaction: transaction ?? _session.transaction,
+    );
+  }
+
+  /// Updates all [TableRow]s matching the [where] expression with the specified column values.
+  /// Returns the list of updated rows.
+  @internal
+  Future<List<T>> updateWhere<T extends TableRow>(
+    List<ColumnValue> columnValues, {
```
marcelomendoncasoares: Same as above, to reflect the `DatabaseConnection` scheme.

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
```diff
@@ -701,6 +703,127 @@ class BuildRepositoryClass {
...
+            ..type = refer('Session', 'package:serverpod/serverpod.dart')
+            ..name = 'session'),
+          Parameter((p) => p
+            ..type = idTypeReference.rebuild((u) => u.isNullable = false)
+            ..name = 'id'),
+          Parameter((p) => p
+            ..type = TypeReference((b) => b
+              ..symbol = 'ColumnValueListBuilder<${className}Table>'
+              ..url = 'package:serverpod/serverpod.dart')
+            ..name = 'columns'),
```
marcelomendoncasoares: To keep the same standard of the `Database` class, may be valid renaming this to `columnValues` and making it a named parameter.

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
```diff
@@ -701,6 +703,127 @@ class BuildRepositoryClass {
...
+            ..name = 'transaction'
+            ..named = true),
+        ])
+        ..modifier = MethodModifier.async
+        ..body = refer('session')
+            .property('db')
+            .property('updateById')
+            .call([
+              refer('id'),
+              refer('columns').call([
```
marcelomendoncasoares: This call will have to be moved to the named section below to reflect the named parameters change in the `Database` class - right before the `transaction` parameter below.

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
```diff
@@ -701,6 +703,127 @@ class BuildRepositoryClass {
...
+        )
+        ..requiredParameters.addAll([
+          Parameter((p) => p
+            ..type = refer('Session', 'package:serverpod/serverpod.dart')
+            ..name = 'session'),
+          Parameter((p) => p
+            ..type = TypeReference((b) => b
+              ..symbol = 'ColumnValueListBuilder<${className}Table>'
+              ..url = 'package:serverpod/serverpod.dart')
+            ..name = 'columns'),
```
marcelomendoncasoares: Same as above, to keep the same standard of the `Database` class, may be valid renaming this to `columnValues` and making it a named parameter.

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,205 @@
...
+      var updated = await UniqueData.db.updateById(
+        session,
+        existingEntry.id!,
+        (t) => [t.number(100), t.email('updated@serverpod.dev')],
+      );
+
+      expect(updated, isNotNull);
+      expect(updated!.number, 100);
+      expect(updated.email, 'updated@serverpod.dev');
+    });
```
marcelomendoncasoares: This test seems a bit redundant with the one on line 125 - also tests multiple columns, but checking that other columns remain unchanged. May be worth keeping only the most complete of the tests.

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,205 @@
...
+        (t) => [t.anInt(42), t.aBool(true), t.aString('value')],
+      );
+
+      expect(updated, isNotNull);
+      expect(updated!.anInt, 42);
+      expect(updated.aBool, true);
+      expect(updated.aString, 'value');
+      expect(updated.aDouble, isNull);
+      expect(updated.aDateTime, isNull);
+    });
```
marcelomendoncasoares: This test seems mostly duplicated with the previous one of line 125. The only difference is the `aBool` value being set, but not actually testing anything new.

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,205 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+import 'package:serverpod_test_server/test_util/test_serverpod.dart';
+import 'package:test/test.dart';
+
+void main() async {
```
marcelomendoncasoares: As a general comment on the tests, it may be interesting to use only the `Types` class to make all tests, as it has more columns and allow testing all behaviors. This would avoid future contributors facing the misguiding question on why some tests uses `UniqueData` and others use `Types` - as happened to me when reviewing.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,212 @@
...
+        where: (t) => t.number.equals(1),
+      );
+
+      expect(updated.length, 2);
+      expect(updated.every((row) => row.number == 42), isTrue);
+      // Emails should remain unchanged
+      var emails = updated.map((e) => e.email).toSet();
+      expect(emails,
+          containsAll(['first@serverpod.dev', 'duplicate@serverpod.dev']));
+    });
```
marcelomendoncasoares: This test is mostly duplicated with the one below - only the number itself is different, from 42 to 100. The description on the below test seems clearer to me.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,212 @@
...
+
+    tearDown(() async {
+      await Types.db.deleteWhere(
+        session,
+        where: (_) => Constant.bool(true),
+      );
+    });
+
+    test(
+        'when updating where anInt equals 1 then only matching rows are updated',
```
marcelomendoncasoares: On the other test file, the corresponding test for this one had a clearer description:

`when updating specific columns by id then only those columns are updated and others remain unchanged`

To achieve its full scope, would be needed to only add a check that other fields on updated are not changed.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,212 @@
...
+
+      expect(updated.length, 2);
+      expect(updated.every((row) => row.number == 200), isTrue);
+      // Emails should remain unchanged
+      var emails = updated.map((e) => e.email).toSet();
+      expect(
+          emails, containsAll(['second@serverpod.dev', 'third@serverpod.dev']));
+    });
+  });
+
```
marcelomendoncasoares: The other test file covered an interesting use case that seems missing in this one (line 55):

`when updating by id within a transaction then update is part of transaction`

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,212 @@
...
+      var updated = await Types.db.updateWhere(
+        session,
+        (t) => [t.anInt(99), t.aString('was_null')],
+        where: (t) => t.anInt.equals(null),
+      );
+
+      expect(updated.length, 1);
+      expect(updated.first.anInt, 99);
+      expect(updated.first.aString, 'was_null');
+    });
```
marcelomendoncasoares: The other test file covered a "set to null" use case that seems missing in this one (line 182):

`when updating a column to null then the column is set to null`

marcelomendoncasoares (commented):
Seems very solid to me! Only some minor comments to enforce standards between database methods and to improve the test coverage.

modules/new_serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/generated/legacy_email_password.dart
```diff
@@ -411,6 +411,36 @@ class LegacyEmailPasswordRepository {
...
+    return session.db.updateById<LegacyEmailPassword>(
+      id,
+      columnValues: columnValues(LegacyEmailPassword.t),
+      transaction: transaction,
+    );
+  }
+
+  /// Updates all [LegacyEmailPassword]s matching the [where] expression with the specified [columnValues].
+  /// Returns the list of updated rows.
+  Future<List<LegacyEmailPassword>> updateWhere(
```
SandPod, line 452: Suggestion: Add `limit`, `offset`, `orderBy`, `orderByList`.

As specified in the issue.

Should `Include` also be added to this interface? Since this is a result?

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,174 @@
...
+  var session = await IntegrationTestServer().session();
```
SandPod: Suggestion: Use the new test framework for these new tests.

Should we use the test framework for these tests instead of using the legacy type?

SandPod (changes requested):
Interace does not cover what was specified in the issue.

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
SandPod: `IntegrationTestServer` was out first attempt at enabling integration tests when developing Serverpod. But since then we have created a test framework that is accessible through the `withServerpod(....)` function. If you search for it in the project you can find examples of it.

A big benefit of using the test framework `withServerpod(...)` is that we can then run the test concurrently. Tests using `IntegrationTestServer` commit their database changes to the database, thus polluting the database disabling the use of concurrent tests. `withServerpod(...)` instead runs all database operations inside a transaction that is reverted at the end of the test so these are never commited to the database allowing concurrent tests to be run.

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,153 @@
...
+
+import '../../test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod(
+    'Given updateById operations',
+    (sessionBuilder, endpoints) {
+      var session = sessionBuilder.build();
+
+      group('Given an existing UniqueData entry', () {
```
exaby73: Should we be grouping like this, or should be be using multiple `withServerpod`s. My current approach duplicates `Given`s which seem counter-intuitive to me

SandPod: It is fine to have multiple `withServerpod` in the same file and do the grouping there. I think this gives us better test descriptions, so please use that pattern.

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
```diff
@@ -701,6 +703,166 @@ class BuildRepositoryClass {
     });
   }
 
+  Method _buildUpdateByIdMethod(
```
SandPod, line 706: Suggestion: Please add a code generation test that validates that these methods are generated.

You can find such tests here: tools/serverpod_cli/test/generator/dart/server_code_generator/database_repository_class_test.dart

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,153 @@
...
+              anInt: null,
+              aBool: null,
+              aDouble: null,
+              aString: null,
+              aDateTime: null,
+            ),
+          );
+        });
+
+        test('when updating a column to null then the column is set to null',
```
SandPod: Suggestion: Have a non-null value as given.

Why do you need to set the value as `null` only to start the test by modifying the given?
What prevents us from not having `null` from the start and then validating that we can set it back?

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,153 @@
...
+
+        setUp(() async {
+          existingEntry = await UniqueData.db.insertRow(
+            session,
+            UniqueData(number: 1, email: 'original@serverpod.dev'),
+          );
+        });
+
+        test(
+            'when updating by id with a single column then only that column is updated',
```
SandPod:
```suggestion
            'when updating by id for a single column then only that column is updated',
```

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,153 @@
...
+            columnValues: (t) => [t.number(42)],
+          );
+
+          expect(updated, isNotNull);
+          expect(updated!.number, 42);
+          expect(updated.email, 'original@serverpod.dev');
+        });
+
+        test(
+            'when updating by id within a transaction then update is part of transaction',
```
SandPod:
```suggestion
            'when updating by id within an aborted transaction then update reverted as part of transaction',
```

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,153 @@
...
+                existingEntry.id!,
+                columnValues: (t) => [t.number(42)],
+                transaction: transaction,
+              );
+
+              expect(updated, isNotNull);
+              expect(updated!.number, 42);
+
+              // Throw to rollback the transaction
+              throw Exception('Rollback');
```
SandPod: Suggestion: Use `transaction.cancel()` here or is the Exception important?

Then we can remove the try catch around I think.

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,153 @@
...
+          }
+
+          // Verify the update was rolled back
+          var result = await UniqueData.db.findById(session, existingEntry.id!);
+          expect(result!.number, 1);
+        });
+      });
+
+      group('Given no existing data', () {
+        test('when updating by non-existent id then null is returned',
```
SandPod: Suggestion(NIT): Combine the group and the test since it is only a single test.

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
SandPod: This would allow us to add this test to the previous group.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,593 @@
...
+            'when updating where number equals 1 then all matching rows are updated',
+            () async {
+          var updated = await UniqueData.db.updateWhere(
+            session,
+            columnValues: (t) => [t.number(42)],
+            where: (t) => t.number.equals(1),
+          );
+
+          expect(updated.length, 2);
+          expect(updated.every((row) => row.number == 42), isTrue);
```
SandPod: Suggestion: Change this expect to validate the actual values.

When this fails, this will throw `expected true but was false` which is not that helpful for the developer reading the error message.
You could pull out the numbers and expect that list to be contain only 2.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,593 @@
...
+        test(
+            'when updating where number equals 1 then all matching rows are updated',
+            () async {
+          var updated = await UniqueData.db.updateWhere(
+            session,
+            columnValues: (t) => [t.number(42)],
+            where: (t) => t.number.equals(1),
+          );
+
+          expect(updated.length, 2);
```
SandPod: Suggestion: Use `hasLength(2)`.

This gives a better error message if the test fails.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,593 @@
...
+          expect(updated.every((row) => row.number == 42), isTrue);
+          // Emails should remain unchanged
```
SandPod: Suggestion: Remove redundant comment and reflect this in the `then` section of the test.

If we are expecting on this, this is probably relevant for the test and should therefore be reflected in the description.
This tells me that we are possibly testing two things:
1. We update selected columns.
2. We do not update unselected columns.

Isolating these can help guide future devs if they accidentally break one of these.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,593 @@
...
+          var emails = updated.map((e) => e.email).toSet();
+          expect(emails,
+              containsAll(['first@serverpod.dev', 'duplicate@serverpod.dev']));
+        });
+
+        test(
+            'when updating with complex where clause then only matching rows are updated',
```
SandPod: Suggestion: Same comment as above, this is really validating two things that could break idependently.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,593 @@
...
+                aDouble: 3.5,
+                aString: 'third',
+                aDateTime: DateTime(2024, 3, 1),
+              ),
+            ],
+          );
+        });
+
+        test(
+            'when updating specific columns then only those columns are updated and others remain unchanged',
```
SandPod: Suggestion: Make it clear how this test relates and differentiates with the test on line 28.

There we are testing selection on a single column, now we are testing selection on multiple columns. Perhaps we should just use Types for all tests so that these can be grouped?

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,593 @@
...
+            columnValues: (t) => [t.anInt(99), t.aString('was_null')],
+            where: (t) => t.anInt.equals(null),
+          );
+
+          expect(updated.length, 1);
+          expect(updated.first.anInt, 99);
+          expect(updated.first.aString, 'was_null');
+        });
+
+        test('when updating a column to null then the column is set to null',
```
SandPod: Suggestion: Modify this description and possibly update the test in accordance with the guidance in `update_by_id_test.dart`.

This description currently reads `Given Types entries with null values when updating a column to null the ncolumn is set to null`. But we are actually testing that we can change a non-null value back to null.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,593 @@
...
+          expect(updated.length, 1);
+          expect(updated.first.anInt, isNull);
+          expect(updated.first.aString, isNull);
+          // Other fields should remain unchanged
+          expect(updated.first.aBool, true);
+          expect(updated.first.aDouble, 1.5);
+        });
+
+        test(
+            'when using limit parameter then only specified number of rows are updated',
```
SandPod: Suggestion: Break this out into a separate group so that the description can reflect how many entries there are an how many we are matching.

`Given 5 matching entries when updating with limit 3 then 3 rows are updated`

This should probably also be combined with a test that validate the other entries are unaffected.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,593 @@
...
+            session,
+            where: (t) => t.anInt.equals(null),
+          );
+          var updatedCount =
+              allNullRows.where((r) => r.aString == 'limited_update').length;
+          expect(updatedCount, 1);
+        });
+
+        test(
+            'when using orderBy parameter then only ordered limited rows are updated',
```
SandPod: Suggestion: Validate this first, then you can use it in the `limit` test to ensure consistent order.

Telling a story with the tests from the top of the file downwards.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,593 @@
...
+            where: (t) => t.anInt.equals(1) | t.anInt.equals(2),
+          );
+          expect(notUpdatedRows.any((r) => r.aString == 'top_two_updated'),
+              isFalse);
+
+          // Clean up additional test data
+          await Types.db.deleteWhere(
+            session,
+            where: (t) => t.anInt.equals(4) | t.anInt.equals(5),
+          );
```
SandPod: Chore: Remove this.

This looks like teardown test code and will not be executed if any of the expects fails, so the test will leak data if it fails.

Since we are using the test framework the tests should clean themselves up, so you can probably just delete this.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,593 @@
...
+          var updatedCount =
+              allRows.where((r) => r.aString == 'top_two_updated').length;
+          expect(updatedCount, 2);
+
+          // Verify rows with anInt 1 and 2 were NOT updated
+          var notUpdatedRows = await Types.db.find(
+            session,
+            where: (t) => t.anInt.equals(1) | t.anInt.equals(2),
+          );
```
SandPod: Question: this test has a lot of expects, that suggests to me that we are validating many different things.

Could this be split up to different tests but with a shared `when` group to provide some more information to future developers if this fails?

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
SandPod: Any improvement to the limit tests should also apply to the offset test.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
SandPod: This also applies to the offset tests.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
SandPod: This also applies to the offset tests.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -234,6 +235,120 @@ class DatabaseConnection {
...
+    }
+
+    var setClause = columnValues.map((cv) {
+      var value = DatabasePoolManager.encoder.convert(cv.value);
+      return '"${cv.column.columnName}" = $value::${_convertToPostgresType(cv.column)}';
+    }).join(', ');
+
+    String updateQuery;
+
+    if (limit != null ||
```
SandPod: Suggestion: Break this out into a variable that explains what the comment indicates and remove the comment.

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
exaby73: Due to `withServerpod` assuming control over the transaction, when it encounters an error, it rolls back to a savepoint in a transaction it manages rather than rolling back the entire transaction. So when we call `.cancel()`, we rollback the entire transaction which then breaks subsequent tests

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,164 @@
...
+        var updated = await UniqueData.db.updateById(
+          session,
+          999999,
+          columnValues: (t) => [t.number(123)],
+        );
+
+        expect(updated, isNull);
+      });
+
+      group('Given a Types entry', () {
```
SandPod: Suggestion: These are two different givens. `Given a Types entry` and `Given a Types entry with null values`.

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,164 @@
...
+            () async {
+          var updated = await Types.db.updateById(
+            session,
+            existingEntry.id!,
+            columnValues: (t) => [t.anInt(42), t.aString('updated')],
+          );
+
+          expect(updated, isNotNull);
+          expect(updated!.anInt, 42);
+          expect(updated.aString, 'updated');
```
SandPod: Suggestion: Use shared constants for these values. Also use the values from the original model to reflect that a field should not have been updated.

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,164 @@
...
+          expect(updated!.anInt, isNull);
+          expect(updated.aString, isNull);
+          // Other fields should remain unchanged
+          expect(updated.aBool, true);
+          expect(updated.aDouble, 1.5);
+          expect(updated.aDateTime, DateTime.utc(2024, 1, 1));
+        });
+
+        test(
+            'when updating null columns to non-null values then the columns are set correctly',
```
SandPod: Suggestion: This test also validates that non selected fields are not updated, thus givin this test multiple reasons to fail. Please create a separate test for this.

What is bellow your comment in the test, same for the other tests above.

SandPod: Is this validation possibly also a duplication of the single column test (that should be split up)?

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,164 @@
...
+          expect(updated.aString, 'was_null');
+          // Other fields should remain null
+          expect(updated.aBool, isNull);
+          expect(updated.aDouble, isNull);
+          expect(updated.aDateTime, isNull);
+        });
+      });
+
+      test(
+          'when updating by id within an aborted transaction then update reverted as part of transaction',
```
SandPod:
```suggestion
          'when updating by id within an aborted transaction then update is reverted as part of transaction',
```

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
SandPod: NIT(Style - Close if you do not agree) - Then I would suggest creating a savepoint and rolling back to it instead. That removes the error handling which just bloats the test.

Check out https://github.com/serverpod/serverpod/blob/5afeafc0753447785ee499541d56c0e4bd6cfa85/tests/serverpod_test_server/test_integration/logging/migrate_session_logs_test.dart#L80-L87

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,164 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+import 'package:test/test.dart';
+
+import '../../test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod(
+    'Given updateById operations',
```
SandPod: Suggestion: Remove wrapping group.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,437 @@
...
+              UniqueData(number: 1, email: 'first@serverpod.dev'),
+              UniqueData(number: 2, email: 'second@serverpod.dev'),
+              UniqueData(number: 3, email: 'third@serverpod.dev'),
+              UniqueData(number: 1, email: 'duplicate@serverpod.dev'),
+            ],
+          );
+        });
+
+        test(
+            'when updating where number equals 1 then all matching rows are updated',
```
SandPod: Discussion: These tests do not validate that the value in the database has been updated. They just validates that what we return from the call contains the updated value.

I think the "evaluation" should be done on a "find" result since that reads the actual value from the database.
This is actually an error with the other update tests in `update_test.dart` as well. We simply validate the result from the method but not what is stored in the database.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
SandPod: @exaby73 - This has not been fixed for all places in the file. Search for `.length`

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
SandPod: this has not been fixed for all places in the file. We still have redundant comments in the test.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
SandPod: This has not been addressed.

tests/serverpod_test_server/test_integration/database_operations/crud/update_where_test.dart
```diff
@@ -0,0 +1,437 @@
...
+          // Verify non-matching row was not updated
+          var nonMatching = await Types.db.findFirstRow(
+            session,
+            where: (t) => t.anInt.equals(2),
+          );
+          expect(nonMatching!.aString, 'second');
+        });
+      });
+
+      group('Given Types entries with null values', () {
```
SandPod: Suggestion: This is actually a mix and not strictly null values. Please reflect this if it is important.

packages/serverpod/lib/src/database/concepts/column_value.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:serverpod/src/database/concepts/columns.dart';
+import 'package:serverpod/src/database/concepts/table.dart';
+
+/// A function that returns a list of [ColumnValue]s for a [Table].
```
SandPod:
```suggestion
/// A function that returns a list of [ColumnValue]s for a [UpdateTable].
```

packages/serverpod/lib/src/database/concepts/column_value.dart
```diff
@@ -0,0 +1,17 @@
...
+typedef ColumnValueListBuilder<T> = List<ColumnValue> Function(T);
```
SandPod: Should we introduce a base type so that we can type this method? 

It could be defined in this file.

```suggestion
typedef ColumnValueListBuilder<T extends UpdateTable> = List<ColumnValue> Function(T);
```

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -234,6 +235,131 @@ class DatabaseConnection {
...
+        'WHERE "${table.id.columnName}" = ${DatabasePoolManager.encoder.convert(id)} '
+        'RETURNING *';
+
+    var result = await _mappedResultsQuery(
+      session,
+      query,
+      transaction: transaction,
+    );
+
+    if (result.isEmpty) {
```
SandPod, line 271: Question: Should we throw an exception that can be caught instead so that the user does not have to handle the `nullable` return type?

We do this in the updateRow case.

Add a test that validates this behavior if we change it.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -234,6 +235,131 @@ class DatabaseConnection {
...
+      if (orderBy != null || orderByList != null) {
+        var orders = _resolveOrderBy(orderByList, orderBy, orderDescending);
+        if (orders != null && orders.isNotEmpty) {
+          var orderStrings = orders.map((o) {
+            var orderStr = o.toString();
+            return orderStr.replaceAll('"${table.tableName}".', '');
+          }).join(', ');
+          orderByClause = ' ORDER BY $orderStrings';
+        }
+      }
```
SandPod: Suggestion: Please simplify this piece of code. Perhaps a switch expression can be used?

Shouldn't this operate on the same result from the `_resolveOrderBy` on line 320? And if the result from this is not null or empty we know we either have an `orderBy` or an `orderByList` clause.

packages/serverpod_test/lib/src/transaction_manager.dart
```diff
@@ -122,4 +122,18 @@ class TransactionManager {
 
     await savepoint.release();
   }
+
+  /// Ensures the transaction stack is unlocked and cleans up any remaining savepoints.
+  Future<void> ensureTransactionIsUnlocked() async {
```
SandPod, line 127: Question: Did you add a test that validates this behaves as expected now for the test framework?

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -0,0 +1,364 @@
...
+
+import '../../test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod(
+    'Given a database entry',
+    (sessionBuilder, endpoints) {
+      var session = sessionBuilder.build();
+
+      group('when updating by id for a single column', () {
```
SandPod, line 56: Suggestion: Add a test for when no columns are specified.

SandPod: Same for the update where class.

packages/serverpod/lib/src/database/concepts/column_value.dart
exaby73: Something like:

```dart
abstract base class UpdateTable<T extends Table> {
    final T table;

    const UpdateTable(this.table);
}
```

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
exaby73: Since we need to contruct this order by clause by it's own, we can't resuse existing query builders like `SelectQueryBuilder` has. And there is this special case where we need to remove the tableName from the resolved orders from the return of `_resolveOrderBy()`. I think this should be handled by https://github.com/serverpod/serverpod/issues/3862 when we get to it

packages/serverpod_test/lib/src/transaction_manager.dart
exaby73: Where would this test live? I see no `tests` dir in `serverpod_test`

packages/serverpod/lib/src/database/concepts/column_value.dart
SandPod: Yes, then we can reference that type in all places where applicable.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
SandPod: I'm not saying we should reuse built order by clause, only the result from `_resolveOrderBy`, it is used with the same arguments here as it is above. And the return value removes the need for the outer if statement here.

packages/serverpod_test/lib/src/transaction_manager.dart
SandPod: The tests are located here `tests/serverpod_test_server/test_integration/test_tools`

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
exaby73: Done

packages/serverpod/lib/src/database/concepts/column_value.dart
exaby73: Done

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
exaby73: Done

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
exaby73: Done

packages/serverpod/lib/src/database/concepts/column_value.dart
exaby73: Done

Issue #3841, opened 2025-07-31
Switch to `Bearer` token for `authorization` header


tp:
For all Serverpod `Client` instances.

Server needs to handle this (plain text, no need for extra wrapping) and the legacy `base64` encoded `Basic` headers.

The auth key interface can be simplified in the same go, which was already started in https://github.com/serverpod/serverpod/pull/3713/files#diff-0c7b22252a03629d67df2e53d682108ea3a729da7334cb798a5d4d9e1c949b1c

PR #3875, opened 2025-08-12
feat: Switch to `Bearer` token for `authorization` header

marcelomendoncasoares:
Closes issue #3841.

There is one minor unrelated change on `packages/serverpod/lib/src/server/session.dart` to remove the unwrap call - never used, as auth is passed in plain text, not base64 encoded. It was identified while analyzing with @SandPod and was causing confusion on understanding the authentication workflow.

## Pre-launch Checklist

## Breaking changes

For clients using the old authentication packages and a custom `AuthenticationKeyManager` extension, the extended class must be changed to `BackwardsCompatibleAuthKeyManager` to keep using `Basic` authorization header. No change is needed for clients using the default `FlutterAuthenticationKeyManager` implementation.

### DIFF
--- a/packages/serverpod_client/lib/src/auth_key_manager.dart
+++ b/packages/serverpod_client/lib/src/auth_key_manager.dart
@@ -18,10 +18,10 @@ abstract class AuthenticationKeyManager {
     return toHeaderValue(key);
   }
 
-  /// Converts an authentication key to a format that can be used in a transport header.
-  /// The default implementation encodes and wraps the key in a 'Basic' scheme.
-  /// (This will automatically be unwrapped again on the server side
-  /// before being handed to the authentication handler.)
+  /// Converts an authentication key to a format that can be used in a transport
+  /// header. The default implementation encodes and wraps the key in a 'Bearer'
+  /// scheme. This will automatically be unwrapped again on the server side
+  /// before being handed to the authentication handler.
   ///
   /// To use a different scheme, override this method.
   /// The value must be compliant with the HTTP header format defined in
@@ -29,6 +29,20 @@ abstract class AuthenticationKeyManager {
   /// See:
   /// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization
   /// https://httpwg.org/specs/rfc9110.html#field.authorization
+  Future<String?> toHeaderValue(String? key) async {
+    if (key == null) return null;
+    return wrapAsBearerAuthHeaderValue(key);
+  }
+}
+
+/// Manages keys for authentication with the server using the Basic auth scheme.
+@Deprecated('Use AuthenticationKeyManager instead.')
+abstract class BackwardsCompatibleAuthKeyManager
+    extends AuthenticationKeyManager {
+  /// Converts an authentication key to a format that can be used in a transport
+  /// header. Overrides the default implementation of 'Bearer' scheme to ensure
+  /// backwards compatibility with the previous default 'Basic' scheme.
+  @override
   Future<String?> toHeaderValue(String? key) async {
     if (key == null) return null;
     return wrapAsBasicAuthHeaderValue(key);
--- a/packages/serverpod_serialization/test/auth_encoding_test.dart
+++ b/packages/serverpod_serialization/test/auth_encoding_test.dart
@@ -13,6 +13,15 @@ void main() {
     ),
   ];
 
+  /// conventional bearer tokens
+  var conventionalBearerTokens = [
+    ('abc123', 'simple token'),
+    ('eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9', 'JWT-like token'),
+    ('dGVzd,;.<>_-/\\|"\'`~!@#\$%^&*()+=[]{}_', 'special characters token'),
+    ('', 'empty token'),
+    ('very-long-token-' * 10, 'long token'),
+  ];
+
   /// standard basic auth schemes' key formats, which is also equivalent to the wrapped key format
   var standardBasicAuthKeys = [
     ('Basic dGVzd', 'valid basic scheme with base64 encoded value'),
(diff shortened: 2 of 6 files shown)


marcelomendoncasoares:
It seems one of the vector tests (that has some hard-to-remove degree of randomness) failed unrelated to the changes in this PR:

```
serverpod_test_server-1  | +1446 ~2: test_integration/database_operations/runtime_parameters/vector_query_test.dart: Given a large number of vectors with very similar distances when using conservative and aggressive index query parameters then intersection ratio of found vectors is less than 0.5.
serverpod_test_server-1  | Conservative IDs: {4762, 4802, 4748, 4735, 4614, 4643, 4702, 4680, 4671, 4685, 4742, 4634, 4826, 4655, 4667, 4825, 4805, 4871, 4882, 4839, 4558, 4540, 4568, 4556, 4874, 4572, 4619, 4865, 4819, 4541}
serverpod_test_server-1  | Aggressive IDs: {4614, 4762, 4802, 4643, 4558, 4540, 4748, 4735, 4961, 4702, 4922, 4680, 4671, 4685, 4742, 4634, 4826, 4655, 4667, 4825, 4568, 4805, 4871, 4882, 4556, 4839, 4874, 4572, 4619, 4865}
serverpod_test_server-1  | Intersection ratio: 0.875
serverpod_test_server-1  | +1446 ~2 -1: test_integration/database_operations/runtime_parameters/vector_query_test.dart: Given a large number of vectors with very similar distances when using conservative and aggressive index query parameters then intersection ratio of found vectors is less than 0.5. [E]
serverpod_test_server-1  |   Expected: a value less than <0.5>
serverpod_test_server-1  |     Actual: <0.875>
serverpod_test_server-1  |      Which: is not a value less than <0.5>
serverpod_test_server-1  |   
serverpod_test_server-1  |   package:matcher                                                                      expect
serverpod_test_server-1  |   test_integration/database_operations/runtime_parameters/vector_query_test.dart 93:9  main.<fn>.<fn>.<fn>
```

I'll open an issue to investigate on how to fix this flaky test.

packages/serverpod_serialization/test/auth_encoding_test.dart
```diff
@@ -13,6 +13,15 @@ void main() {
     ),
   ];
 
+  /// conventional bearer tokens
+  var conventionalBearerTokens = [
+    ('abc123', 'simple token'),
+    ('eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9', 'JWT-like token'),
```
SandPod: Question(NIT): Shouldn't a JWT token contain dots between the different parts of the token?

packages/serverpod_client/lib/src/auth_key_manager.dart
```diff
@@ -18,17 +18,31 @@ abstract class AuthenticationKeyManager {
...
   /// https://httpwg.org/specs/rfc9110.html#field.authorization
+  Future<String?> toHeaderValue(String? key) async {
+    if (key == null) return null;
+    return wrapAsBearerAuthHeaderValue(key);
+  }
+}
+
+/// Manages keys for authentication with the server using the Basic auth scheme.
+@Deprecated('Use AuthenticationKeyManager instead.')
+abstract class BackwardsCompatibleAuthKeyManager
```
SandPod: Suggestion: Rename this to `BasicHeaderAuthenticationKeyManager` or something more fitting.

Using Basic auth should be allowed, it is just that we will be using Bearer for our new tokens.

Perhaps we should remove the default implementation of `toHeaderValue` from `AuthenticationKeyManager` and make the user pick between `BearerHeaderAuthenticationKeyManager` and `BasicHeaderAuthenticationKeyManager` that have the implementations? That would make it an explicit choice that makes this breaking change visible for all users?

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
```diff
@@ -143,10 +143,95 @@ void main() async {
...
+      var reflectedKey = await client.echoRequest.echoAuthenticationKey();
+      expect(reflectedKey, key);
+
+      var receivedToken = await tokenInspectionCompleter.future;
+      expect(receivedToken, key);
+    });
+
+    test(
+        'when calling an endpoint method with a parameter '
+        'then it should receive plain auth key (i.e. in original format)',
```
SandPod: Suggestion: Make it clear that the token we are inspecting is what the authentication handler receives.

General question: Is it helpfull that we strip the schema before the authentication handler receives the request? perhaps this is valuable information and makes it a bit inconsistent?

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
```diff
@@ -143,10 +143,95 @@ void main() async {
       );
     });
   });
+
+  group('Given auth key with Bearer HTTP header format', () {
```
SandPod, line 147: Question: What happens if the user uses a token that is invalid? For example an extra space at the end after the token?

SandPod (commented):
Nice minimal change!

Had a couple of points for discussion and improvements.

SandPod:
We should also write down a migration documentation for this in the docs Repo. I suggest starting with this already now so that we keep track of what we have changed.

packages/serverpod_serialization/test/auth_encoding_test.dart
marcelomendoncasoares: You're right, we can improve this example to be more like a real JWT token. Done!

packages/serverpod_client/lib/src/auth_key_manager.dart
marcelomendoncasoares: Agreed! If our goal is to support both `Basic` and `Bearer` as valid alternatives, having the `toHeaderValue` abstract on the base class makes total sense. Done!

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
marcelomendoncasoares: I've added a test for an invalid token case and it surfaced an untreated exception that led to 500 internal error - instead of the expected 400 bad request. Fixed!

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
marcelomendoncasoares: Suggestion done. About the question, I don't see an issue from the user perspective, as they set up an `AuthenticationKeyManager` choosing the scheme and just pass keys to it, without having to worry about when the keys are wrapped or not. On Serverpod's internal code, however, I think we lose a bit of clarity by not knowing when the `String` is of the plain key or the wrapped key at each point in the code.

I don't think this refactor would be worth exclusively for the internal perspective, but if we see value in exposing the scheme info to the user, an alternative would be to make the `toHeaderValue` convert the `String` key to an `AuthHeader` class storing the `scheme` and `key`. Then, this object could be passed to the `authenticationHandler` instead of the key as `String`. It would be a breaking change to the `authenticationHandler` contract, so now would be the best moment to do it - if it makes sense. Or leave to create a secondary `headerAuthenticationHandler` in the future to support both.

Do you think its worth creating a separate issue for this discussion, @SandPod?

SandPod: I think it makes sense to separate the discussion from this change, as you said, this would more be an internal change in Serverpod regarding the auth key contract.

I'm thinking about how this should work when working with the Relic web server, should you be able to connect the same authenticationHandler there as well. I do belive that would make sense. But lets leave it out of this PR and discuss it separately.

Feel free to close this conversation when that discussion is posted somewhere.

packages/serverpod/lib/src/server/server.dart
```diff
@@ -477,12 +477,16 @@ class Server {
...
       authenticationHeaderValue = request.headers.authorization?.headerValue;
       authenticationKey = unwrapAuthHeaderValue(authenticationHeaderValue);
-    } on InvalidHeaderException catch (_) {
-      // Use authHeaderValueFromHeader in the error message as it's the (potentially problematic) value we read
-      return ResultStatusCode(
-        400,
-        'Request has invalid "authorization" header: $authenticationHeaderValue',
-      );
+    } catch (e) {
+      if (e is AuthHeaderEncodingException || e is InvalidHeaderException) {
```
SandPod, line 481: Question: Where did this excpetion come from? Relic?

@nielsenko - If the static validation on all headers is still present in relic, that needs to be fixed before we can release it (making it configurable per path and not a global true or false)

nielsenko: @SandPod Typed headers are only validated on access in relic.

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
marcelomendoncasoares: @coderabbitai Create an issue for this internal improvement from the discussion.

PR #3878, opened 2025-08-12
Feat: Custom object return for password hash validator and named parameters

### LINKED ISSUE #3698: Improve `PasswordHashValidator`
### Problem to Solve

```
typedef PasswordHashValidator = Future<bool> Function(
  String password,
  String email,
  String hash, {
  void Function({required String passwordHash, required String storedHash})?
      onValidationFailure,
  void Function(Object e)? onError,
});
```

The onError & onValidationFailure don't feel like dart functions. Why not just throw a custom Exception that needs to be catched instead?

### Proposal

Use named params to prevent a user to pass the incorrect value if the same order is not used. 
Using named params will make it less likely to cause mistakes
```
typedef PasswordHashValidator = Future<bool> Function({
  required String password,
  required String email,
  required String hash, 
  void Function({required String passwordHash, required String storedHash})?
      onValidationFailure,
  void Function(Object e)? onError,
});
```

### Use Case

Making it less likely to make mistakes & handle errors

### Alternatives

I am creating a wrapper with a completer around the PasswordHashValidator

### Additional context

_No response_

- [ ] I want to work on this

yashas-hm:
Fixes #3698
Closes #3698

## Summary of Changes
- Structured password-hash validation added:
  - New PasswordHashValidatorResponse { success: bool, error?: PasswordHashValidatorException }.
  - New exceptions: PasswordHashValidatorException and PasswordHashValidationFailedException (includes computed vs stored hash for safe debugging).
- Email auth flow updates:
  - Emails.authenticate uses structured validator; safe logging;
- Exports:
  - module.dart now exports password_hash_validator_exception.dart.

## Pre-launch Checklist

## Breaking Changes
- Emails.validatePasswordHash return type changed from bool to PasswordHashValidatorResponse.
- PasswordHashValidator typedef updated to:
  - Future<PasswordHashValidatorResponse> Function({required String password, required String email, required String hash})
- Removed callbacks onError and onValidationFailure; callers must inspect response.error instead.

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_server/lib/src/business/email_auth.dart
+++ b/modules/serverpod_auth/serverpod_auth_server/lib/src/business/email_auth.dart
@@ -25,10 +25,10 @@ Future<String> defaultGeneratePasswordHash(String password) =>
 /// Warning: Using a custom hashing algorithm for passwords
 /// will permanently disrupt compatibility with Serverpod's
 /// password hash validation and migration.
-Future<bool> defaultValidatePasswordHash(
-  String password,
-  String email,
-  String hash, {
+Future<bool> defaultValidatePasswordHash({
+  required String password,
+  required String email,
+  required String hash, 
   void Function({
     required String passwordHash,
     required String storedHash,
@@ -627,9 +627,9 @@ class Emails {
     void Function(Object e)? onError,
   }) =>
       AuthConfig.current.passwordHashValidator(
-        password,
-        email,
-        hash,
+        password: password,
+        email: email,
+        hash: hash,
         onError: onError,
         onValidationFailure: onValidationFailure,
       );
--- a/modules/serverpod_auth/serverpod_auth_server/lib/src/business/email_auth.dart
+++ b/modules/serverpod_auth/serverpod_auth_server/lib/src/business/email_auth.dart
@@ -25,15 +25,10 @@ Future<String> defaultGeneratePasswordHash(String password) =>
 /// Warning: Using a custom hashing algorithm for passwords
 /// will permanently disrupt compatibility with Serverpod's
 /// password hash validation and migration.
-Future<bool> defaultValidatePasswordHash({
+Future<PasswordHashValidatorResponse> defaultValidatePasswordHash({
   required String password,
   required String email,
-  required String hash, 
-  void Function({
-    required String passwordHash,
-    required String storedHash,
-  })? onValidationFailure,
-  void Function(Object e)? onError,
+  required String hash,
 }) async {
   try {
     return await PasswordHash(
@@ -41,10 +36,15 @@ Future<bool> defaultValidatePasswordHash({
       legacySalt: EmailSecrets.legacySalt,
       legacyEmail: AuthConfig.current.extraSaltyHash ? email : null,
       pepper: EmailSecrets.pepper,
-    ).validate(password, onValidationFailure: onValidationFailure);
-  } catch (e) {
-    onError?.call(e);
-    return false;
+    ).validate(password);
+  } catch (e, s) {
+    return PasswordHashValidatorResponse(
+      success: false,
+      error: PasswordHashValidatorException(
+        errorMessage: e,
+        stackTrace: s,
+      ),
+    );
   }
 }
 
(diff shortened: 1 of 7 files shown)


SandPod:
@yashas-hm - As discussed in the issue, to make the interface responsibility as clear as possible for the implementer we should use typed responses that the framework can then act on.

The `onError` can probably be caught in the code that calls the method. However, the method should either return a success response or a validation error response with the information required to achieve the functionality that exists today. I suggest creating a sealed base class that has two final implementaiton that can be returned.

yashas-hm:
@SandPod would something like this work?

```dart
sealed class PasswordValidationResult {
  const PasswordValidationResult();
}

final class PasswordValidationSuccess extends PasswordValidationResult {
  const PasswordValidationSuccess();
}

final class PasswordValidationError extends PasswordValidationResult {
  final String message;
  final int? errorCode; // optional

  const PasswordValidationError(this.message, {this.errorCode});
}
```

is yes i can go ahead and implement an approach like this?

SandPod:
@yashas-hm - Yes, I was thinking about something in that direction, but instead of PasswordValidationError, use PasswordValidationFailed and include the information that was used in the validation failed callback from before.

yashas-hm:
Sure sounds good.

tests/serverpod_test_server/test_integration/auth_module/email_auth_provider/email_auth_provider_test.dart
```diff
@@ -349,18 +349,17 @@ void main() async {
...
       );
+      if (validationResponse is! PasswordValidationSuccess) {
+        if (validationResponse is PasswordValidationFailed) {
+          actualPasswordHash = validationResponse.passwordHash;
+          actualStoredHash = validationResponse.storedHash;
+        }
+      }
 
       expect(actualStoredHash, hunter4PasswordHash);
       expect(actualPasswordHash, notHunter4PasswordHash);
```
SandPod: Chore: Replace this with an expect for `PasswordValidationFailed` and then validate the fields of the object.

Remove the nested if statements and the assigning of the values.

tests/serverpod_test_server/test_integration/auth_module/email_auth_provider/email_auth_provider_test.dart
```diff
@@ -339,7 +339,7 @@ void main() async {
           'test7@serverpod.dev',
           hunter4PasswordHash,
         ),
-        isFalse,
+        isNot(isA<PasswordValidationSuccess>()),
```
SandPod:
```suggestion
        isA<PasswordValidationFailed>(),
```

tests/serverpod_test_server/test_integration/auth_module/email_auth_provider/email_auth_provider_test.dart
SandPod: Also update the test description to reflect the returned object. This test could probably be combined with the previous test.

modules/serverpod_auth/serverpod_auth_server/lib/src/business/email_auth.dart
```diff
@@ -187,17 +186,20 @@ class Emails {
     }
 
     // Check old password
-    if (!await validatePasswordHash(
+    final validationResponse = await validatePasswordHash(
```
SandPod: Suggestion: Remove the `PasswordValidationError` type and wrap this call in a try catch instead and log and return false if something is thrown.

modules/serverpod_auth/serverpod_auth_server/lib/src/business/email_auth.dart
```diff
@@ -187,17 +186,20 @@ class Emails {
...
+    if (validationResponse is! PasswordValidationSuccess) {
+      if (validationResponse is PasswordValidationError) {
         session.log(
-          ' - error when validating password hash: $e',
+          ' - error when validating password hash: ${validationResponse.errorMessage}',
           level: LogLevel.error,
         );
-      },
-    )) {
+      }
```
SandPod: Suggestion: Flatten this code with an early return for `PasswordValidationFailed`

modules/serverpod_auth/serverpod_auth_server/test/password_hash/argon2id_password_hash_test.dart
```diff
@@ -156,7 +159,8 @@ void main() {
         legacySalt: salt,
       );
 
-      expect(await passwordHash.validate('chaser1'), isFalse);
+      expect(await passwordHash.validate('chaser1'),
+          isNot(isA<PasswordValidationSuccess>()));
```
SandPod: Chore: Update this for all tests that are expected to fail validation.

Remember to update the descriptions, we are not returning `true` or `false` anymore.

```suggestion
          isA<PasswordValidationFailed>());
```

SandPod (changes requested):
- Class should be removed
- Tests need to be updated
- Call to validation method refined

Then this is ready to be merged.

tests/serverpod_test_server/test_integration/auth_module/email_auth_provider/email_auth_provider_test.dart
```diff
@@ -332,35 +335,22 @@ void main() async {
...
-          'test7@serverpod.dev',
-          hunter4PasswordHash,
-        ),
-        isFalse,
-      );
-    });
-
     test(
-        'then validation failure callback is called with generated hash and passed in hash.',
+        'then validation fails and validation failure callback is called with generated hash and passed in hash.',
```
SandPod:
```suggestion
        'then validation returns PasswordValidationFailed with generated and passed in hash.',
```

Issue #3879, opened 2025-08-12
Running integration tests on Windows fail with `Internal server error. Zoned exception.`


marcelomendoncasoares:
### Describe the bug

Due to the way the `SystemResources` class is implemented on the [system_resources](https://pub.dev/packages/system_resources) package, starting a serverpod server on Windows for the second time (as happens in tests) results in a zoned exception:

```
SERVERPOD version: 3.0.0-alpha.1, dart: 3.8.1 (stable) (Wed May 28 00:47:25 2025 -0700) on "windows_x64", time: 2025-08-12 17:31:29.880357Z
runMode: production
serverId: default
role: monolith
loggingMode: normal
applyMigrations: false
applyRepairMigration: false
SERVERPOD initialized, time: 2025-08-12 17:31:29.888452Z
2025-08-12 17:31:31.083136Z ERROR: Internal server error. Zoned exception.
2025-08-12 17:31:31.083136Z ERROR: ProcessException: O sistema não pode encontrar o arquivo especificado.

  Command: uname -sm
#0      _ProcessImpl._runAndWait (dart:io-patch/process_patch.dart:516:7)
#1      _runNonInteractiveProcessSync (dart:io-patch/process_patch.dart:683:18)
#2      Process.runSync (dart:io-patch/process_patch.dart:79:12)
#3      _filename (package:system_resources/src/dylib.dart:15:25)
#4      loadLibsysres.<anonymous closure> (package:system_resources/src/dylib.dart:50:32)
<asynchronous suspension>
#5      SystemResources.init (package:system_resources/src/base.dart:11:18)
<asynchronous suspension>
#6      HealthCheckManager.start (package:serverpod/src/server/health_check_manager.dart:42:7)
<asynchronous suspension>
#7      Serverpod._unguardedStart (package:serverpod/src/server/serverpod.dart:734:7)
<asynchronous suspension>
#8      Serverpod.start.<anonymous closure> (package:serverpod/src/server/serverpod.dart:612:9)
<asynchronous suspension>
#9      Serverpod.start (package:serverpod/src/server/serverpod.dart:611:7)
<asynchronous suspension>
#10     main.<anonymous closure>.<anonymous closure> (file:///C:/Users/Marcelo/GitHub/serverpod/tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart:33:7)
auth_header_test.dart:33
<asynchronous suspension>
#11     Future._kTrue (dart:async/future.dart:660:3)
[...]

PR #3880, opened 2025-08-12
fix: Fix `system_resources` error on Windows when starting Serverpod multiple times

marcelomendoncasoares:
Quick fix for #3879.

A more permanent alternative might be to hide the [system_resources](https://pub.dev/packages/system_resources) behind a custom package that uses [win32](https://pub.dev/packages/win32) to access Windows resources and provide full API compatibility. Then, this fix can be removed with no harm.

## Pre-launch Checklist

## Breaking changes

N/A.

### DIFF
--- a/packages/serverpod/lib/src/server/health_check_manager.dart
+++ b/packages/serverpod/lib/src/server/health_check_manager.dart
@@ -32,6 +32,12 @@ class HealthCheckManager {
   Future<void> start() async {
     _running = true;
 
+    if (Platform.isWindows) {
+      stderr.writeln(
+          'WARNING: CPU and memory usage metrics are not supported on Windows.');
+      return;
+    }
+
     try {
       await SystemResources.init();
     } catch (e) {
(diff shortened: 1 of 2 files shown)


packages/serverpod/lib/src/server/health_check_manager.dart
```diff
@@ -32,6 +32,12 @@ class HealthCheckManager {
   Future<void> start() async {
     _running = true;
 
+    if (Platform.isWindows) {
+      stderr.writeln(
```
exaby73, line 36: Would this error possibly always be printed for users whenever they start the dev server?

nielsenko (commented):
This seems like a pragmatic approach for now. We could choose to implement the needed libsysres for [windows](https://github.com/jonasroussel/system_resources/tree/main/lib/src/libsysres) as well, and upstream a PR.

packages/serverpod/lib/src/server/health_check_manager.dart
marcelomendoncasoares: This is actually the same approach that happens today, but avoiding the exception that is leaking through the try-catch block that is below.

PR #3885, opened 2025-08-13
fix: Possible corrupted state for downloaded templates

### LINKED ISSUE #3790: Downloading templates may be left in a corrupted state
### Describe the bug

As per https://github.com/serverpod/serverpod/pull/3781#discussion_r2200726192, if the download templates fails but the dir exists, the next time a create is attempted, we only check if the dir exists and therefore the case may be that the actual template files does not exist

### To reproduce

Run `serverpod create`. Delete the template files leaving the dir. Templates are not re-downloaded on next `serverpod create`

### Expected behavior

The templates should always be in a valid state

### Library version

2.9.1

### Platform information

Any

### Additional context

_No response_

- [x] I want to work on this

exaby73:
When downloading templates, there can be a case where the templates are in a corrupted state if the download was stopped unexpectedly.
Since we only check if the template dir exists, this causes us to possible copy over corrupted files during `serverpod create`

This PR changes the download strategy to first download templates in a temp dir, then move it to the final location.
On top of that, we also generate and store checksums in the repo, which upon `serverpod create` we use to verify the templates are not corrupted.

Closes #3790

## Pre-launch Checklist

### DIFF
--- a/tools/serverpod_cli/lib/src/downloads/resource_manager.dart
+++ b/tools/serverpod_cli/lib/src/downloads/resource_manager.dart
@@ -139,9 +139,7 @@ class ResourceManager {
     log.info('Downloading templates for version $templateVersion');
     if (!versionedDir.existsSync()) versionedDir.createSync(recursive: true);
 
-    var tempDir = Directory(p.join(
-        versionedDir.path, 'temp_${DateTime.now().millisecondsSinceEpoch}'));
-    tempDir.createSync(recursive: true);
+    var tempDir = await Directory.systemTemp.createTemp('serverpod_templates_');
 
     try {
       var response = await http.get(Uri.parse(packageDownloadUrl));


tests/bootstrap_project/test/template_checksum_test.dart
```diff
@@ -0,0 +1,208 @@
...
+      final (:projectName, :commandRoot) = createRandomProjectName(tempPath);
+
+      var serverpodDir = Directory(path.join(
+        Platform.environment['HOME'] ?? Platform.environment['USERPROFILE']!,
+        '.serverpod',
+      ));
+      if (serverpodDir.existsSync()) {
+        try {
+          await serverpodDir.delete(recursive: true);
+        } catch (_) {}
```
nielsenko, line 67: Why is this try catch here?

tests/bootstrap_project/test/template_checksum_test.dart
```diff
@@ -0,0 +1,208 @@
...
+      expect(
+        Directory(path.join(tempPath, projectName)).existsSync(),
+        isTrue,
+        reason: 'Project directory was not created',
+      );
+
+      try {
+        await Directory(path.join(tempPath, projectName))
+            .delete(recursive: true);
+      } catch (_) {}
```
nielsenko, line 94: Why not test explicitly for existence instead of squashing all exceptions?

tools/serverpod_cli/lib/src/downloads/resource_manager.dart
```diff
@@ -103,33 +104,72 @@ class ResourceManager {
...
-      var outFileName = p.join(templateDirectory.path, file.name);
-      if (file.isFile) {
-        var outFile = File(outFileName);
-        outFile = await outFile.create(recursive: true);
-        await outFile.writeAsBytes(file.content);
-      } else {
-        await Directory(outFileName).create(recursive: true);
+    var tempDir = Directory(p.join(
+        versionedDir.path, 'temp_${DateTime.now().millisecondsSinceEpoch}'));
+    tempDir.createSync(recursive: true);
```
nielsenko:
```suggestion
    var tempDir = Directory.systemTemp.createTempSync();
```
this will allow the OS to cleanup if we fail

tools/serverpod_cli/lib/src/downloads/resource_manager.dart
```diff
@@ -103,33 +104,72 @@ class ResourceManager {
...
+
+      // Const constructor was introduced in version 4.0.0 of archive package.
+      // At the time this was written, that version was 2 days old and we don't
+      // want to force a constraint on the package for this.
+      // ignore: prefer_const_constructors
+      var unzipped = GZipDecoder().decodeBytes(data);
+      var archive = TarDecoder().decodeBytes(unzipped);
+
+      var tempTemplateDir =
+          Directory(p.join(tempDir.path, 'serverpod_template'));
```
nielsenko, line 156: Could be handled with the `prefix` argument to `createTempSync` as well

tools/serverpod_cli/lib/src/downloads/template_checksum_manager.dart
```diff
@@ -0,0 +1,183 @@
...
+            'projectname_client',
+            'projectname_flutter',
+            'github'
+          ];
+          compositeChecksum = templateChecksums['server'] as String?;
+          break;
+        case 'module':
+          dirsToCheck = ['modulename_server', 'modulename_client'];
+          compositeChecksum = templateChecksums['module'] as String?;
+          break;
```
nielsenko: This feels fishy with the hardcoded dir names..

tools/serverpod_cli/test/downloads/template_checksum_manager_test.dart
```diff
@@ -0,0 +1,330 @@
...
+import 'package:path/path.dart' as p;
+import 'package:serverpod_cli/src/downloads/template_checksum_manager.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('TemplateChecksumManager', () {
+    late Directory tempDir;
+
+    setUp(() {
+      tempDir = Directory.systemTemp.createTempSync('test_checksum_');
```
nielsenko, line 11: So you about this..

nielsenko (changes requested):
Why are you using a bash script to calculate the checksums when updating but dart code when checking? Why not just use dart code for both. Expecting bash on windows, while it works on github actions due to WSL, is still a bit opinionated.

Also, I feel this is perhaps a bit over-engineered. Downloading to a tmp directory and moving it in place with an atomic move upon success seems fine. Calculating the checksums on every create will add to an already slow experience.

Perhaps someone else can give their perspective. @SandPod / @marcelomendoncasoares

Issue #3448, opened 2025-04-22
Support non-`int` `userId` in `SessionLogEntry`


tp:
Currently `SessionLogEntry` only supports `int` on the `authenticatedUserId` field, and since that model is used in the database, moving to String or `json` would entail a big data migration (+ breaking change for any `int` ID specific queries).

We do need to define a migration plan and support arbitrary user IDs.

Relates to https://github.com/serverpod/serverpod/pull/3443#discussion_r2053962243

SandPod:
We are OK with changing this as a table dropping change. In the upgrade instructions we will need to encourage the developer to backup this table if the information is valuable.

PR #3893, opened 2025-08-18
fix!: Switch to logging authenticated user id as String.

SandPod:
This PR updates the logging of authenticated user IDs in Serverpod Insights to use String values instead of integers. This change improves flexibility and ensures compatibility with systems where user identifiers may not always be numeric.

It also introduces a utility method that allows users to migrate their existing session log user IDs to this new `String` field.

Fixes: https://github.com/serverpod/serverpod/issues/3448

## Pre-launch Checklist

## Breaking changes

Yes, this change will require developers to run a database migration.

### DIFF
--- a/packages/serverpod/lib/serverpod.dart
+++ b/packages/serverpod/lib/serverpod.dart
@@ -24,6 +24,7 @@ export 'package:serverpod/database.dart';
 export 'package:serverpod_serialization/serverpod_serialization.dart';
 export 'package:serverpod/src/util/request_extension.dart';
 export 'package:serverpod/src/generated/log_level.dart';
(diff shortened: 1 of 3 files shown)


SandPod:
@coderabbitai review

SandPod:
@coderabbitai full review

marcelomendoncasoares (commented):
Is the PR tittle correct? Or is the exclamation mark ("fix*!*:") on the prefix just a typo? On the code, LGTM!

SandPod:
@marcelomendoncasoares - I marked this as a breaking change (by adding `!` to the type) since the user will need to run a migration in order for this to work, otherwise there will be exceptions from the database if the table is missing.

PR #3894, opened 2025-08-18
feat: Add parallel generation script

exaby73:
This PR adds a new script to parallelize the code generation in batches. By default, the batch size is `4` which can be configured using the `BATCH_SIZE` env var.

Example usage:

```bash
# Use default batch size of 4
./util/generate_all_parallel

# Configure batch size to 2
BATCH_SIZE=2 ./util/generate_all_parallel
```

## Pre-launch Checklist


util/generate_all_parallel
```diff
@@ -0,0 +1,120 @@
...
+        eval "$project &"
+    done
+    
+    # Wait for this batch to complete
+    wait
+    echo -e "\n✅ Batch $batch_num completed!\n"
+}
+
+# Makes script exit on first non-zero error code
+set -e
```
SandPod: Question: Now that it is running in parallel, should it abort on first non-zero exit code?

Perhaps a better solution would be to record any non-zero exit code and then exit with this at the end.

util/generate_all_parallel
```diff
@@ -0,0 +1,120 @@
...
+
+BASE=`pwd`
+CLI_DIR=$BASE/tools/serverpod_cli
+CLI=$CLI_DIR/bin/serverpod_cli.dart
+
+# Set batch size (adjust this to control how many processes run at once)
+# Default is 4, but you can change this based on your system capabilities
+BATCH_SIZE=${BATCH_SIZE:-4}
```
SandPod: Suggestion/Discussion: Could we derive this from the number of cores available on the platform we are running on?

That would give a reasonable default in all places.

util/generate_all_parallel
```diff
@@ -0,0 +1,120 @@
...
+
+echo "pub get cli"
+cd $CLI_DIR
+dart pub get
+
+echo -e "\n🚀 Starting batched parallel generation (batch size: $BATCH_SIZE)...\n"
+
+# Define all generation tasks
+declare -a tasks=(
```
SandPod: Suggestion: Could we combine the first commit i have in my draft branch (will break this out into a separate PR) with this? That make this file only focus on running a something in parallel.

Draft PR here: https://github.com/serverpod/serverpod/pull/3893

util/generate_all_parallel
```diff
@@ -0,0 +1,120 @@
...
+    'generate_project "serverpod" "$BASE/packages/serverpod" ""'
+    'generate_project "chat_server" "$BASE/examples/chat/chat_server" ""'
+    'generate_project "auth_example_server" "$BASE/examples/auth_example/auth_example_server" ""'
+    'generate_project "modulename_server" "$BASE/templates/serverpod_templates/modulename_server" ""'
+    'generate_project "projectname_server" "$BASE/templates/serverpod_templates/projectname_server" ""'
+    '(cd "$BASE/templates/serverpod_templates/projectname_server_upgrade" && echo "Starting: projectname_server_upgrade" && generateServerUpgradeFiles 2>&1 | while read line; do echo "[projectname_server_upgrade] $line"; done && echo "✅ Done: projectname_server_upgrade")'
```
SandPod: Question: Why combine these into a single generate?

exaby73: Could you elaborate on this? If you're asking about this special handling for `projectname_server_upgrade` then we have a special case for this. Now that I look at it, I think I can refactor this to be simpler and have this `echo`s and `cd`s in the `generateServerUpgradeFiles` function. This was a copy-paste from the existing sequential script that I forgot to refactor. Let me know if that answers your question

util/generate_all_parallel
exaby73: I didn't want to use all cores since in my experience, when I had a script use 10 cores (my laptop has 10 cores), it would cause the entire machine to lag to the point that the mouse was lagging too. So since most people for the most part run 6+ cores at least, I made 4 the default to keep some room for other stuff to run since this script still does take around a minute to run (down from 2.5 minutes on my machine)

util/generate_all_parallel
exaby73: My idea was that exiting immediately would give better feedback to the developer and the error messages would be available right there instead of having to possibly scroll up to find the failed job. Maybe we can introduce a `--fail-fast` flag to still support this while by default changing it to your suggestion?

util/generate_all_parallel
marcelomendoncasoares: Maybe a good default could be `max(<all_cores> - 2, 1)`? Because assuming 4 could be bad for people contributing from less capable machines, or two few for people with more capable ones. By using the `max`, we also ensure to only run in parallel if the machine has at least 4 cores (which would then use 2).

SandPod: Understand the motivation!

I think @marcelomendoncasoares solution could be a good middle ground. Could even use `max(<all_cores> / 2, 1)` if we would like to preserve some more room.

util/generate_all_parallel
SandPod: I don't think I read this correctly. I was under the impression we were patching multiple projects. Sorry!

util/generate_all_parallel
SandPod: This assumes that the exit of the script will not flush other outputs.
Perhaps recoding failed calls and printing them at the end would be a better solution? Given this is not to complicated.

Otherwise, a flag for this sounds like a good solution; the default value could even be true, and you could turn it off instead if we believe this provides a better developer experience.

util/generate_all_parallel
exaby73: As discussed in Slack, I'll wait for https://github.com/serverpod/serverpod/pull/3897 to be merged and rebase this PR onto main, then proceed to refactor to use the new shared definitions

util/generate_all_parallel
exaby73: Yes this sounds like a sound solution. I think having a `-2` sounds better since we can maximize parallelism while not completely blocking system resources

marcelomendoncasoares:
One issue I've just experienced with this script is that, if I try to cancel, it will only cancel the current batch. For a low `BATCH_SIZE`, this means I'll have to fire `^C` several times to abort the script - up to once per package if using concurrency of 1.

exaby73:
Since this is a good utility to have currently even if it's not exactly what we want, we can merge it and think about improvements in the future.

PR #3901, opened 2025-08-20
chore: Use defaults for Dart analyzer language settings

christerswahn:
Use defaults for Dart analyzer language settings. Instead of enabling strict-casts, strict-inference, and strict-raw-types language rules centrally in serverpod_lints, and then disabling them again in each serverpod package, don't specify them at all and use the Dart defaults.

## Pre-launch Checklist

## Breaking changes

n/a

### DIFF
--- a/packages/serverpod_lints/lib/internal.yaml
+++ b/packages/serverpod_lints/lib/internal.yaml
@@ -18,9 +18,3 @@ linter:
     sort_pub_dependencies: true
     unawaited_futures: true
     unnecessary_final: false
-
-analyzer:
-  language:
-    strict-casts: true
-    strict-inference: true
-    strict-raw-types: true
--- a/modules/serverpod_auth/serverpod_auth_server/analysis_options.yaml
+++ b/modules/serverpod_auth/serverpod_auth_server/analysis_options.yaml
@@ -9,10 +9,3 @@ linter:
     prefer_final_locals: false
     prefer_final_parameters: false
     sort_pub_dependencies: false
-
-analyzer:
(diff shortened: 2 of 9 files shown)


packages/serverpod_lints/lib/internal.yaml
```diff
@@ -18,9 +18,3 @@ linter:
     sort_pub_dependencies: true
     unawaited_futures: true
     unnecessary_final: false
-
-analyzer:
-  language:
-    strict-casts: true
-    strict-inference: true
-    strict-raw-types: true
```
nielsenko, line 26: It is odd that we enable it in the serverpod_lints package, only to disable it everywhere we use it (except relic). 

I do think it was/is supposed to be a temporary measure, and that _we_ intended/intends for all code to align with the `serverpod_lints` package eventually.

However, I also think we have become wiser since. Perhaps we should revisit all the extra rules enabled in `serverpod_lints` and decide if they carry their weight.

I approve this PR, but I suggest we discuss the lints in general in a wider forum.

nielsenko (approved):
I think we should discuss the lints in a larger forum.

packages/serverpod_lints/lib/internal.yaml
christerswahn: Yes, `serverpod_lints` ought not impose stricter rules for the _language_ section  than the Dart defaults IMHO.

Defining a common style, e.g. final / var, etc, can be of value to help consistency in the code base.

Agree that it's good to bring it up in the team.

PR #3903, opened 2025-08-20
feat: Add cache control headers to dynamic routes

exaby73:
This PR adds cache control headers to dynamic routes.

Fixes: https://github.com/serverpod/serverpod/issues/2534

## Pre-launch Checklist

### DIFF
--- a/packages/serverpod/lib/src/web_server/web_server.dart
+++ b/packages/serverpod/lib/src/web_server/web_server.dart
@@ -153,23 +153,7 @@ class WebServer {
   ) async {
     // TODO: route.setHeaders(request.response.headers);
     try {
-      final result = await route.handleCall(session, context);
-
-      if (route is WidgetRoute && result is ResponseContext) {
-        final transformedResponse = result.response.copyWith(
-          headers: result.response.headers.transform(
-            (mh) => mh.cacheControl = CacheControlHeader(
-              maxAge: 0,
-              sMaxAge: 0,
-              noCache: true,
-              noStore: true,
-            ),
-          ),
-        );
-        return result.withResponse(transformedResponse);
-      }
-
-      return result;
+      return await route.handleCall(session, context);
     } catch (e, stackTrace) {
       final request = context.request;
       await _reportException(
@@ -313,8 +297,17 @@ abstract class WidgetRoute extends Route {
     }
 
     final mimeType = widget is JsonWidget ? MimeType.json : MimeType.html;
+
+    final headers = Headers.build(
+      (mh) => mh.cacheControl = CacheControlHeader(
+        noCache: true,
+        privateCache: true,
+      ),
+    );
+
     return context.withResponse(Response.ok(
       body: Body.fromString(widget.toString(), mimeType: mimeType),
+      headers: headers,
     ));
   }
 }
--- a/packages/serverpod/lib/src/web_server/web_server.dart
+++ b/packages/serverpod/lib/src/web_server/web_server.dart
@@ -153,7 +153,23 @@ class WebServer {
   ) async {
     // TODO: route.setHeaders(request.response.headers);
     try {
-      return await route.handleCall(session, context);
+      final result = await route.handleCall(session, context);
+
+      if (route is WidgetRoute && result is ResponseContext) {
+        final transformedResponse = result.response.copyWith(
+          headers: result.response.headers.transform(
+            (mh) => mh.cacheControl = CacheControlHeader(
+              maxAge: 0,
+              sMaxAge: 0,
+              noCache: true,
+              noStore: true,
+            ),
+          ),
+        );
+        return result.withResponse(transformedResponse);
+      }
+
+      return result;
     } catch (e, stackTrace) {
       final request = context.request;
       await _reportException(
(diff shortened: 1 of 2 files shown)


packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -153,7 +153,23 @@ class WebServer {
   ) async {
     // TODO: route.setHeaders(request.response.headers);
     try {
-      return await route.handleCall(session, context);
+      final result = await route.handleCall(session, context);
+
+      if (route is WidgetRoute && result is ResponseContext) {
```
SandPod: Suggestion: Move this header adaption into the `WidgetRoute` code.

This solution `taints` all `WidgetRoute` to use no cache headers. This both removes the flexibility from the developer to decide this on their own but it will also override any value that was explicitly specified.

By adding this as a resonable default for the `WidgetRoute` component, developers can chose a different caching strategy if needed.

SandPod: You should be able to achiev this by making an addition in the `WidgetRoute.handleCall` method. Either through a call to a different method or an optional constructor paramter.

packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -153,7 +153,23 @@ class WebServer {
...
+        final transformedResponse = result.response.copyWith(
+          headers: result.response.headers.transform(
+            (mh) => mh.cacheControl = CacheControlHeader(
```
SandPod: Question: To be safe in all environments, should we also add `must-revalidate, proxy-revalidate`?

According to this article here this might be needed for HTTP and old outdated proxies: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#shared_cache

SandPod: However, This is also noted on the page:


> max-age=0 means that the response is immediately stale, and must-revalidate means that it must not be reused without revalidation once it is stale — so, in combination, the semantics seem to be the same as no-cache.
> 
> However, that usage of max-age=0 is a remnant of the fact that many implementations prior to HTTP/1.1 were unable to handle the no-cache directive — and so to deal with that limitation, max-age=0 was used as a workaround.
> 
> But now that HTTP/1.1-conformant servers are widely deployed, there's no reason to ever use that max-age=0 and must-revalidate combination — you should instead just use no-cache.

SandPod: Also: Note this section: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#whats_lost_by_no-store

> You may think adding no-store would be the right way to opt-out of caching.

> However, it's not recommended to grant no-store liberally, because you lose many advantages that HTTP and browsers have, including the browser's back/forward cache.

> Therefore, to get the advantages of the full feature set of the web platform, prefer the use of no-cache in combination with private.

packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -153,7 +153,23 @@ class WebServer {
...
+              maxAge: 0,
+              sMaxAge: 0,
+              noCache: true,
+              noStore: true,
```
SandPod: Discussion: For personlized content, `private` should be used to avoid request collapse. But that is perhaps a different configured route that should be used? A `PersonalWidgetRoute` or something like that?

Source: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#request_collapse

SandPod: This seems to be the recommended default setting: `Cache-Control: no-cache, private`

Source: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#default_settings

SandPod: Here is also two recommended configurations: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#main_resources

packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -153,7 +153,23 @@ class WebServer {
...
     try {
-      return await route.handleCall(session, context);
+      final result = await route.handleCall(session, context);
+
+      if (route is WidgetRoute && result is ResponseContext) {
+        final transformedResponse = result.response.copyWith(
+          headers: result.response.headers.transform(
+            (mh) => mh.cacheControl = CacheControlHeader(
+              maxAge: 0,
+              sMaxAge: 0,
```
SandPod: Question: Is this required?

SandPod (changes requested):
Would like to have the implementation moved so that it allows developers and us to control this behavior without modifying the web server.

I would also like to get a better understanding of why these headers are configured in relation to the documentation that I provided. Please read it through and update or motivate the choices made here.

Tests looked good! 🌟

packages/serverpod/lib/src/web_server/web_server.dart
SandPod: Also, could we check what for example RubyOnRails or Django uses for their dynamically served content?

exaby73:
Rationale for changes to the values of the cache control header:

- MDN recommends `no-cache, private` as the default for dynamic content
- `no-store` should be avoided as it disables browser back/forward cache
- `maxAge: 0` and `sMaxAge: 0` are redundant with `noCache`
- Rails uses similar approach: `private, must-revalidate` by default

exaby73:
Add in https://github.com/serverpod/serverpod/pull/3903/commits/4d75cb12c030af13cfe5f0bf57f2d791b431aec7 @SandPod

PR #3905, opened 2025-08-20
fix: Password length validation in reset and change password

### LINKED ISSUE #3891: Password length validation missing on email authentication password reset/change
### Describe the bug

In the old authentication package, when users reset or change their password via email authentication, the minimum and maximum password length are not validated. This allows users to set passwords that are too short or excessively long, potentially compromising security.

### To reproduce

1. Use email authentication to reset or change a user's password.
2. Enter a password shorter than the intended minimum or longer than the intended maximum.
3. Observe that the password is accepted without validation.

### Expected behavior

Password resets and password changes via email authentication should enforce minimum and maximum password length validation, rejecting passwords that do not meet the criteria.

### Library version

2.9.1 and older.

### Platform information

All platforms where the old authentication package is used.

### Additional context

This bug could allow users to set insecure passwords and should be addressed to maintain proper security standards.

- [ ] I want to work on this

yashas-hm:
Fixes #3891 
Closes #3891 

Fixes password length validation when a reset password or change password request is sent via email auth provider.

## Pre-launch Checklist

### DIFF
--- a/tests/serverpod_test_server/test_integration/auth_module/email_auth_provider/password_length_validation_test.dart
+++ b/tests/serverpod_test_server/test_integration/auth_module/email_auth_provider/password_length_validation_test.dart
@@ -1,9 +1,35 @@
+import 'dart:async';
+import 'dart:math';
+
 import 'package:serverpod_auth_server/serverpod_auth_server.dart';
 import 'package:test/test.dart';
 
 import '../../test_tools/serverpod_test_tools.dart';
 
 void main() async {
+  const minPasswordLength = 8;
+  const maxPasswordLength = 128;
+
+  // Capture latest reset code here
+  late Completer<String> latestResetCode;
+
+  setUpAll(() async {
+    // Configure sending reset emails to capture the verification code
+    AuthConfig.set(
+      AuthConfig(
+        sendPasswordResetEmail: (sess, userInfo, validationCode) async {
+          latestResetCode.complete(validationCode);
+          return true;
+        },
+        // Some other tests set this to false; keep defaults suitable for test
+        extraSaltyHash: false,
+        sendValidationEmail: (sess, email, code) async => true,
+        minPasswordLength: minPasswordLength,
+        maxPasswordLength: maxPasswordLength,
+      ),
+    );
+  });
+
   withServerpod('Given an existing user ', (sessionBuilder, _) {
     final session = sessionBuilder.build();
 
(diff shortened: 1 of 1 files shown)


modules/serverpod_auth/serverpod_auth_server/lib/src/business/email_auth.dart
```diff
@@ -179,6 +179,18 @@ class Emails {
       return false;
     }
 
+    if (newPassword.length < AuthConfig.current.minPasswordLength ||
```
SandPod: Suggestion: Break this out into a method and reuse the logic in all three places.

SandPod (changes requested):
Implementation looks good, but it should be broken out and shared for all places.

Please also add tests that validate that this validation is used here: tests/serverpod_test_server/test_integration/auth_module/email_auth_provider

yashas-hm:
> Implementation looks good, but it should be broken out and shared for all places.
> 
> Please also add tests that validate that this validation is used here: tests/serverpod_test_server/test_integration/auth_module/email_auth_provider

Should the test just verify that the validate password length function or he whole email reset/change password with length validation?

SandPod:
It should validate that email reset and change password uses and respects length validation.

Password length validation is then validated through these public methods.

modules/serverpod_auth/serverpod_auth_server/lib/src/business/email_auth.dart
```diff
@@ -179,6 +207,10 @@ class Emails {
       return false;
     }
 
+    if (!isValidPasswordLength(password: newPassword, session: session)) {
```
SandPod: Suggestion: Make this a private method.

Don't need to bloat the interface with this logic.

tests/serverpod_test_server/test_integration/auth_module/email_auth_provider/password_length_validation_test.dart
```diff
@@ -0,0 +1,129 @@
+import 'package:serverpod_auth_server/serverpod_auth_server.dart';
+import 'package:test/test.dart';
+
+import '../../test_tools/serverpod_test_tools.dart';
+
+void main() async {
+  group('Given the password of a particular length', () {
```
SandPod: Suggestion: Remove these specific tests and only validate it through the public API.

The password validation method is an internal implementation detail, by not testing it directly but instead validating it works as expected through the public interface, we make it possible to refactor the internals of this logic without having to change our tests, thus not reducing confidence the code works as expected.

tests/serverpod_test_server/test_integration/auth_module/email_auth_provider/password_length_validation_test.dart
```diff
@@ -0,0 +1,129 @@
...
+
+    const userName = 'user-change';
+    const email = 'user.change@serverpod.dev';
+    const oldPassword = 'oldPassword1';
+
+    setUp(() async {
+      await Emails.createUser(session, userName, email, oldPassword);
+    });
+
+    test('then new password shorter than minimum is rejected', () async {
```
SandPod: Suggestion: Fix test so that the combined description is `Given an existing user when changing to a password that is shorter than minimum required password lenght then password change is rejected`

Follow the same pattern for the description of the other tests.

tests/serverpod_test_server/test_integration/auth_module/email_auth_provider/password_length_validation_test.dart
```diff
@@ -0,0 +1,129 @@
...
+      expect(success, isTrue);
+    });
+  });
+
+  withServerpod('Given a created user when resetting password',
+      (sessionBuilder, _) {
+    final session = sessionBuilder.build();
+
+    const userName = 'user-reset';
+    const email = 'user.reset@serverpod.dev';
```
SandPod: Suggestion: Combine set up with the previous test.

The `Given` part is the same here as for th previous test, we can utalize this and differentiateb based on the `When` part.

tests/serverpod_test_server/test_integration/auth_module/email_auth_provider/password_length_validation_test.dart
```diff
@@ -0,0 +1,129 @@
...
+      final success = await Emails.resetPassword(
+        session,
+        latestResetCode!,
+        '12345678', // min length
+      );
+
+      expect(success, isTrue);
+    });
+  });
+}
```
SandPod, line 213: Question: Should we add a test for creating an account as well here? Since this file now covers password validation scenarios.
Or is this better located in the create account test file?

yashas-hm: I can add a create account test just covering the password validation here. It would be suitable to put all tests that validate password length in the same file.

Issue #3887, opened 2025-08-14
Test case for password reset of non-migrated user


SandPod:
**Test**: After authentication migration, if a user’s first action is to request a password reset instead of logging in, the flow should work as expected.

**Expected behavior**

- Password reset completes successfully.
- User can log in with the new password immediately.
- All migration-dependent actions (e.g., clear `LegacyEmailPassword` entry) are correctly set, as if the user had logged in first.

nicowalter256:
I want to contribute. Is this fixed?

SandPod:
Hey @nicowalter256 👋 

That would be awesome!

This is not fixed yes, and I believe you can add the test here: tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart.

It could be good to split the test up into different files based on scenario.

nicowalter256:
thanks Let me try it out.

nicowalter256:
@SandPod could you check the PR https://github.com/serverpod/serverpod/pull/3913. Thanks

nicowalter256:
I didnt have a way to add you as a reviewer

SandPod:
Solved by https://github.com/serverpod/serverpod/pull/3913

PR #3913, opened 2025-08-25
test(auth): added comprehensive authentication integration tests

nicowalter256:
## Test Coverage
- Login with valid/invalid credentials  
- Password reset and change functionality
- Blocked user authentication handling
- Non-migrated user password reset scenarios
- Migrated user password reset scenarios 
- Legacy password fallback behavior
- Proper test isolation and cleanup procedures

## Issues Addressed
- #3887 - Test case for password reset of non-migrated user

The test specifically covers the scenario where a user who exists in the legacy system (non-migrated) attempts to reset their password, ensuring the system handles this correctly by either requiring migration first or providing appropriate error handling.

## Breaking changes
None - this is purely a test addition.

### DIFF
--- a/tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
+++ b/tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
@@ -2,7 +2,6 @@ import 'package:serverpod/serverpod.dart';
 import 'package:serverpod_auth_idp_server/core.dart';
 import 'package:serverpod_auth_idp_server/providers/email.dart';
 import 'package:test/test.dart';
-
 import '../util/test_tags.dart';
 import 'test_tools/serverpod_test_tools.dart';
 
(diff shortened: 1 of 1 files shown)


nicowalter256:

nicowalter256:
@SandPod could you have a look at this PR

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -2,7 +2,6 @@ import 'package:serverpod/serverpod.dart';
 import 'package:serverpod_auth_idp_server/core.dart';
 import 'package:serverpod_auth_idp_server/providers/email.dart';
 import 'package:test/test.dart';
-
```
marcelomendoncasoares: Worth keeping this space, as it separates different import groups (package vs local).

SandPod:
@nicowalter256 - Big thanks for this PR! @marcelomendoncasoares Will take on the review :)

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -407,6 +406,372 @@ void main() {
...
+            sendPasswordResetVerificationCode: (
+              final session, {
+              required final email,
+              required final passwordResetRequestId,
+              required final transaction,
+              required final verificationCode,
+            }) {
+              receivedVerificationCode = verificationCode;
+            },
+          );
```
marcelomendoncasoares: May be worth moving this block to the `setUp`. This way you can safely reduce the boilerplate code inside each test and make them clearer to future developers - same patter as the first two `withServerpod` groups of the test file. Commenting here, but valid for all new tests.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -407,6 +406,372 @@ void main() {
     testGroupTagsOverride: TestTags.concurrencyOneTestTags,
     rollbackDatabase: RollbackDatabase.disabled,
   );
+
+  withServerpod(
```
marcelomendoncasoares: As all tests are related to user migration, it is valid to move them to a separate file.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -407,6 +406,372 @@ void main() {
...
+          );
+
+          await endpoints.emailAccount.startPasswordReset(
+            sessionBuilder,
+            email: email,
+          );
+
+          expect(receivedVerificationCode, isNotNull);
+        },
+      );
```
marcelomendoncasoares: This test seems to be redundant with the last one from the former group, right? If so, may be worth removing to avoid duplicate validations

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -407,6 +406,372 @@ void main() {
...
+
+          await endpoints.emailAccount.finishPasswordReset(
+            sessionBuilder,
+            passwordResetRequestId: receivedPasswordResetRequestId,
+            verificationCode: receivedVerificationCode,
+            newPassword: newPassword,
+          );
+
+          // Reset config for login test
+          EmailAccounts.config = EmailAccountConfig();
```
marcelomendoncasoares: Question: Why is this reset needed here? The previous password reset configuration should not need to change for subsequent login attempts, right?

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -524,4 +889,6 @@ Future<void> _cleanUpDatabase(final Session session) async {
     session,
     where: (final _) => Constant.bool(true),
   );
+
+  // LegacyEmailPassword entries are cleaned up automatically via foreign key constraints
```
marcelomendoncasoares: This comment can be removed, as it is not tied to any code.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -407,6 +406,372 @@ void main() {
...
+            authSuccess.token,
+          );
+
+          expect(authInfo, isNotNull);
+          expect(authInfo!.authUserId, newAuthUserId);
+        },
+      );
+
+      test(
+        'when completing password reset, then legacy password entry remains as fallback.',
```
marcelomendoncasoares: Question(about the feature): Is this behavior something we really want? The legacy password remain as fallback **during** the reset makes sense, but should it remain valid after reset completed? It seems kind of unexpected to me (and a possible security flaw) that resetting the password could result in a previous password still valid, but I may be missing some edge case here. @SandPod, what is your take on this?

There is a `clearLegacyPassword` that is automatically called inside the `importLegacyPasswordIfNeeded`. Should it also be automatically called upon completing password reset?

marcelomendoncasoares (changes requested):
Great work, @nicowalter256! Only minor formatting comments and the separation to a different file. The tests highlighted one important feature discussion for us to define while proceeding with the other changes.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
nicowalter256: Got it 👍 Thanks for pointing that out. I’ll move the block into the setUp

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
nicowalter256: let me remove it

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
nicowalter256: Thanks for catching that - it was indeed redundant boilerplate code!

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_migration_test.dart
```diff
@@ -0,0 +1,333 @@
...
+            sendPasswordResetVerificationCode: (
+              final session, {
+              required final email,
+              required final passwordResetRequestId,
+              required final transaction,
+              required final verificationCode,
+            }) {
+              receivedVerificationCode = verificationCode;
+            },
+          );
```
marcelomendoncasoares: It is valid to also move this config (and the one in 91-101 lines) to the `setUp`, as you did on the other test.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_migration_test.dart
```diff
@@ -0,0 +1,333 @@
...
+          }) {
+            receivedPasswordResetRequestId = passwordResetRequestId;
+            receivedVerificationCode = verificationCode;
+          },
+        );
+      });
+
+      tearDown(() async {
+        EmailAccounts.config = EmailAccountConfig();
+
```
marcelomendoncasoares: Suggestion(nitpick): Remove this empty line to follow same standard as you have used in the other `tearDown`.

marcelomendoncasoares (changes requested):
Changes are in great shape! Just two more small comments and we are good to go!

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
SandPod: This does not seem like correct behavior. Once a password reset is complete, the old password should not work, and there is no need to preserve it as a fallback. You have uncovered a bug in the system.

Please patch this as part of this PR as well.

marcelomendoncasoares: @nicowalter256, can you address this on the PR, then? Update the test with the intended behavior (old password revoked after reset) and fix the code to make it pass.

tp: To be clear, the old/legacy password could not be used (in any of the new APIs at least, that password would / will still work via the old auth endpoints if they're not taken offline).

The test may make it seem like the password is still usable, because it calls into a internal/implementation method, but does not execute a real login.

> There is a clearLegacyPassword that is automatically called inside the importLegacyPasswordIfNeeded. Should it also be automatically called upon completing password reset?

Yes, calling `clearLegacyPassword` in the reset password flow is exactly what was missing. The idea of the "legacy tables" was that they would get their rows reduced one by one whenever a user is fully migrated, and resetting their password is one of those options where they would count as migrated :)

nicowalter256: @tp   I've added the `clearLegacyPassword` call in the password reset flow

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_accounts.dart
```diff
@@ -500,6 +501,13 @@ abstract final class EmailAccounts {
           transaction: transaction,
         );
 
+        // Clear legacy password after successful password reset for security
+        await AuthBackwardsCompatibility.clearLegacyPassword(
+          session,
+          emailAccountId: account.id!,
+          transaction: transaction,
+        );
+
```
marcelomendoncasoares: The problem with this approach is that it introduced a dependency of the migration package on the default package, which is not desired - new users should never have the migration package installed.

A possible approach to circumvent this is to expose an `onPasswordResetCallback` on the `EmailAccountConfig` - similar to the `sendRegistrationVerificationCode`. Then we can instruct on the docs that, when migrating, this parameter must be configured to call the `AuthBackwardsCompatibility.clearLegacyPassword`.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/pubspec.yaml
```diff
@@ -7,12 +7,13 @@ version: 3.0.0-alpha.1
...
 
 environment:
-  sdk: '>=3.5.0 <4.0.0'
+  sdk: ">=3.5.0 <4.0.0"
 
 dependencies:
   serverpod: 3.0.0-alpha.1
   serverpod_shared: 3.0.0-alpha.1
   serverpod_auth_core_server: 3.0.0-alpha.1
+  serverpod_auth_bridge_server: 3.0.0-alpha.1
```
marcelomendoncasoares: With the removal of the dependency, all changes to the `pubspec` files can be reverted.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_accounts.dart
nicowalter256: Thanks for the guidance. I’m working on it

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_account_config.dart
```diff
@@ -30,6 +30,14 @@ typedef SendPasswordResetVerificationCodeFunction = void Function(
   required Transaction transaction,
 });
 
+/// Function to be called after a password reset is successfully completed.
+typedef OnPasswordResetCompletedFunction = void Function(
+  Session session, {
+  required String email,
```
marcelomendoncasoares: Suggestion: Do we have an use case in mind for needing the `email` parameter here? If not, by removing the parameter, we make the callback signature fully compatible with the `AuthBackwardsCompatibility.clearLegacyPassword` function and it can be used like:

```dart
  EmailAccounts.config = EmailAccountConfig(
    // other parameters goes here
    onPasswordResetCompleted:
        AuthBackwardsCompatibility.clearLegacyPassword,
  );
```

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/pubspec.yaml
```diff
@@ -7,7 +7,7 @@ version: 3.0.0-alpha.1
 repository: https://github.com/serverpod/serverpod
 
 environment:
-  sdk: '>=3.5.0 <4.0.0'
```
marcelomendoncasoares: Please revert all changes to the `pubspec` files, as there are no more dependencies added. This will also resolve the conflicts with the `main` branch.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_migration_test.dart
```diff
@@ -0,0 +1,331 @@
...
+            required final emailAccountId,
+            required final transaction,
+          }) async {
+            await AuthBackwardsCompatibility.clearLegacyPassword(
+              session,
+              emailAccountId: emailAccountId,
+              transaction: transaction,
+            );
+          },
+        );
```
marcelomendoncasoares: If you remove the `email` parameter from the callback, reflect the direct set of `AuthBackwardsCompatibility.clearLegacyPassword` to the parameter without the closure.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_migration_test.dart
```diff
@@ -0,0 +1,331 @@
...
+          // Verify legacy password remains cleared after password importing login
+          expect(
+            await endpoints.emailAccountBackwardsCompatibilityTest
+                .checkLegacyPassword(
+              sessionBuilder,
+              email: email,
+              password: legacyPassword,
+            ),
+            isFalse,
+          );
```
marcelomendoncasoares, line 328: Question: If I'm understanding correctly, the use case we want to validate with this three checks is that, after clearing the legacy password, it does not get re-imported again, right? If so, it may be worth separating this use case to a different test to isolate it.

marcelomendoncasoares (changes requested):
Great progress! A few final comments in style and to understand a possible test separation.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_migration_test.dart
nicowalter256: You're absolutely right - the test was mixing concerns. I've refactored it to separate the validation into two focused tests

marcelomendoncasoares (changes requested):
Everything looking good now! Just fix the failing test and we are good to merge.

Issue #2071, opened 2024-03-26
serverpod create-migration exits with non zero exit code if no changes are detected


dkbast:
**Describe the bug**
To make sure, that developers don't forget to run code generation I usually run "any" code generator involved in a project in a pipeline step. Running serverpod create-migration exits with a non 0 exit code if no changes have been detected, breaking with the exit code convention.

It would also be worth considering to set a non zero exit code if anything has changed ( or add a flag, like "set-exit-if-changed" similar to dartfmt ).

**To Reproduce**
1. run `serverpod create-migration` twice in a row

**Expected behavior**
The cli exits with 0
(Optional: if something has changed exit with non zero)

**Screenshots**
```
serverpod create-migration
⠹ Creating migration... (0.1s)
No changes detected.
✗ Creating migration (0.1s)
make: *** [generate] Error 1
```

**Serverpod versions**
 -  _CLI Version:_ 1.2.5
 - _Library version:_ 1.2.5

**Platform information**
Macos arm

SandPod:
Good suggestion!

We should fix this!

When fixing this we should also introduce the following flags:
- `--empty` - Creates an empty migration without evaluating the state of the project (Advanced use case).
- `--check` - Returns with exit code 0 if no changes have been detected. (A bit less verbose than `--set-exit-if-changed` and used in both Djangos migration tool and Prettier).

SandPod:
@nicowalter256 - Could you take a look at this one?

There is already an AI-suggested implementation, but I'm not sure that it is completely valid.

Check out [#3952](https://github.com/serverpod/serverpod/pull/3952) for inspiration.

nicowalter256:
cool can take on this

nicowalter256:
check this out https://github.com/serverpod/serverpod/pull/3978 @SandPod

SandPod:
Exit code has now been changed with this PR: https://github.com/serverpod/serverpod/pull/4388

Will create new issues to cover the  `--empy` and `--check` flags

PR #3915, opened 2025-08-25
fix: Misc auth fixes related to shared config.

SandPod:
This PR adds some minor fixes required to use a shared authentication config for the token providers.

Changes in the PR:
- Adds `method` field to refresh tokens to keep track of what provider created it and updates the interfaces accordingly.
- Makes scopes optional when creating refresh tokens and uses the auth users scopes if none are provided.
- Respect blocked user when creating refresh token.
- Use a single call to clock for the same refresh tokens fields.
- Recreate all migration scripts and tag them with 3.0.0 for everything related with this release.
  - Migration for `serverpod_test_server` does not include the `v3-0-0` tag because tags are used for the migration tests.

Related to: #3856 

**Additional fixes**
A bit jumbled up but this PR also includes a utility script to remove all the latest migrations or to remove all migrations from a specific tag.

This was used to recreate all migrations with the `v3-0-0` tag.

Also removes the `-e` from the `create_migrations` script since we report no migration created with exit code 1, tracked here: https://github.com/serverpod/serverpod/issues/2071

## Pre-launch Checklist

## Breaking changes

No, since the force migration change is added in an unreleased module.


SandPod:
I have manually validated that the change of the migrations will not negatively impact users of the alpha release.
They simply need to create a new migration for their own project.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/models/refresh_token.spy.yaml
```diff
@@ -23,6 +23,11 @@ fields:
   ### This is only stored as a serialized String in the database due to schema limitations.
   extraClaims: String?
 
+  ### The method through which this token was created.
+  ###
+  ### This can be either an email or social login, a personal access token, service account etc.
+  method: String
```
marcelomendoncasoares, line 29: Question: is it worth adding this `method` in the `AuthSucess` as well? Since provider configs can generate either a JWT or SAS tokens.

SandPod: I see this more as a server-side property. 

The main use case I see for this is traceability on the Server that created a token. There is also an opportunity to display these to the user so that they can revoke specific tokens. But that will be up to the implementer to decide.

Therefore, I don't think this should be part of the `AuthSuccess` response since this should just reflect your credentials used to authenticate towards the server.

Issue #3674, opened 2025-06-18
JWT Session manager


SandPod:
Session manager for the new Authentication Module that can interact with JWT sessions.

tp:
When sending the JWT access tokens to the server, we probably run into the issue I now encountered when attempting to change the session key format for `serverpod_auth_session`: Since the client always uses a `Basic` authorization header it requires the session key to contain a `:` to confirm to the `user:password` schema on the server (additional `:` are fine, as they are just part of the password). (See https://github.com/serverpod/serverpod/pull/3834 for more info.)

Since we can not control the access token format for JWT, we might want add support for `Bearer` token headers instead. In those scenarios the client should send the token verbatim then (thus requiring it to be "header safe"), so it might be a good idea to have `SessionManager`s opt into it.

If the above idea gets added for JWT, it might prove useful to also employ that approach for the DB-backed "session keys" and switch to a "single base64" key format as attempted here: https://github.com/serverpod/serverpod/commit/dbb836c35b1479c1f92f621a612e2c1e9bc9c2c8

PR #3927, opened 2025-08-26
feat: Introduce the new `authKeyProvider` and a new client session manager experience

marcelomendoncasoares:
This PR is part of the work to support JWT session manager (#3674) and introduce a cleaner experience for the user when interacting with our authentication package.

## Features introduced

- The `SessionManager` on the client is now called `ClientAuthSessionManager` and no longer requires the `caller` parameter. This makes the user experience smoother, since the `caller` is for internal usage on the session manager. The new pattern reuses the same as with the `FlutterConnectivityMonitor`:

  ```dart
  final client = Client('http://localhost:8080/')
        ..connectivityMonitor = FlutterConnectivityMonitor()  // follows same pattern as this
        ..authSessionManager = ClientAuthSessionManager(storage: storage);  // storage is optional
  
  await client.auth.initialize();
  ```

- The `authenticationKeyManager` is no longer used on the new auth. An `authKeyProvider` is expected instead, which only implements the `authHeaderValue` getter. To allow native support for JWT session managers, there is an extension of the `ClientAuthKeyProvider` with a `refreshAuthKey` method.

- If the `authKeyProvider` implements the `refrehsAuthKey`, when encountering a 401 (unauthorized) error for the first time, the the `ServerpodClientShared` (base class for the `Client`) will call it and retry the original request. It won't retry a second time, as it will not likely be an expiration problem. This covers edge cases where eagerly refreshing meets tight thresholds or with short expiration timeouts and network delays.

- To persistently store the auth key, the `ClientAuthSessionManager` requires a storage. As default storage, we previously used the traditional `shared_preferences` package, which was not an ideal solution to store secrets. Now, the default storage was replaced by the most suitable `flutter_secure_storage` package.

## Fixes

- The new auth package fixes an issue of the previous where using different clients with separate authentication managers would lead to one overriding the other due to the singleton pattern. Each client now has its own session manager.

## Tests considerations

The `SecureClientAuthInfoStorage`  class is intentionally left untested because a next PR will be opened replacing the `shared_preferences` package by the `flutter_secure_storage` package, and then introduce proper tests.

## Next steps

- Implement the automatic lock mechanism inside the `RefresherClientAuthKeyProvider` to avoid concurrent refresh attempts.
- Implement the status endpoints and the respective endpoint calls inside the `ClientAuthSessionManager`.
- Allow configuration of `ClientAuthSessionManager` to use JWT when managing user sessions.

## Pre-launch Checklist

## Breaking changes

For users of the alpha package, the client interface changes considerably. However, the new session manager was not fully implemented yet.


packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -449,13 +449,11 @@ abstract class ServerpodClientShared extends EndpointCaller {
...
   /// when sent to the server.
-  Future<void> updateStreamingConnectionAuthenticationKey(
-    String? authKey,
-  ) async {
+  Future<void> updateStreamingConnectionAuthenticationKey() async {
     if (streamingConnectionStatus == StreamingConnectionStatus.disconnected) {
       return;
     }
-    var authValue = await authenticationKeyManager?.toHeaderValue(authKey);
+    var authValue = await authenticationKeyManager?.getHeaderValue();
```
SandPod: Question: Can you remind me of how this works? I thought we removed the header encoding from the old streaming endpoints?

Will this work with backwards compatabaility?

packages/serverpod_client/lib/src/client_method_stream_manager.dart
```diff
@@ -79,8 +79,9 @@ final class ClientMethodStreamManager {
   /// Throws a [ConnectionAttemptTimedOutException] if the connection attempt
   /// takes too long.
   Future<void> openMethodStream(
-    MethodStreamConnectionDetails connectionDetails,
-  ) async {
+    MethodStreamConnectionDetails connectionDetails, [
+    bool shouldRetryOnAuthFailed = true,
```
SandPod: Suggestion: Move this so that it is configured on the Client.

What would be the use case for configuring it per call?
(NIT) Use named parameters so that we can grow this list without thinking about order.

packages/serverpod_client/lib/src/client_method_stream_manager.dart
```diff
@@ -115,6 +116,14 @@ final class ClientMethodStreamManager {
     if (openResponse != OpenMethodStreamResponseType.success) {
       _inboundStreams.remove(inboundStreamKey);
       _tryCloseConnection();
+
+      final authKeyProvider = connectionDetails.authKeyProvider;
+      if (openResponse == OpenMethodStreamResponseType.authenticationFailed &&
+          authKeyProvider is RefresherClientAuthKeyProvider &&
+          shouldRetryOnAuthFailed &&
```
SandPod: Suggestion(Style): Break these three out into a separate variable and name it `shouldRefreshAuth`

packages/serverpod_client/lib/src/auth_key_provider.dart
```diff
@@ -0,0 +1,36 @@
...
+    // TODO: Use Completer to implement a lock under the future of the getter to
+    // avoid multiple concurrent refresh attempts.
+    return await notLockedAuthHeaderValue;
+  }
+
+  /// Refreshes the authentication key. If the refresh is successful, should
+  /// return true to retry requests that failed due to authentication errors.
+  /// Be aware that the refresh endpoint must be annotated with @unauthenticated
+  /// to avoid a deadlock on the [authHeaderValue] getter on refresh call.
+  Future<bool> refreshAuthKey();
```
SandPod, line 35: Suggestion: This method also needs to be synchronized for the client to avoid multiple concurrent calls to `refreshAuthKey`.

Otherwise, there is a risk of a data race where an older request overrides the new auth key.

SandPod: I would like to see a test that validates that we handle this situation.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -205,12 +206,17 @@ abstract class ServerpodClientShared extends EndpointCaller {
...
+  /// unauthenticated.
+  ClientAuthKeyProvider? authKeyProvider;
+
   /// Creates a new ServerpodClientShared.
   ServerpodClientShared(
     this.host,
     this.serializationManager, {
     dynamic securityContext,
-    required this.authenticationKeyManager,
+    this.authenticationKeyManager,
```
SandPod, line 219: Question: When can we remove this and introduce the breaking change?

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -493,6 +504,17 @@ abstract class ServerpodClientShared extends EndpointCaller {
     } catch (e, s) {
       onFailedCall?.call(callContext, e, s);
 
+      // A first failure here with 401 can be due to an access token expiration.
```
SandPod: Question: You mention this should be tried only once, but the current configuration will try it once for every request from the server with a 401 response. Shouldn't we ensure that we only retry once for a 401?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
```diff
@@ -1,158 +1,174 @@
...
+
+import 'package:flutter/material.dart';
+import 'package:serverpod_auth_core_client/serverpod_auth_core_client.dart';
+
+import '/src/storage/base.dart';
+import '/src/storage/secure.dart';
+
+/// As applications can have more than one Serverpod client, they also need to
+/// have separate auth session managers for each.
+final _authSessionManagersCache =
```
SandPod: Question: Why do we need a global auth session manager cache?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
```diff
@@ -1,158 +1,174 @@
...
     }
-    _didInit = true;
-
-    return _init();
-  }
 
-  Future<void> _init() async {
-    final sessionKey = await _secureStorage.get(
-      SessionManagerStorageKeys.sessionKey.key,
+    throw StateError(
```
SandPod, line 41: Question: Didn't we move away from this consolidation pattern?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
```diff
@@ -1,158 +1,174 @@
...
+  var _initialized = false;
 
-    if (sessionKey == null) {
-      _authInfo.value = null;
+  AuthSuccess? _signedInUser;
 
-      return;
+  /// Returns information about the signed in user, if any.
+  AuthSuccess? get signedInUser {
+    if (!_initialized) {
```
SandPod: Question: Can we follow "Resource acquisition is initialization" here and always initialize the session manager when it is created?

packages/serverpod_client/lib/src/auth_key_provider.dart
```diff
@@ -0,0 +1,36 @@
...
+  Future<String?> get authHeaderValue;
+}
+
+/// Provides the authentication key for the client, with a method to refresh it.
+abstract class RefresherClientAuthKeyProvider implements ClientAuthKeyProvider {
+  /// Authentication header value getter that must be implemented. This getter
+  /// must get the header value directly from the source, without any locking.
+  /// The [authHeaderValue] getter will handle refreshing the key if needed,
+  /// with a lock to avoid concurrent refresh attempts.
+  Future<String?> get notLockedAuthHeaderValue;
```
SandPod, line 17:
```suggestion
  Future<String?> get notSyncronizedAuthHeaderValue;
```

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/base.dart
```diff
@@ -0,0 +1,83 @@
...
+    _cachedData = data;
+    _cached = true;
+    return _cachedData;
+  }
+}
+
+/// Implements the [ClientAuthInfoStorage] for key-value based storages.
+abstract class KeyValueClientAuthInfoStorage
+    extends CachedClientAuthInfoStorage {
+  /// Set the [AuthSuccess] as JSON-encoded string value in the storage.
```
SandPod: Question: Why doesn't this interface take in an `AuthSuccess`? Same for the next method.

packages/serverpod_client/lib/src/auth_key_provider.dart
marcelomendoncasoares: I preferred to address this in a separate PR to avoid this one growing in too many directions. So I left the `TODO` here and added the lock mechanism with tests on this commit for a next PR: https://github.com/serverpod/serverpod/commit/4dfa63bcd93e908543a85132530015141c9de369.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
marcelomendoncasoares: I have a partial commit with the deprecation of this parameter in a separate branch, but I thought better to only apply it once we have the new authentication fully implemented - with the new endpoints and providers working together.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
marcelomendoncasoares: Not anymore! Thanks for catching this. It was necessary before moving the `ClientAuthKeyProvider` implementation to the `ClientAuthSessionManager`, as we decided during the development. I'll clean this up.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
marcelomendoncasoares: As we investigated, this header wrapping happens inside the `updateStreamingConnectionAuthenticationKey`, which calls the command to update the auth. So wrapping the header does no harm because the command on the server will unwrap the auth header value:

https://github.com/serverpod/serverpod/blob/5afeafc0753447785ee499541d56c0e4bd6cfa85/packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart#L69-L72

packages/serverpod_client/lib/src/client_method_stream_manager.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/3927/commits/670c7dcc41e3d8925ad6797d5515085f36259eaa. The parameter exists no more, so no need to change to named instead of positional.

packages/serverpod_client/lib/src/client_method_stream_manager.dart
marcelomendoncasoares: Also done at https://github.com/serverpod/serverpod/pull/3927/commits/670c7dcc41e3d8925ad6797d5515085f36259eaa.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
marcelomendoncasoares: As we discussed, ensuring that new requests never retry after the token is deemed invalid will be a responsibility of the `RefresherClientAuthKeyProvider` implementation. It must return `false` from the `refreshAuthKey` method for all further requests until a new valid token is received.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/3927/commits/3d4027950e25a0ab18617781b1bf35196dfdd932.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
marcelomendoncasoares: This change was refactored on further commits, so I'll just resolve this one.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
marcelomendoncasoares: Unfortunately we can't due to the initialization being async and the value notifier - as was later refactored - being a sync call.

packages/serverpod_client/lib/src/auth_key_provider.dart
marcelomendoncasoares: We discussed the `notMutexLockedAuthHeaderValue` alternative as well, but we shall move this discussion to the next PR that will address the implementation of the lock mechanism.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/base.dart
marcelomendoncasoares: I'll refactor the classes to hide the storage and serialization implementation details from the user, leaving only the `get` and `set` methods from the base class exposed.

marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/3927/commits/b655c0ddd80cb92dd13c1cdbcacdc16dd20d0528.

As a major refactor was necessary, I squashed the other commit with the replacement of the `shared_preferences` by the `flutter_secure_storage` package to avoid wasting work on a dying piece of code. So one of the `TODO`s that would be addressed in a separate PR was already addressed. Also updated the PR description to include this change.

packages/serverpod_client/lib/src/client_method_stream_manager.dart
```diff
@@ -80,6 +80,29 @@ final class ClientMethodStreamManager {
...
+  ) async {
+    try {
+      return await _openMethodStream(connectionDetails);
+    } on OpenMethodStreamException catch (e) {
+      final authKeyProvider = connectionDetails.authKeyProvider;
+
+      // A first failure here with 401 can be due to an access token expiration.
+      // We will retry only once in such case, and only if the `authKeyProvider`
+      // exposes a `refreshAuthKey` method (like JWT).
+      final shouldRefreshAuth =
```
SandPod, line 92: Suggestion(NIT): Could use early returns here instead, where we check each condition at a time to not make the logic to complicated?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/cached.dart
```diff
@@ -0,0 +1,47 @@
...
+
+  /// Control whether the value can be recovered from the internal cache on
+  /// [get] method. Will be false only before the first method call.
+  var _cached = false;
+
+  AuthSuccess? _cachedData;
+
+  /// Returns the cached [AuthSuccess] instance. Will raise an error if trying to
+  /// access this attributed before any call to [get] or [set].
+  AuthSuccess? get cachedAuthInfo {
```
SandPod: Suggestion: Remove this method.

I only see it used in tests, wouldn't it be better to only expose `set` and `get` and then validate the behavior by controlling the `delegate` storage?
That way we actually test the external interface.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/key_value.dart
```diff
@@ -0,0 +1,41 @@
...
+  /// The key-value storage delegate to perform the actual storage operations.
+  final KeyValueStorage keyValueStorage;
+
+  /// Creates a new [KeyValueClientAuthInfoStorage].
+  KeyValueClientAuthInfoStorage({
+    required this.keyValueStorage,
+
+    /// Override the default key name to store the auth info on the storage.
+    String? authInfoStorageKey,
+  }) : _storageKey = authInfoStorageKey ?? 'serverpod_userinfo_key';
```
SandPod: Suggestion: this should probably be `serverpod_auth_success_key` right?

And the variable should be `authSuccessStorageKey`.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/key_value.dart
```diff
@@ -0,0 +1,41 @@
...
+
+  /// The key-value storage delegate to perform the actual storage operations.
+  final KeyValueStorage keyValueStorage;
+
+  /// Creates a new [KeyValueClientAuthInfoStorage].
+  KeyValueClientAuthInfoStorage({
+    required this.keyValueStorage,
+
+    /// Override the default key name to store the auth info on the storage.
+    String? authInfoStorageKey,
```
SandPod: Suggestion: Could this be `String authInfoStorageKey = 'value'` instead? Since the assigned value is const.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/secure.dart
```diff
@@ -0,0 +1,61 @@
...
+
+import '/src/storage/base.dart';
+import '/src/storage/cached.dart';
+import '/src/storage/key_value.dart';
+
+/// Exposes the [FlutterSecureStorage] class for convenience.
+export 'package:flutter_secure_storage/flutter_secure_storage.dart';
+
+/// A [ClientAuthInfoStorage] based on [FlutterSecureStorage].
+class SecureClientAuthInfoStorage extends CachedClientAuthInfoStorage {
```
SandPod: Question: All the classes in storage still confuse me, It all seems a bit complex. Do we really need all of them?

Who would find and use the different levels of abstractions?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/secure.dart
```diff
@@ -0,0 +1,61 @@
...
+  }) : super(
+          delegate: KeyValueClientAuthInfoStorage(
+            keyValueStorage: FlutterSecureKeyValueStorage(secureStorage),
+            authInfoStorageKey: authInfoStorageKey,
+          ),
+        );
+}
+
+/// A [KeyValueStorage] wrapper for [FlutterSecureStorage].
+class FlutterSecureKeyValueStorage implements KeyValueStorage {
```
SandPod: Question: Shouldn't this implementation be moved to the `key_value.dart` file as a concrete implementation there?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/base.dart
```diff
@@ -0,0 +1,15 @@
+import 'package:serverpod_auth_core_client/serverpod_auth_core_client.dart';
+
+export 'package:serverpod_auth_core_client/serverpod_auth_core_client.dart'
+    show AuthSuccess;
+
+/// Exposes a method for the session manager to update the [AuthSuccess].
+abstract interface class ClientAuthInfoStorage {
```
SandPod: Suggestion: Rename file to reflect the actual implemented interface, same for the other files in this directory. This will be hard to find when searching for the file.

`base` -> `client_auth_info_storage`
`cached` -> `cached_client_auth_info_storage` (should perhaps be implemented in this file directly)
etc...

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/key_value.dart
SandPod: This should also be reflected in the file names of all these files. Since we are only managing `AuthSuccess` objects.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
```diff
@@ -1,158 +1,154 @@
...
-    await _storage.set(SessionManagerStorageKeys.scopeNames.key, null);
+  /// Updates the signed in user on the storage and for open connections.
+  Future<void> updateSignedInUser(AuthSuccess? authInfo) async {
+    await storage.set(authInfo);
+    _authInfo.value = authInfo;
+    await caller.client.updateStreamingConnectionAuthenticationKey();
+  }
 
-    await _init();
+  // MAYBE: Is this true/false return enough? A connection error could return
```
SandPod, line 81: Should perhaps reflect the reason why the refresh failed. But can we safely provider an enum for this that we know won't expand? I guess so since this is inside of our core module.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
```diff
@@ -1,158 +1,154 @@
...
+    Caller? caller,
+
+    /// The secure storage to keep user authentication info. If missing, the
+    /// session manager will create a [SecureClientAuthInfoStorage].
+    ClientAuthInfoStorage? storage,
+  })  : _caller = caller,
+        storage = storage ?? SecureClientAuthInfoStorage();
+
+  /// Sets the caller from the client's module lookup.
+  void setCallerFromClient(ServerpodClientShared client) {
```
SandPod: Question: Why is this needed? Could this take in the caller directly instead? I think this will be needed in order to find the specific caller.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
```diff
@@ -1,158 +1,154 @@
...
+    if (currentProvider == null) {
+      throw StateError(
+        'To access the auth instance, first instantiate the session manager '
+        'and set it as the "authKeyProvider" on the client.',
+      );
+    }
+    if (currentProvider is! ClientAuthSessionManager) {
+      throw StateError(
+        'The "authKeyProvider" is set to an unsupported type. '
+        'Expected "ClientAuthSessionManager", got "$currentProvider".',
```
SandPod:
```suggestion
        'Expected "ClientAuthSessionManager", got "${currentProvider.runtimeType()}".',
```

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
```diff
@@ -0,0 +1,160 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:serverpod_auth_core_flutter/serverpod_auth_core_flutter.dart';
+
+import '../test_utils/storage_delegate.dart';
+
+void main() {
+  group('Given a CachedClientAuthInfoStorage implementation', () {
```
SandPod: Suggestion: Change to `Given unitialized CachedClientAuthInfoStorage`, `when accessing cachedAuthInfo then it throws StateError`

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
```diff
@@ -0,0 +1,160 @@
...
+      });
+
+      test('then it stores the data and caches it', () async {
+        final cached = storage.cachedAuthInfo!;
+
+        expect(cached.toString(), _authSuccess.toString());
+        expect(storage.delegate.lastSetValue, isNotNull);
+      });
+
+      test('then subsequent calls to cachedAuthInfo return the same data',
```
SandPod: Suggestion: I have already commented this somewhere else, but this should be calls to the `get` method to validate that this works as expected, otherwise we are bypassing public interface and are testing implementation details.

I think this method should be removed completely and all tests should be reformated to test the promised public interface.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
```diff
@@ -0,0 +1,160 @@
...
+      });
+
+      test('then it clears the data and caches null', () async {
+        expect(storage.cachedAuthInfo, isNull);
+        expect(storage.delegate.lastSetValue, isNull);
+      });
+    });
+
+    group('when calling get for the first time', () {
+      test('with data in storage then it retrieves from storage', () async {
```
SandPod: Suggestion: This is the Given part of the test. `Given data in storage when calling get for the first time then data is retrieved from storage`.

This could even be shortened to: `Given data in storage when calling get then data is retrieved`.
First time isn't really that important.

SandPod: Same for the other `with` descriptions in this group.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
```diff
@@ -0,0 +1,160 @@
...
+
+      test('with empty storage then it returns null', () async {
+        final result = await storage.get();
+
+        expect(result, isNull);
+        expect(storage.cachedAuthInfo, isNull);
+      });
+    });
+
+    group('when calling get after data is cached', () {
```
SandPod: Suggestion: Reformat to `Given data in storage when calling get multiple times then data is only retrieved from storage once`.

Then we add a helper method to our test storage that returns the `getCallCount` or something like that.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
```diff
@@ -0,0 +1,160 @@
...
+
+        expect(result, isNotNull);
+        expect(result!.toString(), _authSuccess.toString());
+
+        final cached = storage.cachedAuthInfo!;
+        expect(cached.toString(), _authSuccess.toString());
+      });
+    });
+
+    group('when storage operations throw exceptions', () {
```
SandPod: Suggestion: This is also the `Given` part, the action or `when` is calling `set`.

`Given storage that throws exception on set when calling set then exception is propagated`.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
```diff
@@ -0,0 +1,160 @@
...
+      test('then set propagates the exception', () async {
+        storage.delegate.shouldThrowOnSet = true;
```
SandPod: Suggestion: Restructure this to a `setOverride`, this allows us to define the expected exception and behavior in the test.

That way, readers of this test have all the details required to understand what is going on without having to understand the implementation details of the delegate. And we also do not need to refactor all the tests if we where to change the error message of the exception or exception type.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
```diff
@@ -0,0 +1,160 @@
...
+          throwsA(isA<Exception>().having(
+            (e) => e.toString(),
+            'message',
+            contains('Storage error on get'),
+          )),
+        );
+      });
+    });
+
+    test('when updating data then new cached data replaces old cached data',
```
SandPod: Suggestion: This should test the `get` methods, that is where this is important.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/key_value_client_auth_info_storage_test.dart
```diff
@@ -0,0 +1,152 @@
...
+      storage = TestKeyValueAuthInfoStorage.create();
+    });
+
+    group('when calling set with AuthSuccess data', () {
+      setUp(() async {
+        await storage.set(_authSuccess);
+      });
+
+      test('then it stores the data using the default key', () async {
+        const defaultKey = 'serverpod_userinfo_key';
```
SandPod: Comment: It is actually quite importnat that this is defined as a separate variable as opposed to the one used in the production code. This helps us detect if we would accidentally change that key.
Alternatively we could expose the key and have a specific test that validates that it does not change.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/key_value_client_auth_info_storage_test.dart
```diff
@@ -0,0 +1,152 @@
...
+        expect(storage.delegate.values[defaultKey], _authSuccess.toString());
+      });
+
+      test('then it encodes and stores the data as JSON string', () async {
+        final value = storage.delegate.lastSetValue;
+        expect(value, isA<String>());
+        expect(value, _authSuccess.toString());
+      });
+
+      test('then decoding the stored value returns the original object',
```
SandPod: Question: Isn't decoding the when action of this test? `Given stored value when decoding stored value then original object is returned`?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/key_value_client_auth_info_storage_test.dart
```diff
@@ -0,0 +1,152 @@
...
+    });
+
+    test('when calling set with null then it stores null value', () async {
+      await storage.set(null);
+
+      expect(storage.delegate.lastSetValue, isNull);
+    });
+
+    test(
+        'when calling get with stored data then it decodes and returns the AuthSuccess object',
```
SandPod: Suggestion: This is a different Given group.

`Given stored data when calling get then returns the AuthSuccess object` or something in that direction.

SandPod: Same for the other scenarios that describe the setup in the when section (`with` statements that are not parameters are usually a red flags)

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
SandPod: Same for the `get` throw test.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/secure_client_auth_info_storage_test.dart
```diff
@@ -0,0 +1,130 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:serverpod_auth_core_flutter/serverpod_auth_core_flutter.dart';
+
+void main() {
```
SandPod, line 4: Suggestion: Apply the same fixes to this file as we did for `cached_client_auth_info_storage_test.dart` and `key_value_client_auth_info_storage_test.dart`.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/client_auth_key_provider_test.dart
```diff
@@ -0,0 +1,61 @@
...
+
+  test(
+    'Given a client with authSessionManager set, '
+    'when updateSignedInUser is called, '
+    'then the client authKeyProvider provides the correct auth header.',
+    () async {
+      final client = Client('http://localhost:8080/')
+        ..authSessionManager = ClientAuthSessionManager(storage: TestStorage());
+
+      expect(await client.authKeyProvider?.authHeaderValue, isNull);
```
SandPod: Suggestion: Remove this expect, this is explicitly validated in the test before.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/client_auth_key_provider_test.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:serverpod_auth_core_flutter/serverpod_auth_core_flutter.dart';
+import 'package:serverpod_new_auth_test_client/serverpod_new_auth_test_client.dart';
+
+import 'utils/test_storage.dart';
+
+void main() {
+  test(
```
SandPod, line 8: Suggestion: Complement this test with a test that validates it is accessible when set.

The following tests assumes this is true but there is no validation on it.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/client_auth_key_provider_test.dart
```diff
@@ -0,0 +1,61 @@
...
+    },
+  );
+
+  test(
+    'Given a client with authSessionManager set, '
+    'when updateSignedInUser is called, '
+    'then the client authKeyProvider provides the correct auth header.',
+    () async {
+      final client = Client('http://localhost:8080/')
+        ..authSessionManager = ClientAuthSessionManager(storage: TestStorage());
```
SandPod, line 25: Suggestion: Provide a simple implementation of the `authKeyProvider` in this file and use it in the tests.

That way we can be explicit with the behavior. For example what header we are expecting now depends on an implementation detail outside of this test file.

SandPod: If we are not testing our implementation of the ClientAuthSessionManager, but I think that should be a different test?
Perhaps these tests could be moved to unit tests in the Serverpod client package since these should be preserved even if we remove the new auth implementation.

packages/serverpod_client/test/integration/call_server_endpoint_retry_test.dart
```diff
@@ -0,0 +1,200 @@
...
+        {'arg': 'value'},
+      );
+
+      expect(result, 'success');
+      expect(requestCount, 1);
+      expect(authKeyProvider.refreshCallCount, 0);
+    });
+
+    test(
+        'when first call fails with 401 and refresh succeeds '
```
SandPod:
```suggestion
        'when first call fails with 401 but refresh succeeds '
```

packages/serverpod_client/test/integration/call_server_endpoint_retry_test.dart
```diff
@@ -0,0 +1,200 @@
...
+
+      expect(result, 'success');
+      expect(requestCount, 2);
+      expect(authKeyProvider.refreshCallCount, 1);
+      expect(receivedAuthHeaders[0], contains('initial-token'));
+      expect(receivedAuthHeaders[1], contains('refreshed-token-1'));
+    });
+
+    test(
+        'when first call fails with 401 but refresh fails '
```
SandPod:
```suggestion
        'when first call fails with 401 and refresh fails '
```

packages/serverpod_client/test/integration/call_server_endpoint_retry_test.dart
```diff
@@ -0,0 +1,200 @@
...
+        client.callServerEndpoint<String>('test', 'method', {'arg': 'value'}),
+        throwsA(isA<ServerpodClientUnauthorized>()),
+      );
+
+      expect(requestCount, 1);
+      expect(authKeyProvider.refreshCallCount, 1);
+    });
+
+    test(
+        'when first and second calls fails with 401 '
```
SandPod:
```suggestion
        'when first call and refresh call fails with 401 '
```

packages/serverpod_client/test/integration/call_server_endpoint_retry_test.dart
```diff
@@ -0,0 +1,200 @@
...
+      expect(authKeyProvider.refreshCallCount, 0);
+    });
+
+    test(
+        'when exception is not ServerpodClientUnauthorized '
+        'then no retry is attempted.', () async {
+      closeServer = await TestHttpServer.startServer(
+        httpRequestHandler: (request) async {
+          requestCount++;
+          throw Exception('Network error');
```
SandPod: Question: How does this result in an exception in the `TestServerpodClient`, This should result in a 500 status response as far as I understand. 
Do we not handle such responses in the client?

packages/serverpod_client/test/integration/open_method_stream_retry_test.dart
```diff
@@ -0,0 +1,243 @@
...
+      streamManager = ClientMethodStreamManager(
+        connectionTimeout: const Duration(seconds: 5),
+        webSocketHost: webSocketHost,
+        serializationManager: TestSerializationManager(),
+      );
+    });
+
+    tearDown(() async => await closeServer());
+
+    test('when first call succeeds then no retry is attempted.', () async {
```
SandPod: Suggestion: Use connection secceeds or fails.

Since we are connecting streams. This phrasing should probably be updated for the complete file.
Should probably have some more information in the descriptions here as well so that it is clear we are attempting to authenticate through streams if this fails in ci.
The current descriptions do not really communicate this.

SandPod (commented):
Nice work! Took a more detailed look at the tests now.
There is some restructuring required there, but I believe I only had minor comments regarding implementation.

packages/serverpod_client/lib/src/client_method_stream_manager.dart
marcelomendoncasoares: I started applying this change, but it actually started looking worse, because we would end up with several `rethrow` calls and the `is RefresherClientAuthKeyProvider` can not be decoupled from the `refreshAuthKey` call.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/secure.dart
marcelomendoncasoares: My intention with the separation was exactly because of their complexity, to allow a good experience for users that need some custom alternative and maintain good testability.

While I was doing this last refactor, I started merging together the cached and key value, but it seemed to get worse. As each layer works on different aspects of the data, it also seemed as loosing single responsibility principle. Here are the separate use cases for each class:

- The `CachedClientAuthInfoStorage` only implements a `ClientAuthInfoStorage` with a cache layer, which is needed to offer a good performance (otherwise the storage will be hit everytime). It does not care about how the data is stored, which could be done on a key-value storage, a file or a database.
- The `KeyValueClientAuthInfoStorage` implements the most common storage interface, with the serialization and deserialization of the `AuthSuccess` already taken care of. But it does not enforce which key-value storage to use, as the user can have its preference. Two very common are the `SharedPreferences` and the `FlutterSecureStorage`.
- The `SecureClientAuthInfoStorage` is our default implementation, that combines the cached interface with a key-value storage implemented with the `FlutterSecureStorage` underneath. It is the most recommended alternative for storing credentials and also an implementation reference for users that prefer to use a `SharedPreferences`.

With this separation, an user could directly use the `KeyValueClientAuthInfoStorage` with an in-memory storage and pass it directly to the `ClientAuthSessionManager` - with no need for a redundant cache layer. Or combine the cache layer with an actual storage like we are doing.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/secure.dart
marcelomendoncasoares: I preferred to leave it separate since its implementation is only there to be used by our `SecureClientAuthInfoStorage` class. The goal for the `key_value.dart` file was to bear the `KeyValueClientAuthInfoStorage` class - which I think will get clearer after renaming the files as you suggested in other comments.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
marcelomendoncasoares: May be a good alternative. We can then interpretate which cases from this enum will be `true` or `false` on for the `refreshAuthKey` method - that receives this return - to retry or not a failed call from 401. I think we can keep this in mind and move this discussion to when we actually implement the endpoints.

packages/serverpod_client/test/integration/call_server_endpoint_retry_test.dart
marcelomendoncasoares: We do handle this! The unknown exception will be translated by `relic` into a 500. The test here is actually to show that a different error code received by the client won't trigger the refresh and retry. This `throw` could be replaced by a `return Response.badRequest()`, for example.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
marcelomendoncasoares: Done!

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
marcelomendoncasoares: Done!

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
marcelomendoncasoares: Done.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
marcelomendoncasoares: Way better with a storage hit count, indeed. Done!

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
marcelomendoncasoares: Done!

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
marcelomendoncasoares: Done!

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/cached_client_auth_info_storage_test.dart
marcelomendoncasoares: Done!

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/cached.dart
marcelomendoncasoares: Indeed, it is only used in the tests. I had added it to expose the cache to users, but it might not have a real use case, since the authSuccess is available on the `ValueListenable`. Removed, then.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/key_value_client_auth_info_storage_test.dart
marcelomendoncasoares: My idea with this test was exactly this, to "pin down" the default key so that we never change it by mistake. Do you think we need more than this test for it?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/key_value_client_auth_info_storage_test.dart
marcelomendoncasoares: You are right. I've separated the test with the improved description.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/key_value_client_auth_info_storage_test.dart
marcelomendoncasoares: This a good thumbs rule to have in mind! I'll keep that in mind. As for the tests, beside this suggestion, I've reviewed all groups following the same considerations you did on the cache test file.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/secure_client_auth_info_storage_test.dart
marcelomendoncasoares: Done!

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/key_value.dart
marcelomendoncasoares: The idea for not doing it is to allow initializing with null to go with default, like its done on the tests. Then we users can abstract the storage creation without having to copy-paste the key to other signatures, if desired.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
marcelomendoncasoares: Done!

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
marcelomendoncasoares: I had coupled the extraction from the client here, but you're right that we can do it separately. I'll move the `getCaller` call to the `authSessionManager` setter on the client, so anyone can set the desired caller directly using `setCaller`.

packages/serverpod_client/test/integration/call_server_endpoint_retry_test.dart
marcelomendoncasoares: Done.

packages/serverpod_client/test/integration/call_server_endpoint_retry_test.dart
marcelomendoncasoares: Done!

packages/serverpod_client/test/integration/call_server_endpoint_retry_test.dart
marcelomendoncasoares: It is actually `first and second calls`, because we are testing that a 401 -> refresh success -> 401 won't result in a second retry attempt. Maybe the text is not clear, then. I'll change to: `when first call fails with 401, refresh succeeds and second call also fails with 401`

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/base.dart
marcelomendoncasoares: Done!

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/client_auth_key_provider_test.dart
marcelomendoncasoares: Done.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/client_auth_key_provider_test.dart
marcelomendoncasoares: I haven't added the test here to avoid duplicating the validation that is already done on the `session_manager_setup_test.dart` file:

`Given a Client declaration when using the authSessionManagerExtension then client.auth is available`

Do you think we should still have an additional test here?

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/client_auth_key_provider_test.dart
marcelomendoncasoares: This test is indeed to test our implementation of the `ClientAuthSessionManager`. It is validating that changing the auth value in the `ClientAuthSessionManager` will reflect in auth header changes for the client - as the session manager implements the `ClientAuthKeyProvider`.

I don't think we gain much by moving this to the client with a simple implementation of the `ClientAuthKeyProvider`, because we will be mostly testing the test implementation - as the client only has the interface with the `authHeaderValue` getter. Or have I missed something?

packages/serverpod_client/test/integration/open_method_stream_retry_test.dart
marcelomendoncasoares: Done!

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/key_value.dart
SandPod: The file and class still references AuthInfo, is this deliberate?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/test/storage/key_value_client_auth_info_storage_test.dart
SandPod: No, I think this test should suffice!

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/client_auth_key_provider_test.dart
SandPod: Aah, `authSessionManager` is the keyword here, my mistake for missing the point of the test.

Yeah, might not be important to test a simple `setter` ignore my comment.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/client_auth_key_provider_test.dart
SandPod: No, that should be enough. Thanks!

packages/serverpod_client/test/integration/call_server_endpoint_retry_test.dart
SandPod: I don't think I understand this test setup completetly, in the test I only see one call being made. Perhaps we could walk this one through when you are available..

packages/serverpod_client/test/integration/call_server_endpoint_retry_test.dart
SandPod: What confuses me is that we then expect the call to the server to throw a `Exception` which looks like the servers exception. But this is really a Serverpod exception right?

packages/serverpod_client/test/integration/call_server_endpoint_retry_test.dart
marcelomendoncasoares: It's because the second call goes automatically due to the refresh-retry mechanism:

1. The first 401 triggers the refresh.
2. The refresh true triggers the retry.
3. The retry fails with 401 and does not trigger refresh-retry again.

Sure, we can take another look together!

packages/serverpod_client/test/integration/call_server_endpoint_retry_test.dart
marcelomendoncasoares: Yes, it will reach the client as a Serverpod exception. I'll change the test to make it clearer, returning an exception directly, since there is no need for a developer looking at this test to pass through this confusion.

packages/serverpod_client/test/integration/call_server_endpoint_retry_test.dart
marcelomendoncasoares: As we discussed, just improved the test setup to make it clearer for the server on what to return for each call.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/key_value.dart
marcelomendoncasoares: Yes! As we discussed, the idea is that the storage retains the authentication information, which can be an `AuthSuccess` for authenticated users, or `null`.

PR #3934, opened 2025-08-29
feat: Add support for jsonb columns

### LINKED ISSUE #3919: Support for jsonb indexable columns
### Problem to Solve

It would be great if data for a complex type, such as **another model**, a List, or a Map, would be **indexable** at the **properties level**.

Currently these types are stored as a `json` object in the database:
- https://docs.serverpod.dev/concepts/database/models#data-representation
- https://github.com/serverpod/serverpod/blob/a638ce48d6df09790e27d15a3c778c5e487d97bc/packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart#L740

### Proposal

There is a `jsonb` type in PostgreSQL. `jsonb` supports indexing, which can be a significant advantage (https://www.postgresql.org/docs/current/datatype-json.html).

It would be great if `jsonb` was used instead of `json`, but it probably cannot be changed at this stage (it might be a breaking change for a small percentage of the existing data).

Is it possible to add another `ColumnSerializableX` column type that would use `jsonb` in the database?

Or could the used database type for `ColumnSerializable` be configurable per project (`json` or `jsonb`)?

### Use Case

The `Address` class has `state` and `city` properties and we would like to index these in the database.

```
class: Company
table: company
fields:
  address: Address # This would be possible if stored as a `jsonb` instead of `json` column
```

### Alternatives

Only ugly ones like extracting `state` and `city` properties as separate fields from the `Address` class.

```
class: Company
table: company
fields:
  address: Address # Stored as a json column
  address_state: String  # Duplicate of `Address.state`
  address_city: String  # Duplicate of `Address.city`
```

### Additional context

_No response_

- [x] I want to work on this

borutjures:
This PR enables the user to specify whether to use `json` or `jsonb` when serializing serializable fields.

This is done by adding a new keyword `serialize` to models and fields:
```
class: Greeting
table: greeting
serialize: jsonb
fields:
  author: Author
  tags: List<String>, serialize=json
```

`serialize` at the model level means that all serializable fields in the model will use the specified type (`jsonb` in the above example). This can still be overridden by adding `serialize` to a field (`tags` field in the above example).

The implementation adds two experimental features:
- serializeAsJsonb: allows a use of `serialize` keyword for models and fields
- serializeAsJsonbAsDefault: uses `jsonb` as a default type when serializing

Both experimental features can be used independently:
- Enable only `serializeAsJsonbAsDefault` to use `jsonb` for all serializable fields.
- Enable only `serializeAsJsonb` to specify which models/fields should use `jsonb`.
- Enable both to use `jsonb` as default but specify some fields to use `json`.

Note: There is an existing `serialized` keyword used for enums. I intentionally use a different keyword since they have different values (`EnumSerialization(byName, byIndex` vs. `CustomSerialization(json, jsonb)`).

This PR tries to close #3919


borutjures:

tools/serverpod_cli/lib/src/analyzer/models/model_analyzer.dart
```diff
@@ -38,7 +37,7 @@ class SerializableModelAnalyzer {
 
   /// Best effort attempt to extract a model definition from a yaml file.
   static SerializableModelDefinition? extractModelDefinition(
-      ModelSource modelSource, List<TypeDefinition> extraClasses) {
+      ModelSource modelSource, GeneratorConfig config) {
```
borutjures, line 40: Changed the parameter from `config.extraClasses` to more general `config` since the downstream code needs `config.isExperimentalFeatureEnabled()`. This way `_parseModelFieldDefinition()` gets access to both by using a single parameter.

packages/serverpod/lib/src/models/database/column_type.spy.yaml
```diff
@@ -31,3 +31,5 @@ values:
   - sparsevec
   ### Dart type: [Bit]
   - bit
+  ### Esp. for serializable objects as [jsonb].
```
borutjures, line 35: Added at the bottom so that the column type ids don't change.

packages/serverpod/lib/src/database/analyze.dart
```diff
@@ -169,6 +169,7 @@ WHERE t.relname = '$tableName' AND n.nspname = '$schemaName';
       }
 
       var parameters = <String, String>{};
+      GinOperatorClass? ginOperatorClass;
```
marcelomendoncasoares, line 172: Shouldn't this be populated from the `queryResult`? Otherwise, when starting the database, it would flag an inconsistency with the database and throw an error.

packages/serverpod/lib/src/models/database/column_type.spy.yaml
marcelomendoncasoares: Good that you had this into consideration! It is indeed important to preserve the order, as this enum is serialized by index.

packages/serverpod/lib/src/models/database/custom_serialization.spy.yaml
```diff
@@ -0,0 +1,4 @@
+enum: CustomSerialization
```
marcelomendoncasoares: It is worth making this class name clearer, since serialization is very broad. Maybe something like `JsonSerializationStrategy`?

Might also be good to make serialized by name and add comments to the values - describing one as text and other as binary - to better guide less familiar users.

packages/serverpod/lib/src/models/database/gin_operator_class.spy.yaml
```diff
@@ -0,0 +1,8 @@
+### Operator classes used by GIN indexes.
```
marcelomendoncasoares, line 1: Also worth adding docs to each value of this class.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -75,6 +75,9 @@ final class ModelClassDefinition extends ClassDefinition {
   /// If set to true the class is sealed.
   final bool isSealed;
 
+  /// If set, the type of database column type this class should use for custom serialization.
+  final CustomSerialization? serialize;
```
marcelomendoncasoares: In the same direction as the change of the class name, it is worth clarifying this to something like `jsonSerializationStrategy`, following the same semantics as the class name. This way it gets clear of which serialization we are talking here.

This name should be the same across all usages of the feature (project-level, class-level, field-level) and when referencing it on the code (variables, functions, parameters) to ensure consistency.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -346,6 +365,15 @@ class ModelParser {
     var scope = _parseClassFieldScope(node, serverOnlyClass);
     var shouldPersist = _parseShouldPersist(node);
 
+    typeResult.serialize = _parseClassFieldSerialize(node);
+    if (typeResult.serialize == null) {
+      final serializeAsJsonbAsDefault =
+          config.isExperimentalFeatureEnabled(ExperimentalFeature.serializeAsJsonbAsDefault);
+      if (serializeAsJsonbAsDefault) {
+        typeResult.serialize = CustomSerialization.jsonb;
```
marcelomendoncasoares: Are we setting this for every field, even the ones that are not serialized as JSON in the database? If so, we should add a condition here to only configure it for the fields that are actual JSON.

tools/serverpod_cli/lib/src/analyzer/models/model_analyzer.dart
marcelomendoncasoares: Ok! It doesn't seem like a problem to me, as adding and removing experimental features will keep happening in the future. This way we avoid having to add and remove experimental parameters often.

packages/serverpod/lib/src/models/database/gin_operator_class.spy.yaml
```diff
@@ -0,0 +1,8 @@
...
+enum: GinOperatorClass
+serialized: byName
+values:
+  - array
+  - jsonb
+  - jsonbPath
+  - tsvector
```
marcelomendoncasoares, line 9: It is good to add an `unknown` here in case Postgres start supporting new types. This way, older code keeps forward compatible while eventual new types are not yet supported.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -175,7 +181,7 @@ class ModelParser {
     );
 
     var serverOnly = _parseServerOnly(documentContents);
-    var serializeAs = _parseSerializedAs(documentContents);
+    var serializedAs = _parseSerializedAs(documentContents);
```
marcelomendoncasoares, line 188: As we will change the `serialize` to something like `jsonSerializationStrategy`, we don't need to worry about the name conflict and leave the original here.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -1496,6 +1519,28 @@ class Restrictions {
...
+    var definition = documentDefinition;
+    if (definition is! ClassDefinition) return [];
+
+    var errors = <SourceSpanSeverityException>[];
+
+    if ((definition is ModelClassDefinition) &&
+        (definition.tableName != null) &&
+        (parentNodeName == defaultPrimaryKeyName)) {
+      errors.add(SourceSpanSeverityException(
+        'The "${Keyword.scope}" key is not allowed on the "id" field.',
```
marcelomendoncasoares: Here should be `Keyword.serialize` (jsonSerializationStrategy) instead of `Keyword.scope`, right?

tools/serverpod_cli/lib/src/config/experimental_feature.dart
```diff
@@ -15,7 +15,11 @@ enum ExperimentalFeature {
 
   // TODO: Remove when inheritance is enabled by default.
   // Tracked by issue: https://github.com/serverpod/serverpod/issues/2711
-  inheritance;
+  inheritance,
+
+  // Tracked by issue: https://github.com/serverpod/serverpod/issues/3919
+  serializeAsJsonb,
+  serializeAsJsonbAsDefault;
```
marcelomendoncasoares: Do we really need two experimental flags? I understand the idea of allowing users to opt-in with the jsonb serialization on models or fields without changing all of the existing models, but it feels like we are using experimental flags to set project-wide configurations. It creates a problem to further remove the experimental flag, as it would set the jsonb as default for everyone with no opt-out mechanism.

To properly provide this global option, the best way is to add an optional field on the `generator.yaml` file. Then, we can have only one `jsonb` experimental flag,

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -2144,6 +2147,42 @@ class SerializableModelLibraryGenerator {
     });
   }
 
+  Expression _buildModelTableCustomSerializedFieldTypeReference(
```
marcelomendoncasoares: Do we really need a new function here? Can't we bundle the custom serialization login inside the already existing `_buildModelTableGeneralFieldExpression`? It seems to me that the only actual difference between the return of the two is that the new one adds this line commented below:

```dart
    return TypeReference((t) => t
      ..symbol = field.type.columnType
      ..url = serverpodUrl(true)
      ..types.addAll([])).call([
      literalString(field.name),
      refer('this'),
    ], {
      if (isCustomSerializedType && serializeAsJsonb) 'serialize': serializeAs!,  // <-- This line is new
      if (field.type.isVectorType)
    });
```

To add this line, the main difference between the body of the two functions is checking the experimental feature and default serialization type. But do we need to parse the config settings here again? Since the code generation only happens after the analyzer completes, we should expect that all configs required to generate the code can be found inside the specification classes - as you have already done.

So we can safely rely on the `field.type.serialize` (jsonSerializationStrategy) as not null here - if the type is custom serialized - and remove the redundant configuration parsing.

tools/serverpod_cli/lib/src/generator/types.dart
```diff
@@ -398,6 +407,16 @@ class TypeDefinition {
...
+      switch (customSerialization) {
+        case CustomSerialization.json:
+          return 'json';
+        case CustomSerialization.jsonb:
+          return 'jsonb';
+      }
+    }
+
     return 'json';
   }
```
marcelomendoncasoares: Suggestion(NIT): Although the switch here makes us check match completeness with the enum, it feels redundant with the final return. Could be simplified to:

```suggestion
  if (isCustomSerializedType && serialize == CustomSerialization.jsonb) {
    return 'jsonb';
  }
  return 'json';
```

marcelomendoncasoares (changes requested):
Great work on this one, @borutjures! Very clean implementation, with everything in the right places! I made some considerations regarding names used for the feature and other smaller code improvements.

I haven't thoroughly reviewed the tests yet, but I've noticed that you only included unit tests on the `serverpod_cli`. I suggest taking a look at this merge commit https://github.com/serverpod/serverpod/commit/7979b3d19f9183780b845e379205cc7e4e49e479 as a reference to other tests that we must cover for this feature.

As you can see in the commit, beside unit tests on the packages, we use the `tests/serverpod_test_server` to add integration tests with actual models for all of the framework features. So it is good to also create some models with jsonb there and add similar tests.

Let me know if you need any further guidance or if you want to schedule a sync to discuss any implementation detail! I'll keep an eye here as well to iterate with you on the next changes.

packages/serverpod/lib/src/database/analyze.dart
borutjures: Thank you for catching this. I added the variable as a placeholder but forgot to mark it with TODO.
The code now reads the value from the database.

packages/serverpod/lib/src/models/database/custom_serialization.spy.yaml
borutjures: I followed the pattern from `EnumSerialization`. I renamed it to `JsonSerializationDataType`. Notice `DataType` instead of `Strategy`. Strategy sounded "too much" – I picked DataType as used in the PostgreSQL documentation.

Instead of comments to values I added a link to the PostgreSQL documentation page.

packages/serverpod/lib/src/models/database/gin_operator_class.spy.yaml
borutjures: Instead of comments to values I added a link to the PostgreSQL documentation page.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
borutjures: Done.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
borutjures: Fixed. Thank you for catching it.

packages/serverpod/lib/src/models/database/gin_operator_class.spy.yaml
borutjures: I'm not sure about the `unknown`. My understanding is that the `GinOperatorClass` is an optional parameter for an index. If a new type is added to Postgres in the future, the old code will keep working and the new type will be available in Serverpod only after it is implemented.

I checked how `unknown` is used in other places. E.g. `ColumnType.unknown` allows for new unsupported types when db is not created by Serverpod.

`SerializableModelIndexDefinition.ginOperatorClass` is optional and I would prefer to follow the example of `SerializableModelIndexDefinition.vectorDistanceFunction` which also doesn't use `unknown` – it uses `null` instead.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
borutjures: This change to the variable name was to make it consistent with the keyword and method name for the enum file.

Keyword: serialized
Method name: ...serializedAs
Old variable: serializeAs
New variable: **serializedAs**

The variable name was renamed (`d` added) unrelated to the new JsonSerializationStrategy feature. Are such small spelling corrections allowed or should I remove it from this PR?

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
borutjures: Yes. Thank you.

tools/serverpod_cli/lib/src/config/experimental_feature.dart
borutjures: I moved `serializeAsJsonbAsDefault` to an optional field on the generator.yaml file as suggested.

Moving it to `enabledFeatures` might also be an option.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
borutjures: I was extra careful not to change the existing code. I moved the code as suggested.

tools/serverpod_cli/lib/src/generator/types.dart
borutjures: Done.

packages/serverpod/lib/src/models/database/custom_serialization.spy.yaml
borutjures: Another candidates for the name: `ColumnSerializableDataType` or `SerializableDataType`.

Instead of `ColumnSerializable.jsonSerializationDataType`, we would have `ColumnSerializable.columnSerializableDataType` or shorter `ColumnSerializable.serializableDataType`.

This would make the enum ready in case we need to serialize to something different than JSON in the future.

borutjures:
@marcelomendoncasoares I went through the recommended merge commit and looked at the integration tests it added.

I added only one integration test. It looks to me that `jsonb` fields are so close to the database layer that the existing integration tests for `json` already cover `jsonb` fields too.

Please let me know if I missed anything.

The only thing I would like to change before going further is renaming of `JsonSerializationDataType` to `SerializationDataType` (as mentioned in a comment above).

packages/serverpod/lib/src/models/database/gin_operator_class.spy.yaml
marcelomendoncasoares: Makes sense, since it is nullable. We can go without the unknown.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
marcelomendoncasoares: As it is a minor naming change that improves consistency, not actually a problem to leave the change here. We usually avoid bundling changes that are not related to the scope of the PR to facilitate revision, but this is too small to bother. No worries, then!

tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -367,6 +372,8 @@ class GeneratorConfig implements ModelLoadConfig {
       }
     }
 
+    var serializeAsJsonbAsDefault = _loadSerializeAsJsonbAsDefault(file, generatorConfig);
```
marcelomendoncasoares: If the config is set to true, but the experimental feature is set to false, we should raise an error here instead of passing the configuration onwards.

tools/serverpod_cli/lib/src/config/experimental_feature.dart
marcelomendoncasoares: The generator configuration is just fine, since this feature just tweaks a mechanism inside the database. The `enabledFeatures` is used for configs that broadly impact the environment - such as having a database or not.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -2113,6 +2111,23 @@ class SerializableModelLibraryGenerator {
...
+          break;
+        case JsonSerializationDataType.json:
+          serializeAs = jsonSerializationDataTypeType.property('json');
+          break;
+        case JsonSerializationDataType.jsonb:
+          serializeAs = jsonSerializationDataTypeType.property('jsonb');
+          break;
+      }
+    }
+
```
marcelomendoncasoares: Suggestion(NIT): We could simplify this by replacing the switch by a simple `.name` call on the data type directly on the `TypeReference` return - similar to what it's done with the vector function. Something like:

```dart
if (field.type.isColumnSerializable && field.type.jsonSerializationDataType != null)
  'jsonSerializationDataType': refer(
    JsonSerializationDataType.${field.type.jsonSerializationDataType!.name}',
    serverpodProtocolUrl(serverCode),
  )
```

packages/serverpod/lib/src/models/database/custom_serialization.spy.yaml
marcelomendoncasoares: Good suggestion! Having it ready for other types in the future is a valid concern. I think your suggestion of the enum on the other comment as `SerializationDataType` is very clear. Then we can use `serializationDataType` everywhere, making it consistent and clear for the users.

Regarding the doc, I would just suggest that, for this enum, it could be better to add a simple doc comment to each option instead of only the link to the Postgres documentation at the top. Just because if we add new serialization data types in the future, they could be from different sources than this one. It is still worth to mention the Postgres link on each JSON option, though.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
marcelomendoncasoares: Don't forget to change the `Keyword.serialize` now to the new `serializationDataType` name.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -738,11 +738,11 @@ class DatabaseConnection {
     if (column is ColumnSparseVector) return 'sparsevec';
     if (column is ColumnBit) return 'bit';
     if (column is ColumnSerializable) {
```
marcelomendoncasoares, line 863: Suggestion(NIT): Instead of the switch, as all the fields are just mapping to the name, could be simplified (and made clearer) by (already using the new field names):

```suggestion
    if (column is ColumnSerializable) return column.dataType?.name ?? 'json';
```

packages/serverpod/lib/src/models/database/json_serialization_datatype.spy.yaml
```diff
@@ -0,0 +1,6 @@
+### Serialization types for JSON columns.
+### See: https://www.postgresql.org/docs/current/datatype-json.html
+enum: JsonSerializationDataType
+values:
```
marcelomendoncasoares: A comment here that passed by me on the other revision: it is better to serialize `byName` instead of the current default `byIndex`, as with `gin_operator_class`.

tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -383,13 +390,19 @@ class GeneratorConfig implements ModelLoadConfig {
...
       modules: modules,
       extraClasses: extraClasses,
       enabledFeatures: enabledFeatures,
       experimentalFeatures: enabledExperimentalFeatures,
     );
   }
 
+  static bool _loadSerializeAsJsonbAsDefault(File file, Map config) {
+    if (!file.existsSync()) return false;
+    return config['serializeAsJsonbAsDefault'] ?? false;
```
marcelomendoncasoares: Suggestion(NIT): To avoid the doubled "as" we could use `serializeAsJsonbByDefault`.

marcelomendoncasoares (changes requested):
Good suggestion on the naming! You can proceed with the rename for the enum as `SerializationDataType` and the field/variables as `serializationDataType`. Also great progress on previous comments! I made a few follow-ups on the changes and will proceed with reviewing the tests later today.

tools/serverpod_cli/lib/src/config/config.dart
borutjures: Done. I looked which Exceptions are available but ended up using the plain `Exception` for now.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
borutjures: Good NIT ;-). Much better code.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
borutjures: Done. Cleaner code again. Thank you.

packages/serverpod/lib/src/models/database/json_serialization_datatype.spy.yaml
borutjures: Done.

tools/serverpod_cli/lib/src/config/config.dart
borutjures: Done.

tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -372,15 +372,24 @@ class GeneratorConfig implements ModelLoadConfig {
...
       ..._enabledExperimentalFeatures(file, generatorConfig),
       ...CommandLineExperimentalFeatures.instance.features,
     ];
 
+    var serializeAsJsonbByDefault = _loadSerializeAsJsonbByDefault(file, generatorConfig);
+    if (serializeAsJsonbByDefault) {
+      final isSerializeAsJsonbExperimentalFeatureEnabled =
+          enabledExperimentalFeatures.contains(ExperimentalFeature.serializeAsJsonb) ||
+              enabledExperimentalFeatures.contains(ExperimentalFeature.all);
+      if (!isSerializeAsJsonbExperimentalFeatureEnabled) {
```
marcelomendoncasoares: As the `GeneratorConfig` class already has an `isExperimentalFeatureEnabled` that replicates this same behavior, it might be better to move the exception to the constructor body. There you can check if the `serializeAsJsonbByDefault` is set and `!isExperimentalFeatureEnabled(ExperimentalFeature.serializeAsJsonb)` to throw without duplicating the experimental check code.

tools/serverpod_cli/lib/src/config/config.dart
marcelomendoncasoares: Not a problem. We don't expect this to be caught, so plain `Exception` is fine as long as its message is informative enough to the user.

packages/serverpod/test/database/columns/column_serializable_test.dart
```diff
@@ -23,4 +23,26 @@ void main() {
...
+    });
+
+    test('when columnName getter is called then column name is returned.', () {
+      expect(column.columnName, columnName);
+    });
+
+    test('when type is called then List<String> is returned.', () {
+      expect(column.type, List<String>);
+    });
+  });
```
marcelomendoncasoares: What are we expecting with this new test? It does not seem to add much in comparison to the already existing `ColumnSerializable<String>` test and it seems unrelated to the `jsonb` serialization. Or have I missed something?

tests/serverpod_test_server/test_integration/index_test.dart
```diff
@@ -316,4 +316,46 @@ void main() async {
...
+      var databaseDefinition = await DatabaseAnalyzer.analyze(session.db);
+
+      var table = databaseDefinition.tables.firstWhere(
+        (table) => table.name == 'object_with_jsonb',
+      );
+
+      indexes = table.indexes;
+    });
+
+    test('then the implicitly declared gin index exists.', () {
```
marcelomendoncasoares: This test happens on an index with type `explicitly` declared to `gin`. It would be good to also include an index on the model with no type declaration and validate that it is `implicitly` recognized as `gin`. Then, we should have tests for `implicitly` and `explicitly` declared type, such as with vector types.

tests/serverpod_test_server/test_integration/index_test.dart
```diff
@@ -316,4 +316,46 @@ void main() async {
       expect(index.parameters, {'lists': '300'});
     });
   });
+
+  group('Given declared ObjectWithJsonb class when analyzing database schema',
+      () {
+    late List<IndexDefinition> indexes;
+
+    setUpAll(() async {
+      // final config = GeneratorConfigBuilder().withEnabledExperimentalFeatures([ExperimentalFeature.serializeAsJsonb]).build();
```
marcelomendoncasoares: This comment can be removed.

tests/serverpod_test_server/test_integration/index_test.dart
```diff
@@ -316,4 +316,46 @@ void main() async {
...
+      expect(indexes[0].indexName, 'object_with_index_test_index1');
+      expect(indexes[0].isPrimary, false);
+      expect(indexes[0].isUnique, false);
+      expect(indexes[0].predicate, isNull);
+      expect(indexes[0].tableSpace, isNull);
+      expect(indexes[0].type, 'gin');
+      expect(indexes[0].ginOperatorClass, isNull);
+      expect(indexes[0].elements, hasLength(1));
+      expect(indexes[0].elements[0].type, IndexElementDefinitionType.column);
+      expect(indexes[0].elements[0].definition, 'indexed');
```
marcelomendoncasoares: We should reduce the checks here to only the ones necessary to validate the behavior we are testing. This way, we avoid duplication with other tests. Here we are only interested in the `type`, `ginOperatorClass` and that the `index.elements.first.definition` is the declared `indexed` column.

```suggestion
      expect(index.first.type, 'gin');
      expect(index.elements.length, 1);
      expect(index.first.ginOperatorClass, isNull);
      expect(index.elements.first.definition, 'indexed');
```

tests/serverpod_test_server/test_integration/index_test.dart
```diff
@@ -316,4 +316,46 @@ void main() async {
...
+
+      var table = databaseDefinition.tables.firstWhere(
+        (table) => table.name == 'object_with_jsonb',
+      );
+
+      indexes = table.indexes;
+    });
+
+    test('then the implicitly declared gin index exists.', () {
+      expect(indexes[0].indexName, 'object_with_index_test_index1');
```
marcelomendoncasoares: It is best to get the index by its name rather than expect the exact index position to avoid coupling on implementation details - like the order of the parsing. You can use the same strategy as the vector index tests:

```suggestion
      var index = indexes.firstWhere(
        (idx) => idx.indexName == 'object_with_index_test_index1',
      );

```

The same applies for the other tests in this file.

tests/serverpod_test_server/test_integration/index_test.dart
```diff
@@ -316,4 +316,46 @@ void main() async {
...
+      expect(indexes[0].predicate, isNull);
+      expect(indexes[0].tableSpace, isNull);
+      expect(indexes[0].type, 'gin');
+      expect(indexes[0].ginOperatorClass, isNull);
+      expect(indexes[0].elements, hasLength(1));
+      expect(indexes[0].elements[0].type, IndexElementDefinitionType.column);
+      expect(indexes[0].elements[0].definition, 'indexed');
+    });
+
+    test('then the implicitly declared gin index with operator class exists.', () {
```
marcelomendoncasoares: Apply the same comments from the test above to this and the new ones.

tests/serverpod_test_server/lib/src/models/object_with_jsonb.spy.yaml
```diff
@@ -0,0 +1,12 @@
+class: ObjectWithJsonb
+table: object_with_jsonb
+fields:
+  indexed: List<String>, serializationDataType=jsonb
+indexes:
+  object_with_index_test_index1:
```
marcelomendoncasoares: We can use a more descriptive name here, like `object_with_jsonb_default` and `object_with_jsonb_with_operator_class`. So it is clear to future developers to understand what each index is testing here.

tests/serverpod_test_server/lib/src/models/object_with_jsonb.spy.yaml
```diff
@@ -0,0 +1,12 @@
...
+    fields: indexed
+    type: gin
```
marcelomendoncasoares, line 11: As mentioned in the test below, we should also have a test index here with no declared `type` for us to validate that it is correctly assigned as `gin` if the field is serialized with `jsonb`.

tests/serverpod_test_server/lib/src/models/object_with_jsonb.spy.yaml
```diff
@@ -0,0 +1,12 @@
+class: ObjectWithJsonb
```
marcelomendoncasoares, line 1: We should also have other test model to validate the model-level config - and that we can override the model-level configuration at the field level.

tests/serverpod_test_server/lib/src/models/object_with_jsonb.spy.yaml
```diff
@@ -0,0 +1,12 @@
...
+table: object_with_jsonb
+fields:
+  indexed: List<String>, serializationDataType=jsonb
```
marcelomendoncasoares: It is worth adding another `List` or `Map` column here to assert on the integration test that its serialization data type is `json`, as the model has not modified the top-level config.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
```diff
@@ -0,0 +1,158 @@
+import 'package:serverpod_cli/src/analyzer/models/definitions.dart';
```
marcelomendoncasoares, line 1: Just a reminder: When changing the tests, also rename the files to match the `serializationDataType` feature name.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
```diff
@@ -0,0 +1,158 @@
...
+import '../../../../test_util/builders/generator_config_builder.dart';
+import '../../../../test_util/builders/model_source_builder.dart';
+
+void main() {
+  var config = GeneratorConfigBuilder().withEnabledExperimentalFeatures([ExperimentalFeature.serializeAsJsonb]).build();
+  var configWithEnabledSerializeAsJsonbByDefault = GeneratorConfigBuilder()
+      .withEnabledSerializeAsJsonbByDefault()
+      .withEnabledExperimentalFeatures([ExperimentalFeature.serializeAsJsonb]).build();
+
+  group('Given a valid class definition when validating', () {
```
marcelomendoncasoares: It is worth improving the description on the given for this test, like:

```suggestion
  group('Given class definition with no `serializationDataType` key when validating', () {
```

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
```diff
@@ -0,0 +1,158 @@
...
+
+    var definition = definitions.first as ModelClassDefinition;
+
+    test('then no errors are collected', () {
+      expect(collector.errors, isEmpty);
+    });
+
+    test('then serializationDataType is set to jsonb.', () {
+      expect(definition.serializationDataType, SerializationDataType.jsonb);
+    });
```
marcelomendoncasoares, line 75: Also add a test to validate that the field serialization did inherited the model-level configuration, which is what we really care about. Same for the test below.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
```diff
@@ -0,0 +1,158 @@
...
+import 'package:serverpod_cli/src/generator/code_generation_collector.dart';
+import 'package:serverpod_service_client/serverpod_service_client.dart';
+import 'package:test/test.dart';
+
+import '../../../../test_util/builders/generator_config_builder.dart';
+import '../../../../test_util/builders/model_source_builder.dart';
+
+void main() {
+  var config = GeneratorConfigBuilder().withEnabledExperimentalFeatures([ExperimentalFeature.serializeAsJsonb]).build();
+  var configWithEnabledSerializeAsJsonbByDefault = GeneratorConfigBuilder()
```
marcelomendoncasoares: It is worth moving this config to the specific test (or group) that is going to use it, so it does not confuse developers reading preceding tests.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
```diff
@@ -0,0 +1,158 @@
...
+    test('then no errors are collected', () {
+      expect(collector.errors, isEmpty);
+    });
+
+    test('then serializationDataType is set to json.', () {
+      expect(definition.serializationDataType, SerializationDataType.json);
+    });
+  });
+
+  group('Given a valid class definition with serializationDataType set and the default serialization for project is set, the field should be defined to serialize to json when validating', () {
```
marcelomendoncasoares: Also add a test to validate the default project configuration without override and with invalid values. It is good to move the default configuration tests to another file to follow the same pattern with the model and field level tests.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
```diff
@@ -0,0 +1,158 @@
...
+
+    var definitions = analyzer.validateAll();
+
+    var definition = definitions.first as ModelClassDefinition;
+
+    test('then no errors are collected', () {
+      expect(collector.errors, isEmpty);
+    });
+
+    test('then serializationDataType is set to jsonb.', () {
```
marcelomendoncasoares: In the same direction as the above comment (it was also mistyped as `jsonb`):

```suggestion
    test('then serializationDataType is set to the model config.', () {
```

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
marcelomendoncasoares: Since we are testing the behavior here - and not the specific combination of json in one and jsonb in the other -, it is good to describe the test as the broader behavior. This applies to other tests as well.

```suggestion
  group('Given a project with default serializationDataType set and a class definition with serializationDataType set to a different value, when validating then field serializationDataType inherits model config', () {
```

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_serialize_as_test.dart
```diff
@@ -0,0 +1,188 @@
...
+      var collector = CodeGenerationCollector();
+      var analyzer = StatefulAnalyzer(configWithEnabledSerializeAsJsonbByDefault, models, onErrorsCollector(collector));
+
+      var definitions = analyzer.validateAll();
+
+      var definition = definitions.first as ModelClassDefinition;
+      expect(definition.fields.last.type.serializationDataType, SerializationDataType.json);
+    },
+  );
+}
```
marcelomendoncasoares: This test should be on the separate file for the default project configuration.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_serialize_as_test.dart
```diff
@@ -0,0 +1,188 @@
...
+
+      var collector = CodeGenerationCollector();
+      var analyzer = StatefulAnalyzer(configWithEnabledSerializeAsJsonbByDefault, models, onErrorsCollector(collector));
+
+      var definitions = analyzer.validateAll();
+
+      var definition = definitions.first as ModelClassDefinition;
+      expect(definition.fields.last.type.serializationDataType, SerializationDataType.json);
+    },
+  );
```
marcelomendoncasoares: If we have already tested that the project overrides the model and the model overrides the field, this test gets redundant and can be removed.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_serialize_as_test.dart
```diff
@@ -0,0 +1,188 @@
...
+
+      var collector = CodeGenerationCollector();
+      var analyzer = StatefulAnalyzer(configWithEnabledSerializeAsJsonbByDefault, models, onErrorsCollector(collector));
+
+      var definitions = analyzer.validateAll();
+
+      var definition = definitions.first as ModelClassDefinition;
+      expect(definition.fields.last.type.serializationDataType, SerializationDataType.jsonb);
+    },
+  );
```
marcelomendoncasoares: This test should be on the separate file for the default project configuration.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_serialize_as_test.dart
```diff
@@ -0,0 +1,188 @@
...
+
+      var collector = CodeGenerationCollector();
+      var analyzer = StatefulAnalyzer(config, models, onErrorsCollector(collector));
+
+      var definitions = analyzer.validateAll();
+
+      var definition = definitions.first as ModelClassDefinition;
+      expect(definition.fields.last.type.serializationDataType, SerializationDataType.json);
+    },
+  );
```
marcelomendoncasoares: Since here we are testing the model override, it is better suited in the model tests file instead of this one - the same applies for the project, as already commented. This way, when inspecting each test file, we can see that all expected use cases for each feature are validated. Otherwise, a developer wanting to change model level configuration would have to look to both the class test and this one because tests are split between the two.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_serialize_as_test.dart
```diff
@@ -0,0 +1,188 @@
...
+import '../../../../../test_util/builders/model_source_builder.dart';
+
+void main() {
+  var config = GeneratorConfigBuilder().withEnabledExperimentalFeatures([ExperimentalFeature.serializeAsJsonb]).build();
+  var configWithEnabledSerializeAsJsonbByDefault = GeneratorConfigBuilder()
+      .withEnabledSerializeAsJsonbByDefault()
+      .withEnabledExperimentalFeatures([ExperimentalFeature.serializeAsJsonb]).build();
+
+  test(
+    'Given a class with a field with serializationDataType set to jsonb, then the generated field should be serialized as defined at field level.',
```
marcelomendoncasoares: It is also worth adding a test that, if not setting the `serializationDataType`, then default will be `json`.

tools/serverpod_cli/test/database/extentions/pgsql_code_generation_test.dart
```diff
@@ -597,4 +597,76 @@ END
...
+            'List',
+          )
+          .build(),
+    ];
+
+    var databaseDefinition = createDatabaseDefinitionFromModels(models, 'example', []);
+    var tableDefinition = databaseDefinition.tables.first;
+
+    test(
+        'Given a table definition with an GIN index on a jsonb field, then the SQL should include the correct GIN parameters.',
```
marcelomendoncasoares: The `Given... when.... then...` should compose a single sentence for each test to be clear what is being tested. You can follow the same pattern as the vector tests by using a `when defining` or `when creating` indexes without violating the pattern. The same applies to the other GIN tests here.

tools/serverpod_cli/test/database/extentions/index_extensions_test.dart
```diff
@@ -251,4 +251,34 @@ void main() {
       expect(index1.like(index2), isTrue);
     });
   });
+
+  group('Given gin index column types', () {
```
marcelomendoncasoares:
```suggestion
  group('Given gin index on jsonb serialized column types', () {

```

tools/serverpod_cli/test/database/extentions/index_extensions_test.dart
```diff
@@ -251,4 +251,34 @@ void main() {
...
+              [IndexElementDefinition(definition: 'gin_col', type: IndexElementDefinitionType.column)]).build();
+
+      var index2 = index1.copyWith(
+        ginOperatorClass: GinOperatorClass.jsonb,
+      );
+
+      expect(index1.like(index2), isFalse);
+    });
+
+    test('when comparing gin indexes with same gin operator classes then no mismatch is found.', () {
```
marcelomendoncasoares:
```suggestion
    test('when comparing gin indexes with same gin operator class then no mismatch is found.', () {

```

tools/serverpod_cli/test/database/extentions/column_extensions_test.dart
```diff
@@ -181,4 +181,24 @@ void main() {
...
+      ColumnDefinition sameJsonbColumn = ColumnDefinition(
+        name: 'jsonb_field',
+        columnType: ColumnType.jsonb,
+        isNullable: false,
+        dartType: 'Map<String, dynamic>',
+      );
+
+      expect(jsonbColumn.like(sameJsonbColumn), isTrue);
+    });
+  });
```
marcelomendoncasoares: Since we are storing the `jsonb` as a column type, I think we don't really need this test. Because what makes it pass is the same that makes all type-to-other-type comparisons - which are not covered here. The other tests in this file all validate cases where some of the properties doesn't match (such as int to bigint, or dimension of vector columns).

marcelomendoncasoares (changes requested):
Great work! We're mostly done with the feature implementation. I have now reviewed the tests so we can ensure proper coverage according to the repo standards. Let me know if you want to discuss anything!

Beside the comments on the tests, we always include some real world tests using the column type in the `serverpod_test_server` to ensure it works properly - or catch any edge case we might miss in the other tests. You can use the `ObjectWithJsonb` class and add a test to `tests\serverpod_test_server\test_integration\column_operations`.

tools/serverpod_cli/lib/src/config/config.dart
borutjures: `GeneratorConfig` constructor is a `const` constructor and cannot have a body. I believe there are benefits to keeping it `const`. I've added the check after the constructor.

The only other use of `GeneratorConfig` constructor is in `test/test_util/builders/generator_config_builder.dart`. I tried to write a test for `GeneratorConfigBuilder` when only `serializeAsJsonbByDefault` is set and the experimental feature `serializeAsJsonb` is not enabled. Hovewer such a test is pointless since the test should be done wherever the `GeneratorConfigBuilder` is used.

I ended up putting the same check form `GeneratorConfig.load()` into `GeneratorConfigBuilder.build()`. Is this code duplication between the library and test code acceptable?

packages/serverpod/test/database/columns/column_serializable_test.dart
borutjures: Removed. This was one of the first tests I added when learning the existing tests related to the new feature.

tests/serverpod_test_server/test_integration/index_test.dart
borutjures: I added a test for implicit index type on a `jsonb` column (the existing two tests are actually for explicit indexes).

As it turns out, the default index type for a `jsonb` column is `btree`. The only useful operators are “=”, <, <=, >, >=. This can only be used for whole object comparisons, which has a very limited use case.

I'll leave the test for now but indexes on `jsonb` columns should always specify `gin` as their type. Maybe there should be a warning when no type is specified in a model?

tests/serverpod_test_server/test_integration/index_test.dart
borutjures: Done

tests/serverpod_test_server/test_integration/index_test.dart
borutjures: Done

tests/serverpod_test_server/test_integration/index_test.dart
borutjures: Removed

tests/serverpod_test_server/test_integration/index_test.dart
borutjures: Done

tests/serverpod_test_server/lib/src/models/object_with_jsonb.spy.yaml
borutjures: Done. I used `jsonb_index_gin` and `jsonb_index_gin_with_operator_class` so it is similar to `ObjectWithVector`.

tests/serverpod_test_server/lib/src/models/object_with_jsonb.spy.yaml
borutjures: Done (see above). I also learned that the default is `btree` and not `gin`. Defaults are dangerous ;-)

tests/serverpod_test_server/lib/src/models/object_with_jsonb.spy.yaml
borutjures: Done with `ObjectWithJsonbClassLevel`.

tests/serverpod_test_server/lib/src/models/object_with_jsonb.spy.yaml
borutjures: I added three fields to `ObjectWithJsonbClassLevel`:

```
  jsonb1: List<String>
  jsonb2: List<String>, serializationDataType=jsonb
  json: List<String>, serializationDataType=json
```

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
borutjures: Missed these. Thank you.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
borutjures: Done

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
borutjures: The field doesn't store the information when it uses the model-level configuration. The model-level configuration is used if field-level one is not specified. This happens in `ModelParser._parseModelFieldDefinition()`.

There are separate tests in `test/analyzer/models/stateful_analyzer/model_validation/field/field_serialization_data_type_test.dart` that test how model-level configuration affects the field-level configuration. These should cover the described issue.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
borutjures: Done.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
borutjures: I added tests for project-level and model-level configuration combinations for the fields. See:
- `field_serialization_data_type_field_level_test.dart`
- `field_serialization_data_type_class_level_default_test.dart`
- `field_serialization_data_type_project_level_default_test.dart`

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
borutjures: Done

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
borutjures: I refactored tests for project-level and model-level configuration combinations for the fields. See:
- `field_serialization_data_type_field_level_test.dart`
- `field_serialization_data_type_class_level_default_test.dart`
- `field_serialization_data_type_project_level_default_test.dart`

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_serialize_as_test.dart
borutjures: Done (see above).

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_serialize_as_test.dart
borutjures: This is true if there is no edge case that is not handled correctly in the code. In this case this works because the `ModelClassDefinition` constructor handles it correctly when overriding the `serializationDataType` for its fields.

But somebody could easily "optimize" the code and neglect to understand how `_initializeFromClassFields()` works with its `initialize` function parameter. Current implementation expects that fields already have their `serializationDataType` set before the `ModelClassDefinition` constructor is called.

With so many other tests I would rather keep this one too ;-)

Edit:
Removed setting `field.type.serializationDataType` in `ModelClassDefinition` constructor – it is now set only in `ModelParser._parseModelFieldDefinition()`.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_serialize_as_test.dart
borutjures: Moved to `field_serialization_data_type_project_level_default_test.dart`.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_serialize_as_test.dart
borutjures: Done by refactoring to the three separate files (see above).

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_serialize_as_test.dart
borutjures: Done.

tools/serverpod_cli/test/database/extentions/pgsql_code_generation_test.dart
borutjures: Done

tools/serverpod_cli/test/database/extentions/index_extensions_test.dart
borutjures: Done

tools/serverpod_cli/test/database/extentions/index_extensions_test.dart
borutjures: Done

tools/serverpod_cli/test/database/extentions/column_extensions_test.dart
borutjures: Removed

marcelomendoncasoares:
For this next review it would be good to have the CI passing, so we avoid the need for a next review if everything is ok. Can you ping me when the branch is in mergeable state, @borutjures?

borutjures:
I've:
- merged the latest changes from `main`
- successfully run the `tools/serverpod_cli/test` tests
- successfully run the `serverpod_cli_e2e_test` tests
- executed `util/run_tests_e2e` and it ran for ~2 minutes (with no "red" lines) but it ended wish some errors:

```
serverpod_test_server-1  | Malformed call: Method "helloNonExistent" not found in endpoint: serverpod_test_module.module
...
serverpod_test_server-1  | 2025-09-13 15:00:41.993978Z ERROR: Internal server error. Zoned exception.
serverpod_test_server-1  | 2025-09-13 15:00:41.993978Z ERROR: WebSocketConnectionClosed: Connection Closed
serverpod_test_server-1  | #0      IORelicWebSocket._send (package:relic/src/adapter/io/io_relic_web_socket.dart:88:27)
serverpod_test_server-1  | #1      IORelicWebSocket.sendText (package:relic/src/adapter/io/io_relic_web_socket.dart:85:36)
serverpod_test_server-1  | #2      Endpoint.sendStreamMessage (package:serverpod/src/server/endpoint.dart:95:23)
serverpod_test_server-1  | #3      CustomClassProtocolEndpoint.streamOpened.<anonymous closure> (package:serverpod_test_server/src/endpoints/custom_class_protocol.dart:21:17)
...
serverpod_test_server-1  | 
tests-1 exited with code 0
Aborting on container exit...
 Container tests_e2e-tests-1  Stopping
 Container tests_e2e-tests-1  Stopped
 Container tests_e2e-serverpod_test_server-1  Stopping
 Container tests_e2e-serverpod_test_server-1  Stopped
 Container tests_e2e-redis-1  Stopping
 Container tests_e2e-postgres-1  Stopping
serverpod_test_server-1 exited with code 137
```

@marcelomendoncasoares Did you plan to run CI here on GitHub?

marcelomendoncasoares:
I'll keep an eye here to approve the runs as you push changes.

borutjures:
### dart format

- Some files that weren't part of this PR require formatting to pass this check. Should I format them?
- Some files are reported as changed, but formatting produces the same code (e.g. `tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialization_data_type_test.dart`)


### Serverpod integration tests (server)

2 tests fail but they don't seem to be related to the changes in this PR:
- WebServer ERROR: UnimplementedError: ExceptionRoute not implemented
- ERROR: Failed to connect to the database.
  ERROR: SocketException: Connection refused (OS Error: Connection refused, errno = 111), address = localhost, port = 34926


### Migration e2e tests

3 tests fail but they don't seem to be related to the changes in this PR:
- test_e2e_migrations/migrations_roundtrip_test.dart: Given an existing table when a new table is added that is lexically sorted after the existing table and the existing table references the new table when creating and applying migrations then both tables and the relation exist in the database.
- test_e2e_migrations/defaults/migrations_datetime_default_persist_test.dart: Given an empty database and new protocol model with table and fields with "defaultPersist" value when creating and applying migration, then the database contains the new table with the correct "defaultPersist" value for "dateTimeDefaultNow".

marcelomendoncasoares:
The dart format issue is actually due to an update to an external dependency that happened recently and require an upgrade for the local env to match the CI - since the CI always run with the latest available dependencies. Just run this and generate will stop producing diffs:

```bash
melos exec --no-flutter --category="noTemplates" -- dart pub upgrade
melos exec --flutter --category="noTemplates" -- flutter pub upgrade
```

Regarding the error on the tests, they are probably happening because you have added new models to the `serverpod_test_server` project but did not make a migration, so the server is failing to start. Just calling the create migration command on the test server should solve them.

borutjures:
### dart format (3.6.0)

- I formatted all the listed files but they are still reported as changed.


### Serverpod integration tests (server) (3.35.2)

- Doesn't run with `--apply-migrations`.


@marcelomendoncasoares Please advise on the two remaining CI issues above.

marcelomendoncasoares:
I'll try to help you with this issues. If you need further assistance, do not hesitate to message me so we can take a closer look together.

## dart format (3.6.0)

Is it possible that you are running the `dart format` command with any custom line length option? After cloning your branch and running the command, it generated the same changes as the CI action. From the changes, it seems to be a mismatch in a local line length configuration, as yours is generating longer lines.

## serverpod generate

Did you ran the melos commands to upgrade your env? We started having the same problem of local removing empty lines added on CI due to a recent upgrade in a formatter external dependency. Upgrading your local dependencies will fix the issue.

## Serverpod integration tests / Migration e2e tests

This one seems to be some bug in the implementation that must be fixed. The database is failing to start with the following message:

```
Latest database migration already applied.
WARNING: The database does not match the target database:
 - Table "object_with_jsonb" is not like the target database:
 - Index "jsonb_index_gin" mismatch: 
   - Missing Index "GIN index operator class".
 ```

This warning comes from the `serverpod/lib/src/database/extensions.dart` index comparison, on line 293:

```dart
    if (ginOperatorClass != other.ginOperatorClass) {
      mismatches.add(
        IndexComparisonWarning(
          name: 'GIN index operator class',
          expected: ginOperatorClass?.name,
          found: other.ginOperatorClass?.name,
        ),
      );
    }
```

This comparison will check the live database against the `definition.json` stored in the migration folder. The message indicates that the analysis of the current indexes using `DatabaseAnalyzer._getIndexDefinitions` did not return the expected `jsonb_index_gin` index, which is correctly present in the definition file and in the `.sql` files.

After applying the database and directly running the analyzer SQL query in the database, I can see that the index is correctly created, but there must be an issue with how it is being parsed in the `_getIndexDefinitions` function.

Could you try to debug it? It might also highlight a missing test, if we are seeing this problem only in the integration tests.

borutjures:
### dart format (3.6.0)

Adding explicit line length to `analysis_options.yaml` would solve this. I used:

```
formatter:
  page_width: 80
```

### serverpod generate

I've ran the melos commands to upgrade my env. I also ran `serverpod generate`.

### Serverpod integration tests

It turns out that PostgreSQL implicitly adds the `jsonb_ops` operator class to the gin index if none is specified. This means that the live database had this `jsonb_ops` operator class while the tables deserialized from the models didn't have it.

I added this default to the library generator.

borutjures:
I fixed the dart format errors. Other CI failures seam to be unrelated to this feature.

packages/serverpod/test/database/extensions/index_comparison_test.dart
```diff
@@ -1703,4 +1703,299 @@ void main() {
...
+        var mismatches = tableA.like(tableB);
+
+        expect(mismatches.length, 1);
+        expect(mismatches.first, isA<IndexComparisonWarning>());
+        expect(mismatches.first.subs.length, 1);
+        expect(mismatches.first.subs.first.expected, equals('jsonb'));
+        expect(mismatches.first.subs.first.found, equals('jsonbPath'));
+        expect(mismatches.first.subs.first.isMismatch, isTrue);
+      },
+    );
```
marcelomendoncasoares, line 1780: This test seems duplicated with the last one.

tests/serverpod_test_server/test_integration/index_test.dart
```diff
@@ -316,4 +316,91 @@ void main() async {
...
+      );
+
+      expect(index.type, 'gin');
+      expect(index.ginOperatorClass, GinOperatorClass.jsonb);
+      expect(index.elements.length, 1);
+      expect(index.elements.first.type, IndexElementDefinitionType.column);
+      expect(index.elements.first.definition, 'indexed2');
+    });
+
+    test('then the explicitly declared gin index with operator class exists.',
```
marcelomendoncasoares, line 348:
```suggestion
    test('then the explicitly declared gin index with operator class exists with correct operator class.',

```

tests/serverpod_test_server/lib/src/models/object_with_jsonb.spy.yaml
```diff
@@ -0,0 +1,15 @@
+class: ObjectWithJsonb
+table: object_with_jsonb
+fields:
+  indexed0: List<String>
+  indexed1: List<String>, serializationDataType=jsonb
+  indexed2: List<String>, serializationDataType=jsonb
+  indexed3: List<String>, serializationDataType=jsonb
```
marcelomendoncasoares, line 7: Suggestion: It seems the `indexed1` field is not being used on any test, right? It would be nicer if we name the fields after what we are testing on them, such as:

```suggestion
  notJsonb: List<String>
  jsonb: List<String>, serializationDataType=jsonb
  jsonbIndexed: List<String>, serializationDataType=jsonb
  jsonbIndexedGin: List<String>, serializationDataType=jsonb
  jsonbIndexedGinJsonbPath: List<String>, serializationDataType=jsonb
```

This way it gets clear on why each field is here and we can also analyze tests easier, without having to come back to the model to see how is each `index#` field used.

tests/serverpod_test_server/lib/src/models/object_with_jsonb_class_level.spy.yaml
```diff
@@ -0,0 +1,7 @@
+class: ObjectWithJsonbClassLevel
+table: object_with_jsonb_class_level
+serializationDataType: jsonb
+fields:
+  jsonb1: List<String>
+  jsonb2: List<String>, serializationDataType=jsonb
+  json: List<String>, serializationDataType=json
```
marcelomendoncasoares, line 7: Same as comment above regarding the names of the fields.

```suggestion
  implicitJsonb: List<String>
  explicitJsonb: List<String>, serializationDataType=jsonb
  json: List<String>, serializationDataType=json
```

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/indexes/index_operator_class_test.dart
```diff
@@ -0,0 +1,120 @@
...
+import 'package:serverpod_cli/src/generator/code_generation_collector.dart';
+import 'package:test/test.dart';
+
+import '../../../../../test_util/builders/generator_config_builder.dart';
+import '../../../../../test_util/builders/model_source_builder.dart';
+
+void main() {
+  var config = GeneratorConfigBuilder().withEnabledExperimentalFeatures(
+      [ExperimentalFeature.serializeAsJsonb]).build();
+
```
marcelomendoncasoares, line 13: There are a two tests that we could add here (similar to the ones at `index_distance_function_test.dart`):

- Given a class with a gin index with with non-string operator class, then collect an error about parameter type.
- Given a class with a non-gin index and gin operator class, then collect an error that operator class can only be used with gin indexes.

If we change to make `gin` default index type for `jsonb` fields on Serverpod (from previous discussion), we should also add a test here to ensure that indexes with no explicit type are parsed as `gin`.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_serialization_data_type_field_level_test.dart
```diff
@@ -0,0 +1,122 @@
...
+
+          expect(collector.errors, isNotEmpty);
+
+          var error = collector.errors.last;
+
+          expect(error.message,
+              '"Invalid" is not a valid property. Valid properties are (json, jsonb).');
+        },
+      );
+    });
```
marcelomendoncasoares, line 120: We should add a test for this on the `field_serialization_data_type_field_level_test.dart` test file, similar to the one of the others on forbidden keys, such as persist:

```dart
test(
  'Given a class with a declared id field with the "serializationDataType" key set, then an error is collected.',
  () {
    var models = [
      ModelSourceBuilder().withYaml(
        '''
        class: Example
        table: example
        fields:
          id: int?, serializationDataType=jsonb
        ''',
      ).build()
    ];

    var collector = CodeGenerationCollector();
    StatefulAnalyzer(config, models, onErrorsCollector(collector))
        .validateAll();

    expect(
      collector.errors.first.message,
      'The "serializationDataType" key is not allowed on the "id" field.',
    );
  },
);
```

tests/serverpod_test_server/test_integration/index_test.dart
marcelomendoncasoares: I agree, `btree` has little usage for `jsonb` fields and it's counterintuitive to find it as default - as just happened to both of us. I think we could safely define the default on Serverpod for `jsonb` field indexes to be `gin` if the type is omitted. Then, if an user wants a `btree`, it is always possible to explicitly specify.

tools/serverpod_cli/lib/src/config/config.dart
marcelomendoncasoares: Good call about the `const` constructor! How about moving the validation the `serializeAsJsonbByDefault` as a getter? Take a look on the suggestion I made on commit https://github.com/serverpod/serverpod/pull/3934/commits/7ae4c72d9ebd4fb7ba0cf972e048c4be45542d71.

This way we may be able to remove the code duplication and enforce the experimental feature validation if it gets used without the flag. Then it makes sense to include the test you mentioned.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
marcelomendoncasoares: I suggest we centralize the field configuration only on the field level. It makes understanding the code easier, since the serialization will be evaluated only on the model parsing and just used everywhere else. Any developer can be sure that, after the model is parsed, each field's serialization will be correctly set. And we also avoid having rules spread across the code of how to treat null on field level, precedence of model over project config, etc.

For the CLI package, all definition types (`ClassDefinition`, `EndpointDefinition`, `TypeDefinition`, etc.) are first-class citizens - not implementation details. They are the product of the parsing and the main artifacts that we rely on to build the code generation. So we can build the feature around them and use the tests to ensure their correct building during parse phase.

With this change, tests gets as simple and isolated as my previous comment, because we only need to ensure that the field serialization is being correctly overridden with each configuration setup. Once the field serialization is ensured, we already know it will be applied to the model due to the tests of the field level.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
marcelomendoncasoares: After the change from my previous comment to centralize the serialization evaluation (and override) on the parser, we can reorganize the tests and move the model and project level tests to outside of the `field` folder where they are.

File names can also be shortened to:

- `.../field/field_serialization_data_type_test.dart`
- `.../class_serialization_data_type_test.dart`
- `.../project_serialization_data_type_test.dart`

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_serialize_as_test.dart
marcelomendoncasoares: This is the kind of reduction in complexity (and uncertainty) that centralizing the field serialization definition on the parsing gives us. And we rely on tests to ensure that no optimization inadvertedly skip setting the field level property and break the feature.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/class_serialize_as_test.dart
marcelomendoncasoares: One general comment on the descriptions after the refactor is that they are now out of the `Given [scenario] when [acting] then [result]` pattern. All tests are having part of the scenario described on the `when`, instead of the `Given`. A good rule of thumb is to avoid `with` that is not related to parameter values on the `when`.

The refactor will simplify the test descriptions because we won't need to mention project-level and model-level configs on every test, since each will be scoped to one isolated level.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_serialize_as_test.dart
marcelomendoncasoares: This is the separation we are aiming for the tests, which reflects the feature being evaluated in one place only. Each upper level (model and project) is only concerned that it is overriding the below levels:

- The `class_serialization_data_type_test.dart` should only worry about overriding the field level config when it is null and not otherwise.
- The `project_serialization_data_type_test.dart` should only worry about overriding the field level config when it is null and model level is not set.

Tests should not worry about upper level configuration (model and project on field tests, project on model tests).

marcelomendoncasoares (changes requested):
Great progress, I think we are close to finish now.

I have fixed the formatting and code generation tests, and also the integration/migration tests due to the `jsonb` default value. The cause for the failures on the latter was that your fix was on the code generation, which was causing a mismatch on the database state after startup. I have moved the fix to the parsing level, in the same way that we do for vector fields, so that the default value is also present on the `definition.json` file.

To close this feature, we have mainly three things to address:

- Add tests for the experimental feature flag being required, now that it is no longer being evaluated outside of the `GeneratorConfig` class.
- Make `gin` type the default index type for `jsonb` fields when the `type` is omitted. This is what you suggested and I agree that it is a good default. We can show a hint to users during code generation that it will default to `gin` when the `type` is missing so we have no surprises for users that expect Postgres `btree` default - which is unlikely.
- Centralize the model and project level configuration on the field type during parse to avoid having the override rules spread across the codebase. This will require refactoring the tests to remove cross-configuration checking, since every configuration will be isolated from the others.

Once we are done with these ones, I think we are in good shape to one last review by Alex and then merge!

borutjures:
@marcelomendoncasoares Thank you for the changes and all the suggestions! You are an equal co-author of this feature!

Talking about co-authorship. I’ve run out of time between my client projects. Since I dedicate 100% to their project when I work for clients, I will not be able to work on this for a while. I hope you can implement what you already suggested and bring this PR to a merge.

marcelomendoncasoares:
No problem, @borutjures! We can take it from here and finish the implementation. You have done a great job so far and this will be a valuable addition to the framework!

ancill:
I’m really struggling with the outbox pattern when the payload is a plain string. Can we agree this feature would make our lives easier. If it’s already done, I hope to see it in a release soon

Thanks for your work!

t-kietzmann:
Will this be in 3.0.0?

marcelomendoncasoares:
@ancill @t-kietzmann We can not guarantee that this will be in before 3.0 release, but we will do our best for it! 

Nevertheless, even if it doesn't make before the release, it will be added very shortly after, since this PR is nearly ready and it does not introduce any breaking change. The neat part of adding it before 3.0 would be the possibility to set `jsonb` as default for the framework, but it can get in as an opt-in otherwise.

FantaMagier:
@marcelomendoncasoares Do you know when the release will be ready? :D

marcelomendoncasoares:
> @marcelomendoncasoares Do you know when the release will be ready? :D

@FantaMagier We're aiming for the release until mid-December! Then we will be able to review our backlog and prioritize having this `jsonb` feature in.

ancill:
Hey [marcelomendoncasoares](https://github.com/marcelomendoncasoares) thanks for you answer! Do you plan it in January release?

marcelomendoncasoares:
> Hey [marcelomendoncasoares](https://github.com/marcelomendoncasoares) thanks for you answer! Do you plan it in January release?

Unfortunately I can't promise a deadline, but we do want to have this in soon.

I'll post an update here once this gets to work.

Issue #1359, opened 2023-09-12
Immutability support


Isakdl:
**Is your feature request related to a problem? Please describe.**
Today we only support generated mutable classes. But Immutability comes with a lot of benefits and generally leads to code with fewer bugs. It would be great if we could allow both types of classes to be generated and it could be toggled by the user.

**Describe the solution you'd like**
Introduce a new top-level keyword on classes `immutable` that takes a bool. Defaults to `false` to keep backward compatibility (could be changed to `true` in 2.0?).  

Example:
```yaml
class: Example
immutable: true
fields:
  value: String
```

Creates the following class
```dart
class Example {
  final String value;

  Example({required this.value});
  ...
}
```

**Additional context**
Together with this, we should generate `hashCode` and `==operator` overrides to fully support the immutable use-case. There is a PR that was never merged that can be used as a basis for this here: https://github.com/serverpod/serverpod/pull/976.

jason-downs:
I'm sad to see that this has been bumped off of the roadmap entirely. Looks like the duplicate issue #1864 which was previously on the roadmap was closed since it's a duplicate of this one, which was not on the roadmap, and thus the whole thing got booted off the roadmap. Hoping this doesn't slip through the cracks, especially the related and important issue #1468, which as can be seen by the comments there, is still a major headache for me and others. Not being able to use == on model classes is still a huge blocker for converting an existing large codebase over to serverpod, requiring hacky workarounds at best.

obiwanzenobi:
I can take it, we are already using this feature in our fork of serverpod. I created pr that can be a good starting point.

SandPod:
This also needs to consider inheritance. But we might resolve that in a separate PR.

PR #3935, opened 2025-08-29
feat: Generate immutable models

obiwanzenobi:
This PR allows generating immutable models with working `equals` operator and `hash code` function. Similar to default `freezed` package default behavior.  To use it simple put immutable: true in your model yaml.

Example:

```yaml
class: ImmutableObject
immutable: true
fields:
  variable: String
```

This PR will close https://github.com/serverpod/serverpod/issues/1359


obiwanzenobi:

tests/serverpod_test_server/lib/src/generated/immutable/immutable_object.dart
```diff
@@ -0,0 +1,82 @@
...
+            (identical(
+                  other.variable,
+                  variable,
+                ) ||
+                other.variable == variable);
+  }
+
+  @override
+  int get hashCode {
+    return Object.hash(
```
SandPod, line 50: Suggestion: Use `Object.hashAll([...])` so that we can support any number of fields and not just 20.

tests/serverpod_test_server/lib/src/models/immutable/immutable_object.spy.yaml
```diff
@@ -0,0 +1,4 @@
+class: ImmutableObject
```
SandPod, line 1: Suggestion: Add an immutable object that contains:
- Another immutable object
- Records
- Multiple fields with different types (validates that we don't just support 1 but n fields)
- No fields (validates we support 0)

This validates that we behave correctly for these cases.

tools/serverpod_cli/lib/src/analyzer/models/yaml_definitions/class_yaml_definition.dart
```diff
@@ -40,6 +40,10 @@ class ClassYamlDefinition {
         isHidden: !restrictions.config
             .isExperimentalFeatureEnabled(ExperimentalFeature.inheritance),
       ),
+      ValidateNode(
```
SandPod, line 43: Question: How does immutability work in combination with inheritance? Please add a model that exhibits this behavior and a test that validates it is preserved over time (if applicable).

tests/serverpod_test_server/lib/src/models/immutable/immutable_object.spy.yaml
```diff
@@ -0,0 +1,4 @@
...
+immutable: true
```
SandPod, line 2: Suggestion: Please add an immutable model that has a `table` and use it in database interactions.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -410,6 +418,13 @@ class SerializableModelLibraryGenerator {
           hasImplicitClass: hasImplicitClass,
         ));
       }
+
+      // Immutability
+      if (classDefinition.isImmutable) {
```
SandPod, line 429: Chore: Please add a unit test that validates that these methods are added for immutable classes.

The test can be added in a new tesfile in `tools/serverpod_cli/test/generator/dart/server_code_generator/` and  `tools/serverpod_cli/test/generator/dart/client_code_generator/`.

You can use `class_constructors_test.dart` as inspiration for how to structure the test.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -21,6 +21,8 @@ class ModelParser {
   ) {
     var isSealed = _parseIsSealed(documentContents);
 
+    var isImmutable = _parseIsImmutable(documentContents);
```
SandPod, line 24: Chore: Please add a unit test that validates that this property is parsed correctly.

This test should be added in a new file in `tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation`. Inspiration can be taken from `class_type_test.dart`.

tests/serverpod_test_client/lib/src/protocol/immutable/immutable_object.dart
```diff
@@ -0,0 +1,76 @@
...
+  /// with some or all fields replaced by the given arguments.
+  @_i1.useResult
+  ImmutableObject copyWith({String? variable});
+  @override
+  bool operator ==(Object other) {
+    return identical(
+          other,
+          this,
+        ) ||
+        other.runtimeType == runtimeType &&
```
SandPod, line 38: Question/Discussion: Should `runtimeType` be included in the equals operator?

This will prevent a subclass without any additional fields to be equal to a parent class. What is best practice on this?
If we remove `runtimeType` it should also be removed from the hashCode.

tests/serverpod_test_server/test/generated_methods/immutable_test.dart
```diff
@@ -0,0 +1,93 @@
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('Equals behavior', () {
+    test(
+        'Given an immutable object when comparing to other object with same value then objects should be equal',
```
SandPod: Suggestion: Change to `Given two identical immutable objects when comparing equality then equality comparison should return true`.

Right now, the `Given` part does not accurately reflect what the prerequisite are for the test, and the `When` part does not only contain the action performed. Formulating these correctly will provide better guidance for future developers if they accidentally break this behavior and especially if they break in CI.

Please apply the same fix to all tests in this file.

tests/serverpod_test_server/test/generated_methods/immutable_test.dart
```diff
@@ -0,0 +1,93 @@
...
+        () {
+      final firstObject = ImmutableObject(variable: 'value');
+      final secondObject = ImmutableObject(variable: 'value');
+      expect(firstObject == secondObject, true);
```
SandPod: Suggestion/Question: Can you use `equals()` here?

This will provide a better error message if the test fails, as this will just yield `expected true got false`

SandPod (changes requested):
Nice work with this PR!

The solution looks great. I have a couple of additional requirements around testing to validate this works as expected and continues to function as we evolve the framework. I also had a question around including `runtimeType`.

Please address the comments, but leave them open, and request a re-review when solved. I'll resolve them as part of the re-review.

tests/serverpod_test_server/lib/src/generated/immutable/immutable_object.dart
obiwanzenobi: It is done by purpose. Generator is using `hashAll` if there is more than 20 arguments, and `hash` for less than 20.
I will add additional example model that will have it - so it will be clear.

Using `hashAll` always, will make generator code more readable, but there are also some points that are against that:

1.  It is used the same way by libraries like freezed: 
https://github.com/rrousselGit/freezed/blob/master/packages/freezed/lib/src/templates/concrete_template.dart#L501
2. We will introduce the overhead of creating a list everytime `hashCode` is used.
3. Dart lang implementation looks more performant for [hash function](https://api.flutter.dev/flutter/dart-core/Object/hash.html) in comparison to [hash all]( https://api.flutter.dev/flutter/dart-core/Object/hashAll.html).

SandPod: Thank you for motivating the choice. This makes sense to me.

Please add a test that validates the switch to `hashAll` works as expected.

kamil-matula:

tests/serverpod_test_client/lib/src/protocol/immutable/immutable_object.dart
kamil-matula: Structural equality between instances of two different classes, even if one is a subclass of another, is super rare in programming languages. Libraries like [equatable](https://pub.dev/packages/equatable) or [freezed](https://pub.dev/packages/freezed) use `runtimeType` in their implementations of overridden `==` operator and `hashCode` getter, so I agree with having it there as in @obiwanzenobi's solution

tests/serverpod_test_server/lib/src/models/immutable/immutable_object.spy.yaml
kamil-matula: In current implementation there is no way to have an immutable model that represents a database table. Even though the `id` field can be non-nullable ([docs](https://docs.serverpod.dev/concepts/database/models#default-uuid-model-value)), it cannot be marked as `final` and the constructor cannot be marked as `const`. Combining these two functionalities would mean that row in db cannot be updated, which would require definitely more work. For now I will just mark `isImmutable` keyword as mutually exclusive to `table` in `class_yaml_definition.dart`

tools/serverpod_cli/lib/src/analyzer/models/yaml_definitions/class_yaml_definition.dart
kamil-matula: Due to usage of `runtimeType` in `==` operator and `hashCode` getter subclass's instance cannot be compared with superclass's instance, but two instances of subclass can be normally compared to each other - I will add tests for both cases

obiwanzenobi:
Thanks to the help of @kamil-matula from TCB this can be re reviewed

tests/serverpod_test_server/test/generated_methods/immutable_test.dart
```diff
@@ -142,6 +142,34 @@ void main() {
       );
       expect(firstObject, isNot(equals(secondObject)));
     });
+
+    test(
+        'Given two immutable child objects with identical fields when comparing equality then equality comparison returns false',
```
SandPod: Suggestion: Description is wrong, change to `then equality comparison returns true`.

tests/serverpod_test_server/test/generated_methods/immutable_test.dart
```diff
@@ -222,5 +250,20 @@ void main() {
 
       expect(copy, isNot(equals(original)));
     });
+
+    test(
+        'Given an immutable child object when creating a copy with a new scalar value then returned object has updated value original remains unchanged and objects are not equal',
```
SandPod: Suggestion: Break this apart into multiple tests asserting the different expected.

This test group should have a shared setup that creates the immutable child and its copy with the new value.

Then, the tests should target a single scenario at a time. This will give any future developer more precise information about what went wrong when any of these tests fail.

tests/serverpod_test_client/lib/src/protocol/immutable/immutable_object.dart
SandPod: Sounds good! 👍 

Then we preserve it as is.

tests/serverpod_test_client/lib/src/protocol/immutable/immutable_object.dart
```diff
@@ -0,0 +1,63 @@
...
+// ignore_for_file: non_constant_identifier_names
+// ignore_for_file: public_member_api_docs
+// ignore_for_file: type_literal_in_constant_pattern
+// ignore_for_file: use_super_parameters
+// ignore_for_file: invalid_use_of_internal_member
+
+// ignore_for_file: no_leading_underscores_for_library_prefixes
+import 'package:serverpod_client/serverpod_client.dart' as _i1;
+
+class ImmutableObject implements _i1.SerializableModel {
```
SandPod, line 16: Question: Shouldn't we add the `immutable` annotation to the class as well?

So that we get the static analysis benefit of the class being immutable.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/immutable_test.dart
```diff
@@ -0,0 +1,80 @@
+import 'package:serverpod_cli/analyzer.dart';
+import 'package:serverpod_cli/src/analyzer/models/stateful_analyzer.dart';
+import 'package:serverpod_cli/src/generator/code_generation_collector.dart';
+import 'package:serverpod_service_client/serverpod_service_client.dart';
```
SandPod: chore: Remove unused import.

tests/serverpod_test_server/lib/src/models/immutable/immutable_object.spy.yaml
SandPod: Wouldn't this minor fix enable this?
I think this is how it should be; we shouldn't really require that this field exists, just require that we can fetch the value it contains from getter.


```
diff --git a/packages/serverpod/lib/src/database/concepts/table.dart b/packages/serverpod/lib/src/database/concepts/table.dart
index c85df092d..6c742a102 100644
--- a/packages/serverpod/lib/src/database/concepts/table.dart
+++ b/packages/serverpod/lib/src/database/concepts/table.dart
@@ -6,11 +6,11 @@ import 'package:serverpod_serialization/serverpod_serialization.dart';
 /// typically generated. Instances of [TableRow] can also be serialized and
 /// either passed to clients or cached.
 abstract interface class TableRow<T_ID> implements SerializableModel {
-  TableRow(this.id);
+  TableRow();
 
   /// The id column of the row. Can be null if this row is not yet stored in
   /// the database.
-  T_ID id;
+  T_ID get id;
 
   /// The table that this row belongs to.
   Table<T_ID> get table;
```

With this change, and manually adding the `final` to the id field in the generated code, I could generate an immutable object with a table connection.

Then I could run this test to validate that it works as expected:

```dart

import 'test_tools/serverpod_test_tools.dart';

void main() {
  withServerpod('Given an immutable object with a table', (sessionBuilder, _) {
    late Session session;
    setUp(() async {
      session = await sessionBuilder.build();
    });

    test('When inserting, then it should be inserted correctly', () async {
      var object = ImmutableObjectWithTable(variable: 'initial value');
      var insertedObject =
          await ImmutableObjectWithTable.db.insertRow(session, object);
      expect(insertedObject.id, isNotNull);
      expect(insertedObject.variable, 'initial value');

      var fetchedObject = await ImmutableObjectWithTable.db
          .findById(session, insertedObject.id!);
      expect(fetchedObject, isNotNull);
      expect(fetchedObject!.id, insertedObject.id);
      expect(fetchedObject.variable, 'initial value');
    });

    test('When updating, then it should be updated correctly', () async {
      var object = ImmutableObjectWithTable(variable: 'initial value');
      var insertedObject =
          await ImmutableObjectWithTable.db.insertRow(session, object);

      var updatedObject = await ImmutableObjectWithTable.db.updateRow(
          session, insertedObject.copyWith(variable: 'updated value'));
      expect(updatedObject.id, insertedObject.id);
      expect(updatedObject.variable, 'updated value');

      var fetchedObject = await ImmutableObjectWithTable.db
          .findById(session, insertedObject.id!);
      expect(fetchedObject, isNotNull);
      expect(fetchedObject!.id, insertedObject.id);
      expect(fetchedObject.variable, 'updated value');
    });
  });
}
```

SandPod (commented):
Great next step, almost there.

I think it is very simple to make sure we support this with tables as well. Please implement my suggested fix for that and remove the mutual exclusion.

I added a suggestion for adding the annotation to the generated model to preserve the static analysis support.

Should we also support this keyword for exceptions?

tests/serverpod_test_server/test/generated_methods/immutable_test.dart
kamil-matula: Of course, good thought. Already updated

tests/serverpod_test_server/test/generated_methods/immutable_test.dart
kamil-matula: Good catch, thanks!

tests/serverpod_test_client/test/immutable_test.dart
```diff
@@ -183,60 +183,105 @@ void main() {
   });
 
   group('copyWith', () {
-    test(
-        'Given an immutable object when creating a copy with a new scalar value then returned object has updated value original remains unchanged and objects are not equal',
+    group(
+        'Given an immutable object when creating a copy with a new scalar value',
         () {
       final original = ImmutableObject(variable: 'original');
       final copy = original.copyWith(variable: 'updated');
```
SandPod: Suggestion: Define these as late and assign them in a `Setup`.

By placing these definitions in the group, they are invoked each time any test in this file is run.

This becomes a problem if a developer would like to debug a single test in this file without having all the other tests run. Imagine you place a debug breakpoint inside of `copyWith` to validate a single scenario.

Please make this modification to all concerned tests in this file.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -428,6 +428,8 @@ class SerializableModelLibraryGenerator {
       if (classDefinition.isImmutable) {
         classBuilder.methods.add(_buildEqualOperator(classDefinition, fields));
         classBuilder.methods.add(_buildHashCodeMethod(classDefinition, fields));
+        classBuilder.annotations
+            .add(refer('immutable', 'package:meta/meta.dart'));
```
SandPod: Suggestion: Please re-export this from `serverpod` and `serverpod_client` and import it from there by replacing `'package:meta/meta.dart'` to `serverpodUrl(serverCode)`.

This removes the need for our end users to add this import again.

Add the `immutable` export in these places:
- https://github.com/serverpod/serverpod/blob/d453752bbc23432286f648b12989da132bd57cda/packages/serverpod/lib/serverpod.dart#L40
- https://github.com/serverpod/serverpod/blob/d453752bbc23432286f648b12989da132bd57cda/packages/serverpod_client/lib/serverpod_client.dart#L10

tests/serverpod_test_client/lib/src/protocol/immutable/immutable_object.dart
kamil-matula: For some reason this file did not get `Outdated` badge and shows old state - it was updated in `chore: add immutable annotation to immutable models` commit (https://github.com/serverpod/serverpod/pull/3935/commits/e52bb048cf818e9a9369f227b7eefe0b1931fcb1).

kamil-matula:
Thank you for another review, @SandPod. I have applied all of your suggestions. Feel free to take a look when you have some time

tests/serverpod_test_server/config/test.yaml
```diff
@@ -19,15 +19,17 @@ webServer:
   publicScheme: http
 
 database:
-  host: postgres
```
SandPod, line 22: Chore: Please revert any modifications to this file.

Password is not a valid field in this file and our CI infrastructure depend on these host names.

tests/serverpod_test_server/test_integration/immutable_test.dart
```diff
@@ -0,0 +1,57 @@
...
+    late Session session;
+
+    setUp(() async {
+      session = await sessionBuilder.build();
+      await session.db.unsafeExecute('''
+        CREATE TABLE IF NOT EXISTS $tableName (
+          id SERIAL PRIMARY KEY,
+          variable TEXT NOT NULL
+        );
+      ''');
```
SandPod: Suggestion: Create a migration and remove table management from the test.

When adding new models with tables we need create a new migration in the test project.

Please run `serverpod create-migration` in `tests/serverpod_test_server`. This will create a migration that is automatically applied the database.

When that is done you can remove creation and removal of the table from this test.

tests/serverpod_test_server/test_integration/immutable_test.dart
```diff
@@ -0,0 +1,57 @@
...
+    });
+
+    tearDown(() async {
+      await session.db.unsafeExecute('DROP TABLE $tableName;');
+    });
+
+    test('When inserting, then it should be inserted correctly', () async {
```
SandPod: Suggestion: Break this test up into two tests. One test that validates that an object can simply be inserted in the database (that the call succeeds), and one test that validates that `Given an immutable model stored in the database, when fetching it, then it is successfully returned`. 

This will give future developers more detailed information on what went wrong. Serialization down to the database or fetching it.

Also cover these additional scenarios:
- Assert that equals between the constructed object and the one retrieved from the database fails, since the database object now contains an ID.
- Assert that the returned object from the `insertRow` method and a fetched object from the database pass the equals check.
- Assert that equals between two objects fetched from the database pass.


For the tests that require an object to already exist in the database. Please insert that object in the setup for the test. This can be used for the "fetch" tests and for the "update" tests.

tests/serverpod_test_server/config/test.yaml
kamil-matula: Oh, sorry, this didn't mean to be committed 😅 Thanks for the comment!

tests/serverpod_test_server/lib/src/models/immutable/immutable_object.spy.yaml
kamil-matula: Worked! Thanks

kamil-matula:
Thanks for review, @SandPod. Could you re-run CI? At the moment I'm not sure how to handle errors in `dart analyze downgrade` tests - I have downgraded Flutter on my side and IDE has no problems with the syntax 🤔 

EDIT: Nevermind, my Dart stayed at 3.7.0+ anyway, so it wasn't a valid test on my side

kamil-matula:
Hopefully the newest update fixes `update pubspecs` tests - the `collection` dependency was not in template, but in the client's pubspec file instead

kamil-matula:
The `dart analyze downgrade` tests are using `code_builder` v4.5.0 (instead of the newest v4.11.0), which has no `Expression.parenthesized` that we introduced in our code. This usage has been replaced with custom method, so the error should no longer be thrown

The `dart analyze` tests started failing after moving `collection: ^1.17.1` from serverpod_client's pubspec file to templates - unfortunately I forgot to run `util/update_pubspecs` script as Contributing Guidelines suggest. It has been run, so the problem should be gone + it should also fix `update pubspecs` tests

SandPod:
@kamil-matula - I'll take a look a last look at this tomorrow morning, if there is anything simple to fix, is it OK if I patch the branch directly?

Looking forward to getting this in, both because this is an excellent addition to the framework, but also because it will make you a contributor, meaning I won't need to approve the CI runs 😄

kamil-matula:
Sure, go ahead 😇 I've applied all your suggestions so far and it looks like all tests have finally passed, so there shouldn't be too much to add. Thanks so much for your reviews during past weeks, especially today and yesterday, that sped up the process 💪

SandPod:
@kamil-matula - It seems like i'm not allowed to push to the fork. I just have a minor cleanup for the tests and also have included a merge to main. Could you configure the fork so that maintainers are allowed to push?

obiwanzenobi:
I can do it, give me a moment

obiwanzenobi:
We have a little problem because of https://github.com/orgs/community/discussions/5634

obiwanzenobi:
@SandPod because of the github limitations for organizations I have to give you temporary write permission to our fork. Please check if with that you can do your part

SandPod:
@obiwanzenobi - ~~Still seem to have issues pushing to your branch:~~

```
➜  generate-immutable-models git:(feat/generate-immutable-models) git push
ERROR: Permission to thecodebrothers/serverpod.git denied to SandPod.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Sorry, my bad. I had not accepted the invitation! 🙃

SandPod:
@obiwanzenobi - Would you be up for adding this to our documentation as well? You can find the documentation here: https://github.com/serverpod/serverpod_docs

PR #3938, opened 2025-09-01
feat: Added support for Cloudflare R2 storage

tylersavery:
Hi Serverpod team!

I wanted to create support for Cloudflare R2 as a storage backend. My first attempt was just to allow the endpoint to be overridden since I was under the impression the two APIs were basically interchangeable, but I learned that Cloudflare uses PUT requests while S3 allows for POST (which is what the s3 integration uses). 

My next attempt (which I'm PRing with here) was to create a new integration called `serverpod_cloud_storage_r2` which uses the same pattern as the s3 integration as well as the AWS same library (`amazon_cognito_identity_dart_2`) for identity. It was important to me that no new packages would be required for this implementation.

Additionally, i needed to modify the serverpod_client FileUploader class to optionally allow overriding the method when uploading the file to the bucket. It is able to parse this from the upload description but falls back to POST by default.

In doing this, I noticed that the mimetypes were always application/octet-stream (I believe for S3 as well) so I've included a helper function that will detect the mime based on the extension of the file. I did not incorporate this change with the other two integrations.  

In the spirit of following the patterns of the s3 integration, I did not add tests. I assume this was purposefully omitted previously because it would require an S3 bucket / credentials for the tests to run. Would be happy to do so (I'd just need a bit of guidance on the best way to manage keys and such for this and I'd just follow the same test cases as the database upload).

Additionally, I believe this new integration will also support users who want to use other s3 alternatives such as minos or hetzner because they also use PUT requests. I haven't taken the time to test these out yet but I expect it should work – or at least be very close.

Here is an example implementation. See the integration's readme for more details!

server.dart
```
import 'package:serverpod_cloud_storage_r2/serverpod_cloud_storage_r2.dart' as r2;

pod.addCloudStorage(r2.R2CloudStorage(
    serverpod: pod,
    storageId: 'public',
    public: true,
    region: 'auto',
    bucket: 'my-bucket',
    accountId: "cloudflare-account-id",
    publicHost: "https://mydomain.com",
  ));
  ```

> The asset endpoint doesn't require any changes from what is in serverpod's docs.
> The flutter app doesn't require any changes either from what is in serverpod's docs.

Thanks for considering this PR and I hope you're having a great time at Flutter & Friends!

### DIFF
--- a/integrations/serverpod_cloud_storage_r2/CHANGELOG.md
+++ b/integrations/serverpod_cloud_storage_r2/CHANGELOG.md
@@ -1,0 +1,7 @@
+## 3.0.0-alpha.1
+
+- Initial release of Cloudflare R2 integration for Serverpod
+- Support for basic file operations (upload, download, delete)
+- Direct upload support with both POST multipart and PUT presigned URL methods
+- S3-compatible API integration
+- Public and private file access modes
--- a/integrations/serverpod_cloud_storage_r2/README.md
+++ b/integrations/serverpod_cloud_storage_r2/README.md
@@ -1,0 +1,141 @@
+# Serverpod Cloud Storage R2
+
+This package provides Cloudflare R2 cloud storage integration for Serverpod applications.
+
+## Features
+
+- **File Operations**: Upload, download, delete files in Cloudflare R2
+- **Direct Upload Support**: PUT (presigned URL) method optimized for R2
+- **Public/Private Files**: Support for both public and private file access
+- **S3 Compatibility**: Uses S3-compatible API for seamless integration
+
+## Setup
+
+### 1. Add dependency
+
+Add this to your `pubspec.yaml`:
+
+```yaml
+dependencies:
+  serverpod_cloud_storage_r2: ^3.0.0-alpha.1
+```
+
+### 2. Configure R2 credentials
+
+Set your Cloudflare R2 credentials as environment variables:
+
+```bash
+SERVERPOD_R2_ACCESS_KEY_ID=your_r2_access_key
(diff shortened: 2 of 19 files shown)


tylersavery:

SandPod:
Hey @tylersavery 👋 

Thanks for taking the time to contribute to Serverpod! 🎉 

The storage interface is something we are looking to rework, so we are happy for all initiatives that improve this experience. I hope that we will be able to take a look at this PR fairly soon, but we do have a couple of higher priority issues we are battling at the moment!

tylersavery:
Sounds great! Let me know if you have any q's once you have a chance to review.

tananga:
@tylersavery Thanks for your youtube video about this PR. I learned a lot how to contibute to serverpod myself. You are great🤘

SandPod:
Hey @tylersavery!

Sorry for the delay in getting back to you on this one. Things have been quite busy as we’re closing in on the first RC cut of Serverpod 3. Once that’s out, the team will finally have some development bandwidth freed up.

The main reason we haven’t done a quick review of this PR yet is that we’re still uncertain about the long-term direction for the storage integration. We’ve identified a few bugs that we suspect could all be addressed with some dedicated engineering effort, and our current thinking is to consolidate into a single storage package that supports multiple integrations, rather than maintaining separate ones.

Given that the direction is still up in the air, we didn’t feel comfortable merging and committing to maintaining another integration package just yet.

Really appreciate your patience. We’ll revisit this once we have more clarity on where the storage module is headed.

marcelomendoncasoares:
Closing this since R2 support was added with #4771.

PR #3942, opened 2025-09-01
feat: Add `@unauthenticatedClientCall` annotation option to endpoints

marcelomendoncasoares:
Adds a new annotation `@unauthenticated` that can be used on endpoint classes and methods. When calling annotated methods from the client, it will not include the auth header.

As required for issue #3674, this PR enables building JWT refresh endpoints according to the best practice: only pass the refresh token on the body, with no auth header.

If an endpoint class is annotated, all methods will be considered unauthenticated. This behavior will pass on to subclasses of the endpoint, just as we also expect subclasses of `immutable` classes to also inherit the behavior.

As a side effect, this PR also unifies the annotation parsing throughout Serverpod. Previous `doNotGenerate` (and `ignoreEndpoint`) are now parsed together with `deprecated` and `unauthenticated`.

## Pre-launch Checklist

## Breaking changes

N/A.


tools/serverpod_cli/test/generator/dart/client_code_generator/unauthenticated_annotation_test.dart
```diff
@@ -0,0 +1,278 @@
...
+      ],
+      models: [],
+    );
+
+    var codeMap = generator.generateProtocolCode(
+      protocolDefinition: protocolDefinition,
+      config: config,
+    );
+
+    test('then method calls have no authenticated parameter.', () {
```
nielsenko: Perhaps explain in description that this is because the default value is `true`. As is the test reads a bit funny/unexpected.

tools/serverpod_cli/test/generator/dart/client_code_generator/unauthenticated_annotation_test.dart
```diff
@@ -0,0 +1,278 @@
...
+'''));
+    });
+
+    test(
+        'then annotated method has not propagated the @unauthenticated annotation.',
+        () {
+      var clientCode =
+          codeMap.values.where((code) => code.contains('hello')).first;
+
+      expect(clientCode, isNot(contains('''\
```
nielsenko: Try to avoid `isNot(` as it is easy to end in a situation where you are not testing what you expect. You could write random stuff in the inner matcher and it would still pass, like `expect(clientCode, isNot(contains('random stuff')));` which mean that a tiny formatting chance will hide errors here.

nielsenko (approved):
LGTM 👍 

A few comments on the test. Try to avoid using `isNot(contains(...))` matcher, since it is brittle to changes in the generated output. Say we introduce the error they are designed to catch, but change some formatting in the same go, then these kind of tests will still pass.

Running the analyzer over the generated code to extract annotations, and check there are none (`isEmpty`) would be more robust, but perhaps overkill here.

Anyway, just food for thought.

SandPod:
> If an endpoint class is annotated, all methods will be considered unauthenticated. This behavior will pass on to subclasses of the endpoint, just as we also expect subclasses of immutable classes to also inherit the behavior.

`@unauthenticated` only influence if the client passes authentication information or not right? It does not change the behavior around if an endpoint requires authentication right?

Is there a risk that `unauthenticated` might be mixed up with an unauthenticated endpoint? Should we explore another word that could be used that is more precise in term of modifying the client behavior?

tests/serverpod_test_module/serverpod_test_module_server/lib/src/endpoints/unauthenticated.dart
```diff
@@ -0,0 +1,34 @@
+import 'package:serverpod/serverpod.dart';
+
+/// An endpoint with all methods unauthenticated.
+@unauthenticated
+class UnauthenticatedEndpoint extends Endpoint {
```
SandPod, line 5: Suggestion: Please add a test where an endpoint requires authentication but `@unauthenticated` is used. All requests to that endpoint should be rejected even if the client is authenticated normally.

SandPod (commented):
I trust @nielsenko review,
Just browsed over the PR and would like to see a scenario verified.

Since the annotation is inherited, we need to ensure that we do not drop any requirements for authentication.

tools/serverpod_cli/test/generator/dart/client_code_generator/unauthenticated_annotation_test.dart
marcelomendoncasoares: Good catch! I'll enhance the description and add quotes to the parameter name to reduce ambiguity.

tools/serverpod_cli/test/generator/dart/client_code_generator/unauthenticated_annotation_test.dart
marcelomendoncasoares: Indeed... Even changing the function name, but not the assert, would make anything pass. I'll change it to a regex-based approach that asserts the expected text.

tests/serverpod_test_module/serverpod_test_module_server/lib/src/endpoints/unauthenticated.dart
marcelomendoncasoares: Done!

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_unauthenticated_annotation_test.dart
```diff
@@ -124,4 +125,112 @@ class ExampleEndpoint extends Endpoint {
...
+
+      var error = collector.errors.first as SourceSpanSeverityException;
+      expect(error.severity, SourceSpanSeverity.info);
+      expect(
+        error.message,
+        'The endpoint class "ExampleEndpoint" overrides "requireLogin" '
+        'getter and is annotated with @unauthenticated. Be aware that this '
+        'combination may lead to all endpoint calls failing due to client '
+        'not sending a signed in user. To fix this, either remove the getter '
+        'override or remove the @unauthenticated annotation.',
```
SandPod: Question: Do we display this message even if the override is `false`? Would be good to have a test that shows this behavior in that case. This could also be quite annoying if a project always overrides these to be explicit. But on the other hand, `@unauthenticatedClientCall` is new so it is probably not widely used and we could remove this hint in that case.

marcelomendoncasoares: Unfortunately yes, because we can not know the value of the getter while analyzing. This can indeed be annoying for users doing override with `false`, but it is the tradeoff to prevent the most expected behavior of users adding the `@unauthenticatedClientCall` to endpoints that previously had `requireLogin` set to `true`. This is one of the arguments that corroborate to a shift towards annotation-based configuration instead of the `requireLogin` getter.

Regarding the test, I did not add another with `requireLogin` override to `false` because it seemed redundant with this one. Since the test description is explicitly aimed at `overriding requireLogin`, with no specification of with which return, do you think another test here is still necessary?

PR #3947, opened 2025-09-02
refactor: Make ColumnSerializable generic

exaby73:
We need generics for `ColumnSerializable` in https://github.com/serverpod/serverpod/pull/3861 to make the partial update API type-safe

## Pre-launch Checklist

## Summary by CodeRabbit

- New Features
  - Serializable database columns are now generic, improving compile-time type safety for nested and container types.
  - Optional encoder support lets you provide custom serialization for values stored in json-like columns.
  - Code generation now detects serializable columns and automatically wires encoders for record-type fields.

- Tests
  - Expanded test coverage for generic serializable columns, nested/container types, and encoder behavior.

### DIFF
--- a/tests/serverpod_test_server/lib/src/generated/types.dart
+++ b/tests/serverpod_test_server/lib/src/generated/types.dart
@@ -500,7 +500,6 @@ class TypesTable extends _i1.Table<int?> {
     aRecord = _i1.ColumnSerializable<(String, {Uri? optionalUri})>(
       'aRecord',
       this,
-      encodeFn: (value) => _i5.mapRecordToJson(value),
     );
   }
 


exaby73:
@coderabbitai full review

packages/serverpod/test/database/columns/column_serializable_test.dart
```diff
@@ -4,7 +4,10 @@ import 'package:test/test.dart';
 void main() {
   group('Given a ColumnSerializable', () {
     var columnName = 'configuration';
-    var column = ColumnSerializable(columnName, Table<int?>(tableName: 'test'));
+    var column = ColumnSerializable<String>(
```
SandPod, line 7: Suggestion: Update the `Given` description to match the generic type. `Given a ColumnSerializable containing a String`.

tests/serverpod_test_server/lib/src/generated/types.dart
```diff
@@ -485,21 +485,22 @@ class TypesTable extends _i1.Table<int?> {
...
-    aSet = _i1.ColumnSerializable(
+    aSet = _i1.ColumnSerializable<Set<int>>(
       'aSet',
       this,
     );
-    aRecord = _i1.ColumnSerializable(
+    aRecord = _i1.ColumnSerializable<(String, {Uri? optionalUri})>(
       'aRecord',
       this,
+      encodeFn: (value) => _i5.mapRecordToJson(value),
```
SandPod: Question: Why is this needed? I don't see any usages of this in the code.

If it is required to support update by id, I think this should be introduced in that PR.
Without that context, I'm not convinced this is added in the right place.

packages/serverpod/lib/src/database/concepts/columns.dart
```diff
@@ -51,14 +51,24 @@ class ColumnByteData extends Column<ByteData> {
 
 /// A [Column] holding an [SerializableModel]. The entity will be stored in the
 /// database as a json column.
-class ColumnSerializable extends Column<String> {
+class ColumnSerializable<T> extends _ValueOperatorColumn<T> {
```
SandPod: Question: By extending `_ValueOperatorColumn` we now apply encoding for all ColumnSerializable types. What is the purposer of this?

tools/serverpod_cli/lib/src/generator/types.dart
```diff
@@ -156,6 +156,8 @@ class TypeDefinition {
 
   bool get isEnumType => enumDefinition != null;
 
+  bool get isColumnSerializable => columnType == 'ColumnSerializable';
```
SandPod, line 159: Suggestion: Please add a test that validates this.

If you search for uses of `isEnumType` you can get insipration for such a test.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -2092,26 +2093,58 @@ class SerializableModelLibraryGenerator {
...
+                config: config,
+              )
+            ]
+          : [])).call(constructorArgs, {
       if (field.type.isVectorType)
         'dimension': literalNum(field.type.vectorDimension!),
       if (field.defaultPersistValue != null) 'hasDefault': literalBool(true),
+      if (encoderFunction != null) 'encodeFn': encoderFunction,
     });
   }
```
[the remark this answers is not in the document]
SandPod: Is this a concern @exaby73?

Records contained inside of other containers, for example `List<(String, String)>`?

SandPod (commented):
Question regarding the enconder function. Since we do not us it anywhere I'm not sure it should be added in this PR. Would perhaps be good to get the context on how it is used?

packages/serverpod/lib/src/database/concepts/columns.dart
exaby73: This was for adding:

```dart
  @override
  Expression _encodeValueForQuery(T value) {
    return EscapedExpression(encodeFn?.call(value) ?? value);
  }
```

Since `_encodeValueForQuery` comes from `_ValueOperatorColumn`

tests/serverpod_test_server/lib/src/generated/types.dart
exaby73: This is required because the encoder doesn't know how to serialize records. Since we have a generic now, we need to add support for serializing `T` to an object that can be encoded to either a `String` or an object that `ValueEncoder.convert` understands. Hence extending `_ValueOperatorColumn` and adding an encode function for handling records since only the generated code knows how to encode a record to a `Map` that then in turn `ValueEncoder` can convert

packages/serverpod/test/database/columns/column_serializable_test.dart
exaby73: Done

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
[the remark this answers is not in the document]
exaby73: I'm investigating this one now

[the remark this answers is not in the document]
exaby73: In the current context, this works as the model has a `toJson()` which calls `mapContainerToJson` which in turn calls `mapRecordToJson` for any records. This is an issue when we have partial updates as we are using the columns directly in that case which wouldn't use the `toJson` of the model class. I think adding `encodeFn` for collection types should be done in that PR then after rebasing it instead of this one.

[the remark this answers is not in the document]
exaby73: Though on the other hand, this can be justified to be added into this PR to keep parity with the original functionality of `ColumnSerializable` being able to support an object that `ValueEncoder.convert` knows how to convert

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_datatype_test.dart
```diff
@@ -21,14 +21,24 @@ void main() {
...
+    var vectorTypes = [
       'Vector(512)',
       'Vector(512)?',
       'HalfVector(256)',
       'HalfVector(256)?',
       'SparseVector(128)',
       'SparseVector(128)?',
+    ];
+
+    var bitTypes = [
```
SandPod: Discussion: Bit are also Vector types as far as I understand. The only difference is the value it holds.

tests/serverpod_test_server/lib/src/generated/types.dart
SandPod: Hmm, but as you mention, the encoder does not work directly on the field type. It operates based on the result of the `toJson` method. This column now contains information on how to serialize the data structure, and no other column contains that information. I believe we are patching this in the wrong place.

That is why I don't think this should be introduced in this PR, as the usecase for it is not clear.

packages/serverpod/lib/src/database/concepts/columns.dart
SandPod: Adding `EscapedExpression` would have modified how we pass in values for queries towards this column, but since we did not have any ways to query the column this doesn't really do anything. But we should be mindfull on how that type of change would modify potential values.

exaby73: Reverted

tests/serverpod_test_server/lib/src/generated/types.dart
exaby73: Reverted

PR #3957, opened 2025-09-03
feat: Add support for nested template subdirectories

nicowalter256:
- Implement recursive template loading in Templates.loadAll()
- Add _loadTemplatesRecursively() method for subdirectory scanning
- Support template paths like 'admin/dashboard' for admin/dashboard.html
- Maintain full backward compatibility with flat templates
- Add comprehensive test coverage for nested template scenarios
- Include working example with admin dashboard template
- Update TemplateWidget documentation for nested template support

This change allows developers to organize templates in logical subdirectories while maintaining the existing flat template functionality.

### DIFF
--- a/packages/serverpod/lib/src/web_server/templates.dart
+++ b/packages/serverpod/lib/src/web_server/templates.dart
@@ -10,20 +10,36 @@ final Templates templates = Templates();
 class Templates {
   final Map<String, Template> _templates = {};
 
-  /// Loads all templates from web/templates
+  /// Loads all templates from web/templates recursively
   Future<void> loadAll(Directory templateDirectory) async {
     if (!await templateDirectory.exists()) return;
 
-    for (var entity in await templateDirectory.list().toList()) {
+    await _loadTemplatesRecursively(templateDirectory, '');
+  }
+
+  /// Recursively loads templates from subdirectories
+  Future<void> _loadTemplatesRecursively(
+      Directory dir, String relativePath) async {
+    for (var entity in await dir.list().toList()) {
       if (entity is File && extension(entity.path).toLowerCase() == '.html') {
         var file = entity;
-        var name = basenameWithoutExtension(file.path);
+        var fileName = basenameWithoutExtension(file.path);
+
+        // Create template key with relative path
+        var templateKey =
+            relativePath.isEmpty ? fileName : '$relativePath/$fileName';
+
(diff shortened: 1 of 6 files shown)


packages/serverpod/lib/src/web_server/templates.dart
```diff
@@ -10,20 +10,36 @@ final Templates templates = Templates();
...
     if (!await templateDirectory.exists()) return;
 
-    for (var entity in await templateDirectory.list().toList()) {
+    await _loadTemplatesRecursively(templateDirectory, '');
+  }
+
+  /// Recursively loads templates from subdirectories
+  Future<void> _loadTemplatesRecursively(
+      Directory dir, String relativePath) async {
+    for (var entity in await dir.list().toList()) {
```
nielsenko:
```suggestion
    await for (var entity in dir.list()) {
```
why the `toList()` call?

packages/serverpod/test/relic/templates_test.dart
```diff
@@ -5,15 +5,148 @@ import 'package:serverpod/serverpod.dart';
...
+      expect(templates['base'], isNotNull);
+      expect(templates['admin/page'], isNotNull);
+      expect(templates['base']!.renderString({'content': 'Hello'}),
+          equals('<html><body>Hello</body></html>'));
+      expect(templates['admin/page']!.renderString({}),
+          equals('<h1>Admin Page</h1>'));
+    });
+
+    test(
+        'Given deeply nested structure when loading templates then all levels are accessible',
```
nielsenko: The test description doesn't really match the test as only a template at the bottom most level is tested.
Changing the description is fine.

nielsenko (commented):
Overall looks good. Thank you for taking the time to contribute to serverpod.
I have a few questions/comments I would like you to look over, but I'm sure we can land this soon.

packages/serverpod/lib/src/web_server/templates.dart
nicowalter256: I used the .toList() call  to convert the Stream<FileSystemEntity> from dir.list() into a List<FileSystemEntity> so it could be used with a regular for loop.   You're absolutely right. The .toList() call is less efficient. Using await for with the stream directly is better for memory usage. I'll make this change.

nielsenko (approved):
LGTM 🚀

PR #3965, opened 2025-09-09
fix: Fix JWT access tokens not being unique if generated in the same second

### LINKED ISSUE #3964: JWT access tokens not unique due to jti claim using refresh token ID
## Describe the bug

JWT access tokens generated within the same second can be identical when using the same refresh token, causing test flakiness and potential security concerns. This occurs because the `jti` (JWT ID) claim uses the refresh token ID instead of a unique identifier.

The JWT creation in `JwtUtil.createJwt()` sets the `jti` claim to the refresh token ID:

```dart
jwtId: refreshToken.id!.toString(),
```

When multiple JWT tokens are created from the same refresh token within the same second, they have:
- **Same `jti`**: Refresh token ID (doesn't change during rotation)
- **Same `iat`**: Issued at timestamp (identical when created within same second)
- **Same payload**: All other claims are identical

This results in completely identical JWT tokens, violating the uniqueness principle of JWTs according to RFC 7519.

## To reproduce

1. Create a JWT token using `AuthenticationTokens.createTokens()`
2. Immediately call `AuthenticationTokens.refreshAccessToken()` with the refresh token
3. Compare the original and refreshed JWT access tokens
4. Observe that they can be identical when created within the same second

This can be reproduced consistently by running the test:
```bash
flutter test test/jwt_auth_key_provider_test.dart --plain-name "then it updates auth info."
```

The test fails intermittently because it expects different tokens after refresh operations.

## Expected behavior

Each JWT access token should be unique, even when created from the same refresh token within the same timeframe. The `jti` claim should contain a unique identifier per token as specified in RFC 7519.

From test debugging, tokens created in rapid succession show identical payloads:
```
Original JWT:  eyJ...jti":"01992c94-527e-7ccd-ac14-bcfd718a0ca9"...
Refreshed JWT: eyJ...jti":"01992c94-527e-7ccd-ac14-bcfd718a0ca9"...
```

The server-side tests already acknowledge this issue by using `withClock()` to artificially advance time:
```dart
final newTokenPair = await withClock(
  // Need to move forward the clock, as otherwise the new access token has the same expiry date and thus looks equal.
  Clock.fixed(DateTime.now().add(const Duration(seconds: 2))),
  () => AuthenticationTokens.rotateRefreshToken(...)
);
```

## Library version

- Serverpod: Latest (main branch)
- Dart: 3.5.4
- Flutter: 3.24.4

## Platform information

- OS: Windows 11
- Environment: Development/Testing
- Affected: Server-side JWT generation in `serverpod_auth_core_server`

## Additional context

**Impact:**
1. **Test Flakiness**: Tests expecting different tokens after refresh operations fail intermittently
2. **Security Concerns**: Identical tokens could potentially be confused or cached incorrectly
3. **JWT Best Practices Violation**: The `jti` claim should be unique per token

**Proposed Solution:**
Replace the refresh token ID with a unique UUID for the `jti` claim:

```dart
// Before
jwtId: refreshToken.id!.toString(),

// After  
jwtId: const Uuid().v4obj().toString(),
```

**Breaking Changes Assessment:**
Code that depends on `jti` being the refresh token ID may break:
1. `verifyJwt()` extracts `refreshTokenId` from `jti` - **BREAKING**
2. `AuthenticationInfo.authId` uses `refreshTokenId` from JWT - **BREAKING** 
3. Tests expect `jti` to match refresh token ID - **BREAKING**

**Files Affected:**
- `modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/jwt_util.dart`
- `modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/authentication_info_from_jwt.dart`
- Various test files that depend on `jti` being refresh token ID

marcelomendoncasoares:
Fixes #3964.

## Pre-launch Checklist

## Breaking changes

N/A.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/jwt/integration/authentication_tokens_test.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/jwt/integration/authentication_tokens_test.dart
@@ -134,17 +134,39 @@ void main() {
     test(
         'when rotating the tokens, then a new refresh and access token is returned.',
         () async {
-      final newTokenPair = await withClock(
-        // Need to more forward the clock, as otherwise the new access token has the same expiry date and thus looks equal.
-        Clock.fixed(DateTime.now().add(const Duration(seconds: 2))),
-        () => AuthenticationTokens.rotateRefreshToken(
-          session,
-          refreshToken: authSuccess.refreshToken!,
-        ),
+      final newTokenPair = await AuthenticationTokens.rotateRefreshToken(
+        session,
+        refreshToken: authSuccess.refreshToken!,
       );
 
       expect(newTokenPair.accessToken, isNot(authSuccess.token));
       expect(newTokenPair.refreshToken, isNot(authSuccess.refreshToken));
+    });
+
+    test(
+        'when rotating multiple new tokens in succession, then new tokens are returned.',
+        () async {
+      // Pin the clock to ensure all tokens are created at the same time.
+      final newTokenPairs = await withClock(
+          Clock.fixed(DateTime.now()),
+          () => Future.wait(
+                List.generate(
+                  3,
+                  (final _) => AuthenticationTokens.rotateRefreshToken(
+                    session,
+                    refreshToken: authSuccess.refreshToken!,
+                  ),
+                ),
+              ));
+
+      final tokens = newTokenPairs.map((final t) => t.accessToken).toSet();
+      expect(tokens, hasLength(3));
+      expect(tokens.add(authSuccess.token), isTrue);
+
+      final refreshTokens =
+          newTokenPairs.map((final t) => t.refreshToken).toSet();
+      expect(refreshTokens, hasLength(3));
+      expect(refreshTokens.add(authSuccess.refreshToken!), isTrue);
     });
 
     test(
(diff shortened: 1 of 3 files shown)


modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/jwt/integration/authentication_tokens_test.dart
```diff
@@ -134,19 +134,41 @@ void main() {
...
       );
 
       expect(newTokenPair.accessToken, isNot(authSuccess.token));
       expect(newTokenPair.refreshToken, isNot(authSuccess.refreshToken));
     });
 
+    test(
+        'when rotating multiple new tokens in succession, then new tokens are returned.',
+        () async {
+      // Pin the clock to ensure all tokens are created at the same time.
```
SandPod: Suggestion: Please add this to the test description as this is critical for this test and should be front and center if it fails. Even in CI.

SandPod (approved):
LGTM!

Please update the description to motivate why `jti` is the right approach instead of `nonce`.

As I understand it `nonce` is more of a one time challange used by openID connect to prevent replay attacks whereas the `jti` clam should gurantee to be unique (or at leas have a negligable chance of colliding).

Just had one suggestion regarding a test description.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/jwt/integration/authentication_tokens_test.dart
marcelomendoncasoares: Both PR description and test description updated!

PR #3966, opened 2025-09-09
feat: Add a `MutexRefresherClientAuthKeyProvider` to avoid concurrent refreshes

marcelomendoncasoares:
This work is part of the solution to #3674.

As a follow-up of #3927, this PR adds a `RefresherAuthKeyProvider` that uses a mutex lock to avoid concurrent refresh attempts. It also prevents subsequent refreshes when a refresh operation fails with unauthorized for a given key - such as when a refresh token is invalid.

## Pre-launch Checklist

## Breaking changes

The `refreshAuthKey` signature has changed, but it is still unused in the codebase and it is not expected to already be used externally.

### DIFF
--- a/packages/serverpod_client/lib/src/auth_key_provider.dart
+++ b/packages/serverpod_client/lib/src/auth_key_provider.dart
@@ -11,11 +11,27 @@ abstract interface class ClientAuthKeyProvider {
 /// Provides the authentication key for the client, with a method to refresh it.
 abstract interface class RefresherClientAuthKeyProvider
     implements ClientAuthKeyProvider {
-  /// Refreshes the authentication key. If the refresh is successful, should
-  /// return true to retry requests that failed due to authentication errors.
-  /// Be sure to annotate the refresh endpoint with @unauthenticatedClientCall
-  /// to avoid a deadlock on the [authHeaderValue] getter on refresh call.
-  Future<bool> refreshAuthKey();
+  /// Refreshes the authentication key and returns the result of the operation.
+  /// If the refresh is successful, should return [RefreshAuthKeyResult.success]
+  /// to retry requests that failed due to authentication errors. Be sure to
+  /// annotate the refresh endpoint with @unauthenticatedClientCall to avoid a
+  /// deadlock on the [authHeaderValue] getter on a refresh call.
+  Future<RefreshAuthKeyResult> refreshAuthKey();
+}
+
+/// Represents the result of an authentication key refresh operation.
+enum RefreshAuthKeyResult {
+  /// Refresh was skipped because the key is not expiring.
+  skipped,
+
+  /// Refresh was successful and a new key was obtained.
+  success,
+
+  /// Refresh failed due to invalid refresh credentials (such as expired token).
+  failedUnauthorized,
+
+  /// Refresh failed due to other reasons (network, server error, etc.).
+  failedOther,
 }
 
 /// A [RefresherClientAuthKeyProvider] decorator that adds a mutex lock to
@@ -29,7 +45,7 @@ class MutexRefresherClientAuthKeyProvider
   MutexRefresherClientAuthKeyProvider(this._delegate);
 
   /// Shared future that serves as a lock to prevent concurrent refresh calls.
-  Future<bool>? _pendingRefresh;
+  Future<RefreshAuthKeyResult>? _pendingRefresh;
 
   @override
   Future<String?> get authHeaderValue async {
@@ -39,7 +55,7 @@ class MutexRefresherClientAuthKeyProvider
 
   /// Refreshes the authentication key with locking to prevent concurrent calls.
   @override
-  Future<bool> refreshAuthKey() async {
+  Future<RefreshAuthKeyResult> refreshAuthKey() async {
     final pendingRefresh = _pendingRefresh;
     if (pendingRefresh != null) return pendingRefresh;
 
--- a/packages/serverpod_client/lib/src/auth_key_provider.dart
+++ b/packages/serverpod_client/lib/src/auth_key_provider.dart
@@ -59,7 +59,7 @@ class MutexRefresherClientAuthKeyProvider
     final pendingRefresh = _pendingRefresh;
     if (pendingRefresh != null) return pendingRefresh;
 
-    final refreshFuture = _delegate.refreshAuthKey();
+    final refreshFuture = _refreshAuthKey();
     _pendingRefresh = refreshFuture;
 
     try {
@@ -69,5 +69,27 @@ class MutexRefresherClientAuthKeyProvider
         _pendingRefresh = null;
       }
     }
+  }
+
+  String? _lastAuthHeaderValue;
+  RefreshAuthKeyResult? _lastRefreshResult;
+
+  /// Actual refresh operation that is protected by the mutex lock. Will not
+  /// perform a refresh if the last refresh failed due to invalid credentials
+  /// and the auth header value has not changed. This prevents further request
+  /// failures due to invalid credentials when refresh credential is already
+  /// proven invalid.
+  Future<RefreshAuthKeyResult> _refreshAuthKey() async {
+    final currentAuthHeaderValue = await _delegate.authHeaderValue;
+    if (_lastAuthHeaderValue != null &&
+        _lastAuthHeaderValue == currentAuthHeaderValue &&
+        _lastRefreshResult == RefreshAuthKeyResult.failedUnauthorized) {
+      return RefreshAuthKeyResult.failedUnauthorized;
+    }
+
+    final result = await _delegate.refreshAuthKey();
+    _lastAuthHeaderValue = await _delegate.authHeaderValue;
+    _lastRefreshResult = result;
+    return result;
   }
 }
--- a/packages/serverpod_client/lib/src/auth_key_provider.dart
+++ b/packages/serverpod_client/lib/src/auth_key_provider.dart
@@ -9,28 +9,49 @@ abstract interface class ClientAuthKeyProvider {
 }
 
 /// Provides the authentication key for the client, with a method to refresh it.
-abstract class RefresherClientAuthKeyProvider implements ClientAuthKeyProvider {
-  /// Authentication header value getter that must be implemented. This getter
-  /// must get the header value directly from the source, without any locking.
-  /// The [authHeaderValue] getter will handle refreshing the key if needed,
-  /// with a lock to avoid concurrent refresh attempts.
-  Future<String?> get notLockedAuthHeaderValue;
-
-  @override
-  Future<String?> get authHeaderValue async {
-    // TODO: In the beginning of the getter, check validity and refresh if it is
-    // about to expire. Either way, before refreshing, check again if it's still
-    // about to expire to avoid racing conditions when multiple endpoints try to
-    // refresh and one has already refreshed. Then skip refresh and return true.
-
-    // TODO: Use Completer to implement a lock under the future of the getter to
-    // avoid multiple concurrent refresh attempts.
-    return await notLockedAuthHeaderValue;
-  }
-
+abstract interface class RefresherClientAuthKeyProvider
+    implements ClientAuthKeyProvider {
   /// Refreshes the authentication key. If the refresh is successful, should
   /// return true to retry requests that failed due to authentication errors.
   /// Be sure to annotate the refresh endpoint with @unauthenticatedClientCall
   /// to avoid a deadlock on the [authHeaderValue] getter on refresh call.
   Future<bool> refreshAuthKey();
+}
+
+/// A [RefresherClientAuthKeyProvider] decorator that adds a mutex lock to
+/// prevent concurrent refresh calls. Actual auth header getter and refresh
+/// logic is delegated to the [_delegate] provider.
+class MutexRefresherClientAuthKeyProvider
+    implements RefresherClientAuthKeyProvider {
+  final RefresherClientAuthKeyProvider _delegate;
+
+  /// Creates a new [MutexRefresherClientAuthKeyProvider].
+  MutexRefresherClientAuthKeyProvider(this._delegate);
+
+  /// Shared future that serves as a lock to prevent concurrent refresh calls.
+  Future<bool>? _pendingRefresh;
+
+  @override
+  Future<String?> get authHeaderValue async {
+    await refreshAuthKey();
+    return _delegate.authHeaderValue;
+  }
+
+  /// Refreshes the authentication key with locking to prevent concurrent calls.
+  @override
+  Future<bool> refreshAuthKey() async {
+    final pendingRefresh = _pendingRefresh;
+    if (pendingRefresh != null) return pendingRefresh;
+
+    final refreshFuture = _delegate.refreshAuthKey();
+    _pendingRefresh = refreshFuture;
+
+    try {
+      return await refreshFuture;
+    } finally {
+      if (identical(_pendingRefresh, refreshFuture)) {
+        _pendingRefresh = null;
+      }
+    }
+  }
 }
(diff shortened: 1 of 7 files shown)


packages/serverpod_client/lib/src/auth_key_provider.dart
```diff
@@ -9,28 +9,49 @@ abstract interface class ClientAuthKeyProvider {
...
+/// logic is delegated to the [_delegate] provider.
+class MutexRefresherClientAuthKeyProvider
+    implements RefresherClientAuthKeyProvider {
+  final RefresherClientAuthKeyProvider _delegate;
+
+  /// Creates a new [MutexRefresherClientAuthKeyProvider].
+  MutexRefresherClientAuthKeyProvider(this._delegate);
+
+  /// Shared future that serves as a lock to prevent concurrent refresh calls.
+  Future<bool>? _pendingRefresh;
```
SandPod: Question: How does this behave when running the client call in an isolate? Is this an option for our client to begin with? I think it already has quite a lot if internal state.

Perhaps we should run the indiviudal calls inside of client in isolates instead to remove that burden from the user.

packages/serverpod_client/lib/src/auth_key_provider.dart
```diff
@@ -9,28 +9,49 @@ abstract interface class ClientAuthKeyProvider {
...
+
+  @override
+  Future<String?> get authHeaderValue async {
+    await refreshAuthKey();
```
SandPod, line 57: Question: Should this really trigger a `refreshAuthKey` call? Shouldn't this simply wait on the `_pendingRefresh` if there is one ongoing?

packages/serverpod_client/lib/src/auth_key_provider.dart
```diff
@@ -9,28 +9,49 @@ abstract interface class ClientAuthKeyProvider {
...
+    final pendingRefresh = _pendingRefresh;
+    if (pendingRefresh != null) return pendingRefresh;
+
+    final refreshFuture = _delegate.refreshAuthKey();
+    _pendingRefresh = refreshFuture;
+
+    try {
+      return await refreshFuture;
+    } finally {
+      if (identical(_pendingRefresh, refreshFuture)) {
```
SandPod: Suggestion: Throw state error on `!identical(...)`.

If this check fails that means our mutex it now working. We should probably throw a state error in that case since we might be messing up more things. Better to make us informed if that is the case.

packages/serverpod_client/test/mutex_refresher_client_auth_key_provider_test.dart
```diff
@@ -0,0 +1,146 @@
...
+
+    test('when refresh succeeds then returns new auth header value.', () async {
+      delegate.setRefresh(() {
+        delegate.updateAuthKey();
+        return true;
+      });
+
+      final result = await provider.authHeaderValue;
+
+      expect(result, 'refreshed-token-1');
```
SandPod: Question: Where does this string come from? Would be good if the new auth header was defined in the test as well.

And do we really need to update it? Shouldn't we just validate that regardless if it fails or succeeds, the header is read from the delegate?

packages/serverpod_client/test/mutex_refresher_client_auth_key_provider_test.dart
```diff
@@ -0,0 +1,146 @@
...
+
+      final futures = List.generate(3, (_) => provider.refreshAuthKey());
+      final results = await Future.wait(futures);
+
+      expect(results, everyElement(false));
+      expect(delegate.refreshCallCount, 1);
+    });
+
+    test(
+        'when refreshing throws an exception '
```
SandPod, line 174: Question: Is this true even when multiple calls are made concurrently?

packages/serverpod_client/lib/src/auth_key_provider.dart
```diff
@@ -70,4 +70,26 @@ class MutexRefresherClientAuthKeyProvider
...
+
+  String? _lastAuthHeaderValue;
+  RefreshAuthKeyResult? _lastRefreshResult;
+
+  /// Actual refresh operation that is protected by the mutex lock. Will not
+  /// perform a refresh if the last refresh failed due to invalid credentials
+  /// and the auth header value has not changed. This prevents further request
+  /// failures due to invalid credentials when refresh credential is already
+  /// proven invalid.
+  Future<RefreshAuthKeyResult> _refreshAuthKey() async {
```
SandPod, line 88: Question: How will this work if we initiate multiple clients towards the same server? Are there any limitations there that we need to take into consideration? I guess the `Session Manager` needs to be shared. Perhaps we should have documentaiton around that.

SandPod (commented):
Changes look good overall.

Have a couple of questions regarding concurrency, isolates and multiple client instances. Please have a look and let me know if anything is unclear.

packages/serverpod_client/lib/src/auth_key_provider.dart
marcelomendoncasoares: This call is here to allow delegates to refresh eagerly - checking if the key is about to expire and refresh before it generates any unauthorized. The delegate would implement the logic to either perform the actual refresh call or just skip it inside the `refreshAuthKey`. See https://github.com/marcelomendoncasoares/serverpod/commit/542ea586c80dd09c264ac860154587732bab6001 for an example of this implementation.

Given your question, if there is no other consideration on this approach, I can add a doc to the getter to make this expectation explicit.

packages/serverpod_client/test/mutex_refresher_client_auth_key_provider_test.dart
marcelomendoncasoares: Yes, it will throw the exception on each call. I'll add a test to cover this use case as well.

packages/serverpod_client/lib/src/auth_key_provider.dart
marcelomendoncasoares: I don't think we have limitations, and maybe the best approach is indeed to share the session manager. The only consideration in this case is that all calls to refresh will be done from a single one of the clients - the one configured on the session manager. No concurrent refreshes would happen, which is what we care more.

We can cover this shared session manager setup in a test and recommend on the docs for similar cases.

packages/serverpod_client/test/mutex_refresher_client_auth_key_provider_test.dart
marcelomendoncasoares: This was being set on the `TestRefresherAuthKeyProvider.updateAuthKey` method to ensure each refresh key had the `refreshCallCount` appended. But it is somewhat redundant with the `refreshCallCount` check, so I'll remove those.

About the need to update the value, I think we can remove it from the tests that are not asserting the return of the `authKeyHeader`, but keep them when we assert that the provider is correctly forwarding header updates. Otherwise, we could miss poorly implemented cache variables being set once, but not updated after refresh.

packages/serverpod_client/lib/src/auth_key_provider.dart
marcelomendoncasoares: I don't think running the client in an isolate is much of an option with our client, since it has other stateful variables. To name a few, from the `ServerpodClientShared` class:

```dart
abstract class ServerpodClientShared extends EndpointCaller {
  /// Full url to the Serverpod server. E.g. "https://example.com/"
  final String host;

  late final ServerpodClientRequestDelegate _requestDelegate;

  WebSocketChannel? _webSocket;

  Timer? _connectionTimer;

  final List<VoidCallback> _websocketConnectionStatusListeners = [];
```

Regarding the delegation of the client calls to isolates, which would be the major concern to address? I might be missing something here, but wouldn't it create a lot of overhead due to spinning isolares + serialization/deserialization for operations that are I/O-bound (server requests)?

packages/serverpod_client/lib/src/auth_key_provider.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/3966/commits/8250b143b65f7607e46c51667253975b8bc2303e.

packages/serverpod_client/test/mutex_refresher_client_auth_key_provider_test.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/3966/commits/1f2f3dadec1410f3c63df343aba8922cbba16020.

But it seems that we hit something unexpected in WASM:
```
+29 ~4 -1: [Chrome, Dart2Wasm] test/mutex_refresher_client_auth_key_provider_test.dart: Given a mutex protected auth key provider with an initial token when multiple authHeaderValue calls are made concurrently and refresh throws an exception then authHeaderValue rethrows the exception for all calls. [E]
  Exception: Refresh failed
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:542476  module0.main closure at file:///home/runner/work/serverpod/serverpod/packages/serverpod_client/test/mutex_refresher_client_auth_key_provider_test.dart:206:27
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:581947  module0.closure wrapper at file:///home/runner/work/serverpod/serverpod/packages/serverpod_client/test/mutex_refresher_client_auth_key_provider_test.dart:206:27 trampoline
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:581256  module0.TestRefresherAuthKeyProvider.refreshAuthKey inner
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:581079  module0.TestRefresherAuthKeyProvider.refreshAuthKey
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:580751  module0.MutexRefresherClientAuthKeyProvider._refreshAuthKey inner
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:486341  module0._awaitHelper closure at org-dartlang-sdk:///lib/_internal/wasm/lib/async_patch.dart:104:5
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:486470  module0.closure wrapper at org-dartlang-sdk:///lib/_internal/wasm/lib/async_patch.dart:104:5 trampoline
  
  Exception: Refresh failed
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:542476  module0.main closure at file:///home/runner/work/serverpod/serverpod/packages/serverpod_client/test/mutex_refresher_client_auth_key_provider_test.dart:206:27
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:581947  module0.closure wrapper at file:///home/runner/work/serverpod/serverpod/packages/serverpod_client/test/mutex_refresher_client_auth_key_provider_test.dart:206:27 trampoline
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:581256  module0.TestRefresherAuthKeyProvider.refreshAuthKey inner
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:581079  module0.TestRefresherAuthKeyProvider.refreshAuthKey
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:580751  module0.MutexRefresherClientAuthKeyProvider._refreshAuthKey inner
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:486341  module0._awaitHelper closure at org-dartlang-sdk:///lib/_internal/wasm/lib/async_patch.dart:104:5
  mutex_refresher_client_auth_key_provider_test.dart.browser_test.dart.wasm 1:486470  module0.closure wrapper at org-dartlang-sdk:///lib/_internal/wasm/lib/async_patch.dart:104:5 trampoline
```

I don't know the best approach for this one... Because it should be the same as the `refreshAuthKey` test that passed, with the difference that in this test the exception is behind one more layer of future. Should we mark it as expected to fail on WASM?

packages/serverpod_client/lib/src/auth_key_provider.dart
marcelomendoncasoares: Test added at https://github.com/serverpod/serverpod/pull/3966/commits/8174f47973264cf767cce59d9b1d8b9dc3991c1b.

packages/serverpod_client/test/mutex_refresher_client_auth_key_provider_test.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/3966/commits/78bf82c4b9755d4b92aaee75844fcb252bae7db7.

packages/serverpod_client/lib/src/auth_key_provider.dart
SandPod: The call in itself is not a big concern. But I do think that serialization/deserialization would be beneficial to run in an isolate. Especially for large objects. However, that is separate from the client's operations.

packages/serverpod_client/lib/src/auth_key_provider.dart
SandPod: Aah got it! This was aslo made clear in later commits.

Yes, it would be good to clarify the expectation with a doc comment. I think

packages/serverpod_client/lib/src/auth_key_provider.dart
SandPod: The scenario that concerns me with multiple clients is this:

1. Both clients call the server and are rejected, which initializes a refresh
2. Client A initiates its refresh first, creating a new token.
3. Client B initiates its refresh, now with an outdated token that would revoke any newly created token.

Now, no valid token is available.

But perhaps this is not an issue since the `Session Manager` or `authKeyProvider` is shared and that is where we do our syncronization?

packages/serverpod_client/test/mutex_refresher_client_auth_key_provider_test.dart
SandPod: I don't understand why this is failing either. The test looks straightforward.

The only difference is that we are using a `getter` instead of calling a method on the class. Are we not capturing the exception correctly?

Can we reproduce this in an isolated example? If so we could create an issue with the Dart team and reference that issue in the `skip` for WASM.

But this is concerning, as this might not behave correctly in WASM.

SandPod: The test framework isn't really telling us that the expectation failed; it is more telling us that the `Exception` could not be captured.

packages/serverpod_client/lib/src/auth_key_provider.dart
marcelomendoncasoares: This was my thought exactly. I started adding tests for the actual authentication and refresh with the shared session, but it was getting redundant with the already existing tests, since we have ensured that the session manager is the same. So I left only this test asserting that the session is shared. Then each client will behave the same as in the other tests.

packages/serverpod_client/test/mutex_refresher_client_auth_key_provider_test.dart
marcelomendoncasoares: I'll try to reproduce this in isolation so we can report to the Dart team. It seems a lot like a bug in WASM, since the code is correct and it does not fail for other platforms.

packages/serverpod_client/lib/src/auth_key_provider.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/3966/commits/f5c6ffd0cc26511938752227db95f3c0c02e6964.

packages/serverpod_client/test/mutex_refresher_client_auth_key_provider_test.dart
marcelomendoncasoares: Issue reported at https://github.com/dart-lang/sdk/issues/61483 and mentioned in the skip message for the test at https://github.com/serverpod/serverpod/pull/3966/commits/936b09039b6f63d24245aa60d20207d9b280674a.

PR #3969, opened 2025-09-10
feat: implement automatic static file cache busting

nicowalter256:
- Add StaticAssetManager for on-demand hash computation and caching
- Add StaticAssetsConfig for server startup configuration
- Add TemplateHelpers for template integration
- Enhance RouteStaticDirectory with automatic hash stripping and fallback
- Support multiple hash patterns (content, Flutter, version, timestamp)
- Add CDN URL support for static assets
- Include comprehensive test suite with 6 passing tests
- Add example usage documentation

Features:
- On-demand content hashing with SHA-256
- Automatic cache invalidation when files change
- Memory-efficient caching with LRU cleanup
- Robust hash pattern recognition
- Fallback logic for file serving
- Template integration with assetUrl() helper
- CDN support for production deployments


packages/serverpod/lib/src/web_server/static_asset_manager.dart
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  /// Simple SHA-256-like hash implementation.
+  String _sha256(List<int> bytes) {
+    var hash = 0;
+    for (var byte in bytes) {
+      hash = ((hash << 5) - hash + byte) & 0xffffffff;
+    }
+    return hash.abs().toRadixString(16).padLeft(8, '0');
+  }
```
nielsenko: Replace with SHA256 from [crypto](https://pub.dev/packages/crypto) package.

packages/serverpod/lib/src/web_server/template_helpers.dart
```diff
@@ -0,0 +1,36 @@
...
+
+  /// Synchronous version of assetUrl for use in templates.
+  /// Note: This will return the asset path without hash if the file
+  /// hasn't been processed yet. Use the async version when possible.
+  static String assetUrlSync(String assetPath) {
+    // For synchronous templates, we can't compute hashes on-demand
+    // Return the original path - the static directory route will handle
+    // hash removal if needed
+    return '/$assetPath';
+  }
```
nielsenko: Can we get rid of this? Has weird semantics.

packages/serverpod/lib/src/web_server/template_helpers.dart
```diff
@@ -0,0 +1,36 @@
...
+  ///
+  /// The function will:
+  /// 1. Compute a hash of the file content
+  /// 2. Append the hash to the filename
+  /// 3. Return the versioned URL
+  ///
+  /// If the file doesn't exist, returns null.
+  static Future<String?> assetUrl(String assetPath) async {
+    return await StaticAssetManager().assetUrl(assetPath);
+  }
```
nielsenko: Why do we need a separate class for this?

packages/serverpod/lib/src/web_server/static_asset_manager.dart
```diff
@@ -0,0 +1,204 @@
...
+/// - Invalidate cache when files change
+/// - Support CDN URLs
+class StaticAssetManager {
+  static final StaticAssetManager _instance = StaticAssetManager._internal();
+
+  /// Creates a new instance of [StaticAssetManager].
+  /// Returns the singleton instance.
+  factory StaticAssetManager() => _instance;
+
+  StaticAssetManager._internal();
```
nielsenko:
```suggestion
  StaticAssetManager._();
```

packages/serverpod/lib/src/web_server/static_asset_manager.dart
```diff
@@ -0,0 +1,204 @@
...
+/// - Generate versioned URLs for cache busting
+/// - Cache hash results for performance
+/// - Invalidate cache when files change
+/// - Support CDN URLs
+class StaticAssetManager {
+  static final StaticAssetManager _instance = StaticAssetManager._internal();
+
+  /// Creates a new instance of [StaticAssetManager].
+  /// Returns the singleton instance.
+  factory StaticAssetManager() => _instance;
```
nielsenko: Not fan of this singleton pattern. Global state makes testing difficult.

packages/serverpod/lib/src/web_server/static_asset_manager.dart
```diff
@@ -0,0 +1,204 @@
...
+
+  StaticAssetManager._internal();
+
+  /// Cache of file hashes: file path -> (hash, last modified)
+  final Map<String, _HashCacheEntry> _hashCache = {};
```
nielsenko: Make it explicit that this is a `LinkedHashMap`.

packages/serverpod/lib/src/web_server/static_asset_manager.dart
```diff
@@ -0,0 +1,204 @@
...
+      }
+
+      // Compute new hash
+      var hash = await _computeFileHash(file);
+      if (hash == null) {
+        return null;
+      }
+
+      // Cache the result
+      _hashCache[filePath] = _HashCacheEntry(hash, lastModified);
```
nielsenko: Have you considered removing then adding, so it goes to the end of your `_hashCache` which is a `LinkedHashMap`?  This will make `_cleanupCache` simpler.

packages/serverpod/lib/src/web_server/static_asset_manager.dart
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  /// Clean up the hash cache to prevent memory leaks.
+  void _cleanupCache() {
+    if (_hashCache.length <= _maxCacheEntries) return;
+
+    // Remove oldest entries (simple LRU-like behavior)
+    var entries = _hashCache.entries.toList();
+    entries
+        .sort((a, b) => a.value.lastModified.compareTo(b.value.lastModified));
```
nielsenko: If you kept the `LinkedHashMap` ordered by access, then you could avoid this sort.

packages/serverpod/lib/src/web_server/static_asset_manager.dart
```diff
@@ -0,0 +1,204 @@
...
+    } catch (e) {
+      return null;
+    }
+  }
+
+  /// Compute the content hash of a file.
+  Future<String?> _computeFileHash(File file) async {
+    try {
+      var bytes = await file.readAsBytes();
+      var hash = _sha256(bytes);
```
nielsenko: Don't read full file into memory. Use streaming.

examples/static_assets_example/web/templates/index.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Static Assets Example</title>
+    <!-- Cache-busted CSS -->
+    <link rel="stylesheet" href="{{ assetUrl('css/style.css') }}">
```
nielsenko: How is this supposed to work? How will the template engine call out to `assetUrl`?

BTW, this example project is missing `pubspec.yaml`

packages/serverpod/lib/src/web_server/routes/static_directory.dart
```diff
@@ -130,36 +193,57 @@ class RouteStaticDirectory extends Route {
...
-              ? CacheControlHeader(
-                  maxAge: 0, sMaxAge: 0, noCache: true, noStore: true)
-              : CacheControlHeader(maxAge: pathCacheMaxAge.inSeconds));
-
-      final body = Body.fromDataStream(
-        file.openRead().cast(),
-        mimeType: mimeType,
-      );
-      return context.withResponse(Response.ok(body: body, headers: headers));
+      return await _serveFile(file, path, context);
```
nielsenko, line 214: So, if a request arrives for a file with a specific hash, we may potentially return a file that would hash differently?! Is that intended?

We strip hash from requested file. Check if we can find it on disk (regardless of content). Then return whatever is in there, if found? Only if we don't find it we try to look for the hashed path (`originalFilePath`).

packages/serverpod/lib/src/web_server/static_asset_manager.dart
nielsenko: Actually this is a weird LRU. You store `lastModified` from the `file.stat()` call. Then you clean out the oldest files from `_hashCache`, but these may well be the hottest in the cache!? The correct approach would be to track access time in the cache.

Or just use the fact that `LinkedHashMap` maintains insert order.

vlidholt:
A thought on this. @nielsenko, perhaps this could mostly be done in the Relic package and the static file handler? As I understand it, does it already create hashes to be used in the headers? When looking up a path there, it could optionally remove the hash part of the name, and there could be a method that retrieves the hash for a file. The static file handler may even have some awareness of which path it is mounted under?

examples/static_assets_example/web/templates/index.html
vlidholt: Also, does this work if the assets are mounted under a custom path?

packages/serverpod/lib/src/web_server/static_asset_manager.dart
nicowalter256: I've refactored the cache to use LinkedHashMap's insertion order for true LRU behavior

packages/serverpod/lib/src/web_server/static_asset_manager.dart
nicowalter256: Great insight

packages/serverpod/lib/src/web_server/static_asset_manager.dart
nicowalter256: okay

SandPod:
@nicowalter256 - Thanks for the swift work of drafting up an implementation on this.

However, after reviewing the code, we have realized that we need to rethink the specification of this feature before it can be implemented.

I suggest we close this PR, and I'll direct you towards a more refined feature you could support us with.

Does that sound OK with you?

nicowalter256:
@SandPod  Thanks for the feedback. That sounds good to me. Please go ahead and point me towards the refined feature—I’d be happy to contribute there.

SandPod:
I have reached out to you in a separate  issue.

Closing this with the motivation above.

PR #3978, opened 2025-09-13
Fix/create migration exit code

nicowalter256:
## Description

This PR adds two new flags to the `serverpod create-migration` command to address CI/CD pipeline compatibility issues and provide advanced migration creation capabilities.

### Changes Made

**Added `--check` flag (`-c`):**
- Checks if there are pending migrations without creating any files
- Returns exit code `0` when no changes are detected (CI/CD friendly)
- Returns exit code `1` when changes exist
- Designed for CI/CD pipelines that need to verify if migrations are pending

**Added `--empty` flag (`-e`):**
- Creates an empty migration without evaluating the current project state
- Bypasses model analysis and database state evaluation
- Useful for advanced use cases requiring manual migration creation
- Always creates a migration regardless of project changes

### Implementation Details

- Enhanced `CreateMigrationCommand` with new option definitions
- Added `createEmptyMigration()` method to `MigrationGenerator` class
- Updated command logic to handle both check and empty modes
- Maintained full backward compatibility with existing behavior
- Added proper help text and validation for flag combinations

### Usage Examples

```bash
# CI/CD friendly check (returns 0 if no changes)
serverpod create-migration --check

# Create empty migration for manual editing
serverpod create-migration --empty

# Create empty migration with tag
serverpod create-migration --empty --tag manual-fix
```

## Issues Fixed

- Fixes #XXXX - `serverpod create-migration` command returns non-zero exit code even when no changes are detected, breaking CI/CD pipelines

#
## Breaking changes

No breaking changes. This PR maintains full backward compatibility with existing `serverpod create-migration` behavior while adding new optional flags.

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/create_migration.dart
+++ b/tools/serverpod_cli/lib/src/commands/create_migration.dart
@@ -13,8 +13,8 @@ import 'package:serverpod_shared/serverpod_shared.dart' hide ExitException;
 enum CreateMigrationOption<V> implements OptionDefinition<V> {
   force(CreateMigrationCommand.forceOption),
   tag(CreateMigrationCommand.tagOption),
-  empty(CreateMigrationCommand.emptyOption),
-  check(CreateMigrationCommand.checkOption);
+  check(CreateMigrationCommand.checkOption),
+  empty(CreateMigrationCommand.emptyOption);
 
   const CreateMigrationOption(this.option);
 
@@ -40,6 +40,15 @@ class CreateMigrationCommand extends ServerpodCommand<CreateMigrationOption> {
     customValidator: _validateTag,
   );
 
+  static const checkOption = FlagOption(
+    argName: 'check',
+    argAbbrev: 'c',
+    negatable: false,
+    defaultsTo: false,
+    helpText: 'Check if there are changes without creating a migration. '
+        'Returns exit code 0 if no changes detected, 1 if changes exist.',
+  );
+
   static const emptyOption = FlagOption(
     argName: 'empty',
     argAbbrev: 'e',
@@ -48,16 +57,6 @@ class CreateMigrationCommand extends ServerpodCommand<CreateMigrationOption> {
     helpText:
         'Creates an empty migration without evaluating the state of the project. '
         'Advanced use case for manual migration creation.',
-  );
-
-  static const checkOption = FlagOption(
-    argName: 'check',
-    argAbbrev: 'c',
-    negatable: false,
-    defaultsTo: false,
-    helpText:
-        'Returns with exit code 0 if no changes have been detected. '
-        'Useful for CI/CD pipelines to check if migrations are needed.',
   );
 
   static void _validateTag(String tag) {
@@ -83,25 +82,8 @@ class CreateMigrationCommand extends ServerpodCommand<CreateMigrationOption> {
   ) async {
     bool force = commandConfig.value(CreateMigrationOption.force);
     String? tag = commandConfig.optionalValue(CreateMigrationOption.tag);
-    bool empty = commandConfig.value(CreateMigrationOption.empty);
     bool check = commandConfig.value(CreateMigrationOption.check);
-
-    // Validate flag combinations
-    if (empty && check) {
-      log.error(
-        'The --empty and --check flags cannot be used together. '
-        'Use --empty to create an empty migration, or --check to verify if changes are needed.',
-      );
-      throw ExitException(ServerpodCommand.commandInvokedCannotExecute);
-    }
-
-    if (empty && force) {
-      log.error(
-        'The --empty and --force flags cannot be used together. '
-        'The --empty flag already creates a migration without evaluation.',
-      );
-      throw ExitException(ServerpodCommand.commandInvokedCannotExecute);
-    }
+    bool empty = commandConfig.value(CreateMigrationOption.empty);
 
     GeneratorConfig config;
     try {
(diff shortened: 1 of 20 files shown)


tools/serverpod_cli/lib/src/commands/create_migration.dart
```diff
@@ -88,13 +111,25 @@ class CreateMigrationCommand extends ServerpodCommand<CreateMigrationOption> {
     );
 
     MigrationVersion? migration;
-    await log.progress('Creating migration', () async {
+    await log.progress(
+        check
+            ? 'Checking for changes'
+            : (empty ? 'Creating empty migration' : 'Creating migration'),
```
marcelomendoncasoares: May be worth separating to a variable to improve readability by avoiding nested ternaries.

tools/serverpod_cli/lib/src/commands/create_migration.dart
```diff
@@ -88,13 +111,25 @@ class CreateMigrationCommand extends ServerpodCommand<CreateMigrationOption> {
...
       try {
-        migration = await generator.createMigration(
-          tag: tag,
-          force: force,
-          config: config,
-        );
+        if (empty) {
+          migration = await generator.createEmptyMigration(
+            tag: tag,
+            write: !check, // Don't write files in check mode
```
marcelomendoncasoares: Does it make sense to allow calling with both `empty` and `check`? What would be the expected use cases for such combination? If I understand correctly, it would actually ignore the `empty` and just perform a check, since the `createEmptyMigration` will be called with `write: false`, right?

tools/serverpod_cli/lib/src/commands/create_migration.dart
```diff
@@ -114,6 +149,27 @@ class CreateMigrationCommand extends ServerpodCommand<CreateMigrationOption> {
       return migration != null;
     });
 
+    if (check) {
+      // In check mode, exit with appropriate code based on whether changes were detected
+      if (empty) {
+        // Empty migrations always create a migration, so always indicate changes exist
+        log.info('Empty migration would be created.', type: TextLogType.bullet);
+        // Exit with code 1 (changes exist)
+        throw ExitException(1);
```
marcelomendoncasoares: As commented above, I think we should forbid the call with both `empty` and `check` parameters, as it would be basically an expensive alias to `(exit 1)`.

tools/serverpod_cli/lib/src/commands/create_migration.dart
```diff
@@ -114,6 +149,27 @@ class CreateMigrationCommand extends ServerpodCommand<CreateMigrationOption> {
...
+      } else if (migration == null) {
+        // No changes detected
+        log.info('No changes detected.', type: TextLogType.success);
+        // Exit with code 0 (success)
+        return;
+      } else {
+        // Changes detected
+        log.info('Changes detected.', type: TextLogType.bullet);
+        // Exit with code 1 (changes exist)
+        throw ExitException(1);
```
marcelomendoncasoares: Suggestion: Use the `ExitException.error()` method to follow the code standard.

tools/serverpod_cli/lib/src/migrations/generator.dart
```diff
@@ -34,6 +34,68 @@ class MigrationGenerator {
...
+  /// If [write] is false, the migration will not be written to disk.
+  ///
+  /// Returns the migration version, or null if no migration was created.
+  ///
+  /// Throws [MigrationVersionAlreadyExistsException] if the migration version
+  /// already exists.
+  Future<MigrationVersion?> createEmptyMigration({
+    String? tag,
+    bool write = true,
+  }) async {
```
marcelomendoncasoares: Suggestion: If the empty migration is just a subset of the code of the normal migration, I'd suggest we try to add the `empty` functionality to the existing `createMigration` to avoid code duplication - the same way the `force` parameter is implemented.

tools/serverpod_cli/lib/src/migrations/generator.dart
```diff
@@ -34,6 +34,68 @@ class MigrationGenerator {
...
+      migrationRegistry.add(versionName);
+      await migrationRegistry.write();
+    }
+
+    return migrationVersion;
+  }
+
   /// Creates a new migration version.
   /// If [tag] is specified, the migration will be tagged with the given name.
   /// If [force] is true, the migration will be created even if there are
```
marcelomendoncasoares: Suggestion: Since we are introducing an `--empty` flag, we should also indicate the possibility of using it when there are no changes in the warning below (L175-L180):

```dart

    if (migration.isEmpty && !force) {
      log.info(
-       'No changes detected. Use --force to create an empty migration.',
+       'No changes detected. Use --force or --empty to create an empty migration.',
      );
      return null;
    }
```

We should still support the `force` command generating the empty migration in such cases, as it is compatible with its meaning (and a commonly used workflow), but also suggest the usage of the `empty` flag.

tools/serverpod_cli/lib/src/commands/create_migration.dart
```diff
@@ -114,6 +149,27 @@ class CreateMigrationCommand extends ServerpodCommand<CreateMigrationOption> {
...
+      }
+    }
+
+    // Normal mode - create migration
     var projectDirectory = migration?.projectDirectory;
     var migrationName = migration?.versionName;
     if (migration == null ||
```
marcelomendoncasoares: Here we must change to no longer throw an error if the `migration` is null due to no changes. This is a breaking change requested in the issue to allow developers to call the `create-migration` twice in a row.

marcelomendoncasoares (changes requested):
Hi, @nicowalter256! Good start on this one. Had a couple of comments regarding the feature behavior and the implementation.

We will need tests for this one, that can follow the same existing pattern as the ones for `create-migration`. Check the `tests/serverpod_test_server/test_e2e_migrations` directory for reference.

tools/serverpod_cli/lib/src/commands/create_migration.dart
nicowalter256: Your interpretation was correct - when both flags were used, it would call createEmptyMigration with write: false, but then the check logic would always exit with code 1 saying "Empty migration would be created", which was misleading since no actual migration was being created.
Thanks for catching this!

tools/serverpod_cli/lib/src/migrations/generator.dart
nicowalter256: ok

tools/serverpod_cli/lib/src/commands/create_migration.dart
```diff
@@ -61,6 +84,18 @@ class CreateMigrationCommand extends ServerpodCommand<CreateMigrationOption> {
...
     String? tag = commandConfig.optionalValue(CreateMigrationOption.tag);
+    bool check = commandConfig.value(CreateMigrationOption.check);
+    bool empty = commandConfig.value(CreateMigrationOption.empty);
+
+    // Validate that check and empty flags are not used together
+    if (check && empty) {
+      log.error(
+        'Cannot use both --check and --empty flags together. '
+        'The --check flag checks for changes without creating migrations, '
+        'while --empty creates an empty migration.',
```
marcelomendoncasoares: Suggestion(NIT): We can avoid the repeated command description, since it is already present in the help. The first line is already clear for the user.

```suggestion
        'Cannot use both --check and --empty flags together. ',
```

tools/serverpod_cli/lib/src/migrations/generator.dart
```diff
@@ -53,11 +55,58 @@ class MigrationGenerator {
...
+    );
+
+    // Handle empty migration case
+    if (empty) {
+      // Create an empty migration
+      var migration = DatabaseMigration(
+        actions: [],
+        warnings: [],
+        migrationApiVersion: DatabaseConstants.migrationApiVersion,
+      );
```
marcelomendoncasoares: Suggestion(NIT): Move this to inline on the `MigrationVersion` creation, since the variable wont be used and all other fields are being created inline.

tools/serverpod_cli/lib/src/commands/create_migration.dart
```diff
@@ -114,14 +162,31 @@ class CreateMigrationCommand extends ServerpodCommand<CreateMigrationOption> {
       return migration != null;
     });
 
-    var projectDirectory = migration?.projectDirectory;
-    var migrationName = migration?.versionName;
-    if (migration == null ||
-        projectDirectory == null ||
-        migrationName == null) {
```
marcelomendoncasoares: Question: Any reason why you are no longer exiting with an error if `projectDirectory` or `migrationName` are `null`? Or was this just forgotten when updating the condition to not throw anymore when there are no migrations to create?

tools/serverpod_cli/lib/src/commands/create_migration.dart
```diff
@@ -114,14 +162,31 @@ class CreateMigrationCommand extends ServerpodCommand<CreateMigrationOption> {
       return migration != null;
```
marcelomendoncasoares: We should not return `false` here if there is no migration to create. Otherwise, a red `✗` will be printed instead of the green `✓` that we expect.

When using the `check` flag, the behavior must also be consistent: green check for no migration, red X for changes detected.

marcelomendoncasoares (changes requested):
Good progress. Beside the review comments, also remove the comment lines that are just code explanation comments. AI frequently generates a lot of them, but it makes the code worst to maintain and they tend get out-of-sync with the code. Good comments are the ones that highlight decisions or unexpected behaviors that future developers might need to understand the code, but I don't think we have any in this PR.

We also need to work on the migration tests - both to fix the ones that broke and to introduce new ones for `check` and `empty` flags. As I've mentioned, migration tests are located at `tests/serverpod_test_server/test_e2e_migrations`. You can run them by calling `util/run_tests_migrations_e2e` from the repository root. If you want to run just a few, alter the `tests\docker\tests_e2e_migrations\run-tests.sh` on line 28 to specify the tests you want to run (instead of the entire `test_e2e_migrations` folder).

The debugging of the tests was a little bit rough, but it is being fixed on PR #3988. So, if you get to work on this before the change gets to main, just copy the changes on `run-tests.sh` file to your branch and you will have clean logs on the migration tests.

If you need any help, feel free to reach me!

SandPod:
@nicowalter256 - Are you still working on this?

nicowalter256:
> @nicowalter256 - Are you still working on this?

yes. was away but back to it

marcelomendoncasoares:
Hey @nicowalter256! I see you made some progress here. For the next review, we will need to have the CI passing - and the addition of those tests I have mentioned in the last comment.

marcelomendoncasoares:
@nicowalter256 Are you still planning to continue this work or can we take it from here?

nicowalter256:
hello @marcelomendoncasoares been off my machine for past 3 weeks. I guess you take it from here since will be back in next 3-4 weeks for active contributions. Sorry for any incovinience caused

marcelomendoncasoares:
No problem @nicowalter256. We thank you for your contribution so far!

Issue #2498, opened 2024-07-17
Improved hosting of static files in Relic


vlidholt:
Serverpod's web server, Relic, supports hosting static files through the `RouteStaticDirectory`. It currently has limited functionality and is a bit fiddly and difficult to understand how to configure. We can improve the interface and take inspiration from `shelf_static`, who seems to have done a better job at handling MIME types, listing directories, etc. We can move code from `shelf_static` to `serverpod_relic_helpers` as needed.

`RouteStaticDirectory` can have the following constructor. It will serve files from the provided directory but place them under the path specified by the `Router` (see #2489).

```dart
RouteStaticDirectory({
  required String fileSystemPath,
  String? defaultDocument,
  bool listDirectories = false,
  bool useHeaderBytesForContentType = false,
  MimeTypeResolver? contentTypeResolver,
});
```

coderxsle:
👍 Amazing! Every step forward is a historic leap, propelling Serverpod toward an even greater future!  💪Keep pushing forward，greatness is just ahead! 🚀

🔥🎉 I absolutely love Serverpod and hope to contribute to its growth as well! ❤️🌍

I plan After completing my project, 🎥 to record a series  of tutorials on Serverpod。I want more people to learn about it and discover its true value! 🚀📚

🚀  Serverpod series tutorials 
 🌐 Serverpod API backend
📱 Serverpod Flutter App
 🖥️  Serverpod Flutter Web Admin 
 🔥 Promoting Serverpod to a wider

PR #3980, opened 2025-09-16
feat!: Defer static handling to Relic

nielsenko:
## Description

- Updates `RouteStaticDirectory` to use Relic's static handler with improved caching and HTTP feature support
- Migrates `Route` class from single method to support multiple HTTP methods
- Refactors web server request handling to use Relic's pipeline and routing system
- Updates all `withResponse` calls to use Relic's `respond` method

Fixes: #2498
Fixes: #3758 
Fixes: #3986
Fixes: #3983 

## Pre-launch Checklist

## Breaking changes

This changes the interface of `RouteStaticDirectory`, `Route`, and how to setup routing of static assets (using tail-matching `/prefix/**`)

## Summary by CodeRabbit

- New Features
  - Improved static file serving: recursive matching (/**), nested paths, automatic content types, configurable Cache-Control, ETag/If-None-Match, Last-Modified/If-Modified-Since, Range (partial content), and HEAD support.
- Refactor
  - Routing supports multiple HTTP methods per route; simplified response handling.
  - Static directory route configuration simplified; basePath parameter removed.
- Tests
  - Expanded integration tests for static files, caching, conditional requests, ranges, and HEAD behavior.
- Chores
  - Updated dependency resolution for “relic” to a pinned Git reference across packages and templates.

### DIFF
--- a/packages/serverpod/lib/src/web_server/routes/static_directory.dart
+++ b/packages/serverpod/lib/src/web_server/routes/static_directory.dart
@@ -3,6 +3,7 @@ import 'package:serverpod/serverpod.dart';
 // NOTE: This class has been cleared out and marked abstract.
 // It is only left behind to add a deprecation annotation
 // to help people move to use StaticRoute.
+/// DEPRECATED: Use [StaticRoute.directory] with [CacheControlFactory] instead
 /// A path pattern to match, and the max age that paths that match the pattern
 /// should be cached for, in seconds.
 @Deprecated('Use StaticRoute.directory and CacheControlFactory instead')
@@ -11,6 +12,7 @@ class PathCacheMaxAge {}
 // NOTE: This class has been cleared out and marked abstract.
 // It is only left behind to add a deprecation annotation
 // to help people move to use StaticRoute.
+/// DEPRECATED: Use [StaticRoute.directory] instead
 /// Route for serving a directory of static files.
 @Deprecated('Use StaticRoute.directory instead')
 abstract class RouteStaticDirectory extends Route {}


packages/serverpod/pubspec.yaml
```diff
@@ -29,7 +29,10 @@ dependencies:
   path: ^1.8.3
   postgres: ^3.4.0
   redis: ^4.0.0
-  relic: ^0.4.1
+  relic: 
+    git:
+      url: https://github.com/nielsenko/relic.git
+      ref: merged
```
nielsenko: Once the outstanding PRs in relic lands and a release is cut, then this will be replaced with a proper version.

nielsenko:
@coderabbit review

templates/pubspecs/packages/serverpod/pubspec.yaml
```diff
@@ -28,7 +28,10 @@ dependencies:
   path: ^1.8.3
   postgres: ^3.4.0
   redis: ^4.0.0
-  relic: ^0.4.1
+  relic: 
+    git:
+      url: https://github.com/nielsenko/relic.git
+      ref: merged
```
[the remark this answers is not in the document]
nielsenko: If you read my comment https://github.com/serverpod/serverpod/pull/3980#discussion_r2355444398 you would know this is only a temporary measure until some outstanding PRs land in relic.

tests/serverpod_test_server/test_integration/static_files_test.dart
```diff
@@ -1,19 +1,34 @@
+@Timeout.none
+library;
+
 import 'dart:io';
 
 import 'package:serverpod/serverpod.dart';
 import 'package:serverpod_test_server/test_util/test_serverpod.dart';
 import 'package:test/test.dart';
 import 'package:http/http.dart' as http;
+import 'package:http/io_client.dart' as http;
```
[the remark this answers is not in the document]
nielsenko: This is not a problem in Dart as long as there are no conflicts.

packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -94,56 +99,29 @@
...
+        if (cookie.name == 'auth') {
+          authenticationKey = cookie.value;
+        }
       }
-    }
 
-    var queryParameters = request.url.queryParameters;
-    authenticationKey ??= queryParameters['auth'];
+      var queryParameters = request.url.queryParameters;
+      authenticationKey ??= queryParameters['auth'];
```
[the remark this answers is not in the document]
nielsenko: We'll need to keep for legacy reason a little longer.

packages/serverpod/lib/src/web_server/routes/static_directory.dart
```diff
@@ -62,104 +45,43 @@ class RouteStaticDirectory extends Route {
...
+            PathCacheMaxAge.oneYear;
+
+        // Set Cache-Control header
+        return pathCacheMaxAge == PathCacheMaxAge.noCache
+            // Don't cache this path
+            ? CacheControlHeader(
+                maxAge: 0, sMaxAge: 0, noCache: true, noStore: true)
+            : CacheControlHeader(maxAge: pathCacheMaxAge.inSeconds);
+      },
+    );
```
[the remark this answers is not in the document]
nielsenko: This is not a dynamically obtained value from the request, but set server-side during startup. We don't need to guard against escaping `web/`. The developer can do so if he chooses.

packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -71,12 +71,17 @@ class WebServer {
...
-      var context = _securityContext;
       final server = await serve(
-        _handleRequest,
+        const Pipeline()
+            .addMiddleware(
+              routeWith(router as Router<Route>, toHandler: _routeToHandler),
+            )
+            .addHandler(
+              respondWith((_) => Response.notFound()),
+            ),
```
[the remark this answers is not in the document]
nielsenko: Actually `routeWith` does not return 405 on method mismatch. The `Router` class does not distinguish between miss due to path, or miss due to method.

But this is valid feedback, we should return 405 with an appropriate `Allow` header.

Please raise an issue to address this in relic repo, and create an issue in serverpod that depends on this.

tests/serverpod_test_server/test_integration/static_files_test.dart
```diff
@@ -1,19 +1,34 @@
+@Timeout.none
+library;
```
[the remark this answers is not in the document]
nielsenko: The `@Timeout.none` need to be attached to a `library` element. Your suggestion is not valid. However this is cruft left over from development to prevent tests from timing out during debugging. I'll remove this.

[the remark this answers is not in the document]
nielsenko: Debug code removed

tests/serverpod_test_server/test_integration/static_files_test.dart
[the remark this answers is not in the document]
nielsenko: Debug code removed

tests/serverpod_test_server/test_integration/static_files_test.dart
```diff
@@ -112,17 +121,202 @@ void main() {
...
+          expect(headResponse.statusCode, 200);
+          expect(headResponse.body, isEmpty); // HEAD should have no body
+          expect(headResponse.headers['content-length'],
+              getResponse.headers['content-length']);
+          expect(headResponse.headers['etag'], getResponse.headers['etag']);
+          expect(headResponse.headers['last-modified'],
+              getResponse.headers['last-modified']);
+        });
+      });
+    });
```
marcelomendoncasoares: These new tests should also follow the `Given`/`When`/`Then` pattern.

packages/serverpod/lib/src/web_server/web_server.dart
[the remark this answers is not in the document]
nielsenko: @coderabbit Create an issue to remove this legacy code.

tests/serverpod_test_server/test_integration/static_files_test.dart
nielsenko: Fixed

packages/serverpod/lib/src/web_server/routes/static_directory.dart
```diff
@@ -62,104 +45,43 @@ class RouteStaticDirectory extends Route {
   /// The path to the directory to serve relative to the web/ directory.
   final String serverDirectory;
 
-  /// The path to the directory to serve static files from.
-  final String? basePath;
```
SandPod, line 66: Question(For my knowledge): This is now configured through the path of the route right?

packages/serverpod/lib/src/web_server/routes/static_directory.dart
```diff
@@ -62,104 +45,43 @@ class RouteStaticDirectory extends Route {
...
   /// Creates a static directory with the [serverDirectory] as its root.
-  /// If [basePath] is provided, the directory will be served from that path.
   /// If [pathCachePatterns] is provided, paths matching the requested
   /// patterns will be cached for the requested amount of time. Paths that
   /// are do not match any provided pattern are cached for one year.
   RouteStaticDirectory({
     required this.serverDirectory,
-    this.basePath,
-    this.serveAsRootPath,
     List<PathCacheMaxAge>? pathCachePatterns,
```
SandPod: Question/Discussion: Does this still make sense? Perhaps we should upgrade the interface in a different PR?

packages/serverpod/lib/src/web_server/routes/static_directory.dart
```diff
@@ -62,104 +45,43 @@ class RouteStaticDirectory extends Route {
...
+    _handler = createStaticHandler(
+      p.join('web', serverDirectory),
+      cacheControl: (ctx, fileInfo) {
+        // Get the max age for the path
+        var pathCacheMaxAge = _pathCachePatterns
+                ?.firstWhereOrNull(
+                    (pattern) => pattern._shouldCache(ctx.remainingPath.path))
+                ?.maxAge ??
+            // Default to a max age of one year if no pattern matched
+            PathCacheMaxAge.oneYear;
```
SandPod: Suggestion: This is really not helpful, should we make this explicit in some way? Since we don't have a cache busting mechanism its quite easy to accidentally deploy something that is cached forever.

packages/serverpod/lib/src/web_server/routes/static_directory.dart
```diff
@@ -62,104 +45,43 @@ class RouteStaticDirectory extends Route {
   /// The path to the directory to serve relative to the web/ directory.
```
SandPod: Question: In cloud we use this to server a single page application. How would one configure such an application using this route now? Perhaps we should supply such an implementation out of the box?

packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -71,12 +71,17 @@ class WebServer {
     }
 
     try {
-      var context = _securityContext;
       final server = await serve(
-        _handleRequest,
+        const Pipeline()
+            .addMiddleware(
+              routeWith(router as Router<Route>, toHandler: _routeToHandler),
```
SandPod: Question: Should `router` be a private property of the web server? Since we provide the `addRoute` method?

packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -94,56 +99,29 @@ class WebServer {
...
-      logDebug('Malformed call, invalid uri. Client IP: ${request.remoteInfo}');
-      return context.withResponse(Response.badRequest());
-    }
-
-    String? authenticationKey;
-    for (var cookie in request.headers.cookie?.cookies ?? const <Cookie>[]) {
-      if (cookie.name == 'auth') {
-        authenticationKey = cookie.value;
+      String? authenticationKey;
+      for (var cookie in request.headers.cookie?.cookies ?? const <Cookie>[]) {
```
SandPod: Question: This should be removed, right?

And why not removing the query params reading directly?

packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -94,56 +99,29 @@ class WebServer {
...
-      uri.path,
-    );
-    if (match != null) {
-      final route = match.value;
-      return await _handleRouteCall(route, session, context);
-    }
+      final session = WebCallSession(
+        server: serverpod.server,
+        endpoint: request.requestedUri.path,
+        authenticationKey: authenticationKey,
```
SandPod: Question: This should simply be the auth header right?

packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -164,7 +142,7 @@ class WebServer {
         request: request,
       );
 
-      return context.withResponse(Response.internalServerError());
+      return context.respond(Response.internalServerError());
```
SandPod, line 135: Suggestion: Remove the `TODO` in this method. This should be handled now right?

packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -226,26 +204,17 @@ class WebServer {
...
 abstract class Route {
-  /// The method this route will respond to, i.e. HTTP get or post.
-  final RouteMethod method;
+  /// The methods this route will respond to, i.e. HTTP get or post.
+  final Set<Method> methods;
   String? _matchPath;
 
   /// Creates a new [Route].
-  Route({this.method = RouteMethod.get});
+  Route({this.methods = const {Method.get}});
```
SandPod, line 206: Question: The `getBody` method here seems like it should be delegated to Relic. Can we do that?

packages/serverpod/lib/src/web_server/routes/static_directory.dart
nielsenko: This is consistent with existing behaviour, but I agree. Should we simply not set Cache-Control if not matched? 

Or should we overhaul this interface completely? Getting rid of `PathCacheMaxAge` class?

SandPod: This would be the time to change existing behavior.

The way this is configured now attempts to give a reasonable default with some configuration support for specific paths. Since we modified how this works in relic, are we happy with that approach? If so, do you think we should reflect that pattern here? Perhaps the `PathCacheMaxAge` should be explicit instead of implicit? So you need to make some form of decision around caching?

Or do you have a suggestion for what would make this a simple to use interface?

packages/serverpod/lib/src/web_server/routes/static_directory.dart
nielsenko: Yes

packages/serverpod/lib/src/web_server/web_server.dart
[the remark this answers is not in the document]
nielsenko: Fixed in relic

packages/serverpod/lib/src/web_server/web_server.dart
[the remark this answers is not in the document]
nielsenko: Fixed

packages/serverpod/lib/src/web_server/routes/static_directory.dart
nielsenko: I have added a new `StaticRoute` class that support both explicit files and root directories. I have kept the `RouteStaticDirectory` but marked it deprecated.

packages/serverpod/lib/src/web_server/routes/static_directory.dart
nielsenko: I have added a new `StaticRoute` that follows relic, and marked `RouteStaticDirectory` as deprecated.

packages/serverpod/lib/src/web_server/routes/static_directory.dart
nielsenko:
```dart
pod.webServer.addRoute(StatidRoute.file(File(´index.html´)), ´/**´);
```
will route all paths except those with literal first segment (say fx `/api/..`) to `index.html`.

packages/serverpod/lib/src/web_server/web_server.dart
nielsenko: Done

packages/serverpod/lib/src/web_server/web_server.dart
nielsenko: Done

packages/serverpod/lib/src/web_server/web_server.dart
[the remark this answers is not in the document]
nielsenko: This PR fixes #3983

packages/serverpod/lib/src/web_server/web_server.dart
nielsenko: This is dead code. Removed

packages/serverpod/lib/src/web_server/web_server.dart
nielsenko: I have removed `addRoute`. Instead you should use `router.add`.

SandPod: Hmm, from an interface perspective, I think I preferred the `addRoute` call directly on webServer. That more directly communicates what should be done.

packages/serverpod/lib/src/web_server/routes/static_directory.dart
SandPod: How about if our navigation actually uses subpaths and these should route to some specified screen in the app? That would require any unmatched paths, even those containing a first segment, to map to index.html.

SandPod (commented):
Couple of last comments and questions.

packages/serverpod/lib/src/web_server/web_server.dart
nielsenko: I'm afraid we will and up replicating the interface on `WebServer`. Isn't it better to delegate to the router?

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/lib/server.dart
```diff
@@ -15,17 +17,15 @@ void run(final List<String> args) async {
...
-  pod.webServer.addRoute(RootRoute(), '/index.html');
-  // Serve all files in the /static directory.
-  pod.webServer.addRoute(
-    RouteStaticDirectory(serverDirectory: 'static', basePath: '/'),
-    '/*',
-  );
+  pod.webServer.router.add(RootRoute(), '/');
+  pod.webServer.router.add(RootRoute(), '/index.html');
+  // Serve all files in the web/static relative directory under /.
+  final root = Directory(Uri(path: 'web/static').toFilePath());
```
SandPod, line 27: Suggestion: This seems cumbersome, could we use this instead?

Fix for all places where we configure this.
```dart
pod.webServer.router.add(
    StaticRoute.directory(Directory('static')),
    '/*',
  );
```

packages/serverpod/lib/src/web_server/routes/static_directory.dart
```diff
@@ -58,108 +41,49 @@ class PathCacheMaxAge {
 }
 
 /// Route for serving a directory of static files.
+@Deprecated('Use StaticRoute.directory instead')
```
SandPod, line 17: Suggestion: Make this class abstract and clear out the implementation.

This forces everyone to upgrade with a nice nudge and we don't need to maintain this old interface.

packages/serverpod/lib/src/web_server/routes/static_route.dart
```diff
@@ -0,0 +1,42 @@
...
+
+  StaticRoute._(this._handler) : super(methods: {Method.get, Method.head});
+
+  /// Use [StaticRoute.directory] to serve everything below a given [root].
+  ///
+  /// Use [cacheControlFactory] to customize what [CacheControlHeader] to
+  /// return for a given asset. Default is to leave caching behavior to client
+  /// side heuristics.
+  factory StaticRoute.directory(Directory root,
+      [CacheControlFactory cacheControlFactory = _defaultFactory]) {
```
SandPod, line 38: Question/Discussion: Two things.

1. We now expose Relic primitives, do we want to do that? Just to make sure this is a deliberate decision.
2. Are there any nice default value builders on the CacheControlFactory? Say for caching content for a year?

packages/serverpod/lib/src/web_server/routes/static_directory.dart
nielsenko: In relic
```dart
const Pipeline()
            .addMiddleware(routeWith(router, toHandler: _routeToHandler))
            .addHandler(_routeToHandler(Static.file(File('index.html)));
```
but this is the stuff we need to expose an interface for.

packages/serverpod/lib/src/web_server/routes/static_route.dart
SandPod: Decision:
1. This is encouraged and allows users to reuse their relic middleware or modules.
2. We will add some default cache control factories that are available for the user. We will set the default to validate the etag before fetching the content.

packages/serverpod/lib/src/web_server/routes/static_directory.dart
SandPod: Discussed in slack, works as expected.

packages/serverpod/lib/src/web_server/routes/static_route.dart
nielsenko: Done

packages/serverpod/lib/src/web_server/routes/static_directory.dart
```diff
@@ -40,50 +40,9 @@ class PathCacheMaxAge {
...
-    );
-  }
-
-  late final Handler _handler;
-
-  @override
-  FutureOr<HandledContext> handleCall(Session session, NewContext context) =>
-      _handler(context);
-}
+abstract class RouteStaticDirectory extends Route {}
```
SandPod, line 18: Question: Shouldn't the same fix be applied to the rest of this file? `PathCacheMaxAge`, for example?

SandPod (approved):
Last considerations otherwise it looks good! 👍

packages/serverpod/lib/src/web_server/routes/static_directory.dart
nielsenko: Done

PR #3982, opened 2025-09-17
feat: Implement `ClientAuthSessionManager` completely with JWT and SAS support

### LINKED ISSUE #3674: JWT Session manager

marcelomendoncasoares:
This PR closes #3674. 

To facilitate the revision, the commit history was separated into preparation commits that include the necessary support features and the actual session manager changes.

## Features introduced

- The `AuthSuccess` class now carry the access token expiration, if applicable, to allow early refreshing.
- Users now have a `UnifiedAuthTokens.authenticationHandler` to use on the server to handle both JWT and SAS tokens.
- JWT server-side configuration keys no longer have dots (which would be problematic to use with env vars).
- A simple status endpoint will be available out-of-the-box with our IDP package, together with the refresh JWT endpoint.
- Users can now force-refresh the tokens using the `refreshAuthKey` method (which is mutex protected).
- The `ClientAuthSessionManager` is now fully functional and support both SAS and JWT tokens.

> The term "SAS" stands for "Serverpod Authentication Session", and is the acronym given to our server-side session authentication method. A PR will follow this one with a refactor to change "session" references to "sas" whenever we refer to the authentication session. This change will eliminate the confusion with the other session objects throughout the codebase.

## Additional notes

There is one pending `TODO` on the `StatusEndpoint` to be done after the [TokenManager](https://github.com/serverpod/serverpod/pull/3970/files#diff-e200ece62a44fa80cae9cdc8867583938136558e79fa400a2a06094a588f4d5b) class is implemented according to this [POC](https://github.com/serverpod/serverpod/pull/3970/files). The `signOut` methods should delegate the token destruction to the `TokenManager` instead of exhaustively trying our known token handlers, as it was temporarily implemented.

Because of this, integration tests to the `StatusEndpoint` were skipped - although the relevant behavior of the class is being tested through the `ClientAuthSessionManager` integration tests.

## Pre-launch Checklist

## Breaking changes

The PR introduces a new `force` parameter in the `RefresherClientAuthKeyProvider` interface, which would require client code to update the `refreshAuthKey` method signature. However, this class is not expected to be used externally yet, as it is still part of the latest alpha version.


modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/jwt_util.dart
```diff
@@ -129,12 +129,27 @@ class JwtUtil {
...
       refreshTokenId: refreshTokenId,
+      tokenExpiresAt: _extractExpirationDate(allClaims),
       authUserId: authUserId,
       scopes: scopes,
       extraClaims: extraClaims,
     );
   }
 
+  /// Extracts the expiration date from a JWT token.
+  DateTime extractExpirationDate(final String accessToken) {
```
SandPod, line 140: Suggestion(NIT): Perhaps `_extractPayload` should be the private method?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/jwt/integration/authentication_tokens_test.dart
```diff
@@ -44,6 +44,25 @@ void main() {
       );
     });
 
+    test('when requesting a new token pair, then expiration date is returned.',
+        () async {
+      final authSuccess = await AuthenticationTokens.createTokens(
```
SandPod: Suggestion/Discussion: Can you fix the clock using the `Clock` package here?

It would be nice to move away from an approximate test.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/endpoints/jwt_tokens_endpoint.dart
```diff
@@ -0,0 +1,34 @@
...
+  ///   only if it has not been used within configured `refreshTokenLifetime`.
+  /// -[RefreshTokenInvalidSecretException]: refresh token is incorrect, meaning
+  ///   it does not refer to the current secret refresh token. This indicates
+  ///   either a malfunctioning client or a malicious attempt by someone who has
+  ///   obtained the refresh token. In this case the underlying refresh token
+  ///   will be deleted, and access to it will expire fully when the last access
+  ///   token is elapsed.
+  ///
+  /// This endpoint is unauthenticated, meaning it won't receive authentication
+  /// information in the [session] object.
```
SandPod: Suggestion: `This endpoint is unauthenticated, meaning the client won't include any authentication information with the call`?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/authentication_tokens.dart
```diff
@@ -136,6 +136,32 @@ abstract final class AuthenticationTokens {
...
 
+  /// Returns a access token while also rotating the refresh token.
+  ///
+  /// Invalidates the previous refresh token as security best practice.
+  static Future<AuthSuccess> refreshAccessToken(
+    final Session session, {
+    required final String refreshToken,
+    final Transaction? transaction,
+  }) async {
+    final refreshesTokenPair = await rotateRefreshToken(
```
SandPod, line 147: Question: Is `rotateRefreshToken` now an implementation detail and we should only expose `refreshAccessToken`?

SandPod: Then we should probably move all those tests over to the exposed one.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/unified.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_core_server/jwt.dart';
+import 'package:serverpod_auth_core_server/session.dart';
+
+/// Authentication handler that supports both JWT and SAS auth tokens.
+abstract final class UnifiedAuthTokens {
```
SandPod, line 6: @exaby73 Here is the introduction of the unified authentication handler.

We will need to change the shape of this a bit so that the `authenticationHandler` can be configured before it is used.

My thinking around this now is that we simply remove the authenticationHandler as a constructor parameter of serverpod and have it as a setter method instead. That would allow us to use configuration read through Serverpod when configuring our `authenticationHandler`.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/endpoints/status_endpoint.dart
```diff
@@ -0,0 +1,52 @@
...
+
+/// Endpoint for getting status and managing a signed in user.
+class StatusEndpoint extends Endpoint {
+  /// Returns true if the client user is signed in.
+  Future<bool> isSignedIn(final Session session) async {
+    final authInfo = await session.authenticated;
+    return authInfo?.authId != null;
+  }
+
+  // TODO: Replace signout methods implementation by using the [TokenManager]
```
SandPod, line 13: 👍  FYI @exaby73

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/endpoints/status_endpoint.dart
```diff
@@ -0,0 +1,52 @@
...
+    if (authUserIdStr == null) return;
+    final authUserId = UuidValue.withValidation(authUserIdStr);
+
+    try {
+      await AuthenticationTokens.destroyAllRefreshTokens(
+        session,
+        authUserId: authUserId,
+      );
+      return;
+    } on RefreshTokenNotFoundException catch (_) {}
```
SandPod: Discussion: This exposes a difference in the API for AuthSessions and AuthenticationTokens. Perhaps we should return a list of the entries that were removed, or a count if we want to be anonymous. Or should we throw?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/endpoints/status_endpoint.dart
```diff
@@ -0,0 +1,52 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_core_server/serverpod_auth_core_server.dart';
+import 'package:serverpod_auth_core_server/session.dart';
+import 'package:serverpod_auth_core_server/src/jwt/business/authentication_tokens.dart';
+
+/// Endpoint for getting status and managing a signed in user.
+class StatusEndpoint extends Endpoint {
```
SandPod, line 6: Suggestion: Please add some tests that validate that this works as expected.

When inspecting commit 141e244b127f2b762a08384baacd12bcb67649d7 I can't see any tests added.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/auth_key_providers/jwt_auth_key_provider.dart
```diff
@@ -0,0 +1,85 @@
+import 'package:clock/clock.dart';
+import 'package:serverpod_auth_core_client/serverpod_auth_core_client.dart';
+
+/// The [JwtAuthKeyProvider] keeps track of and manages the signed-in state of
+/// the user for JWT-based authentication. This is the default authentication
+/// method for the client.
```
SandPod: Question: Why is this the default authentication method for the client? Do we need to have an opinion?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
```diff
@@ -53,17 +64,56 @@ class ClientAuthSessionManager implements ClientAuthKeyProvider {
...
 
+  /// The authentication key provider to use for the current auth strategy.
+  ClientAuthKeyProvider? get authKeyProviderDelegate {
+    final authStrategy = authInfo.value?.authStrategy;
+    if (authStrategy == null) return null;
+
+    var authKeyProvider = _authKeyProviderDelegates[authStrategy];
+    if (authKeyProvider != null) return authKeyProvider;
+
+    switch (authStrategy) {
```
SandPod: Question: Have we switched this over to a String now to allow further expansion of this list?
Perhaps this should also be a lookup in a configurable map?

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_auth_key_provider_test.dart
```diff
@@ -0,0 +1,360 @@
...
+      },
+      refreshEndpoint: refreshEndpoint,
+      refreshJwtTokenBefore: const Duration(seconds: 30),
+    );
+
+    final testUserId = await client.authTest.createTestUser();
+    jwtAuthSuccess = await client.authTest.createJwtToken(testUserId);
+
+    // Need to wait a second to ensure the next JWT token has a different
+    // expiry date. Otherwise it would look equal.
```
SandPod: Question: This is not true anymore right? All JWT tokens should be unique now right? If so, please remove.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_auth_key_provider_test.dart
```diff
@@ -0,0 +1,360 @@
...
+
+    expect(result, 'Bearer ${jwtAuthSuccess.token}');
+  });
+
+  group(
+      'Given a JwtAuthKeyProvider with auth info that has no expiration time '
+      'when refreshing auth key ', () {
+    setUp(() async {
+      storedAuthInfo = jwtAuthSuccess.copyWith()..tokenExpiresAt = null;
+      result = await provider.refreshAuthKey();
```
SandPod, line 85: Discussion: This tripped me a bit but remembering that this method takes a `force` parameter made me understand why an explicit call could be optimized away.

packages/serverpod_client/test/test_utils/test_auth_key_providers.dart
```diff
@@ -19,6 +20,7 @@ class TestRefresherAuthKeyProvider implements RefresherClientAuthKeyProvider {
   @override
   Future<RefreshAuthKeyResult> refreshAuthKey({bool force = false}) async {
     refreshCallCount++;
+    forcedRefresh = force;
```
SandPod: Question: This value would be overridden; perhaps it is better to store a list of the `force status` for the calls?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/jwt/integration/authentication_tokens_test.dart
marcelomendoncasoares: Done.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/jwt_util.dart
marcelomendoncasoares: It is also being used on `AuthenticationTokens.createTokens` (L129), so it must be a public method.

```dart
    return AuthSuccess(
      authStrategy: AuthStrategy.jwt,
      token: token,
      tokenExpiresAt: _jwtUtil.extractExpirationDate(token), // <-- here
      refreshToken: RefreshTokenString.buildRefreshTokenString(
        refreshToken: refreshToken,
        rotatingSecret: secret,
      ),
      authUserId: authUserId,
      scopeNames: scopes.names,
    );
```

We need it there to extract the `expirationDate` from the newly created access token.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/endpoints/jwt_tokens_endpoint.dart
marcelomendoncasoares: Done.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/authentication_tokens.dart
marcelomendoncasoares: I wouldn't say so, because users can rely on our JWT implementation, but not necessarily use our `AuthSuccess` and the rest of the authentication package. In such custom implementations, they could still use the `rotateRefreshToken` as the vanilla method for a common JWT operation.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/endpoints/status_endpoint.dart
marcelomendoncasoares: This was actually intentional, as stated in the PR description:

> There is one pending TODO on the StatusEndpoint to be done after the [TokenManager](https://github.com/serverpod/serverpod/pull/3970/files#diff-e200ece62a44fa80cae9cdc8867583938136558e79fa400a2a06094a588f4d5b) class is implemented according to this [POC](https://github.com/serverpod/serverpod/pull/3970/files). The signOut methods should delegate the token destruction to the TokenManager instead of exhaustively trying our known token handlers, as it was temporarily implemented.

> Because of this, integration tests to the StatusEndpoint were skipped - although the relevant behavior of the class is being tested through the ClientAuthSessionManager integration tests.

I have skipped it to avoid a refactor of such tests after we add `TokenIssuer`s, which is going to be done in sequence. The endpoint methods signature might not change much, but its internals - and, consequentially, the tests setup and use cases - sure will. Since the `TokenIssuer` implementation will require a refactor on this class already, do you think it is necessary that we add tests here before and refactor next?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/auth_key_providers/jwt_auth_key_provider.dart
marcelomendoncasoares: This comment is actually left from a previous implementation and I forgot to remove.

As a general perspective regarding your question, though, I think we offer a best experience modelling our auth system as close as possible to zero-config. Being able to add a few imports and calls and see it working without first thinking about configuration reduces friction and accelerates development, specially for less advanced users. Lately, users can tweak configurations as needed for their use cases. Just something for us to keep in mind when working the server side and defining default options - if you and @exaby73 also agree.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_auth_key_provider_test.dart
marcelomendoncasoares: Good catch, it was added before the fix on the other PR and I forgot to remove. Removing this and the delay below will also speed up the tests. Done!

packages/serverpod_client/test/test_utils/test_auth_key_providers.dart
marcelomendoncasoares: Good idea, it will make the parameter check on tests clearer. Done!

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/endpoints/status_endpoint.dart
marcelomendoncasoares: This was actually a confusion of mine... The `RefreshTokenNotFoundException` is only thrown in the `rotateTokens` method, not in the deletion. I have changed the delete methods to return:

- A `bool` for the individual delete.
- A `List<UuidValue>` with the IDs of the deleted tokens.

Then, we now only call the other token manager if the first returned `false`/`isEmpty`.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
marcelomendoncasoares: I was thinking in making this switch in the refactor PR that will follow this one to standardize `AuthSessions` and `AuthenticationTokens`, but can do it right here as well.

marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/3982/commits/0a0e9951839da4102d2eb960b1ec49695b7d3a96.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/endpoints/status_endpoint.dart
marcelomendoncasoares: When making this change at https://github.com/serverpod/serverpod/pull/3982/commits/580622b248b57752623bec28aa55b43e67c427ca, I actually found a few blind spots on the tests - both for previous SAS/JWT destruction and for `signOutDevice`/`signOutAllDevices` methods. Adding the tests uncovered two minor bugs that are now fixed.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/jwt_util.dart
SandPod: Sorry, the comment was unclear. I was more suggestng that the contents of this method mostly concerns extracting the payload, while extrating the expiration date is abstracted away in a private method.

Given the name of this method, I would more expect the extracting of the payload to be abstracted awat in a private method, and this method to contain the lofic for extracting the expriation date.

But this is a NIT comment.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/authentication_tokens.dart
SandPod: Good call, this is about exposing the atoms. I lost track of that design goal. Thanks for guarding it.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/auth_key_providers/jwt_auth_key_provider.dart
SandPod: I agree, this should be under the hood configuration that is available if you would like to do a deep dive.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
SandPod: As far as I can tell the switch to String was done but we still have a switch statement making it not dynamic to register different auth key providers. Should we include this in this PR as well?

SandPod (approved):
Two comments still open but the PR LGTM 👍

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
marcelomendoncasoares: It is, actually. We do allow users to register their providers in the constructor and use the switch statement here only to ensure we exhaust our own options when no provider is specified.

PR #3988, opened 2025-09-18
test: Fix verbose logging on tests that require the CLI

### LINKED ISSUE #3987: Excessive verbose logging during migration tests due to CLI dependency resolution
### Describe the bug

When running e2e migration tests in Docker containers (on the CI as well), the test output is flooded with verbose Dart pub dependency resolution logs (`FINE: Pub 3.8.0`, `SLVR: fact:`, `MSG : Resolving dependencies...`, etc.) for each test that creates a migration. This makes it extremely difficult to read test results and debug issues.

The verbose logs appear because the Serverpod CLI tool is performing dependency resolution for each `serverpod create-migration` command, despite the CLI being activated once at the beginning of the test run.

### To reproduce

1. Run the e2e migration tests:
   ```bash
   util/run_tests_migrations_e2e
   ```

2. Observe the output - each test that calls `MigrationTestUtils.createMigrationFromProtocols()` generates extensive pub dependency resolution logs like:
   ```
   tests-1 | +0: test_e2e_migrations/migrations_roundtrip_test.dart: Given new protocol model with table when creating and applying migration then database contains new table.
   tests-1 | FINE: Pub 3.8.0
   tests-1 | 
   tests-1 | MSG : Resolving dependencies...
   tests-1 | 
   tests-1 | SLVR: fact: serverpod_cli is 3.0.0-alpha.1
   tests-1 | SLVR: derived: serverpod_cli
   tests-1 | SLVR: fact: serverpod_cli depends on archive >=3.3.7 <5.0.0
   tests-1 | [... hundreds more lines of dependency resolution logs ...]
   ```

### Expected behavior

The CLI should cache its dependencies after the initial activation and not perform dependency resolution for each command invocation. Test output should be clean and focused on actual test results, with dependency resolution happening only once during CLI activation.

### Library version

- Serverpod: 3.0.0-alpha.1
- Dart: 3.8.0

### Platform information

- Docker container running `dart:3.8.0` base image
- Linux environment
- Container-based test execution

### Additional context

There is a simple (and somewhat hacky) solution that is quite simple: after activating the CLI in the `tests_e2e_migrations/run_tests.sh` file, call the serverpod CLI and activate the CLI again. Then, dependencies will be fully cached before any tests execute.

```bash
cd tools/serverpod_cli
dart pub global activate -s path . > /dev/null 2>&1

# Force the CLI to be activated and cache dependencies
serverpod version > /dev/null 2>&1
dart pub global activate -s path . > /dev/null 2>&1
```

- [x] I want to work on this

marcelomendoncasoares:
Fixes #3987.

Also apply the same fix to other scripts and tests that activate the CLI.

## Pre-launch Checklist

## Breaking changes

N/A.

### DIFF
--- a/tests/docker/tests_e2e_migrations/run-tests.sh
+++ b/tests/docker/tests_e2e_migrations/run-tests.sh
@@ -13,6 +13,11 @@ echo "### Serverpod home: $SERVERPOD_HOME"
 
 cd tools/serverpod_cli
 dart pub global activate -s path .
+
+# Force the CLI to be activated with cached dependencies
+serverpod version
+dart pub global activate -s path .
+
 cd ../..
 
 # Wait for the server to be up (timeout after 60 seconds)
--- a/tests/bootstrap_project/test/module_test.dart
+++ b/tests/bootstrap_project/test/module_test.dart
@@ -22,6 +22,19 @@ void main() {
       workingDirectory: cliPath,
     );
 
+    // Run command and activate again to force cache pub dependencies.
+    await runProcess(
+      'serverpod',
+      ['version'],
+      workingDirectory: cliPath,
+    );
+
+    await runProcess(
+      'dart',
+      ['pub', 'global', 'activate', '-s', 'path', '.'],
+      workingDirectory: cliPath,
+    );
+
     await Directory(tempPath).create();
   });
 
--- a/tests/bootstrap_project/test/project_mini_test.dart
+++ b/tests/bootstrap_project/test/project_mini_test.dart
@@ -21,6 +21,19 @@ void main() async {
       workingDirectory: cliPath,
     );
 
+    // Run command and activate again to force cache pub dependencies.
+    await runProcess(
+      'serverpod',
+      ['version'],
+      workingDirectory: cliPath,
+    );
+
+    await runProcess(
+      'dart',
+      ['pub', 'global', 'activate', '-s', 'path', '.'],
+      workingDirectory: cliPath,
+    );
+
     await Directory(tempPath).create();
   });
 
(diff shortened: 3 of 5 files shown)


tests/docker/tests_e2e_migrations/run-tests.sh
```diff
@@ -13,6 +13,11 @@ echo "### Serverpod home: $SERVERPOD_HOME"
 
 cd tools/serverpod_cli
 dart pub global activate -s path .
+
+# Force the CLI to be activated with cached dependencies
+serverpod version
+dart pub global activate -s path .
```
SandPod, line 19: Question: I don't understand why a second activation would make a different? Do you have some reference documentation for this?

SandPod (approved):
I'll approve since it actually does reduce the number of lines written from 39k to 7k. But I'm not sure why a double activation would make a difference.

tests/docker/tests_e2e_migrations/run-tests.sh
marcelomendoncasoares: The closest issue I found was this https://github.com/dart-lang/pub/issues/4536, that running an activated package triggers upgrade. It is solved, but seems to be only on `dev` release and to be published on stable `3.10+`.

I don't understand 100% either, but this is aligned with what I do in my local environment (either, Windows, MacOS or Ubuntu on WSL) to avoid dependencies being downloaded for every command. And it is not only the double activation: if the executable is not called before the second activation, it has no effect. I tried multiple combinations, including only calling `pub get` instead, but this `activate` -> `call` -> `activate` was the only one that solved.

Issue #4007, opened 2025-09-26
create-migration blows up if you set a default value in a .yaml ORM file that's not a string


parnesen:
### Describe the bug

I tried to add a new column to a table with: 

  tombstoned:
    type: bool
    default: false

and got this:

PS D:\code\dart\repos\tapestry\dart\tapestry_library_server> serverpod create-migration
⠹ Creating migration... (90ms)
Yikes! It is possible that this error is caused by an internal issue with the Serverpod tooling. We would appreciate if you filed an issue
over at Github. Please include the stack trace below and describe any steps you did to trigger the error.
https://github.com/serverpod/serverpod/issues

type 'bool' is not a subtype of type 'String?'

#0      getColumnDefault (package:serverpod_cli/src/database/create_definition.dart:155:14)
#1      createDatabaseDefinitionFromModels (package:serverpod_cli/src/database/create_definition.dart:36:34)
#2      MigrationGenerator.createMigration (package:serverpod_cli/src/migrations/generator.dart:77:37)
<asynchronous suspension>
#3      CreateMigrationCommand.runWithConfig.<anonymous closure> (package:serverpod_cli/src/commands/create_migration.dart:93:21)
<asynchronous suspension>
#4      StdOutLogger.progress (package:cli_tools/src/logger/loggers/std_out_logger.dart:132:20)
<asynchronous suspension>
#5      CreateMigrationCommand.runWithConfig (package:serverpod_cli/src/commands/create_migration.dart:91:5)
<asynchronous suspension>
#6      CommandRunner.runCommand (package:args/command_runner.dart:212:13)
<asynchronous suspension>
#7      BetterCommandRunner.runCommand (package:cli_tools/src/better_command_runner/better_command_runner.dart:299:14)
<asynchronous suspension>
#8      ServerpodCommandRunner.runCommand (package:serverpod_cli/src/runner/serverpod_command_runner.dart:100:5)
<asynchronous suspension>
#9      _main (file:///C:/Users/patri/AppData/Local/Pub/Cache/hosted/pub.dev/serverpod_cli-2.9.1/bin/serverpod_cli.dart:59:5)
<asynchronous suspension>
#10     main.<anonymous closure> (file:///C:/Users/patri/AppData/Local/Pub/Cache/hosted/pub.dev/serverpod_cli-2.9.1/bin/serverpod_cli.dart:32:9)
<asynchronous suspension>
[...]

PR #4011, opened 2025-09-29
fix: Analyzer type parsing handled consistently under the hood

Isakdl:
This PR aims to resolve a bug where the analyzer and code generator would crash if the wrong type was interpreted from the yaml definition. This could happen because of a discrepancy in how a Yaml map and a single line string was parsed. 

To resolve issue the single line resolver was modified to always return the same datatype as the normal yaml parser would. Meaning, return an int, bool, or double if possible, otherwise the value is interpreted as a string. Previously all values was interpreted as strings.

The idea is to consolidate the two different styles to behave exactly the same so that this type of error cannot occur in the future. 

Example:

```yaml
# Default values previously interpreted as Strings, but as the proper types after this PR.
class: Example
fields:
   name: String, default=Hello
   age: int, default=0
   ratio: double, default=0.5
   happy: bool, default=true
```

```yaml
# Default values interpreted as their respective types.
class: Example
fields:
   name: 
     type: String
     default: Hello
   age: 
     type: int
     default: 0
   ratio:
      type: double
      default: 0.5
   happy: 
     type: bool
     default: true
```

This change lead to many downstream effects especially in many tests as they where validating the wrong behaviours.

I have also added tests for BigInt and Uri default value types as these tests were missing from the test suites.

*Review notes*: I have not dealt with the redix problem when parsing ints, @SandPod do you have an example case where the parsing would fail? I think guarding against the problem with a test is a must if we add it!

Closes: https://github.com/serverpod/serverpod/issues/4007

## Pre-launch Checklist

## Breaking changes

There is a breaking change in how BigInt values that are larger than what fits into a normal int. These values must now be quoted or else the user will get a lint error saying that the value cannot be parsed. For values that fits into an int we allow them to be converted under the hood.

This is only a behaviour change for a single line value, the yaml map structure already had this limitation previously.

Example:

```yaml
class: Example
fields:
   age: BigInt, default=0 # This is still okay
   errorMass: BigInt, default=9999999999999999999 # This is now considered an error
   validMass: BigInt, default="9999999999999999999" # This is valid
```

### DIFF
--- a/tools/serverpod_cli/lib/src/database/create_definition.dart
+++ b/tools/serverpod_cli/lib/src/database/create_definition.dart
@@ -203,5 +203,6 @@ String? getColumnDefault(
 /// _escapeSqlString("This is a \"default\" persist value") // Returns "This is a ""default"" persist value"
 /// ```
 String _escapeSqlString(String value) {
-  return "'${value.replaceAll("\\'", "'").replaceAll("\\\"", '"')}'";
+  return '\$\$$value\$\$';
+  return "${value.replaceAll("\\'", "'").replaceAll("\\\"", '"')}'";
 }
--- a/tools/serverpod_cli/test/database/create_definition/defaults/string_default_definition_test.dart
+++ b/tools/serverpod_cli/test/database/create_definition/defaults/string_default_definition_test.dart
@@ -9,7 +9,7 @@ void main() {
     group('when "defaultPersist" is set', () {
       var field = FieldDefinitionBuilder()
           .withName('stringDefault')
-          .withTypeString(defaultPersistValue: "'This is a default value'")
+          .withTypeString(defaultPersistValue: 'This is a default value')
           .build();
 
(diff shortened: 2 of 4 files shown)


SandPod:
@Isakdl - I drafted out the cases I could think of, and we can observe different behavior for hexadecimal values based on the radix:

```dart
void main() {
  print(
      '1a: ${int.tryParse("010")}'); // Could be 10 (decimal) or 8 (octal) (Returns 10)
  print('1b: ${int.tryParse("010", radix: 10)}'); // Always returns 10

  print('2a: ${int.tryParse("0x10")}'); // Returns 16 (hexadecimal)
  print(
      '2b: ${int.tryParse("0x10", radix: 10)}'); // Returns null (invalid decimal)

  print('3a: ${int.tryParse("FF")}'); // Returns null (invalid decimal)
  print(
      '3b: ${int.tryParse("FF", radix: 10)}'); // Returns null (invalid decimal)
}
```

PR #4023, opened 2025-10-01
feat: Add abstract endpoint inheritance on client

### LINKED ISSUE #4000: Allow users to create code that dependes on the generated client for abstract endpoints
### Problem to Solve

As abstract endpoints are currently not generated on the client, it is not possible for users to create code that depends on concrete implementations of such endpoints.

### Proposal

To solve this, abstract endpoints can generate a corresponding abstract/interface class on the client. Concrete implementations of such endpoint will also be reflected on client classes that subclass/implements the abstract generated class.

This will allow proper inheritance on endpoints, since the client hierarchy will closely mirror the server code, as would be expected. The only exception would be allowing users to mark an abstract class as `@doNotGenerate` to opt-out from the abstract client generation. In this case, the classes will behave as today, with the generated class as a fully independent class.

So, an implementation must ensure:
- Generating abstract client classes/interfaces when server endpoint classes are abstract.
- Excluding server-only base endpoint classes (annotated with @doNotGenerate) from client code entirely.
- Maintaining correct client-side inheritance: concrete endpoints should extend generated abstract base classes.

### Use Case

When building the new authentication module, provider endpoints must only exist on the client if the user explicitly wants to. This will be possible by making endpoint classes abstract and requiring the user to subclass the ones they want to be available. However, on the flutter package, we need to depend on the generated client interfaces to create widgets for each provider.

The solution for this problem will be to always generate an abstract class on the client when the endpoint is declared as abstract and make all descendants of this endpoint subclass it. This way, users can create code that depend on the concrete implementation of the class.

### Alternatives

There is no current way to allow writing code that depends on a generated client without having the endpoint classes declared as concrete.

### Additional context

_No response_

- [x] I want to work on this

marcelomendoncasoares:
Closes #4000.

With this PR, we add support for proper endpoint inheritance, which means that generated client endpoint classes now mirror the original endpoint inheritance chain.

- Abstract endpoint classes now generate abstract client endpoint classes.
- Concrete endpoint classes will inherit from generated client endpoint superclasses.

This means that module developers can declare base abstract classes and leave up to the user to extend them and expose the endpoints on their server. As such endpoints will inherit from the abstract class, the module can have code that depend on it.

A helper method `getEndpointOfType<T extends EndpointRef>` was also added to the `EndpointCaller` class (base for the `ServerpodClientShare`) to facilitate retrieving concrete endpoints by their type.

To opt out of the abstract client endpoint class generation, users can annotate endpoints with `@doNotGenerate`. Then, classes that inherit from this endpoint will be generated as extending `EndpointRef` only - ignoring the inheritance chain.

## Pre-launch Checklist

## Breaking changes

None.


tools/serverpod_cli/lib/src/analyzer/dart/definitions.dart
```diff
@@ -23,6 +25,12 @@ class EndpointDefinition {
   /// The annotations of this endpoint.
   final List<AnnotationDefinition> annotations;
 
+  /// Whether this endpoint is abstract.
+  final bool isAbstract;
```
SandPod, line 29: Question: Should this be a "type" of `EndpointDefinition` instead? Perhaps a subclass?

tools/serverpod_cli/test/test_util/builders/endpoint_definition_builder.dart
```diff
@@ -31,6 +34,11 @@ class EndpointDefinitionBuilder {
     return this;
   }
 
+  EndpointDefinitionBuilder withExternalModule(String serverPackageName) {
```
SandPod: Suggestion: Add a helper method to build the correct file path instead of this parameter.

This is an abstract construction that makes the `build` method ignore a specific set value for `_filePath`. Perhaps the module could simply be an optional parameter to `withFilePath` or a static `moduleFilePath` method could be added that takes the two parameters?

tools/serverpod_cli/test/generator/dart/client_code_generator/abstract_endpoint_client_test.dart
```diff
@@ -0,0 +1,629 @@
...
+          .withName(streamMethodName)
+          .buildMethodStreamDefinition(),
+    ]).build();
+
+    var protocolDefinition = ProtocolDefinition(
+      endpoints: [abstractEndpoint],
+      models: [],
+    );
+
+    var codeMap = generator.generateProtocolCode(
```
SandPod: Suggestion: Mark these as late so that we don't evaluate it until it is actually needed. 

This allows us to debug a single scenario in the test file and place debuggers in the code generator that are only triggered by a specific test.

SandPod: Apply these to all added generator tests.

tools/serverpod_cli/test/generator/dart/client_code_generator/abstract_endpoint_client_test.dart
```diff
@@ -0,0 +1,629 @@
...
+        expect(concreteClass!.abstractKeyword, isNull);
+      });
+
+      test('extends external module abstract base endpoint.', () {
+        var baseEndpoint = 'Endpoint${baseEndpointName.pascalCase}';
+        var concreteEndpoint = 'Endpoint${concreteEndpointName.pascalCase}';
+
+        expect(
+          codeMap[expectedFileName],
+          contains('$concreteEndpoint extends _i1.$baseEndpoint'),
```
SandPod: Suggestion: Remove `_i1` from validation.

This makes the test depend on import order and will break the test if the import order ever changes. Please make this test not depend on import order.

SandPod: Fix in all places if the comment applies in other places.

tools/serverpod_cli/test/generator/dart/client_code_generator/abstract_endpoint_client_test.dart
```diff
@@ -0,0 +1,629 @@
...
+      expect(codeMap, contains(expectedFileName));
+    });
+
+    test('then the generated client file has the endpoint class defined.', () {
+      var clientCompilationUnit =
+          parseString(content: codeMap[expectedFileName]!).unit;
+
+      var endpointClass = CompilationUnitHelpers.tryFindClassDeclaration(
+        clientCompilationUnit,
+        name: 'Endpoint${endpointName.pascalCase}',
```
SandPod, line 69: Comment: Never thought about that we flip the order of the name when we generate it on the client 😅

tools/serverpod_cli/test/generator/dart/client_code_generator/abstract_endpoint_client_test.dart
```diff
@@ -0,0 +1,629 @@
...
+        expect(
+          codeMap[expectedFileName],
+          contains('$concreteEndpoint extends _i1.$baseEndpoint'),
+        );
+      });
+    });
+  });
+
+  group(
+      'Given protocol definition with abstract > concrete > abstract > concrete endpoint hierarchy when generating client file',
```
SandPod: Suggestion(NIT): Please reflect the subsclass in the description.

This makes the description a bit easier to understand.

`abstract > concrete > abstract subclass > concrete subclass`

SandPod: Apply this to the server, test tools and endpoint description test.

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/abstract_endpoint_analysis_test.dart
```diff
@@ -0,0 +1,280 @@
...
+    });
+
+    test('then concrete endpoint has both base and concrete methods.', () {
+      expect(
+        concreteEndpoint.methods.map((m) => m.name).toSet(),
+        {'baseMethod', 'concreteMethod', 'overriddenMethod'},
+      );
+    });
+
+    test(
```
SandPod, line 179: Suggestion: Break this out into it's own test group with a separate setup and note down the condition in the `Given` part of the test description.

That will allow us to understand if everything except the override is behaving as expected in the analyzer. (For example if overrides start throwing exceptions)

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_class_analyzer.dart
```diff
@@ -3,28 +3,95 @@ import 'package:serverpod_cli/src/analyzer/code_analysis_collector.dart';
...
+  /// another endpoint class, that parent class is also parsed and added to the
+  /// [endpointDefinitions] list. [EndpointDefinition]s that have already been
+  /// parsed are not parsed again.
+  ///
+  /// For [EndpointDefinition]s that are not part of this package, the field
+  /// [EndpointDefinition.filePath] will be set to the library uri.
+  static void parse(
     ClassElement element,
-    List<MethodDefinition> methodDefinitions,
+    Map<String, List<SourceSpanSeverityException>> validationErrors,
```
SandPod, line 22: Suggestion: Should we also document that we also modify the `validationErrors` argument?

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_class_analyzer.dart
```diff
@@ -102,6 +170,50 @@ abstract class EndpointClassAnalyzer {
   }
 }
 
+extension on ClassElement {
+  /// Returns all endpoints methods from the class.
+  ///
+  /// Those defined directly on the class, as well as those inherited from the base classes.
+  List<MethodElement> collectionEndpointMethods({
```
SandPod: Suggestion(NIT): Rename to `collectEndpointMethods`.

Seems like that is what we actually want here right?

tools/serverpod_cli/lib/src/analyzer/dart/endpoints_analyzer.dart
```diff
@@ -21,10 +22,12 @@ class EndpointsAnalyzer {
 
   final String absoluteIncludedPaths;
 
+  final GeneratorConfig config;
```
SandPod: Question: When inspecting commit 2ff55e3df409f07c18d6a892f4628472ddf90ddd it doesn't seem like this is used anywhere?

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -874,6 +909,9 @@ class LibraryGenerator {
       p.joinAll([...config.generatedServeModelPathParts]);
 
   String _endpointPath(EndpointDefinition endpoint) {
+    // For endpoints defined in other packages, the filePath is the library uri.
+    if (endpoint.filePath.startsWith('package:')) return endpoint.filePath;
```
SandPod, line 913: Discussion: `package:` could be used for imports inside a package. Do we handle this?

tests/serverpod_test_client/test/endpoint_inheritance_test.dart
```diff
@@ -0,0 +1,154 @@
...
+
+  test(
+      'Given an abstract endpoint class that extends a concrete endpoint class from a module'
+      'then it is subclass of the generated module concrete class.', () {
+    expect(isSubClass<EndpointAbstractModuleSubClass, m.EndpointConcreteBase>(),
+        isTrue);
+  });
+}
+
+bool isSubClass<S, T>() => <S>[] is List<T>;
```
SandPod, line 154: Comment: Is there no normal matcher for this? Couldn't `isA<T>()` be used for this?

The way this is used, it will not provide a good error message when it fails.
`expect(isSubClass<EndpointAbstractModuleBase, m.EndpointAbstractBase>(), isTrue);` will just return expected `true` but got `false`.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -675,6 +675,26 @@ abstract class EndpointCaller {
...
+
+  /// Returns an endpoint of type [T]. If more than one endpoint of type [T]
+  /// exists, [name] can be used to disambiguate.
+  T getEndpointOfType<T extends EndpointRef>([String? name]) {
+    if (name != null) return endpointRefLookup[name] as T;
+
+    var foundEndpoints = endpointRefLookup.values.whereType<T>();
+    switch (foundEndpoints.length) {
+      case 0:
+        throw StateError('No endpoint of type $T found.');
```
SandPod: Suggestion: Throw a specific exception for this case.

This allows module developers to develop a widget that can automatically work if their endpoint lookup passes, while it can be disabled if the endpoints are not found.

Since `StateError` should never be catched it would be good to introduce a specific error for this case that can be catched.

SandPod (changes requested):
Great work! I agree with the implementation and the approach.

I have a couple of fixes, questions, and suggestions.

We should complement this with a documentation page.

tools/serverpod_cli/test/test_util/builders/endpoint_definition_builder.dart
marcelomendoncasoares: I like the approach of using a module parameter in the `withFilePath`, since it will end up in this field after all. Done at https://github.com/serverpod/serverpod/pull/4023/commits/80e3df0da4263b89082cb14f3643884c733796b2.

tools/serverpod_cli/test/generator/dart/client_code_generator/abstract_endpoint_client_test.dart
marcelomendoncasoares: This is kind of generalized problem with the code generator tests. I'll fix all occurrences on test files touched by this PR and keep this in mind to slowly refactor tests touched on future PRs.

tools/serverpod_cli/test/generator/dart/client_code_generator/abstract_endpoint_client_test.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4023/commits/9817882b5f9c01d1a0a9b4e7108f52ac15bfac55.

tools/serverpod_cli/test/generator/dart/client_code_generator/abstract_endpoint_client_test.dart
marcelomendoncasoares: It is a nice solution to avoid name clashing, but one more reason to have our helper method to get the endpoint by superclass on the auth modules 😆

tools/serverpod_cli/test/generator/dart/client_code_generator/abstract_endpoint_client_test.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4023/commits/4c80811506b65789978379f7cb3828acedb35c63.

tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/abstract_endpoint_analysis_test.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4023/commits/a8bcd0381d93fda151f67867673f98ae00e5ba2e.

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_class_analyzer.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4023/commits/575cf3e3cb56902b8619ea2406d711020e12f297.

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_class_analyzer.dart
marcelomendoncasoares: It was the existing name and I just moving without judgment, but you are completely right. Also done at https://github.com/serverpod/serverpod/pull/4023/commits/575cf3e3cb56902b8619ea2406d711020e12f297.

tools/serverpod_cli/lib/src/analyzer/dart/endpoints_analyzer.dart
marcelomendoncasoares: Good catch! It was used in an intermediary solution to get external module references, but I managed to remove its need and forgot the parameter there. Will make the PR a lot clean. Removed at https://github.com/serverpod/serverpod/pull/4023/commits/f283bcef7960d12e806c48b57bedc13385d5e4a5.

tests/serverpod_test_client/test/endpoint_inheritance_test.dart
nielsenko: @SandPod `isA<T>` works on an instance. `isSubClass<S, T>` works on type arguments. I'm not sure if that is important here, just wanted to highlight the difference.

marcelomendoncasoares: Unfortunately there isn't, because in this test we are asserting the type itself, not an object. As the class is abstract, we can't create an object to use the `isA<T>()` matcher - as we are doing on the concrete classes tests. And creating a mock class just for testing with the matcher seemed dirtier to me, since it would be kind of the same as the subclasses.

I agree that this is a less-than-ideal experience, but it also seemed better to me to have this test so we can assert the abstract classes' superclasses.

This solution is proposed [in a dart-lang discussion about checking Type inheritance](https://github.com/dart-lang/sdk/issues/43390#issuecomment-690993957) and it is one of the few ways to to it without `dart:mirrors`.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
marcelomendoncasoares: Good point. Done at https://github.com/serverpod/serverpod/pull/4023/commits/9627ddc094ff0ef5a1e39403ab6d0d8d234b9037.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
marcelomendoncasoares: Currently, we have no such usage. The `filePath` is only used as actual file paths, originated from the `_libraries` getter on the `EndpointsAnalyzer` class:

```dart
  Stream<(ResolvedLibraryResult, String)> get _libraries async* {
    for (var context in collection.contexts) {
      var analyzedFiles = context.contextRoot.analyzedFiles().toList(); // <-- This is the source.
      analyzedFiles.sort();
      var analyzedDartFiles = analyzedFiles
          .where((path) => path.endsWith('.dart'))
          .where((path) => !path.endsWith('_test.dart'));
      for (var filePath in analyzedDartFiles) {
        var library = await context.currentSession.getResolvedLibrary(filePath);
        if (library is ResolvedLibraryResult) {
          yield (library, filePath);
        }
      }
    }
  }
```

From this getter, the list of existing file paths will be propagated to the analyzer and parsed classes will use the received `filePath` instead of references from imports. It can be confirmed by existing tests that use the `filePath` field, which only set it to actual paths.

So the only expected usage for it starting with `package:` would be if the parsed endpoint is coming from outside of the current project, in which case we want the import on the generated file to not be relative. Such endpoints will also be striped from the list of generated endpoints, as they are only evaluated for inheritance resolution.

tools/serverpod_cli/test/generator/dart/client_code_generator/abstract_endpoint_client_test.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4023/commits/99bd2db306d0f34433548808da8a39c308a99880.

tools/serverpod_cli/lib/src/analyzer/dart/definitions.dart
marcelomendoncasoares: I gave some more thought and experimentation with the `AbstractEndpointDefinition` alternative, but it would require considerably more code to do the same as the `isAbstract` bool:

- The subclass would serve merely to do `is` and `whereType` checks, since we need all fields of `EndpointDefinition`.
- The code below to add an endpoint definition would have to be duplicated to create either one of the classes.
- Relying on the class type to do the bool comparison would likely introduce more variables and casts.

I think it would be a better approach if we had a different set of fields to store in case of abstract endpoints, but it isn't the case - at least for now. Since `abstract` is a class modifier and it is already exposed on the same `isAbstract` bool field of the `ClassElement` type, I'd say we are better served with a simple bool.

```dart
    endpointDefinitions.add(
      EndpointDefinition(
        name: endpointName,
        documentationComment: classDocumentationComment,
        className: className,
        methods: _parseEndpointMethods(element, validationErrors, filePath),
        filePath: filePath,
        annotations: annotations,
        isAbstract: element.isAbstract,  // <-- Here
        extendsClass: parentEndpointDefinition,
      ),
    );
```

SandPod: Sounds good to me! Thanks for investigating this.

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_class_analyzer.dart
SandPod: I think I probably introduced it 🙃 , thanks for cleaning this up 🧹

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
SandPod: Thanks for clarifying this! 👍

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -723,3 +739,39 @@ abstract class EndpointRef {
     _streamController = StreamController<SerializableModel>();
   }
 }
+
+/// Thrown if not able to get an endpoint on the client by type.
+abstract class ServerpodClientGetEndpointException implements Exception {
```
SandPod: Discussion: Should we make this sealed and the implementations final so that the user can easily see what implementations of the class exists?

It increases discoverability, but it also means that if we introduce a new exception type, that would be a breaking change...

tests/serverpod_test_client/test/endpoint_inheritance_test.dart
SandPod: Aaah, thank you for clarifying this to me! 👍

Nice find for solving this.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
SandPod: Perhaps the concrete classes should be marked as `final` either way, to ensure we can modify them without introducing breaking changes?

marcelomendoncasoares: Since these extensions are pretty well scoped and we have little expectation of changing, I think the discoverability that the `sealed` provides will be a better benefit. And it also makes clear for developers to treat new exceptions in case we add any. Done at https://github.com/serverpod/serverpod/pull/4023/commits/771647497cdd0732fc44250a1cba53212132f5d3.

PR #4027, opened 2025-10-02
fix: Allow inheritance on `id` field for table models

### LINKED ISSUE #3939: Inheritance of `id` field changes not working for table classes
### Describe the bug

When a table class inherits from a class that changes `id` field type, relations to the subclass will not recognize the type of the `id` field correctly.

### To reproduce

1. Create a class with no table and a changed id.
  ```yaml
  class: ParentWithChangedId
  fields:
    id: UuidValue, defaultModel=random_v7
  ```

2. Create a child class that inherits from it and has a table defined.
  ```yaml
  class: ChildWithInheritedId
  extends: ParentWithChangedId
  table: child_with_inherited_id
  fields:
    name: String
    parent: ChildWithInheritedId?, relation(field=parentId)
    parentId: UuidValue?
  ```

3. Call `serverpod generate --experimental-features=all` and see the analyzer error.
  ```
  ERROR: Found 1 issue.
  Error on line 6, column 49 of lib\src\models\inheritance\child_with_inherited_id.spy.yaml: The field "parentId" is of type "UuidValue" but reference field "id" is of type "int".
    ╷
  1 │ parentId
    │ ^^^^^^^^
    ╵
  ```

### Expected behavior

As occur with all other fields, inheritance should also work for the `id` field. This enables changing the primary key for a group a tables that share a common base and/or audit fields.

### Library version

2.9.0 - 3.0.0-alpha.1

### Platform information

Windows x64

### Additional context

This error occurs because the `id` field is evaluated on the `ModelParser._parseClassFields` method before model dependencies are resolved on `ModelDependencyResolver.resolveModelDependencies`. The fix is to also defer the `id` field resolution to **after** the dependency resolution.

- [x] I want to work on this

marcelomendoncasoares:
Fixes #3939.

To allow inheritance of the `id` field, its parsing had to be moved to after the inheritance resolution on the `ModelDependencyResolver.resolveModelDependencies`.

Since the `id` field has strict rules - that only applies if the model has the `table` keyword - and inheriting it bypasses all validations executed during the YAML parsing, such validations were enforced on the `table` value parsing of child models. This provides a natural experience for the user that will see the error highlighted on the `table` keyword of the model due to incompatibilities of the `id` field defined in parent classes.

## Pre-launch Checklist

## Breaking changes

N/A.

### DIFF
--- a/tests/serverpod_test_client/lib/src/protocol/inheritance/child_with_inherited_id.dart
+++ b/tests/serverpod_test_client/lib/src/protocol/inheritance/child_with_inherited_id.dart
@@ -1,0 +1,119 @@
+/* AUTOMATICALLY GENERATED CODE DO NOT MODIFY */
+/*   To generate run: "serverpod generate"    */
+
+// ignore_for_file: implementation_imports
+// ignore_for_file: library_private_types_in_public_api
+// ignore_for_file: non_constant_identifier_names
+// ignore_for_file: public_member_api_docs
+// ignore_for_file: type_literal_in_constant_pattern
+// ignore_for_file: use_super_parameters
+// ignore_for_file: invalid_use_of_internal_member
+
+// ignore_for_file: no_leading_underscores_for_library_prefixes
+import '../protocol.dart' as _i1;
+import 'package:serverpod_client/serverpod_client.dart' as _i2;
+import '../inheritance/child_with_inherited_id.dart' as _i3;
+
+abstract class ChildWithInheritedId extends _i1.ParentWithChangedId
+    implements _i2.SerializableModel {
+  ChildWithInheritedId._({
+    _i2.UuidValue? id,
+    required this.name,
+    this.parent,
+    this.parentId,
+  }) : id = id ?? _i2.Uuid().v7obj();
+
+  factory ChildWithInheritedId({
+    _i2.UuidValue? id,
+    required String name,
+    _i3.ChildWithInheritedId? parent,
+    _i2.UuidValue? parentId,
+  }) = _ChildWithInheritedIdImpl;
+
+  factory ChildWithInheritedId.fromJson(
+      Map<String, dynamic> jsonSerialization) {
+    return ChildWithInheritedId(
+      id: _i2.UuidValueJsonExtension.fromJson(jsonSerialization['id']),
+      name: jsonSerialization['name'] as String,
+      parent: jsonSerialization['parent'] == null
+          ? null
+          : _i3.ChildWithInheritedId.fromJson(
+              (jsonSerialization['parent'] as Map<String, dynamic>)),
+      parentId: jsonSerialization['parentId'] == null
+          ? null
+          : _i2.UuidValueJsonExtension.fromJson(jsonSerialization['parentId']),
+    );
+  }
+
+  /// The database id, set if the object has been inserted into the
+  /// database or if it has been fetched from the database. Otherwise,
+  /// the id will be null.
+  _i2.UuidValue id;
+
+  String name;
+
+  _i3.ChildWithInheritedId? parent;
+
+  _i2.UuidValue? parentId;
+
+  /// Returns a shallow copy of this [ChildWithInheritedId]
+  /// with some or all fields replaced by the given arguments.
+  @override
+  @_i2.useResult
+  ChildWithInheritedId copyWith({
+    _i2.UuidValue? id,
+    String? name,
+    _i3.ChildWithInheritedId? parent,
+    _i2.UuidValue? parentId,
+  });
+  @override
+  Map<String, dynamic> toJson() {
+    return {
+      'id': id.toJson(),
+      'name': name,
+      if (parent != null) 'parent': parent?.toJson(),
+      if (parentId != null) 'parentId': parentId?.toJson(),
+    };
+  }
+
+  @override
+  String toString() {
+    return _i2.SerializationManager.encode(this);
+  }
+}
+
+class _Undefined {}
+
+class _ChildWithInheritedIdImpl extends ChildWithInheritedId {
+  _ChildWithInheritedIdImpl({
+    _i2.UuidValue? id,
+    required String name,
+    _i3.ChildWithInheritedId? parent,
+    _i2.UuidValue? parentId,
+  }) : super._(
+          id: id,
+          name: name,
+          parent: parent,
+          parentId: parentId,
+        );
+
+  /// Returns a shallow copy of this [ChildWithInheritedId]
+  /// with some or all fields replaced by the given arguments.
+  @_i2.useResult
+  @override
+  ChildWithInheritedId copyWith({
+    _i2.UuidValue? id,
+    String? name,
+    Object? parent = _Undefined,
+    Object? parentId = _Undefined,
+  }) {
+    return ChildWithInheritedId(
+      id: id ?? this.id,
+      name: name ?? this.name,
+      parent: parent is _i3.ChildWithInheritedId?
+          ? parent
+          : this.parent?.copyWith(),
+      parentId: parentId is _i2.UuidValue? ? parentId : this.parentId,
+    );
+  }
+}
--- a/tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
@@ -121,8 +121,12 @@ final class ModelClassDefinition extends ClassDefinition {
 
   /// Returns a list of all fields in the parent class.
   /// If there is no parent class, an empty list is returned.
+  /// Excludes the id field, as it is re-declared on child classes.
   List<SerializableModelFieldDefinition> get inheritedFields =>
-      parentClass?.fieldsIncludingInherited ?? [];
+      parentClass?.fieldsIncludingInherited
+          .where((element) => tableName == null || element.name != 'id')
+          .toList() ??
+      [];
 
   /// Returns a list of all fields in this class, including inherited fields.
   /// It ensures that the 'id' field, if present, is always included at the beginning of the list.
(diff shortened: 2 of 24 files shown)


tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/id_field_inheritance_test.dart
```diff
@@ -0,0 +1,552 @@
...
+    });
+
+    late var childClass = models.last as ModelClassDefinition;
+
+    test('then child class id field is nullable.', () {
+      expect(childClass.idField.type.nullable, true);
+    });
+
+    test(
+        'then child class id field has defaultPersist set to random and defaultModel set to null.',
```
SandPod: Comment(NIT): The `random` value is introduced here but not reflected in the given. Perhaps we should just reference the parents configured value?

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
```diff
@@ -67,6 +76,56 @@ class ModelDependencyResolver {
...
+    var idFieldType = maybeIdField?.type ?? defaultIdType.type.asNullable;
+
+    var defaultPersistValue = (maybeIdField != null)
+        ? maybeIdField.defaultPersistValue
+        : defaultIdType.defaultValue;
+
+    // The 'int' id type can be specified without a default value.
+    if (maybeIdField?.type.className == 'int') {
+      defaultPersistValue ??= SupportedIdType.int.defaultValue;
+    }
```
SandPod, line 102: Question: Can this be expressed through a switch expression? Just to make it even clearer that we are defining `defaultPersistValue` by only inspecting properties of `maybeIdField`.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -295,7 +299,7 @@ class SerializableModelFieldDefinition {
   /// - [shouldIncludeField]
   /// - [shouldSerializeFieldForDatabase]
   bool shouldSerializeField(bool serverCode) {
-    return scope == ModelFieldScopeDefinition.all;
+    return name == 'id' || scope == ModelFieldScopeDefinition.all;
```
SandPod: Suggestion: It would be better to not have a specific case for the ID field. Can't we always give it scope defintion `all` instead?

Would it be possible to give `serverOnly` for the id field, making this fail?

SandPod: This also feels like a potentially bad hardcoding of this relationship.

tests/serverpod_test_client/lib/src/protocol/inheritance/child_with_inherited_id.dart
```diff
@@ -0,0 +1,119 @@
...
+              (jsonSerialization['parent'] as Map<String, dynamic>)),
+      parentId: jsonSerialization['parentId'] == null
+          ? null
+          : _i2.UuidValueJsonExtension.fromJson(jsonSerialization['parentId']),
+    );
+  }
+
+  /// The database id, set if the object has been inserted into the
+  /// database or if it has been fetched from the database. Otherwise,
+  /// the id will be null.
```
SandPod: Suggestion: Update this comment.

This is no longer true when we use UUID and default values. We might always have a value here regardless of if we have inserted it to the database.

This might not be part of this PR but I noticed now.

tools/serverpod_cli/lib/src/analyzer/models/yaml_definitions/class_yaml_definition.dart
```diff
@@ -47,7 +47,7 @@ class ClassYamlDefinition {
       ValidateNode(
         Keyword.table,
         keyRestriction: restrictions.validateTableNameKey,
-        valueRestriction: restrictions.validateTableName,
+        valueRestriction: restrictions.validateTableValue,
```
SandPod: Suggestion: Add this as a separate restriction.

Isn't this a separate restriction that works on the same key? I don't know if we have a previous example of this. But since this validates both the name and the value I think we should have a name that reflects this or we should split it into two separate restrictions.

SandPod: Field is for example split on `validateFieldName` and `validateFieldType`.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -993,6 +971,101 @@ class Restrictions {
...
+  List<SourceSpanSeverityException> _validateTableInheritedIdField(
+    SourceSpan? span,
+  ) {
+    final model = documentDefinition;
+    if (model is! ModelClassDefinition || !model.isIdInherited) return [];
+
+    var errors = _validateIdFieldDataType(model.idField, span);
+
+    if (model.idField.defaultModelValue != null) {
+      var defaultModelErrors = DefaultValueRestriction(
```
SandPod, line 983: Question: Isn't this already validated somewhere else?

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -30,9 +31,13 @@ class LibraryGenerator {
   Library generateProtocol() {
     var library = LibraryBuilder();
 
-    var allModels = protocolDefinition.models
-        .where((model) => serverCode || !model.serverOnly)
-        .toList();
+    // Models will be sorted by file path and then topologically, to ensure a
+    // stable order and that children gets generated before parents.
+    var allModels = (protocolDefinition.models
```
SandPod, line 36: Suggestion/Discussion: Move this to our parse layer.

Shouldn't the output from our parsing have a stable order directly, rather than patching it when trying to generate? Feels like the filenames shouldn't modify the result from the parsing.

SandPod: Decision: Keep this here as it only relates to the generation of the models.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
SandPod: This is not validated in another place.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
marcelomendoncasoares: After our discussion, I recalled that I have added this to avoid the diff on the new serverpod auth models. In reality, they contain an untested edge case: we are using `serverOnly` models that change the `id` type, and they are all generating the `id` on the `toJsonForProtocol`. Classes like this:

```yaml
class: UserProfileImage
serverOnly: true
table: serverpod_auth_core_profile_image
fields:

  ### The [UserProfile] this image belongs to.
  userProfile: UserProfile?, relation(onDelete=Cascade)

  ### The time when this profile image was created.
  createdAt: DateTime, default=now

  ### Storage in which the image is stored.
  storageId: String

  ### Path inside [storageId] at which the image is stored.
  path: String

  ### The public URL to access the image.
  url: Uri
```

So this is actually another unrelated bug that was caught in this PR. After removing this `shouldSerializeField` change, the `id` serialization gets removed.

image

I'll add a test for this as well.

marcelomendoncasoares: While fixing this, I found even another bug detailed here #4039.

tools/serverpod_cli/lib/src/analyzer/models/yaml_definitions/class_yaml_definition.dart
marcelomendoncasoares: As discussed, I have renamed to `validateTable` at https://github.com/serverpod/serverpod/pull/4027/commits/2248c57b4aad2aba22b4c4f9a7b9f10d8541a528.

tests/serverpod_test_client/lib/src/protocol/inheritance/child_with_inherited_id.dart
marcelomendoncasoares: Although not related to the PR, it doesn't hurt to change. Done at https://github.com/serverpod/serverpod/pull/4027/commits/343b826b1bdb2c8e21823ead991c679221be8f0b.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
marcelomendoncasoares: The first issue was fixed at https://github.com/serverpod/serverpod/pull/4027/commits/39df97a606cd5eb15210cade37b3da3e3cc401e6.

The second found issue #4039 was fixed at https://github.com/serverpod/serverpod/pull/4027/commits/04582746a51f05830e101a249d5bb9bd7b5d4880.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/id_field_inheritance_test.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4027/commits/2f3121d37adc68edc3c7c53ca2ac3024e54afcff.

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
marcelomendoncasoares: I have tried multiple ways to make it more readable and this was the one that I found most concise while still being very clear. Do you have a suggestion on how using a `switch` could improve it?

SandPod: We can ignore this NIT comment.

SandPod:
We also have: https://github.com/serverpod/serverpod/pull/4041 that fixes this specifically.

PR #4030, opened 2025-10-03
chore: Bump relic to 0.6.0 in stable branch

SandPod:
Bumps relic to 0.6.0 in stable alpha branch to include fixes for routing static resources that was patched in Relic.

**Additional fix**

A dependency override was added to contraint `image_cropper_platform_interface` since a breaking for older flutter versions was introduced in version 7.2.0. Issue here: https://github.com/hnvn/flutter_image_cropper/issues/594

## Pre-launch Checklist

## Breaking changes

External interface should have been kept the same.


SandPod:
Serverpod generate failure can safely be ignored since it is some formatting that has shifted with the differnt dart versions.

nielsenko (commented):
Looks kosher except for the added newlines that is causing the generate tests to fail currently.

However I thought you wanted to cut a new 3.0.0-alpha.3 release from main. This doesn't include the `StaticRoute` changes in serverpod?

SandPod:
@nielsenko - No, we will just patch the alpha.1 release since main contains additional breaking changes (also to the new auth packages) that are not important for cloud at the moment.

We anticipate more breaking changes around auth so there is no reason to force that migration at the moment.

PR #4032, opened 2025-10-03
fix: Gracefully handle file match misses.

SandPod:
Fixes an issue where file match misses had changed from a nullable type to a specific router type.

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/packages/serverpod/lib/src/web_server/web_server.dart
+++ b/packages/serverpod/lib/src/web_server/web_server.dart
@@ -141,7 +141,10 @@ class WebServer {
     return switch (match) {
       RouterMatch<Route>() =>
         _handleRouteCall(match.asMatch.value, session, context),
-      _ => context.respond(Response.notFound()),
+      PathMiss<Route>() => context.respond(Response.notFound()),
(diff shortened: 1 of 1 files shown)


packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -137,13 +137,12 @@ class WebServer {
...
-      final route = match.asMatch.value;
-      return await _handleRouteCall(route, session, context);
-    }
 
-    // No matching patch found
-    return context.respond(Response.notFound());
+    return switch (match) {
+      RouterMatch<Route>() =>
+        _handleRouteCall(match.asMatch.value, session, context),
+      _ => context.respond(Response.notFound()),
```
nielsenko: You could make a distinction on 404 vs 405 here.

nielsenko (approved):

packages/serverpod/lib/src/web_server/web_server.dart
SandPod: Fixed! 👍

PR #4036, opened 2025-10-06
ci: Add tests for package publish readiness

SandPod:
Introduces a new CI check that validates package publish readiness and fixes an import issue where UUID would be imported from the UUID package on the client instead of through `serverpod_client`.

### Why inspecting the output for ERRORS
Unfortunately, `dart pub publish --dry-run` does not support ignoring `warnings`, meaning that it will exit with exit code `1` if there are any warnings found. And using a strict dependency version is flagged with a warning in the publish check.

Issue for `dart pub publish` here: https://github.com/dart-lang/pub/issues/3807

Therefore, we won't use the exit code to assert if the check failed, but rather a manual `ERROR: ` inspection.

This also makes it possible for us to use `melos` even though it exists with exit code `65`.

### Example

Example of a failed run when the licence file was removed: https://github.com/serverpod/serverpod/actions/runs/18306991635/job/52126422370?pr=4036

### Additional fix

This PR also includes a fix for when the generated client would import `UuidValue` from `package:uuid/uuid_value.dart` or `package:uuid/uuid.dart` instead of from `package:serverpod_client/serverpod_client.dart` that made the check fail.

Fixes: https://github.com/serverpod/serverpod/issues/4040.

**Update** - This was merged as a separate fix here: #4041 

## Pre-launch Checklist

## Breaking changes

None - CI fix plus an import fix in the generated client, but this requires no changes from the user.


SandPod:
@nielsenko - It needs to include the fix in order to pass the check. Would you rather see that change introduced in a separate PR?

nielsenko:
> @nielsenko - It needs to include the fix in order to pass the check. Would you rather see that change introduced in a separate PR?

Yes, but I missed that the PR included a fix on first pass. I'm okay with keeping it in one.

nielsenko (approved):

SandPod:
Great! I updated the description to reflect this change as well.

Will do a last sync with @marcelomendoncasoares before merging since he introduced a similair fix in his PR.

SandPod:
Ended up merging the fix for the imports in a separate PR. This PR now only contains the new CI pipeline.

nielsenko (approved):
LGTM 🎸

Issue #3644, opened 2025-06-13
Add Flutter package for new email authentication endpoint


tp:
The tricky part might be how to set the endpoint, as the one in `serverpod_auth_email` is abstract, and thus none is generated in its client package.

PR #4058, opened 2025-10-14
refactor!: Refactor e-mail IDP provider to support Flutter client package

marcelomendoncasoares:
This PR enables #3644.

### Account registration and Password reset now return a request ID

To allow a simple and secure verification process, we now return a request ID for both account registration and password reset requests. This request ID can be combined with the verification code sent via email to verify the request. As the ID is also sent via email, more advanced use cases such as magic links are still possible. For invalid requests, this ID will be a random UUID. 

### Hide account not found exceptions from the client

To enhance security, the system no longer throws any kind of "account not found" exception when a user tries to register an already existing email or reset the password for a non-existing email. Instead, it behaves as if the request was successful and throws a generic "unauthorized" error message when trying to complete the request. This change ensures that attackers cannot determine whether an email is registered in the system.

### E-mail exception and reason refactor

Exceptions for email accounts have been renamed for better clarity and consistency, together with their reasons. Serialization has also been changed to use names instead of indexes, improving readability and maintainability. All exception reason enums now have a default value for future-proofing.
 
## Pre-launch Checklist

## Breaking changes

- Major refactor on all email exception class names and their reasons. Will impact try-catch blocks that treat email exceptions on the server or client.
- All email exception reason enum values now serialize by name instead of by index. Will break any code that serializes/deserializes or stores these values directly.
- The exception of account not found is no longer thrown when trying to reset a password for an email that does not exist or register an email that exists. Will break client code that relies on these exceptions to show specific error messages.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/models/exceptions/email_account_password_reset_exception_reason.spy.yaml
+++ b/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/models/exceptions/email_account_password_reset_exception_reason.spy.yaml
@@ -3,11 +3,11 @@ enum: EmailAccountPasswordResetExceptionReason
 serialized: byName
 default: unknown
 values:
-  ### Exception to be thrown when the password reset request was used after
-  ### it has expired.
+  ### Exception to be thrown when attempting to complete a valid password reset
+  ### request that has already expired.
   - expired
-  ### Exception to be thrown when an attempt was made to complete a password
-  ### reset with an invalid verification code.
+  ### Exception to be thrown when attempting to complete a password reset with
+  ### an invalid verification code or password reset request id.
   - invalid
   ### Exception to be thrown when attempting to set a password which does not
   ### match the configured policy.
--- a/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_client/lib/src/protocol/providers/email/models/exceptions/email_account_password_reset_exception_reason.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_client/lib/src/protocol/providers/email/models/exceptions/email_account_password_reset_exception_reason.dart
@@ -14,12 +14,12 @@ import 'package:serverpod_client/serverpod_client.dart' as _i1;
 
 /// The reason for why the password reset request was rejected.
 enum EmailAccountPasswordResetExceptionReason implements _i1.SerializableModel {
-  /// Exception to be thrown when the password reset request was used after
-  /// it has expired.
+  /// Exception to be thrown when attempting to complete a valid password reset
+  /// request that has already expired.
   expired,
 
-  /// Exception to be thrown when an attempt was made to complete a password
-  /// reset with an invalid verification code.
+  /// Exception to be thrown when attempting to complete a password reset with
+  /// an invalid verification code or password reset request id.
   invalid,
 
   /// Exception to be thrown when attempting to set a password which does not
--- a/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_client/lib/src/protocol/providers/email/models/exceptions/email_account_request_exception_reason.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_client/lib/src/protocol/providers/email/models/exceptions/email_account_request_exception_reason.dart
@@ -14,12 +14,12 @@ import 'package:serverpod_client/serverpod_client.dart' as _i1;
 
 /// The reason for why the account creation with email request was rejected.
 enum EmailAccountRequestExceptionReason implements _i1.SerializableModel {
-  /// Exception to be thrown when one attempts to complete an email account
-  /// request after it has expired.
+  /// Exception to be thrown when attempting to complete a valid email account
+  /// request that has already expired.
   expired,
 
-  /// Exception to be thrown when the verification code given for an email
-  /// account request is not valid.
+  /// Exception to be thrown when attempting to complete an email account
+  /// request with an invalid verification code or account request id.
   invalid,
 
   /// Exception to be thrown when attempting to set a password which does not
--- a/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/generated/providers/email/models/exceptions/email_account_password_reset_exception_reason.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/generated/providers/email/models/exceptions/email_account_password_reset_exception_reason.dart
@@ -14,12 +14,12 @@ import 'package:serverpod/serverpod.dart' as _i1;
 
 /// The reason for why the password reset request was rejected.
 enum EmailAccountPasswordResetExceptionReason implements _i1.SerializableModel {
-  /// Exception to be thrown when the password reset request was used after
-  /// it has expired.
+  /// Exception to be thrown when attempting to complete a valid password reset
+  /// request that has already expired.
   expired,
 
-  /// Exception to be thrown when an attempt was made to complete a password
-  /// reset with an invalid verification code.
+  /// Exception to be thrown when attempting to complete a password reset with
+  /// an invalid verification code or password reset request id.
   invalid,
 
   /// Exception to be thrown when attempting to set a password which does not
--- a/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/generated/providers/email/models/exceptions/email_account_request_exception_reason.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/generated/providers/email/models/exceptions/email_account_request_exception_reason.dart
@@ -14,12 +14,12 @@ import 'package:serverpod/serverpod.dart' as _i1;
 
 /// The reason for why the account creation with email request was rejected.
 enum EmailAccountRequestExceptionReason implements _i1.SerializableModel {
-  /// Exception to be thrown when one attempts to complete an email account
-  /// request after it has expired.
+  /// Exception to be thrown when attempting to complete a valid email account
+  /// request that has already expired.
   expired,
 
-  /// Exception to be thrown when the verification code given for an email
-  /// account request is not valid.
+  /// Exception to be thrown when attempting to complete an email account
+  /// request with an invalid verification code or account request id.
   invalid,
 
   /// Exception to be thrown when attempting to set a password which does not
--- a/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/models/exceptions/email_account_request_exception_reason.spy.yaml
+++ b/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/models/exceptions/email_account_request_exception_reason.spy.yaml
@@ -3,11 +3,11 @@ enum: EmailAccountRequestExceptionReason
 serialized: byName
 default: unknown
 values:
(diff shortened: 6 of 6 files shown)


modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_accounts.dart
```diff
@@ -338,15 +338,21 @@ abstract final class EmailAccounts {
...
   /// Each reset request will be logged to the database outside of the
   /// [transaction] and can not be rolled back, so the throttling will always be
   /// enforced.
   ///
   /// Throws an [EmailAccountPasswordResetRequestTooManyAttemptsException] in
   /// case the client or account has been involved in too many reset attempts.
-  static Future<PasswordResetResult> startPasswordReset(
+  static Future<
+      ({
+        PasswordResetResult result,
```
SandPod, line 356: Note: This would also be a good candidate for sealed class inheritance if that lands.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_accounts.dart
```diff
@@ -338,15 +338,21 @@ abstract final class EmailAccounts {
...
+        UuidValue? passwordResetRequestId,
```
SandPod, line 357: Suggestion: Always return a UuidValue.

This allows developers that use this module to not need to keep track of what type of Uuid should be constructred in order to keep the implementation in sync.

We would take that responsibility,

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -28,16 +28,27 @@ void main() {
...
-        await endpoints.emailAccount.startRegistration(
-          sessionBuilder,
-          email: 'test@serverpod.dev',
-          password: 'Foobar123!',
-        );
+        setUpAll(() async {
+          clientReceivedRequestId =
+              await endpoints.emailAccount.startRegistration(
+            sessionBuilder,
+            email: _getNewTestEmail(),
```
SandPod: Suggestion: Remove this and fix the underlying issue.

This shouldn't be needed in a test environment, we should be able to trust what is already in the database. Do we have a leak in some test somewhere? Would in that case be better to fix that leak.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -75,7 +86,81 @@ void main() {
...
+          expect(
+            await EmailAccountPasswordResetRequest.db.findFirstRow(
+              sessionBuilder.build(),
+              where: (final t) => t.id.equals(clientReceivedRequestId!),
+            ),
+            isNull,
+          );
+        });
+
+        test(
```
SandPod, line 143: Suggestion: If we move the random uuid creation into the module, this test should be moved to test that interface as well.

SandPod: And the duplicated test can be removed from there (for `startPasswordReset`).

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/auth_email.dart
```diff
@@ -85,40 +86,48 @@ abstract class AuthEmail {
...
+            authUserId: authUserId,
+            transaction: transaction,
+          );
+
+          return admin.createSession(
+            session,
+            authUserId,
+            transaction: transaction,
+          );
+        } on EmailAccountRequestException catch (e) {
```
SandPod: Question: Should there be a client version of this exception group so that we don't accidentally leak this information? Or is there a case where we would like to send `notFound` exceptions?

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_accounts.dart
marcelomendoncasoares: As we discussed, the strategy for keeping the non-nullable `UuidValue` return only on the endpoint business logic is due to this being coupled with our usage of the endpoint. Having it here could be strange for users that are using this on the server only, with no intention to send this information to the client - such generating only magic links to send to email.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4058/commits/f10bbc54a6c416551588f5a0b94ac4caadf1ac44.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/auth_email.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4058/commits/04e98314b9c6a4d368b681e1d50827506b041334. I focused this commit on decoupling the exceptions and added more test coverage (with fixes) on the next one.

marcelomendoncasoares:
While working on this PR, we realized that the `completePasswordReset` method must be split in two: one for first verifying the reset password code and another to set the new password. This approach is standard practice and will improve both usability and security, ensuring that users have completed verification before being allowed to set a new password. @coderabbitai, can you create an issue for it?

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -10,72 +11,160 @@ void main() {
...
+
+          expect(
+            clientReceivedRequestId!.version,
+            validPasswordResetRequest.version,
+          );
+        });
+      });
     },
+    testGroupTagsOverride: TestTags.concurrencyOneTestTags,
+    rollbackDatabase: RollbackDatabase.disabled,
```
SandPod, line 167: Question: What is the reason that this is needed?

Given these tests, I don't really see a reason for disabling the rollbacks?

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -510,6 +879,11 @@ Future<void> _cleanUpDatabase(final Session session) async {
     where: (final _) => Constant.bool(true),
   );
 
+  await EmailAccountRequest.db.deleteWhere(
```
SandPod, line 882: Comment: Good find 👍

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/auth_email.dart
```diff
@@ -192,14 +189,55 @@ abstract class AuthEmail {
...
+    return await fn();
+  } on EmailServerException catch (e) {
+    switch (e) {
+      // Login
+      case EmailAccountNotFoundException():
+      case EmailAuthenticationInvalidCredentialsException():
+        throw EmailAccountLoginException(
+            reason: EmailAccountLoginExceptionReason.invalidCredentials);
+      case EmailAuthenticationTooManyAttemptsException():
+        throw EmailAccountLoginException(
```
SandPod: Question: Is this also leaking the existence of the account?

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/models/exceptions/email_account_password_reset_exception_reason.spy.yaml
```diff
@@ -6,20 +6,11 @@ values:
...
   ### it has expired.
   - expired
-  ### Exception to be thrown when a password reset request could not be found.
-  ###
-  ### This might mean that the password reset request never existed or has been
-  ### removed in the meantime. This reason is only available in the server and
-  ### will be replaced by `unauthorized` when sent to the client.
-  - notFound
+  ### Exception to be thrown when an attempt was made to complete a password
+  ### reset with an invalid verification code.
```
SandPod: Suggestion:`...reset with an invalid verification code or password reset request id.

SandPod: Possibly the same modification for `email_account_request... exception`?

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_account_server_exceptions.dart
```diff
@@ -0,0 +1,68 @@
+/// Base exception for all email related errors.
+///
+/// These exceptions are for internal purposes only and must not be exposed to
+/// clients. For such, see the adequate serializable exceptions.
+sealed class EmailServerException implements Exception {}
```
SandPod, line 5: Suggestion: Create a sub sealed group that extends `EmailServerException` for the exceptions that could be thrown in different scenarios.

This makes it possible to catch all exceptions from a scenario without capturing other cases.

Also add documentation that these groups exist in the method documentation.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/auth_email.dart
marcelomendoncasoares: Not for login, because we log every login attempt, regardless of if it is for an existing account or not.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
marcelomendoncasoares: The `startPasswordReset` tests break with "Concurrent calls to transaction are not supported when database rollbacks are enabled." without it. This is in line with the other existing tests for password reset, that already were in a group with disabled concurrency.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_account_server_exceptions.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4058/commits/c4cf9eafed43e4859f60a8a6424bc933bb15da42.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/models/exceptions/email_account_password_reset_exception_reason.spy.yaml
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4058/commits/a672fefe8d46d622f3c2f01daba9c4626d7aafae,

SandPod (approved):
LGTM! 👍

PR #4059, opened 2025-10-14
feat: Introduces a new auth example project.

SandPod:
Introduces a new example project named `auth` that will be used to showcase the new authentication.

This project currently does not have any implemented authentication.

## Pre-launch Checklist

- [ ] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [ ] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
- [ ] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.
- [ ] I added new tests to check the change I am making.
- [ ] All existing and new tests are passing.
- [ ] Any breaking changes are documented below.

If you need help, consider asking for advice on the [discussion board](https://github.com/serverpod/serverpod/discussions).

## Breaking changes

None - Moves the old example into a `legacy_auth_directory`

PR #4062, opened 2025-10-15
chore: Rename legacy auth example back to auth_example.

SandPod:
This PR https://github.com/serverpod/serverpod/pull/4059 introduced a name change to the legacy auth example.
But we have external sources that uses directly links to main when reffering to that package (in our documentation).

Therefore, this PR reverts the name change of the project.

### DIFF
--- a/examples/auth/README.md
+++ b/examples/auth/README.md
@@ -2,6 +2,6 @@
 
 This project contains an example on how to use the new authentication module in Serverpod.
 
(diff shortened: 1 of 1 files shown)


marcelomendoncasoares (approved):
I understand our short-term goal is to avoid having to change other repos mid-development just to have this new example, but we still aim to declare the old example as "legacy" before release of the next version, right? So users do not get confused on how to use the new auth and end up copying from the old example - which is in higher visibility.

SandPod:
@marcelomendoncasoares Yes! That is the intention.

@coderabbitai - Could you create an issue out of this comment: https://github.com/serverpod/serverpod/pull/4062#pullrequestreview-3339931093

PR #4004, opened 2025-09-24
feat: Token issuer, manager and auth config

exaby73:
This PR adds `AuthConfig`, `TokenManager` and `TokenIssuer` with their respective implementations for JWT and SAS (Serverpod Auth Sessions).

## Pre-launch Checklist

- [x] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [x] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
- [x] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.
- [x] I added new tests to check the change I am making.
- [x] All existing and new tests are passing.
- [x] Any breaking changes are documented below.

If you need help, consider asking for advice on the [discussion board](https://github.com/serverpod/serverpod/discussions).

Issue #3453, opened 2025-04-22
Create Apple authentication module


tp:
## Open Points

- [ ] Support incoming web-hooks from Apple
  - missing documentation, the Relic handler is now exposed
- [ ] Handle "re-registration" (might change name and/or email on a later re-connection)
  - This is probably not critical out of the gate, and there is no 100% great path to handle this right now without overwriting the previous data
- [ ] ~~Consolidation by email address (?)~~
- [ ] Adding Apple login to an existing account (new endpoint method/flow)
  - On the endpoint level, instead of always creating a new account
  - This could also be a later feature, and the initial release would only support 1 authentication method per `AuthUser`

SandPod:
@tp Is this complete?

tp:
@SandPod the "Open Points" as stated above still hold, though not all of them are strictly required.

I removed the "email consolidation", as you already declared that out of scope.

Likewise, the re-registration (if the user disconnected the authentication via their Apple device and then re-registers with a potentially different name and email) is probably less relevant initially (and in practice anyway). Lastly, having an endpoint to add Apple as an additional authentication mechanism is still open, but I think that could likewise make for a follow-up ticket.

So the only thing that would be needed initially from my point of view would be documenting the Relic handler to handle server-to-server notifications from Apple.

Issue #3451, opened 2025-04-22
Create Google authentication module


tp:
Open points:

- [ ] Consider switching to "internal" token verification like we do in Sign in with Apple, which would allow for testing this offline and with an expired token (mocking the time via `clock`)
- [ ] Consider simplifying the configuration to not require the whole JSON configu file, but rather just specify the Google "client ID", which is the only field we currently use on that level (to verify that the incoming token was actually meant for our app).
- [ ] Handle configuration all client IDs centrally (see [comment](https://github.com/serverpod/serverpod/issues/3639#issuecomment-3024191056)). **This is a blocker if using multiple platforms (ios and android)**
- [ ] Handle refresh tokens, and occasionally background checks (like Apple)
  - [ ] Does Google also offer a PUSH web hook for this?
  - [ ] Check when the previous flow using `authenticationCode` (in `GoogleEndpoint.authenticateWithServerAuthCode`) was used (web & current implementation uses an identity token, but would be great to get them aligned to have 1 codepath and similar functionality)

tp:
Just ran a local integration test for this:

<img width="1474" height="1084" alt="Image" src="https://github.com/user-attachments/assets/4f77777e-80cd-4f28-a219-66ed3eb9b122" />

<img width="2278" height="394" alt="Image" src="https://github.com/user-attachments/assets/b16cee5a-035a-49c6-9054-1237c7452e0e" />

It seems like the format of the identify token changed and not the email verification status is given as a String (`"true"`/`"false"`) under key `email_verified`. (At least for the one created on web, maybe app tokens differ.) (This matches the spec published at https://accounts.google.com/.well-known/openid-configuration)

Sadlym I can not commit this test, as it calls to Google's API for reading the token, which will fail as soon as the token is expired (and we should not have external dependencies in the test anyway).

So maybe a later implementation can fetch the keys from Google and verify the token locally with the keys published at https://www.googleapis.com/oauth2/v3/certs, just like the Apple login works (and is tested).

PR #4064, opened 2025-10-15
refactor!: Refactor Google and Apple to follow the new IDP pattern.

SandPod:
## IDP Architecture Pattern

To improve the developer experience when working with identity providers (IDPs), this PR introduces a consistent, modular structure for all providers.  
The goal is to make their internal logic easier to navigate, extend, and customize.

### Proposed Structure

Each identity provider should be organized into three files:

1. **`<provider>_idp.dart`** — Main class for the identity provider.  
   Contains methods intended to be called from endpoints or routes.

2. **`<provider>_idp_admin.dart`** — Admin-focused utilities.  
   Provides administrative functions used to manage and configure the identity provider.

3. **`<provider>_idp_utils.dart`** — Shared building blocks.  
   Defines lower-level utilities used by both the main and admin classes.  
   These “atoms” can also be reused by developers to create custom authentication flows.

This separation clarifies which layer should be modified when adjusting authentication behavior and makes it easier to explore or extend existing providers.

---

## Summary of Changes

- Introduces a temporary `AuthServices` abstraction to decouple identity providers from singleton instances.  
- Refactors the **Google Identity Provider** to follow the new IDP architecture pattern.  
- Refactors the **Apple Identity Provider** to follow the new IDP architecture pattern.  
- Lays the foundation for upcoming work in #3451 and #3453.

---

## About `AuthServices`

The new `AuthServices` class is a **temporary solution** that will later be replaced by the more complete `AuthConfig` implementation currently being developed in [#4004](#4004).

---

## Guidance for review

Focus on the interface and the structure.

Best reviewed commit by commit. But the commit that conforms to the new interface does modify the behavior of the methods as well.

There will be a follow-up PR to ensure that the internal functionality works as expected.

---

## Pre-launch Checklist

## Breaking changes

Yes, this PR contains breaking changes to how Apple and Google IDP work, both in the endpoint layer towards the client and in the business API layer.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/apple/business/apple_idp.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/apple/business/apple_idp.dart
@@ -42,13 +42,9 @@ final class AppleIDP {
       key: config.key,
     );
 
-    final siwa = SignInWithApple(
-      config: siwaConf,
-    );
+    final siwa = SignInWithApple(config: siwaConf);
 
-    utils = AppleIDPUtils(
-      siwa: siwa,
-    );
+    utils = AppleIDPUtils(siwa: siwa);
     admin = AppleIDPAdmin(utils: utils);
   }
 


modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/apple/business/apple_idp.dart
```diff
@@ -0,0 +1,127 @@
...
+      serviceIdentifier: config.serviceIdentifier,
+      bundleIdentifier: config.bundleIdentifier,
+      redirectUri: config.redirectUri,
+      teamId: config.teamId,
+      keyId: config.keyId,
+      key: config.key,
+    );
+
+    final siwa = SignInWithApple(
+      config: siwaConf,
```
marcelomendoncasoares: Suggestion(NIT): Unnecessary trailing comma - here and on line 50.

examples/auth/auth_server/lib/server.dart
```diff
@@ -13,14 +13,33 @@ import 'src/generated/endpoints.dart';
...
+      ),
+    ),
+    appleIDPConfig: AppleIDPConfig(
+      serviceIdentifier: pod.getPassword('appleServiceIdentifier')!,
+      bundleIdentifier: pod.getPassword('appleBundleIdentifier')!,
+      redirectUri: pod.getPassword('appleRedirectUri')!,
+      teamId: pod.getPassword('appleTeamId')!,
+      keyId: pod.getPassword('appleKeyId')!,
+      key: pod.getPassword('appleKey')!,
+    ),
```
marcelomendoncasoares, line 36: Discussion: This feels a bit asymmetric for me: Google configuration as a single `GoogleClientSecret` object - that are extracted from a configuration file - as opposed to Apple having the configuration separate. Does it make sense to also bring the keys on Google side to the top-level of the config object?

SandPod: I think this will depend a bit on how the secret is supplied. I don't know the specific implementations right now but will ensure that we harmonize this.

For example, only read a single line inside of the `googleClientSecret`, this line should probably be supplied in isolation. We also need to add support for configuring mutiple client secrets.

So if it is with you, I'll postpone this improvement untill I have more knowledge about the provider.

Issue #3283, opened 2025-03-04
Rework exception hierarchy in Serverpod.


SandPod:
Placeholder issue where details will need to be added later.

For Serverpod 3.0 we should rework our exception hierarchy in Serverpod to ensure consistency throughout the framework.

One example of a strange exception hierarchy are the specializations for `DatabaseInsertRowException`, `DatabaseUpdateRowException` and `DatabaseDeleteRowException`. These don't really explain what went wrong but more reflect the domain where it went wrong.

### Client Exceptions

This rework should also inspect the client exception hierarchy. We should probably have a base class and then all other exceptions on the client should be derived from that.

marcelomendoncasoares:
When refactoring the client-side exception hierarchy, it is necessary to create a dedicated `ServerpodClientNetworkError` exception to replace the current usage of `ServerpodClientException` with `statusCode: -1` to replace `http.ClientException` on web and `SocketException` on native. This pattern already existed for web and was replicated to native on PR #4072.

EchoEllet:
> ### Client Exceptions


### JSON parsing

I think JSON parsing error handling can be more explicit, with two exceptions:

- `JsonDecodingException`: Malformed JSON sent by the server. Sometimes Nginx, Cloudflare, or a reverse proxy may return HTML in case the server is down. There are other cases of why this may happen. Can be handled by catching `FormatException`

- `JsonDeserializationException`: Even if the response body is valid JSON, models are generated on both server and client side (type-safe, automated), it is still not guaranteed that the JSON structure is compatible with the data class model. Especially if server deployment can be self-hosted (like WordPress).
  - The app user may use an outdated client
  - The server admin may use an outdated Server that sends missing required fields to the client.
  - Can be handled by `on TypeError catch (e)` (not `on Exception catch (e)` if we use `!` in `fromJson` methods). ([Example](https://github.com/KraftLauncher/kraft-launcher/blob/d5894dba2ba75cb3488a420fcc81511884698dae/packages/core/json_utils/lib/json_utils.dart#L76-L78))

### Socket / Network failures

[...]

PR #4072, opened 2025-10-17
fix: Fix `ClientAuthSessionManager` not catching network errors on init

marcelomendoncasoares:
Fix network errors not being caught on `ClientAuthSessionManager` initialization.

While making this fix, it was also identified that there were an asymmetry between web and platform clients: web was replacing network exception (`http.ClientException`) by `ServerpodClientException`, while platform wasn't. This PR also fixes this asymmetry by making platform also replace the exception. Although the solution of using `ServerpodClientException` with code -1 is not ideal, it is best to have both exceptions paired and further improve exception handling when working on issue #3283.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
@@ -126,7 +126,12 @@ class ClientAuthSessionManager implements RefresherClientAuthKeyProvider {
     Duration timeout = const Duration(seconds: 2),
   }) async {
     await restore();
-    return validateAuthentication(timeout: timeout);
+    try {
+      await validateAuthentication(timeout: timeout);
+      return true;
+    } on ServerpodClientException catch (_) {
+      return false;
+    }
   }
 
   /// Restore the current sign in status from the storage. If the underlying
@@ -150,18 +155,14 @@ class ClientAuthSessionManager implements RefresherClientAuthKeyProvider {
 
   /// Verifies the current sign in status of the user with the server and
   /// updates the authentication info, if needed. If the user authentication is
-  /// no longer valid, the user is signed out from the current device. In case
-  /// of network or internal server errors, the user is not signed out. The
+  /// no longer valid, the user is signed out from the current device. If the
+  /// sign out fails for any reason, returns false. Otherwise, returns true.
+  /// Other exceptions during the validation are propagated to the caller. The
   /// [timeout] parameter can be used to override the default client timeout.
   Future<bool> validateAuthentication({Duration? timeout}) async {
-    try {
-      return await _validateAuthentication().timeout(
-        timeout ?? caller.client.connectionTimeout,
-        onTimeout: () => false,
-      );
-    } on ServerpodClientException catch (_) {
-      return false;
-    }
+    return await _validateAuthentication().timeout(
+      timeout ?? caller.client.connectionTimeout,
+    );
   }
 
   Future<bool> _validateAuthentication() async {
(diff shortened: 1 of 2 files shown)


modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
```diff
@@ -1,4 +1,5 @@
 import 'dart:collection';
+import 'dart:io';
```
SandPod: Question: This makes this package incompatible with Flutter Web. Shouldn't this be caught and converted in the serverpod `browser/io` packages?

marcelomendoncasoares: Removed at https://github.com/serverpod/serverpod/pull/4072/commits/57dfe365520ca4eb03d0113dc484cc75882ef3bd after replacing the `SocketException` on the client by the same pattern of the browser implementation:

https://github.com/serverpod/serverpod/blob/dac42c39f52a4ad41bb95e4c3eb9d5fe73961515/packages/serverpod_client/lib/src/serverpod_client_browser.dart#L57-L60

The replacement was done at https://github.com/serverpod/serverpod/pull/4072/commits/961ef60bce46cc96a0d357b7f17b0676d4d6a5e7.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
```diff
@@ -111,10 +111,13 @@ class ClientAuthSessionManager implements RefresherClientAuthKeyProvider {
   /// authentication is no longer valid, the user is signed out from the current
   /// device, updating the [authInfo] value. If the refresh fails due to other
   /// reasons (network error, server error, etc.), returns false, but does not
-  /// sign out the user. Returns true if the authentication was validated.
-  Future<bool> initialize() async {
+  /// sign out the user. Returns true if the authentication was validated. Use
+  /// [timeout] to set a maximum time for the server validation call.
+  Future<bool> initialize({
```
SandPod, line 125: Suggestion: Perhaps we should mention that `restore()` can be used to do the same operation without roundtripping to the server in the doc comment?

SandPod: Or that this is the same thing as calling `restore()` and then `validateAuthentication()`

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
```diff
@@ -138,21 +141,29 @@ class ClientAuthSessionManager implements RefresherClientAuthKeyProvider {
...
-            !await caller.status.isSignedIn()) {
-          return await signOutDevice();
-        }
-      }
-      return true;
+      return await _validateAuthentication().timeout(
+        timeout ?? caller.client.connectionTimeout,
+        onTimeout: () => false,
+      );
     } on ServerpodClientException catch (_) {
```
SandPod: Question: Should there perhaps be a way to call this method where the exceptions are not squashed? So that the client can react to it?

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4072/commits/04d4db036c15b0ec8e68984f1cfff0555b5e8c36.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/session_manager.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4072/commits/7e8b534c6d217fe8dc3cd5b6aab486fb5fd33be8.

PR #4077, opened 2025-10-23
chore: Add valid tests for VS Code extension

Swiftaxe:
## Changes Made

- Tests added to cover the functionality in the VS Code extension

Bug Fixed
- When writing the tests, a possible bug was discovered. Deactivate() did not clear the client variable. It stayed set forever, leaving a stale reference. 😱 It caused a leak between tests.
  It is now fixed. 
  (Let me know if this breaks the single-issue per PR rule and should be filed as a separate issue.)

For the compilation to run, there was a need for a couple of package changes:
 - Upgraded typescript as @types/node required a later version. 
 - As semver is a plain JavaScript package, TypeScript type definitions were needed

Closes: #3998

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tools/serverpod_vscode_extension/src/test/extension.test.ts
+++ b/tools/serverpod_vscode_extension/src/test/extension.test.ts
@@ -1,15 +1,149 @@
 import * as assert from 'assert';
-
-// You can import and use all API from the 'vscode' module
-// as well as import your extension to test it
 import * as vscode from 'vscode';
-// import * as myExtension from '../../extension';
+import * as sinon from 'sinon';
+import * as childProcess from 'child_process';
+import { validVersion, activate, deactivate } from '../extension';
+import { LanguageClient } from 'vscode-languageclient/node';
 
-suite('Extension Test Suite', () => {
-    vscode.window.showInformationMessage('Start all tests.');
+suite('VS Code Extension', () => {
+    let execSyncStub: sinon.SinonStub;
+    let showErrorMessageStub: sinon.SinonStub;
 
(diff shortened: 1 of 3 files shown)


Swiftaxe:

Swiftaxe:
recheck

tools/serverpod_vscode_extension/src/test/extension.test.ts
```diff
@@ -1,15 +1,158 @@
...
+
+            // Temporarily replace LanguageClient
+            const LanguageClientModule = require('vscode-languageclient/node');
+            const originalLC = LanguageClientModule.LanguageClient;
+            LanguageClientModule.LanguageClient = LanguageClientSpy;
+
+            activate(mockContext);
+
+            // Restore
+            LanguageClientModule.LanguageClient = originalLC;
```
SandPod: Question: JS tests suites are not my strong suit, but if something goes wrong after the language client has been replaced that aborts the test (say `activate` crashes), this teardown won't be executed.

Is it possible to move this to a teardown section instead? To ensure we always capture and restore the original language client?

Swiftaxe: Good point! Will do.

Swiftaxe: Fixed! The test now uses a stub instead of altering the actual language client. The stub is restored in the teardown by sinon.restore(). This also makes the test cleaner.

SandPod (approved):
LGTM! 👍

Issue #2224, opened 2024-05-09
StreamingConnectionHandler timer not being cancelled after successful reconnect


inf0rmatix:
**Describe the bug**
The `StreamingConnectionHandler` class creates a timer to schedule retrys for reconnecting to the server when the connection is lost. Once the connection is regained, the timer is not cancelled and keeps running indefinitely until another disconnect happens or the `StreamingConnectionHandler` is disposed.

**To Reproduce**
1. Create a new instance of `StreamingConnectionHandler`
2. Call `StreamingConnectionHandler`.connect()
3. Restart your server to enforce a streaming connection disconnect
4. Observe the function `_onCountDown` in `StreamingConnectionHandler` being called indefinitely even after the connection has been established again

**Expected behavior**
The timer should be cancelled once the `StreamingConnectionStatus.connected` is emitted

**Screenshots**
 - 

**Serverpod versions**
 -  _CLI Version:_ 1.2.6
 - _Library version:_ 1.2.6

**Platform information**
All

**Additional context**
 -

PR #4086, opened 2025-10-24
fix: Fix streaming connection handler.

SandPod:
Fixes an old issue on the endpoint streaming interface with a simple test for validation.

The test unfortunately needs to be timing based as an internal `Timer` is used and the interface currently only accepts seconds and not duration.

Closes: #2224 

## Pre-launch Checklist

## Breaking changes

None - simple bug fix.


SandPod:
Unrelated test failures that will be fixed in a separate PR.

packages/serverpod_client/test/streaming_connection_handler_test.dart
```diff
@@ -0,0 +1,118 @@
...
+    client = MockServerpodClient();
+    receivedStates = [];
+  });
+
+  tearDown(() {
+    handler.dispose();
+  });
+
+  test(
+      'when transitioning to connected status during countdown '
```
marcelomendoncasoares, line 75: This test description is missing a `Given`.

marcelomendoncasoares (approved):

packages/serverpod_client/test/streaming_connection_handler_test.dart
SandPod: Fixed in [8a16eb6](https://github.com/serverpod/serverpod/pull/4086/commits/8a16eb65b82dc4a768c922c5121e06f16c0d9251)

PR #4087, opened 2025-10-24
test: Update expired certs

nielsenko:
Update expired test certs


SandPod:
Seems like something is going wrong here, why is this PR different from the one in Relic?

SandPod (approved):

PR #4088, opened 2025-10-26
feat: Set column name explicitly

### LINKED ISSUE #1238: Allow to set the column name explicitly.
To help us manage implicit relations we will need to generate columns with relations in the database. But if the programmer then specifies this relation explicitly we would have to change the column name holding the relation. To alleviate this problem we should be able to set a different column name for each field in the model. 

While we can completely manage this behind the scenes exposing this functionality may also help programmers to manage migrations and changes within the model / database.

Introduce a new keyword `column` on the `relation` keyword. The value should be a string of the column name that will be used in the database. The default value of `column` would be the field name and this keyword should normally never have to be used in most cases, but can be useful for more advanced situations. 

Example:

```yaml
class: Example
table: example
fields:
  parentId: int, relation(parent=example), column=parent_id
```

jwelmac:
This pull request introduces a new keyword `column` for each field. 
The value should be a string of the column name that will be used in the database. 
By default the column name is null and falls back to the field name.
This is useful in postgres databases to ensure that the [naming convention](https://www.geeksforgeeks.org/postgresql-naming-conventions/) is preserved, or for more advanced situations.

Closes #1238 

## Pre-launch Checklist

## Breaking changes

None

## Summary by CodeRabbit

* **New Features**
  * Support explicit database column names on model fields, applied to serialization, schema generation, and foreign keys.
  * Added validation to ensure a column name is not assigned to multiple fields and a YAML key for column names.

* **Documentation**
  * Added examples and docs showing how to specify explicit column names in model definitions.

* **Tests**
  * Added tests covering validation, database schema output, and generated client/server serialization for explicit column names.


jwelmac:

codescype:
Fantastic!

Welldone 👏🏾 @jwelmac

jwelmac:
> Fantastic!
> 
> Welldone 👏🏾 @jwelmac

Thanks @codescype

SandPod:
@jwelmac Thank you for the contribution 🚀 

We will first conduct an automated review by CodeRabbit and then we will take a look at the PR.
If there are comments that you do not agree with from CodeRabbit, please answer the comments and we can take a look at that in our manual review.

SandPod:
@coderabbitai review

tools/serverpod_cli/test/generator/dart/server_code_generator/explicit_column_name_test.dart
```diff
@@ -0,0 +1,189 @@
...
+        expect(
+          toJsonForProtocolCode.contains(
+            "'$columnName' : $columnFieldName",
+          ),
+          isTrue,
+          reason:
+              'The toJsonForProtocol method should map the field name to its '
+              'serialized variable.',
+        );
+      });
```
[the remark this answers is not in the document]
jwelmac: Good catch. Same fixed!

jwelmac:
@SandPod 
Updated code for all @coderabbitai review changes and fixes.

Looking forward to the manual review.

Once this is complete could I make the documentation updates, or would you guys handle that?

jwelmac:
@coderabbitai review nitpick comments addressed

SandPod:
@jwelmac Thanks for addressing the comments.

I'll take a look at the PR now.

> Once this is complete, could I make the documentation updates, or would you guys handle that?

It would be great if you could submit a documentation update as well.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -971,6 +971,34 @@ class Restrictions {
     return errors;
   }
 
+  List<SourceSpanSeverityException> validateColumnName(
```
SandPod, line 1007: Suggestion: Add validation for `_maxColumnNameLength`.

If the column name is longer than max supported column name length, postgres will not allow the column to be created.

Add a test that validates this works as expected.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -2055,4 +2083,13 @@ class Restrictions {
       },
     );
   }
+
+  List<SerializableModelFieldDefinition> _findFieldsWithColumn(
+    ModelClassDefinition currentModel,
+    String column,
+  ) {
+    return currentModel.fields
```
SandPod, line 2205: Suggestion: Use the `columnName` getter to find collisions.

This will not find collisions with fields that do not have a `column` override.

Please add a test that validates that we can detect collisions with columns that do not have overrides.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -265,6 +265,23 @@ class SerializableModelFieldDefinition {
...
+  /// This should be set if the column name is not a camelCaseString
+  /// and will therefore be different from the field name.
+  ///
+  /// Example yaml use
+  /// ```yaml
+  /// class: Example
+  /// fields:
+  ///   userName: String, column=user_name
+  /// ```
+  final String? column;
```
SandPod: Suggestion: Rename to `columnNameOverride` since that is what it is.

Can we make this private? Since the `columnName` getter is the one that is intended to be used right?

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -265,6 +265,23 @@ class SerializableModelFieldDefinition {
...
+  /// Name of the column in the database
+  /// This should be set if the column name is not a camelCaseString
+  /// and will therefore be different from the field name.
+  ///
+  /// Example yaml use
+  /// ```yaml
+  /// class: Example
+  /// fields:
+  ///   userName: String, column=user_name
+  /// ```
```
SandPod: Suggestion: Remove this documentation.

It is also not only used for removing camel case. It can be used in any scenario.

The internal model is not tightly coupled with the yaml structure. This should only explain what this field is intended to be used for. (And as suggested below, this can probably be private to the class)

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
SandPod: Reflect this change in the serializable entity field definition builder and tests.

tools/serverpod_cli/lib/src/analyzer/models/yaml_definitions/class_yaml_definition.dart
```diff
@@ -203,6 +203,10 @@ class ClassYamlDefinition {
                   Keyword.relation,
                 },
               ),
+              ValidateNode(
```
SandPod, line 202: Suggestion: This should only be allowed if the class has the `table` property and should probably not be allowed in the `id` field, if you don't want to implement dynamic primary key fields, which I believe is a bigger undertaking.

tools/serverpod_cli/test/database/create_definition/database_action_definition_test.dart
```diff
@@ -130,4 +130,82 @@ void main() {
       });
     });
   });
+
+  group('Given a class definition with a table with an explicit column name ',
```
SandPod: Question: What are we testing here?

tools/serverpod_cli/test/generator/dart/client_code_generator/explicit_column_name_test.dart
```diff
@@ -0,0 +1,161 @@
...
+  final expectedFilePath = path.join(
+    '..',
+    'example_project_client',
+    'lib',
+    'src',
+    'protocol',
+    '$testClassFileName.dart',
+  );
+
+  group(
```
SandPod, line 28: Discussion: This change should not propagate to the client as this is pure server changes.

The toJson on the client should probably match the toJsonForProtocol on the server.

tools/serverpod_cli/test/generator/dart/server_code_generator/explicit_column_name_test.dart
```diff
@@ -0,0 +1,189 @@
...
+            "'$noColumnFieldName' : $noColumnFieldName",
+          ),
+          isTrue,
+          reason: 'The toJson method should map the field name to its '
+              'serialized variable.',
+        );
+      });
+    });
+
+    group('then toJsonForProtocol method should return ', () {
```
SandPod: Suggestion/Discussion: toJsonForProtocol should not be influenced by the database column name.

Modifying the column name would then break communication with old clients which unrelated configuration.
This forces us to add a `fromJsonFromProtocol` so that we can send and receive the format from the client while modifying the names in the database.

SandPod: Perhaps we should break out the `json` methods for the database instead, `toJsonForDatabase` and ´fromJasonForDatabase`.

tools/serverpod_cli/lib/src/analyzer/models/yaml_definitions/class_yaml_definition.dart
jwelmac: I will add the restrictions stated.
I agree that dynamic primary key fields should ideally be handled as a separate pull request.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
jwelmac: Excellent suggestions.

tools/serverpod_cli/test/database/create_definition/database_action_definition_test.dart
jwelmac: Here we are testing that the database definition generated when the field has a column name override has the correct overrides as set in [tools/serverpod_cli/lib/src/database/create_definition.dart](https://github.com/serverpod/serverpod/pull/4088/files/6ac8e16e973ce5ef34f9e519b6bc814a1937ee92#diff-ab498fe61d260a8374867a564ce42d43f47b3c631b2d9a233a048019bba46d1b)

tools/serverpod_cli/test/generator/dart/server_code_generator/explicit_column_name_test.dart
jwelmac: > Suggestion/Discussion: toJsonForProtocol should not be influenced by the database column name.
> 
> Modifying the column name would then break communication with old clients which unrelated configuration. This forces us to add a `fromJsonFromProtocol` so that we can send and receive the format from the client while modifying the names in the database.

This is a very good point. Since the column name is database specific the protocol payloads could match the model fields. This allows changes to the database to be isolated without downstream effects when changes occur. With this we would have a clear separation of concerns.

jwelmac: > Perhaps we should break out the `json` methods for the database instead, `toJsonForDatabase` and ´fromJasonForDatabase`.

Interesting thought, but would this still be needed with a clear separation between protocol and database payloads as suggested previously.

jwelmac:
@SandPod 
Thanks for your detailed feedback.
This is one of the reasons I love PR Reviews and collaborative coding, it allows you to see your blind spots, learn and grow quickly. This is especially important when you have not yet developed a deep contextual awareness of a project.

I will address the various scenario you have mentioned.
Looking forward to getting this completed satisfactorily so other devs can benefit.

tools/serverpod_cli/test/generator/dart/server_code_generator/explicit_column_name_test.dart
jwelmac: As I dig into this deeper I am definitely appreciating more the concerns you have raised.
Isolating the serialization and deserialization of the models between client and database requires quite a few changes to ensure that it happens seamlessly while maintaining communication with old clients.
After much consideration I am leaning towards the `to/FromJsonForDatatabase` too.
I will keep you posted.

jwelmac: @SandPod

Here is my current recommendation on the implementation of the database related json methods.

## Serializing json for database
  
1. Add a new interface class `TableRowWithColumnOverride`

```dart
/// Holds data correspoding to a row in the database where one or more column
/// names have been overriden in their model representation.
abstract interface class TableRowWithColumnOverride<T_ID>
    implements TableRow<T_ID> {
  /// Returns a serialized JSON structure of the model which
  /// ensures keys match the column names used by the database.
  dynamic toJsonForDatabase();
}
```

2. Add an extension method to the `TableRow` interface to ensure that `toJsonForDatabase` can
be called for all instances.
```dart
/// Extension on a [TableRow] to enable database serialization
extension TableRowDatabaseJsonExtension on TableRow {
  /// Returns the json representation to be sent to the database for storage
  dynamic toJsonForDatabase() {
    // Save the current instance to a variable to allow type promotion
    // as type-promotion for `this` is not yet supported.
    // See: <https://github.com/dart-lang/language/issues/1397>
    // Adapted from: https://stackoverflow.com/a/71327677/5372784
    final self = this;
    if (self is TableRowWithColumnOverride) {
      return self.toJsonForDatabase();
    }
    return toJson();
  }
}
```
  
3. Generate the `toJsonForDatabase` method on each `TableRowWithColumnOverride<T>` implementation.

```dart
  @override
  Map<String, dynamic> toJsonForDatabase() {
    return {
      if (id != null) 'id': id,
      if (bigIntDefaultPersistStr != null)
        'big_int_default_persist_str': bigIntDefaultPersistStr?.toJson(),
    };
  }
```

4. Replace the `toJson` method calls to serialize database rows with the `toJsonForDatabase` method.
```dart
class DatabaseConnection {
  ...
  
  String _createQueryValueList(
    Iterable<TableRow> rows,
    Iterable<Column> column,
  ) {
    ---     return rows.map((row) => row.toJson() as Map<String, dynamic>).map((row) {
    +++     return rows
    +++         .map((row) => row.toJsonForDatabase() as Map<String, dynamic>)
    +++         .map((row) {
                 ...
                 }).join(', ');
              }).join(', ');
    }

  ... 

  List<Map<String, dynamic>> Function(Iterable<Map<String, dynamic>>)
      _mergeResultsWithNonPersistedFields<T extends TableRow>(
    List<T> rows,
  ) {
    return (Iterable<Map<String, dynamic>> dbResults) =>
        List<Map<String, dynamic>>.generate(dbResults.length, (i) {
          return {
            // Add non-persisted fields from the original object
---         ...rows[i].toJson(),
+++         ...rows[i].toJsonForDatabase(),
            // Override with database fields (common fields)
            ...dbResults.elementAt(i),
          };
        });
  }
}
```

```dart
class InsertQueryBuilder {
  ...

  String? _build(bool onlyWithIdNull) {
    ...
    var values = filteredValues
    --- .map((row) => row.toJson() as Map<String, dynamic>)
    +++ .map((row) => row.toJsonForDatabase() as Map<String, dynamic>)
    ...

  }
}
```

## Parsing json from database

1. Generate a `fromJsonFromDatabase` factory constructor for  each `TableRowWithColumnOverride<T>` class implementation.

```dart
  factory BigIntDefaultPersist.fromJsonFromDatabase(
      Map<String, dynamic> jsonSerialization) {
    return BigIntDefaultPersist(
      id: jsonSerialization['id'] as int?,
      bigIntDefaultPersistStr:
          jsonSerialization['big_int_default_persist_str'] == null
              ? null
              : _i1.BigIntJsonExtension.fromJson(
                  jsonSerialization['big_int_default_persist_str']),
    );
  }
```

2. Add a `deserializeFromDatabase` method to the `SerializationManagerServer` class that by default forwards the call to the `deserialize` method

```dart
/// Deserialize the provided json [data] from the database to an object
/// of type [t] or [T].
T deserializeFromDatabase<T>(
  dynamic data, [
  Type? t,
]) {
  return deserialize(data, t);
}
```

3. Update the generation of the server `protocol.dart` file to generate a `deserializeFromDatabase` if one or more 
classes have column overrides. This method would fall back to using deserialize for any other classes.

```dart
  
  @override
  T deserializeFromDatabase<T>(
    dynamic data, [
    Type? t,
  ]) {
    if (t == _i25.BigIntDefaultPersist) {
      return _i25.BigIntDefaultPersist.fromJsonFromDatabase(data) as T;
    }
    return deserialize<T>(data, t);
  }
```

4. In the `DatabasePoolManger` update the `SerializationManager` to `SerializationManagerServer`

```dart
---   late SerializationManager _serializationManager;
+++   late SerializationManagerServer _serializationManager;

---   SerializationManager get serializationManager => _serializationManager;
+++   SerializationManagerServer get serializationManager => _serializationManager;
```

5. Update the calls for `deserialize` to `deserializeFromDatabase` in `database_connection.dart`

```dart
    --- _poolManager.serializationManager.deserialize<T>
    +++ _poolManager.serializationManager.deserializeFromDatabase<T>
```

## Questions
- Are there any potential gotchas, or downstream effects that you foresee that I need to account for upfront?
- Are there any alternatives you would like me to consider?

jwelmac: @SandPod 
I was awaiting a response before proceeding but its possibly you may be swamped with other responsibilities. So in the meantime I am going to proceed with the changes above and implementing the validations you previously mentioned.
If any changes or pivots are needed I will make the necessary adjustments.

SandPod: Hey @jwelmac! Sorry for the delayed response. We are closing in on the first RC for Serverpod 3. Give me a couple of days and I'll come back with some feedback on your approach.

Sorry for any inconvenience and thanks for staying dedicated to the contribution 🙏

jwelmac: You are most welcome.
Looking forward to it.

SandPod: @jwelmac - The first RC is now out freeing me up for a bit, thanks for waiting for my reply.

I think the solution you are suggesting above _could_  work. While tinker on this I had another idea that I think is worth exploring before adding a different serialization approach.

When serializing data into the database, we store the column name as part of the `Table` definition. We could also, only in the database layer, translate our "fields" of the model into the `column` specified.

I believe this would require the following changes:

1. `Column` receives a new field named `fieldName` that holds the name of the column in the model it is connected to.
2. When inserting data to the database, for example in the `InserQueryBuilder.build(...)` method, we do a lookup in the json using the `Column.fieldName` and assign it to a `Column.columnName` key.
3. When deserializing results from the database in `DatabaseConnection._deserializedMappedQuery` (and possible other places) we add a step that translates the `Column.columnName` back into the `Column.fieldName` before deserialization.

In the database layer, we already have aliases happening to protect ourselves from long table or column names.

This would isolate the change to only the database layer and save us from adding another toJson method.

jwelmac: @SandPod 
Congratulations to you and the team on the recent release. 🥳
I know those releases are never easy on everyone.

Thank you for taking the time to review my approach.
I think that is a great approach which minimizes the impact to current code and generated files, while producing the intended functionality. 

I deliberately chose to ask for your insight because I suspected there may be another approach that is not readily apparent to my adjusting eyes, that with your awareness you would be able to better identify. 

I will get those changes done so you have something to look at early next week. Then we can possibly get this in the next RC.

jwelmac: @SandPod 
I was able to make the changes to maintain the serialization between server and client whilst ensuring that the database serialization and deserialization works. 
I used a combination of my previous recommendation which was made simpler by your suggestion.
Thanks again.

I will complete the tests on the `serverpod_test_server` next to ensure all the pieces work in harmony.

tools/serverpod_cli/test/generator/dart/client_code_generator/explicit_column_name_test.dart
jwelmac: Same updated.

tools/serverpod_cli/test/generator/dart/server_code_generator/explicit_column_name_test.dart
SandPod: Make sure that you add tests that validate all CRUD operations to the database when a column name has been modified. This will help us catch any inconsistencies in our implementation between the operations.

jwelmac: Will do!

jwelmac:
@SandPod 
I have added the following `test_server` models

- [x] [Model with a column override](https://github.com/Khuwn-Soulutions/serverpod/commit/899a1624a632f570d22dc1fd8b88c2b101563a29)
- [x] [Migration that contains a model with column name overrides.](https://github.com/Khuwn-Soulutions/serverpod/commit/12655e31e0cd778dd97307feee844eeeb5fe2333)
- [x] [Model with a relation that uses column override ](https://github.com/Khuwn-Soulutions/serverpod/commit/1ab8920748344ae3b266caba637e037ccc2b168c)
- [x] [Migration that contains a model with column name overrides on index](https://github.com/Khuwn-Soulutions/serverpod/commit/11b2e0b2460c1a48be30545554c058e8d5b590bb)
- [x] Model that has modified the column name (showing client should is unchanged and migration changes). [before](https://github.com/Khuwn-Soulutions/serverpod/commit/10dd96fbe542a9a2762e45ba2c2d4c20d177c1e5) [after](https://github.com/Khuwn-Soulutions/serverpod/commit/de64de4201a60d314ee04d612227ffc19f00e65f)

Of note to ensure reliability with relations I have added a [restriction](https://github.com/Khuwn-Soulutions/serverpod/commit/f6c239065f4f66ebb54b453cf1aa6c07459649fa) so that in a relation column overrides can only be used on foreign key fields. 
If user wants to ensure that the foreign key is different from the field name they need to specify the foreign key field and column override.
If a column override is placed on a symbolic relation field an error will be thrown.

I have added relevant tests so far except for database operations including explicit column names on the test server

jwelmac:
@SandPod 

Database operations for tables with explicit column names are now added.

- [x] [Create, read, update and delete of table with explicit column names](https://github.com/Khuwn-Soulutions/serverpod/commit/e3817a133341d6417d4a41be835493aa37f3e694)
- [x] [Query on table with a relation list with an explicit column name on foreign key field](https://github.com/Khuwn-Soulutions/serverpod/commit/3fa28a57f765765a1bce3411a41f2aa335ece054)

@coderabbitai Please review

SandPod:
@jwelmac - Hey! I just want to give you an update of the review.

I made several attempts to review this PR the last couple of working days but have not been able to find the time to complete it yet. The work with Serverpod 3 has been stealing most of my attention.

This is top of my stack of tasks to complete but since it is such a critical change in the core of the database connection I want to ensure that I give it the attention it deserves. Especially in regards to these extra layers when it comes to serialization performance.

I'm aiming to complete a review of this PR over the next couple of days.

jwelmac:
@SandPod 
Thanks for the follow up. 
I am looking forward to the review.

In the meantime, I have been doing some work on my package [supabase_codegen](https://pub.dev/supabase_codegen) for which I created a companion plugin to generate model files from supabase tables and enums for serverpod ([supabase_codegen_serverpod](https://github.com/Khuwn-Soulutions/supabase_codegen/tree/main/packages/supabase_codegen_serverpod)). 
I think it will be a great companion release alongside Serverpod 3. While I wait for this PR to get merged and released so I can release the first version of that package I am working on some other great features.

tests/serverpod_test_server/test_integration/database_operations/explicit_column_name/crud_test.dart
```diff
@@ -0,0 +1,108 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+import 'package:serverpod_test_server/test_util/test_serverpod.dart';
+import 'package:test/test.dart';
+import '../../../../../packages/serverpod/lib/src/database/sql_query_builder.dart';
+
+void main() async {
```
SandPod: Suggestion: Run these tests using the `withServerpod` test framework.

I belive we can use the test framework for these tests, since all database operations are actually performed in the test framework as well.
We just never commit them to the database and utalize the WAL.

Moving to the test framework will allow us to run these tests concurrently with other tests.

tests/serverpod_test_server/test_integration/database_operations/explicit_column_name/crud_test.dart
```diff
@@ -0,0 +1,108 @@
...
+      test('the explicit column name is used in the sql', () {
+        final query = InsertQueryBuilder(table: table, rows: [data]).build();
+        expect(
+          query,
+          'INSERT INTO "${table.tableName}" '
+          '("user_name", "user_description") '
+          'VALUES (\'userName\', \'description\') RETURNING *',
+        );
+      });
+    });
```
SandPod: Suggestion: this type of test, that captures the written sql, should be captured in packages/serverpod/test/database/database_query/insert_sql_query_test.dart.

This allows us to focus on the syntax in a single place and only need to update it there if it changes.

SandPod: Move all other sql syntax changes to appropriate files in that directory.

tests/serverpod_test_server/test_integration/database_operations/explicit_column_name/crud_test.dart
```diff
@@ -0,0 +1,108 @@
...
+
+    group('when an object is retrieved by its id', () {
+      test('it returns the same object after insertion', () async {
+        final inserted = await db.insertRow(session, data);
```
SandPod: Suggestion: Make the insertion part of the setup.

Create a `Given` group (this will be a `withServerpod` group) names `Given an entry in the database for a table with an explicit column name` and then have all the `When` focus on the action performed and the `Then` focus on the outcome.


`When` and `Then` can be be in the same description (for a test for example) if the action is not sharable or worth breaking out.

SandPod: And use the `Given, When, Then` pattern for all tests. This makes our tests consistent and makes it easier to parse for future developers reading them.

tests/serverpod_test_server/test_integration/database_operations/explicit_column_name/relational_query_test.dart
```diff
@@ -0,0 +1,81 @@
...
+      session,
+      where: (t) => pod.Constant.bool(true),
+    );
+    await Department.db.deleteWhere(
+      session,
+      where: (t) => pod.Constant.bool(true),
+    );
+  });
+
+  group(
```
SandPod: Suggestion: Could you add a model that validates a one to one relation as well?

The code path is slightly different when building the query, so this assures this works as expected for this case as well.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -2682,6 +2684,8 @@ class SerializableModelLibraryGenerator {
       if (field.type.isVectorType)
         'dimension': literalNum(field.type.vectorDimension!),
       if (field.defaultPersistValue != null) 'hasDefault': literalBool(true),
+      if (field.columnName != field.name)
```
SandPod: Suggestion(NIT): Make this an extension on the class to remove code duplication and enable `field.hasColumnNameOverride`

tools/serverpod_cli/test/generator/dart/server_code_generator/table_class_test.dart
```diff
@@ -826,4 +829,106 @@ void main() {
...
+          var constructor =
+              CompilationUnitHelpers.tryFindConstructorDeclaration(
+                maybeClassNamedExampleTable!,
+                name: null,
+              );
+
+          expect(
+            constructor?.toSource(),
+            contains(
+              "$fieldName = _i1.Column$columnType('$columnName', this, "
```
SandPod: Suggestion: Remove the dependency on `_i1.` in the pattern matcher.

This is an implementation detail that might change because of unrelated changes. We don't want these tests to start failing because we added a new import to the default generated file.

Please limit the string pattern matching and remove anything that might change.

Apply this to the other tests.

packages/serverpod/lib/src/serialization/serialization_manager.dart
```diff
@@ -14,4 +14,22 @@ abstract class SerializationManagerServer extends SerializationManager {
...
+  ]) {
+    t ??= T;
+    final table = getTableForType(t);
+
+    if (table == null || data is! Map<String, dynamic>) {
+      return deserialize<T>(data, t);
+    }
+
+    // Create json with expected field name for SerializableModel
+    final json = table.parseRow(data);
```
SandPod: Discussion: It would have been great if we could have streamed the data from parseRow. Feels a bit unfortunate that we build a completely new map just to replace a couple of keys. Is there anything we can do there?

SandPod (commented):
I think the approach looks good, there are a couple of suggestions to fix. But I don't see any major concerns with this.

I am missing a test for migrating a column to a new name. Could you add such a test in migration e2e tests?

But to safeguard against us missing something, would it make sense to release this as an experimental feature first? And then we can promote it once we see all scenarios are covered in real world usage?

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
jwelmac: Very good point. I should have recognized the repetition. And that makes it more readable too.

packages/serverpod/lib/src/serialization/serialization_manager.dart
jwelmac: Let me think about this and see what could be done.

jwelmac:
@SandPod 

> I think the approach looks good, there are a couple of suggestions to fix. But I don't see any major concerns with this.

Thank you so much for the timely review.

> 
> I am missing a test for migrating a column to a new name. Could you add such a test in migration e2e tests?

I didn't write a test for this but actually [committed the code](https://github.com/Khuwn-Soulutions/serverpod/commit/12655e31e0cd778dd97307feee844eeeb5fe2333) showing same.
Having a test to check this is definitely wise though. As it will confirm the feature and identify any regressions during further development.

> But to safeguard against us missing something, would it make sense to release this as an experimental feature first? And then we can promote it once we see all scenarios are covered in real world usage?

I think this is a great way to get it into the hands of others to stress test it, get feedback and give everyone time to think through the permutations and repercussions of this change. 
As we interact with it we may also find other pathways to optimize or improve this feature.

SandPod:
@jwelmac 

> In the meantime, I have been doing some work on my package...

That looks very interesting! And will be really helpful if interacting with or migrating from Supabase. Cool project!


> I think this is a great way to get it into the hands of others to stress test it, get feedback and give everyone time to think through the permutations and repercussions of this change.

Great, it will be enough to guard against using the column keyword if the experiment is not enabled. This PR shows how to remove experimental from a keyword, but you should be able to use it as instructions for how to hide this keyword: https://github.com/serverpod/serverpod/pull/4236

tests/serverpod_test_server/test_integration/database_operations/explicit_column_name/crud_test.dart
jwelmac: Done!

tests/serverpod_test_server/test_integration/database_operations/explicit_column_name/crud_test.dart
jwelmac: Done!

tests/serverpod_test_server/test_integration/database_operations/explicit_column_name/crud_test.dart
jwelmac: Done!

tests/serverpod_test_server/test_integration/database_operations/explicit_column_name/relational_query_test.dart
jwelmac: Done!

packages/serverpod/lib/src/serialization/serialization_manager.dart
jwelmac: @SandPod 
Wow. So I thought long and hard about this, when I was first implementing it and again while attempting the refactor. In the end I came upon a simple solution I wish I had thought of sooner. 
Use aliases in the sql to map the column name to the field name if there is a column name override.
This eliminated the need for row parsing and the deserializeFromDatabase since the database returns the json with the expected field names.

SandPod: Interesting, is this solution already implemented? If so, I'll take a new full look at the solution. This should limit the impact of this PR quite a bit right?

jwelmac: @SandPod 
Yes it is. It does limit the impact. Take a look.

jwelmac:
Previously failing test now fixed.

jwelmac:
@SandPod 
Please review and let me know if with the new approach and comprehensive tests if its good for release. Or do I need to still set it up as experimental?

tests/serverpod_test_server/test_integration/database_operations/explicit_column_name/crud_test.dart
```diff
@@ -0,0 +1,81 @@
...
+      );
+
+      test(
+        'when updating the entry then the updated object is returned',
+        () async {
+          const newUserName = 'newUserName';
+          inserted.userName = newUserName;
+          final updated = await db.updateRow(session, inserted);
+
+          expect(updated.userName, newUserName);
```
SandPod: Suggestion: Please validate that the updated entry can be retrieved from the database.

When we run updateRow we copy some data over from the inserted object into the returned one. This will help us detect if this logic is ever broken.

I would suggest converting this test into a `group` for the `when updating the entry` section that does this in a setup and then has two tests that validate these two things.

tests/serverpod_test_server/test_integration/database_operations/explicit_column_name/relational_query_test.dart
```diff
@@ -0,0 +1,127 @@
+import 'package:serverpod/serverpod.dart' as pod;
+
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+import 'package:serverpod_test_server/test_util/test_serverpod.dart';
+
+import 'package:test/test.dart';
+
+void main() async {
+  final session = await IntegrationTestServer().session();
```
SandPod: Suggestion: Use `withServerpod` here as well.

packages/serverpod/lib/src/database/concepts/table_relation.dart
```diff
@@ -95,7 +95,7 @@ class TableRelation {
 
   /// The foreign field name joined on.
   String get foreignFieldName {
-    return _tableRelationEntries.last.foreignField.columnName;
+    return _tableRelationEntries.last.foreignField.fieldName;
```
SandPod, line 98: Question: Shouldn't other parts of this class also be modified as a result of this? Say for example the `foreignFieldNameWithJoins`?  Why is that field not modified?

jwelmac: Very good point. I will fix those and ensure they are covered by tests.

jwelmac:
Thanks @SandPod for the review.
Yes I do agree on the decision to go experimental first.
I will fix the merge and double check locally that all tests pass.

packages/serverpod/lib/src/database/concepts/table_relation.dart
jwelmac: Ok so after digging into the other fields I noticed a pattern. The `field.columnName` is required for use in building the SQL query to ensure that the database receives the expected column name. Conversely, the `field.fieldName` is needed when parsing a result from the database since all `Select` queries or `Returns` statements will alias the `columnName as fieldName`.

So after running an analysis of the `TableRelation getters` using Antigravity with these constraints the following was found.

> 
> | Getter | Used For | Current Value | Correct? | Needs Change? |
> |---|---|---|---|---|
> | fieldName | Create Query | columnName | Yes | No |
> | fieldQueryAlias | Create Query | queryAlias (contains columnName) | Yes | No |
> | fieldQueryAliasWithJoins | Create Query | columnName | Yes | No |
> | fieldNameWithJoins | Create Query | columnName | Yes | No |
> | foreignFieldName | Process Result | fieldName | Yes | No |
> | foreignFieldNameWithJoins | Create Query | columnName | Yes | No |
> 
> **Analysis:**
> 
> `foreignFieldName` is used in `postgres_result_parser.dart` to map results. Since it operates on resolved rows (where keys are field names), it correctly returns `fieldName`.
> `fieldQueryAliasWithJoins` is used in `database_connection.dart` to extract primary keys from raw result sets. Since raw result sets use column aliases (based on column names), it correctly returns `columnName`.
> All other getters are used in `sql_query_builder.dart` for constructing SQL queries, so they correctly use `columnName`.

It was right on all the above except for `fieldQueryAlias` that was referencing `queryAlias` that  used the `fieldName` for the column. To resolve this I created another getter on the column `fieldQueryAlias` that was using the `fieldName` and reverted the `queryAlias` to using the `columnName`. This ensured that the column name is correctly referenced in queries for relations but accurately aliased and parsed in `Selects` and `Returns`.

tests/serverpod_test_server/test_integration/database_operations/explicit_column_name/relational_query_test.dart
jwelmac: Done!

tests/serverpod_test_server/test_integration/database_operations/explicit_column_name/crud_test.dart
jwelmac: Done!

jwelmac:
@SandPod
Ready for your review again.

packages/serverpod/lib/src/database/sql_query_builder.dart
```diff
@@ -537,13 +539,29 @@ String _buildColumnAliases(List<Column> columns) {
...
 }
 
+/// Builds the returning clause for a query.
+String buildReturningClause(Table table, {String? tableAlias}) {
+  if (!table.hasColumnMapping) return '*';
+  return table.columns
+      .map((column) {
+        final queryColumnName = tableAlias != null
+            ? '$tableAlias.${column.columnName}'
+            : column.toString();
```
marcelomendoncasoares: Suggestion: Use the same quoted pattern on both: `"$tableAlias"."${column.columnName}"`. It might be clearer to remove the `toString` completely and explicitly paste the identifier.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -989,6 +989,74 @@ class Restrictions {
...
+
+    if (definition.tableName == null) {
+      return [
+        SourceSpanSeverityException(
+          'The "${Keyword.table}" property must be defined in the class to set '
+          'the ${Keyword.columnKey} on a field.',
+          span,
+        ),
+      ];
+    }
```
marcelomendoncasoares: Discussion: Should we be restrictive to only allow on table models or leave it be used on non-table models in case they are inherited by table models, so that child table models get the column rename?

Differently from `relation`, that has implications on which field get the relation defined, I would be more inclined to have this looser in favor of inheritance.

jwelmac: Hmmm. That's a good one.

I would say that the restriction could be lifted even though it will only be used by the table class during database querying. The benefit of doing that is that the parent non-table model can define the column name which would be applied to the table class or overwritten.

I will add test cases for this to ensure it works as expected.

packages/serverpod/lib/src/database/sql_query_builder.dart
jwelmac: Will do!

jwelmac:
Thanks for the review @marcelomendoncasoares. 
Two great points were noted. I will address those.

jwelmac:
@SandPod @marcelomendoncasoares 
Any idea why the job for unit tests for IO are failing even though all tests pass?

packages/serverpod/lib/src/database/sql_query_builder.dart
jwelmac: Done!

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
jwelmac: [Restriction removed](https://github.com/serverpod/serverpod/pull/4088/commits/c147e3895eb1ba8daf30094cc6ab7caa0a96be9c), [new model](https://github.com/serverpod/serverpod/pull/4088/commits/ef20d25d19ff2564fda66ab21176983c4e6dba0c) and [tests](https://github.com/serverpod/serverpod/pull/4088/commits/1e5ec6a86b20b686d2a8be002ef0dd41b3b21b46) added to verify database operation

jwelmac:
@SandPod @marcelomendoncasoares @coderabbitai Please review

marcelomendoncasoares:
> @SandPod @marcelomendoncasoares Any idea why the job for unit tests for IO are failing even though all tests pass?

@jwelmac We have one flaky unit test (`endpoint_method_test.dart`) that is being solved in another PR. You can ignore this error, it is unrelated to the changes here. The ones failing now are probably due to some missing migration on your PR for the test projects.

Sorry for the delay in the review, we're with our plates a bit full with the 3.0 release. But we'll try to do a next pass as soon as the CI is passing again.

jwelmac:
I will fix that and run the integration tests locally to ensure the merge is working as expected. 🤦‍♂️

All the best with the upcoming release. Looking forward to it.

jwelmac:
@marcelomendoncasoares @SandPod 
Migration issues fixed. The CI should pass once the workflow is approved.

jwelmac:
Thanks @marcelomendoncasoares.
Seems all is well now except for that flaky test failure.

marcelomendoncasoares:
> Thanks @marcelomendoncasoares. Seems all is well now except for that flaky test failure.

@jwelmac This current failure actually seems related to the changes in this PR:

```
+505 -1: test/database/database_query/insert_sql_query_test.dart: Given model with an explicit column field name when building insert query then output is a valid SQL query with column names. [E]
  Expected: 'INSERT INTO "user" ("id", "user_name", "age") VALUES (33, \'Alex\', 33) RETURNING *'
    Actual: 'INSERT INTO "user" ("id", "user_name", "age") VALUES (33, \'Alex\', 33) RETURNING "user"."id" AS "id", "user"."user_name" AS "userName", "user"."age" AS "age"'
     Which: is different.
            Expected: ... RETURNING *
              Actual: ... RETURNING "user"."id ...
                                    ^
             Differ at offset 80
  
  package:matcher                                                expect
  test/database/database_query/insert_sql_query_test.dart 166:9  main.<fn>.<fn>
  
+741 -1: Some tests failed.

Consider enabling the flag chain-stack-traces to receive more detailed exceptions.
For example, 'dart test --chain-stack-traces'.
```

Also, the flaky test is now fixed, so it will no longer impact the tests.

jwelmac:
Thanks I missed that because I scrolled to the end of the logs and saw that all tests had passed.
I see it now.
Fixing same. Thanks for the update.

jwelmac:
Thanks @marcelomendoncasoares.
Same now fixed.
I will now note that the unit tests job contains multiple test summaries so I need to check throughout and not only at the end of the logs.
I will also take some time to look through all the jobs run in the workflow to better familiarize myself.

jwelmac:
@SandPod @marcelomendoncasoares 
Congratulations on the version 3 release. I have updated the code by merging in the latest changes.

SandPod:
@jwelmac - Thanks you! 🙌 

I'm taking a look at this PR now, but the branch currently has conflict with main. Could you update it one last time? :)

jwelmac:
@SandPod - you're welcome. 

Not a problem. You guys move fast so I have gotten used to checking if I need to update at least daily. 
On it!

jwelmac:
Thanks @SandPod 
I am glad I took on this challenge. I learnt a lot, interacted with some great people and was able to contribute to a very meaningful project.

Thanks for the guidance and questions along the way from you and @marcelomendoncasoares.

jwelmac:
Now on to updating the documentation so its usage can be clear to all who wish to utilize it.

PR #4119, opened 2025-10-28
fix: Returns false if the database is enabled but turned off.

SandPod:
One-off fix without tests since we have no tests for the area.

If the database was enabled but the check failed with an exception, their would not report as unhealthy.

## Pre-launch Checklist

## Breaking changes

None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed health check system to consistently report database health status as unhealthy when exceptions occur, ensuring all health metrics are properly emitted instead of being omitted.


SandPod:
@coderabbitai review

PR #4126, opened 2025-10-29
feat: Add support for RPC middlewares

gitrema:
## Summary

Adds experimental middleware support to Serverpod by integrating relic's native `Middleware` system. Middleware enables interception, modification, and short-circuiting of HTTP requests and responses before they reach endpoint handlers.

## Motivation

Take a look at Issue https://github.com/serverpod/serverpod/issues/4194

**Design Decisions**

  1. Relic Native Types: No custom abstraction - uses relic's Middleware typedef directly for simplicity and familiarity
  2. Immutability: List.unmodifiable() prevents mutations after Server construction
  3. WebSocket Bypass: WebSocket upgrade requests require direct connection access and cannot pass through middleware
  4. ExperimentalApi Pattern: Follows existing pattern for exposing experimental features

## Pre-launch Checklist

## Breaking changes

No breaking changes, the new feature is behind an experimental flag.

## Summary by CodeRabbit

* **New Features**
  * Added middleware support to Serverpod servers for request and response processing.
  * Introduced built-in logging middleware with configurable output destinations and optional verbose mode for detailed header logging.
  * Added middleware configuration validation to warn about large middleware lists.


gitrema:
@nielsenko would you mind to review this one when you have time, thanks.

gitrema:

gitrema:
@coderabbitai review

SandPod:
@gitrema Could you comment on this issue so that I can assign it to you? https://github.com/serverpod/serverpod/issues/4194

packages/serverpod/lib/src/server/experimental_features.dart
```diff
@@ -7,6 +9,85 @@ class ExperimentalFeatures {
...
+  /// ```dart
+  /// final myMiddleware = createMiddleware(
+  ///   onRequest: (request) {
+  ///     print('Request: ${request.method} ${request.requestedUri}');
+  ///     return null; // Continue to next handler
+  ///   },
+  ///   onResponse: (response) {
+  ///     print('Response: ${response.statusCode}');
+  ///     return response;
+  ///   },
```
nielsenko: I suggest we don't put `createMiddleware` in the limelight. This is inherited from shelf, and I prefer if people don't use it and become acquainted with the canonical way

packages/serverpod/lib/src/server/middleware/middleware_validator.dart
```diff
@@ -0,0 +1,108 @@
...
+/// - Combine related middleware when possible
+/// - Avoid registering the same middleware multiple times
+/// - Order middleware from most general to most specific
+///
+/// ## See Also
+///
+/// - [Middleware] - The relic middleware type
+/// - [ExperimentalFeatures.middleware] - How to register middleware
+/// - [Pipeline] - How middleware are composed
+class MiddlewareValidator {
```
nielsenko: Is this really worth it? Why 10? I think I prefer dropping it, but perhaps I can be persuaded.

packages/serverpod/lib/src/server/server.dart
```diff
@@ -104,6 +104,17 @@ class Server implements RouterInjectable {
...
+  /// List of middleware to process HTTP requests.
+  /// Middleware is executed in the order it appears in this list.
+  ///
+  /// This list is immutable to ensure the middleware pipeline remains
+  /// consistent after server construction.
+  ///
+  /// **Note:** WebSocket upgrade requests (`/websocket`, `/v1/websocket`)
+  /// bypass middleware entirely as they require direct access to the
+  /// connection upgrade mechanism.
+  final List<Middleware> _middleware;
```
nielsenko: Do we need to track this? Perhaps just add them as they arrive?

packages/serverpod/lib/src/server/server.dart
```diff
@@ -132,6 +146,12 @@ class Server implements RouterInjectable {
     if (serverpod.config.loggingMode == ServerpodLoggingMode.verbose) {
       router.use('/', _verboseLogging);
     }
+
+    // Register user-provided middleware
+    for (var middleware in _middleware) {
+      router.use('/', middleware);
+    }
+
```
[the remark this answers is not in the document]
nielsenko: Either that, or we should consider giving user the full control over the middleware stack. Ie. allow him to replace the default middleware as well.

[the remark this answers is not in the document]
gitrema: I tend to agree with coderabbit here. I see your point though and I think we should discuss it more and follow up with another PR. It is not the intention of this PR to enable users to replace the default ones so if you don't have further objections I'll just follow the coderabbit suggestion for the time being

packages/serverpod/lib/src/server/middleware/middleware_validator.dart
gitrema: I was not sure either about its worth. At the end I left it to have feedback on it. It might be more useful to capture metrics about how long the middleware chain takes to execute and single middleware as well. Not as part of this PR though, so I'll remove the validator.

packages/serverpod/lib/src/server/server.dart
gitrema: No, we don't need to track it. I am not sure how to pass the list from the `Constructor` to the `injectIn` method where it is needed.

packages/serverpod/lib/src/server/experimental_features.dart
```diff
@@ -7,6 +9,69 @@ class ExperimentalFeatures {
...
+  ///     middleware: middleware,
+  ///   ),
+  /// );
+  /// ```
+  ///
+  /// ## Creating Custom Middleware
+  ///
+  /// ### Using Handler wrapping
+  /// ```dart
+  /// Middleware myMiddleware(Handler innerHandler) {
```
nielsenko:
```suggestion
  /// Handler myMiddleware(Handler innerHandler) {
```

packages/serverpod/lib/src/server/server.dart
```diff
@@ -104,6 +104,17 @@ class Server implements RouterInjectable {
...
 
+  /// List of middleware to process HTTP requests.
+  /// Middleware is executed in the order it appears in this list.
+  ///
+  /// This list is immutable to ensure the middleware pipeline remains
+  /// consistent after server construction.
+  ///
+  /// **Note:** WebSocket upgrade requests (`/websocket`, `/v1/websocket`)
+  /// bypass middleware entirely as they require direct access to the
+  /// connection upgrade mechanism.
```
nielsenko: This is not correct. As is, middleware is applied at `/` which also applies to web-sockets (at `/websocket/` and `/v1/websocket/`)

packages/serverpod/lib/src/server/server.dart
```diff
@@ -132,9 +147,18 @@ class Server implements RouterInjectable {
...
       ..use('/', _headers)
-      ..use('/', _reportException)
+      ..use('/', _reportException);
+
+    // Register user-provided middleware
+    for (var middleware in _middleware) {
+      router.use('/', middleware);
+    }
+
+    router
```
nielsenko: This is fine, but just for clarification. You could also loop over the middleware after adding the handlers. Middleware is applied by hierarchy first, then order, meaning that even though you call `use` after `add` (`get`,  etc.) it will still be applied to all handlers below that path.

The order is only important when attaching middleware at the same path. In that case they are composed first-to-last as outer-to-inner.

In particular the middleware is also applied to `_health`, `/websocket`, `/v1/websoket`, `/serverpod_cloud_storage`, and `/**` (ie. the serverpod endpoints).

tests/serverpod_test_server/test/core_middleware_ordering_test.dart
```diff
@@ -0,0 +1,262 @@
+import 'package:http/http.dart' as http;
+import 'package:serverpod/serverpod.dart';
+import 'package:test/test.dart';
+
+import 'package:serverpod_test_server/src/generated/protocol.dart';
+import 'package:serverpod_test_server/src/generated/endpoints.dart';
+
+/// Tests to verify that user middleware cannot bypass core middleware
+/// (_headers and _reportException).
```
nielsenko, line 9: I'm still wondering if we might want to allow overriding all middleware. But fine for now..

tests/serverpod_test_server/test/middleware/api_middleware_forwarding_test.dart
```diff
@@ -0,0 +1,271 @@
...
+          body: '{"value": 42}',
+        );
+
+        // Wait for middleware to execute (with timeout)
+        await middlewareExecuted.future.timeout(
+          Duration(seconds: 5),
+          onTimeout: () {
+            fail('Middleware did not execute within timeout');
+          },
+        );
```
nielsenko: Do we need a completer? Isn't it enough to just test that the header was added?

tests/serverpod_test_server/test/middleware/logging_middleware_test.dart
```diff
@@ -0,0 +1,594 @@
...
+        final logMessage = logMessages[0];
+
+        // Extract duration from log (format: "...200 (123ms)")
+        final durationMatch = RegExp(r'\((\d+)ms\)').firstMatch(logMessage);
+        expect(durationMatch, isNotNull, reason: 'Should include timing');
+
+        final duration = int.parse(durationMatch!.group(1)!);
+        // Duration should be reasonable (0-2000ms for a simple request)
+        expect(duration, greaterThanOrEqualTo(0));
+        expect(duration, lessThan(2000));
```
nielsenko: I hope to see sub-millisecond request handling eventually 😄 .. also (from experience) adding timing dependent test-cases like this is inherently brittle. Sometimes it is hard to avoid, but i don't think the actual duration is important here.

tests/serverpod_test_server/test/middleware/logging_middleware_test.dart
```diff
@@ -0,0 +1,594 @@
...
+      () async {
+        final beforeTime = DateTime.now().toUtc();
+
+        await httpClient.post(
+          endpoint,
+          headers: {'Content-Type': 'application/json'},
+          body: '{"value": 42}',
+        );
+
+        final afterTime = DateTime.now().toUtc();
```
nielsenko:
```suggestion
        final afterTime = DateTime.timestamp();
```

nielsenko (commented):
Some more comments 😊

packages/serverpod/lib/src/server/server.dart
nielsenko: My thinking was that it could be `addMiddleware` method instead of a ctor argument. What is your thoughts on that?

nielsenko: .. but I guess the current approach where the list is setup as part of the experimental feature is clean.. Just wondering, if once graduated it would be better to have an `addMiddleware` method, similar to what we have on `WebServer`.

packages/serverpod/lib/src/server/server.dart
gitrema: Do you see anything possibly wrong if the middlewares runs for those special endpoints, especially for the websocket ones?

tests/serverpod_test_server/test/middleware/api_middleware_forwarding_test.dart
gitrema: yes, it is

packages/serverpod/lib/src/server/server.dart
gitrema: I'll remove the comment

packages/serverpod/lib/src/server/server.dart
gitrema: As per our discussion, it will be removed as experimental feature and the `addMiddleware` method will be available

packages/serverpod/lib/src/server/server.dart
nielsenko: No, I think it is okay. You can use middleware on web-socket endpoints to drop connections before they are started fx. We just need to make sure the docs are not incorrect.

examples/middleware/middleware_server/lib/src/middleware/logging_middleware.dart
```diff
@@ -0,0 +1,165 @@
...
+///
+/// The middleware logs to stdout by default, but accepts an optional
+/// [logger] function for custom logging implementations.
+///
+/// **Note**: WebSocket requests bypass all middleware and will not be logged
+/// by this middleware.
+///
+/// ## Basic Usage
+///
+/// ```dart
```
nielsenko, line 18: The code blocks in this doc comment are no longer correct

packages/serverpod/test/server/middleware/middleware_export_test.dart
```diff
@@ -0,0 +1,144 @@
...
+/// Tests that verify middleware-related types and implementations are properly
+/// exported from the serverpod.dart public API.
+///
+/// These tests ensure that users can access all necessary middleware functionality
+/// without needing to import internal implementation files.
+void main() {
+  group('Middleware exports', () {
+    group('Relic middleware types', () {
+      test(
+        'Given middleware exports When accessing Middleware type Then it is available',
```
nielsenko: nit.
```suggestion
        'Given middleware exports '
        'when accessing Middleware type '
        'then it is available',
```
So no capitalising of _When_ and _Then_ here, and possibly split across multiple lines, seems to be the preferred format in serverpod tests.

packages/serverpod/test/server/middleware/middleware_export_test.dart
```diff
@@ -0,0 +1,144 @@
...
+        () {
+          // Verify that the Handler type from relic is accessible
+          // ignore: prefer_function_declarations_over_variables
+          Handler handler = (Request req) async {
+            return Response.ok();
+          };
+
+          expect(handler, isA<Handler>());
+        },
+      );
```
nielsenko: This adds little value. The compiler has already enforced that `handler` is a `Handler`

packages/serverpod/test/server/middleware/middleware_export_test.dart
```diff
@@ -0,0 +1,144 @@
...
+          // Verify that the Middleware type from relic is accessible
+          // This is a function type: Handler Function(Handler)
+          // ignore: prefer_function_declarations_over_variables
+          Middleware middleware = (Handler innerHandler) {
+            return (Request req) async {
+              return await innerHandler(req);
+            };
+          };
+
+          expect(middleware, isA<Middleware>());
```
nielsenko: Compiler guarantees this.

packages/serverpod/test/server/middleware/middleware_export_test.dart
```diff
@@ -0,0 +1,144 @@
...
+      );
+
+      test(
+        'Given middleware exports When accessing Pipeline class Then it is available',
+        () {
+          // Verify that the Pipeline class from relic is accessible
+          const pipeline = Pipeline();
+          expect(pipeline, isA<Pipeline>());
+        },
+      );
```
nielsenko: Again guaranteed.

packages/serverpod/test/server/middleware/middleware_export_test.dart
```diff
@@ -0,0 +1,144 @@
...
+
+      test(
+        'Given middleware exports When accessing Request type Then it is available',
+        () {
+          // Verify that Request type is accessible (needed for middleware)
+          expect(Request, isA<Type>());
+        },
+      );
```
nielsenko: Any Dart type is `Type`?

packages/serverpod/test/server/middleware/middleware_export_test.dart
```diff
@@ -0,0 +1,144 @@
...
+    });
+
+    group('Serverpod middleware implementations', () {
+      test(
+        'Given middleware exports When accessing loggingMiddleware Then it is available',
+        () {
+          // Verify that the loggingMiddleware factory is accessible
+          final middleware = loggingMiddleware();
+          expect(middleware, isA<Middleware>());
+        },
```
nielsenko: I'm not a fan of all these "is available tests". Instead of creating runtime test cases, why not just have a file that use/reference them, and fails at compile time (if not available)?

Basically `middleware_export_test.dart` could contain no tests, but just reference the classes/functions you would like to ensure are available.

Then the real test becomes if this file can be loaded or not. That would be a lot shorter, and avoid all the dummy _would_have_been_caught_by_compiler_ runtime tests.

nielsenko (changes requested):
More feedback.

I don't like `middleware_export_test.dart`. It is full of runtime tests that would be trivially caught by the compiler (if they were not true), and hence I was confused by them.

I get that you want to test that these types, functions, etc. are available to the user, but I think it would be better to have a test file that contains not runtime tests. Simply reference the entities needed. Add a big comment explaining why this test file is weird, and warns against simply point importing the missing stuff, if it starts to fail.

packages/serverpod/test/server/middleware/middleware_export_test.dart
gitrema: it makes sense, lemme update the file following your suggestion

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -383,7 +383,6 @@ class Serverpod {
       _initializeServerpod(
         args,
         config: config,
-        experimentalFeatures: experimentalFeatures,
```
nielsenko, line 386: I know this feature is no longer behind an experimental flag, but should we remove this completely?

nielsenko (approved):
I have a single comment to consider. Other than that it looks good. Thank you @gitrema

packages/serverpod/lib/src/server/serverpod.dart
gitrema: good catch!

it was introduced in this commit: https://github.com/gitrema/serverpod/commit/8af44031052dbf11ec7edcd30f304f2fd837754c

but even then it was passed but never used by `_initializeServerpod`.
it does not hurt removing it but of course, technically it is not the job of this PR to correct or remove that.

Shall we merge the PR as it is now?

nielsenko: Okay, I didn't check for use. I'm fine with merging.

examples/middleware/middleware_server/lib/src/middleware/logging_middleware.dart
```diff
@@ -0,0 +1,155 @@
...
+/// - If [errorLogger] is provided, errors use it
+/// - Otherwise, if [logger] is provided, errors use it
+/// - Otherwise, errors go to stderr (default behavior)
+///
+/// ## See Also
+///
+/// - [Middleware] - The relic middleware type
+/// - [Pipeline] - For composing multiple middleware
+/// - [Server.addMiddleware] - How to register middleware
+Middleware loggingMiddleware({
```
SandPod, line 87: Suggestion: Move this into the Example package.

We already supply logging with Serverpod for all endpoints. Even if it is not ideal, I think it canbe confisuing present multiple solutions to the same problem.
But I do think this example is important as it might be used and modified to add other type of monitoring or reporting (New Relic or Sentry logging).

We do have an initiative planned to overlook our general logging in the framework.

packages/serverpod/test/server/middleware/middleware_export_test.dart
```diff
@@ -0,0 +1,35 @@
+// ignore_for_file: unused_local_variable, prefer_function_declarations_over_variables
+
+import 'package:serverpod/serverpod.dart';
+
+/// Compile-time verification that middleware-related types are properly
+/// exported from the serverpod.dart public API.
+///
+/// This file contains no runtime tests. If any referenced type or function
+/// is not exported, this file will fail to compile, which is the test.
+void main() {
```
SandPod: Suggestion: Remove this test.

With the move of the logging middleware to the example project we will test the exports of the package automatically.
Therefore, this test can probably be removed.

tests/serverpod_test_server/test/middleware/logging_middleware_test.dart
```diff
@@ -0,0 +1,621 @@
+import 'package:http/http.dart' as http;
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_server/test_util/test_serverpod.dart';
+import 'package:test/test.dart';
+
+void main() {
```
SandPod: Suggestion(Optional): These test can be removed.

When moving the logging middleware to the example project, we don't need to have specific validation on it. We can keep them if you think it is important.

SandPod (changes requested):
Awesome work! Really nice to see this feature finished! :D

I have some last NIT comments around what we expose. I think we shouldn't cause confusion by providing multiple ways handling logging in the framework. But I do believe that the logging middleware is a good example on how to use middleware that can be adapted for different purposes.

For that reason I think we should move it to the Example project.

I know you have already spent some time on this, if you don't have time to fix this in the near future, we could do a quick push to the branch and solve the last hurdle :)

examples/middleware/middleware_server/lib/src/middleware/logging_middleware.dart
gitrema: Agree, My intention was to showcase how a middleware can be used. So it makes sense to move it to the example package

PR #4134, opened 2025-10-30
refactor: Use RelicApp in Server

### LINKED ISSUE #4101: Refactor request handling to use Router class from relic package
## Description
The current server implementation uses manual path and HTTP verb dispatch. This should be refactored to use the `Router` class from the `relic` package for better maintainability and extensibility.

## Affected Files
- `packages/serverpod/lib/src/server/server.dart:190`

## Context
```dart
// TODO: Use Router instead of manual dispatch on path and verb
if (uri.path == '/') {
  // Perform health checks
  ...
}
```

## Suggested Implementation
- Use the `Router` class from the `relic` package (already a dependency)
- Implement route matching and dispatching logic using relic's Router API
- Migrate existing manual dispatch code to use Router
- Add support for route parameters and pattern matching as provided by relic
- Maintain backward compatibility during transition
- Add tests for route matching edge cases

## Impact
Code maintainability, extensibility for adding new routes, better error handling

## Related
- PR #4085

nielsenko:
## Summary

Use Relic's router for endpoint dispatch in api server.

Closes: #4101

## Changes

- Migrated server to implement `RouterInjectable` and use `RelicApp` for request routing
- Removed `EndpointResult` classes.
- Changed `_handleUriCall` to return `Response` directly instead of `EndpointResult`
- Moved all exception-to-response mapping to the `_reportException` middleware
- Made `EndpointDispatchException` sealed to enable exhaustive pattern matching
- Simplified `NotAuthorizedException` to store `AuthenticationFailureReason` directly
- Updated `canUserAccessEndpoint` to return `AuthenticationFailureReason?` instead of result wrapper
- Removed `sessionLogId` from error responses to avoid exposing internal database IDs
- Removed helper methods `_toResponse` and `_logMalformedCalls` (no longer needed)
- Updated all existing tests to reflect new return types

## Pre-launch Checklist

## Breaking changes

None. This is an internal refactoring that doesn't affect the public API.

## Release Notes

* **New Features**
  * Added router injection capability for streamlined middleware and route registration.

* **Refactor**
  * Simplified authentication error handling, moving from verbose result objects to concise reason-based responses.
  * Restructured request dispatch and exception handling for improved maintainability.
  * Enhanced server startup architecture with injected application management.

### DIFF
--- a/packages/serverpod/lib/src/server/server.dart
+++ b/packages/serverpod/lib/src/server/server.dart
@@ -280,7 +280,14 @@ class Server implements RouterInjectable {
 
       final result = await next(req);
       return switch (result) {
-        Response() => result.copyWith(headers: headers),
+        Response() => result.copyWith(
+            headers: result.headers.isEmpty
+                ? headers
+                : result.headers.transform((mh) {
+                    for (final h in headers.entries) {
+                      mh[h.key] ??= h.value;
+                    }
+                  })),
         _ => result,
       };
     };
--- a/packages/serverpod/lib/src/server/server.dart
+++ b/packages/serverpod/lib/src/server/server.dart
@@ -231,14 +231,14 @@ class Server implements RouterInjectable {
     };
   }
 
-  Future<Result> _health(Request req) async {
+  Future<Result> _health(Request _) async {
     final metrics = (await performHealthChecks(serverpod)).metrics;
     final issues = metrics.where((m) => !m.isHealthy);
     final ok = issues.isEmpty;
-    if (ok) return Response.ok(body: Body.fromString('OK'));
+    final now = DateTime.timestamp();
+    if (ok) return Response.ok(body: Body.fromString('OK $now'));
     return Response(503, body: Body.fromDataStream(() async* {
-      final now = DateTime.timestamp();
-      yield utf8.encode('${ok ? 'OK' : 'SADNESS'} $now\r\n');
+      yield utf8.encode('SADNESS $now\r\n');
       for (final metric in issues) {
         yield utf8.encode('${metric.name}: ${metric.value}\r\n');
       }
--- a/packages/serverpod/lib/src/server/server.dart
+++ b/packages/serverpod/lib/src/server/server.dart
@@ -16,7 +16,7 @@ import 'package:serverpod/src/server/websocket_request_handlers/method_websocket
 
 /// Handling incoming calls and routing them to the correct [Endpoint]
 /// methods.
-class Server {
+class Server implements RouterInjectable {
   // Map of [WebSocket] connected to the server.
   // The key is a unique identifier for the connection.
   // The value is a tuple of a [Future] that completes when the connection is
@@ -124,18 +124,37 @@ class Server {
         _securityContext = securityContext,
         _port = port;
 
+  late final _app = RelicApp()..inject(this);
+
+  @override
+  void injectIn(RelicRouter router) {
+    if (serverpod.config.loggingMode == ServerpodLoggingMode.verbose) {
+      router.use('/', _verboseLogging);
+    }
+    router
+      ..use('/', _headers)
+      ..use('/', _reportException)
+      ..get('/', _health)
+      ..get('/websocket',
+          _dispatchWebSocket(EndpointWebsocketRequestHandler.handleWebsocket))
+      ..get('/v1/websocket',
+          _dispatchWebSocket(MethodWebsocketRequestHandler.handleWebsocket))
+      ..anyOf({Method.get, Method.options, Method.post},
+          '/serverpod_cloud_storage', _cloudStorage)
+      ..any('/**', _endpoints);
+  }
+
   /// Starts the server.
   /// Returns true if the server was started successfully.
   Future<bool> start(
       {required AuthenticationHandler authenticationHandler}) async {
     _authenticationHandler = authenticationHandler;
     try {
-      final server = RelicServer(() => IOAdapter.bind(
-            io.InternetAddress.anyIPv6,
-            port: _port,
-            context: _securityContext,
-          ));
-      await server.mountAndStart(_relicRequestHandler);
+      final server = await _app.serve(
+        address: io.InternetAddress.anyIPv6,
+        port: _port,
+        securityContext: _securityContext,
+      );
       _actualPort = server.port;
       _relicServer = server;
     } catch (e, stackTrace) {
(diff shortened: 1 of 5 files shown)


nielsenko:
@coderabbit review

nielsenko:
@coderabbit review

packages/serverpod/lib/src/server/server.dart
```diff
@@ -155,255 +174,163 @@ class Server {
...
+      });
+
+      // early exit on Method.options
+      if (isOptions) return Response.ok(headers: headers);
+
+      final result = await next(req);
+      return switch (result) {
+        Response() => result.copyWith(headers: headers),
+        _ => result,
+      };
```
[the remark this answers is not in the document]
nielsenko: Actually `result.headers` will always be empty here (currently), but we should perhaps not assume that.
Fixed in ba1561a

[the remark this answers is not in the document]
nielsenko: Actually I think it would be even better to not overwrite any existing headers here. I have changed that in eba4895e

packages/serverpod/lib/src/server/server.dart
```diff
@@ -155,255 +174,170 @@ class Server {
...
+      responseBuffer.writeln('SADNESS ${DateTime.now().toUtc()}');
+    }
+    for (var issue in issues) {
+      responseBuffer.writeln(issue);
     }
+
+    return Response(
+      allOk ? io.HttpStatus.ok : io.HttpStatus.serviceUnavailable,
+      body: Body.fromString(responseBuffer.toString()),
+      //headers: headers,
```
SandPod: Suggestion: Remove comment.

packages/serverpod/lib/src/server/server.dart
```diff
@@ -155,255 +174,170 @@ class Server {
...
-            case Body():
-              value = Response.ok(body: value, headers: headers);
-              continue response;
-            response:
-            case Response():
-              return value;
+  Handler _headers(Handler next) {
+    return (req) async {
+      final isOptions = req.method == Method.options;
+      // TODO: Make httpResponseHeaders a Headers object from the get-go.
```
SandPod: Question: Is this a short lived followup? Otherwise, please remove the TODO comment and file an issue for it.

packages/serverpod/lib/src/server/server.dart
```diff
@@ -155,255 +174,170 @@ class Server {
...
+    if (allOk) {
+      responseBuffer.writeln('OK ${DateTime.now().toUtc()}');
+    } else {
+      responseBuffer.writeln('SADNESS ${DateTime.now().toUtc()}');
+    }
+    for (var issue in issues) {
+      responseBuffer.writeln(issue);
     }
+
+    return Response(
```
SandPod: Suggestion(NIT): Switch to early return in this method to simplify the code?

packages/serverpod/lib/src/server/server.dart
```diff
@@ -155,255 +174,170 @@ class Server {
...
-            SerializationManager.encodeForProtocol(value),
-            mimeType: MimeType.json,
-          ),
-          headers: headers,
-        );
-    }
+      });
+
+      // early exit on Method.options
+      if (isOptions) return Response.ok(headers: headers);
```
SandPod, line 263: Suggestion: Can we register this as a separate route only accepting the options method?

packages/serverpod/lib/src/server/server.dart
```diff
@@ -155,255 +174,170 @@ class Server {
...
+                    }
+                  })),
+        _ => result,
+      };
+    };
+  }
+
+  FutureOr<Result> _cloudStorage(Request req) async {
+    final uri = req.requestedUri;
+    assert(uri.path == '/serverpod_cloud_storage');
```
SandPod, line 282: Question: Why is this assertion needed? We should control this in the registered path, right?

packages/serverpod/lib/src/server/server.dart
```diff
@@ -155,255 +174,170 @@ class Server {
...
+    return await _handleUriCall(uri, '', req);
```
SandPod: Discussion: Should we rename `_handleUriCall` to `_handleEndpointCall`?  Since it now only handles endpoints, right?

packages/serverpod/lib/src/server/server.dart
```diff
@@ -540,9 +459,37 @@ class Server {
           methodCallContext.arguments,
         );
 
-        return ResultSuccess(
-          result,
-          sendAsRaw: methodCallContext.endpoint.sendAsRaw,
+        // Handle success - construct Response directly
+        var value = result;
```
SandPod: Comment: Not a big fan of this switch statement logic. I see that this was already here from before, so we might not need to prove it. But it would have been nice to break it up a bit.

packages/serverpod/lib/src/server/server.dart
```diff
@@ -155,255 +174,170 @@ class Server {
...
-          this,
-          webSocket,
-          req,
-          () => _webSockets.remove(websocketKey),
-        );
+  ) {
+    return (req) async {
+      return WebSocketUpgrade((webSocket) async {
+        try {
+          // TODO(kasper): Should we keep doing this?
```
SandPod: Suggestion: Can we remove this comment now?

SandPod (commented):
Changes looks good, have a couple of comments and minor suggestions.

packages/serverpod/lib/src/server/server.dart
nielsenko: Issue is #4102 . Will remove.

packages/serverpod/lib/src/server/server.dart
nielsenko: This is middleware. We don't have verb specific middleware routing in Relic currently.

packages/serverpod/lib/src/server/server.dart
nielsenko: Yes, but `_handleUriCall` does use the uri. Right now it is basically handled like a regular generated endpoint.

packages/serverpod/lib/src/server/server.dart
nielsenko: It builds up a response depending on how pre-cooked the value is already. I can try to make it clearer, but I doubt it can be much more concise. I will split out in separate method.

packages/serverpod/lib/src/server/server.dart
nielsenko: Yes. It is actually removed in another PR, but I can do it in this instead/as well.

packages/serverpod/lib/src/server/server.dart
nielsenko: Not sure I follow. The previous if is just what header line to put in the body. Can you elaborate?

nielsenko: Now I get it. Old code, but will refactor.

packages/serverpod/lib/src/server/server.dart
nielsenko: Fixed in 34a19c5f. Much shorter and clearer.

packages/serverpod/lib/src/server/server.dart
nielsenko: Fixed in 34a19c5f. Returns early and build body from stream.

packages/serverpod/lib/src/server/server.dart
nielsenko: Fixed in 34a19c5f

packages/serverpod/lib/src/server/server.dart
nielsenko: Fixed in 34a19c5f

packages/serverpod/lib/src/server/server.dart
SandPod: I'm not sure I agree that the format now is easer to read. I was more looking for a:

```dart
if (allOK) {
  return Response.ok(body: Body.fromString('OK'));
}
// else all the compiled stuff when we had an error.
```

nielsenko: But this would be semantically different. We return the actual metrics, also when `allOk` is `true`. 
Not sure it is important, but what you suggest is breaking.

packages/serverpod/lib/src/server/server.dart
nielsenko: I'll split `handleUriCall` in two. We don't actually need to do much of it when serving `/serverpod_cloud_storage`. If my analysis is correct, then that is the only single segment "endpoint" we have.

packages/serverpod/lib/src/server/server.dart
SandPod: Thanks 🙏

packages/serverpod/lib/src/server/server.dart
nielsenko: Sorry, my bad. You are right. Fixed in edc12d3c

SandPod (approved):
👍 LGTM!

PR #4139, opened 2025-10-31
fix: Initialize endpoints on Server start.

SandPod:
Moves the initialize call for endpoint from the Serverpod constructor to the server start sequence.

This enables adding dependencies on global state to be configured before the endpoint is constructed.

This will be used in the new auth modules to ensure that AuthConfig is initialized before the endpoint is initialized, enabling us to throw an exception when the server is started with invalid configuration.

## Pre-launch Checklist

## Breaking changes

We don't give a guarantee today when the endpoint is initialized, so this should not be something that users are depending on.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server startup reliability with enhanced error handling for endpoint initialization failures. Server will now properly report and prevent startup if initialization encounters issues.

* **Refactor**
  * Consolidated endpoint initialization logic to execute at a single point during server startup for more predictable initialization behavior.

### DIFF
--- a/packages/serverpod/lib/src/server/server.dart
+++ b/packages/serverpod/lib/src/server/server.dart
@@ -127,6 +127,14 @@ class Server {
   /// Returns true if the server was started successfully.
   Future<bool> start() async {
     try {
+      endpoints.initializeEndpoints(this);
+    } catch (e, stackTrace) {
+      await _reportFrameworkException(e, stackTrace,
+          message: 'Failed to initialize endpoints.');
+      return false;
+    }
+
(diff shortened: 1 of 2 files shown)


SandPod:
@coderabbitai review

SandPod:
This PR needs to be complemented with lazy endpoint initialization in the test framework.

nielsenko (approved):
🚀

SandPod:
@coderabbitai We will not implement this feature right now since it is not critical. Please log an issue for this.

PR #4155, opened 2025-11-04
refactor!: Use shared challenge model for all user challenges in Email IDP.

SandPod:
Breaks out the stored information related to a challenge so that it can be reused between the two different types of challenges that currently exist.

We will soon introduce a new challenge that will be a one-time code used to reset the password.

## Pre-launch Checklist

## Breaking changes

Yes, this modifies the data structure for the unreleased email IDP.

### DIFF
--- a/tests/serverpod_new_auth_test/serverpod_new_auth_test_server/config/passwords.yaml
+++ b/tests/serverpod_new_auth_test/serverpod_new_auth_test_server/config/passwords.yaml
@@ -22,7 +22,7 @@ development:
   # service protocol.
   serviceSecret: "kYhQvD7M8ye1B2Z0kNymKFl_dzdAdrtu"
 
-  serverpod_auth_idp_email_passwordHashPepper: "emailPepper!"
+  serverpod_auth_idp_email_secretHashPepper: "emailPepper!"
   serverpod_auth_session_sessionKeyHashPepper: "sessionPepper!"
   serverpod_auth_jwt_algorithm: "HS512"
   serverpod_auth_jwt_secretKey: "secretKey!"
@@ -31,7 +31,7 @@ development:
 test:
   database: "A0EV-ij7y7e639gg0PrH5IYphMqAXVAE"
   redis: "xBjXhPyFVTSFrOuGNq1GDi564623qEGU"
-  serverpod_auth_idp_email_passwordHashPepper: "emailPepper!"
+  serverpod_auth_idp_email_secretHashPepper: "emailPepper!"
   serverpod_auth_session_sessionKeyHashPepper: "sessionPepper!"
 
 # Passwords used in your staging environment if you use one. The default setup


modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_idp_utils.dart
```diff
@@ -34,20 +34,20 @@ class EmailIDPUtils {
 
   /// Creates a new instance of [EmailIDPUtils].
   EmailIDPUtils({required final EmailIDPConfig config})
-      : passwordHash = EmailIDPPasswordHashUtil(
-          passwordHashPepper: config.passwordHashPepper,
-          passwordHashSaltLength: config.passwordHashSaltLength,
+      : hashUtil = EmailIDPHashUtil(
+          hashPepper: config.passwordHashPepper,
+          hashSaltLength: config.passwordHashSaltLength,
```
marcelomendoncasoares: Discussion: Since the hash is used for more than just password generation and we now explicitly name `hashUtil` to decouple from passwords only, should we change the configuration to something more generic? Like `secretHashPepper` and `secretHashSaltLength`?

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/models/email_account_challenge.spy.yaml
```diff
@@ -0,0 +1,11 @@
+class: EmailAccountChallenge
```
marcelomendoncasoares: Discussion: Since this challenge engine is generic, is it worth already moving it out of the email provider and treat it as a general utils that can be used by any provider? It would greatly simplify SMS login, for example, since it is nothing more than the combination of a challenge with a specific sending mechanism, or two-factor implementations.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_idp_utils.dart
SandPod: Aren't these configurations specifically used for the password? I don't think they are general for other parts of the code. That was the reason for keeping the names. I'll check

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/models/email_account_challenge.spy.yaml
SandPod: True! I'll move it into common.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_idp_utils.dart
SandPod: It was not specific. Will update the name to reflect this.

Issue #3530, opened 2025-05-10
serverpod create-migration exits with error but no error message


christerswahn:
### Describe the bug

When run in a non-serverpod directory, `serverpod create-migration` command exits with error code 126 and no error message.

### To reproduce

Run `serverpod create-migration` in a non-serverpod directory.

### Expected behavior

It should print an appropriate error message and user guidance. (Perhaps the message from the `ServerpodProjectNotFoundException`.)

### Library version

Dart 3.7

### Platform information

MacOS

### Additional context

It might be relevant to look at this issue together with #2071 .

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

FXschwartz:
@SandPod I'll take this one!

PR #4157, opened 2025-11-05
feat: Enable CLI commands to run from anywhere in project (#3688)

### LINKED ISSUE #3688: Allow Serverpod CLI to run from any directory within the project structure
### Problem to Solve

The Serverpod CLI currently requires project management commands—such as `create-migration` or `generate`—to be run from the `server/` directory of a Serverpod project. This limitation interrupts developer workflows, especially in monorepos or modular setups, where users often work from different directories within the project. It adds unnecessary friction when performing common tasks.

### Proposal

Enhance the CLI to allow project management commands to be executed from any directory within a Serverpod project. The CLI should:

1. Traverse upward or downwards from the current working directory with a limited depth.
2. Identify the Serverpod server root by locating a `pubspec.yaml` with a dependency on `serverpod`.
3. If multiple such projects are found during traversal, abort with a clear message:  
   "Multiple Serverpod projects detected. Please run this command from within the intended server directory."
4. If no project is found, display a helpful hint:  
   "No Serverpod server project detected. Make sure your project includes a `pubspec.yaml` with a dependency on `serverpod`."
5. Consider and support Dart workspace structures (e.g., multiple packages under a common root), ensuring compatibility with tools like `melos` and `dart pub`.

All commands related to project management should implement this behavior (e.g., `generate`, `create-migration`, etc.).

### Use Case

A developer inside the a serverpod project directory can run `serverpod generate` without navigating to `my_project/server/`. The CLI auto-locates the root server project using the `serverpod` dependency as a signal, even in Dart workspace setups.

### Alternatives

n/a

### Additional context

Aligns Serverpod with the usability standards of modern tools like Git and npm. By limiting the traversal depth and checking for a `serverpod` dependency, we avoid incorrect assumptions and improve developer experience, especially in complex monorepos or Dart workspaces.

None

- [ ] I want to work on this

FXschwartz:
Adds the ability to run the CLI commands (generate, create-migration, create-repair-migration) from anywhere in the repo by automatically locating the server package when running commands . The search respects repository boundaries (.git, melos.yaml) and handles monorepo/dart workspace structures with nested applications.

  - Auto-detects server directory from current location
  - Searches children, siblings, and upward through parents
  - Stops at repository/workspace boundaries
  - Prompts when multiple servers found
 
Fixes #3688 and #3530 

## Pre-launch Checklist

## Breaking changes
None


FXschwartz:

SandPod:
Hey @FXschwartz! Big thanks for taking this on!

Before examining your search algorithm, could you inspect the one we use for our Serverpod cloud CLI to ensure that it delivers a somewhat equal experience?

We have failed to port that code over to Serverpod yet, but now that you have area experience, perhaps you could make sure they make sense?

If they differ, feel free to argue for why your approach should be ported to that code instead of copying it.

The code for the Serverpod cloud cli search algorithm is available here: https://github.com/serverpod/serverpod_cloud_clients/blob/ed5f08c7e5da195a8ae4176f93ed1ea30f41fe4e/serverpod_cloud_cli/lib/util/scloud_config/file_finder.dart#L40

SandPod:
If you could comment on the two issues linked I can assign them to you to ensure that other developers to not take on this work in parallel.

FXschwartz:
> If you could comment on the two issues linked I can assign them to you to ensure that other developers to not take on this work in parallel.

Done! 

And yes for sure, I'll take a look at the cloud cli search and get back to you.

Thanks!

FXschwartz:
@SandPod  Alright, looking over the cloud cli code it's definitely cleaner and simpler. I'd vote to copy over the basic structure using the generic FileFinder<T> pattern and the file content validation callback.

There are some things I'd like to keep from my implementation that might not be necessary for the cloud cli but I think are important for this specific use case.

Here's what I see that both solutions do and their key differences. Let me know if I've missed something or gotten the cloud cli behavior wrong.

Both Solutions:
  - Check current directory
  - Search child directories up to 2 levels down

Cloud CLI
  - IF pubspec.yaml exists in current directory:
    - Go up 1 level (to parent)
    - Search down 1 level from parent
  - ELSE:
    - Stop (not found)
  - Limits:
    - Max upward: 1 level
    - Only searches up if pubspec.yaml present
    - Total depth from parent: 1 level

This PR's Solution
  - Check sibling patterns (_flutter → _server, _client → _server, _app → _server)
    - Just a quick win for typical serverpod project naming (my_app_client, my_app_server, etc)
  - Search upward through parent directories (up to 5 levels):
    - At each parent level, check if it's a repository boundary (.git, melos.yaml)
    - Search children of each parent:
        - Regular parent: search 1 level down
      - Boundary parent: search 2 levels down
    - Stop if boundary reached (don't escape repository)
  - Interactive prompt if multiple servers found
  - Limits:
    - Max upward: 5 levels (or until boundary)
    - Always searches up (not conditional)
    - Total depth from boundary parent: 2 levels
    - Total depth from non-boundary parent: 1 level

What I'd like to keep from my PR:
The main thing is the deeper upward search capability. In a dart workspace monorepo where the server is nested several levels deep (like backend/api/server), the cloud CLI would fail to find it, but this solution handles it. I've also added boundary detection to make sure we don't accidentally escape the current repository and trigger permission prompts as well as user prompting if multiple backend projects are found for the user to choose from. For the user prompting I've added an optional flag that would disabled the user prompting if this was used in a CI/CD environment which a prompt could cause it to hang.

Let me know if you have any questions concerns or feedback. Thanks!

SandPod:
@FXschwartz Thanks for comparing your search algorithm with the one used in cloud.

This sounds reasonable!

Please add tests that validate this works as expected and passes the CI.

When creating temporary directories to validate the search algorithm, please use the https://pub.dev/packages/test_descriptor package. It makes it really easy to create temp directory structures that can be validated in parallel.

This pattern is not always applied in the codebase today, but it should be.

FXschwartz:
@SandPod Ok I've updated the search flow here to match the pattern used in `serverpod_cloud_cli` I did make a change to instead of calling it `FileFinder` here I used `DirectoryFinder` as it makes more sense to me in this context but feel free to have me change that.
- On a side note are there any plans to maybe move some of this file/directory search logic into a common place where we can share it? Not sure if there are any other places that search for specific files with logic as to how many levels it searches.

Additionally I've update all my test cases to use the `test_descriptor` package when creating temp directories.

And finally I added some test cases to validate the behavior in CI environments.

Take a look when you have some free time and let me know your thoughts or requested changes.

SandPod:
@FXschwartz Thanks for the update! I'll take a look ASAP.

We don't have a place to share this logic at the moment. So we will have to update it here for now, and then I'll have the team sort that out at some point.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
```diff
@@ -0,0 +1,314 @@
...
+///
+/// If [directoryContentCondition] is provided, it is called with each
+/// directory found. If it returns false, the directory is not considered a match.
+///
+/// The search strategy:
+/// 1. Checks the current directory
+/// 2. Searches child directories (depth 2)
+/// 3. Checks for sibling directories with standard naming patterns (only if not at boundary)
+/// 4. Searches upward through parent directories (max 5 levels or until boundary)
+/// 5. At each parent level, checks its children for server directories
```
SandPod: Comment: Note for self - Do we search the same children multiple times?

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
```diff
@@ -0,0 +1,314 @@
...
+  ) {
+    var dirName = p.basename(dir.path);
+    var parent = dir.parent;
+
+    String? baseName;
+    if (dirName.endsWith('_flutter')) {
+      baseName = dirName.substring(0, dirName.length - 8);
+    } else if (dirName.endsWith('_client')) {
+      baseName = dirName.substring(0, dirName.length - 7);
+    } else if (dirName.endsWith('_app')) {
```
SandPod: Suggestion/Discussion: I understand that you have added this check for your specific usecase. But it is not a name that Serverpod would ever assign.

It feels wrong to include it in the general search algorithm for that reason. What we could do is make it possible to configure file ending additions through the `generator.yaml` config file, so that you can adjust this to fit your needs? It would be a specialized option that will have a very niche use case, but since you are implementing this, I think it is only right that we adapt it a bit to fit your needs.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
```diff
@@ -0,0 +1,314 @@
...
+  /// Attempts to find a sibling server directory based on naming conventions.
+  static Directory? _findSiblingServer(
+    Directory dir,
+    DirectoryContentCondition condition,
+  ) {
+    var dirName = p.basename(dir.path);
+    var parent = dir.parent;
+
+    String? baseName;
+    if (dirName.endsWith('_flutter')) {
```
SandPod: Suggestion: Could you add these to a `const` list instead and then use the same algorithm and stop at the first match? Feels like we could just iterate the list until we have a first match and move on with that, and the length removal could be derived from the entry.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
```diff
@@ -0,0 +1,314 @@
...
+      // 3. Check for standard naming pattern siblings
+      var siblingServer =
+          ServerDirectoryFinder._findSiblingServer(start, condition);
+      if (siblingServer != null) {
+        candidates.add(siblingServer);
+      }
+
+      // 4. Search upward through parent directories
+      var current = start.parent;
+      for (var i = 0; i < 5; i++) {
```
SandPod: Suggestion: Make `5` an input parameter to the function.

If we ever want to tweak this, that will be really helpful.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
```diff
@@ -0,0 +1,314 @@
...
+    var candidates = <Directory>[];
+    var visited = <String>{};
+
+    // 1. Check current directory first (fast path)
+    if (condition(start)) {
+      return start;
+    }
+
+    // 2. Search child directories (depth 2)
+    candidates.addAll(
```
SandPod, line 69: Question: Would it make sense to add an early return after we have searched the children? No real need to step upwards if we have already found a server at this point, right?

Or is there any other reason for continuing the search if we have a match here? What reasonable case would we miss?

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
```diff
@@ -0,0 +1,314 @@
...
+      }
+    } on FileSystemException catch (_) {
+      // skip directories that cannot be accessed
+    }
+
+    return results;
+  }
+
+  /// Prompts the user to select from multiple server directories.
+  static Future<Directory> _promptUserSelection(
```
SandPod, line 280: Suggestion: I believe CLI tools have support for interactive prompts. 

More specifically a `select` prompt is available here: https://github.com/serverpod/cli_tools/blob/main/packages/cli_tools/lib/src/prompts/select.dart

Any reason for not using it?

tools/serverpod_cli/test/util/server_directory_finder_test.dart
```diff
@@ -0,0 +1,494 @@
...
+        expect(result, isNotNull);
+        expect(
+          path.normalize(result!.path),
+          equals(path.normalize(testDir.path)),
+        );
+      },
+    );
+
+    test(
+      'Given sibling directory with _flutter suffix, '
```
SandPod: Suggestion(NIT): `Given _flutter suffix directory with sibling server directory`

Just to make it clear in the given that server directory is the sibling.
Apply same suggestion for the other tests.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
SandPod: Pershaps we should apply the same optimization for the sibling check since that is also just to make it faster when we regonize some pattern? No use in checking the rest of the parents if we recognize our structure?

tools/serverpod_cli/test/util/server_directory_finder_search_test.dart
```diff
@@ -0,0 +1,494 @@
...
+
+        expect(result, isNotNull);
+        expect(
+          path.basename(result!.path),
+          equals('server'),
+        );
+      },
+    );
+
+    test(
```
SandPod, line 263: Suggestion: Isn't the best way to test this to create a structure that looks like this:

server/
-- pubspec.yaml
-- subdir/
---- .git
---- secondSubdir/

Then start the search from `secondSubdir`? That would assert that the `server` directory is not found becuase the .git` stops this?

SandPod: Same for the other "boundary" markers.

SandPod: This might be what you are doing in the next couple of tests, but I don't think it is as clear as this structure.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
```diff
@@ -0,0 +1,314 @@
...
+  );
+
+  /// Finds a server directory, prompting the user if multiple are found.
+  ///
+  /// Throws [ServerpodProjectNotFoundException] if no server directory is found
+  /// or if multiple are found and [interactive] is false.
+  static Future<Directory> findOrPrompt({
+    Directory? startDir,
+    // Optional interactive property in case someone wanted to run these CLI commands in a CI/CD environment.
+    bool interactive = true,
```
SandPod, line 173: Suggestion/Question: I can't see this option being configured from anywhere, and currently it is on by default. How would the user use this?

SandPod (changes requested):
Nice clean PR! The code structure is easy to read and understand. Great work!

I have a couple of comments and suggestions to address I have left in the review, but then I would also like to to address three important things:
1. How does the user configure the interactive prompt?
2. Could you add an end to end test for this in the CLI e2e test directory: tests/serverpod_cli_e2e_test
3. Can we use the prompt tools from cli_tools instead of building a new one here?

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
FXschwartz: @SandPod We "shouldn't" be no. I added a `visited` set that is used to manage already searched children. It gets passed to all `_searchChildrenRecursive` calls and there is a check at line 228 that if the current path exist in the normalized set we do an early return.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
FXschwartz: @SandPod Yeah you are totally right, this doesn't feel right. I wen't ahead and removed the `_app` pattern check. And just to clarify did you want me to remove all of these pattern checks? Or just the `_app` one.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
FXschwartz: Done!

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
FXschwartz: Awkward, I had that hardcoded for testing and forgot to fix it later. Magic number has been replaced with an input parameter called `maxUpwardLevels` and it defaults to 5.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
FXschwartz: Yeah this is definitely a great place to optimize. I've added early returns to both the children check and the sibling check.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
FXschwartz: No reason, just missed that `cli_tools` package. I've updated it to use it.

tools/serverpod_cli/test/util/server_directory_finder_test.dart
FXschwartz: Makes a lot more sense. Updated!

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
FXschwartz: Yeah you are right, sorry I didn't fully think this one through. A couple questions then on how this implementation should ideally work.

Some CI environments like Github Actions set an environment variable called `CI` to true and we could check it like this `Platform.environment['CI'] == 'true'` and just skip the prompt.

Or is this an edge case worth supporting initially? Currently a command like `create-migration` ran in a CI environment and multiple server directories are found then the CI job will hang and fail due to the prompt.

tools/serverpod_cli/test/util/server_directory_finder_search_test.dart
FXschwartz: Sounds good. I refactored the .git boundary tests to better test it respects the boundary and I removed one of the boundary tests that are now redundant.

FXschwartz:
> Nice clean PR! The code structure is easy to read and understand. Great work!
> 
> I have a couple of comments and suggestions to address I have left in the review, but then I would also like to to address three important things:
> 
> 1. How does the user configure the interactive prompt?
> 2. Could you add an end to end test for this in the CLI e2e test directory: tests/serverpod_cli_e2e_test
> 3. Can we use the prompt tools from cli_tools instead of building a new one here?

For your second point is adding a e2e test for the generate cli command enough or do you also want tests for the create-migration? If we wanted one for create-migration I would have to add a lot more logic and code so I want to check before spending that amount of time on it.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
SandPod: I was only concerned with the `_app` check. We can preserve the others as shortcuts for known patterns.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
SandPod: I can't seem to find any tests that validate these early returns work as expected and in the given order that we have configured. Could you add a test that validates this?

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
SandPod: I would expect this to be sourced from ENV and as a possible input flag to the CLI.

We do have a dependency on https://pub.dev/packages/ci (that we don't seem to be using anymore?), perhaps we could use the check there so that we don't need to keep track of all possible CI environments?

If the command line option is supplied, the configuration there should overwrite whatever is deduced from the `ci` package.

In a CI environment, you really want to be able to disable these things. I would suggest that we introduce a flag that silences this. I'm not sure what to call it though, it would be good to have something that is reusable for all future interactive prompt disabling.

I'm not sure `--yes` is the right choice, since we don't want just to pick one, and I'm also unsure about `ci`, as it might be unambiguous regarding its purpose. Is`--<no->input` better? Or would you happen to have a good suggestion?

SandPod:
> For your second point is adding a e2e test for the generate cli command enough, or do you also want tests for the create-migration

I think it is enough to test it with generate. We are only interested in whether the location algorithm works as expected e2e. Possibly also the `ci` override through command lines.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
FXschwartz: Good call! I've added four tests to validate the potential outcomes of the new early return logic.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
FXschwartz: @SandPod What about something like `--non-interactive`?

FXschwartz:
> I think it is enough to test it with generate. We are only interested in whether the location algorithm works as expected e2e. Possibly also the ci override through command lines.

I've added three e2e tests to cover the usage of `generate` in multiple places and levels in a project. Once we nail down how we would like to handle the CI override then I will add another one for that.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
SandPod: That sounds good, but in that case, use `--<no->-interactive`. Using the built-in negatable option.

This allows someone to both do:
`serverpod generate --interactive`
`serverpod generate --no-interactive`

And this should default to true except if we notice that we are in a CI environment.

FXschwartz: @SandPod Alright I've update the CLI commands (`generate`, `create-migration`, and `create-repair-migration`) to have this optional negatable flag. 

The priority is `CLI flag > CI detection > default (true)` with logic `final isInteractive = interactive ?? !ci.isCI;`

I also added the following e2e tests to cover this interactive logic.

  - Tests --no-interactive flag works
  - Tests CI environment auto-detection (CI=true)
  - Tests flag override behavior (--interactive overrides CI)
  - Tests multiple servers with --no-interactive → error message
  - Tests multiple servers with --interactive → shows prompt

I had to make multiple file changes to update the `GeneratorConfig.load()` call sites from positional to named parameters with the addition of the optional `bool? interactive` parameter.

Take a look and let me know what you'd like changed or any kind of feedback. 

Appreciate your time reviewing this PR!

SandPod: Direction sounds good. I'll take a look at it during the weekend.

But could you make this a global flag instead, much like the verbose one?

I think the behaviour is that you don't care about what command you are running, but if there is an interactive component to it, it should never be executed.

FXschwartz: @SandPod Done! I've moved the interactive configuration to be global. 

I have noticed a couple of potentially flaky tests that seem to be intermittently failing locally for me. `endpoint_method_test.dart`, `endpoint_file_test.dart`, and `endpoint_inheritance_analysis_test.dart` are failing sometimes with the error `PathNotFoundException: Deletion failed`. Might be a race condition with the temp directory cleanup?

tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -238,7 +240,27 @@ class GeneratorConfig implements ModelLoadConfig {
...
   /// Create a new [GeneratorConfig] by loading the configuration in the [serverRootDir].
-  static Future<GeneratorConfig> load([String serverRootDir = '']) async {
+  ///
+  /// If [serverRootDir] is empty, the server directory will be automatically
+  /// detected by searching the current directory and nearby locations.
+  ///
+  /// The [interactive] parameter controls whether interactive prompts are enabled.
+  /// Defaults to true unless running in a CI environment (detected via ci package).
+  /// Explicit flag value overrides CI detection.
+  static Future<GeneratorConfig> load(
```
SandPod: Suggestion: This is also called from `create` and `lsp`.

For the `create` command, the interactive flag should also be respected.
For the `lsp` command `interactive` should always be false (at least for now).

You can find them by searching for all references of this method.

Perhaps `interactive` should be nullable required parameter? So we ensure it is always handled.

tools/serverpod_cli/lib/src/runner/serverpod_command_runner.dart
```diff
@@ -125,6 +125,15 @@ enum GlobalOption<V> implements OptionDefinition<V> {
   quiet(BetterCommandRunnerFlags.quietOption),
   verbose(BetterCommandRunnerFlags.verboseOption),
   analytics(BetterCommandRunnerFlags.analyticsOption),
+  interactive(
+    FlagOption(
+      argName: 'interactive',
+      negatable: true,
+      defaultsTo: true,
```
SandPod: Suggestion: Remove default value.

Having a default value will always set this to true, resulting in it being impossible for us to detect if the value was set or not. By removing the default value, it will be set to `null` by default.

Add a test to `serverpod_command_runner_test.dart` that validates this behaves as expected.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
SandPod: Hmm, could depend on how the tests are run. We will have to monitor the potentially flaky in our CI to see if they reproduce there. But most likely this is because we are not using the test descriptor package.

tools/serverpod_cli/test/util/server_directory_finder_test.dart
```diff
@@ -0,0 +1,569 @@
...
+        expect(result, isNotNull);
+        expect(
+          path.basename(result!.path),
+          equals('level_api'),
+        );
+      },
+    );
+  });
+
+  group('ServerDirectoryFinder.search hierarchical order and early returns',
```
SandPod: Suggestion(NIT): Break all "outer" groups out into different files.

We usually try to stay away from creating "outer" groups around tests in favour of breaking them out into separate files instead.

The benefit of this is that we don't get really long test files, which makes it easier to discover what we are validating. By inspecting the directory, it becomes clear what scenarios have been covered.

tools/serverpod_cli/lib/src/config/config.dart
FXschwartz: Nice! I missed those. Updated both to respect the interactive flag with lsp being set to false, and made `interactive` nullable required.

tools/serverpod_cli/lib/src/runner/serverpod_command_runner.dart
FXschwartz: Removed `defaultsTo` and changed `value()` to `optionalValue()` in all commands that get this flag value. Also added tests to validate the if `interactive`, or `non-interactive` is not set then it will be null.

tools/serverpod_cli/test/util/server_directory_finder_test.dart
FXschwartz: Nice I like that! Tests have been split up into individual files with a single group in each.

FXschwartz:
@SandPod I believe all requested changes have been addressed and I've merged main and resolved conflicts. Take another looks when you have the bandwidth and let me know what you think!

SandPod:
It all looks good to me now! Just fix the CI failures, and we are ready to merge 🚀 

Its been great collaborating with you on this @FXschwartz 👍

FXschwartz:
> It all looks good to me now! Just fix the CI failures, and we are ready to merge 🚀
> 
> Its been great collaborating with you on this @FXschwartz 👍

Ran the dart formatter, and fixed the two failing tests! I don't see the tests re running so hopefully all goes well.

Looks like the dart formatter formatted a ton of files which I didn't touch. I'm guessing that's from merging main into my branch that caused that, but let me know if you'd like me to undo those changes and go a different route.

Appreciate your review and time on this one! Definitely a great experience contributing to the repo and here's hoping to some more collaboration in the future 🍻.

SandPod:
@FXschwartz - Something seems to have gone wrong in your last commit as the formatter is still failing, what version of Dart are you running?

FXschwartz:
> @FXschwartz - Something seems to have gone wrong in your last commit as the formatter is still failing, what version of Dart are you running?

I'm on ```Dart SDK version: 3.9.2 (stable) (Wed Aug 27 03:49:40 2025 -0700) on "macos_arm64"```

Also looks like some endpoint validation tests are failing. I think those might be the same ones that are intermittently failing for me locally.

SandPod:
That should be enough. Have you merged with main recently? If not, could you merge main into your branch and then format again?

Drop you last commit so that you don't touch too many files in the project before the merge.

GitHub automatically merges main into your branch if there are no conflicts before running the workflows, so it could just be that you are based on some old main.

FXschwartz:
@SandPod Ok I reverted my dart format changes but kept the fixes to the two tests. I re synced my fork's main branch and then merged main into this branch.

FXschwartz:
Looks like something weird is going on. Merging main pulled in a bunch of other peoples commits and now shows all of their files changes included.

SandPod:
Hmm, how did you merge main into your branch? GitHub doesn't handle all types of merges equally. 

An alternative solution could be to remove the merge and rebase your branch on top of the latest main.

FXschwartz:
@SandPod Ok yeah sorry about this, I actually messed it up yesterday when I merged main into my branch instead of just rebasing which broke the dart formatter test. I've reverted, rebased and cleaned up the history. All tests are passing for me locally.

Should be good to start the workflow again.

SandPod:
@FXschwartz - No problem at all! I can't even count the number of times I have had to panic-restore my Git state based on the reflog... Let's hope all is green now!

FXschwartz:
Well the good news is the dart formatter is failing but only for the files I actually created/touched. One second and let me re run the formatter locally.

FXschwartz:
@SandPod Ok I've manually ran dart format on those files and confirmed running `dart format --output=none --set-exit-if-changed .` locally passes. Fingers crossed we are good to go!

SandPod (approved):
LGTM! :rocket:

tools/serverpod_cli/test/integration/downloads/localCache/latest_cli_version.yaml
```diff
@@ -0,0 +1,4 @@
+{
+  "version": "1.1.1",
+  "valid_until": 1764012255496
+}
```
SandPod: @FXschwartz - Was just about to merge this and noticed this file.

This seems like it was added but I'm not sure it should have been. Is this a residue from one of the tests that was accidentally committed?

FXschwartz: Yup! This was an accident during my dart format commit. Deleted!

PR #4158, opened 2025-11-05
feat: Resolve authentication on Session creation

exaby73:
## Pre-launch Checklist

## Breaking changes

`session.authenticated` is no longer a `Future`

### DIFF
--- a/packages/serverpod/lib/src/server/server.dart
+++ b/packages/serverpod/lib/src/server/server.dart
@@ -405,7 +405,7 @@ class Server implements RouterInjectable {
     try {
       var methodCallContext = await endpoints.getMethodCallContext(
         createSessionCallback: (connector) async {
-          maybeSession = await InternalSessionFactory.createMethodCallSession(
+          maybeSession = await SessionInternalMethods.createMethodCallSession(
             server: this,
             uri: uri,
             body: body,
--- a/packages/serverpod/lib/src/server/session.dart
+++ b/packages/serverpod/lib/src/server/session.dart
@@ -657,9 +657,10 @@ class MessageCentralAccess {
   }
 }
 
-/// Internal factory methods for creating Session instances.
-/// This class is hidden from the public API and should only be used within Serverpod.
-abstract interface class InternalSessionFactory {
+/// Internal methods for [Session].
+/// This is used to provide access to internal methods that should not be
+/// accessed from outside the library.
+extension SessionInternalMethods on Session {
   /// Creates a new [MethodCallSession].
   static Future<MethodCallSession> createMethodCallSession({
     required Server server,
@@ -751,12 +752,7 @@ abstract interface class InternalSessionFactory {
     await session.initializeAuthentication();
     return session;
   }
-}
 
-/// Internal methods for [Session].
-/// This is used to provide access to internal methods that should not be
-/// accessed from outside the library.
-extension SessionInternalMethods on Session {
   /// Returns the [LogManager] for the session.
   SessionLogManager? get logManager => _logManager;
 
--- a/packages/serverpod/lib/server.dart
+++ b/packages/serverpod/lib/server.dart
@@ -14,8 +14,7 @@ export 'src/generated/protocol.dart'
 export 'src/server/run_mode.dart';
 export 'src/server/server.dart' hide ServerInternalMethods;
 export 'src/server/serverpod.dart' hide ServerpodInternalMethods;
-export 'src/server/session.dart'
-    hide SessionInternalMethods, InternalSessionFactory;
+export 'src/server/session.dart' hide SessionInternalMethods;
 export 'src/config/security_context_config.dart' show SecurityContextConfig;
 
 export 'package:serverpod_shared/serverpod_shared.dart'
--- a/packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
+++ b/packages/serverpod/lib/src/server/websocket_request_handlers/endpoint_websocket_request_handler.dart
@@ -23,7 +23,7 @@ abstract class EndpointWebsocketRequestHandler {
     void Function() onClosed,
   ) async {
     try {
-      var session = await InternalSessionFactory.createStreamingSession(
+      var session = await SessionInternalMethods.createStreamingSession(
         server: server,
         uri: request.requestedUri,
         request: request,
--- a/packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
+++ b/packages/serverpod/lib/src/server/websocket_request_handlers/method_websocket_request_handler.dart
@@ -301,7 +301,7 @@ class MethodWebsocketRequestHandler {
       methodStreamCallContext =
           await server.endpoints.getMethodStreamCallContext(
         createSessionCallback: (connector) async {
-          maybeSession = await InternalSessionFactory.createMethodStreamSession(
+          maybeSession = await SessionInternalMethods.createMethodStreamSession(
             server: server,
             authenticationKey: unwrapAuthHeaderValue(message.authentication),
             endpoint: message.endpoint,
--- a/packages/serverpod/lib/src/web_server/web_server.dart
+++ b/packages/serverpod/lib/src/web_server/web_server.dart
@@ -170,7 +170,7 @@ class _SessionMiddleware extends MiddlewareObject {
   @override
   Handler call(Handler next) {
     return (req) async {
-      final session = await InternalSessionFactory.createWebCallSession(
+      final session = await SessionInternalMethods.createWebCallSession(
         server: _server,
         endpoint: req.requestedUri.path,
         authenticationKey: req.headers.authorization?.headerValue,


modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/endpoints/status_endpoint.dart
```diff
@@ -12,13 +12,13 @@ class StatusEndpoint extends Endpoint {
...
   Future<bool> isSignedIn(final Session session) async {
-    final authInfo = await session.authenticated;
+    final authInfo = session.authenticated;
     return authInfo?.authId != null;
   }
 
   /// Signs out a user from the current device.
   Future<void> signOutDevice(final Session session) async {
-    final authInfoIdStr = (await session.authenticated)?.authId;
+    final authInfoIdStr = (session.authenticated)?.authId;
```
SandPod: Chore: Remove the `()` around the call.

Check your diff and fix this in all places.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -383,6 +422,30 @@ class MethodStreamSession extends Session {
     required this.connectionId,
   })  : _method = method,
         super(method: method);
+
+  /// Creates a new [MethodStreamSession] with authentication initialized.
+  ///
+  /// This factory method creates a session and initializes authentication
+  /// before returning it.
+  static Future<MethodStreamSession> _create({
```
SandPod: Question: Why the separate `_create` method here and not only the internal extension?

I think this might be redundant since we could do the full flow in the internal extension.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -681,6 +761,94 @@ class MessageCentralAccess {
   }
 }
 
+/// Internal factory methods for creating Session instances.
+/// This class is hidden from the public API and should only be used within Serverpod.
+abstract interface class InternalSessionFactory {
```
SandPod: Suggestion: Add all new methods to the already existing `SessionInternalMethods`.

Should be OK to have everything there. Then we only need to exclude one export.

tests/serverpod_test_server/test_integration/session/session_authenticated_test.dart
```diff
@@ -0,0 +1,195 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod/src/server/session.dart';
+import 'package:test/test.dart';
+
+import '../test_tools/serverpod_test_tools.dart';
+
+void main() {
```
SandPod: Question: This is just validating the test framework, don't we already have tests that validate this?

packages/serverpod/lib/src/server/server.dart
```diff
@@ -62,10 +63,11 @@ class Server implements RouterInjectable {
...
   final SerializationManager serializationManager;
 
-  late AuthenticationHandler _authenticationHandler;
-
   /// [AuthenticationHandler] responsible for authenticating users.
-  AuthenticationHandler get authenticationHandler => _authenticationHandler;
+  // DISCUSSION: Made non-final to support testing scenarios where authentication
+  // behavior needs to be mocked or overridden. Consider if there's a better approach
+  // that maintains immutability while supporting test cases.
+  late AuthenticationHandler authenticationHandler;
```
SandPod: Question: Why was this modified? I can't seem to see us using this anywhere? Could the change be reverted?

packages/serverpod/lib/src/server/session.dart
```diff
@@ -706,6 +874,48 @@ extension SessionInternalMethods on Session {
...
+  /// by calling the server's authentication handler with the session's authentication key.
+  /// The authentication information is cached in the session, making subsequent accesses
+  /// to [Session.authenticated] synchronous and efficient.
+  ///
+  /// This method is idempotent - calling it multiple times will only initialize
+  /// authentication once. Subsequent calls will return immediately without performing
+  /// any work.
+  Future<void> initializeAuthentication() async {
+    var authKey = authenticationKey;
+    if (authKey != null) {
```
SandPod, line 788: Suggestion: Null out the `_authenticated` field here if authKey is null.

That keeps the logic for determining what the `_authenticated` field should be tightly connected.
Right now, if a user is signed in and authentication key is nulled out and initializeAuthentication is called, without being called through `updateAuthenticationKey`, the user would still be authenticated.

(Remove nulling out the field in `updateAuthenticationKey`)

tests/serverpod_test_server/test_integration/session_authentication_test.dart
```diff
@@ -0,0 +1,328 @@
...
+    const testUserId = 'test-user-123';
+    const testAuthId = 'auth-id-456';
+    final testScopes = {Scope('test'), Scope('admin')};
+
+    setUpAll(() async {
+      server = IntegrationTestServer.create(
+        authenticationHandler: (final session, final token) async {
+          // Web routes do not strip "Bearer " prefix automatically
+          // This is likely a bug. For now, we handle it here.
+          final t = token.replaceFirst('Bearer ', '');
```
SandPod, line 25: Suggestion: Is this a bug that we should fix? We shouldn't handle this in a test. The same token should work in the RPC as in the web server.

tests/serverpod_test_server/test_integration/session_authentication_test.dart
```diff
@@ -0,0 +1,328 @@
...
+      () {
+        late Client client;
+
+        setUp(() {
+          final authKeyManager = TestAuthKeyManager();
+          client = Client(
+            'http://localhost:8080/',
+            authenticationKeyManager: authKeyManager,
+          );
+          authKeyManager.put('valid-test-token');
```
SandPod: Suggestion: Break this out into a variable so that it can be referenced in all places in the test file.

This also makes it easy to assert we have an invalid token as we can just add something to the valid one.

tests/serverpod_test_server/test_integration/session_authentication_test.dart
```diff
@@ -0,0 +1,328 @@
...
+
+          expect(userId, equals(testUserId));
+        });
+
+        test('then getAuthenticatedScopes returns authenticated scopes',
+            () async {
+          final scopes =
+              await client.sessionAuthentication.getAuthenticatedScopes();
+
+          expect(scopes, containsAll(['test', 'admin']));
```
SandPod: Suggestion: Validate towards the testScopes variable.

tests/serverpod_test_server/test_integration/session_authentication_test.dart
```diff
@@ -0,0 +1,328 @@
...
+import 'package:http/http.dart' as http;
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_test_client/serverpod_test_client.dart';
+import 'package:serverpod_test_server/src/web/routes/session_test_route.dart';
+import 'package:serverpod_test_server/test_util/test_key_manager.dart';
+import 'package:serverpod_test_server/test_util/test_serverpod.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('Given a server with an authenticationHandler', () {
```
SandPod, line 11: Suggestion: Completement these tests with an invalid token use to cover all cases.

SandPod (changes requested):
Couple of comments.

Most importantly we should address if there is an inconsistency on how the token is handled between the RPC and the web server. There MUST be the same.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/endpoints/status_endpoint.dart
exaby73: Would it be okay for me to add a lint to automatically detect unecessary parens?

https://dart.dev/tools/linter-rules/unnecessary_parenthesis

packages/serverpod/lib/src/server/session.dart
exaby73: Done in https://github.com/exaby73/serverpod/commit/3e13aef7853e16b99b8770ea96ce09fcbba6b7e7

packages/serverpod/lib/src/server/session.dart
exaby73: Done in https://github.com/exaby73/serverpod/commit/36708280f6a872b0ea080278bde57ba626555093

tests/serverpod_test_server/test_integration/session/session_authenticated_test.dart
exaby73: We don't have any tests testing the `authenticated` property in isolation, always with the old auth module. And these tests also expect `authentication` to be a non `Future` as well. These tests aren't super important but they might be worth keeping for completions sake. Let me know if you disagree as I don't feel too strongly about keeping them and I'm okay with removing them altogether

packages/serverpod/lib/src/server/server.dart
exaby73: Slipped through the cracks from my previous implementation. Reverted in https://github.com/exaby73/serverpod/commit/ac051025bd61237316e9d408f5597cbfb9c3bbac

packages/serverpod/lib/src/server/session.dart
exaby73: Fixed in https://github.com/exaby73/serverpod/commit/786cb301b7018c883b49c2c4c2fb2d454a4f0106

tests/serverpod_test_server/test_integration/session_authentication_test.dart
exaby73: Fixed in https://github.com/exaby73/serverpod/commit/dd9212423b7f3427f07b47ed37b2aebdd8853469

tests/serverpod_test_server/test_integration/session_authentication_test.dart
exaby73: Fixed in https://github.com/exaby73/serverpod/commit/f43784209a6d19c2787ac412b6d59ba89fb9e29f

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/endpoints/status_endpoint.dart
SandPod: Sure! As long as it doesn't explode the repo :)

tests/serverpod_test_server/test_integration/session/session_authenticated_test.dart
SandPod: Aren't these basically the same tests: tests/serverpod_test_server/test_integration/test_tools/authentication_validation_test.dart

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/endpoints/status_endpoint.dart
exaby73: If it does, I'll add it temporarily, and remove it. But might be worth adding it in the future to `serverpod_lints` possibly as sometimes it can be hard to spot useless parens :D

tests/serverpod_test_server/test_integration/session/session_authenticated_test.dart
exaby73: Ahh you're right! I missed these. Will remove these tests then :)

exaby73: Removed in https://github.com/exaby73/serverpod/commit/ed45201b4d36d944796ee0236767dd7481e20b38

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/endpoints/status_endpoint.dart
exaby73: Fixed in https://github.com/exaby73/serverpod/commit/1b05c8bb59f68abe41c436ec2f85d93276a23343

tests/serverpod_test_server/test_integration/session_authentication_test.dart
exaby73: Added a comment tracking it in https://github.com/exaby73/serverpod/commit/fbf96a956348955bd2a44506f759ab8673097dea

tests/serverpod_test_server/test_integration/session_authentication_test.dart
exaby73: Added tests in https://github.com/exaby73/serverpod/commit/fbf96a956348955bd2a44506f759ab8673097dea

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/analysis_options.yaml
```diff
@@ -3,4 +3,8 @@ include: package:serverpod_lints/public.yaml
 analyzer:
   exclude:
     - lib/src/generated/**
-    - test/serverpod_test_tools.dart
\ No newline at end of file
+    - test/serverpod_test_tools.dart
+
+linter:
```
SandPod, line 8: Question: Should this rule be applied more widely? For example in the Serverpod lints package?

packages/serverpod/lib/src/server/session.dart
SandPod: This one still remains: `InternalSessionFactory` Please remove it as well together with the export.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/analysis_options.yaml
SandPod: You already suggested this and that could be a good idea :)

exaby73: I could create an issue for this. `serverpod_auth_core_server` doesn't actually use `serverpod_lints`. Maybe we should change that as well?

packages/serverpod/lib/src/server/session.dart
exaby73: Removed in https://github.com/exaby73/serverpod/commit/91099d56fa0e3cdae14bc27a0f5f9298a6fec561

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/analysis_options.yaml
SandPod: Sounds good! 👍

PR #4159, opened 2025-11-05
refactor!: Split complete password reset step in two for Email IDP.

SandPod:
This PR refactors the password reset process by splitting the final step into two separate actions:
- Validation of the verification code
- Setting the new password

This separation enables independent validation for each stage and introduces flexibility for future MFA integration.

## New flow
Screenshot 2025-11-05 at 10 40 58

Once the validation code has been verified, a reset token code is generated that can be submitted together with the `password`. The separation allows the possibility of injecting user-enforced MFA steps.

The validation code is only returned once from the verify method, meaning the user will need to store it if additional steps are added, and then submit it once all their validation have passed.

## Validation change

I moved the expired check to happen before we validate the security code and check if the request has been verified. That was to ensure that we remove expired requests. The user still needs to have the unique ID of the request in order to get the `expired` message, which is a UUID and should be almost impossible to brute force.

Closes: #4067 

## Pre-launch Checklist

## Breaking changes

Yes, this changes the flow of a password reset. But the changes are related to a non released package.

## Release Notes

* **New Features**
  * Added password reset code verification endpoint to the email authentication flow.
  * Implemented two-step password reset process: verify reset code before completing password change.
  
* **Bug Fixes**
  * Enhanced error handling with new exception types for password reset scenarios.

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/utils/email_idp_password_reset_util.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/utils/email_idp_password_reset_util.dart
@@ -43,7 +43,7 @@ class EmailIDPPasswordResetUtil {
   /// Can throw the following [EmailPasswordResetServerException] subclasses:
   /// - [EmailPasswordResetRequestNotFoundException] if no reset request could
   ///   be found for [passwordResetRequestId].
-  /// - [EmailPasswordResetSetPasswordTokenNotFoundException] if the set
+  /// - [EmailPasswordResetNotVerifiedException] if the set
   ///   password token has not been set.
   /// - [EmailPasswordResetRequestExpiredException] if the reset request has
   ///   expired and has not been cleaned up yet.
@@ -82,19 +82,9 @@ class EmailIDPPasswordResetUtil {
       throw EmailPasswordResetRequestNotFoundException();
     }
 
-    if (resetRequest.isExpired(_config.passwordResetVerificationCodeLifetime)) {
-      await EmailAccountPasswordResetRequest.db.deleteRow(
-        session,
-        resetRequest,
-        // passing no transaction, so this will not be rolled back
-      );
-
-      throw EmailPasswordResetRequestExpiredException();
-    }
-
     final setPasswordChallenge = resetRequest.setPasswordChallenge;
     if (setPasswordChallenge == null) {
-      throw EmailPasswordResetSetPasswordTokenNotFoundException();
+      throw EmailPasswordResetNotVerifiedException();
     }
 
     if (!await _passwordHashUtil.validateHash(
@@ -103,6 +93,16 @@ class EmailIDPPasswordResetUtil {
       salt: setPasswordChallenge.challengeCodeSalt.asUint8List,
     )) {
       throw EmailPasswordResetInvalidVerificationCodeException();
+    }
+
+    if (resetRequest.isExpired(_config.passwordResetVerificationCodeLifetime)) {
+      await EmailAccountPasswordResetRequest.db.deleteRow(
+        session,
+        resetRequest,
+        // passing no transaction, so this will not be rolled back
+      );
+
+      throw EmailPasswordResetRequestExpiredException();
     }
 
     await EmailAccountPasswordResetRequest.db.deleteRow(
@@ -190,16 +190,6 @@ class EmailIDPPasswordResetUtil {
       throw EmailPasswordResetTooManyVerificationAttemptsException();
     }
 
-    if (resetRequest.isExpired(_config.passwordResetVerificationCodeLifetime)) {
-      await EmailAccountPasswordResetRequest.db.deleteRow(
-        session,
-        resetRequest,
-        // passing no transaction, so this will not be rolled back
-      );
-
-      throw EmailPasswordResetRequestExpiredException();
-    }
-
     if (resetRequest.isVerificationCodeUsed) {
       throw EmailPasswordResetVerificationCodeAlreadyUsedException();
     }
@@ -212,6 +202,16 @@ class EmailIDPPasswordResetUtil {
       salt: challenge.challengeCodeSalt.asUint8List,
     )) {
       throw EmailPasswordResetInvalidVerificationCodeException();
+    }
+
+    if (resetRequest.isExpired(_config.passwordResetVerificationCodeLifetime)) {
+      await EmailAccountPasswordResetRequest.db.deleteRow(
+        session,
+        resetRequest,
+        // passing no transaction, so this will not be rolled back
+      );
+
+      throw EmailPasswordResetRequestExpiredException();
     }
 
     final setPasswordToken = const Uuid().v4();
--- a/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/email/email_idp/email_idp_verify_password_reset_code_test.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/email/email_idp/email_idp_verify_password_reset_code_test.dart
@@ -184,7 +184,7 @@ void main() {
       });
 
       test(
-          'when verifyPasswordResetCode is called with invalid credentials after expiration then it throws EmailAccountPasswordResetException with reason "expired"',
+          'when verifyPasswordResetCode is called with invalid credentials after expiration then it throws EmailAccountPasswordResetException with reason "invalid" to not leak that the request exists',
           () async {
         await withClock(
             Clock.fixed(DateTime.now().add(
@@ -202,7 +202,7 @@ void main() {
               isA<EmailAccountPasswordResetException>().having(
                 (final e) => e.reason,
                 'reason',
-                EmailAccountPasswordResetExceptionReason.expired,
+                EmailAccountPasswordResetExceptionReason.invalid,
               ),
             ),
           );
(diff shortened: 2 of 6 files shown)


SandPod:
@coderabbitai review

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/utils/email_idp_password_reset_util.dart
```diff
@@ -133,6 +127,114 @@ class EmailIDPPasswordResetUtil {
...
+      resetRequest.copyWith(
+        setPasswordChallengeId: setPasswordChallenge.id!,
+      ),
+      transaction: transaction,
+    );
+
+    return (
+      passwordResetRequestId: resetRequest.id!,
+      verificationCode: setPasswordToken,
+    );
```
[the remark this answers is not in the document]
SandPod: This is fixed in the last commit with test validating the issue.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/utils/email_idp_password_reset_util.dart
```diff
@@ -444,4 +578,12 @@ extension on EmailAccountPasswordResetRequest {
...
     return challenge!;
   }
+
+  bool get isVerificationCodeUsed => setPasswordChallenge != null;
 }
+
+/// The credentials for setting the password for the password reset request.
+typedef SetPasswordCredentials = ({
+  UuidValue passwordResetRequestId,
+  String verificationCode,
```
marcelomendoncasoares, line 627: Discussion: While reviewing, I was a bit confused with having two methods that accept verification codes, with one being the final step and the other an intermediary. Then I understood that the verification method is issuing a new verification code to be used on the complete method.

From an external user perspective, when inspecting the signature of the methods, it might generate a question of why should they call the verify method since the complete already receives the verification code. To avoid this confusion, we could rename this `verificationCode` here (and on the `completePasswordReset` method) to `verificationToken` (or even `validatedVerificationToken` to enforce the need of previously validating). It even enforces that this token comes from other method, since we never expect a user to enter a token manually, but we do expect them to enter verification codes. Does it make sense?

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -606,8 +771,7 @@ void main() {
 
       test(
           'when calling `finishPasswordReset` with the wrong verification code after the password reset request has expired, '
-          'then it fails with reason `invalid` to not leak that the request exists.',
-          () async {
+          'then it fails with reason `expired`.', () async {
         await expectLater(
           () => withClock(
             Clock.fixed(DateTime.now().add(verificationCodeLifetime)),
```
marcelomendoncasoares, line 771: We can't return expired when the code is wrong, otherwise we leak if the account exists. An attacker could wait enough time to know that the request should have expired if it is valid, and then try to complete with any verification code. To avoid it, wrong code must have the same return as non-existing request ID in all cases.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/email/email_idp/email_idp_verify_password_reset_code_test.dart
```diff
@@ -0,0 +1,669 @@
...
+            throwsA(
+              isA<EmailAccountPasswordResetException>().having(
+                (final e) => e.reason,
+                'reason',
+                EmailAccountPasswordResetExceptionReason.expired,
+              ),
+            ),
+          );
+        });
+      });
```
marcelomendoncasoares, line 210: Similar to the previous comment, making this reveal expired create an asymmetry on the response for existing and non-existing requests. It is important to inform expired when the verification code is right so the UI can be helpful for real users, but otherwise should just show invalid.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/email/email_idp/email_idp_finish_password_reset_test.dart
```diff
@@ -192,12 +209,12 @@ void main() {
     });
 
     test(
-        'when finishPasswordReset is called with invalid verification code then it throws EmailAccountPasswordResetException with reason "invalid"',
+        'when finishPasswordReset is called with invalid verification code then it throws EmailAccountPasswordResetException with reason "expired"',
```
marcelomendoncasoares: Same comment as before also reflects on this test.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
```diff
@@ -527,14 +527,179 @@ void main() {
...
+              verificationCode: 'wrong',
+            ),
+          ),
+          throwsA(isA<EmailAccountPasswordResetException>().having(
+            (final exception) => exception.reason,
+            'Reason',
+            EmailAccountPasswordResetExceptionReason.expired,
+          )),
+        );
+      });
```
marcelomendoncasoares, line 622: Same comment as before also reflects on this test.

marcelomendoncasoares (changes requested):
Just two comments regarding:
- Name `verificationCode` for the set password token parameter.
- Possible leakage of existing accounts when throwing expired for wrong verification code.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
SandPod: Good catch, you are correct. I will update the code to reflect this.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/utils/email_idp_password_reset_util.dart
SandPod: Yes, this cause a bit of confusion for me as well. Perhaps the passwordResetToken should actually contain the request ID as well? So that is the only input to that method?

SandPod: Fixed here: [8171fa0](https://github.com/serverpod/serverpod/pull/4159/commits/8171fa09ca8d490fbe83e1a89a4ec77d0c7a308f)

We should really adopt this pattern for the create account flow as well... The user might want to introduce more steps before the account is created, and currently we have two validation calls but we don't create a unique token to complete the account creation.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/email_account_endpoint_test.dart
SandPod: Fixed: [37c4c00](https://github.com/serverpod/serverpod/pull/4159/commits/37c4c0023a820fa7f1527a97be1103edfbeace61)

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/email/email_idp/email_idp_finish_password_reset_test.dart
SandPod: Fixed: [37c4c00](https://github.com/serverpod/serverpod/pull/4159/commits/37c4c0023a820fa7f1527a97be1103edfbeace61)

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/email/email_idp/email_idp_verify_password_reset_code_test.dart
SandPod: Fixed: [37c4c00](https://github.com/serverpod/serverpod/pull/4159/commits/37c4c0023a820fa7f1527a97be1103edfbeace61)

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/utils/email_idp_password_reset_util.dart
marcelomendoncasoares: It is indeed much better with the unified token approach! Is it worth also changing on the start reset password in this PR, since it is also a breaking change on the endpoint signatures?

SandPod: If it is OK with you, I would much rather do that in a separate PR since this PR is currently blocking progress on other related changes. I can file an issue for this and start working on it straight away.

marcelomendoncasoares: Agreed!

SandPod: @coderabbitai Please create en issue for using a unified token in the create account process as discussed in this thread.

PR #4175, opened 2025-11-10
refactor!: Use Headers class instead of Map<String, String> when passing headers

### LINKED ISSUE #4102: Refactor httpResponseHeaders to use Headers class from relic package
## Description
The `httpResponseHeaders` should use the `Headers` class from the `relic` package instead of the current Map implementation for better type safety and API consistency.

## Affected Files
- `packages/serverpod/lib/src/server/server.dart:180`

## Context
```dart
// TODO: Make httpResponseHeaders a Headers object from the get-go.
// or better yet, use middleware
final headers = Headers.build((mh) {
  for (var rh in httpResponseHeaders.entries) {
    mh[rh.key] = ['${rh.value}'];
  }
});
```

## Suggested Implementation
- Change `httpResponseHeaders` from `Map<String, String>` to use the `Headers` class from the `relic` package directly
- Refactor code to use the Headers type throughout the codebase
- Update related APIs to work with the new type
- Consider using middleware as suggested in the comment
- Ensure backward compatibility or provide migration guide
- Add tests for header manipulation

## Impact
Type safety, API consistency, better developer experience

## Related
- PR #4085

nielsenko:
## Description

Refactors HTTP header handling to use relic's `Headers` type instead of `Map<String, dynamic>`.

### Changes:

- Updated `httpResponseHeaders` and `httpOptionsResponseHeaders` fields in `Server` and `Serverpod` classes from `Map<String, dynamic>` to `Headers`
- Simplified `_headers()` handler to work directly with `Headers` instances using `Headers.transform()`
- Converted default header constants to use `Headers.build()`

Fixes #4102 

## Pre-launch Checklist

## Breaking changes

Users who directly instantiate `Serverpod` or `Server` with custom headers need to use `Headers.build()` instead of Map literals. Users relying on default headers are not affected.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced type safety for HTTP header configuration with improved internal header management.

### DIFF
--- a/packages/serverpod/lib/src/server/serverpod.dart
+++ b/packages/serverpod/lib/src/server/serverpod.dart
@@ -323,21 +323,29 @@ class Serverpod {
 
   /// HTTP headers used by all API responses. Defaults to allowing any
   /// cross origin resource sharing (CORS).
-  final Map<String, dynamic> httpResponseHeaders;
+  final Headers httpResponseHeaders;
 
   /// HTTP headers used for OPTIONS responses. These headers are sent in
   /// addition to the [httpResponseHeaders] when the request method is OPTIONS.
-  final Map<String, dynamic> httpOptionsResponseHeaders;
+  final Headers httpOptionsResponseHeaders;
 
-  static const _defaultHttpResponseHeaders = {
-    'Access-Control-Allow-Origin': '*',
-    'Access-Control-Allow-Methods': 'POST',
-  };
+  static final _defaultHttpResponseHeaders = Headers.build((mh) {
+    mh.accessControlAllowOrigin =
+        const AccessControlAllowOriginHeader.wildcard();
+    mh.accessControlAllowMethods =
+        AccessControlAllowMethodsHeader.methods(methods: [Method.post]);
+  });
 
-  static const _defaultHttpOptionsResponseHeaders = {
-    'Access-Control-Allow-Headers':
-        'Content-Type, Authorization, Accept, User-Agent, X-Requested-With',
-  };
+  static final _defaultHttpOptionsResponseHeaders = Headers.build((mh) {
+    mh.accessControlAllowHeaders =
+        AccessControlAllowHeadersHeader.headers(headers: [
+      'Content-Type',
+      'Authorization',
+      'Accept',
+      'User-Agent',
+      'X-Requested-With',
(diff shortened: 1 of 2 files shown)


nielsenko:
@coderabbitai review

nielsenko:
To be honest I think setting default headers, etc. belong in the _configure middleware_ realm. However we currently only support middleware on `WebServer`. This may change in the near future with #4126

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -323,21 +323,29 @@ class Serverpod {
...
-        'Content-Type, Authorization, Accept, User-Agent, X-Requested-With',
-  };
+  static final _defaultHttpOptionsResponseHeaders = Headers.build((mh) {
+    mh.accessControlAllowHeaders =
+        AccessControlAllowHeadersHeader.headers(headers: [
+      'Content-Type',
+      'Authorization',
+      'Accept',
+      'User-Agent',
+      'X-Requested-With',
```
exaby73, line 346: It might be overkill but do you think these could be typed too? Maybe something like:

```suggestion
      AccessControlAllowHeadersHeaderValue,contentType,
      AccessControlAllowHeadersHeaderValue.authorization,
      AccessControlAllowHeadersHeaderValue.accept,
      AccessControlAllowHeadersHeaderValue.userAgent,
      AccessControlAllowHeadersHeaderValue.custom('X-Requested-With'),
```

exaby73 (commented):
Small discussion point but I think it might be overkill

packages/serverpod/lib/src/server/serverpod.dart
nielsenko: I think this is valid idea, it would be something like:
```suggestion
      Headers.contentType,
      Headers.authorization,
      Headers.accept,
      Headers.userAgent,
      Headers.xRequestedWith,
```
However, perhaps we can defer that to a later PR. 

Also, it does require all headers to be defined as typed headers. Perhaps we should still allow just setting some by string here.

exaby73: I agree. And we should definitely have a way of setting custom headers and possibly typing them, similar to the way you made typed path params

PR #4195, opened 2025-11-12
fix!: Deprecate old streaming endpoints and chat module.

SandPod:
Deprecates the streaming interface and removes the chat module from publishable packages.

The chat module needs to be reworked with the new authentication and streaming methods in mind.

It currently has around 600 downloads on pub.dev, so it doesn't seem to have that heavy usage at the moment, which should make it fine to not publish for a while.

### Deprecation and overrides
Unfortunately, deprecation on methods that are overridden is not propagated to the override site: https://github.com/dart-lang/sdk/issues/42991?utm_source=chatgpt.com

## Pre-launch Checklist

## Breaking changes


PUBLISHABLE_PACKAGES
```diff
@@ -15,9 +15,6 @@ modules/serverpod_auth/serverpod_auth_apple_flutter
 modules/serverpod_auth/serverpod_auth_google_flutter
 modules/serverpod_auth/serverpod_auth_email_flutter
 modules/serverpod_auth/serverpod_auth_firebase_flutter
-modules/serverpod_chat/serverpod_chat_server
-modules/serverpod_chat/serverpod_chat_client
-modules/serverpod_chat/serverpod_chat_flutter
```
marcelomendoncasoares, line 20: Discussion: Do we really need to remove the `chat` package from publishing for this deprecation? By marking the old streaming deprecated, we only need to update the package to use the new streaming before actually removing the old on the next major version after serverpod 3.0. We could leave it there to avoid loosing feature for eventual users of this package - even thought there are only a few -, and still keep the possibility of removing the old streaming soon.

SandPod: The reason we are holding of from deploying it is that it has a hard dependency on the old Auth module.

PR #4197, opened 2025-11-12
chore: Bump dart SDK version to 3.8.0

SandPod:
Bumps to minimum Dart version 3.8.0.

### Why 3.8.0 and not 3.7.0

3.8.0 allowed us to use the formatting rule preserve trailing commas, giving us some form of control over the formatting style.

### Code changes

In order to support the new formatting style, the code generator was updated to use Dart 3.8.0 with preserved trailing commas.

### analyzer_options.yaml updates

The rule to preserve trailing commas was added to all analyzer options in the project.
In the future, we should only depend on the serverpod lints in the project and adhere to the rules defined there.

Fixing this was out of scope for this bump.

### CI changes

The formatter now depends on all dependencies being pulled in order to give consistent results. Therefore we needed to bootstrap the project before formatting.

Closes: #3937 

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/tools/serverpod_cli/test/generator/dart/client_code_generator/unauthenticated_annotation_test.dart
+++ b/tools/serverpod_cli/test/generator/dart/client_code_generator/unauthenticated_annotation_test.dart
@@ -57,10 +57,10 @@ void main() {
             clientCode,
             contains('''\
   Future<String> hello() => caller.callServerEndpoint<String>(
-        'example',
-        'hello',
-        {},
-      );
+    'example',
+    'hello',
+    {},
+  );
 '''),
           );
         },
@@ -139,11 +139,11 @@ void main() {
             clientCode,
             contains('''\
   Future<String> hello() => caller.callServerEndpoint<String>(
-        'example',
-        'hello',
-        {},
-        authenticated: false,
-      );
+    'example',
+    'hello',
+    {},
+    authenticated: false,
+  );
 '''),
           );
         },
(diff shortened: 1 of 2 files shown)


templates/pubspecs/tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/pubspec.yaml
```diff
@@ -5,8 +5,8 @@ description: Auth module client-side tests.
 version: 0.0.1
 
 environment:
-  sdk: '>=3.5.0 <4.0.0'
-  flutter: '>=3.24.0'
+  sdk: DART_VERSION
+  flutter: FLUTTER_VERSION
```
nielsenko, line 9: 👀

.github/workflows/dart-tests.yaml
```diff
@@ -24,12 +24,17 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
-      - uses: dart-lang/setup-dart@v1.3
+      - uses: subosito/flutter-action@v2
         with:
-          sdk: ${{ matrix.dart-version }}
+          flutter-version: ${{ matrix.flutter-version }}
+          cache: true
+      - name: Run pub get
+        run: dart pub get
```
nielsenko, line 35: Do we need this? Doesn't `melos bootstrap` do the trick?

SandPod: Doesn't seem like Melos bootstrap included the root directory for some reason. I ended up with formatting issues when not included.

Issue #4177, opened 2025-11-10
Investigate whether streams from JWT tokens are properly closed when a frequently-refreshed refresh token is revoked


SandPod:
### Describe the bug

Investigate if streams initiated from JWT tokens are properly closed when a refresh token, which has been refreshed multiple times, is revoked. This may result in resource leaks or lingering sessions if streams are not shut down when the associated tokens are invalidated, especially after multiple refreshes.

### To reproduce

1. Generate a refresh token and use it to obtain new JWT tokens multiple times.
2. Revoke the refresh token after several uses.
3. Observe whether any streams associated with those JWT tokens remain open after revocation.
4. Monitor for resource leaks, dangling connections, or unexpected session continuance.

### Expected behavior

All streams associated with JWT tokens issued via a refresh token should be closed immediately when the refresh token is revoked, regardless of how many times it has been refreshed.

### Library version

N/A

### Platform information

N/A

### Additional context

This investigation aims to ensure the robustness of session and stream management in scenarios where refresh tokens are used repeatedly before eventual revocation.

### Are you interested in working on a PR for this?

- [ ] I want to work on this

PR #4198, opened 2025-11-12
fix: Publish authentication revoked on refresh token rotate expiry and invalid secret failures

exaby73:
Closes https://github.com/serverpod/serverpod/issues/4177

## Pre-launch Checklist

### DIFF
--- a/tests/serverpod_new_auth_test/serverpod_new_auth_test_client/lib/src/protocol/client.dart
+++ b/tests/serverpod_new_auth_test/serverpod_new_auth_test_client/lib/src/protocol/client.dart
@@ -72,11 +72,11 @@ class EndpointAuthTest extends _i1.EndpointRef {
       );
 
   /// Destroys a specific refresh token by ID.
-  _i2.Future<bool> destroySpecificRefreshToken() =>
+  _i2.Future<bool> destroySpecificRefreshToken(String token) =>
       caller.callServerEndpoint<bool>(
         'authTest',
         'destroySpecificRefreshToken',
-        {},
+        {'token': token},
       );
 
   /// Checks if the session is authenticated for the test user.
--- a/tests/serverpod_new_auth_test/serverpod_new_auth_test_server/lib/src/generated/endpoints.dart
+++ b/tests/serverpod_new_auth_test/serverpod_new_auth_test_server/lib/src/generated/endpoints.dart
@@ -184,13 +184,22 @@ class Endpoints extends _i1.EndpointDispatch {
         ),
         'destroySpecificRefreshToken': _i1.MethodConnector(
           name: 'destroySpecificRefreshToken',
-          params: {},
+          params: {
+            'token': _i1.ParameterDescription(
+              name: 'token',
+              type: _i1.getType<String>(),
+              nullable: false,
+            ),
+          },
           call:
               (
                 _i1.Session session,
                 Map<String, dynamic> params,
               ) async => (endpoints['authTest'] as _i2.AuthTestEndpoint)
-                  .destroySpecificRefreshToken(session),
+                  .destroySpecificRefreshToken(
+                    session,
+                    params['token'],
+                  ),
         ),
         'checkSession': _i1.MethodConnector(
           name: 'checkSession',
--- a/tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/test_tools/serverpod_test_tools.dart
+++ b/tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/test_tools/serverpod_test_tools.dart
@@ -362,6 +362,7 @@ class _AuthTestEndpoint {
 
   _i3.Future<bool> destroySpecificRefreshToken(
     _i1.TestSessionBuilder sessionBuilder,
+    String token,
   ) async {
     return _i1.callAwaitableFunctionAndHandleExceptions(() async {
       var _localUniqueSession =
@@ -374,7 +375,7 @@ class _AuthTestEndpoint {
           createSessionCallback: (_) => _localUniqueSession,
           endpointPath: 'authTest',
           methodName: 'destroySpecificRefreshToken',
-          parameters: _i1.testObjectToJson({}),
+          parameters: _i1.testObjectToJson({'token': token}),
           serializationManager: _serializationManager,
         );
         var _localReturnValue =


modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/models/refresh_token_expired_exception.spy.yaml
```diff
@@ -1 +1,5 @@
 exception: RefreshTokenExpiredException
+
+fields:
```
SandPod: Question: should we leak this to the client? This means that any client holding on to an old refresh token could us that to get the auth user id and the refresh token id associated with it?

Any concerns with this?

We could also have a server side exception that returns these things that we handle and then convert to a client side exception that we return from the endpoint.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/authentication_tokens.dart
```diff
@@ -21,7 +21,7 @@ final class AuthenticationTokens {
   final AuthenticationTokenConfig config;
 
   /// Admin-related functions for managing authentication tokens.
-  final AuthenticationTokensAdmin admin;
+  late final AuthenticationTokensAdmin admin;
```
SandPod: Suggestion: Remove `late` by using a private constructor and adding a public `factory` to the now defined constructor.

This was a pattern that @nielsenko encouraged me to use that looks quite nice.
You can see an example here: modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_idp.dart

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/jwt/integration/authentication_tokens_test.dart
```diff
@@ -321,75 +234,130 @@ void main() {
...
+          expect(
+            authTokensForUser.single.extraClaims,
+            {'string': 'foo', 'int': 1},
           );
         },
       );
 
       test(
-        'when trying to rotate the token with a wrong fixed secret, then it throws a "not found" error.',
+        'when calling rotateRefreshToken with an invalid secret, then authenticationRevoked message is published with correct authId',
```
SandPod: Suggestion: Split this test in two, one that expects the throw to be the right thing and one that expect the revoked message to be sent.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/jwt/integration/authentication_tokens_test.dart
```diff
@@ -847,6 +815,81 @@ void main() {
...
+          );
+
+          tokenId = authenticationTokens.jwtUtil
+              .verifyJwt(authSuccess.token)
+              .refreshTokenId;
+        },
+      );
+    });
+
+    test(
```
SandPod, line 712: Suggestion: Same suggestion as above, please split this test in two.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/authentication_tokens.dart
exaby73: Fixed in 9ccc2c2e3329

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/jwt/integration/authentication_tokens_test.dart
exaby73: Fixed in 63006f1e7052

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/jwt/integration/authentication_tokens_test.dart
exaby73: Fixed in 63006f1e7052

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/models/refresh_token_expired_exception.spy.yaml
exaby73: Doesn't look like we allow `scope=severOnly` on exceptions

```terminal
Found 2 issues.
Error on line 4, column 26 of lib/src/jwt/models/refresh_token_expired_exception.spy.yaml: The "scope" property is not allowed for authUserId type. Valid keys are {type, required, default, defaultModel}.
  ╷
1 │ scope
  │ ^^^^^
  ╵
Error on line 5, column 30 of lib/src/jwt/models/refresh_token_expired_exception.spy.yaml: The "scope" property is not allowed for refreshTokenId type. Valid keys are {type, required, default, defaultModel}.
  ╷
1 │ scope
  │ ^^^^^
  ╵
  ```

Comment (SandPod) [reply modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/models/refresh_token_expired_exception.spy.yaml]: I think the suggestion is to not have it serializable at all, and then convert it to a serializable exception when we want to sent it to the client.

Check out how we do it in modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_idp_server_exceptions.dart

These are then converted to a client exception in the path that leads to the endpoint.

Comment (exaby73) [reply modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/models/refresh_token_expired_exception.spy.yaml]: Fixed in 90be02b16faa

File: modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/jwt_util.dart [hunk 5]
```diff
@@ -154,6 +155,27 @@ class JwtUtil {
     return _extractExpirationDate(payload);
   }
 
+  /// Replaces server-side exceptions by client-side exceptions, hiding details
+  /// that could leak account information.
+  static Future<T> withReplacedServerJwtException<T>(
```
SandPod: Suggestion: Move this to `authentication_tokens.dart`.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/authentication_tokens.dart
```diff
@@ -186,117 +208,67 @@ final class AuthenticationTokens {
...
-    if (!await refreshTokenSecretHash.validateHash(
-      secret: refreshTokenData.rotatingSecret,
-      hash: Uint8List.sublistView(refreshTokenRow.rotatingSecretHash),
-      salt: Uint8List.sublistView(refreshTokenRow.rotatingSecretSalt),
-    )) {
-      await RefreshToken.db.deleteRow(
-        session,
-        refreshTokenRow,
-        transaction: transaction,
+    } on RefreshTokenExpiredServerException catch (e) {
```
SandPod: Question: Shouldn't this use the substitution?

And remember to update the description.

SandPod (commented):
Thanks for fixing my comments. Still have some last suggestion.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/jwt_util.dart
exaby73: Fixed in f7c27c56dab4

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/authentication_tokens.dart
exaby73: `rotateRefreshToken` is called by `refreshAccessToken` which is in turn called in the endpoint. So `rotateRefreshToken` should throw the server error, which is caught and substituted in `refreshAccessToken` before rethrowing the client exceptions for the endpoint

SandPod: In that case, would it make sense to move this into the admin class since it throws a server side exception? That might also help clear out the confugsion around what the difference between this and the refreshAccessToken (After our call I now see that this is a bad name...) is?

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/lib/src/endpoints/authenticated_streaming_test_endpoint.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:serverpod/serverpod.dart';
+
+class AuthenticatedStreamingTestEndpoint extends Endpoint {
+  @override
+  bool get requireLogin => true;
+
+  Stream<int> openAuthenticatedStream(final Session session) async* {
```
SandPod, line 7: Suggestion: Echo an input stream parameter to the output stream.

That way, we get full control from the client when we want to close the stream or if we would like to test to see if the communication with the server is still open.
Check out `tests/serverpod_test_server/lib/src/endpoints/method_streaming.dart` on line 93.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/jwt_refresh_streaming_endpoint_test.dart
```diff
@@ -0,0 +1,150 @@
...
+
+void main() {
+  setUp(() {
+    AuthServices.set(
+      primaryTokenManager: jwtTokenManagerFactory,
+      identityProviders: [],
+    );
+  });
+
+  withServerpod('Given a streaming endpoint authenticated with JWT', (
```
SandPod: chore: This isn't using the end to end code but working through the test framework.

We want to use a real e2e test to validate this behavior.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/authentication_tokens.dart
exaby73: The issue is that the `rotateRefreshToken` method in the "user" class AKA `AuthenticationTokens` publishes that `authenticationRevoked` message. Else it's just a call to `admin.rotateRefreshToken`. How do you propose we separate that? Do we just have one `refreshAccessToken` method on `AuthenticationTokens` and it call `admin.rotateRefreshToken` and publish the messages as well?

SandPod: Yes, that sounds reasonable right? No need to have both of them publishing messages? I'm not really sure why both would be exposed at all. That would scope down their responsibility as well.

exaby73: Fixed in a242c02e1183

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/test/integration/jwt_refresh_streaming_endpoint_test.dart
exaby73: Fixed in 19f78204f190

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/lib/src/endpoints/authenticated_streaming_test_endpoint.dart
exaby73: I was getting a hanging test on this change. I spent a bit of time trying to debug it but decided to leave it out for now since the current code is working for my test cases. If you feel strongly on this, we can pair to understand what is going on

templates/pubspecs/tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/pubspec.yaml
```diff
@@ -13,6 +13,7 @@ dependencies:
     sdk: flutter
   serverpod_auth_bridge_flutter: SERVERPOD_VERSION
   serverpod_auth_client: SERVERPOD_VERSION
+  serverpod_auth_core_client: SERVERPOD_VERSION
```
SandPod: Question: Why is this dependency required? I believe that everything we need should be exposed through the `serverpod_auth_idp_client`? It should include the `core_client`.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
```diff
@@ -0,0 +1,122 @@
...
+    late Stream<int> stream;
+
+    setUp(() async {
+      userId = await client.authTest.createTestUser();
+      authSuccess = await client.authTest.createJwtToken(userId);
+
+      final tokenParts = authSuccess.token.split('.');
+      if (tokenParts.length != 3) {
+        throw Exception('Invalid JWT token format');
+      }
```
SandPod: Suggestion: Remove, we don't care about the format in this test. Other tests are responsible for validating this works as expected.

We shouldn't have to change this test if we modify the format.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
```diff
@@ -0,0 +1,122 @@
...
+        tearDown(() async {
+          await streamSubscription.cancel();
+        });
+
+        test('then stream remains open', () {
+          expect(streamClosedCompleter.isCompleted, isFalse);
+        });
+
+        test(
+          'when the last refreshed token is revoked, then the stream closes with an error.',
```
SandPod: Suggestion: `When`, `When` in the test structure.

Please straighten this out.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
```diff
@@ -0,0 +1,122 @@
...
+      'http://localhost:8080/',
+      authenticationKeyManager: authKeyManager,
+    );
+  });
+
+  tearDown(() {
+    client.close();
+  });
+
+  group('Given a streaming endpoint authenticated with JWT', () {
```
SandPod: Suggestion: How about you change the setup and description to: 
`Given a streaming method authenticated with a JWT token that has since been refreshed multiple times`
`when latest revision of token is revoked`
`then streaming method is closed with "<Right exceptio>"`
?

That would make the setup heavy but the test case simple.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
```diff
@@ -0,0 +1,122 @@
...
+              refreshToken: authSuccess.refreshToken!,
+            );
+
+            authSuccess = newAuthSuccess;
+            await authKeyManager.put(authSuccess.token);
+
+            await Future<void>.delayed(const Duration(milliseconds: 150));
+          }
+
+          final tokenParts = authSuccess.token.split('.');
```
SandPod: Suggestion: Run this through the authentication handler for the token instead. That gives us the same value that would be received when creating streaming connections and we don't need to know the internal structure of the JWT token.

We might change it and this test should not fail in that case.

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/lib/src/endpoints/authenticated_streaming_test_endpoint.dart
SandPod: OK, this is not critical but it shouldn't be a problem getting this working. Perhaps you were not closing the right stream? (output or input)?

templates/pubspecs/tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/pubspec.yaml
exaby73: Ahh I missed that one. My IDEs auto import didn't pick that up. Fixed in 9ac778ccd6de

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
exaby73: Fixed in 8b417e77cf9e

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
exaby73: Fixed in 75438c431ff9

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
exaby73: Since I removed this one in 8b417e77cf9e, is this still valid?

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
exaby73: Given the changes in 75438c431ff9, is this one valid?

tests/serverpod_new_auth_test/serverpod_new_auth_test_server/lib/src/endpoints/authenticated_streaming_test_endpoint.dart
exaby73: I needed the stream to be closed by the `authenticationRevoked` event so I wasn't closing the stream in the tests. But I think the way it was set up caused race condition (my hypothesis, I didn't investigate too deeply on it)

templates/pubspecs/tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/pubspec.yaml
```diff
@@ -30,8 +30,6 @@ dependency_overrides:
     path: ../../../modules/new_serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_flutter
   serverpod_auth_client:
     path: ../../../modules/serverpod_auth/serverpod_auth_client
-  serverpod_auth_core_client:
```
SandPod, line 33: I think the depdency override is still needed as it is a transitive dependency of the package.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
```diff
@@ -0,0 +1,112 @@
...
+
+        await valueReceivedCompleter.future;
+
+        for (var i = 0; i < 3; i++) {
+          final newAuthSuccess = await client.jwtRefresh.refreshAccessToken(
+            refreshToken: authSuccess.refreshToken!,
+          );
+
+          authSuccess = newAuthSuccess;
+          await authKeyManager.put(authSuccess.token);
```
SandPod: Suggestion: Remove this as it only causes confusion. We are not going to make any more requests so we want the old key to be untainted.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
```diff
@@ -0,0 +1,112 @@
...
+
+          await Future<void>.delayed(const Duration(milliseconds: 150));
```
SandPod: Question: Why is this required? How about you await the completion of `valueReceivedCompleter` instead? Then we asert the connection is up but don't care about anything else.

templates/pubspecs/tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/pubspec.yaml
exaby73: Reverted in c1324d0ee55a

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
```diff
@@ -0,0 +1,112 @@
...
+        }
+      });
+
+      tearDown(() async {
+        await streamSubscription.cancel();
+      });
+
+      test('then the stream remains open', () {
+        expect(streamClosedCompleter.isCompleted, isFalse);
+      });
```
SandPod: Suggestion: Remove this and add this as an assert after the value has been received in the setup.

We don't need to validate this here.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
SandPod: No, then we can close this! 👍

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
SandPod: Yes, this was still part of the code (in the setup). Please remove it.

exaby73: Fixed in 442bd7ab7c33

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
exaby73: I removed it but then added it back below the for loop to save the latest token as that's what needed to be revoked in the following test

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
exaby73: Fixed in 442bd7ab7c33

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
exaby73: Removed in 442bd7ab7c33

SandPod: Assert is missing.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
SandPod: We need to store `authSuccess.token` but we don't need to add it to the authKeyManager.

Please make it explicit what token we are revoking in the call to `revokeSpecificToken` by adding it as an input parameter. Just like you had before when you saved it as `finalToken`.

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart
exaby73: I added this: https://github.com/exaby73/serverpod/blob/5b934eef047f5c2492fcf404c90080618c5b9bbf/tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/jwt_refresh_streaming_endpoint_test.dart#L60

SandPod: Please use asserts in `setup` and expects in tests.

SandPod (approved):
LGTM :rocket:

PR #4200, opened 2025-11-13
feat: Add compatibility improvements to S3CloudStorage for S3-compatible providers (MinIO, LocalStack, etc.)

mathis6787:
This PR improves the compatibility of S3CloudStorage with S3-compatible providers such as MinIO, LocalStack, and other self-hosted or local S3 implementations.
It aligns the behavior of Serverpod with non-AWS S3 providers while preserving full AWS S3 compatibility.

These adjustments make the S3 storage backend usable in local development environments and with S3-compatible services.

⸻

💡 Why

Many developers use S3-compatible storage such as MinIO or LocalStack for:
	•	local development
	•	CI testing
	•	on-premise/self-hosted deployments

Serverpod’s current S3 implementation is AWS-specific, which makes it difficult to use these environments.
This PR expands compatibility without introducing breaking changes.

⸻

❌ Breaking changes

None — existing AWS configurations continue to work exactly the same.


mathis6787:

integrations/serverpod_cloud_storage_s3/lib/src/aws_s3_upload/aws_s3_upload.dart
```diff
@@ -29,8 +29,13 @@ class AwsS3Uploader {
 
     /// The AWS region. Must be formatted correctly, e.g. us-west-1
     required String region,
+
+    /// Optional custom endpoint for S3-compatible storage (e.g. http://localhost:9000/my-bucket)
+    String? endpointUrl,
   }) async {
-    final endpoint = 'https://$bucket.s3-$region.amazonaws.com';
+    final endpoint = (endpointUrl != null && endpointUrl.isNotEmpty)
+        ? '${endpointUrl.endsWith('/') ? endpointUrl : '$endpointUrl/'}$bucket'
```
nielsenko: This is the crux of this PR right?

integrations/serverpod_cloud_storage_s3/lib/src/aws_s3_upload/aws_s3_upload.dart
```diff
@@ -107,11 +112,16 @@ class AwsS3Uploader {
...
+    String? endpointUrl,
+
     /// The filename to upload as. If null, defaults to the given file's current filename.
     required String uploadDst,
     bool public = true,
   }) async {
-    final endpoint = 'https://$bucket.s3-$region.amazonaws.com';
+    final endpoint = (endpointUrl != null && endpointUrl.isNotEmpty)
+        ? '${endpointUrl.endsWith('/') ? endpointUrl : '$endpointUrl/'}$bucket'
+        : 'https://$bucket.s3-$region.amazonaws.com';
```
nielsenko: Could we extract into method to avoid repetition?

integrations/serverpod_cloud_storage_s3/lib/src/aws_s3_upload/aws_s3_upload.dart
```diff
@@ -191,13 +201,18 @@ class AwsS3Uploader {
...
     /// The filename to upload as. If null, defaults to the given file's current filename.
     required String uploadDst,
     Duration expires = const Duration(minutes: 10),
     int maxFileSize = 10 * 1024 * 1024,
     bool public = true,
   }) async {
-    final endpoint = 'https://$bucket.s3-$region.amazonaws.com';
+    final endpoint = (endpointUrl != null && endpointUrl.isNotEmpty)
+        ? '${endpointUrl.endsWith('/') ? endpointUrl : '$endpointUrl/'}$bucket'
+        : 'https://$bucket.s3-$region.amazonaws.com';
```
nielsenko: again..

nielsenko (changes requested):
Some first feedback

integrations/serverpod_cloud_storage_s3/lib/src/aws_s3_upload/aws_s3_upload.dart
mathis6787: Yes

integrations/serverpod_cloud_storage_s3/lib/src/aws_s3_upload/aws_s3_upload.dart
mathis6787: done

nielsenko (commented):
You have added a number of optional arguments at various places, that is needed when building endpoints in the case we are not calling aws s3.

Would it be an idea, to abstract the code that needs these arguments into a callback type, use a default version for S3, and allow caller to pass another factory if needed?

This way you get more flexibility and less custom params (only an optional callback)? 

Just food for thought. Otherwise, I'm prepared to accept this PR.

mathis6787:
Hi! Thanks a lot for the feedback — I appreciate it.

I am  new to open source contribution so before implementing the factory/callback abstraction, I want to confirm one important point to avoid going in the wrong direction because I am not sure to understand your intention.

✔️ My current understanding

The idea is to avoid having multiple optional parameters (endpointUrl, host, useHttps, etc.) and replace them with a single abstraction (factory or similar) that lets callers customize the S3 client in a cleaner, more flexible way.

✔️ Why I’m unsure about the factory-only approach

A clientFactory gives full control over how AwsS3Client is created.
But the issue is that AwsS3Uploader does not use AwsS3Client at all.

It builds its own URL internally:
	•	hard-coded AWS URL:
https://$bucket.s3-$region.amazonaws.com
	•	and in my PR I added:
endpointUrl so MinIO / LocalStack works

Because AwsS3Uploader is completely independent from AwsS3Client:

➤ Even if we add a clientFactory,

we would still need endpointUrl on AwsS3Uploader anyway.

So the factory alone does not remove optional parameters from the constructor.

This is why I’m unsure how the factory solves the “too many optional S3 params” problem.

✔️ Why an override object may be more consistent

If we had something like:
```dart
 class S3Overrides {
  final String? endpointUrl;
  final String? host;
  final bool useHttps;
}
```
Then both components could use it:
	•	AwsS3Client → host, useHttps
	•	AwsS3Uploader → endpointUrl,
	•	getPublicUrl → useHttps 
	•	createDirectFileUploadDescription → endpointUrl

In that approach, the only optional parameter of S3CloudStorage would be a single overrides argument, and callers would pass a S3Overrides object.
This removes all the individual optional parameters (endpointUrl, host, useHttps, etc.) from the constructor and centralizes S3-compatibility settings in one place.

1.	Do you prefer that both AwsS3Client and AwsS3Uploader receive the same S3Overrides object?
So overrides are applied consistently everywhere.
2.	Or do you want clientFactory callback and we let in the S3CloudStorage constructor the optional  params endpointUrl (may be modify to endpointOverride to be more explicit) ?
(Factory only handles the client, override object handles uploader)

nielsenko:
@mathis6787 Thank you for taking the time to consider my suggestion. A simple callback is indeed too simple. I like your `S3Overrides` object, but I would like you to also encapsulated the actual `endpointUrl` calculation in that, and have a default version for regular AWS S3. Does that make sense?

mathis6787:
## Refactor S3 storage configuration with S3Overrides

Replaced multiple optional S3 configuration parameters (`endpointUrl`, `host`, `useHttps`) with a single `S3Overrides` object to improve API clarity and maintainability.

### Changes
- Created `S3Overrides` class to encapsulate endpoint configuration for S3-compatible storage
- Refactored `S3CloudStorage` constructor to accept `S3Overrides?` instead of individual parameters
- Updated `AwsS3Uploader` to require pre-built endpoint (removed internal `_buildEndpointUrl`)
- `S3Overrides` handles endpoint URL normalization (supports with/without trailing slash)
- Kept `publicHost` as separate parameter for flexibility

### Usage
**AWS S3 (unchanged behavior):**
```dart
S3CloudStorage(
  serverpod: pod,
  storageId: 'default',
  region: 'us-east-1',
  bucket: 'my-bucket',
  public: true,
)
```

**MinIO/S3-compatible:**
```dart
S3CloudStorage(
  serverpod: pod,
  storageId: 'default',
  region: 'us-east-1',
  bucket: 'my-bucket',
  public: false,
  overrides: S3Overrides(
    endpointUrl: 'http://localhost:9000',
    host: 'localhost:9000',
    useHttps: false,
  ),
)
```

Tested with MinIO and with AWS S3.

integrations/serverpod_cloud_storage_s3/lib/src/aws_s3_client/client/client.dart
```diff
@@ -97,23 +103,33 @@ class AwsS3Client {
...
+/${unencodedPath.split('/').map(Uri.encodeComponent).join('/')}
 $canonicalQuery
-host:$_host
-x-amz-content-sha256:$payload
-x-amz-date:$datetime
-x-amz-security-token:${_sessionToken ?? ""}
-
-host;x-amz-content-sha256;x-amz-date;x-amz-security-token
+${canonicalHeaders.toString()}
+${signedHeaders.toString()}
```
nielsenko:
```suggestion
$canonicalHeaders
$signedHeaders
```
Drop redundant `toString()`

integrations/serverpod_cloud_storage_s3/lib/src/aws_s3_client/client/client.dart
```diff
@@ -122,17 +138,20 @@ $payload''';
...
     final signature = SigV4.calculateSignature(signingKey, stringToSign);
 
-    final authorization = [
-      'AWS4-HMAC-SHA256 Credential=$_accessKey/$credentialScope',
-      'SignedHeaders=host;x-amz-content-sha256;x-amz-date;x-amz-security-token',
-      'Signature=$signature',
-    ].join(',');
+    final authorization =
+        'AWS4-HMAC-SHA256 Credential=$_accessKey/$credentialScope, '
+        'SignedHeaders=${signedHeaders.toString()}, '
```
nielsenko:
```suggestion
        'SignedHeaders=$signedHeaders, '
```
.. again

integrations/serverpod_cloud_storage_s3/lib/src/cloud_storage/s3_overrides.dart
```diff
@@ -0,0 +1,55 @@
+/// Configuration for S3 or S3-compatible storage endpoint overrides.
+///
+/// Use this class to configure connections to S3-compatible storage services
+/// like MinIO, LocalStack, or custom S3 implementations.
+class S3Overrides {
+  S3Overrides({
+    this.endpointUrl,
+    this.host,
+    this.useHttps = true,
+  });
```
nielsenko: Instead of making this configurable, couldn't we have a default implementation for AWS, and have people who want to configure it implement the interface directly?

integrations/serverpod_cloud_storage_s3/lib/src/aws_s3_upload/aws_s3_upload.dart
```diff
@@ -107,13 +108,13 @@ class AwsS3Uploader {
     /// The AWS region. Must be formatted correctly, e.g. us-west-1
     required String region,
 
+    /// Full endpoint URL including bucket (e.g. https://my-bucket.s3-us-east-1.amazonaws.com or http://localhost:9000/my-bucket)
+    required String endpoint,
+
```
nielsenko, line 115: Adding a required parameter is a breaking change. Could we make it configurable, similar to how you did with `S3Options`?

nielsenko: .. so that user can pass a custom object that will build the endpoint, from the bucket and whatever is required by the S3 wanna-be service?

Then this parameter could be optional and default to regular AWS S3 behavior.

mathis6787:
Thanks for the feedback!  
I applied the requested changes:

- Introduced ` change S3Overrides for S3Options` and changed it to an abstract class.
- Added `AwsS3Options` and `MinioS3Options` as concrete implementations.
- Remove required for endpoint in AWsUploader
- Cleaned up the redundant `toString()` usage.

I hope this aligns with what you had in mind

mathis6787:
@nielsenko 
Any update ?

nielsenko:
@mathis6787 Sorry for the delay. I do have a pending review, but the race to release 3.0 took precedence. I'm currently on vacation, but will return in the new year.

nielsenko:
@mathis6787 We decided to do  an overhaul of the storage implementations in 3.4.  You may want to try building your min-io integration on top.

mathis6787:
Your new package works very well with s3 Compat. I am closing this PR

Issue #4205, opened 2025-11-13
Duplicated deserialization code for classes referenced in multiple relation fields


marcelomendoncasoares:
### Describe the bug

When a list, set or map of a model class is referenced in multiple relation fields across different models (or even the same model), the code generator creates duplicate deserialization code in the generated `protocol.dart` file. Although this causes no real problem, it is duplicated and dead code on the `SerializationManager.deserialize` method.

### To reproduce

Create one or more models that have relation fields referencing the same list, set or map type. It can be on the same model or on different models, like what happens in the `serverpod_test_server` test project with `List<Person>?` fields.

```yaml
# tests/serverpod_test_server/lib/src/models/models_with_list_relations/city.spy.yaml
class: City
table: city
fields:
  name: String
  citizens: List<Person>?, relation
  organizations: List<Organization>?, relation(name=city_org)
```

```yaml
# tests/serverpod_test_server/lib/src/models/models_with_list_relations/organization.spy.yaml
class: Organization
table: organization
fields:
  name: String
  people: List<Person>?, relation(name=org_person)
  city: City?, relation(name=city_org, optional)
```

Because the `City` model references both the `List<Person>?` and the `List<Organization>?`, and the `Organization` model references the `List<Person>?`, the code generator will create duplicate deserialization code for the `List<Person>?` type.

https://github.com/serverpod/serverpod/blob/cf276aa9ac00c2e3e616139c09b99dfd483b5614/tests/serverpod_test_server/lib/src/generated/protocol.dart#L8336-L8355

### Expected behavior

Each entry in the `SerializationManager.deserialize` method should be unique, regardless of how many models reference that type.

### Library version

3.0.0-alpha.3

### Platform information

Any

### Additional context

The issue is caused in the `library_generator.dart` file:

https://github.com/serverpod/serverpod/blob/a9f471564b406e428149c2e27393470e6d1972f1/tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart#L184-L193

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

Issue #4135, opened 2025-10-31
Deserialization in tests fails if endpoint returns a record


TheLastGimbus:
### Describe the bug

When return type of an endpoint function is a record, deserialization logic iniside generated test utilities breaks, and tries to deserialize-already-deserialized objects:

```dart
// greeting_endpoint.dart
Future<(Greeting, String)> hello(Session session, String name) async {
  return (
    Greeting(
      message: 'Hello $name',
      author: 'Serverpod',
      timestamp: DateTime.now(),
    ),
    name
  );
}
```

```dart
withServerpod('Given Greeting endpoint', (sessionBuilder, endpoints) {
  test(
    'when calling `hello` with name then returned greeting includes name',
    () async {
      // Call the endpoint method by using the `endpoints` parameter and
      // pass `sessionBuilder` as a first argument. Refer to the docs on
      // how to use the `sessionBuilder` to set up different test scenarios.
      final greeting = await endpoints.greeting.hello(sessionBuilder, 'Bob');
      expect(greeting.$1.message, 'Hello Bob');
    },
  );
});
```

```
Server default listening on port 8080
package:serverpod_records_test_server/src/generated/protocol.dart 37:36  Protocol.deserialize
package:serverpod_records_test_server/src/generated/protocol.dart 44:9   Protocol.deserialize
test/integration/test_tools/serverpod_test_tools.dart 135:32             _GreetingEndpoint.hello.<fn>.<fn>
===== asynchronous gap ===========================
test/integration/test_tools/serverpod_test_tools.dart 129:33             _GreetingEndpoint.hello.<fn>
===== asynchronous gap ===========================
package:serverpod_test/src/function_call_wrappers.dart 12:12             callAwaitableFunctionAndHandleExceptions
===== asynchronous gap ===========================
test/integration/greeting_endpoint_test.dart 19:26                       main.<fn>.<fn>

type '_GreetingImpl' is not a subtype of type 'Map<String, dynamic>'
```

For some reason, stuff inside to-deserialize map that represents a record, is, already, deserialized

<img width="625" height="758" alt="Image" src="https://github.com/user-attachments/assets/70f8e9ce-ff4f-42a3-9b4f-2908a33b4fcd" />

Then, it tries to deserialize them individually:

<img width="629" height="823" alt="Image" src="https://github.com/user-attachments/assets/0f1fede5-d324-4a5d-ae2e-d3332f1efd38" />

...but it already is:

[...]

PR #4227, opened 2025-11-15
feat: Implement support for polymorphism on endpoints

### LINKED ISSUE #2909: Deserialization of subclasses in inheritance
Subtask to: https://github.com/serverpod/serverpod/issues/1873

When using inheritance through serialization, the runtime type must be preserved to ensure that the subclasses can be instantiated correctly after deserialization.

In the following scenarios, the runtime type of the subclass is lost:
1. Return type of a non-streaming endpoint method.
2. As a parameter to any endpoint method.
3. As a field on a model that is passed through a serialization roundtrip (endpoint, database, etc.).
4. Contained inside of a container type such as `List` or `Map` passed through a serialization roundtrip (endpoint, database, etc.). 

Below are some scenarios that need to be supported:

#### Inheritance in endpoint
Given a `ChildClass` that inherits from a `ParentClass`, the runtime type of the child class would be lost in the following scenarios:

```dart
Future<ParentClass> myEndpointMethod(Session session) async {
  // The type of the return value would assume the ParentClass in the client.
  return ChildClass(....);
}
```

```dart
// The type of `value` would assume the parrent class even if ChildClass is passed from the client.
Future<ParentClass> myEndpointMethod(Session session, ParentClass value) async {
  assert(value.runtimeType == ChildClass);
  ....
}
```

#### Field serialization roundtrip
Given a `ContainerClass` where a field named `contents` has the type `ParentClass` but the instance of that field is a `ChildClass` that inherits from `ParentClass`, this should be supported:

```dart
ContainerClass container(contents: ChildClass(...));
// Serialization strips the knowledge of the runtime type for the child class
final afterSerializationRoundtrip = serializationRoundtrip(container);
final contents = afterSerializationRoundtrip.contents;

assert(container.contents == contents.runtimeType);
```

**Additional context**
- Streaming endpoints already handle this by always encoding transported objects with their runtime type through the `encodeWithType` method in the Serialization Manager. But note that this does not apply to any fields of the model.

marcelomendoncasoares:
This PR implements the missing piece for model inheritance on the Serverpod framework: support for polymorphism on the serialization for non-streaming endpoints. The feature implementation consisted in:

- Add a `__className__` field to the JSON serialization of all models carrying the name of the model.
- Make the `Protocol().deserialize` check for such field and call `deserializeByClassName` if present.
- Centralize all deserialization calls for `List`, `Set` and `Map` on the `Protocol().deserialize` method instead of directly inside the `fromJson` method of each class.

The central piece on the polymorphic deserialization is the `deserializeByClassName` method, which was already implemented for usage in the streaming endpoints with the exact feature of parsing the model from the specified `className` - that now goes inside the JSON.

Closes #2909.

During development, two bugs were found and fixed in the serialization of models:

Fixes: #4135
Fixes: #4205

Since these fixes were blocking of this feature and would also touch all of the models, it was preferred to solve them here together instead of creating a blocker and conflicts.

## Pre-launch Checklist

## Breaking changes

The implementation is fully backwards compatible and introduces no breaking changes.

### DIFF
--- a/packages/serverpod_serialization/lib/src/extensions/serialization_extensions.dart
+++ b/packages/serverpod_serialization/lib/src/extensions/serialization_extensions.dart
@@ -143,24 +143,6 @@ extension SetJsonExtension<T> on Set<T> {
 
     return map<dynamic>(valueToJson).toList();
   }
-
-  /// Returns a deserialized version of the [Set].
-  static Set<T>? fromJson<T>(
-    dynamic value, {
-    required T Function(dynamic) itemFromJson,
-  }) {
-    if (value is Set<T>) return value;
-
-    var set = (value as List?)?.map(itemFromJson).toSet();
-
-    if (set != null && value!.length != set.length) {
-      throw Exception(
-        'Input list for Set contained duplicate items. List with length ${value.length} resulted in a set with only ${set.length} item(s).',
-      );
-    }
-
-    return set;
-  }
 }
 
 /// Expose toJson on Vector
--- a/tests/serverpod_test_server/test/generated_methods/from_json_test.dart
+++ b/tests/serverpod_test_server/test/generated_methods/from_json_test.dart
@@ -626,20 +626,13 @@ void main() {
     );
 
     test(
-      'when deserializing from JSON with duplicate values, then a TypeError is thrown',
+      'when deserializing from JSON with duplicate values, then duplicates are removed',
       () {
-        expect(
-          () => TypesSet.fromJson({
-            'aString': ['test', 'test'],
-          }),
-          throwsA(
-            isA<Exception>().having(
-              (e) => e.toString(),
-              'message',
-              contains('Input list for Set contained duplicate items'),
-            ),
-          ),
-        );
+        var typeSet = TypesSet.fromJson({
+          'aString': ['test', 'test'],
+        });
+        expect(typeSet.aString?.length, 1);
+        expect(typeSet.aString?.first, 'test');
       },
     );
 


tools/serverpod_cli/lib/src/generator/dart/library_generators/util/class_generators_util.dart
```diff
@@ -278,160 +268,51 @@ Expression _buildEnumTypeFromJson(
...
-    );
-  }
-
   return CodeExpression(
-    Block.of([
-      valueExpression
-          .asA(
-            CodeExpression(
-              Code('List${type.nullable ? '?' : ''}'),
+    _getProtocolReference(serverCode, config)
```
SandPod, line 279: Question: Shouldn't the changes in this commit [16e2d356392ffa68b227b3fa382ed81db5c5e082] trigger a bunch of failures in unit tests?

tests/serverpod_test_client/lib/src/protocol/changed_id_type/many_to_many/course.dart
```diff
@@ -32,9 +33,11 @@ abstract class CourseUuid implements _i1.SerializableModel {
           ? null
           : _i1.UuidValueJsonExtension.fromJson(jsonSerialization['id']),
       name: jsonSerialization['name'] as String,
-      enrollments: (jsonSerialization['enrollments'] as List?)
-          ?.map((e) => _i2.EnrollmentInt.fromJson((e as Map<String, dynamic>)))
-          .toList(),
+      enrollments: jsonSerialization['enrollments'] == null
+          ? null
+          : _i3.Protocol().deserialize<List<_i2.EnrollmentInt>>(
```
SandPod, line 38: Question: `Protocol()` is a hidden singleton, so this shouldn't have a cost. But if we ever change that this might become expensive depending on how much we store in `Protocol`.

To avoid this, is it possible to make `deserialize` a static method? Or do we work on a runtime instance for deserialization somewhere?

SandPod (commented):
Solution looks good!

I'm, however,  missing some backwards compatibility tests to ensure that the serialization format used in Serverpod 1.x and 2.x is still supported.

Perhaps manually calling an endpoint with the old format can validate that the response is parsed correctly?

And what is our view on porting our streaming methods over to use this format? We could save some space by not using the old serializeByClassName. There is no code modifying that flow at the moment.

tests/serverpod_test_client/lib/src/protocol/changed_id_type/many_to_many/course.dart
marcelomendoncasoares: I agree with you on this, but I would consider this as a concern for such future since this `Protocol()` call is spread across the repository. Here I only reused the same pattern that we were already applying for record types and now is standard for all non-primitive types.

tools/serverpod_cli/lib/src/generator/dart/library_generators/util/class_generators_util.dart
marcelomendoncasoares: Indeed, tests that expected the exact string of the serialized model broke due to missing the `__className__` key. Those were fixed at https://github.com/serverpod/serverpod/pull/4227/commits/586403d82e75fbb976699037884745f31f5e29ee - and one last regarding `Set` using a custom `fromJson` at https://github.com/serverpod/serverpod/pull/4227/commits/2637ae3763f5ecb41a1ff6eb0e2045fddc634f59.

PR #4232, opened 2025-11-17
refactor: Rename AuthenticationTokens and AuthSessions

### LINKED ISSUE #4222: Rename AuthenticationToken and AuthSessions
Provide a better name for these two classes that better reflect the technology used for authentication.

A suggestion could be JWT and Server Side Sessions.

Make sure to rename the models and implementation classes as well and all documentation.

exaby73:
Rename `AuthenticationTokens` and `AuthSessions` to match their underlying technologies better

Closes #4222

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/business/server_side_sessions.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/business/server_side_sessions.dart
@@ -9,7 +9,7 @@ import '../../generated/protocol.dart';
 import '../util/session_key_hash.dart';
 import 'server_side_sessions_admin.dart';
 import 'server_side_sessions_config.dart';
-import 'session_key.dart';
+import 'server_side_sessions_token.dart';
 
 /// Management functions for [ServerSideSession]s.
 ///
@@ -45,7 +45,7 @@ final class ServerSideSessions {
     final Session session,
     final String key,
   ) async {
-    final sessionKeyParts = tryParseSessionKey(session, key);
+    final sessionKeyParts = tryParseServerSideSessionToken(session, key);
     if (sessionKeyParts == null) {
       return null;
     }
@@ -211,7 +211,7 @@ final class ServerSideSessions {
 
     return AuthSuccess(
       authStrategy: AuthStrategy.session.name,
-      token: buildSessionKey(
+      token: buildServerSideSessionToken(
         secret: secret,
         serverSideSessionId: serverSideSession.id!,
       ),
--- a/modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/business/server_side_sessions_token.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/business/server_side_sessions_token.dart
@@ -13,9 +13,9 @@ final _sessionKeyPrefix = utf8.encode('sas');
 /// characters and thus does not change with the later data.
 final _sessionKeyPrefixBase64 = base64Url.encode(_sessionKeyPrefix);
 
-/// Creates the String representation to map to an `ServerSideSession`.
+/// Creates the String representation to map to a `ServerSideSession`.
 @internal
-String buildSessionKey({
+String buildServerSideSessionToken({
   required final UuidValue serverSideSessionId,
   required final Uint8List secret,
 }) {
@@ -26,12 +26,12 @@ String buildSessionKey({
   ]);
 }
 
-/// Tries parsing a session key String created by [buildSessionKey] into its
+/// Tries parsing a server side session token String created by [buildServerSideSessionToken] into its
 /// parts.
 ///
 /// Returns `null` if it does not match the spec or parsing fails for any reason.
 @internal
-SessionKeyData? tryParseSessionKey(
+SessionKeyData? tryParseServerSideSessionToken(
   final Session session,
   final String key,
 ) {
--- a/modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/session/integration/server_side_sessions_authentication_handler_test.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/session/integration/server_side_sessions_authentication_handler_test.dart
@@ -3,7 +3,7 @@ import 'dart:convert';
 import 'package:clock/clock.dart';
 import 'package:serverpod/serverpod.dart';
 import 'package:serverpod_auth_core_server/session.dart';
-import 'package:serverpod_auth_core_server/src/session/business/session_key.dart';
+import 'package:serverpod_auth_core_server/src/session/business/server_side_sessions_token.dart';
 import 'package:test/test.dart';
 
 import '../../serverpod_test_tools.dart';
@@ -62,14 +62,14 @@ void main() {
   ) {
     late Session session;
     late UuidValue authUserId;
-    late String sessionKey;
+    late String serverSideSessionToken;
 
     setUp(() async {
       session = sessionBuilder.build();
 
       authUserId = (await serverSideSessions.authUsers.create(session)).id;
 
-      sessionKey = (await serverSideSessions.createSession(
+      serverSideSessionToken = (await serverSideSessions.createSession(
         session,
         authUserId: authUserId,
         scopes: {},
@@ -82,7 +82,7 @@ void main() {
       () async {
         final authInfo = await serverSideSessions.authenticationHandler(
           session,
-          sessionKey,
+          serverSideSessionToken,
         );
 
         expect(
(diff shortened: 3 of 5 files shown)


examples/auth/auth_server/config/template.passwords.yaml
```diff
@@ -19,9 +19,9 @@ development:
   serviceSecret: 'P6bXXWpVVtxlN1aaiQOZG0J2cZXXXVcj'
 
   # Authentication managers configuration
-  authSessionsSessionKeyHashPepper: 'test_session_key_pepper'
-  authenticationTokenRefreshTokenHashPepper: 'test_refresh_token_pepper'
-  authenticationTokenPrivateKey: 'test_private_key'
+  serverSideSessionsSessionKeyHashPepper: 'test_session_key_pepper'
```
marcelomendoncasoares: Discussion: Since we use `jwt` as an acronym, should we also use `sss` instead of the more verbose `serverSideSessions`? This would also prevent the "SessionsSession" sequence that is confusing to read.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/lib/src/protocol/protocol.dart
```diff
@@ -16,27 +16,27 @@ import 'auth_user/models/auth_user_model.dart' as _i3;
...
-import 'jwt/models/authentication_token_info.dart' as _i7;
+import 'jwt/models/jwt_token_info.dart' as _i7;
 import 'jwt/models/refresh_token_expired_exception.dart' as _i8;
 import 'jwt/models/refresh_token_invalid_secret_exception.dart' as _i9;
 import 'jwt/models/refresh_token_malformed_exception.dart' as _i10;
 import 'jwt/models/refresh_token_not_found_exception.dart' as _i11;
 import 'jwt/models/token_pair.dart' as _i12;
 import 'profile/models/user_profile_model.dart' as _i13;
-import 'session/models/auth_session_info.dart' as _i14;
+import 'session/models/server_side_session_info.dart' as _i14;
```
marcelomendoncasoares, line 26: Discussion: Should the top level folder still call `session` or also be changed to `sss` to follow the same `jwt` pattern?

```suggestion
import 'sss/models/server_side_session_info.dart' as _i14;
```
> This suggestion is only here to highlight the proposal, not to be applied.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/lib/src/protocol/protocol.dart
```diff
@@ -182,8 +182,8 @@ class Protocol extends _i1.SerializationManager {
         return 'TokenPair';
       case _i13.UserProfileModel():
         return 'UserProfileModel';
-      case _i14.AuthSessionInfo():
-        return 'AuthSessionInfo';
+      case _i14.ServerSideSessionInfo():
```
marcelomendoncasoares, line 216: Question: Is this the dual for the `JwtTokenInfo`? If so, shouldn't it be called `<sssPrefix>TokenInfo` as well?

```suggestion
      case _i14.ServerSideSessionTokenInfo():
```
> This suggestion is only here to highlight the proposal, not to be applied.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/jwt.dart
```diff
@@ -32,11 +32,11 @@ final class AuthenticationTokens {
   /// Management functions for auth users.
   final AuthUsers authUsers;
 
-  /// Creates a new instance of [AuthenticationTokens].
-  AuthenticationTokens({
+  /// Creates a new instance of [Jwt].
+  Jwt({
```
marcelomendoncasoares, line 37: Discussion: It seems to me that calling this class just `Jwt` does not communicate exactly what it does. Does it represents a JWT token? Is it for managing JWT tokens? Maybe calling it `JwtTokens` might clarify it - the same applies to SSS to call something like `sssTokens` instead of only `ServerSideSessions`.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/business/jwt.dart
```diff
@@ -32,11 +32,11 @@ final class AuthenticationTokens {
...
     required this.config,
     this.authUsers = const AuthUsers(),
-  }) : admin = AuthenticationTokensAdmin(
+  }) : admin = JwtAdmin(
```
marcelomendoncasoares: If we change `Jwt` to `JwtTokens`, we should also change this to `JwtTokensAdmin`.

marcelomendoncasoares (commented):
Only a few discussions regarding the new names.

exaby73:
`AuthUsers` cannot be renamed to `AuthUser` as `AuthUser` exists as the database model and `AuthUsers` as the management class. Same with `UserProfile` and `UserProfiles`

examples/auth/auth_server/lib/server.dart
```diff
@@ -21,27 +21,29 @@ void run(List<String> args) async {
   );
 
   // Configure our token managers.
-  final authSessionsConfig = AuthSessionsConfig(
-    sessionKeyHashPepper: pod.getPassword('authSessionsSessionKeyHashPepper')!,
+  final serverSideSessionsConfig = ServerSideSessionsConfig(
+    sessionKeyHashPepper: pod.getPassword(
+      'serverSideSessionsSessionKeyHashPepper',
```
SandPod: Suggestion: Use `serverSideSessionKeyHashPepper`.

Singular like the class.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/business/server_side_sessions.dart
```diff
@@ -7,31 +7,31 @@ import 'package:serverpod_shared/serverpod_shared.dart';
...
 import 'session_key.dart';
 
-/// Management functions for [AuthSession]s.
+/// Management functions for [ServerSideSession]s.
 ///
-/// This should be used instead of [AuthSession.db].
-final class AuthSessions {
-  final AuthSessionsConfig _config;
+/// This should be used instead of [ServerSideSession.db].
+final class ServerSideSessions {
```
SandPod, line 17: Question: Shouldn't this be `ServerSideSession`?

both according to the documentation but also according to what was agreed before?

SandPod: This change should then be propagated to all places.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/business/session_key.dart
```diff
@@ -13,15 +13,15 @@ final _sessionKeyPrefix = utf8.encode('sas');
 /// characters and thus does not change with the later data.
 final _sessionKeyPrefixBase64 = base64Url.encode(_sessionKeyPrefix);
 
-/// Creates the String representation to map to an `AuthSession`.
+/// Creates the String representation to map to an `ServerSideSession`.
 @internal
 String buildSessionKey({
```
SandPod: Suggestion: Should this class and the method be changed to `server_side_session_token.dart` and `buildServerSideSessionToken`?

examples/auth/auth_server/lib/server.dart
exaby73: Fixed in 15149d43b423

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/business/server_side_sessions.dart
exaby73: `ServerSideSession` is a model class at `modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/models/server_side_session.spy.yaml` so unfortunately, we cannot rename this one as there are name conflicts

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/business/session_key.dart
exaby73: FIxed in 9ab0a334074a

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/business/server_side_sessions_token.dart
exaby73, line 9: Question: Should we change this to:

```suggestion
/// "sss" being abbreviated of "server_side_session"
final _sessionKeyPrefix = utf8.encode('sss');
```

SandPod: I don't think so as this is a marker for our specific tokens. We don't use this as a marker for what it is, only to filter out so we know we created it.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/business/server_side_sessions.dart
SandPod: Got it, leave as is!

PR #4248, opened 2025-11-19
feat: Create a simplified yaml-based auth configuration experience

### LINKED ISSUE #4245: Introduce convenience `defaults` factory method on `AuthServices` for simple auth onboarding
While our current configuration on server is robust and well built for extensibility, it also presents the user with a lot of nested objects upfront, which can be daunting for new users that are only interested in a default auth experience. This became clearer while writing the new auth documentation, since all objects are already present on the first snippets of code.

To avoid this, we should provide on the server a similar zero-config "just works" experience as we have on the client, that can be replaced as needed for more complex use cases. Below is a possible implementation example:

```dart

  final pod = Serverpod(
    args,
    Protocol(),
    Endpoints(),
  );

  // Only this addition would be necessary for a clean experience
  pod.initializeAuthServices(
    emailIDPOptions: EmailIDPOptions(
      sendRegistrationVerificationCode: _sendRegistrationVerificationCode,
      sendPasswordResetVerificationCode: _sendPasswordResetVerificationCode,
    ),
  );

  // Start the server.
  await pod.start();
}
```

The `initializeAuthServices` extension method could do all the heavy lifting of parsing the configuration keys from the `passwords.yaml` file or other location and initialize the providers with the default configurations and a default token manager.

marcelomendoncasoares:
Closes #4245.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/examples/auth/auth_server/lib/server_yaml_based_setup.dart
+++ b/examples/auth/auth_server/lib/server_yaml_based_setup.dart
@@ -15,10 +15,35 @@ void run(List<String> args) async {
 
   // Initialize
   pod.initializeAuthServices(
-    emailIDPOptions: EmailIDPOptions(
-      sendRegistrationVerificationCode: _sendRegistrationVerificationCode,
-      sendPasswordResetVerificationCode: _sendPasswordResetVerificationCode,
-    ),
+    tokenManagers: [
+      // AuthSessionsTokenManagerFactory(
+      //   AuthSessionsConfig(
+      //     sessionKeyHashPepper: pod.getPassword(
+      //       'authSessionsSessionKeyHashPepper',
+      //     )!,
+      //   ),
+      // ),
+      AuthenticationTokensTokenManagerFactory(
+        AuthenticationTokenConfig(
+          refreshTokenHashPepper: pod.getPassword(
+            'authenticationTokenRefreshTokenHashPepper',
+          )!,
+          algorithm: AuthenticationTokenAlgorithm.hmacSha512(
+            SecretKey(pod.getPassword('authenticationTokenPrivateKey')!),
+          ),
+        ),
+      ),
+    ],
+    identityProviders: [
+      // GoogleIdentityProviderFactory.fromKeys(pod.getPassword),
+      // AppleIdentityProviderFactory.fromKeys(pod.getPassword),
+      EmailIdentityProviderFactory.fromKeys(
+        pod.getPassword,
+        sendRegistrationVerificationCode: _sendRegistrationVerificationCode,
+        sendPasswordResetVerificationCode: _sendPasswordResetVerificationCode,
+      ),
+      // PasskeyIdentityProviderFactory.fromKeys(pod.getPassword),
+    ],
   );
 
   // Start the server.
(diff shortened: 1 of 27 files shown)


modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_idp_config.dart
```diff
@@ -44,10 +44,67 @@ typedef SendRegistrationVerificationCodeFunction =
...
+    sendRegistrationVerificationCode: options.sendRegistrationVerificationCode,
+    sendPasswordResetVerificationCode:
+        options.sendPasswordResetVerificationCode,
+    onPasswordResetCompleted: options.onPasswordResetCompleted,
+    failedLoginRateLimit: options.failedLoginRateLimit,
+    passwordValidationFunction: options.passwordValidationFunction,
+    maxPasswordResetAttempts: options.maxPasswordResetAttempts,
+    secretHashSaltLength: options.secretHashSaltLength,
+  );
+}
```
marcelomendoncasoares: Discussion: Is it worth changing this to composition to remove this redundancy - and the need for tests that ensure all optional properties are mapped? This would pass the burden of another layer of config object for all that use `EmailIDPConfig`, but these will be the most advanced users.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/serverpod_auth_core_server.dart
```diff
@@ -3,8 +3,18 @@
...
+// export 'src/common/business/multi_token_manager.dart';
 export 'src/common/integrations/provider_factory.dart';
 export 'src/common/integrations/token_manager.dart';
+export 'src/common/integrations/token_manager_factory.dart';
 export 'src/generated/endpoints.dart';
 export 'src/generated/protocol.dart';
+
+// export 'src/jwt/jwt.dart';
+// export 'src/profile/profile.dart';
+// export 'src/session/session.dart';
```
marcelomendoncasoares: Discussion: Should we unify the export API? Since Dart already does tree shaking, is it really necessary for users to have to find where specific imports are and add several imports to their projects?

examples/auth/auth_server/lib/server_yaml_based_setup.dart
```diff
@@ -0,0 +1,50 @@
...
+    Endpoints(),
+  );
+
+  // Initialize
+  pod.initializeAuthServices(
+    emailIDPOptions: EmailIDPOptions(
+      sendRegistrationVerificationCode: _sendRegistrationVerificationCode,
+      sendPasswordResetVerificationCode: _sendPasswordResetVerificationCode,
+    ),
+  );
```
marcelomendoncasoares, line 47: Comment: This is the simplified experience that we provide with this PR, with all providers being automatically initialized according to the configurations present on the `passwords.yaml` or the provided `getConfigFunction` parameter.

modules/new_serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/serverpod_auth_core_server.dart
marcelomendoncasoares: Decision was to unify this, since the core package is meant to be used as a whole.

Issue #4054, opened 2025-10-13
Add logic to disable execution of future calls when none are registered


SandPod:
### Problem to Solve

Serverpod currently attempts to execute future calls even when there are no future calls registered. This leads to unnecessary queries to the database.

### Proposal

Update the future call execution logic to first check if any future calls are registered. If none are found, exit the execution routine immediately and skip further processing.

Registering a future call should enable the execution of future calls, even if this happens when the server is running.

### Use Case

Support hybernating the database when no future calls are known and the server does not do any database operations.

### Alternatives

N/A

### Additional context

This enhancement would improve overall server efficiency.

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [x] I want to work on this

AliYar-Khan:
@SandPod I would like to work on this. I have checked the code and my proposal is below in future_call_manager.dart file:

```
 void _dispatchEntries(List<FutureCallEntry> entries) {
    if (entries.isNotEmpty) { //check if the entries list is not empty
      final callbacks = entries.map<TaskCallback?>((entry) {
        final futureCall = _futureCalls[entry.name];

        if (futureCall == null) {
          // TODO: this should be logged or caught otherwise.
          // https://github.com/serverpod/serverpod/issues/3485
          return null;
        }

        return () => _runFutureCall(
              futureCallEntry: entry,
              futureCall: futureCall,
            );
      });

      _scheduler.addTaskCallbacks(callbacks.nonNulls.toList());
    }
  }
```

SandPod:
@AliYar-Khan - Could we add this check on an even higher level?
As far as I understand, there are two ways to "start" the future call manager, either through `runScheduledFutureCalls()` or `start()`.

Couldn't we have a mechanism in the FutureCallManager that does not trigger the `scanner.start` if `registerFutureCall` has never been called?

And if `start` has already been called (and `stop` has not been called in between) and `registerFutureCall` is called, then we actually start the `scanner`?

AliYar-Khan:
@SandPod i will further check it if we can move it to higher level or not

AliYar-Khan:
@SandPod yes we can introduce a mechanism to not trigger ```scanner.start``` at all if ```registerFutureCall``` is not called at all

To support this behavior, I am proposing these changes:

- Track whether start() was requested.
- Track whether the scanner is currently started.
- Track whether any future calls have been registered.
[...]

PR #4283, opened 2025-11-23
feat: Implement lazy scanning for future calls in FutureCallManager and FutureCallScanner

AliYar-Khan:
Serverpod currently attempts to execute future calls even when there are no future calls registered. This leads to unnecessary queries to the database.
This PR enhances the FutureCallManager and FutureCallScanner classes to check if any future calls are registered, and only then run the queries.

Fixes: #4054 

## Pre-launch Checklist

### DIFF
--- a/packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
+++ b/packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
@@ -43,6 +43,7 @@ class FutureCallManager {
 
   late final ServerpodTaskScheduler _scheduler;
   late final FutureCallScanner _scanner;
+  bool _startRequested = false;
 
   /// Creates a new [FutureCallManager]. Typically, this is instantiated
   /// internally by the [Serverpod].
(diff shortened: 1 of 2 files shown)


AliYar-Khan:
@SandPod Please check this pull request. Thank you

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -149,13 +157,20 @@ class FutureCallManager {
   /// Starts the [FutureCallManager], enabling it to monitor the database
   /// for overdue future calls and execute them automatically.
   void start() {
-    _scanner.start();
+    _startRequested = true;
+    if (_futureCalls.isNotEmpty && !_scanner.isRunning) {
```
SandPod, line 161: Question: Isn't `_scanner.start()` idempotent, which means we don't need to guard against calling `start` multiple times?

That would make it possible to remove the check for `!_scanner.isRunning`.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -149,13 +157,20 @@ class FutureCallManager {
...
   }
 
   /// Stops the [FutureCallManager], preventing it from monitoring and
   /// executing overdue future calls.
   Future<void> stop() async {
     await _scanner.stop();
+    _startRequested = false;
+    if (_scanner.isRunning) {
+      await _scanner.stop();
+    }
```
SandPod, line 173: Chore: This looks incorrect, we are calling stop multiple times. Please look over this logic.

Do we need to guard against `_scanner.isRunning`? Can't we always call `stop()`?

SandPod:
@AliYar-Khan - Are you still interested on working on this? This has become a required feature we are going to need for Serverpod 3, so if you don't have the capacity to work on this we will implement it.

AliYar-Khan:
If it is urgent then you guys can continue to work on this. 

> @AliYar-Khan - Are you still interested on working on this? This has become a required feature we are going to need for Serverpod 3, so if you don't have the capacity to work on this we will implement it.

PR #4240, opened 2025-11-17
fix: Misc fixes related to cloud migration.

SandPod:
Misc fixes that were identified when migrating Serverpod Cloud to the new auth RC.

### Changes
- [01b5732bb7fdd4e7516485a80cf3bd0945fa0af7, 2a98f4f21fa0cbce2f2bcb6a5f3e202138bae138] - Removes requirement for redundant parameters for `AuthServices` and `EmailIDPUtils`
- [765fd92f9d4931c9fbaffe5c8b041d35065a048f, 609a8b07bad72fedbf69bc3485fc892d8e7f18d8] - Organizes exports from core module.
- [98b5169b8d3e66d92ccf94f1c2182cbf69c261f1] - Make authId non-nullable. Required to be set by all authentication handlers now.
- [fe988cc2957b0c69a649e168f6cb4593daf01421] - Add callback for when an account is created.
- [df47a8897d13280e11b9b6d5cf4ca77bf74874db] - Return auth user id from password reset. (**Should we return email account id here instead?**)
- [27cfdca9554243ac58dcca7388ba104978f48d8c] - Makes our cache control factories static.
- [9ce232ce1bb0a03617c2ceef09327554bd5bc34b] - Use Duration to configure max age.
- [bc07515f58f184db1e1ddf2b64afb900d88edca8] - Add symmetric getter extension for ClientAuthSessionManager.

## Pre-launch Checklist

- [ ] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [ ] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
- [ ] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.
- [ ] I added new tests to check the change I am making.
- [ ] All existing and new tests are passing.
- [ ] Any breaking changes are documented below.

If you need help, consider asking for advice on the [discussion board](https://github.com/serverpod/serverpod/discussions).

## Breaking changes

_If you have done any breaking changes, make sure to outline them here, so that they can be included in the notes for the next release._

## Summary by CodeRabbit

# Release Notes

* **New Features**
[...]

PR #4288, opened 2025-11-24
chore: Bumps version to 3.0.0-rc.2 and updates CHANGELOG.

marcelomendoncasoares:
Release of the next release candidate version.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/examples/auth/auth_client/pubspec.yaml
+++ b/examples/auth/auth_client/pubspec.yaml
@@ -8,8 +8,8 @@ environment:
   sdk: '^3.8.0'
 
 dependencies:
-  serverpod_auth_idp_client: 3.0.0-rc.1
-  serverpod_client: 3.0.0-rc.1 
+  serverpod_auth_idp_client: 3.0.0-rc.2
+  serverpod_client: 3.0.0-rc.2 
 
 dependency_overrides: 
   serverpod_auth_idp_client:
--- a/examples/auth/auth_flutter/pubspec.yaml
+++ b/examples/auth/auth_flutter/pubspec.yaml
@@ -30,8 +30,8 @@ dependencies:
   extended_image: ^10.0.1
   flutter:
     sdk: flutter
-  serverpod_auth_idp_flutter: 3.0.0-rc.1
-  serverpod_flutter: 3.0.0-rc.1
+  serverpod_auth_idp_flutter: 3.0.0-rc.2
+  serverpod_flutter: 3.0.0-rc.2
 
 dev_dependencies:
   flutter_lints: '>=3.0.0 <7.0.0'
--- a/examples/auth/auth_server/pubspec.yaml
+++ b/examples/auth/auth_server/pubspec.yaml
@@ -10,12 +10,12 @@ environment:
   sdk: '^3.8.0'
 
 dependencies:
-  serverpod: 3.0.0-rc.1
-  serverpod_auth_idp_server: 3.0.0-rc.1
+  serverpod: 3.0.0-rc.2
+  serverpod_auth_idp_server: 3.0.0-rc.2
 
 dev_dependencies:
   lints: '>=3.0.0 <7.0.0'
-  serverpod_test: 3.0.0-rc.1
+  serverpod_test: 3.0.0-rc.2
   test: ^1.25.5
 
 dependency_overrides: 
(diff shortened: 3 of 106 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,36 @@
...
+
+Release candidate for Serverpod 3.
+
+Serverpod 3 is a major overhaul of the authentication system and the web server.
+This release candidate is **not yet production-ready**. It is still under active development and may contain bugs or breaking changes.
+
+### Polymorphism support
+Polymorphic models are now supported on endpoints, making inheritance no longer an experimental feature on the framework. It works both for sealed and unsealed models.
+
+This has been a long-awaited feature that solves one of the very few areas where invoking client endpoints behave differently than regular method calls. See the [next documentation](https://docs.serverpod.dev/next/concepts/models) for more details.
```
SandPod: Suggestion/Discussion: Add this to the feats list.

I'm not sure we want to feature this as the same level as Auth and Web server? In the real release notes, would we list this as a Feat (with some credits to ben) or a proper section?

CHANGELOG.md
```diff
@@ -1,3 +1,36 @@
...
+
+### New auth improvements from last rc
```
SandPod, line 14: Suggestion: List these changes as a list, possible even in the same list as the rest? Or is it worth separating it?

CHANGELOG.md
```diff
@@ -1,3 +1,36 @@
...
+- fix: Clarifies DB migration version not found error message.
+- fix: SessionLogEntry time field now uses session start time.
+- fix: Publishes authentication revoked on refresh token rotate expiry and invalid secret failures.
+- fix: Prevents null check error when relation defined without table.
+- fix: Uses daemon exit code conventions for SIGTERM graceful shutdown.
+- fix: Make connectionTimeout final to prevent post-initialization mutation
+
+#### Misc
+- chore(deps): Bumps analyzer from 8.4.1 to 9.0.0 in serverpod_cli.
+- chore: Updates various dependencies.
```
SandPod: Suggestion: Remove, we don't usually list dependency bumps.

CHANGELOG.md
```diff
@@ -1,3 +1,36 @@
...
+### New auth improvements from last rc
+All services that use secrets now support multiple fallback mechanisms to allow secret rotation without invalidating existing tokens. This includes JWT and server-side session secrets, as well as Email hash peppers.
+
+The `ClientAuthSessionManager` has been moved to the `serverpod_auth_core_client` package, decoupling from Flutter  dependencies to allow the usage of Serverpod auth in pure Dart projects. For Flutter apps, there is a new `FlutterAuthSessionManager` that extends the base class and adds Flutter-specific functionality.
+
+A major rename has also taken place in the new authentication module to clarify purpose of each class and method. t should now be easier to understand and extend core functionality and identity providers.
+
+### Additional changes
+
+#### Fixes
```
SandPod, line 37: Suggestion: Add a fix for making `authId` required in AuthenticationInfo and that Duration is used as max age for cache control.

These breaking changes are a bit hidden in this PR: https://github.com/serverpod/serverpod/pull/4240

CHANGELOG.md
marcelomendoncasoares: I agree that it should not feature the same spot as the web server and the new auth, but, being a feature that is long coming and awaited by the community, I thought it would be worth giving it some higher visibility - both here and on the actual release notes. My thought on this is that it contributes to the marketing of the release due to the impact on the user experience. However, since this is a decision on how we want to market the release, your input is the most relevant on whether we should or not give it a greater highlight.

CHANGELOG.md
marcelomendoncasoares: I preferred to separate since the new auth is a focus of the new release, and users that have started adopting the first RC could spot clearly on what has improved on this version - and why we are releasing it. But it could be reduced to a list and together with the rest if we don't think this highlight is important.

CHANGELOG.md
SandPod: Then let's keep it and decide what to do for the final release later. I agree that this is a bigger feature.

CHANGELOG.md
SandPod: Sounds good, we can keep it as a separate section, but transform it to a list following the rest of the changelogs.

Issue #1730, opened 2024-01-08
Have Serverpod auto-add custom SQL commands on initial migration


lukehutch:
**Is your feature request related to a problem? Please describe.**

When creating an initial migration (initializing a database), I have to manually run additional SQL commands, to create tables and indexes that Serverpod doesn't yet know how to create (e.g. for Geolocation columns, or to create a partial index -- there are issues for each of these).

**Describe the solution you'd like**

It would be nice if there was a directory that custom `.sql` files could be put in, which are merged in with Serverpod's own SQL files when a migration is created.

SandPod:
Great suggestion, we have been discussing this in the team as well and are planning to expand the migrations system with the capability of adding custom SQL commands.

lukehutch:
Thanks. The custom commands will have to be run after the Serverpod commands, because a user might have to override something due to lack of Serverpod support.

However, the table diff system is going to have to handle these sorts of manual modifications though, to avoid "the table is not like the protocol definition" type warnings, and migration problems.

Isakdl:
I think there might be a need for one pre- and one post- SQL file. Imagine you have a manually managed model and you want to create a relation to it for example. Then the table must exist before the serverpod SQL is executed.

SandPod:
A possible solution is that the Serverpod Migration system generates four additional SQL files `pre_database_setup.sql`, `post_database_setup.sql`, `pre_migration.sql` and `post_migration.sql`.

The `pre/post_database_setup.sql` files are utilized if a database is created from scratch. Any SQL code contained in these files would be applied as a single transaction with the `definition.sql` contents. The `pre_database_setup.sql` contents would be inlined before the `definition.sql` contents and the `postDatabaseSetup` contents would be inlined after.

The `pre/post_migration.sql` files are utilized when the database is rolled forward. These files live inside of each migration. Any SQL code in these fields would be applied as a single transaction with the `migration.sql` contents. The `pre_migration.sql` contents would be inlined before the `migration.sql` contents and the `post_migration.sql` contents would be inlined after.

### Modules

Special consideration would need to be taken for modules. Any custom commands added to modules would need to be inlined in the correct places as well. And if custom code is supported in modules, module changes would need to always be applied migration by migration in order to catch any custom code.

mathis6787:
Hi!
I’ve opened a PR  https://github.com/serverpod/serverpod/pull/4319 for this issue.

https://github.com/serverpod/serverpod/pull/4319

[...]

Issue #3873, opened 2025-08-11
Programmatic method to apply migrations before server start


SandPod:
### Problem to Solve

Currently, database migrations in Serverpod can only be applied as part of the server startup process. This limits workflows where a developer needs to apply migrations before starting the server, for example data transformations that depend on the new schema.

### Proposal

Introduce a dedicated method (e.g., `applyMigrations()`) that can be called programmatically without starting the server. This method should:
- Use the same logic as migrations during startup.
- Apply only pending migrations.
- Be idempotent to prevent reapplying completed migrations.
- Provide clear feedback on applied changes and any errors.

This dedicated method could possibly live under a `maintenance` namespace that allows other maintenance mode operations such as running all scheduled future calls, performing the health check, and updating the log settings.

### Use Case

A developer wants to apply the latest schema changes, then run a one-off Dart script to backfill or transform data before starting the server. This avoids starting the server in a partially migrated state.

### Alternatives

Manually starting the server with migrations enabled, then stopping it to perform pre-startup tasks. This is inefficient and risks unexpected behavior if the server begins handling requests during the migration process.

### Additional context

-

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

christerswahn:
Should this maybe be designed in conjunction with the service lifecycle interface? E.g. standardizing something along these lines:

1. Serverpod object creation
2. Configurations ingestion
3. Backend services initialization (connecting to backend services like DB and Redis)
4. Migrations
5. Start serving requests (incl future calls)

Today there is only 1 and 5 with user-added "stuff" in between. A well-defined pattern on how to do this, how to add new services to this process, and inject tasks pre/post these steps would be nice!

PR #4319, opened 2025-11-27
feat: Add support for custom sql

mathis6787:
This PR introduces a complete implementation of custom SQL support in Serverpod migrations, matching the design discussed in the issue. Closes: #1730 

It adds four SQL hook files for every migration:
	•	pre_database_setup.sql
	•	post_database_setup.sql
	•	pre_migration.sql
	•	post_migration.sql

These files allow developers to inject custom SQL at every stage of the migration lifecycle.

## Features

✅ 1. Custom SQL Execution Pipeline
Custom SQL is now executed in strict order during:

#### Fresh database creation
	1.	pre_database_setup.sql
	2.	definition.sql
	3.	post_database_setup.sql

#### Migration
	1.	pre_migration.sql
	2.	migration.sql
	3.	post_migration.sql

All custom SQL is executed in a single transaction, after being combined safely.

⸻

#### 2. CLI Support — Files Are Auto-Generated and Accumulated
The CLI now:
	•	Automatically creates all four files for each migration
	•	Copies previous custom SQL forward when creating a new migration
	•	Accumulates migration SQL into the next setup SQL so fresh DB creation is consistent

This ensures feature-parity between:
	•	rolling forward migrations
	•	creating a fresh database

⸻

#### 3. Warning Categorization for Schema Integrity
Database integrity verification now distinguishes:
	•	Informational warnings (e.g., custom indexes, triggers)
	•	Critical mismatches that must block server startup

Custom SQL can no longer wrongly trigger catastrophic schema mismatch failures.

⸻

#### 4. Full Backward Compatibility
If the four custom SQL files are empty:
	•	Nothing changes
	•	The system behaves exactly as before
	•	No SQL is executed
	•	No informational warnings appear

No breaking changes for existing users.

⸻

### Why this is needed

Serverpod currently lacks a clean way to support custom index:
	•	Custom indexes (BM25 (paradeDB, POSTGIST,Partial index…)

At the moment, adding these manually results in “the table is not like the protocol definition” warnings, and in debug mode Serverpod quits the process entirely, blocking development and causing migration problems.

⸻

##🧱 Implementation follows the design proposed by SandPod in #1730

## About Modules (I need guidance)

I implemented safe behavior for modules during migration execution (their pre/post_migration.sql runs normally).

However, I am still not 100% certain how to tackle the module guideline and if it's actually needed.

⸻

## Known Design Issue: Fresh Database Consistency

####❗ Problem

post_migration.sql and pre_migration.sql are only executed when rolling forward.

If a developer deletes the database and recreates it from scratch, custom SQL added in migration N (pre/post_migration.sql, is not applied until migration N+1 is created.

Example
	1.	Developer adds a Postgis index  on a table inside post_migration.sql
	2.	--apply-migrations → Index created ✅
	3.	The database is dropped and recreated
	4.	--apply-migrations →  the postgis index will be missing ❌
	5.	Only after creating the next migration does the index appear in post_database_setup.sql

🔍 Possible solutions (need maintainers feedback)
	1.	Dynamic accumulation at runtime (best DX)
	•	When creating a fresh DB, read all pre/post_migration.sql and inject them into setup
	2.	Runtime update of setup files
	3.	Keep current behavior + documentation
	•	Expected behavior: developers must create a new migration to propagate custom SQL
	4.	Add a CLI command
	•	Regenerates setup files from migration SQL

Any other ideas ? 

I can implement any of these depending on feedback.

⸻

📦 Breaking Changes

None.


FantaMagier:
@mathis6787 Great RP! I need a GIN Index so it could be also used for that

mathis6787:
> @mathis6787 Great RP! I need a GIN Index so it could be also used for that

Hi @FantaMagier thanks for the comment , you should be able to use GIN index withouth my PR https://docs.serverpod.dev/concepts/database/indexing

FantaMagier:
@mathis6787 Thanks. Yes but I need Indexes on List<String> at best as JSONB Format for smart array search

marcelomendoncasoares:
Hi @mathis6787! 👋 

Thanks for this great contribution! We are currently heads down on the release of Serverpod 3.0, but we will soon be able to review your PR.

There are also some concerns that we will have to review carefully with this feature regarding the database analyzer and how to deal with conflicts of objects created in the additional SQL files (like the mentioned custom indexes).

marcelomendoncasoares:
Hi @mathis6787! Now that we have released Serverpod 3.0, we are slowly getting back to PRs pending review. Regarding this one, we still have some discussions on the best approach to allow custom SQL commands, but without handling a footgun to users - that could seem to allow arbitrary commands, but would quickly end up in inconsistent database state and prevent server startup. This was an aspect that was still pending an answer on the issue and must be solved before we consider this feature complete.

As we are soon going to work on migration hooks for #3873, I'll propose for us to park this one for now and consider the needs for the parent issue of this PR on that development. Then, we will be able to perform a broader analysis on the best approach to solve both problems and come back here. Would you be ok with this approach?

mathis6787:
Hi @marcelomendoncasoares , thanks for the update, I am totally fine with it.

mathis6787:
Hey 👋 @marcelomendoncasoares 

I noticed that issue #3873 was recently moved back in the roadmap, and I also went through the new migrations RFC (#4604). From what I understand, support for custom SQL hooks is currently out of scope for this iteration, and there are still some ongoing discussions around the broader migration architecture.

Given those changes, I wanted to check in regarding this PR and the original discussion around it.

Do you think this PR is still aligned with the current direction, or has the scope shifted enough that it would make more sense to close it and revisit the topic once the migration work is further along?

Additionally, do you have any rough timeframe in mind (even very approximate) for when custom SQL / migration hooks might be revisited? This would help me decide whether to wait for the framework support or plan a workaround on my side for now, as this is starting to become blocking for some use cases (e.g. custom indexes or extensions like pg_textsearch / ParadeDB).

Totally fine either way.
Thanks!

marcelomendoncasoares:
> Hey 👋 @marcelomendoncasoares
> 
> I noticed that issue #3873 was recently moved back in the roadmap, and I also went through the new migrations RFC (#4604). From what I understand, support for custom SQL hooks is currently out of scope for this iteration, and there are still some ongoing discussions around the broader migration architecture.
> 
> Given those changes, I wanted to check in regarding this PR and the original discussion around it.

Hi @mathis6787! Sorry for the delay in getting back to you here. In fact, migration hooks are out of the migrations UX RFC, but our plan was to revisit them for #3873 that could be required for auth user migration. Turns out we'll probably not need to work on that now, given a change in the strategy for the user migration.

> Do you think this PR is still aligned with the current direction, or has the scope shifted enough that it would make more sense to close it and revisit the topic once the migration work is further along?

The major issue with the current approach - [the same proposed on the issue](https://github.com/serverpod/serverpod/issues/1730#issuecomment-2077048006) - is that it might end up putting a footgun in the hand of users. [As one comment already states](https://github.com/serverpod/serverpod/issues/1730#issuecomment-1882539524), if the extra files are used to create additional objects, the analyzer might flag the database state as inconsistent and prevent the server start. So we can't expose an official way for users to run custom SQL without being able to allow ignoring the created elements in the migration system.

> Great RP! I need a GIN Index so it could be also used for that

This comment above already shows the exact expectation that such files would create. Even worst, if such objects are of custom types unknown to Serverpod, the migration system might not even be able to repair itself, leaving the database in a state that can only be repaired by manually destroying the custom objects.

> Additionally, do you have any rough timeframe in mind (even very approximate) for when custom SQL / migration hooks might be revisited?

Unfortunately I can't give a timeline when we plan to work on it, but I can say that it should not be that soon. First we intend to enhance the migrations UX - executing part or most of the RFC - to then revisit other parts of the process. It is possible that these priorities change, but this is the current overview.

> This would help me decide whether to wait for the framework support or plan a workaround on my side for now, as this is starting to become blocking for some use cases (e.g. custom indexes or extensions like pg_textsearch / ParadeDB).

If you know well what you're doing, you can always go by with changing the manually generated SQL files. But if what you need is just support for custom indexes or extensions, do open an issue that we can evaluate extending Serverpod to support it!

Given all above, I'll then close this PR to be revisited when we have a better solution for the migration system to support custom objects. Thanks for your contribution and comprehension!

PR #4320, opened 2025-11-28
refactor: Simplify UX of interacting with `AuthServices` and IDPs on the server

### LINKED ISSUE #4245: Introduce convenience `defaults` factory method on `AuthServices` for simple auth onboarding

marcelomendoncasoares:
Closes #4245.

This PR is best reviewed commit by commit, since it contains multiple changes to the authentication module. Commit messages contain more detailed descriptions of each change.

## Pre-launch Checklist

## Breaking changes

Several changes in names of parameters, classes, and methods.


examples/auth/auth_server/lib/server.dart
```diff
@@ -90,20 +90,16 @@ void run(List<String> args) async {
   );
 
   AuthServices.set(
-    primaryTokenManager: ServerSideSessionsTokenManagerFactory(
-      serverSideSessionsConfig,
-    ),
+    tokenManagers: [
```
SandPod: Suggestion: Shouldn't we make this change for the normal constructor as well? No need to split it into different lists.

examples/auth/auth_server/lib/server_from_passwords.dart
```diff
@@ -0,0 +1,72 @@
...
+import 'package:serverpod_auth_idp_server/providers/apple.dart';
+import 'package:serverpod_auth_idp_server/providers/email.dart';
+import 'package:serverpod_auth_idp_server/providers/google.dart';
+import 'package:serverpod_auth_idp_server/providers/passkey.dart';
+import 'package:serverpod_auth_idp_server/core.dart';
+
+import 'src/generated/protocol.dart';
+import 'src/generated/endpoints.dart';
+
+// This is the starting point of your Serverpod server. In most cases, you will
```
SandPod, line 17: Question: Should we have some form of description connected to this file? Either in the file directly or externally?

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/utils/email_idp_account_creation_util.dart
```diff
@@ -652,21 +652,21 @@ class EmailIdpCompleteAccountCreationResult {
...
-/// The result of the [EmailIdpAccountCreationUtil.finalizeAccountRequest] operation.
+/// The result of the [EmailIdpAccountCreationUtil.completeAccountCreation] operation.
 ///
 /// This describes the detailed status of the operation to the caller.
 ///
 /// In the general case the caller should take care not to leak this to clients,
 /// such that outside clients can not use this result to determine whether a
 /// specific account is registered on the server.
-class EmailIdpFinalizeAccountRequestResult {
+class EmailIdpCompleteAccountRequestResult {
```
SandPod, line 656: Question: There was actually a subtle difference between finalize and complete, are we sure we want to remove it?

SandPod (approved):
LGTM! Couple of smaller questions.

examples/auth/auth_server/lib/server.dart
marcelomendoncasoares: This was intentional since we do use the first with a priority, to communicate this expressively on the constructor of the class - and reflect its intention on the usage.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/utils/email_idp_account_creation_util.dart
marcelomendoncasoares: This only made the name compatible with the method that actually returns this class, which was changes previously from `finalizeAccountRequest` to `completeAccountRequest`.

examples/auth/auth_server/lib/server_from_passwords.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4320/commits/e8078f7e16834e95f1c41981f72a723857f1021f.

Issue #4253, opened 2025-11-19
Deprecate and remove AuthenticationKeyManager from serverpod_client


SandPod:
## Problem to Solve

The `AuthenticationKeyManager` class in `serverpod_client` is currently deprecated and serves only as a thin wrapper around `ClientAuthKeyProvider`. This creates confusion for developers and adds unnecessary complexity to the codebase. The class is already marked as deprecated but still exists in the codebase, and there are still references to it throughout the codebase that need to be migrated.

Having both `AuthenticationKeyManager` and `ClientAuthKeyProvider` creates:
- Confusion about which interface to use
- Maintenance burden of keeping deprecated code
- Potential for new code to accidentally use the deprecated class
- Inconsistency in the API surface

## Proposal

1. **Remove `AuthenticationKeyManager` from `serverpod_client`**:
   - Delete `packages/serverpod_client/lib/src/auth_key_manager.dart`
   - Remove the deprecated `authenticationKeyManager` getter from `ServerpodClientShared`
   - Update all internal references to use `ClientAuthKeyProvider` directly

2. **Migrate all usages**:
   - Update all classes that extend `AuthenticationKeyManager` to implement `ClientAuthKeyProvider` directly
   - Update all client constructors that accept `authenticationKeyManager` parameter to use `authKeyProvider` instead
   - Update examples, tests, and documentation

3. **Migration path for users**:
   - The replacement is straightforward: `AuthenticationKeyManager` already implements `ClientAuthKeyProvider`, so users can simply change their type annotations and imports
   - Update migration guide in CHANGELOG.md

**Code changes needed:**
- Remove `packages/serverpod_client/lib/src/auth_key_manager.dart`
- Update `packages/serverpod_client/lib/src/serverpod_client_shared.dart` to remove deprecated getter
- Migrate classes like `FlutterAuthenticationKeyManager`, `ServiceKeyManager`, `TestAuthKeyManager`, etc.
- Update generated client code templates
- Update all examples and tests

## Use Case

This cleanup will:
- Simplify the API for new developers who won't encounter the deprecated class
- Reduce maintenance burden by removing deprecated code
- Make the codebase more consistent by having a single interface for authentication key management
- Improve code clarity by removing an unnecessary abstraction layer

**Example migration:**
```dart
// Before
class MyKeyManager extends AuthenticationKeyManager {
  @override
  Future<String?> get() async => ...;
  // ...
}

// After
class MyKeyManager implements ClientAuthKeyProvider {
[...]

PR #4323, opened 2025-11-28
refactor: Deprecate AuthenticationKeyManager

exaby73:
Closes https://github.com/serverpod/serverpod/issues/4253

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/tests/serverpod_test_server/test_integration/test_tools/unauthenticated_annotation_test.dart
+++ b/tests/serverpod_test_server/test_integration/test_tools/unauthenticated_annotation_test.dart
@@ -1,5 +1,3 @@
-// ignore_for_file: deprecated_member_use
-
 import 'package:serverpod/serverpod.dart';
 import 'package:serverpod_test_client/serverpod_test_client.dart';
 import 'package:serverpod_test_server/test_util/config.dart';
@@ -10,6 +8,7 @@ import 'package:test/test.dart';
 void main() {
   late Serverpod server;
   late Client client;
+  // ignore: deprecated_member_use
   late AuthenticationKeyManager authKeyManager;
 
   setUpAll(() async {
@@ -19,6 +18,7 @@ void main() {
     authKeyManager = TestAuthKeyManager();
     client = Client(
       serverUrl,
+      // ignore: deprecated_member_use
       authenticationKeyManager: authKeyManager,
     );
 
--- a/tests/serverpod_test_server/test_integration/websockets/method_websockets/method_streams/reconnect_to_server_test.dart
+++ b/tests/serverpod_test_server/test_integration/websockets/method_websockets/method_streams/reconnect_to_server_test.dart
@@ -1,5 +1,3 @@
-// ignore_for_file: deprecated_member_use
-
 import 'dart:async';
 
 import 'package:serverpod/serverpod.dart';
@@ -19,6 +17,7 @@ void main() {
 
     client = c.Client(
       serverUrl,
+      // ignore: deprecated_member_use
       authenticationKeyManager: TestAuthKeyManager(),
     );
   });
--- a/tests/serverpod_test_server/test_integration/websockets/method_websockets/method_streams/revoked_authentication_stream_test.dart
+++ b/tests/serverpod_test_server/test_integration/websockets/method_websockets/method_streams/revoked_authentication_stream_test.dart
@@ -1,5 +1,3 @@
-// ignore_for_file: deprecated_member_use
-
 import 'dart:async';
 
 import 'package:serverpod/serverpod.dart';
@@ -35,6 +33,7 @@ void main() {
     authKeyManager = TestAuthKeyManager();
     client = c.Client(
       serverUrl,
+      // ignore: deprecated_member_use
       authenticationKeyManager: authKeyManager,
     );
   });
--- a/tests/serverpod_test_server/test_e2e_migrations/apply_repair_migration_test.dart
+++ b/tests/serverpod_test_server/test_e2e_migrations/apply_repair_migration_test.dart
@@ -1,5 +1,4 @@
 @Timeout(Duration(minutes: 5))
-// ignore_for_file: deprecated_member_use
 import 'package:serverpod_service_client/serverpod_service_client.dart';
 import 'package:serverpod_test_server/test_util/config.dart';
 import 'package:serverpod_test_server/test_util/migration_test_utils.dart';
@@ -9,6 +8,7 @@ import 'package:test/test.dart';
 void main() {
   var serviceClient = Client(
     serviceServerUrl,
+    // ignore: deprecated_member_use
     authenticationKeyManager: TestServiceKeyManager(
       '0',
       'super_SECRET_password',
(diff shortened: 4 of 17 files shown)


tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -954,6 +973,7 @@ class LibraryGenerator {
                           ],
                           {
                             'securityContext': refer('securityContext'),
+                            'authKeyProvider': refer('authKeyProvider'),
                             'authenticationKeyManager': refer(
```
exaby73: I couldn't figure out a way to add an ignore comment here to ignore the deprecation warning, so I added it in the header in `tools/serverpod_cli/lib/src/generator/code_generator.dart` which caused every file to be generated with an `ignore_for_file`. I'm open to suggestions on this one

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -214,6 +221,10 @@ abstract class ServerpodClientShared extends EndpointCaller {
     this.host,
     this.serializationManager, {
     dynamic securityContext,
+    ClientAuthKeyProvider? authKeyProvider,
```
marcelomendoncasoares: Discussion: Do we really need to expose the `authKeyProvider` as a parameter on the constructor? The ideia of its usage was to follow the `connectivityMonitor` pattern (set through cascade), and even using the extension method from the auth core instead of  on the parameter directly.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -134,10 +134,17 @@ abstract class ServerpodClientShared extends EndpointCaller {
...
-      authKeyProvider as AuthenticationKeyManager?;
+  AuthenticationKeyManager? get authenticationKeyManager {
+    final provider = authKeyProvider;
+    if (provider == null) return null;
+    if (provider is AuthenticationKeyManager) return provider;
+    throw StateError(
+      'The authKeyProvider is not an AuthenticationKeyManager. '
+      'Use authKeyProvider directly instead of authenticationKeyManager.',
+    );
+  }
```
marcelomendoncasoares, line 147: Question: Wasn't this listed for deletion on the issue?

packages/serverpod_service_client/lib/src/service_key_manager.dart
```diff
@@ -2,7 +2,7 @@ import 'package:serverpod_client/serverpod_client.dart';
 import 'package:serverpod_shared/serverpod_shared.dart';
 
 /// The key manager used for the service protocol.
-class ServiceKeyManager extends BearerAuthenticationKeyManager {
+class ServiceKeyManager implements ClientAuthKeyProvider {
```
marcelomendoncasoares: Suggestion: Should we now call it `ServiceKeyProvider` instead to conform to the new API?

tests/serverpod_new_auth_test/serverpod_new_auth_test_flutter/test/session_manager_migration_test.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: deprecated_member_use
```
marcelomendoncasoares: Question: If we initialize the `Client` using the `authKeyProvider` instead of the `authenticationKeyManager`, will this top level ignore still be needed? Or we are going to need only one or two deprecation ignores?

tests/serverpod_test_flutter/test_integration/auth_session_manager_signout_test.dart
```diff
@@ -29,6 +29,7 @@ void main() {
     setUp(() async {
       primaryClient = Client(
         serverUrl,
+        // ignore: deprecated_member_use
         authenticationKeyManager: FlutterAuthenticationKeyManager(
```
marcelomendoncasoares, line 33: Question: Same of the previous question - can't we use the `authKeyProvider` directly?

packages/serverpod_client/lib/src/serverpod_client_shared.dart
exaby73: Yes but after some discussion, we need to deprecate first. This field did not have a deprecated annotation in 2.X.X so we can't remove it yet

packages/serverpod_service_client/lib/src/service_key_manager.dart
exaby73: Wouldn't that be a breaking change unless since this package is only used internally, we're okay with that?

packages/serverpod_client/lib/src/serverpod_client_shared.dart
exaby73: Would it make the deprecation notice confusing though? Other than writing a long deprecation message with usage of cascade to set it, if we keep the now `Use authKeyProvider instead. This will be removed in future releases.` it would be confusing as users would not know immediately that it's supposed to be set via the cascade syntax.

tests/serverpod_test_flutter/test_integration/auth_session_manager_signout_test.dart
exaby73: These tests are against the legacy auth API so I didn't migrate these to the new API

tests/serverpod_test_server/test_integration/test_tools/unauthenticated_annotation_test.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: deprecated_member_use
```
marcelomendoncasoares: Can we change this for individual deprecated ignores as we did on the other files?

tests/serverpod_test_server/test_integration/websockets/method_websockets/method_streams/reconnect_to_server_test.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: deprecated_member_use
```
marcelomendoncasoares: Same here.

tests/serverpod_test_server/test_integration/websockets/method_websockets/method_streams/revoked_authentication_stream_test.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: deprecated_member_use
```
marcelomendoncasoares: Same here.

marcelomendoncasoares (changes requested):
Just some `ignore_for_file` to be replaced by individual ignores and we are good to go!

tests/serverpod_test_server/test_integration/websockets/method_websockets/method_streams/revoked_authentication_stream_test.dart
exaby73: Fixed in 83bb447879aa

PR #4327, opened 2025-11-28
fix: Fix `serverpod_cli` endpoint generate flaky tests

### LINKED ISSUE #4301: Flaky `serverpod_cli` endpoint method tests failing randomly
While running tests on PR #4276, unit tests run ended with some flaky error on the [endpoint_method_test.dart](https://github.com/serverpod/serverpod/actions/runs/19677527134/job/56362764457?pr=4276) file that were not related to the changes.

```
Running unit tests in all project that contain a test directory

==== Running tests in ./tools/serverpod_cli ====
+181 -1: test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart: Given an endpoint method with Stream parameter that returns a Future with nullable type when analyzed then endpoint definition is created. [E]
  Expected: an object with length of <1>
    Actual: []
     Which: has length of <0>
  
  package:matcher                                                                     expect
  test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart 804:9  main.<fn>.<fn>
  
+182 -2: test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart: Given an endpoint method with Stream parameter that returns a Future with nullable type when analyzed then endpoint method definition is a method stream definition. [E]
  Expected: <Instance of 'MethodStreamDefinition'>
    Actual: <null>
     Which: is not an instance of 'MethodStreamDefinition'
  
  package:matcher                                                                                                                                expect
  /home/runner/work/serverpod/serverpod/tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart 811:11  main.<fn>.<fn>.<fn>
  
+182 -3: test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart: Given an endpoint method with Stream parameter that returns a Future with nullable type when analyzed then endpoint method definition has future return type. [E]
  Expected: 'Future'
    Actual: <null>
     Which: not an <Instance of 'String'>
  
  package:matcher                                                                                                                                expect
  /home/runner/work/serverpod/serverpod/tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/endpoint_method_test.dart 817:11  main.<fn>.<fn>.<fn>
  
+3692 -3: Some tests failed.
```

After running the pipeline again, [3 of the 4 unit tests passed, but the Windows one failed again](https://github.com/serverpod/serverpod/actions/runs/19677527134/job/56364851372?pr=4276) with a similar issue:

```
==== Running tests in ./tools/serverpod_cli ====
+3158 -1: test\integration\analyzer\dart\endpoint_validation\endpoint_method_test.dart: Given a valid endpoint method with "@Deprecated(<string literal>)" annotation then endpoint definition is created. [E]
  Expected: an object with length of <1>
    Actual: []
     Which: has length of <0>
  
  package:matcher                                                                      expect
  test\integration\analyzer\dart\endpoint_validation\endpoint_method_test.dart 1079:9  main.<fn>.<fn>
  
+3158 -2: test\integration\analyzer\dart\endpoint_validation\endpoint_method_test.dart: Given a valid endpoint method with "@Deprecated(<string literal>)" annotation then endpoint definition method has expected annotations. [E]
  Expected: <1>
    Actual: <null>
  
  package:matcher                                                                      expect
  test\integration\analyzer\dart\endpoint_validation\endpoint_method_test.dart 1085:9  main.<fn>.<fn>
  
+3693 -2: Some tests failed.

Consider enabling the flag chain-stack-traces to receive more detailed exceptions.
For example, 'dart test --chain-stack-traces'.
```

Errors on the same test file, but not the same errors. Finally, running only the unit tests on Windows again passed.

It seems that files are being deleted before the tests end, possibly due to concurrent tests cleaning up the directory and leading to this flakiness. It was the first time that I encountered this problem, but we should investigate if it happen again on other PRs.

marcelomendoncasoares:
This PR is an attempt to fix #4301.

My current suspect is the concurrent execution of any CI step that runs `actions/checkout@v4` while the CLI tests are running. Since [the checkout step will run a `git clean -ffdx && git reset --hard HEAD`](https://github.com/actions/checkout#:~:text=%23%20Whether%20to%20execute%20%60git%20clean%20%2Dffdx%20%26%26%20git%20reset%20%2D%2Dhard%20HEAD%60%20before%20fetching%0A%20%20%20%20%23%20Default%3A%20true%0A%20%20%20%20clean%3A%20%27%27), this would discard the files created by the CLI tests. They are currently not even "gitignored", but this wouldn't matter due to the `x` `git clean` flag.

This fix moves the test artifacts location to the system temp, so it doesn't get removed by `git reset`.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/util/server_directory_finder.dart
+++ b/tools/serverpod_cli/lib/src/util/server_directory_finder.dart
@@ -133,7 +133,7 @@ Directory? _returnCandidates(List<Directory> candidates) {
   // Remove duplicates by normalized path
   var uniquePaths = <String>{};
   candidates = candidates.where((dir) {
-    var canonicalized = p.canonicalize(dir.path);
+    var canonicalized = p.canonicalize(dir.resolveSymbolicLinksSync());
     if (uniquePaths.contains(canonicalized)) {
       return false;
     }


marcelomendoncasoares:
Moving the test to the temp folder created an unexpected conflict with the generate from anywhere test:

```
Running unit tests in all project that contain a test directory

==== Running tests in ./tools/serverpod_cli ====
+179 -1: test/integration/generate/generate_from_anywhere_test.dart: GeneratorConfig.load with auto-detection Given in server subdirectory (lib/src/protocol), when GeneratorConfig.load is called without directory parameter, then it finds the server directory by searching upward [E]
  Multiple Serverpod projects detected:
    - /tmp/serverpod_anywhere_test_JAXNMV/myapp_server
    - /tmp/cli_test_KCUBZN
  Please navigate to one of these directories or use the --directory flag.
  package:serverpod_cli/src/util/server_directory_finder.dart 191:9  ServerDirectoryFinder.findOrPrompt
  package:serverpod_cli/src/config/config.dart 278:51                GeneratorConfig.load
  test/integration/generate/generate_from_anywhere_test.dart 151:46  main.<fn>.<fn>
  
+3709 -1: Some tests failed.
```

It seems that this test is expecting no other packages to existe on the entire temp folder, which is a fragile premise. We should also refactor this tests to make them all run into a root folder that has no conflicts.

SandPod:
I think the conflict would also be solved if we where to use the [test_descriptor](https://pub.dev/packages/test_descriptor) package as it creates a unique sandbox for each test.

marcelomendoncasoares:
> I think the conflict would also be solved if we where to use the [test_descriptor](https://pub.dev/packages/test_descriptor) package as it creates a unique sandbox for each test.

I tried to use the `test_descriptor` in this fix, but its call to `addTearDown` would require deeper changes on the CLI generator tests. These tests share one common environment created before running all groups and also run a delete on `tearDownAll`, so I saw no reason to completely refactor them just to use the package.

```dart
// Contents of the `d.sandbox` function
String get sandbox {
  if (_sandbox != null) return _sandbox!;
  // Resolve symlinks so we don't end up with inconsistent paths on Mac OS where
  // /tmp is symlinked.
  final sandbox = _sandbox = Directory.systemTemp
      .createTempSync('dart_test_')
      .resolveSymbolicLinksSync();

  addTearDown(() async {  // <-- The exact same as we do
    final sandbox = _sandbox!;
    _sandbox = null;
    await Directory(sandbox).delete(recursive: true);
  });

  return sandbox;
}
```

Then, I only changed the `generate_from_anywhere` tests to use the `test_descriptor` and kept the direct `systemTemp` approach with the CLI generator tests.

But I'm still facing a conflict that seems unexpected. When running generate from `lib/src/protocol` with the following folder structure, shouldn't the finder stop at the `myapp_server` level?

```
==== Running tests in ./tools/serverpod_cli ====
+189 -1: test/integration/generate/generate_from_anywhere_test.dart: GeneratorConfig.load with auto-detection Given in server subdirectory (lib/src/protocol), when GeneratorConfig.load is called without directory parameter, then it finds the server directory by searching upward [E]
  Multiple Serverpod projects detected:
    - /tmp/dart_test_SITNIK/myapp_server
    - /tmp/cli_test_FUAAMI
  Please navigate to one of these directories or use the --directory flag.
  package:serverpod_cli/src/util/server_directory_finder.dart 191:9  ServerDirectoryFinder.findOrPrompt
  package:serverpod_cli/src/config/config.dart 278:51                GeneratorConfig.load
  test/integration/generate/generate_from_anywhere_test.dart 138:46  main.<fn>.<fn>
```

marcelomendoncasoares:
Seems that now we have changed the tests that are flaky:

```
==== Running tests in ./tools/serverpod_cli ====
+3662 -1: test\util\server_directory_finder_interactive_test.dart: ServerDirectoryFinder.findOrPrompt with interactive flag Given no server directory exists, when findOrPrompt is called with interactive=false, then it throws ServerpodProjectNotFoundException [E]
  Expected: throws <Instance of 'ServerpodProjectNotFoundException'> with `message`: contains 'No Serverpod server project detected'
    Actual: <Closure: () => Future<Directory>>
     Which: returned a Future that emitted _Directory:<Directory: 'C:\Users\runneradmin\AppData\Local\Temp\cli_test_19ca2863'>
  
  package:test_api  OutstandingWork.complete
  
+3668 -2: test\util\server_directory_finder_search_test.dart: ServerDirectoryFinder.search Given no server directories exist, when search is called, then it returns notFound status [E]
  Expected: null
    Actual: _Directory:<Directory: 'C:\Users\runneradmin\AppData\Local\Temp\cli_test_19ca2863'>
  
  package:matcher                                           expect
  test\util\server_directory_finder_search_test.dart 161:9  main.<fn>.<fn>
  
+3670 -3: test\util\server_directory_finder_search_test.dart: ServerDirectoryFinder.search Given search depth is limited to 2, when server directory is 3 levels deep, then it does not find the server directory [E]
  Expected: null
    Actual: _Directory:<Directory: 'C:\Users\runneradmin\AppData\Local\Temp\cli_test_19ca2863'>
  
  package:matcher                                           expect
  test\util\server_directory_finder_search_test.dart 232:9  main.<fn>.<fn>
  
+3733 -3: Some tests failed.

Consider enabling the flag chain-stack-traces to receive more detailed exceptions.
For example, 'dart test --chain-stack-traces'.
```

In one run, only the first failed. In the last, the three above. The failure is just on Windows, but I can't seem to reproduce on my machine - either by running tests simultaneously or trying to isolate on a test.

Despite forbidding the directory searcher to go up towards the system temp, it seems that this is failing on Windows running in  the CI, leading to the search to find artifacts from concurrent tests.

```dart
      // Stop at first level within top-level directories to prevent
      // cross-contamination between concurrent or leftover of previous tests
      final topLevelDirectories = [
        p.canonicalize('/'),
        p.canonicalize(Directory.systemTemp.path),
      ];
      if (topLevelDirectories.contains(p.canonicalize(dir.parent.path))) {
        return true;
      }
```

marcelomendoncasoares:
After adding some (a ton of) debugging prints, I found the flakiness source of the directory search on Windows:

```
+3684: test\util\server_directory_finder_interactive_test.dart: ServerDirectoryFinder.findOrPrompt with interactive flag Given no server directory exists, when findOrPrompt is called with interactive=false, then it throws ServerpodProjectNotFoundException
[_isRepositoryBoundary] Parent is not a top-level directory (dir: c:\users\runneradmin\appdata\local\temp\dart_test_1f5ec89c, top-level directories: d:\, c:\users\runner~1\appdata\local\temp)
[_isRepositoryBoundary] ✗ Not a boundary
```

The `path.canonicalize` function should return the same path if the input point to the same location, as stated in the docs:

> Canonicalizes [path].
> 
> This is guaranteed to return the same path for two different input paths if and only if both input paths point to the same location. Unlike [normalize], it returns absolute paths when possible and canonicalizes ASCII case on Windows.
> 
> Note that this does not resolve symlinks.
> 
> If you want a map that uses path keys, it's probably more efficient to use a Map with [equals] and [hash] specified as the callbacks to use for keys than it is to canonicalize every key.

But it is resolving the temp folder to `c:\users\runner~1\appdata\local\temp`. The tricky issue is that this is not a symlink, but a "short filename" (also known as legacy DOS alias) that is probably being received from some env var, so we have no native tools to "expand" it.

Maybe we could work around this by trying to match:
1. If the path starts with what comes before the `~` (`c:\users\runner`)
2. If the path ends with what comes after the next `\` (`\appdata\local\temp`)

Then, if both are true (for Windows only), we treat as the same path - either for temp/root or for the user folder.

tools/serverpod_cli/test/integration/generate/generate_from_anywhere_test.dart
```diff
@@ -111,11 +100,10 @@ dependencies:
       'then it finds the sibling server directory',
       () async {
         await createMockServerpodProject(
-          parent: tempDir,
           projectName: 'myapp',
         );
 
-        var clientDir = Directory(path.join(tempDir.path, 'myapp_client'));
+        var clientDir = Directory(path.join(d.sandbox, 'myapp_client'));
```
nielsenko: Nit!: This is a bit of a weird mix I think - using `test_descriptor` package to get a path, but not to create files and directories?

I think using either `test_descriptor` fully, or using `Directory.systemTemp.createTempSync(prefix)` to create a test specific directory, is more idiomatic.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
```diff
@@ -321,12 +321,67 @@ class ServerDirectoryFinder {
...
+    final prefix = maybeParent.substring(0, tildeIndex);
+    final suffix = afterTilde.substring(nextSlashIndex);
+
+    // Check if this matches: starts with prefix and ends with suffix
+    final pattern = RegExp(
+      '^${RegExp.escape(prefix)}.*${RegExp.escape(suffix)}\$',
+      caseSensitive: false,
+    );
+    return pattern.hasMatch(this);
+  }
```
nielsenko: This seems a reasonable implementation to me, given the objective. However, I do not understand why we need to handle this?  Why do we see legacy path on CI?

marcelomendoncasoares: After moving the flaky endpoint generator tests to the system temp, we started having eventual conflicts with the generate from anywhere tests. While implementing a fix to never go up to the temp to search for other projects, we hit this legacy path comparison problem on Windows.

My guess is that at least some of the machines that are assigned to run our job on Windows in the CI have short path name (or 8.3 filename) enabled, leading to this problem.

nielsenko: But today (like since Windows95!!) the 8.3 name is always just an alias to the real name. It is possible to disable it (since Windows 7 I believe), but you always have the long name.

nielsenko: https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#short-vs-long-names
https://en.wikipedia.org/wiki/8.3_filename

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
```diff
@@ -321,12 +321,67 @@ class ServerDirectoryFinder {
       }
       var homeDir =
           Platform.environment['HOME'] ?? Platform.environment['USERPROFILE'];
-      if (homeDir != null && p.normalize(dir.path) == p.normalize(homeDir)) {
+      if (homeDir != null &&
+          p.canonicalize(dir.path) == p.canonicalize(homeDir)) {
```
nielsenko: I have a hunch you should use `p.equals` here instead of `String` compare. I don't think `canonicalize` resolves symlinks and stuff like 8.3 filename aliases.

nielsenko: Actually `p.equals` doesn't either.

Before comparing you need `FileSystemEntity.resolveSymbolicLinks` which uses [GetFinalPathNameByHandle](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfinalpathnamebyhandlea) on windows

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
nielsenko: You should think of 8.3 aliases as symlinks. They need to be resolved.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
```diff
@@ -321,12 +321,67 @@ class ServerDirectoryFinder {
...
         return true;
       }
+
+      // Stop at first level within top-level directories to prevent
+      // cross-contamination between concurrent or leftover of previous tests
+      final topLevelDirectories = {
+        // Root directory, that can be another drive on Windows (like "D:\")
+        p.rootPrefix(p.canonicalize('/')),
+        // Root directory of the current directory, in case it's in another drive
```
nielsenko: What do you expect this string manipulation to return? On windows it will just return `\` (ie. backslash instead of forward slash)

marcelomendoncasoares: This will return the drive of the root path. On the CI, for example, this returns `d:\`, even though the temp dir is under `c:\`. Then having both root prefixes ensures that we never navigate up to the root of the drive.

But thinking now, since we are only traversing upwards from the path, we may be able to suppress this `p.rootPrefix(p.canonicalize('/'))` and leave only the other `p.rootPrefix(p.canonicalize(dir.path))` - which will return `c:\` on the CI, as expected.

nielsenko: Actually, it doesn't. It will just return `\` (to my knowledge).

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
marcelomendoncasoares: Good find! I'll replace the hacky comparison with a `.resolveSymbolicLinks()` call before cannonicalization.

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
nielsenko: It would need to infer that you where thinking of the root of the current drive, and then consult CWD, and use something like [GetFinalPathNameByHandle](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfinalpathnamebyhandlea) to find the drive. 

I don't have a windows machine at hand right now, but I would be surprised if it did.

marcelomendoncasoares: > Actually, it doesn't. It will just return `\` (to my knowledge).

It is surprising indeed that it did return `d:\` (print above) and on my machine it also returns `c:\`. This is actually unclear seeing the documentation of the method, since there is no example of a path that starts with only one slash:

> Returns the root of [path], if it's absolute, or the empty string if it's relative.
> 
> // Unix
> p.rootPrefix('path/to/foo'); // -> ''
> p.rootPrefix('/path/to/foo'); // -> '/'
> 
> // Windows
> p.rootPrefix(r'path\to\foo'); // -> ''
> p.rootPrefix(r'C:\path\to\foo'); // -> r'C:\'
> p.rootPrefix(r'\\server\share\a\b'); // -> r'\\server\share'

nielsenko: TIL

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
```diff
@@ -321,9 +321,22 @@ class ServerDirectoryFinder {
       }
       var homeDir =
           Platform.environment['HOME'] ?? Platform.environment['USERPROFILE'];
-      if (homeDir != null && p.normalize(dir.path) == p.normalize(homeDir)) {
+
+      final parentPath = p.canonicalize(dir.parent.resolveSymbolicLinksSync());
```
nielsenko, line 325: I would suggest doing `final dirPath = dir. resolveSymbolicLinksSync()` very early (just after entry), and then don't use `dir.path` but `dirPath` throughout the method.

Also, I think you should _resolve_ `homeDir` as well. Might be 8.3 file alias as well.

marcelomendoncasoares: Good suggestion, I'll move the resolve to the top!

> Also, I think you should _resolve_ `homeDir` as well. Might be 8.3 file alias as well.

The home dir is being resolved right below before the comparison.

```dart
  final topLevelDirectories = {
    p.rootPrefix(parentPath),
    p.canonicalize(Directory.systemTemp.path),
    if (homeDir != null)
      p.canonicalize(Directory(homeDir).resolveSymbolicLinksSync()),
  };
  if (topLevelDirectories.any((d) => d == parentPath)) {
    return true;
  }
```

tools/serverpod_cli/lib/src/util/server_directory_finder.dart
```diff
@@ -321,9 +321,22 @@ class ServerDirectoryFinder {
...
+
+      // Stop at first level within top-level directories to prevent
+      // cross-contamination between concurrent or leftover of previous tests
+      // Paths must be resolved first to avoid legacy DOS Windows paths.
+      final topLevelDirectories = {
+        p.rootPrefix(parentPath),
+        p.canonicalize(Directory.systemTemp.path),
+        if (homeDir != null)
+          p.canonicalize(Directory(homeDir).resolveSymbolicLinksSync()),
```
nielsenko, line 334: Arhh 🤦 .. Anyway, I think it would be even clearer if it happens straight after reading the environment variables.

tools/serverpod_cli/test/integration/generate/generate_from_anywhere_test.dart
marcelomendoncasoares: You're right on this, I did a fast replacement and forgot to update all dir and file interactions. Standardized at https://github.com/serverpod/serverpod/pull/4327/commits/62047b11117b4310e8882bd4ff6b34aed6776422. It would be better if all tests used the `test_descriptor` package, but this would require a greater refactor that is not worth right now. So I only ensured that the ones that we changed to use are using it fully.

PR #4330, opened 2025-11-28
chore: Update relic to 0.11.0

nielsenko:
- **chore: Upgrade relic to 0.11.0**
- **feat: remoteInfo now fallback to connectionInfo.remote.address instead of 'unknown'**
- **fix: Update code after upgrade to relic 0.11.0**

### DIFF
--- a/packages/serverpod/lib/src/util/request_extension.dart
+++ b/packages/serverpod/lib/src/util/request_extension.dart
@@ -1,15 +1,28 @@
+import 'package:postgres/postgres.dart';
 import 'package:relic/relic.dart';
 
 /// Extends [Request] with useful methods.
 extension RequestExtension on Request {
   /// Returns information about the remote client that made the request.
   ///
-  /// The information is extracted from HTTP headers, preferring the
-  /// `Forwarded` header, and using the `X-Forwarded-For` header as a fallback.
-  /// If no information can be extracted, it returns 'unknown'.
+  /// The information is extracted from HTTP headers, preferring the `Forwarded`
+  /// header, then the `X-Forwarded-For` header, and using
+  /// [ConnectionInfo.remote.address] as fallback.
   String get remoteInfo {
-    return headers.forwarded?.elements.first.forwardedFor?.identifier ??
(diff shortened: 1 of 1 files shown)


packages/serverpod/lib/src/util/request_extension.dart
```diff
@@ -1,15 +1,27 @@
...
+    // node-id = ( IPv4address / "[" IPv6address "]" / "unknown" / obfuscated )
+    // obfuscated = "_" 1*( ALPHA / DIGIT / "." / "_" / "-" )
+    //
+    // Examaples: “192.0.2.43”, “[2001:db8::1]”, “unknown”, or “_08fusc4t3d"
+    final forwardedFor =
+        headers.forwarded?.elements.first.forwardedFor?.identifier;
+
+    return forwardedFor ??
+        headers.xForwardedFor?.addresses.first ?? // anything really ¯\_(ツ)_/¯
+        connectionInfo.remote.address.toString();
```
nielsenko, line 26: Fallback to use remote IPAddress of TCP connection in `remoteInfo`

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/util/session_extension.dart
```diff
@@ -4,9 +4,10 @@ import 'package:serverpod/serverpod.dart';
 extension SessionExtension on Session {
   /// Returns the client's IP address, or empty string in case it could not be
   /// determined.
-  String get remoteIpAddress {
+  IPAddress? get remoteIpAddress {
```
nielsenko, line 7: Now actually returns the remote IPAddress of the TCP connection. This extension is only used Email IDP and only for logging. I believe this should be changed, but let us discuss and do so in a separate PR:

SandPod:
@coderabbitai Please create an issue for the existence of the `remoteInfo` field on session that is only populated when a `WebCallSession` is created but not when a `MethodCallSession` is created.

PR #4339, opened 2025-12-01
feat: Allow inheriting indexes declared on base models

### LINKED ISSUE #4239: Support defining indexes on base class that automatically apply to child tables
### Problem to Solve

Currently, indexes in Serverpod need to be defined separately for each table, even if they derive from a common base class. This can lead to code duplication and potential inconsistencies when indexes are needed on base common fields for the child classes.

### Proposal

Introduce support for defining indexes on base classes so that these are automatically applied to all child tables inheriting from the base. This would streamline index management, promote DRY practices, and reduce the risk of errors. The name of the index in the base class would be a prefix to be applied to the table class name, so that each index receive a unique name.

### Use Case

Have a base class with audit fields and indexes on such fields so that audit queries can be speed up.

### Alternatives

Manually defining the same indexes repeatedly on all child classes. Alternatively, tooling could be built to help ensure index consistency, but that introduces additional steps and complexity.

### Additional context

This feature combined with inheritance on models will close https://github.com/serverpod/serverpod/issues/792.

- [ ] I want to work on this

marcelomendoncasoares:
Closes #4239.

## Pre-launch Checklist

## Breaking changes

None;

### DIFF
--- a/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
@@ -238,13 +238,14 @@ class Restrictions {
     SourceSpan? span,
   ) {
     return [
-      ..._validateTableName(parentNodeName, tableName, span),
+      ..._validateTableName(tableName, span),
       ..._validateTableInheritedIdField(span),
+      if (tableName is String)
+        ..._validateTableInheritedIndexNames(tableName, span),
     ];
   }
 
   List<SourceSpanSeverityException> _validateTableName(
-    String parentNodeName,
     dynamic tableName,
     SourceSpan? span,
   ) {
@@ -471,15 +472,28 @@ class Restrictions {
           ),
         ];
       }
-    }
 
-    if (indexName.length > DatabaseConstants.pgsqlMaxNameLimitation) {
-      return [
-        SourceSpanSeverityException(
-          'The index name "$indexName" exceeds the ${DatabaseConstants.pgsqlMaxNameLimitation} character index name limitation.',
-          span,
-        ),
-      ];
+      if (indexName.length > DatabaseConstants.pgsqlMaxNameLimitation) {
+        final baseMessage =
+            'The index name "$indexName" exceeds the ${DatabaseConstants.pgsqlMaxNameLimitation} character index name limitation.';
+
+        if (definition.inheritedIndexes.any((i) => i.name == indexName)) {
+          return [
+            SourceSpanSeverityException(
+              '$baseMessage '
+              'Consider shortening either the base index name or the table name. '
+              'Note that changing the base index name may require dropping and '
+              'recreating indexes for other tables that already extend the '
+              'model where it is defined.',
+              span,
+            ),
+          ];
+        }
+
+        return [
+          SourceSpanSeverityException(baseMessage, span),
+        ];
+      }
     }
 
     return [];
@@ -1036,6 +1050,20 @@ class Restrictions {
     }
 
     return errors;
+  }
+
+  List<SourceSpanSeverityException> _validateTableInheritedIndexNames(
+    String tableName,
+    SourceSpan? span,
+  ) {
+    var definition = documentDefinition;
+    if (definition is! ModelClassDefinition) return [];
+
+    var indexNames = definition.inheritedIndexes.map((i) => i.name);
+    return [
+      for (var indexName in indexNames)
+        ...validateTableIndexName(tableName, indexName, span),
+    ];
   }
 
   List<SourceSpanSeverityException> _validateIdFieldDataType(
--- a/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/indexes/index_inheritance_test.dart
+++ b/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/indexes/index_inheritance_test.dart
@@ -126,4 +126,70 @@ void main() {
       expect(inheritedIndex.fields, ['indexed']);
     },
   );
+
+  group(
+    'Given an inherited index whose name combined with the table name exceeds the 63 character index name limitation, then collect an error that the index name is too long.',
+    () {
+      var models = [
+        ModelSourceBuilder().withYaml(
+          '''
+class: ParentBase
+fields:
+  indexed: int
+indexes:
+  base_index_that_is_very_very_long:
+    fields: indexed
+          ''',
+        ).build(),
+        ModelSourceBuilder().withFileName('child_table').withYaml(
+          '''
+class: ChildTable
+table: child_table_with_a_very_long_name
+extends: ParentBase
+fields:
+  ownField: String
+          ''',
+        ).build(),
+      ];
+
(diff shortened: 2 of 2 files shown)


tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -364,6 +384,18 @@ class SerializableModelIndexDefinition {
 
   /// Whether the index is of vector type.
   bool get isVectorIndex => VectorIndexType.values.any((e) => e.name == type);
+
+  /// Copy the index with a new name that is prefixed with [prefix].
+  SerializableModelIndexDefinition copyWithPrefix(String prefix) {
+    return SerializableModelIndexDefinition(
+      name: '${prefix}_$name',
```
SandPod, line 391: Question: Any caution that needs to be taken in regard to the length of the index name? For example if an index is inherited through multiple levels of inheritance?

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
```diff
@@ -140,7 +140,7 @@ class ModelDependencyResolver {
     SerializableModelFieldDefinition fieldDefinition,
     ModelClassDefinition classDefinition,
   ) {
-    var indexes = classDefinition.indexes;
+    var indexes = classDefinition.indexesIncludingInherited;
```
SandPod, line 143: Question: What is the normal case? Wanting the indexes with or without inheritance? Perhaps we should have a specific getter for fetching the indexes without inheritance instead?

tests/serverpod_test_server/lib/src/models/inheritance/grandparent_with_id.spy.yaml
```diff
@@ -2,3 +2,8 @@ class: GrandparentClassWithId
 fields:
     id: UuidValue?, defaultPersist=random
     grandParentField: String
+
+indexes:
```
SandPod, line 6: Question: Is this allowed? I thought we had a restriction on indexes so that they are only applicable on classes with tables? Not sure if this makes sense out side of inheritance 🤔

We restrict a lot of other keywords that require a table, for example the `relation` keyword. Same logic could be applied there to allow these to be defined. Should we?

SandPod (commented):
Couple of questions for you

tests/serverpod_test_server/lib/src/models/inheritance/grandparent_with_id.spy.yaml
marcelomendoncasoares: It also surprised me that this is allowed when I first tried, given keys are very restricted to where they are expected. And I think it makes sense for most keys or if there is no valid use case for a given combination. However, now that we have inheritance, I think it might be interesting to relax restrictions on keys that are valid to use for child table models.

The `relation` keyword is an example of one that we can not allow, since the relation gets stored in the field - and could leak to the parent model with the current implementation. But `indexes` (and the upcoming `column`) would do no harm to allow on non-table models.

Another approach would be requiring an identifier to enable table features on models to be used as base, but I fail to see a strong reason to do so instead of just allowing the keys. Or, if we want to be restrictive without adding complexity for the users, collect all non-table models that has table properties to issue a warning/error in the end if they do not get extended by table models.

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
marcelomendoncasoares: I'd say the normal is to get with inheritance, but the `indexes` field is the representation of the model file. The pattern here follows the already existing `fields`/`fieldsIncludingInherited`, which is harder to invert because the base `ClassDefinition` does not have inheritance. Given the usage of these classes, I don't think we would have much gain doing all this refactor.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
marcelomendoncasoares: The number of levels of inheritance does not impact the name length, since it is only possible for a table model to inherit a non-table models. An index could propagate through N levels of inheritance that the index name would still be base prefix + table name.

However, for longer prefixes combined with long table names, it is possible to exceed the limit. Right now, the index creation will be forbidden by the generator, since there is a restriction that validates the index name length. This would force the user to change either the table name or the index prefix - which is not the ideal experience.

Given that the limit is quite high (63 chars) and no automatic solution is that much straight forward, I would consider to wait until someone actually have this problem for us to work on it.

SandPod: I think it is fine to take the pragmatic approach here only question is if the user will understand how they need to modify the table structures to solve this issue.

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
SandPod: The difference between this and `fields`/`fieldsIncludingInherited` is that the normal case, when this was implemented, was that you wanted the explicit fields of the model and the inherited fields was only included in a subsection of the code.

If it is easy for us to populate `indexes` including inherited indexedes that might be the way to go. But I leave the up to your decision.

tests/serverpod_test_server/lib/src/models/inheritance/grandparent_with_id.spy.yaml
SandPod: I know we implemented such a mechanism for interfaces, only enabling database keywords if the model that implemented it contained the table keyword.

In this case, we should perhaps require the subclass to implement the table keyword if we include db related keywords? Either way, this seems like a larger discussion and direction to take.

marcelomendoncasoares: @coderabbitai Please open an issue regarding this discussion on which should be the best direction regarding the usage of keys that are table related for models that have no table key but could be used as base for tables. Then we can move the discussion and take a unified approach.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
marcelomendoncasoares: I'll modify the error message that we show to the user for it to be informative enough when the naming is too long due to index inheritance.

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
marcelomendoncasoares: Since we already use the separate check now with the dedicated message when the name is too long, I'll prefer to leave with the same pattern as the `fields` - reflecting exactly the model file.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4339/commits/a0759543fe35a2db6c4792614aa3329470829662.

PR #4347, opened 2025-12-02
fix!: Fallback route cannot access session

### LINKED ISSUE #4344: Fallback route error with: `Bad state: ContextProperty value not found.`
### Describe the bug

When a fallback route is configured and it is triggered, it fails with the following error:

```bash
ERROR - 2025-12-01 22:38:02.369683
Method.get /hla/hla
Unhandled error in mounted handler.
Bad state: ContextProperty value not found. Property: Session. Ensure middleware has set this value for the request token.
package:relic/src/middleware/context_property.dart 52:10  ContextProperty.get
package:serverpod/src/web_server/web_server.dart 227:43   SessionEx.session
package:serverpod/src/web_server/web_server.dart 259:31   Route.call
package:relic/src/relic_server.dart 128:35                _RelicServer._handleRequest
```

### To reproduce

1. Create a new project from the template.
2. Modify the routes so that it looks like this:

```dart
  pod.webServer.addRoute(RootRoute(), '/');
  pod.webServer.addRoute(RootRoute(), '/index.html');
  pod.webServer.fallbackRoute = RootRoute();
```

3. Request a path that does not exist
4. Observe the issue.

### Expected behavior

Expects the fallback route to function without any issues.

### Library version

`3.0.0-rc.3`

### Platform information

MacOS.

### Additional context

_No response_

None

- [ ] I want to work on this

nielsenko:
- Defer session creation until actually needed using a `_Deferred` wrapper class
- Sessions are only created when `request.session` is awaited, avoiding unnecessary allocations
- Apply session and exception middleware to fallback route for consistent error handling
- Change `Request.session` return type from `Session` to `Future<Session>`

Fixes #4344

## Pre-launch Checklist

## Breaking changes

- `Request.session` now returns `Future<Session>` instead of `Session`
- `Request.sessionOrNull` now returns `Future<Session>?` instead of `Session?`

### DIFF
--- a/packages/serverpod/lib/src/web_server/web_server.dart
+++ b/packages/serverpod/lib/src/web_server/web_server.dart
@@ -193,7 +193,7 @@ class _SessionMiddleware extends MiddlewareObject {
       try {
         return await next(req);
       } finally {
-        await deferredSession.value?.close(); // only close if actually resolved
+        await deferredSession.ifInitiatedRun((s) => s.close());
       }
     };
   }
@@ -225,14 +225,23 @@ class _ReportExceptionMiddleware extends MiddlewareObject {
 
 class _Deferred<T> {
   final Future<T> Function() _futureFactory;
-  T? value;
-  Future<T>? _cachedFuture;
 
   _Deferred(this._futureFactory);
 
+  Future<T>? _cachedFuture;
   Future<T> get future async {
-    _cachedFuture ??= _futureFactory().then((v) => value = v);
+    _cachedFuture ??= _futureFactory();
     return _cachedFuture!;
+  }
+
+  bool get initiated => _cachedFuture != null;
+
+  Future<R?> ifInitiatedRun<R>(FutureOr<R> Function(T) action) async {
+    final future = _cachedFuture;
+    if (future != null) {
+      return action(await future);
+    }
+    return null;
   }
 }
 


tests/serverpod_test_server/test/middleware_test.dart
```diff
@@ -117,11 +117,12 @@ extension on Request {
 
 class _AuthMiddleware {
   Handler call(Handler next) {
-    return (ctx) async {
-      final info = await ctx.session.authenticated;
+    return (req) async {
+      final session = await req.session;
+      final info = await session.authenticated;
```
SandPod: Question: This is not an await anymore, right?

packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -68,7 +68,12 @@ class WebServer {
   /// a request.
   ///
   /// If not set, default behavior is to return "404 Not Found".
-  set fallbackRoute(Route route) => _app.fallback = route.asHandler;
+  set fallbackRoute(Route route) =>
```
SandPod, line 71: Suggestion: Doesn't seem like we actually had any tests on fallback route. Could you add one?

tests/serverpod_test_server/test/middleware_test.dart
nielsenko: Since this is middleware (not a handler - `Route`), then we don't get the session as a parameter, but need to retrieve it from the request. Hence the `await req.session;`.

nielsenko: But yes, `await session.authenticated;` is pointless. Funny that doesn't trigger a lint 🤔

nielsenko:
Rebased on main

packages/serverpod/lib/src/web_server/web_server.dart
nielsenko: Fixed in [e3e687a](https://github.com/serverpod/serverpod/pull/4347/commits/e3e687af93704e4c4de0671d2bbe11f071a3b9ad)

Note that I don't actually test that you can avoid session creation, by not touching `req.session`. I was a little lost at how to test this without accessing the private `_sessionPropterty` `ContextProperty` in _web_server.dart_

tests/serverpod_test_server/test/middleware_test.dart
nielsenko: Fixed

nielsenko:
As futures start as soon as they are constructed in Dart, it is not enough to not await them. Hence `_Deferred` has been re-worked to take a `futureFactory` in [a229b97](https://github.com/serverpod/serverpod/pull/4347/commits/a229b97997897239b312621b8169f5e9282eb841)

SandPod:
But it seems like they have broken some other functionality.

tests/serverpod_test_server/test/middleware_test.dart
SandPod: Leaving this comment open as a reminder that the CI should be green before we merge.

nielsenko:
@SandPod I have re-worked `_Deferred` a bit. Now I track if future has been constructed, not if it ran to completion. See [30cd35c](https://github.com/serverpod/serverpod/pull/4347/commits/30cd35ccf31c0b446905887cce2c78ecee9f87fe).

SandPod (approved):
👍

PR #4377, opened 2025-12-05
feat: Add SpaRoute and FlutterRoute

exaby73:
Closes https://github.com/serverpod/serverpod/issues/4372

## Pre-launch Checklist

## Summary by CodeRabbit

* **New Features**
  * Added middleware for automatic fallback routing based on response conditions.
  * Added middleware to inject WebAssembly security headers for multi-threading support.
  * Added routes for serving Flutter web applications with WASM support and customizable caching.
  * Added routes for serving single-page applications with client-side routing fallback.

* **Tests**
  * Added comprehensive integration tests for Flutter routes, SPA routes, middleware behavior, and fallback routing scenarios.

### DIFF
--- a/tests/serverpod_test_server/test_integration/web_server/flutter_route_test.dart
+++ b/tests/serverpod_test_server/test_integration/web_server/flutter_route_test.dart
@@ -4,14 +4,13 @@ import 'package:http/http.dart' as http;
 import 'package:path/path.dart' as path;
 import 'package:serverpod/serverpod.dart';
 import 'package:test/test.dart';
+import 'package:test_descriptor/test_descriptor.dart' as d;
 
 import 'package:serverpod_test_server/src/generated/protocol.dart';
 import 'package:serverpod_test_server/src/generated/endpoints.dart';
 
 void main() {
   late Directory webDir;
-  late File indexFile;
-  late File appJsFile;
   late http.Client client;
 
   final portZeroConfig = ServerConfig(
@@ -23,24 +22,17 @@ void main() {
 
   setUpAll(() async {
     client = http.Client();
-    webDir = Directory(
-      path.join(Directory.current.path, 'web', 'flutter_route_test'),
-    );
-    await webDir.create(recursive: true);
 
-    // Create test files
-    indexFile = File(path.join(webDir.path, 'index.html'));
-    await indexFile.writeAsString('<html><body>Flutter App</body></html>');
+    await d.dir('web', [
+      d.file('index.html', '<html><body>Flutter App</body></html>'),
+      d.file('main.dart.js', '// Flutter web JS'),
+    ]).create();
 
-    appJsFile = File(path.join(webDir.path, 'main.dart.js'));
-    await appJsFile.writeAsString('// Flutter web JS');
+    webDir = Directory(path.join(d.sandbox, 'web'));
   });
 
   tearDownAll(() async {
     client.close();
-    if (await webDir.exists()) {
-      await webDir.delete(recursive: true);
-    }
   });
 
   group('Given a web server with FlutterRoute', () {
@@ -134,10 +126,10 @@ void main() {
         ),
       );
 
+      await d
+          .file('web/custom.html', '<html><body>Custom Flutter</body></html>')
+          .create();
       customIndex = File(path.join(webDir.path, 'custom.html'));
-      await customIndex.writeAsString(
-        '<html><body>Custom Flutter</body></html>',
-      );
 
       pod.webServer.addRoute(
         FlutterRoute(webDir, indexFile: customIndex),
--- a/tests/serverpod_test_server/test_integration/web_server/spa_fallback_test.dart
+++ b/tests/serverpod_test_server/test_integration/web_server/spa_fallback_test.dart
@@ -4,6 +4,7 @@ import 'package:http/http.dart' as http;
 import 'package:path/path.dart' as path;
 import 'package:serverpod/serverpod.dart';
 import 'package:test/test.dart';
+import 'package:test_descriptor/test_descriptor.dart' as d;
 
 import 'package:serverpod_test_server/src/generated/protocol.dart';
 import 'package:serverpod_test_server/src/generated/endpoints.dart';
@@ -11,7 +12,6 @@ import 'package:serverpod_test_server/src/generated/endpoints.dart';
 void main() {
   late Directory webDir;
   late File indexFile;
-  late File appJsFile;
   late http.Client client;
 
   final portZeroConfig = ServerConfig(
@@ -23,22 +23,18 @@ void main() {
 
   setUpAll(() async {
     client = http.Client();
-    webDir = Directory(path.join(Directory.current.path, 'web', 'spa_test'));
-    await webDir.create(recursive: true);
 
-    // Create test files
+    await d.dir('web', [
+      d.file('index.html', '<html><body>SPA Index</body></html>'),
+      d.file('app.js', 'console.log("app");'),
+    ]).create();
+
+    webDir = Directory(path.join(d.sandbox, 'web'));
     indexFile = File(path.join(webDir.path, 'index.html'));
-    await indexFile.writeAsString('<html><body>SPA Index</body></html>');
-
-    appJsFile = File(path.join(webDir.path, 'app.js'));
-    await appJsFile.writeAsString('console.log("app");');
   });
 
   tearDownAll(() async {
     client.close();
-    if (await webDir.exists()) {
-      await webDir.delete(recursive: true);
-    }
   });
 
   group(
@@ -58,7 +54,6 @@ void main() {
           ),
         );
 
-        // Add fallback middleware and static route
         pod.webServer.addMiddleware(
           FallbackMiddleware(
             fallback: StaticRoute.file(indexFile),
@@ -131,7 +126,6 @@ void main() {
         ),
       );
 
-      // Use a custom route that returns 404 as the fallback
       pod.webServer.addMiddleware(
         FallbackMiddleware(
           fallback: _NotFoundRoute(),
@@ -174,15 +168,13 @@ void main() {
         ),
       );
 
-      // Create admin SPA directory
-      adminDir = Directory(
-        path.join(Directory.current.path, 'web', 'admin_test'),
-      );
-      await adminDir.create(recursive: true);
-      adminIndexFile = File(path.join(adminDir.path, 'index.html'));
-      await adminIndexFile.writeAsString('<html><body>Admin SPA</body></html>');
+      await d.dir('admin', [
+        d.file('index.html', '<html><body>Admin SPA</body></html>'),
+      ]).create();
 
-      // Admin SPA
+      adminDir = Directory(path.join(d.sandbox, 'admin'));
+      adminIndexFile = File(path.join(adminDir.path, 'index.html'));
+
       pod.webServer.addMiddleware(
         FallbackMiddleware(
           fallback: StaticRoute.file(adminIndexFile),
@@ -195,7 +187,6 @@ void main() {
         '/admin/**',
       );
 
-      // Public SPA
       pod.webServer.addMiddleware(
         FallbackMiddleware(
           fallback: StaticRoute.file(indexFile),
@@ -211,9 +202,6 @@ void main() {
 
     tearDown(() async {
       await pod.shutdown(exitProcess: false);
-      if (await adminDir.exists()) {
-        await adminDir.delete(recursive: true);
-      }
     });
 
     test('when requesting admin path then admin index is served', () async {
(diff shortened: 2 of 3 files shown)


exaby73:
@coderabbitai full review

packages/serverpod/lib/src/web_server/middleware/wasm_headers_middleware.dart
```diff
@@ -0,0 +1,57 @@
...
+/// ```dart
+/// pod.webServer.addMiddleware(WasmHeadersMiddleware(), '/app/**');
+/// ```
+///
+/// ## How It Works
+///
+/// The middleware intercepts all Response objects and adds COOP/COEP headers.
+/// Other Result types (Hijack, WebSocketUpgrade) pass through unchanged.
+/// Existing headers are preserved.
+class WasmHeadersMiddleware extends MiddlewareObject {
```
nielsenko, line 35: Nit. Could consider making this just a `wasmHeaders` function, as it doesn't take any configuration, ie. `Handler wasmHeaders(Handler next) { ... ` or `addWasmHeaders` or `wasmHeadersMiddleware`.

packages/serverpod/lib/src/web_server/routes/flutter_route.dart
```diff
@@ -0,0 +1,112 @@
...
+  FlutterRoute(
+    this.directory, {
+    this.indexFile,
+    this.cacheControlFactory,
+    this.cacheBustingConfig,
+  }) : super(methods: {Method.get, Method.head});
+
+  @override
+  void injectIn(RelicRouter router) {
+    final subRouter = Router<Handler>();
```
nielsenko, line 48: I wonder why you choose to create a `subRouter` that is _attached_ at the end, instead of just working with the `router` directly. 

One reason might be to ensure, that if an exception is raised, then no changes are done to the `router`, ie. `injectIn` is all or nothing, but I don't see how the current set of operations can reasonably fail in a way that should allow the setup to continue.

Anyway, it gave me reason to pause. Should we prefer this "transactional" behavior in general, and let `Router.inject` handle it transparently? 🤔

packages/serverpod/lib/src/web_server/routes/spa_route.dart
```diff
@@ -0,0 +1,85 @@
...
+///
+/// ## How It Works
+///
+/// 1. Creates a sub-router for the route path
+/// 2. Applies fallbackMiddleware to intercept 404 responses
+/// 3. Injects the static route into the sub-router
+/// 4. Attaches the sub-router to the main router
+///
+/// This ensures all 404s from the static route fall back to the specified file.
+class SpaRoute extends Route {
```
nielsenko, line 20: Could `FlutterRoute` not just build on this, and add the wasm headers?

tests/serverpod_test_server/test_integration/web_server/flutter_route_test.dart
```diff
@@ -0,0 +1,229 @@
...
+    setUp(() async {
+      pod = Serverpod(
+        [],
+        Protocol(),
+        Endpoints(),
+        config: ServerpodConfig(
+          apiServer: portZeroConfig,
+          webServer: portZeroConfig,
+        ),
+      );
```
nielsenko, line 127: Could this _setup_ be shared between the groups? (Same issue repeats)

packages/serverpod/lib/src/web_server/routes/flutter_route.dart
```diff
@@ -0,0 +1,112 @@
...
+    router.attach('/', subRouter);
+  }
+
+  @override
+  FutureOr<Result> handleCall(Session session, Request request) {
+    // This should never be called since routing is handled via injectIn
+    throw UnimplementedError(
+      'FlutterRoute handles routing via injectIn and should not be called directly',
+    );
+  }
```
nielsenko, line 77: You could also have composed it here, instead of in `injectIn`, ie. just wrapped `StaticRoute.directory` in `WasmHeadersMiddleware` and `FallbackMiddleware`, but this is just as fine.

I have been thinking if we should make a default implementation for `handleCall` as well, so you avoid this forced overwrite that will never be called. That way you can choose to either implement `handleCall`, `call`, or `injectIn` for progressively more control.

tests/serverpod_test_server/test_integration/web_server/flutter_route_test.dart
```diff
@@ -0,0 +1,229 @@
...
+  late Directory webDir;
+  late File indexFile;
+  late File appJsFile;
+  late http.Client client;
+
+  setUpAll(() async {
+    client = http.Client();
+    webDir = Directory(
+      path.join(Directory.current.path, 'web', 'flutter_route_test'),
+    );
```
nielsenko: Could we use `test_descriptor` package instead? Or at least move the `webDir` to tmp (Ie. `    Directory.systemTemp.createTemp(path.join('web', 'flutter_route_test'))`

packages/serverpod/lib/src/web_server/middleware/wasm_headers_middleware.dart
```diff
@@ -0,0 +1,57 @@
...
+  Handler call(Handler next) {
+    return (Request req) async {
+      final result = await next(req);
+
+      // Only modify Response objects
+      if (result is Response) {
+        return result.copyWith(
+          headers: result.headers.transform((mh) {
+            mh['Cross-Origin-Opener-Policy'] = ['same-origin'];
+            mh['Cross-Origin-Embedder-Policy'] = ['require-corp'];
```
nielsenko: Why not use the typed interface?

packages/serverpod/lib/src/web_server/routes/flutter_route.dart
```diff
@@ -0,0 +1,112 @@
...
+/// );
+/// ```
+///
+/// ## How It Works
+///
+/// 1. Creates a sub-router for the route path
+/// 2. Applies wasmHeadersMiddleware for Flutter WASM multi-threading
+/// 3. Applies fallbackMiddleware to serve index.html on 404
+/// 4. Injects static directory route
+/// 5. Attaches the sub-router to the main router
```
nielsenko: This seems a bit excessive. Probably residue from the AI tooling.

packages/serverpod/lib/src/web_server/routes/spa_route.dart
```diff
@@ -0,0 +1,85 @@
...
+/// ```
+///
+/// ## How It Works
+///
+/// 1. Creates a sub-router for the route path
+/// 2. Applies fallbackMiddleware to intercept 404 responses
+/// 3. Injects the static route into the sub-router
+/// 4. Attaches the sub-router to the main router
+///
+/// This ensures all 404s from the static route fall back to the specified file.
```
nielsenko: Should we explain implementation like this in doc comments?

packages/serverpod/lib/src/web_server/routes/spa_route.dart
```diff
@@ -0,0 +1,85 @@
...
+
+  /// Cache busting configuration for static files
+  final CacheBustingConfig? cacheBustingConfig;
+
+  /// Creates a new SpaRoute.
+  ///
+  /// - [directory] is the root directory containing static files
+  /// - [fallback] is the file to serve when requested files don't exist
+  /// - [cacheControlFactory] customizes cache headers for static assets
+  /// - [cacheBustingConfig] enables cache busting for assets
```
nielsenko: Use prose to explain parameters, return values, and exceptions (See [effective dart ](https://dart.dev/effective-dart/documentation#do-use-prose-to-explain-parameters-return-values-and-exceptions))

packages/serverpod/lib/src/web_server/routes/flutter_route.dart
```diff
@@ -0,0 +1,112 @@
...
+
+  /// Cache busting configuration for static files
+  final CacheBustingConfig? cacheBustingConfig;
+
+  /// Creates a new FlutterRoute.
+  ///
+  /// - [directory] is the root directory containing Flutter web files
+  /// - [indexFile] is the fallback file (defaults to [directory]/index.html)
+  /// - [cacheControlFactory] customizes cache headers for static assets
+  /// - [cacheBustingConfig] enables cache busting for assets
```
nielsenko: https://dart.dev/effective-dart/documentation#do-use-prose-to-explain-parameters-return-values-and-exceptions

packages/serverpod/lib/src/web_server/routes/flutter_route.dart
exaby73: This was a suggestion from @SandPod. I like this "transactional" behavior but I don't know why 🙂 I'm indifferent to it honestly

packages/serverpod/lib/src/web_server/routes/spa_route.dart
exaby73: The code became a bit messy if I wanted to keep the sub router so I opted for a little duplication

tests/serverpod_test_server/test_integration/web_server/flutter_route_test.dart
exaby73: I chose to keep it separate due to each group having it's own routes, and I didn't want to pollute other groups with routes it didn't configure

packages/serverpod/lib/src/web_server/routes/flutter_route.dart
exaby73: I thought about that but then I'd have to call `StaticRoute.directory().handleCall` which felt weird to do so and `injectIn` felt the best place to set up routing related stuff like middlewares instead of in `handleCall` which sounds like the better place for serving content, not setting up routes

tests/serverpod_test_server/test_integration/web_server/flutter_route_test.dart
exaby73: Didn't think about it here, but looking at other tests, they follow the same pattern. Not sure if we should migrate some tests and leave others in an old state, never to be touched. Maybe this should be a tech debt task we take on later to migrate sections of tests at once. Thoughts?

packages/serverpod/lib/src/web_server/middleware/wasm_headers_middleware.dart
exaby73: Slipped my mind :D Fixed in d68c4c9dd771

packages/serverpod/lib/src/web_server/routes/flutter_route.dart
exaby73: I actually intentionally added this to give an explanation on how it works since it's not obvious what it actually does. Should we keep this sort of comments?

exaby73: Actually rereading it and the accompanying constructor docs, It's unnecessary. Removed in 1476bdb10c9b

packages/serverpod/lib/src/web_server/routes/spa_route.dart
exaby73: Removed in 1476bdb10c9b

packages/serverpod/lib/src/web_server/routes/flutter_route.dart
exaby73: FIxed in 1476bdb10c9b

packages/serverpod/lib/src/web_server/middleware/wasm_headers_middleware.dart
exaby73: Given `MiddlewareObject` existed, I chose to reuse that. Should we deprecate `MiddlewareObject` or is this just a styling suggestion?

nielsenko: Just styling. I usually use the object version when I have state to track in the object, which is not the case here. I'm fine with keeping as is. I wish we could disable [implicit_call_tearoffs](https://dart.dev/tools/linter-rules/implicit_call_tearoffs) lint 😊

packages/serverpod/lib/src/web_server/routes/flutter_route.dart
nielsenko: Lets keep it. I'll look into adding it adding it on the relic side.

tests/serverpod_test_server/test_integration/web_server/flutter_route_test.dart
nielsenko: I'm thinking of the pod

packages/serverpod/lib/src/web_server/routes/spa_route.dart
nielsenko: Okay. Not critical.

tests/serverpod_test_server/test_integration/web_server/flutter_route_test.dart
nielsenko: What about using system temp. To ensure we don't litter the workspace when running locally.

PR #4382, opened 2025-12-08
chore: Upgrade serverpod to relic 0.13.0

nielsenko:
Bump relic to ^0.13.0 and use `Request.readAsString` for body reading. 

Also fixes file upload by awaiting both pipe and send concurrently, then draining the response.  This is an alternative (/better) fix for https://github.com/serverpod/serverpod/issues/3213 that avoid draining the too large request server side.

## Pre-launch Checklist

Fixes: #3213 (again, but client-side)

## Breaking changes

None.

## Release Notes

* **Bug Fixes**
  * Improved handling and error reporting for oversized HTTP request bodies across the server.
  * Enhanced file upload reliability through improved concurrent stream and request handling with proper cleanup.

* **Chores**
  * Updated relic dependency from 0.12.0 to 0.13.0 across packages and templates.

### DIFF
--- a/packages/serverpod/lib/src/server/server.dart
+++ b/packages/serverpod/lib/src/server/server.dart
@@ -203,10 +203,12 @@ class Server implements RouterInjectable {
     return (req) async {
       try {
         return await next(req);
-      } on _RequestTooLargeException catch (e) {
+      } on MaxBodySizeExceeded catch (e) {
         return Response(
           io.HttpStatus.requestEntityTooLarge,
-          body: Body.fromString(e.errorDescription),
+          body: Body.fromString(
+            'Request size exceeds the maximum allowed size of ${e.maxLength} bytes.',
+          ),
         );
       } on EndpointDispatchException catch (e) {
         return switch (e) {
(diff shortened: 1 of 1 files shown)


packages/serverpod_client/lib/src/file_uploader.dart
```diff
@@ -49,9 +49,11 @@ class FileUploader {
...
           });
           request.contentLength = length;
-          unawaited(stream.pipe(request.sink));
-
-          var response = await request.send();
+          final (_, response) = await (
+            stream.pipe(request.sink),
+            request.send(),
+          ).wait;
+          await response.stream.drain();
```
nielsenko, line 56: This is the crux of the alternative fix

nielsenko:
@coderabbitai review

packages/serverpod_client/lib/src/file_uploader.dart
SandPod: Question: this seems strange.

To me it looks like we are calling send at the same time as we are starting the read. The old code represented this dependency more explicitly. Why does this new code work as expected?

nielsenko: We didn't await `stream.pipe(request.sink)` previously. With the `().wait` it is now trivial to await both. As you know, leaving un-awaited futures across async gaps is not kosher.

The crux of the fix is call to `drain`.

nielsenko: And no, it is `stream.pipe` and `request.send` that are run concurrently. It did before as well, but we never awaited the pipe (to avoid needless buffering). I know I added the previous _unawaited_, but I'm not a fan of that pattern. What if the pipe was not guarenteed to make progress, unless it was actually awaited.

PR #4386, opened 2025-12-09
refactor: Minor UX improvements to the new auth

marcelomendoncasoares:
Minor improvements realized while writing the docs for the new auth.

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/email/forms/registration_start.dart
+++ b/modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/email/forms/registration_start.dart
@@ -42,7 +42,7 @@ class StartRegistrationForm extends StatelessWidget {
     final onPrivacyPolicyPressed = this.onPrivacyPolicyPressed;
 
     final hasTermsAndConditions =
-        onTermsAndConditionsPressed != null && onPrivacyPolicyPressed != null;
+        onTermsAndConditionsPressed != null || onPrivacyPolicyPressed != null;
 
     return FormStandardLayout(
       title: 'Sign Up with email',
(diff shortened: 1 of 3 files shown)


modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/email/forms/registration_start.dart
```diff
@@ -42,7 +42,7 @@ class StartRegistrationForm extends StatelessWidget {
     final onPrivacyPolicyPressed = this.onPrivacyPolicyPressed;
 
     final hasTermsAndConditions =
-        onTermsAndConditionsPressed != null && onPrivacyPolicyPressed != null;
+        onTermsAndConditionsPressed != null || onPrivacyPolicyPressed != null;
```
nielsenko, line 45: Is this intentional? Quite a significant change in semantics..

nielsenko (commented):
- There is switch from `&& `to `||` that I would like you to confirm. 
- I like the new `VerificationCodeConfig` container.
- The rest seems less important, but I'm okay with it.

modules/new_serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/email/forms/registration_start.dart
marcelomendoncasoares: It is intentional. @SandPod questioned [if it was possible to configure using only one or the other](https://github.com/serverpod/serverpod_docs/pull/360#discussion_r2600508455), so I made it possible with this commit. It is still up for discussion whether we want it or not, since having separate legal docs is the strongly recommended approach - as we were enforcing before. Then, if we consider better to not allow it, we just drop the commit.

SandPod: Tested it out in Storybook and it works as expected, let's keep it in as a ToS is not legally required.

SandPod (approved):
👍

SandPod:
Merging this as I need the export fixes for another PR.

PR #4407, opened 2025-12-10
feat: Add pub workspace support to project and module templates on new project creation

exaby73:
Closes https://github.com/serverpod/serverpod/issues/3193

## Pre-launch Checklist


tools/serverpod_cli/lib/src/create/create.dart
```diff
@@ -733,7 +758,7 @@ void _copyServerTemplates(
         replacement: '.gitignore',
       ),
     ],
-    ignoreFileNames: ['pubspec.lock'],
+    ignoreFileNames: ['pubspec.lock', 'pubspec_overrides.yaml'],
```
exaby73, line 852: This was causing issues running tests locally due to `melos` generated `pubspec_overrides.yaml` files

tests/bootstrap_project/test/module_test.dart
```diff
@@ -287,6 +287,57 @@ void main() {
...
+      group('then the workspace', () {
+        test('has a root pubspec.yaml file', () {
+          expect(
+            File(path.join(tempPath, projectName, 'pubspec.yaml')).existsSync(),
+            isTrue,
+            reason: 'Root workspace pubspec file does not exist.',
+          );
+        });
+
+        test('root pubspec.yaml has name: _', () {
```
SandPod, line 300: Question: This seems a bit odd, why should the name be `_`? What does this mean? Shouldn't the name be the same as the project name but without the `_server`etc suffix?

tests/bootstrap_project/test/module_test.dart
```diff
@@ -287,6 +287,57 @@ void main() {
           reason: 'Client protocol client file does not exist.',
         );
       });
+
+      group('then the workspace', () {
```
SandPod, line 291: Question: Have you validated that we correctly handle modules added that are workspaces? Or do we have an automated test that validates this?

Could be good to ensure that we don't have a dependency solely on the pubspec file in the module for something.

tests/serverpod_cli_e2e_test/test/generate_from_anywhere_test.dart
```diff
@@ -10,24 +10,13 @@ void main() async {
   late Directory tempDir;
   var rootPath = path.join(Directory.current.path, '..', '..');
   var cliPath = path.join(rootPath, 'tools', 'serverpod_cli');
+  var cliEntrypoint = path.join(cliPath, 'bin', 'serverpod_cli.dart');
 
   setUpAll(() async {
+    // Run pub get to ensure dependencies are up to date
```
SandPod: Question: Why was this modified? Doesn't seem to do the same action as before now.

SandPod: Since these are e2e tests, I believe we should perform the same operations that our end users are doing.

tools/serverpod_cli/lib/src/create/copier.dart
```diff
@@ -83,6 +91,28 @@ class Copier {
...
+    var lines = str.split('\n');
+    var processedLines = <String>[];
+
+    for (var line in lines) {
+      var processedLine = line;
+      if (line.contains(_uncommentMarker)) {
+        processedLine = processedLine
+            .replaceAll(_uncommentMarker, '')
+            .trimRight();
+        // Remove leading # while preserving indentation
```
SandPod: Question: What happens If we have `####`? 

Perhaps we should capture all leading `#`?

tools/serverpod_cli/lib/src/create/copier.dart
```diff
@@ -15,13 +17,16 @@ class Copier {
...
+  bool processUncommentMarker;
+
   Copier({
     required this.srcDir,
     required this.dstDir,
     required this.replacements,
     required this.fileNameReplacements,
     this.removePatterns = const <String>[],
     this.ignoreFileNames = const <String>[],
+    this.processUncommentMarker = false,
```
SandPod: Suggestion: Should we flip the default here to minimize the footprint of the change? Seems like we normally want to process it.

tests/bootstrap_project/test/module_test.dart
exaby73: That's the Dart convention for monorepo. See https://dart.dev/tools/pub/workspaces

tests/serverpod_cli_e2e_test/test/generate_from_anywhere_test.dart
exaby73: The issue is with dev time cache bug that exists in dart. And before we were not exactly doing what our users would which is activating from hosted. `dart run` and `dart pub global activate --source path` do the same thing, with the difference being that the former is perfectly reliable for running tests locally.

tools/serverpod_cli/lib/src/create/copier.dart
exaby73: Fixed in 41ab53f3faea

tools/serverpod_cli/lib/src/create/copier.dart
exaby73: Changed in 0798e899997f

tests/bootstrap_project/test/module_test.dart
exaby73: Dart pub workspaces do not allow nested workspaces. But in a workspace, you can have modules and they resolve correctly. I have manually tested by adding a module to the workspace of a new workspace project and successfully was able to create references in tables to a module's tables

exaby73: While testing this, I also ran into https://github.com/serverpod/serverpod/issues/4430

SandPod:
Let's not merge this until https://github.com/serverpod/serverpod/issues/4430 is resolved.

tests/serverpod_cli_e2e_test/test/generate_from_anywhere_test.dart
marcelomendoncasoares: @exaby73 Just one question on this one before we can merge. What exactly was the issue that made this change required? The ideal would be to make it work the way it was before.

exaby73: `global activate` would not update the CLI so dev time, the tests would not work with changes in the CLI. Kasper added a fix if I remember correctly to precompile the CLI then run the tests

marcelomendoncasoares: But was the issue for local development only or was it impacting the CI? If it was only on local dev, then it is best to revert it to be as close as possible to what users have.

PR #4416, opened 2025-12-10
feat: Add serverpod run command for running scripts

### LINKED ISSUE #4402: feat(cli): Add `serverpod run` command to execute scripts from pubspec.yaml
### Problem to Solve

Currently, there is no convenient built-in way to execute project-specific scripts directly through the serverpod CLI. This limits automation and makes it harder to run workflows like starting a Flutter client or executing custom operations defined in a project's configuration.

### Proposal

Introduce a new `serverpod run <script>` command in the serverpod CLI. This command will look for a `serverpod_scripts` section in the project's pubspec.yaml and execute the corresponding script value. For example:

```yaml
serverpod_scripts:
  flutter_run: flutter run -d chrome ../<my_project>/.
```

Running `serverpod run flutter_run` would execute the specified command. This enables teams to easily integrate and automate cross-tool behaviors within their workflow, managed through pubspec.yaml.

### Use Case

A developer can add frequently-used scripts to the pubspec.yaml and trigger them via `serverpod run <script>`. For example, starting the frontend and backend simultaneously for development, database migrations, or custom checks—without manually hunting for the correct shell command.

### Alternatives

- Manually running scripts from the terminal, which reduces consistency and automation
- Using other CLI wrappers, but these do not integrate with Serverpod's workflow
- Custom bash scripts, which are less discoverable and not linked to Serverpod's project structure

### Additional context

This will enhance developer experience and enable complex integration scenarios involving Flutter, backend, and third-party tools—while keeping workflow management in a single configuration file

- [ ] I want to work on this

nielsenko:
## Description

Adds `serverpod run` command to execute scripts defined in `serverpod_scripts` section of pubspec.yaml.

- New `RunCommand` with `--list` flag to show available scripts
- `Scripts` class parses `serverpod_scripts` from pubspec.yaml with source span error messages
- Uses `ServerDirectoryFinder` to locate the server project's pubspec
- Extends `UnmodifiableMapBase` for readonly map semantics

Fixes #4402

## Pre-launch Checklist

## Breaking changes

None.

## Summary by CodeRabbit

* **New Features**
  * Added a `run` command to the CLI for executing custom scripts defined in your project configuration
  * Users can list all available scripts or execute a specific script directly from the command line
  * Provides comprehensive error handling for missing scripts, invalid script definitions, and configuration issues with helpful error messages

### DIFF
--- a/tests/serverpod_cli_e2e_test/test/run_command_test.dart
+++ b/tests/serverpod_cli_e2e_test/test/run_command_test.dart
@@ -152,23 +152,32 @@ serverpod:
     },
   );
 
-  test(
-    'when running a non-existent script, then an error is shown',
-    () async {
-      final result = await runServerpod(
+  group('when running a non-existent script', () {
+    late ProcessResult result;
+
+    setUpAll(() async {
+      result = await runServerpod(
         ['run', 'nonexistent'],
         workingDirectory: serverDir,
       );
+    });
 
+    test('then the exit code is non-zero', () {
       expect(result.exitCode, isNot(0));
+    });
+
+    test('then an error message is shown', () {
       final output = '${result.stdout}${result.stderr}';
       expect(output, contains('Script "nonexistent" not found'));
+    });
+
+    test('then all available scripts are listed', () {
       expect(
         result.stdout,
         containsLines(['hello:', 'fail:', 'start:', 'test:']),
       );
-    },
-  );
+    });
+  });
 
   test(
     'when running without script name, then available scripts are listed',
@@ -219,4 +228,60 @@ serverpod:
     },
     timeout: Timeout(const Duration(minutes: 3)), // allow time to compile
   );
+
+  group('given pubspec without serverpod namespace', () {
+    late String serverDirNoNamespace;
+
+    setUpAll(() async {
+      await d.dir('test_server_no_namespace', [
+        d.file('pubspec.yaml', '''
+name: test_server_no_namespace
+environment:
+  sdk: ^3.0.0
+dependencies:
+  serverpod: any
+'''),
+      ]).create();
+      serverDirNoNamespace = p.join(d.sandbox, 'test_server_no_namespace');
+    });
+
+    test(
+      'when running a script, then an error about no scripts is shown',
+      () async {
+        final result = await runServerpod(
+          ['run', 'hello'],
+          workingDirectory: serverDirNoNamespace,
+        );
+
+        expect(result.exitCode, isNot(0));
+        expect(result.stdout, contains('No scripts defined'));
+      },
+    );
+
+    test(
+      'when running with --list flag, then an error about no scripts is shown',
+      () async {
+        final result = await runServerpod(
+          ['run', '--list'],
+          workingDirectory: serverDirNoNamespace,
+        );
+
+        expect(result.exitCode, isNot(0));
+        expect(result.stdout, contains('No scripts defined'));
+      },
+    );
+
+    test(
+      'when running without script name, then an error about no scripts is shown',
+      () async {
+        final result = await runServerpod(
+          ['run'],
+          workingDirectory: serverDirNoNamespace,
+        );
+
+        expect(result.exitCode, isNot(0));
+        expect(result.stdout, contains('No scripts defined'));
+      },
+    );
+  });
 }


nielsenko:
@coderabbitai review

tools/serverpod_cli/lib/src/commands/run.dart
```diff
@@ -0,0 +1,133 @@
...
+    final stdoutFuture = process.stdout
+        .transform(utf8.decoder)
+        .forEach((data) => stdout.write(data));
+    final stderrFuture = process.stderr
+        .transform(utf8.decoder)
+        .forEach((data) => stderr.write(data));
+
+    await [stdoutFuture, stderrFuture].wait;
+    return process.exitCode;
+  }
```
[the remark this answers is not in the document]
nielsenko: Fixed in [47f23e8](https://github.com/serverpod/serverpod/pull/4416/commits/47f23e8f5b7d2921ebe16ad21e67b1b577176c9d)

Along with more serious problems

tools/serverpod_cli/test/scripts/scripts_test.dart
```diff
@@ -0,0 +1,393 @@
...
+
+import 'package:serverpod_cli/src/scripts/script.dart';
+import 'package:serverpod_cli/src/scripts/scripts.dart';
+import 'package:term_glyph/term_glyph.dart' as glyph;
+import 'package:test/test.dart';
+import 'package:test_descriptor/test_descriptor.dart' as d;
+import 'package:yaml/yaml.dart';
+
+void main() {
+  glyph.ascii = false; // for windows to render spans correct
```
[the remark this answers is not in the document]
nielsenko: Fixed in [33cb738](https://github.com/serverpod/serverpod/pull/4416/commits/33cb7388f5cfa69fa6e9fc88d405560be156b080) 🙄

tools/serverpod_cli/lib/src/scripts/scripts.dart
```diff
@@ -0,0 +1,122 @@
...
+    return Scripts.fromPubspec(pubspecPlus);
+  }
+
+  /// Loads scripts from [pubspecPlus]
+  ///
+  /// Returns an empty [Scripts] if the file doesn't exist or doesn't contain
+  /// a `serverpod_scripts` section.
+  ///
+  /// Throws [ScriptsParseException] if the scripts section is malformed.
+  factory Scripts.fromPubspec(PubspecPlus pubspecPlus) {
```
SandPod, line 94: Suggestion: Hide this.

This is only used internally and seems to be an implementation detail. Hide it to slim down the public Scripts interface removing the need for having it explicitly tested.

tools/serverpod_cli/lib/src/scripts/scripts.dart
```diff
@@ -0,0 +1,122 @@
...
+
+  /// Finds the pubspec.yaml file for the Serverpod server project.
+  ///
+  /// Uses [ServerDirectoryFinder] to locate the server directory,
+  /// which handles walking up directories, sibling patterns, and
+  /// repository boundary detection.
+  ///
+  /// Returns null if no server directory is found.
+  static File? findPubspecFile(Directory directory) {
+    final serverDir = ServerDirectoryFinder.search(directory);
```
SandPod: Suggestion: Add the interactive flag.

In the run command, parse the global interracive flag and pass it on to the ServerDirectoryFinder. You can look at the create_migration.dart command for inspiration.

tools/serverpod_cli/lib/src/scripts/scripts.dart
```diff
@@ -0,0 +1,122 @@
...
+    return Scripts.fromYaml(scriptsYaml);
+  }
+
+  /// Finds the pubspec.yaml file for the Serverpod server project.
+  ///
+  /// Uses [ServerDirectoryFinder] to locate the server directory,
+  /// which handles walking up directories, sibling patterns, and
+  /// repository boundary detection.
+  ///
+  /// Returns null if no server directory is found.
```
SandPod: Suggestion(NIT): The interface of this method is a bit strange where a `non-null` return doesn't mean that the file actually exist.

One option is to call the directory finder directly in the command and then act accordingly but I understand why you always want to end up with a File.

tools/serverpod_cli/test/commands/run_command_test.dart
```diff
@@ -0,0 +1,174 @@
...
+name: test_server
+environment:
+  sdk: ^3.0.0
+dependencies:
+  serverpod: any
+
+serverpod_scripts:
+  hello: echo "Hello from script"
+  fail: exit 42
+  start: dart bin/main.dart
```
SandPod: Question: Does this path work on windows? Shouldn't the path be platform dependent?

tools/serverpod_cli/test/commands/run_command_test.dart
```diff
@@ -0,0 +1,174 @@
...
+        final result = await runServerpod(
+          ['run', '--list'],
+          workingDirectory: serverDir,
+        );
+
+        expect(result.exitCode, 0);
+        expect(result.stdout, contains('hello:'));
+        expect(result.stdout, contains('fail:'));
+        expect(result.stdout, contains('start:'));
+        expect(result.stdout, contains('test:'));
```
SandPod: Suggestion: This misses the `trap` command.

Should we convert this to a `containsAll`?

tools/serverpod_cli/test/commands/run_command_test.dart
```diff
@@ -0,0 +1,174 @@
...
+      () async {
+        final result = await runServerpod(
+          ['run', 'nonexistent'],
+          workingDirectory: serverDir,
+        );
+
+        expect(result.exitCode, isNot(0));
+        final output = '${result.stdout}${result.stderr}';
+        expect(output, contains('Script "nonexistent" not found'));
+        expect(output, contains('hello:'));
```
SandPod: Suggestion: Split the tests based on the same scenario.

Why is hello printed? Please add a separate validating test that explains this. Same for the other tests.

tools/serverpod_cli/test/scripts/scripts_test.dart
```diff
@@ -0,0 +1,402 @@
...
+import 'package:term_glyph/term_glyph.dart' as glyph;
+import 'package:test/test.dart';
+import 'package:test_descriptor/test_descriptor.dart' as d;
+import 'package:yaml/yaml.dart';
+
+void main() {
+  // Use Unicode glyphs for better and consistent visual output
+  // (Not default on windows)
+  glyph.ascii = false;
+  group('Script', () {
```
SandPod: Suggestion: Remove the outer group and break this out into different files.

tools/serverpod_cli/lib/src/commands/run.dart
```diff
@@ -0,0 +1,156 @@
...
+
+  void _listScripts(Scripts scripts) {
+    log.info('Available scripts:');
+    for (final script in scripts.values) {
+      log.info('  ${script.name}: ${script.command}');
+    }
+  }
+
+  Future<int> _executeScript(Script script, String workingDirectory) async {
```
SandPod, line 123: Suggestion: Make this a method on the `Script` class and take in the working directory as a parameter.

tools/serverpod_cli/lib/src/commands/run.dart
```diff
@@ -0,0 +1,156 @@
...
+    // Find and execute the script
+    final script = scripts[scriptName];
+    if (script == null) {
+      log.error('Script "$scriptName" not found.');
+      log.info('Available scripts:');
+      _listScripts(scripts);
+      throw ExitException(ServerpodCommand.commandInvokedCannotExecute);
+    }
+
+    final workingDirectory = pubspecFile.parent.path;
```
SandPod, line 108: Suggestion: Locate the server directory outside of `Scripts` so that we have it available here.

Might be best just to resolve it in this context since we require it here instead of stepping out of the pubspec files path.

SandPod (commented):
Nice quick work!

Two additional things:
1. Can we add an e2e test that validates this works as expected? We have our e2e cli tests we can use.
2. Is it a big addition to allow additional arguments to be passed down to the script? This is scope creep and could be added later on upon request.

tools/serverpod_cli/lib/src/commands/run.dart
nielsenko: Disagree. We have resolved the pubspec, which is needed to parse the scripts, the commands will run relative to the placement of the resolved pubspec.

tools/serverpod_cli/test/commands/run_command_test.dart
nielsenko: We list all available scripts in the output when we cannot find the one requested

nielsenko: I will make it explicit that we expect them all. Or do you prefer us not to list the available scripts on a missed find?

tools/serverpod_cli/test/commands/run_command_test.dart
nielsenko: Run just calls the script. It does nothing to its arguments meaning. Trying to parse the arguments are bound to lead to trouble. There are a ton of corner cases.

tools/serverpod_cli/lib/src/scripts/scripts.dart
nielsenko: I consider this the primary interface. [PubspecPlus] is no more an implementation detail than [Scripts] is. If we should hide a ctor it should be `fromYaml` as this can be called with the wrong `YamlNode`, but I kept it for easier testing, ie. I don't need to setup a fully valid pubspec.

`Scripts.fromPubspecFile` is just a convenience ctor.

tools/serverpod_cli/lib/src/commands/run.dart
```diff
@@ -0,0 +1,156 @@
...
+    final shellArg = Platform.isWindows ? '/c' : '-c';
+
+    // NOTE: We invoke a shell instead of the script.command directly (with
+    // runInShell: true). This avoid a lot of edge cases regarding quoting,
+    // repeated spaces, etc.
+    final process = await Process.start(
+      shell,
+      [shellArg, script.command],
+      workingDirectory: workingDirectory,
+    );
```
Isakdl, line 136: Question: Can we guarantee that bash is available on none windows machines?

nielsenko: I'l be hard pressed to mention a platform (should be some version of Linux) where we can reasonably run serverpod and that does not have `bash`. I think it feels a bit weird to keep using `sh` for stuff like this. I'm an old `fish` fan myself, but `bash` is everywhere.

tools/serverpod_cli/lib/src/scripts/scripts.dart
SandPod: I don't agree with this. The interface we are using from the run command is the one we should validate works as expected. Everything else in this class should be susceptible for a refactor without loosing confidence that what is actually used in the production code is still validated.

So either you would need to duplicate all tests for all interfaces to cove all levels, which I believe is unecessary and just hardens the code structure, or hide the implementation details and only expose the interface that is used

tools/serverpod_cli/test/commands/run_command_test.dart
SandPod: Aah, I confused the level we are testing on here. Disregard my comment.

tools/serverpod_cli/test/commands/run_command_test.dart
SandPod: I think it is a good feature to list all available scripts when one is missing! So you should keep it :) Just add a separate test explaining that we are expecting this given the same `when` statement.

tools/serverpod_cli/lib/src/commands/run.dart
nielsenko: I prefer `Script` to ba a plain ~~`Pojo`~~`Podo` No need for it to depend on `dart:io`.

tools/serverpod_cli/lib/src/scripts/scripts.dart
nielsenko: Now it tests for existence, and passing a non-existent file to ctor will cause `PathNotFoundException`. Since ae8a723

tools/serverpod_cli/test/commands/run_command_test.dart
nielsenko: Introduced a `containsLines` matcher and used it throughout. Since ae8a723

tools/serverpod_cli/test/commands/run_command_test.dart
nielsenko: Fixed in ae8a723 with `containsLines` matcher.

tools/serverpod_cli/lib/src/commands/run.dart
Isakdl: Alright, ill copy this solution for use in the cloud side, there is an argument to be made to move this into CLI tools, as we need this code in both repos.

Isakdl: Actually, shouldn't we do that? Seems reasonable too me, _executeScript would have to be a public method in CLI tools though, as I will loop several scripts on my side.

tools/serverpod_cli/lib/src/scripts/scripts.dart
nielsenko: I'll remove `fromPubspecFile` then and only test `fromPubspec`. Do you think all the test of `_fromYaml` should be converted to using `fromPubspec` as well?

nielsenko:
> Nice quick work!
> 
> A couple of comments and suggestions.
> 
> Two additional things:
> 
> 1. Can we add an e2e test that validates this works as expected? We have our e2e cli tests we can use.
> 2. Is it a big addition to allow additional arguments to be passed down to the script? This is scope creep and could be added later on upon request.

Due to some conflict between args and cli_tools (or my lagging knowledge) I'm having trouble getting rest args (those after --) to work with the `run` command. I will return to this later, as to not waste more time now.

@coderabbitai Will you create a follow-up issue for this?

tools/serverpod_cli/lib/src/scripts/scripts.dart
nielsenko: Added in 32525c1e2efc

tools/serverpod_cli/lib/src/scripts/scripts.dart
nielsenko: `_fromYaml` is now truly private (no `ScriptsInternal.fromYaml` side-channel). All tests, now use either `Scripts.fromPubspec` or `Scripts.fromPubspecFile`. I kept tests of `Scripts.fromPubspecFile` as there are some subtle differences in error reporting (the rendered spans are different, when a file uri source is known), and it can obviously throw file related exceptions (such as `PathNotFound`).

tools/serverpod_cli/test/scripts/scripts_test.dart
nielsenko: Outer groups removed. Still just one file (314 lines is not that bad for a test file).

tools/serverpod_cli/lib/src/commands/run.dart
nielsenko: There is a PR to move the functionality to `cli_tools` (https://github.com/serverpod/cli_tools/pull/96)

tools/serverpod_cli/lib/src/commands/run.dart
nielsenko: See https://github.com/serverpod/cli_tools/pull/96

nielsenko:
> * Can we add an e2e test that validates this works as expected? We have our e2e cli tests we can use.

Added (/moved) in 80beac5

nielsenko:
@coderabbitai Create issue for using the `runServerpod` function from `run_command_test.dart` in all `serverpod_cli_e2e_test` tests.

nielsenko:
@coderabbitai Create issue to use `execute` from https://github.com/serverpod/cli_tools/pull/96 instead of `RunCommand._executeScript`.

tests/serverpod_cli_e2e_test/test/run_command_test.dart
```diff
@@ -0,0 +1,222 @@
...
+environment:
+  sdk: ^3.0.0
+dependencies:
+  serverpod: any
+
+serverpod:
+  scripts:
+    hello: echo "Hello from script"
+    fail: exit 42
+    start: dart bin/main.dart
```
SandPod, line 106: Question: Wouldn't this script fail if run on windows because of the path?

tools/serverpod_cli/lib/src/commands/run.dart
```diff
@@ -0,0 +1,167 @@
...
+    // Parse scripts
+    final Scripts scripts;
+    try {
+      scripts = Scripts.fromPubspecFile(pubspecFile);
+    } on ScriptsParseException catch (e) {
+      log.error('Error parsing "serverpod/scripts":\n$e');
+      throw ExitException(ServerpodCommand.commandInvokedCannotExecute);
+    }
+
+    if (scripts.isEmpty) {
```
SandPod, line 86: Suggestion: Treat this the same as missing.

Would it make sense to allow this but treat it as the same as missing? So you can always run `serverpod run list` or `serverpod run` to see what scripts are available, even in a project that has no scripts?

tests/serverpod_cli_e2e_test/test/run_command_test.dart
```diff
@@ -0,0 +1,222 @@
...
+  );
+}
+
+void main() {
+  late String serverDir;
+
+  setUpAll(() async {
+    // Create a single test server with all scripts needed by tests
+    await d.dir('test_server', [
+      d.file('pubspec.yaml', '''
```
SandPod, line 95: Suggestion: Missing tests for when the `serverpod` namespace is missing in the pubspec.

This will be the default behavior for everyone that migrates an earlier project, so lets adds som automated tests there to ensure we behave correctly when we start expanding or modifying this code.

tests/serverpod_cli_e2e_test/test/run_command_test.dart
```diff
@@ -0,0 +1,222 @@
...
+      expect(result.exitCode, 0);
+      expect(
+        result.stdout,
+        containsLines(['hello:', 'fail:', 'start:', 'test:']),
+      );
+    },
+  );
+
+  test(
+    'when running a non-existent script, then an error is shown',
```
SandPod: Suggestion: Make this a `when` group and break the expects out into to tests that validated each scenario with clear descriptions.

Let's help future devs understand that the `containsLines` here actually means that we print all the available scripts.

SandPod (approved):
Last couple of comments then we are good to go! 👍

tests/serverpod_cli_e2e_test/test/run_command_test.dart
nielsenko: No.

Windows APIs accept / as a path separator, so commands like `type C:/Users/file.txt` works just fine. 

`cmd.exe` itself uses / for command switches (e.g., `dir /w`). So some commands get confused when a path starts with / or in certain argument positions, but `dart` in particular does not.

I really don't think we should try to add smartness here.

tools/serverpod_cli/lib/src/commands/run.dart
nielsenko: You can always run `serverpod run`, but if `scripts.isEmpty` there is nothing to list?

Now we just log "No scripts defined ..."

tests/serverpod_cli_e2e_test/test/run_command_test.dart
nielsenko: Addressed in [daafdba](https://github.com/serverpod/serverpod/pull/4416/commits/daafdba9237f9d544e4ff830f0b81165a4953166)

tests/serverpod_cli_e2e_test/test/run_command_test.dart
nielsenko: Addressed in [daafdba](https://github.com/serverpod/serverpod/pull/4416/commits/daafdba9237f9d544e4ff830f0b81165a4953166)

tools/serverpod_cli/lib/src/commands/run.dart
SandPod: Sorry, It is already treated as command missing 👍 

This implementation is correct.

PR #4432, opened 2025-12-12
fix: Throw PasswordNotFoundException instead of null assertion in IDP configs

exaby73:
## Pre-launch Checklist

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/config/from_passwords_config_test.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/config/from_passwords_config_test.dart
@@ -36,10 +36,16 @@ void main() {
         Endpoints(),
         config: ServerpodConfig(apiServer: portZeroConfig),
       );
-    });
 
-    tearDownAll(() async {
-      Directory.current = originalDir;
+      // Teardown need to be done using `addTearDown` instead of `tearDownAll`
+      // because the call to `addTearDown` inside `d.sandbox` will add a tear
+      // down that will run before any declared `tearDown` functions. Otherwise,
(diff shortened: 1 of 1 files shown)


modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/apple/business/apple_idp_config.dart
```diff
@@ -56,16 +57,16 @@ class AppleIdpConfigFromPasswords extends AppleIdpConfig {
...
-        )!,
-        redirectUri: Serverpod.instance.getPassword('appleRedirectUri')!,
-        teamId: Serverpod.instance.getPassword('appleTeamId')!,
-        keyId: Serverpod.instance.getPassword('appleKeyId')!,
-        key: Serverpod.instance.getPassword('appleKey')!,
+        ),
+        redirectUri: Serverpod.instance.getPasswordOrThrow('appleRedirectUri'),
+        teamId: Serverpod.instance.getPasswordOrThrow('appleTeamId'),
+        keyId: Serverpod.instance.getPasswordOrThrow('appleKeyId'),
+        key: Serverpod.instance.getPasswordOrThrow('appleKey'),
```
marcelomendoncasoares, line 69: Suggestion: For configurations that rely on multiple keys, this individual approach will create a poor UX if several are missing (one error -> fix other error -> fix, etc.). In this case, it would be better to collect all and inform the missing ones at once.

To enforce a repeatable standard, I would consider not using the `getPasswordOrThrow` extension, but instead throw an exception testing null on the group of keys at every IDP (even when this group only contains 1 key for now).

marcelomendoncasoares (approved):

Issue #4403, opened 2025-12-10
Support building and serving Flutter web from project template


SandPod:
### Problem to Solve

Currently, when a project is created with Serverpod, the Flutter web app is not automatically built and served by the web server. Users must manually build the Flutter app and configure server-side serving, which adds complexity and friction for new projects.

### Proposal

Enable Serverpod’s template to automatically build and serve the Flutter web application when the server is started. After running:

`serverpod create <project_name>`
`dart bin/main.dart --apply-migrations`

…users should be able to visit the web server and see their Flutter frontend running, without manual web build or configuration steps. The process should:
- Build the Flutter app (in the correct directory) as part of the server startup or migration process.
- Serve the Flutter web build through the serverpod web server.

### Use Case

A developer spins up a new Serverpod project, applies migrations, and immediately accesses the Flutter web frontend in their browser without additional setup. This streamlines development for fullstack Dart/Flutter projects and enables instant web previews for project templates.

### Alternatives

Current alternatives require manually building the Flutter web app and copying files into the correct location, and configuring the web server to serve these files. This adds manual work and can cause onboarding confusion for new users.

### Additional context

Integration could leverage Dart’s build tools and Flutter’s CLI. May require adding a build step to template setup or server startup, and documentation updates. Consider environments for development vs. production builds.

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

PR #4442, opened 2025-12-14
feat: Serve a flutter app for new project templates.

SandPod:
This PR introduces updates to the template to serve the Flutter web app directly after the project has been created.

### Guidance for review

Best reviewed commit by commit.

#### First and last commit
The first and last PR contain the creation and removal of an example project that was the target when creating the templates.

I simply built the project I wanted and then removed and recreated it until there was no meaningful difference between the generated output and my target.

I have kept these in case further iterations of the templates are needed.

### Upcoming work

Once https://github.com/serverpod/serverpod/pull/4416 is in, a new command should be added that builds and copies the Flutter app over to the Server directory. This command should be run by running `serverpod run flutter_build`.

### Outstanding questions

- Should we serve a placeholder page if the app has not been built? The drawback is that it complicates the initial server setup quite significantly.
- Should we always build the app when creating a new project? The time overhead added on my machine significantly outranks all other create steps.

### Windows and `-o` for Flutter build
There seems to be a problem with flutter build when building on windows and using the `-o` flag. Therefore we need to have a separate move step to move the contents of the build Flutter app to the server directory for serving.

Issue tracked here: https://github.com/flutter/flutter/issues/157886

Closes: #4403

## Pre-launch Checklist

## Breaking changes

None, this change will only affect newly created Serverpod projects.

### DIFF
--- a/examples/legacy/auth_example/auth_example_server/lib/server.dart
+++ b/examples/legacy/auth_example/auth_example_server/lib/server.dart
@@ -39,7 +39,6 @@ void run(List<String> args) async {
   // Serve all files in the /static directory.
   pod.webServer.addRoute(
     StaticRoute.directory(Directory('static')),
-    '/**',
   );
 
   // Configuration for sign in with email.
--- a/examples/middleware/middleware_server/lib/server.dart
+++ b/examples/middleware/middleware_server/lib/server.dart
@@ -21,7 +21,7 @@ void run(List<String> args) async {
   pod.webServer.addRoute(RootRoute(), '/index.html');
   // Serve all files in the web/static relative directory under /.
   final root = Directory(Uri(path: 'web/static').toFilePath());
-  pod.webServer.addRoute(StaticRoute.directory(root), '/**');
+  pod.webServer.addRoute(StaticRoute.directory(root));
 
   // ## Verbose Mode
   //
--- a/modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/generated/protocol.dart
+++ b/modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/generated/protocol.dart
@@ -290,6 +290,9 @@ class Protocol extends _i1.SerializationManagerServer {
     if (t == _i1.getType<_i7.LegacySession?>()) {
       return (data != null ? _i7.LegacySession.fromJson(data) : null) as T;
     }
+    if (t == Set<String>) {
+      return (data as List).map((e) => deserialize<String>(e)).toSet() as T;
+    }
     try {
       return _i3.Protocol().deserialize<T>(data, t);
     } on _i1.DeserializationTypeNotFoundException catch (_) {}
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/generated/protocol.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/generated/protocol.dart
@@ -1030,6 +1030,21 @@ class Protocol extends _i1.SerializationManagerServer {
               : null)
           as T;
     }
+    if (t == Map<String, String>) {
+      return (data as Map).map(
+            (k, v) => MapEntry(deserialize<String>(k), deserialize<String>(v)),
+          )
+          as T;
+    }
+    if (t == _i1.getType<Map<String, String>?>()) {
+      return (data != null
+              ? (data as Map).map(
+                  (k, v) =>
+                      MapEntry(deserialize<String>(k), deserialize<String>(v)),
+                )
+              : null)
+          as T;
+    }
     if (t == _i1.getType<({_i25.ByteData challenge, _i1.UuidValue id})>()) {
       return (
             challenge: deserialize<_i25.ByteData>(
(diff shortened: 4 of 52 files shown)


templates/serverpod_templates/projectname_flutter/lib/src/app_config.dart
```diff
@@ -0,0 +1,32 @@
...
+    return jsonDecode(data);
+  }
+}
+
+extension on Map<String, dynamic> {
+  T? get<T>(String key) {
+    final value = this[key];
+    return value as T?;
+  }
+}
```
nielsenko: isn't
```dart
final map = <String, dynamic>{ 'a': 42 };
int? i = map['a'];
```
the same as:
```dart
final map = <String, dynamic>{ 'a': 42 };
final i = map.get<int>(['a']);
``` 
?  - Due to the way _dynamic_ is cast (or do we have some lint enabled that complains?).

tools/serverpod_cli/lib/src/util/command_line_tools.dart
```diff
@@ -39,6 +40,47 @@ class CommandLineTools {
     return true;
   }
 
+  static Future<bool> flutterBuild(Directory dir, Directory serverDir) async {
+    log.debug('Running `flutter build` in ${dir.path}', newParagraph: true);
+
+    // Unfortunately, we can't use the `-o` flag directly because of an error
+    // that happens on windows. Issue is tracked by in flutter
```
nielsenko:
```suggestion
    // that happens on windows. Issue is tracked in flutter
```

tools/serverpod_cli/lib/src/util/pubspec_helpers.dart
```diff
@@ -1,12 +1,16 @@
 import 'dart:io';
 
 import 'package:path/path.dart' as p;
+import 'package:pub_semver/pub_semver.dart';
 import 'package:pubspec_parse/pubspec_parse.dart';
+import 'package:source_span/source_span.dart';
+import 'package:yaml/yaml.dart';
+import 'package:yaml_edit/yaml_edit.dart';
 
 Pubspec parsePubspec(File pubspecFile) {
```
nielsenko, line 10: I can recommend using `PubspecPlus` that combines a `Pubspec` instance with the `Yaml` it was parsed from. This allows it inflate spans, if you need to raise errors. It has functionality for iterating over typed deps and getting the `type` (hosted, path, git, or sdk) and `kind`, (normal, dev, override), as well as the `span` dep is in the pubspec.yaml (if you need to raise an error).

templates/serverpod_templates/projectname_server_upgrade/web/pages/build_flutter_app.html
```diff
@@ -0,0 +1,41 @@
...
+      <div class="info-box">
+        <p style="margin-bottom: 12px; font-weight: 600; color: #333;">Flutter app not built</p>
+        <p>
+          The Flutter web app has not been built yet.
+        </p>
+        <p>
+          To build the app, run the following command in your Flutter project directory:
+        </p>
+        <div style="background-color: #f5f5f5; padding: 12px; border-radius: 6px; margin-top: 12px; margin-bottom: 12px; font-family: 'Courier New', monospace; font-size: 13px; color: #333; overflow-x: auto;">
+          flutter build web --base-href /app/ --wasm -o ../flutter_web_server/web/app
```
SandPod: Chore: Replace `flutter_web_server` with replaceable name referencing the server directory.

templates/serverpod_templates/projectname_flutter/lib/src/app_config.dart
SandPod: Yes it is, this was a residue from the copied code. Will remove the extension.

tools/serverpod_cli/lib/src/util/pubspec_helpers.dart
SandPod: Could be a good addition if we ever need to raise exceptions. Currently we use a file that we have created so shouldn't be a problem. (if the project has been modified this part of the code will never run).

templates/serverpod_templates/projectname_flutter/lib/src/app_config.dart
SandPod: Fixed here: [5076eb2](https://github.com/serverpod/serverpod/pull/4442/commits/5076eb2626c7ea8a39a9eda8fe5d8b09139b27f4)

tools/serverpod_cli/lib/src/util/command_line_tools.dart
SandPod: Fixed: [c351ee4](https://github.com/serverpod/serverpod/pull/4442/commits/c351ee4663e8e3041976e1324b47310a5dd06686)

templates/serverpod_templates/projectname_server_upgrade/web/pages/build_flutter_app.html
SandPod: Fixed: 9dd1618

tools/serverpod_cli/lib/src/util/pubspec_helpers.dart
nielsenko: Okay

PR #4443, opened 2025-12-14
fix: Ensure tailmatch is the default for `StaticRoute.directory`

nielsenko:
Currently `router.group('/**').add('/**/', o)` isn't illegal. However, it makes no sense, and `o` will not be looked up on say `a/b` as you might expect. _The tail is a tail_, you don't get to attach anything to a tail.

We managed to trip us self on this, as `addRoute`  is implemented as `group(path).inject(handlerObject)`.  

To fix this Relic has been updated to ^0.14.0 and will now throw whenever an attempt is made on attaching a router on a tail-path. _**UNLESS**_ the attached router contains a single root value, _**AND**_ we are allowed to consume the nodes of the attached router (as opposed to sharing them). The permission to consume is given with the new `attach(..., consume: true)` flag.

Now `addRoute` uses `injectAt`, that is now implemented like:

```dart
  void injectAt(final String path, final InjectableIn<Router<T>> injectable) {
    final subRouter = Router<T>()..inject(injectable);
    attach(path, subRouter, consume: true);
  }
```
Notice it no longer uses `group`, but instead uses `attach` directly with the new `consume: true` flag. 

This will allow the somewhat common corner-case: `webServer.addRoute(MyRoute(), '/**')`, iff `MyRouter` doesn't touch `path` and `injectIn`, ie. it simply attach a single handler at the '/' of the given `router`.
 
```dart
class MyRoute {
  @override
  void injectIn(RelicRouter router) => router.anyOf(methods, '/', _myHandler);
  // ...
}
```

which is the default behaviour inherited from `Route`.

This PR also moves responsibility of the tail-match segment down into `StaticRoute` and ensures no _earlier_ segment uses `**`. This ensures it goes as late as possible, and is never omitted.

Tests has been updated not to use `/**` when setting up `FlutterRoute`, `SpaRoute`, and `StaticRoute.directory` (this is now left to `StatidRoute.directory` ctor). Tests for the `/` corner-case has been added.

## Pre-launch Checklist

## Breaking changes

None. No interface changes. Semantics has changed, but this is reasonably a bug-fix.

### DIFF
--- a/packages/serverpod/lib/src/web_server/web_server.dart
+++ b/packages/serverpod/lib/src/web_server/web_server.dart
@@ -58,9 +58,10 @@ class WebServer {
   /// Adds [route] to the server at [path].
   ///
   /// The [path] supports:
-  /// - Literal segments: `/users/profile`
-  /// - Parameters: `/users/<id>`
-  /// - Wildcards: `/files/*` (single segment) or `/files/**` (tail match)
+  /// - Literal segments: `/admin/profile/edit`
+  /// - Parameters: `/users/:id/posts`
+  /// - Wildcards: `/api/*/list` (single segment)
+  /// - Tails: `/files/**` (tail match)
   ///
   /// The full route path combines [path] with [Route.path]. For example,
   /// a route with `path: '/edit'` added via `addRoute(route, '/users')`


exaby73 (approved):
LGTM

nielsenko:
@coderabbitai review

packages/serverpod/test/web_server/add_route_test.dart
```diff
@@ -0,0 +1,136 @@
...
+
+void main() {
+  final portZeroConfig = ServerConfig(
+    port: 0,
+    publicScheme: 'http',
+    publicHost: 'localhost',
+    publicPort: 0,
+  );
+
+  group('WebServer.addRoute', () {
```
marcelomendoncasoares: Missing "Given" on the test description.

packages/serverpod/test/web_server/add_route_test.dart
```diff
@@ -0,0 +1,136 @@
...
+              ),
+              '/**',
+            ),
+            throwsArgumentError,
+          );
+        },
+      );
+
+      test(
+        'rejects route with non-rooted path',
```
marcelomendoncasoares: Missing "then" to compose the Given/When/Then pattern. Same for below tests and groups.

nielsenko: Seems to be a common theme 😬

packages/serverpod/test/web_server/add_route_test.dart
nielsenko: Fixed in b49fc07

packages/serverpod/test/web_server/add_route_test.dart
nielsenko: Fixed in b49fc07

SandPod:
@coderabbitai please create an issue for adding test that validate the routing behavior of `StaticRoute.file`. It seems like this is never tested.

PR #4458, opened 2025-12-16
fix: Ensure consistent copy behavior for flutter_build script.

SandPod:
If the directory existed, the build script would place the contents inside of a web folder inside of the directory.
If it didn't exist it would correctly create an app folder with the contents.

This PR modified the behavior so that it becomes consistent.

CI should inform us if this works as expected on Windows.

## Pre-launch Checklist

## Breaking changes

None, fixes an unreleased script.


nielsenko (approved):
.. but do check up on `-p` on windows. I believe `mkdir` will create the path recursively be default on windows.

SandPod:
Went with `rm -rf` since that is always run even if the directory does not exist.

Always scary to introduce an `-f` but I think it should be OK in this environment. And the user can always modify the script if another behavior is desired.

PR #4463, opened 2025-12-16
feat: Prevent database operations on health check when the database is idle

### LINKED ISSUE #4313: Allow the database to sleep when the server is idle
### Problem to Solve

Currently, the database remains active even when the Serverpod server is idle, particularly due to periodic health checks. This can prevent the database from entering a sleep state, which may lead to unnecessary resource usage and costs.

The health check service (see `health_check.dart` and `health_check_manager.dart`) regularly pings the database to check its status, and this keeps the connection open and prevents sleep. There may also be other services or mechanisms in the code that keep database connections alive regardless of server activity.

### Proposal

Modify the health check service logic so the database is only pinged when the server is active to allow it to sleep. This means that RAM and CPU stats collection will only happen when the server is under usage.

It might be needed to revisit other mechanisms, such as future call scheduling, cleanup routines, and connection pooling, to identify and adapt any code that prevents sleep.

### Use Case

A deployment running on cloud infrastructure where database resources are billed per uptime or connection activity, or when a server is designed for low-traffic or event-driven workloads. Letting the database sleep could reduce costs and environmental impact.

### Alternatives

None at the moment.

### Additional context

_No response_

- [ ] I want to work on this

marcelomendoncasoares:
Closes #4313.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tests/serverpod_test_server/test_integration/health_check_manager_test.dart
+++ b/tests/serverpod_test_server/test_integration/health_check_manager_test.dart
@@ -52,10 +52,14 @@ void main() {
         await server.shutdown(exitProcess: false);
       });
 
-      test('when server starts then one health check is recorded.', () async {
-        final healthChecks = await server.countHealthChecks(session);
-        expect(healthChecks, equals(1));
-      });
+      test(
+        'when server starts then one health check is recorded.',
+        () async {
+          final healthChecks = await server.countHealthChecks(session);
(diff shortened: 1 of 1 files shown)


nielsenko (approved):
👍

marcelomendoncasoares:
@coderabbitai Please open an issue regarding the flaky health check test that needs fixing after this PR.

PR #4474, opened 2025-12-18
feat: Add Firebase IDP

### LINKED ISSUE #3454: Create Firebase authentication module

exaby73:
Closes #3454

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/examples/auth/auth_flutter/README.md
+++ b/examples/auth/auth_flutter/README.md
@@ -82,3 +82,69 @@ needed for Android and web platforms.
 
 > For more information on how to configure Apple Sign-In credentials, see the
 > [Sign in with Apple documentation](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms).
+
+## Sign-In With Firebase Setup
+
+Firebase Authentication allows you to use Firebase's authentication system with
+Serverpod. This enables you to leverage Firebase's various sign-in providers
+(email/password, phone, social providers, etc.) while syncing the authenticated
+user with Serverpod's auth system.
+
+### 1. Create a Firebase Project
+
+1. Go to the [Firebase Console](https://console.firebase.google.com/)
+2. Create a new project or select an existing one
+3. Enable the Authentication providers you want to use (e.g., Email/Password)
+
+### 2. Configure FlutterFire
+
+Install the FlutterFire CLI and configure your Flutter app:
+
+```bash
+# Install FlutterFire CLI
+dart pub global activate flutterfire_cli
+
+# Configure your Flutter app (run from the auth_flutter directory)
+flutterfire configure
+```
+
+This will generate the `lib/firebase_options.dart` file with your Firebase
+configuration.
+
+### 3. Enable Firebase in the App
+
+In `lib/firebase.dart`, uncomment the import and options line:
+
+```dart
+// Uncomment this import:
+import 'firebase_options.dart';
+
+Future<void> initializeFirebase() async {
+  await Firebase.initializeApp(
+    // Uncomment this line:
+    options: DefaultFirebaseOptions.currentPlatform,
+  );
+}
+```
+
+### 4. Server Configuration
+
+On the server side, you need to configure the Firebase service account key for
+ID token verification. In your server's `config/passwords.yaml`, add your
+Firebase service account JSON:
+
+```yaml
+firebase:
+  serviceAccountKey: |
+    {
+      "type": "service_account",
+      "project_id": "your-project-id",
+      ...
+    }
+```
+
+You can download the service account key from the Firebase Console under
+**Project Settings > Service Accounts > Generate New Private Key**.
+
+> For more information on Firebase Authentication, see the
+> [Firebase Authentication documentation](https://firebase.google.com/docs/auth/flutter/start)


exaby73:
Ignore failing flaky test

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/firebase/business/firebase_idp_token_verifier.dart
```diff
@@ -0,0 +1,255 @@
...
+import 'package:dart_jsonwebtoken/dart_jsonwebtoken.dart' as dart_jsonwebtoken;
+import 'package:http/http.dart' as http;
+import 'package:synchronized/synchronized.dart';
+
+/// Utility class for verifying Firebase ID tokens.
+///
+/// This implementation follows the pattern from Google's ID token verification,
+/// fetching public keys from Firebase's certificate endpoint and validating
+/// JWT claims.
+class FirebaseIdTokenVerifier {
```
marcelomendoncasoares: Question: The `FirebaseIdTokenVerifier` is very much similar to `GoogleIdTokenVerifier`. Is it possible to abstract the code to reuse for both instead of fully duplicating? The original code from where I built the Google one [even had a method to validate Firebase tokens](https://github.com/googleapis/google-auth-library-python/blob/c767531ce05a89002d109f595187aff1fcaacfb7/google/oauth2/id_token.py#L196), so it is highly possible that they can be combined.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/firebase/business/firebase_idp_utils.dart
```diff
@@ -0,0 +1,207 @@
+import 'package:serverpod/serverpod.dart';
+
+import '../../../../../core.dart';
+import 'firebase_idp_config.dart';
+import 'firebase_idp_token_verifier.dart';
+
+/// Details of the Firebase Account.
+///
+/// Some fields may be null depending on the authentication provider used
+/// within Firebase (e.g., phone auth may not have email).
```
marcelomendoncasoares, line 11: Question: Even when using phone auth the phone is not returned?

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/firebase/business/firebase_idp_token_verifier.dart
```diff
@@ -0,0 +1,255 @@
...
+  /// Certificates are cached for 1 hour to reduce network requests. If the
+  /// request fails, the cache update is postponed for 10 minutes up to 5 times.
+  /// If the request fails 5 times, throws
+  /// [FirebaseIdTokenValidationServerException].
+  static Future<Map<String, dart_jsonwebtoken.JWTKey>> _getKeys() async {
+    if (!_cachedKeySet.shouldUpdate) {
+      return _cachedKeySet.keys;
+    }
+
+    await _lock.synchronized(() async {
```
marcelomendoncasoares: Question: Do we really need the `synchronized` dependency for this? We could avoid the extra dependency by using the same simple locking mechanism present in the `GoogleIdTokenVerifier`.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/firebase/business/firebase_idp_utils.dart
exaby73: What I meant here was that if you logged in with phone, there may not be an email associated with that account

exaby73: TODO(exaby73): Add phone field to the return object

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/firebase/firebase_sign_in_service.dart
marcelomendoncasoares, line 1: Although we can say that using the Firebase IDP is just calling `client.<firebaseIdpEndpoint>.login`, it feels like a less polished experience than the one we offer with other IDPs through their pre-existing controllers.

We don't want to depend on `firebase_ui_auth`, but we might be able to create a callback that uses only the `firebase_auth` classes to reduce the boilerplate for users to link Serverpod Auth with `firebase_ui_auth`.

My proposal is to create a `FirebaseAuthController` that will have only a `login` method with no parameters - analogous to `AppleAuthController`. This controller should have a getter to return a callback that receives a `firebase_auth.User?` class and implements a similar logic as the one in the legacy firebase package:

1. Testing if `User` is null.
2. Getting its `idToken`.
3. Passing the `idToken` to our Firebase `login` endpoint.
4. Calling `client.auth.updateSignedInUser(authSuccess)` to perform the login on the app.

https://github.com/serverpod/serverpod/blob/886a067c688c750e6c6ecf9c4177c1dd4a231775/modules/legacy/serverpod_auth/serverpod_auth_firebase_flutter/lib/src/auth.dart#L27-L86

Then, users will only have to use this callback inside the `AuthStateChangeAction<SignedIn>` action of Firebase's `SignInScreen` widget to have our provider integrated.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/common/id_token_verifier/id_token_verifier_config.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:dart_jsonwebtoken/dart_jsonwebtoken.dart' as dart_jsonwebtoken;
```
marcelomendoncasoares, line 1: Suggestion: Export the `JWTKey` class here so that implementations of the `IdTokenVerifierConfig` don't have to import it from a third package.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/common/id_token_verifier/id_token_verifier_exception.dart
```diff
@@ -0,0 +1,10 @@
+/// Base exception thrown when ID token validation fails.
+///
+/// Provider-specific exceptions should extend this class.
+class IdTokenValidationException implements Exception {
+  /// The exception message that was thrown.
+  final String message;
+
+  /// Creates a new instance.
+  IdTokenValidationException(this.message);
+}
```
marcelomendoncasoares: Question: Do we really need this base class? Since the exceptions that subclass it are isolated in their worlds (Google and Firebase) and we never expect to catch them both in the same place, I think it would be clearer to remove it.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/firebase/business/firebase_idp_config.dart
```diff
@@ -0,0 +1,88 @@
...
+///
+/// The password key should be `firebaseServiceAccountKey` and contain the
+/// entire service account JSON as a string.
+class FirebaseIdpConfigFromPasswords extends FirebaseIdpConfig {
+  /// Creates a new [FirebaseIdpConfigFromPasswords] instance.
+  FirebaseIdpConfigFromPasswords({
+    super.firebaseAccountDetailsValidation,
+  }) : super(
+         credentials: FirebaseServiceAccountCredentials.fromJsonString(
+           Serverpod.instance.getPassword('firebaseServiceAccountKey')!,
```
marcelomendoncasoares: We should use the `getPasswordOrThrow` as in the other providers.

```suggestion
           Serverpod.instance.getPasswordOrThrow('firebaseServiceAccountKey')!,
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/common/id_token_verifier/id_token_verifier.dart
```diff
@@ -0,0 +1,199 @@
...
+  /// 2. Validates standard claims (exp, iat, sub)
+  /// 3. Validates provider-specific claims via the config
+  ///
+  /// The [audience] parameter is passed to the config for audience validation.
+  /// Its interpretation depends on the provider (e.g., projectId for Firebase,
+  /// clientId for Google).
+  ///
+  /// Throws an exception created by the config in case of any validation
+  /// failures.
+  static Future<Map<String, dynamic>> verify(
```
marcelomendoncasoares: Question: To be more expressive of the intent of this method, should we keep the original `verifyOAuth2Token` name?

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/migrations/migration_registry.txt
```diff
@@ -5,3 +5,5 @@
 ### the conflict by removing and recreating the conflicting migration.
 
 20251208110420531-v3-0-0
+20251211185512693
+20260105043727055
```
marcelomendoncasoares: Before merging, we should squash these migrations into a single one to avoid bloating the history. Ideally, each PR should only contain a single migration that covers the feature/fix. This applies to all other packages that received more than one migration during this development.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/serverpod_auth_idp_flutter.dart
```diff
@@ -15,6 +15,7 @@ export 'src/apple/apple_sign_in_style.dart';
 export 'src/apple/apple_sign_in_widget.dart';
 export 'src/email/email_auth_controller.dart';
 export 'src/email/email_sign_in_widget.dart';
+export 'src/firebase/firebase_sign_in_service.dart';
```
marcelomendoncasoares, line 19: Once added, we can't forget to export the `FirebaseAuthController` here.

Issue #4459, opened 2025-12-16
Create a shared test utils package for internal use


exaby73:
We currently use and maintain multiple test projects, along with the tests in each published package as well. This means we often create utils in one package and either copy it over to other packages, or not use the utils and instead write the logic inline.

An example of the latter is `runServerpodCli` and `startServerpodCli` utils in [bootstrap test project's utils](https://github.com/serverpod/serverpod/blob/d5a70dfa334e17866f9ffd488659f707e759944e/tests/bootstrap_project/lib/src/util.dart#L83-L115).

We should create a shared package, and consolidate any such utils in there

PR #4494, opened 2025-12-22
fix: Use getPasswordOrThrow in JwtConfigFromPasswords and ServerSideSessionsConfigFromPasswords

exaby73:
## Pre-launch Checklist

## Breaking changes


modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/utils/get_passwords_extension.dart
```diff
@@ -1,6 +1,5 @@
 import 'package:serverpod/serverpod.dart';
-
-import 'password_not_found_exception.dart';
+import 'package:serverpod_auth_core_server/serverpod_auth_core_server.dart';
 
 /// Convenience helpers for retrieving passwords with clearer errors.
 extension GetPasswordsExtension on Serverpod {
```
marcelomendoncasoares, line 5: Question: Shouldn't this be removed now that we have the same extension on the core?

marcelomendoncasoares (approved):
Just one question of whether we should remove the extension from the IDP package, but otherwise LGTM!

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/utils/get_passwords_extension.dart
exaby73: I did not want to export the extension from `core` just to use in `idp` so I opted for a copy instead. This would be one of the things solved by https://github.com/serverpod/serverpod/issues/4459

marcelomendoncasoares: Seems somewhat contradictory to me that we use the exception from the core package, but duplicate the extension. Since we already export the exception, shouldn't we treat this extension as a core feature?

exaby73: I don't export it because I did not want it to be part of the public API

Issue #4379, opened 2025-12-06
[new_serverpod_auth] Add anonymous accounts


craiglabenz:
### Problem to Solve

Applications should be able to transparently create anonymous accounts for users to begin persisting data on their behalf without forcing cumbersome, up-front account creation flows. Later, applications should allow users to upgrade their anonymous accounts by adding more durable IDPs like email, Apple, Google, or any others an app supports.

### Proposal

As discussed offline, this is a deceivingly large and complex feature. Here is my implementation proposal, broken into areas of new functionality and changes to `examples/auth` which will exercise all of the new features.

### Core feature design
- Add a new `serverpod_auth_idp_server/lib/src/providers/anonymous/` folder which mirrors the structure of the other included providers.
- Single method for the API surface - `AnonymousAuthEndpoint.signIn()`, which accepts no parameters (other than the `Session`), and creates the new account.
- New anonymous accounts get 3 records in the database; `AuthUser`, `UserProfile`, and a session. There is no "anonymous_account" database table like the other IDPs have -- once the session is terminated, the account is defunct.

### Upgrading accounts
- Add new `Future<bool> hasAccount()`-style methods to other IDPs (e.g., `hasEmailAccount()`, `hasGoogleAccount()`, etc) for a client to detect whether a user should be presented with the UI to add other IDPs
- Add new `attach(Session)` methods to Apple and Google IDPs which create records in their respective IDP tables
- Similar for email IDP, but this will still come at the end of the full email verification flow

### Account merging
- Add `mergeAccounts(sourceUserId, targetUserId)` method to `AuthConfig` to allow applications to register custom account merging logic
- In `attach` methods listed above, first check for existing accounts (based on IDP-specific logic), and call `mergeAccounts` if supplied
  - Offer `UserProfile` merging for free, mapping column values over only if not-empty on to-be-deleted UserProfile and empty on to-be-saved UserProfile
    - Relatedly, does `new_serverpod_auth` have a mechanism for applications to swap out their `UserProfile` model? If so, default `UserProfile` mapping may not be possible, or may only work when the model is not swapped.

### Updates to `example/auth`
- Enhance logged-in screen with the following functionality:
  - Add a counter app which persists each user's clicks (serving the role of user data to later test merges)
    - Add a new model to `example/auth/auth_server`, called `Click`, which stores each time a user has clicked the Increment button
  - Add a `Drawer` widget which calls each `has[IDP]Account()` method and shows buttons as appropriate
[...]

PR #4495, opened 2025-12-22
feat: Add Anonymous IDP

craiglabenz:
[Feature design doc](https://docs.google.com/document/d/1hHURFDIIBIvO5n7bf7QhJ1Gbz9D9P2VklsEV28oES7o/edit?usp=sharing).

Adds the ability for applications to create transparent anonymous accounts that the user may not even be aware of toward the end of reducing onboarding friction for new users. Users can then later add more durable authentication methods, like email or social auth, to secure their account against device loss, or to simply allow them to log in to their same account from a second device.

Note that this functionality also allows users who might start with email, Apple, or Google, to add other providers through the account linking mechanisms.

Begins work on https://github.com/serverpod/serverpod/issues/4379.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp_config.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp_config.dart
@@ -43,47 +43,17 @@ class AnonymousIdpConfig extends IdentityProviderBuilder<AnonymousIdp> {
   /// The maximum rate of anonymous accounts that can be created from a single
   /// IP address.
   /// {@endtemplate}
-  final AnonymousIdpLoginRateLimitingConfig? perIpAddressRateLimitConfig;
+  final RateLimit? perIpAddressRateLimit;
 
   /// Creates a new [AnonymousIdpConfig].
-  const AnonymousIdpConfig._({
-    required this.onBeforeAnonymousAccountCreated,
-    required this.onAfterAnonymousAccountCreated,
-    required this.perIpAddressRateLimitConfig,
+  const AnonymousIdpConfig({
+    this.onBeforeAnonymousAccountCreated,
+    this.onAfterAnonymousAccountCreated,
+    this.perIpAddressRateLimit = const RateLimit(
+      maxAttempts: 100,
+      timeframe: Duration(hours: 1),
+    ),
   });
-
-  /// Creates a new [AnonymousIdpConfig] with optional rate limiting. If no
-  /// rate limiting value is provided, none will be used and requests will not
-  /// be throttled.
-  factory AnonymousIdpConfig({
-    final BeforeAnonymousAccountCreatedFunction?
-    onBeforeAnonymousAccountCreated,
-    final AfterAnonymousAccountCreatedFunction? onAfterAnonymousAccountCreated,
-    final RateLimit? perIpAddressRateLimit,
-  }) => AnonymousIdpConfig._(
-    onBeforeAnonymousAccountCreated: onBeforeAnonymousAccountCreated,
-    onAfterAnonymousAccountCreated: onAfterAnonymousAccountCreated,
-    perIpAddressRateLimitConfig:
-        AnonymousIdpLoginRateLimitingConfig.fromRateLimit(
-          rateLimit:
-              perIpAddressRateLimit ??
-              AnonymousIdpLoginRateLimitingConfig.defaultRateLimit,
-        ),
-  );
-
-  /// Creates a new [AnonymousIdpConfig] with completely customized rate
-  /// limiting.
-  factory AnonymousIdpConfig.customRateLimiting({
-    required final AnonymousIdpLoginRateLimitingConfig
-    perIpAddressRateLimitConfig,
-    final BeforeAnonymousAccountCreatedFunction?
-    onBeforeAnonymousAccountCreated,
-    final AfterAnonymousAccountCreatedFunction? onAfterAnonymousAccountCreated,
-  }) => AnonymousIdpConfig._(
-    onBeforeAnonymousAccountCreated: onBeforeAnonymousAccountCreated,
-    onAfterAnonymousAccountCreated: onAfterAnonymousAccountCreated,
-    perIpAddressRateLimitConfig: perIpAddressRateLimitConfig,
-  );
 
   @override
   AnonymousIdp build({
@@ -97,53 +67,4 @@ class AnonymousIdpConfig extends IdentityProviderBuilder<AnonymousIdp> {
       userProfiles: userProfiles,
     );
   }
-}
-
-/// Rate limiting configuration for anonymous login.
-class AnonymousIdpLoginRateLimitingConfig
-    extends RateLimitedRequestAttemptConfig<String> {
-  /// Default maximum number of attempts within [defaultTimeframe].
-  static const defaultMaxAttempts = 100;
-
-  /// Default window for which attempts are capped to [defaultMaxAttempts].
-  static const defaultTimeframe = Duration(hours: 1);
-
-  /// Default [RateLimit].
-  static const defaultRateLimit = RateLimit(
-    maxAttempts: defaultMaxAttempts,
-    timeframe: defaultTimeframe,
-  );
-
-  /// Creates an instance of [AnonymousIdpLoginRateLimitingConfig].
-  ///
-  /// The parameters `nonceToString` and `nonceFromString` are not accepted
-  /// because the nonce type is `String`.
-  AnonymousIdpLoginRateLimitingConfig({
-    super.maxAttempts = defaultMaxAttempts,
-    super.timeframe = defaultTimeframe,
-    super.defaultExtraData,
-    super.onRateLimitExceeded,
-  }) : super(
-         domain: 'anonymous-idp-login',
-         source: 'account-creation',
-         nonceToString: (final String nonce) => nonce,
-         nonceFromString: (final String nonce) => nonce,
-       );
-
-  /// Creates an instance of [AnonymousIdpLoginRateLimitingConfig] from a
-  /// [RateLimit] value.
-  ///
-  /// The parameters `nonceToString` and `nonceFromString` are not accepted
-  /// because the nonce type is `String`.
-  factory AnonymousIdpLoginRateLimitingConfig.fromRateLimit({
-    required final RateLimit rateLimit,
-    final Map<String, String>? defaultExtraData,
-    final Future<void> Function(Session session, String nonce)?
-    onRateLimitExceeded,
-  }) => AnonymousIdpLoginRateLimitingConfig(
-    maxAttempts: rateLimit.maxAttempts,
-    timeframe: rateLimit.timeframe,
-    defaultExtraData: defaultExtraData,
-    onRateLimitExceeded: onRateLimitExceeded,
-  );
 }
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp_utils.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp_utils.dart
@@ -19,9 +19,20 @@ class AnonymousIdpUtils {
     required this.config,
     final AuthUsers authUsers = const AuthUsers(),
   }) : _authUsers = authUsers,
-       _rateLimitUtil = config.perIpAddressRateLimitConfig != null
+       _rateLimitUtil = config.perIpAddressRateLimit != null
            ? DatabaseRateLimitedRequestAttemptUtil(
-               config.perIpAddressRateLimitConfig!,
+               RateLimitedRequestAttemptConfig(
+                 domain: 'anonymous',
+                 source: 'account_creation',
+                 maxAttempts: config.perIpAddressRateLimit?.maxAttempts,
+                 timeframe: config.perIpAddressRateLimit?.timeframe,
+                 onRateLimitExceeded: (final session, final nonce) {
+                   throw AnonymousAccountBlockedException(
+                     reason:
+                         AnonymousAccountBlockedExceptionReason.tooManyAttempts,
+                   );
+                 },
+               ),
              )
            : null;
 
@@ -33,16 +44,10 @@ class AnonymousIdpUtils {
     final Transaction? transaction,
   }) async {
     // Check rate limit and either throw or proceed.
-    final tooManyAttempts = await _rateLimitUtil?.hasTooManyAttempts(
+    await _rateLimitUtil?.hasTooManyAttempts(
       session,
       nonce: session.remoteIpAddress.toString(),
     );
-
-    if (tooManyAttempts == true) {
-      throw AnonymousAccountBlockedException(
-        reason: AnonymousAccountBlockedExceptionReason.tooManyAttempts,
-      );
-    }
 
     final newUser = await _authUsers.create(
       session,
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
@@ -45,8 +45,8 @@ void main() {
       setUp(() async {
         session = sessionBuilder.build();
         fixture = AnonymousIdpTestFixture(
-          config: AnonymousIdpConfig(
-            perIpAddressRateLimit: const RateLimit(
+          config: const AnonymousIdpConfig(
+            perIpAddressRateLimit: RateLimit(
               maxAttempts: 2,
               timeframe: Duration(minutes: 1),
             ),
@@ -81,8 +81,8 @@ void main() {
       setUp(() async {
         session = sessionBuilder.build();
         fixture = AnonymousIdpTestFixture(
-          config: AnonymousIdpConfig(
-            perIpAddressRateLimit: const RateLimit(
+          config: const AnonymousIdpConfig(
+            perIpAddressRateLimit: RateLimit(
               maxAttempts: 1,
               timeframe: Duration(minutes: 1),
             ),
@@ -119,8 +119,8 @@ void main() {
       setUp(() async {
         session = sessionBuilder.build();
         fixture = AnonymousIdpTestFixture(
-          config: AnonymousIdpConfig(
-            perIpAddressRateLimit: const RateLimit(
+          config: const AnonymousIdpConfig(
+            perIpAddressRateLimit: RateLimit(
               maxAttempts: 1,
               timeframe: Duration(minutes: 1),
             ),
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/test_utils/anonymous_idp_test_fixture.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/test_utils/anonymous_idp_test_fixture.dart
@@ -24,7 +24,7 @@ final class AnonymousIdpTestFixture {
         ).tokenManager;
 
     anonymousIdp = AnonymousIdp(
-      config ?? AnonymousIdpConfig(),
+      config ?? const AnonymousIdpConfig(),
       tokenManager: this.tokenManager,
     );
   }


modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/anonymous/anonymous_sign_in_widget.dart
```diff
@@ -0,0 +1,111 @@
...
+          ),
+          const SizedBox(height: 10),
+          // The styling of this button is unimportant, because you would not
+          // show it to users in typical applications. It is only presented here
+          // to force users into anonymous, authenticated states for testing
+          // purposes.
+          OutlinedButton(
+            onPressed: _controller.login,
+            child: const Text('Sign in anonymously'),
+          ),
```
marcelomendoncasoares: Discussion: Although I agree that the best UX is to have this happen in the background and postpone showing a login wall, it might be a better developer experience to have this widget available inside the `SignInWidget` as the other IDPs. It would offer the same discovery and testing experience - just extend the endpoint, hit generate and see the button available on the app - and it is always possible to disable it with a flag.

Beyond a good developer experience, a "Continue without account" button could also be valuable to allow users to postpone signing-up if changing their minds after entering an optional sign-up screen.

My suggestions would be to have this button borderless in the bottom of the screen, after all social sign-ups, bearing the "Continue without account" text that serves either as a first-time option or an option to postpone sign-up.

examples/auth/auth_flutter/lib/main.dart
```diff
@@ -98,19 +98,31 @@ class SignInScreen extends StatelessWidget {
...
+          Padding(
+            padding: const EdgeInsets.only(top: 16),
+            child: AnonymousSignInWidget(
+              client: client,
+              onAuthenticated: () => onAuthenticated(context),
+              onError: (error) => onError(context, error),
+            ),
+          ),
+        ],
       ),
```
marcelomendoncasoares, line 122: Comment: If we go in the direction of baking the button inside the `SignInWidget` as suggested below, this change would no longer be necessary - which highlights the simplicity we aim for experimenting with IDPs from the developer perspective.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/anonymous/anonymous_auth_controller.dart
```diff
@@ -0,0 +1,48 @@
...
+  final VoidCallback? onAuthenticated;
+
+  /// Callback when an error occurs during authentication.
+  ///
+  /// The [error] parameter is an exception that should be shown to the user.
+  /// Exceptions that should not be shown to the user are shown in the debug
+  /// log, but not passed to the callback.
+  final Function(Object error)? onError;
+
+  ///
```
marcelomendoncasoares: Comment: I know the PR is in initial state, so just a reminder that this doc is empty :)

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/providers/anonymous.dart
```diff
@@ -0,0 +1,12 @@
+/// This library contains the Apple authentication provider for the
+/// Serverpod Idp module.
+library;
+
+export 'package:serverpod_auth_core_server/serverpod_auth_core_server.dart'
+    show AuthSuccess;
+
+export '../src/generated/protocol.dart' show AppleAccount;
```
marcelomendoncasoares:
```suggestion
export '../src/generated/protocol.dart'
    show
        AnonymousAccount,
        AnonymousAccountBlockedException,
        AnonymousAccountBlockedExceptionReason;
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/providers/anonymous.dart
```diff
@@ -0,0 +1,12 @@
+/// This library contains the Apple authentication provider for the
```
marcelomendoncasoares:
```suggestion
/// This library contains the Anonymous authentication provider for the
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
```diff
@@ -0,0 +1,29 @@
...
+    testGroupTagsOverride: TestTags.concurrencyOneTestTags,
+    (final sessionBuilder, final endpoints) {
+      late Session session;
+      late AnonymousIdpTestFixture fixture;
+      setUp(() {
+        session = sessionBuilder.build();
+        fixture = AnonymousIdpTestFixture();
+      });
+
+      test('can create a new account', () async {
```
marcelomendoncasoares: Suggestion: Just taking this test as example of the Given/when/then pattern we use for test descriptions - that should form a full sentence describing the test case:

```suggestion
      test('when logging in anonymously then a new account is created', () async {
```
> Preceded by "Given no account logged in"

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/utils/anonymous_idp_account_creation_util.dart
```diff
@@ -0,0 +1,130 @@
...
+class AnonymousIdpAccountCreationUtil {
+  final AnonymousIdpAccountCreationUtilsConfig _config;
+  final DatabaseRateLimitedRequestAttemptUtil<String>? _rateLimitUtil;
+  final AuthUsers _authUsers;
+
+  /// Creates a new [AnonymousIdpAccountCreationUtil] instance.
+  AnonymousIdpAccountCreationUtil({
+    required final AnonymousIdpAccountCreationUtilsConfig config,
+    required final AuthUsers authUsers,
+  }) : _config = config,
```
marcelomendoncasoares: Suggestion: Since the Anonymous IDP is basically account creation, I think we can simplify this architecture and mirror the one from Apple and Google: 1 config class, 1 utils class. Email has this separation of multiple utils due to its complexity, but the anonymous complexity that we expect to add lately (attestation) will most probably result in general utils (like the rate limit and secret challenge) than nested utils (like the email ones).

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/utils/anonymous_idp_account_creation_util.dart
```diff
@@ -0,0 +1,130 @@
...
+      authUserId: newUser.id,
+      scopes: newUser.scopes,
+    );
+  }
+}
+
+/// Configuration for the [AnonymousIdpAccountCreationUtil] class.
+class AnonymousIdpAccountCreationUtilsConfig {
+  /// {@macro before_anonymous_account_created_function}
+  final BeforeAnonymousAccountCreatedFunction? beforeAnonymousAccountCreated;
```
marcelomendoncasoares: Suggestion: Small suggestion for standardization, that also applies to the other config object.

```suggestion
  final BeforeAnonymousAccountCreatedFunction? onBeforeAnonymousAccountCreated;
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp.dart
```diff
@@ -0,0 +1,95 @@
...
+              await config.beforeAnonymousAccountCreated!(
+                session,
+                transaction: transaction,
+              );
+          if (!canCreateAnonymousAccount) {
+            throw AnonymousAccountBlockedException(
+              reason: AnonymousAccountBlockedExceptionReason.denied,
+            );
+          }
+        }
```
marcelomendoncasoares, line 72: Question: Shouldn't `login` pass an optional `nonce` for the `onBeforeAnonymousAccountCreated` to process? Otherwise, how could it define if an account can or cannot be created?

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/utils/anonymous_idp_account_creation_util.dart
```diff
@@ -0,0 +1,130 @@
...
+        nonce: '', // TODO: What to provide here?
+        transaction: transaction,
+      );
+
+      if (attemptCount > _config.rateLimit!.maxAttempts) {
+        throw AnonymousAccountBlockedException(
+          reason: AnonymousAccountBlockedExceptionReason.throttled,
+        );
+      }
+    }
```
marcelomendoncasoares: Suggestion: You can replace all this by the `hasTooManyAttempts` method of the `DatabaseRateLimitedRequestAttemptUtil`. Just configure the `onRateLimitExceeded` to throw (suggestion above) and you will have the same result with less boilerplate.

Regarding the `TODO`, if we pass the optional nonce from the login method, we just reuse it here in the rate limit.

```suggestion
    await _rateLimitUtil?.hasTooManyAttempts(
      session,
      nonce: '', // TODO: What to provide here?
      transaction: transaction,
    );
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/utils/anonymous_idp_account_creation_util.dart
```diff
@@ -0,0 +1,130 @@
...
+  }) : _config = config,
+       _authUsers = authUsers,
+       _rateLimitUtil = config.rateLimit != null
+           ? DatabaseRateLimitedRequestAttemptUtil(
+               RateLimitedRequestAttemptConfig(
+                 domain: 'anonymous',
+                 source: 'account_creation',
+                 maxAttempts: config.rateLimit?.maxAttempts,
+                 timeframe: config.rateLimit?.timeframe,
+               ),
```
marcelomendoncasoares: Suggestion: Add the throw here to reduce the boilerplate below.

```suggestion
               RateLimitedRequestAttemptConfig(
                 domain: 'anonymous',
                 source: 'account_creation',
                 maxAttempts: config.rateLimit?.maxAttempts,
                 timeframe: config.rateLimit?.timeframe,
                 onRateLimitExceeded: (session, nonce) {
                   throw AnonymousAccountBlockedException(
                     reason: AnonymousAccountBlockedExceptionReason.throttled,
                   );
                 },
               ),
```

marcelomendoncasoares (changes requested):
Implementation looks very solid, in line with the discussed specs and following the IDP package patterns. Just a few minor suggestions for improvement and two points for question/discussion!

As we discussed, I focused more on the design and architecture, but I don't think we need much more on the internals to consider it done,

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/anonymous/anonymous_sign_in_widget.dart
craiglabenz: I debated almost exactly that, so that seems reasonable to me. I can make that change.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp.dart
craiglabenz: I think that would position us awkwardly between the two bookend architectures for eventual endpoint protection. This is how I'm thinking about it.

On the one hand is my originally proposed architecture, where anonymous auth is a two-step process, with the first step being only required if endpoint protection is specified in the config. (If it is not configured, the developer simply calls `login`, does not pass the optional `token`, and the account is always created).

On the other hand is (at least how I understood) Alex's recommendation, where the anonymous auth creation is always just 1 method with no special arguments. In this world, attestation is handled by an optional callback (`onBeforeAnonymousAccountCreated`), which will know how to verify app attestation based on the `Session` object alone. This could mean that a specific header is passed which that method can decode, or a database flag is set for that exact session, or... anything else. But in this world, it is up to the developer to create a different endpoint which establishes proof of a valid app installation and then up to `onBeforeAnonymousAccountCreated` to go find and read that proof.

After a lot of thinking about this spectrum, I ended up agreeing with Alex's proposed design. (Or, again, at least how I understood his comments 😆). Thus, once we get to the app attestation phase, I expect to provide a version of this callback that reads proof of attestation as set by Serverpod's default app attestation flow.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/utils/anonymous_idp_account_creation_util.dart
craiglabenz: Did you mean to remove `recordAttempt` in this edit?

craiglabenz: Also, the rate limiting is entirely IP address-based, so in later efforts I've come to think that this needs to supply the IP address as the `nonce`, but I want to discuss that further.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp.dart
marcelomendoncasoares: Although you are right that we could pass some data around the `login` endpoint to retrieve on the callback, it seems to me that it would complicate a lot what could be just an optional parameter on the method. I guess your first suggestion of having the parameter is the simplest and was what Alex also suggested as well in the comments:

> Simplest possible solution:
> 
> 1. Login takes in an optional string,
> 2. Before a new auth use is created in the module a callback is called with that string. The callback can either return null, an auth id or throw.
> 3a. If null is returned a new auth user is created before a new token is issued.
> 3b. If an auth id is returned a token is issued connected to that auth id.
> 3c. If something was thrown the login fails.
>
> I believe this would allow this system to plug in any attestation logic at a later state. If you want to use app check and check that as part of the anonymous auth you could use that.

It also allows the usage of any existing attestation mechanism by users right with the first version of the feature, without depending exclusively in our future attestation solution.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/utils/anonymous_idp_account_creation_util.dart
marcelomendoncasoares: > Did you mean to remove `recordAttempt` in this edit?

The `hasTooManyAttempts` method already records one attempt before checking if should rate limit, so it greatly simplifies the usage of the utils.

> Also, the rate limiting is entirely IP address-based, so in later efforts I've come to think that this needs to supply the IP address as the `nonce`, but I want to discuss that further.

Our concerns with using IP-based rate limiting is that it can either generate problematic false-positives or not be effective at all - in case of real attackers. Some possible problems are:

- Shared IPs (NAT, corporate networks, universities, mobile carrier CGNAT) cause many legitimate users to share the same limit bucket, so one abusive client can throttle everyone behind that IP.
- Dynamic / rotated IPs (mobile networks, residential ISPs) make the limit less effective against a single attacker that reconnects or uses multiple IPs.
- Proxies / VPNs can concentrate many attackers on a single IP or hide the real source, making IP limits noisy and sometimes abusable.

So we must ensure that default limits are generous enough to avoid most false positives - like the one you suggested from Firebase of 100 per IP per hour. To get the IP address as `nonce`, you can pass the `session.remoteIpAddress` (from `utils/session_extension.dart`).

To reduce the potential problems, we might use separate rate limit utils, with the other that receives the `nonce` from the `login` method being called before the IP-based one.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp.dart
craiglabenz: Ah yes, that makes sense to me. I think I must have lost track of the comment you embedded here.

craiglabenz: > 3b. If an auth id is returned a token is issued connected to that auth id.

What is the scenario where this workflow would hit? Is it even possible for a brand new anonymous account creation attempt to end up pointing at an existing `AuthUser`?

marcelomendoncasoares: Maybe allow logging in back to an anonymous account given the `nonce` is the same. Like after reinstalling an app from a device and using a device fingerprint as `nonce`.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
```diff
@@ -0,0 +1,147 @@
...
+
+import '../../test_tags.dart';
+import '../../test_tools/serverpod_test_tools.dart';
+import '../test_utils/anonymous_idp_test_fixture.dart';
+
+void main() {
+  withServerpod(
+    'Given no account',
+    rollbackDatabase: RollbackDatabase.disabled,
+    testGroupTagsOverride: TestTags.concurrencyOneTestTags,
```
craiglabenz, line 15: @marcelomendoncasoares I'm having trouble holding this testing API correctly. Because my rate limiting tests assume an empty database table, (I think) I need to switch this `withServerpod` invocation to look like this:

```dart
withServerpod(
    'Given no account',
    rollbackDatabase: RollbackDatabase.afterEach,
    testGroupTagsOverride: TestTags.concurrencyOneTestTags,
    (sessionBuilder, endpoints) {...},
 )
```

However, when I do that, I get this error:

```
Concurrent calls to transaction are not supported when database rollbacks are enabled. Disable rolling back the database by setting `rollbackDatabase` to `RollbackDatabase.disabled`.
```

I would have expected the `TestTags.concurrencyOneTestTags` disclaimer to have removed concurrent use of the transaction.

Note that I'm also invoking the tests by clicking `Run` in my IDE, which maybe means I am missing some command line flags?

marcelomendoncasoares: What we usually do is to add a `tearDown` step that cleans up the database tables that the test requires when using `RollbackDatabase.disabled`. Since you are using the fixture, you can add a `tearDown` method to it similar to the one on the `EmailIdpTestFixture` and then invoke it as tear down:

```dart
tearDown(() async {
  await fixture.tearDown(session);
});
  ```
  
  See the `serverpod_auth_idp_server/test/email/email_idp/email_idp_login_test.dart` as an example.
  
Using the `run` command from the IDE will only be a problem if you're invoking the tests from a test explorer - since there the `--concurrency=1` won't be added for the tagged tests. But it works fine when running the tests using lens on the tests files, as they run in sequence already (at least using Cursor/VS Code). To run a broader suite of tests, I suggest going for the terminal:

```bash
# Tests with concurrency allowed
dart test -x concurrency_one

# Tests that require concurrency one
dart test -t concurrency_one --concurrency=1
```

craiglabenz: God bless!

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp.dart
craiglabenz: As per offline conversation, I have changed this function signature to accept a `String? token` and have removed the `UuidValue?` return type from on `onBeforeAnonymousAccountCreated`.

craiglabenz:
@marcelomendoncasoares There are a lot of failing e2e tests (at least as per the previous CI execution), but I can't tell whether those connection errors have anything to do with this PR. Is the CI suite having issues this morning?

marcelomendoncasoares:
> @marcelomendoncasoares There are a lot of failing e2e tests (at least as per the previous CI execution), but I can't tell whether those connection errors have anything to do with this PR. Is the CI suite having issues this morning?

I see you already fixed most of the tests! The last one is probably due to missing a migration in the `tests/serverpod_auth_test/serverpod_auth_test_server` project.

craiglabenz:
Now with significantly more green checkmarks 😄

examples/auth/auth_server/lib/server.dart
```diff
@@ -63,12 +64,22 @@ void run(List<String> args) async {
     hostname: 'localhost',
   );
 
+  final anonymousIdpConfig = AnonymousIdpConfig(
+    // Configure rate limiting to protect the anonymous login endpoint from
+    // unlimited abuse.
+    // perIpAddressRateLimit: RateLimit(
+    //   maxAttempts: 100,
+    //   timeframe: const Duration(hours: 1),
+    // ),
```
marcelomendoncasoares: Suggestion: Since we already have this as the default value, we can remove the comments here and leave for users to discover in a second-look - just as we do with all other optional configurations of the other IDPs.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
```diff
@@ -0,0 +1,164 @@
...
+      test(
+        'when logging in anonymously then a new account is created',
+        () async {
+          final result = fixture.anonymousIdp.login(session);
+          await expectLater(result, completion(isA<AuthSuccess>()));
+        },
+      );
+
+      test(
+        'when logging in is not blocked by acceptable amount of attempts',
```
marcelomendoncasoares: Suggestion: We should separate the tests in this file into separate "given" groups:

- Given login request within configured rate limit allowance
  - when loggin in
    - then it is allowed.
- Given login request exceeding configured rate limit 
  - when loggin in
    - then it is blocked with too many attempts exception.
- Given maximum allowed login attempts of 1 for anonymous login
  - when logging in from different IP addresses
    - then all login attempts succeed.
- Given onBeforeAnonymousAccountCreated configured to block login attempts
  - when loggin in
    - then it is blocked with AnonymousAccountBlockedException exception.

I might have forgotten one test case or the other in the suggested descriptions above, but the idea is that every test configuration consists into a different "given" to better follow the tests pattern. The "when" part should only contain the action - not change the scenario -, and the then should only contain the result.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
```diff
@@ -0,0 +1,164 @@
...
+              domain: 'anonymous',
+              source: 'account_creation',
+              nonce: session.remoteIpAddress.toString(),
+              ipAddress: session.remoteIpAddress.toString(),
+              attemptedAt: attemptedAt,
+            ),
+          );
+          final result = rateLimitedFixture.anonymousIdp.login(session);
+          await expectLater(result, completion(isA<AuthSuccess>()));
+        },
```
marcelomendoncasoares: Suggestion: Using the `RateLimitedRequestAttempt` in this test is an implementation detail of how the rate limiting is implemented for the anonymous IDP. Ideally, we want our tests to survive refactors as much as possible by avoiding all implementation details. A way to achieve this is to call a login on the setup of the group and call it again inside the test.

Take the email login test as an inspiration:

https://github.com/serverpod/serverpod/blob/44efa73273e3fcc832b30f8811f79c180c36e669/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/email/email_idp/email_idp_login_test.dart#L149-L218

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
```diff
@@ -0,0 +1,164 @@
...
+                }) => false,
+          ),
+        );
+        await expectLater(
+          () => cannotCreateFixture.anonymousIdp.login(session),
+          throwsA(isA<AnonymousAccountBlockedException>()),
+        );
+      });
+    },
+  );
```
marcelomendoncasoares, line 248: Question: Do we really need two tests for this? From my perspective, having the async should be enough.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/models/exceptions/anonymous_account_blocked_exception_reason.spy.yaml
```diff
@@ -0,0 +1,11 @@
...
+enum: AnonymousAccountBlockedExceptionReason
+serialized: byName
+values:
+  ### Exception to be thrown when the [beforeAnonymousAccountCreation]
+  ### callback returns false.
+  - denied
+
+  ### Exception to be thrown when the anonymous account creation request
+  ### exceeds the configured throttling limits.
+  - throttled
```
marcelomendoncasoares: Suggestion: To mirror the same name we have with email exceptions and create a predictable pattern.

```suggestion
  - tooManyAttempts
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/anonymous/anonymous_sign_in_widget.dart
```diff
@@ -0,0 +1,96 @@
...
+  }) : assert(
+         (controller == null || client == null),
+         'Either controller or client must be provided, but not both. When '
+         'passing a controller, client, onAuthenticated, and onError '
+         'parameters are ignored.',
+       ),
+       assert(
+         (onAuthenticated == null && onError == null) || controller == null,
+         'Provided onAuthenticated or onError when using a controller '
+         'as they are handled by the controller and will be ignored.',
```
marcelomendoncasoares: Suggestion: Reading this I just realized that the description is wrong in the other controllers as well. Would you mind fixing them here together?

```suggestion
         'Do not provide onAuthenticated or onError when using a controller '
         'as they will be handled by the controller and will be ignored.',
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/endpoints/anonymous_idp_base_endpoint.dart
```diff
@@ -0,0 +1,25 @@
...
+  AnonymousIdp get anonymousIdp => AuthServices.instance.anonymousIdp;
+
+  /// {@template anonymous_account_base_endpoint.login}
+  /// Creates a new anonymous account and returns its session.
+  ///
+  /// Invokes the [AnonymousIdp.beforeAnonymousAccount] callback if configured,
+  /// which may prevent account creation if the endpoint is protected.
+  /// {@endtemplate}
+  Future<AuthSuccess> login(final Session session) async {
+    return anonymousIdp.login(session);
```
marcelomendoncasoares: Suggestion: Missing the `String? token` parameter to allow plugging attestation solutions in.

```suggestion
  Future<AuthSuccess> login(final Session session, {String? token}) async {
    return anonymousIdp.login(session, token: token);
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp_config.dart
```diff
@@ -0,0 +1,71 @@
+import 'dart:async';
+
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_idp_server/src/providers/anonymous/business/anonymous_idp.dart';
+import '../../../../../core.dart';
+
+// Ideally, this should be in a shared package
+import '../../../../providers/email.dart' show RateLimit;
```
marcelomendoncasoares: Suggestion: Move the `RateLimit` class to the `rate_limited_request_attempt_config.dart` file under `common`. Just ensure that the class is exported on the provider barrel file, so that this is no breaking change.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp.dart
```diff
@@ -0,0 +1,97 @@
...
+              await config.onBeforeAnonymousAccountCreated!(
+                session,
+                token: token,
+                transaction: transaction,
+              );
+          if (!canCreateAccount) {
+            throw AnonymousAccountBlockedException(
+              reason: AnonymousAccountBlockedExceptionReason.denied,
+            );
+          }
```
marcelomendoncasoares: Discussion: Since the function can be used to run other routines than just validating the token, should we change it to `Future<void>` and expect it to throw appropriate exceptions when something is off - which will properly log the exception on the server - instead of returning bool? We would only need to catch here and replace by the `AnonymousAccountBlockedException` as you are throwing.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/anonymous/anonymous_auth_controller.dart
```diff
@@ -0,0 +1,90 @@
...
+        'No anonymous authentication endpoint found. Make sure you have '
+        'extended and exposed "EndpointAnonymousIdpBase" in your server.',
+      );
+    }
+  }
+
+  /// Initiates the anonymous sign-in process.
+  Future<void> login() async {
+    await _guarded(() async {
+      final authSuccess = await _anonymousEndpoint.login();
```
marcelomendoncasoares: Suggestion: We should have an optional `Future<String?> createAnonymousToken` callback on the controller that can be used to plug in attestation packages - take the name as a suggestion, but feel free to change it if you have a better one.

```suggestion
      final maybeToken = await createAnonymousToken();
      final authSuccess = await _anonymousEndpoint.login(token: maybeToken);
```

This will be the one to allow passing a token that will reach the `onBeforeAnonymousAccountCreated` callback on the server for validation. The parameter must also be exposed on the widget, if users prefer not managing the controller themselves - as we do with the other IDPs.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/sign_in_widget.dart
```diff
@@ -148,6 +159,15 @@ class _SignInWidgetState extends State<SignInWidget> {
               ),
         if (socialProviders.isNotEmpty && hasEmail) const _SignInSeparator(),
         ...socialProviders,
+        if (hasAnonymous) ...[
+          const Spacer(),
```
marcelomendoncasoares: Suggestion: Remove the spacer to avoid pushing the top providers to the top of the screen. 

image

Without the spacer:

image

It also makes the spacing between the widgets consistent in case the height available is just a little bit more than it would require to render all IDPs.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
```diff
@@ -0,0 +1,164 @@
...
+
+      test('is blocked by synchronous beforeAnonymousAccountCreated', () async {
+        final cannotCreateFixture = AnonymousIdpTestFixture(
+          config: AnonymousIdpConfig(
+            onBeforeAnonymousAccountCreated:
+                (
+                  final Session session, {
+                  final String? token,
+                  final Transaction? transaction,
+                }) => false,
```
marcelomendoncasoares: Suggestion: We should also have a test that assert that the token passed on `login` arrives as expected at `onBeforeAnonymousAccountCreated`.

marcelomendoncasoares (changes requested):
Great work, @craiglabenz! Think after iterating on these points we'll be good to merge! I was also going to add a few more comments on the widget/layout, but I think it will be faster if I commit a few suggestions - since I have already done them locally to experiment.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/anonymous/anonymous_sign_in_widget.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/4495/commits/41bb8a69edca05dafc34422719924335e97f06bf.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/sign_in_widget.dart
marcelomendoncasoares: I have already changed at https://github.com/serverpod/serverpod/pull/4495/commits/d4892317fc67880dc127da081edcb497c7c080b5, but let me know what you think!

examples/auth/auth_server/lib/server.dart
craiglabenz: Good point!

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
craiglabenz: Great call, that's smart

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
craiglabenz: Possibly not - I almost never use FutureOr (but I think this method should use it), so I was possibly over-zealous.

Removing.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
craiglabenz: I was able to do this for all of the tests *except* the one which requires a different IP address. However, if there is a cleaner way to mock the IP address of a request, then I could potentially apply that method to the different IP address as well and thus remove the final place where a raw rate limiting request attempt is inserted into the database.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp_config.dart
craiglabenz: Nice, this refactor improved things.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp.dart
craiglabenz: Agreed, that's cleaner and better.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/anonymous/anonymous_auth_controller.dart
craiglabenz: Good thinking. Implemented!

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/sign_in_widget.dart
craiglabenz: SGTM

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
craiglabenz: I've made adjustments in light of these comments. PTAL :)

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/anonymous/anonymous_sign_in_widget.dart
```diff
@@ -0,0 +1,129 @@
...
+  /// In this case, [client] must be provided.
+  ///
+  /// If provided, the widget uses this controller instead of creating one,
+  /// and [client], [onAuthenticated], and [onError] are ignored.
+  final AnonymousAuthController? controller;
+
+  /// The Serverpod client instance.
+  ///
+  /// Required when [controller] is null, ignored otherwise.
+  final ServerpodClientShared? client;
```
marcelomendoncasoares, line 28: Missing the new `createAnonymousToken` parameter here for usage of the widget without managing the controller externally.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp_config.dart
```diff
@@ -0,0 +1,130 @@
+import 'dart:async';
+
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod_auth_idp_server/src/providers/anonymous/business/anonymous_idp.dart';
+import '../../../../../core.dart';
```
marcelomendoncasoares: Suggestion(NIT): Separate package and relative imports.

```suggestion
import 'package:serverpod_auth_idp_server/src/providers/anonymous/business/anonymous_idp.dart';

import '../../../../../core.dart';
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp_config.dart
```diff
@@ -0,0 +1,130 @@
...
+  final BeforeAnonymousAccountCreatedFunction? onBeforeAnonymousAccountCreated;
+
+  /// {@macro after_anonymous_account_created_function}
+  final AfterAnonymousAccountCreatedFunction? onAfterAnonymousAccountCreated;
+
+  /// {@template anonymous_idp_config.rate_limit}
+  /// The maximum rate of anonymous accounts that can be created from a single
+  /// IP address.
+  /// {@endtemplate}
+  final AnonymousIdpLoginRateLimitingConfig? perIpAddressRateLimitConfig;
```
marcelomendoncasoares: Discussion: I think the previous parameter with type `RateLimit` was clearer and simpler for its purpose. I understand that your idea here was to also expose the `defaultExtraData` and `onRateLimitExceeded` to the user, but do we have a use case for this in mind? Or can we keep the simpler approach, as we do with all rate limiting on email?

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp_config.dart
```diff
@@ -0,0 +1,130 @@
...
+    onBeforeAnonymousAccountCreated,
+    final AfterAnonymousAccountCreatedFunction? onAfterAnonymousAccountCreated,
+    final AnonymousIdpLoginRateLimitingConfig? perIpAddressRateLimitConfig,
+  }) => AnonymousIdpConfig(
+    onBeforeAnonymousAccountCreated: onBeforeAnonymousAccountCreated,
+    onAfterAnonymousAccountCreated: onAfterAnonymousAccountCreated,
+    perIpAddressRateLimitConfig:
+        perIpAddressRateLimitConfig ??
+        AnonymousIdpLoginRateLimitingConfig.defaultConfig(),
+  );
```
marcelomendoncasoares: Suggestion: I would prefer that we have the sensible default of 100 attempts/hour than having the default as null (less secure) and requiring users to always pass a `rateLimited` variant of the configs. Our directive is to have users consciously lower the security requirements than having it low by default.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp_utils.dart
```diff
@@ -0,0 +1,86 @@
...
+  /// Creates a new instance of [AnonymousIdpUtils].
+  AnonymousIdpUtils({
+    required this.config,
+    final AuthUsers authUsers = const AuthUsers(),
+  }) : _authUsers = authUsers,
+       _rateLimitUtil = config.perIpAddressRateLimitConfig != null
+           ? DatabaseRateLimitedRequestAttemptUtil(
+               config.perIpAddressRateLimitConfig!,
+             )
+           : null;
```
marcelomendoncasoares: Comment: If we agree on simplifying the rate limit configs back, than we can also return the `RateLimitedRequestAttemptConfig` creation here.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
marcelomendoncasoares: I don't think we have a way to mock the IP address. In this case, we just need to ensure that we lock the implementation in the test in a way that changing it will lead to the test breaking. I'll commit this to the test together with a few suggestions in continuation of the above discussion.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/anonymous/anonymous_idp/anonymous_idp_login_test.dart
marcelomendoncasoares: Nice! I think we can still improve a little to match the pattern of the repository. I'll commit a suggestion of reorganization of the groups and descriptions as an example. :)

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/anonymous/anonymous_sign_in_widget.dart
craiglabenz: Great catch.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/anonymous/business/anonymous_idp_utils.dart
craiglabenz: I think I either am not quite following your train of thought here or I may disagree. I've implemented your other suggestions, but I think this should stay the same. Here is my rationale -- let me know if I am missing something.

1. The default, zero-parameter constructor call for `AnonymousIdpConfig()` now has rate limiting.
2. This means that if the configuration is missing, the developer turned it off by default.
3. Therefore, if `config.perIpAddressRateLimitConfig` is null, falling back to a default `RateLimitedRequestAttemptConfig` would essentially ignore their decision to opt out of rate limiting.

Does that make sense? Or, possibly, were you thinking about this from a different angle?

marcelomendoncasoares: My suggestion here was actually to revert the code to what is was previously, that was simpler and more straightforward (and had the same logic of 1 and 2):

```dart
  AnonymousIdpUtils({
    required this.config,
    final AuthUsers authUsers = const AuthUsers(),
  }) : _authUsers = authUsers,
       _rateLimitUtil = config.perIpAddressRateLimit != null
           ? DatabaseRateLimitedRequestAttemptUtil(
               RateLimitedRequestAttemptConfig(
                 domain: 'anonymous',
                 source: 'account_creation',
                 maxAttempts: config.perIpAddressRateLimit?.maxAttempts,
                 timeframe: config.perIpAddressRateLimit?.timeframe,
                 onRateLimitExceeded: (final session, final nonce) {
                   throw AnonymousAccountBlockedException(
                     reason: AnonymousAccountBlockedExceptionReason.throttled,
                   );
                 },
               ),
             )
           : null;
```

But I get that you might be envisioning some usage for the `defaultExtraData` and `onRateLimitExceeded` callbacks, right? Which is the only reason to have the rate limit config as a separate entity that exposes them instead of created inline from only the `RateLimit` as we had here.

craiglabenz: Mostly correct, yes, but I also want your perspective on that. My assumption was that the extraData Map and specific callback function exist to be user-specified at some point, and so I didn't want to wall them off from developers. But if they're not actually intended to be exposed and are functionally internal then I am not committed to breaking that pattern and exposing them here.

marcelomendoncasoares: The main goal for these parameters was actually to be used by implementations that use the rate limit - like the email and anon IDPs. It is possible to expose them for users if we have clear use cases, but I can't think of one in the current case. On the email IDP, you can see that these parameters are implementation details and hidden away from users, that have access only to the `RateLimit` parameters. So, if we don't have a clear use case, my take would be to simplify the config to the way it was before and keep the pattern.

craiglabenz: This is now back in that earlier original state before the miscommunication :smile:

craiglabenz:
No worries, it's been a fun start to contributing to Serverpod. Feel free to click Merge whenever you're ready :)

Issue #4378, opened 2025-12-06
Flaky `serverpod_cli` directory finder test failing due to cross-contamination with other tests


marcelomendoncasoares:
After fixing #4301 by moving the CLI endpoint generator tests to the temp folder (#4327), we now have another flaky test of the "generate from anywhere" feature. [First occurrence happened](https://github.com/serverpod/serverpod/actions/runs/19979142711/job/57302281200?pr=4088) on PR #4088:

```
==== Running tests in ./tools/serverpod_cli ====
+3733 -1: test\util\server_directory_finder_interactive_test.dart: ServerDirectoryFinder.findOrPrompt with interactive flag Given no server directory exists, when findOrPrompt is called with interactive=false, then it throws ServerpodProjectNotFoundException [E]
  Expected: throws <Instance of 'ServerpodProjectNotFoundException'> with `message`: contains 'No Serverpod server project detected'
    Actual: <Closure: () => Future<Directory>>
     Which: returned a Future that emitted _Directory:<Directory: 'C:\Users\runneradmin\AppData\Local\Temp\dart_test_a8b0b45a\test_server'>
  
  package:test_api  OutstandingWork.complete
  
+3814 -1: Some tests failed.

Consider enabling the flag chain-stack-traces to receive more detailed exceptions.
For example, 'dart test --chain-stack-traces'.
```

The error is happening on Windows, [what is something we have faced during the fix](https://github.com/serverpod/serverpod/pull/4327#issuecomment-3604865622) regarding shortname resolution (8.3 filename). It should have been fixed [by calling `FileSystemEntity.resolveSymbolicLinks`](https://github.com/serverpod/serverpod/pull/4327#discussion_r2588118638), but then we would not be seeing this again.

marcelomendoncasoares:
The error happened again on multiple PRs, [including the #4501 to bump dependencies](https://github.com/serverpod/serverpod/actions/runs/20492496279/job/58887047865?pr=4501).

```
==== Running tests in ./tools/serverpod_cli ====
WARNING: Migration Warnings:
 • Column "toBeDropped" of table "example" will be dropped.
Migration aborted. Use --force to ignore warnings.
WARNING: Migration Warnings:
 • Column "toBeDropped" of table "example" will be dropped.
No changes detected.
[...]

PR #4506, opened 2025-12-25
fix: Fix flaky `serverpod_cli` directory finder test

marcelomendoncasoares:
Fixes: #4378

This PR is a work in progress and was meant to reproduce the bug with added prints so we can see the reason of the failure. However, CI has already run 5 times with no failure yet...

## Pre-launch Checklist

## Breaking changes

None.


marcelomendoncasoares:
It is possible that, by cluttering the directory search code with prints to track what is happening, we are slowing down its execution and the racing condition is not happening... There goes the observer changing the experiment.

I'll update the tracking changes to postpone to a single print only when finishing the search.

marcelomendoncasoares:
Closing since the issue was solved by #4585.

PR #4520, opened 2025-12-30
feat: Add connected IDPs lookup

craiglabenz:
[Feature design doc](https://docs.google.com/document/d/1hHURFDIIBIvO5n7bf7QhJ1Gbz9D9P2VklsEV28oES7o/edit?usp=sharing).

This feature is a component of Phase 2 of the anonymous auth IDP project, which relates to linking IDPs to existing accounts. This targeted PR adds the ability for the frontend to lookup which IDPs a user has connected for the purposes of conditionally rendering buttons to add not-yet-connected IDPs.

I will add tests once we agree upon the basic architecture.

## Design decision to consider

1. This PR introduces the idea of an abstract `Idp` with a bare minimum contract for looking up the existence of a user's account. This would technically be a breaking change for users of the new auth module who have added their own IDPs, albeit, not a challenging one to overcome. **Still, is this feature worth a minor breaking change?**
  1.a. I believe that, if we were to opt out of this abstract class which ensures the existence of `hasAccount`, then this task of fetching all connected accounts would be much more tedious (especially for developers who have added their own IDPs).
  
2. This PR adds a new `IdpEndpoint.idpAccounts()` method. Is this in the correct place and is there a more idiomatic name that we might consider?

## Pre-launch Checklist

## Breaking changes

Possibly the addition of a new required method on IDP classes which decide to extend `Idp`.

### DIFF
--- a/tests/serverpod_auth_test/serverpod_auth_test_client/lib/src/protocol/client.dart
+++ b/tests/serverpod_auth_test/serverpod_auth_test_client/lib/src/protocol/client.dart
@@ -10,13 +10,13 @@
 // ignore_for_file: invalid_use_of_internal_member
 
 // ignore_for_file: no_leading_underscores_for_library_prefixes
-import 'package:serverpod_client/serverpod_client.dart' as _i1;
-import 'dart:async' as _i2;
-import 'package:serverpod_auth_core_client/serverpod_auth_core_client.dart'
-    as _i3;
-import 'package:serverpod_auth_client/serverpod_auth_client.dart' as _i4;
 import 'package:serverpod_auth_idp_client/serverpod_auth_idp_client.dart'
-    as _i5;
+    as _i1;
+import 'package:serverpod_client/serverpod_client.dart' as _i2;
+import 'dart:async' as _i3;
+import 'package:serverpod_auth_core_client/serverpod_auth_core_client.dart'
+    as _i4;
+import 'package:serverpod_auth_client/serverpod_auth_client.dart' as _i5;
 import 'dart:typed_data' as _i6;
 import 'package:serverpod_auth_bridge_client/serverpod_auth_bridge_client.dart'
     as _i7;
@@ -24,31 +24,74 @@ import 'package:serverpod_auth_migration_client/serverpod_auth_migration_client.
     as _i8;
 import 'protocol.dart' as _i9;
 
+/// Endpoint for Apple-based authentication.
+/// {@category Endpoint}
+class EndpointAppleAccount extends _i1.EndpointAppleIdpBase {
+  EndpointAppleAccount(_i2.EndpointCaller caller) : super(caller);
+
+  @override
+  String get name => 'appleAccount';
+
+  /// Signs in a user with their Apple account.
+  ///
+  /// If no user exists yet linked to the Apple-provided identifier, a new one
+  /// will be created (without any `Scope`s). Further their provided name and
+  /// email (if any) will be used for the `UserProfile` which will be linked to
+  /// their `AuthUser`.
+  ///
+  /// Returns a session for the user upon successful login.
+  @override
+  _i3.Future<_i4.AuthSuccess> login({
+    required String identityToken,
+    required String authorizationCode,
+    required bool isNativeApplePlatformSignIn,
+    String? firstName,
+    String? lastName,
+  }) => caller.callServerEndpoint<_i4.AuthSuccess>(
+    'appleAccount',
+    'login',
+    {
+      'identityToken': identityToken,
+      'authorizationCode': authorizationCode,
+      'isNativeApplePlatformSignIn': isNativeApplePlatformSignIn,
+      'firstName': firstName,
+      'lastName': lastName,
+    },
+  );
+
+  @override
+  _i3.Future<bool> hasAccount() => caller.callServerEndpoint<bool>(
+    'appleAccount',
+    'hasAccount',
+    {},
+  );
+}
+
 /// Endpoint for testing authentication.
 /// {@category Endpoint}
-class EndpointAuthTest extends _i1.EndpointRef {
-  EndpointAuthTest(_i1.EndpointCaller caller) : super(caller);
+class EndpointAuthTest extends _i2.EndpointRef {
+  EndpointAuthTest(_i2.EndpointCaller caller) : super(caller);
 
   @override
   String get name => 'authTest';
 
   /// Creates a new test user.
-  _i2.Future<_i1.UuidValue> createTestUser() =>
-      caller.callServerEndpoint<_i1.UuidValue>(
+  _i3.Future<_i2.UuidValue> createTestUser() =>
+      caller.callServerEndpoint<_i2.UuidValue>(
         'authTest',
         'createTestUser',
         {},
       );
 
   /// Creates a new session authentication for the test user.
-  _i2.Future<_i3.AuthSuccess> createSasToken(_i1.UuidValue authUserId) =>
-      caller.callServerEndpoint<_i3.AuthSuccess>(
+  _i3.Future<_i4.AuthSuccess> createSasToken(_i2.UuidValue authUserId) =>
+      caller.callServerEndpoint<_i4.AuthSuccess>(
         'authTest',
         'createSasToken',
         {'authUserId': authUserId},
       );
 
-  _i2.Future<void> deleteSasTokens(_i1.UuidValue authUserId) =>
+  _i3.Future<void> deleteSasTokens(_i2.UuidValue authUserId) =>
       caller.callServerEndpoint<void>(
         'authTest',
         'deleteSasTokens',
@@ -56,15 +99,15 @@ class EndpointAuthTest extends _i1.EndpointRef {
       );
 
   /// Creates a new JWT token for the test user.
-  _i2.Future<_i3.AuthSuccess> createJwtToken(_i1.UuidValue authUserId) =>
-      caller.callServerEndpoint<_i3.AuthSuccess>(
+  _i3.Future<_i4.AuthSuccess> createJwtToken(_i2.UuidValue authUserId) =>
+      caller.callServerEndpoint<_i4.AuthSuccess>(
         'authTest',
         'createJwtToken',
         {'authUserId': authUserId},
       );
 
   /// Deletes all refresh tokens for the test user.
-  _i2.Future<void> deleteJwtRefreshTokens(_i1.UuidValue authUserId) =>
+  _i3.Future<void> deleteJwtRefreshTokens(_i2.UuidValue authUserId) =>
       caller.callServerEndpoint<void>(
         'authTest',
         'deleteJwtRefreshTokens',
@@ -72,7 +115,7 @@ class EndpointAuthTest extends _i1.EndpointRef {
       );
 
   /// Destroys a specific refresh token by ID.
-  _i2.Future<bool> destroySpecificRefreshToken(String token) =>
+  _i3.Future<bool> destroySpecificRefreshToken(String token) =>
       caller.callServerEndpoint<bool>(
         'authTest',
         'destroySpecificRefreshToken',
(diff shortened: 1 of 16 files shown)


craiglabenz:
cc @marcelomendoncasoares For a review of this building block for [Phase 2](https://docs.google.com/document/d/1hHURFDIIBIvO5n7bf7QhJ1Gbz9D9P2VklsEV28oES7o/edit?tab=t.0#heading=h.3cjye6bqivi9) on linking new IDPs to existing users.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/endpoints/idp_endpoint.dart
```diff
@@ -0,0 +1,48 @@
...
+        session.log(
+          'Cannot check accounts for provider $provider because it does '
+          'not subclass Idp from pkg:serverpod_auth_idp_server',
+        );
+        continue;
+      }
+
+      // Add a future which, upon resolving to the existance of an account with
+      // that Idp, results in the Idp's `method` value, or null if no account
+      // exists.
```
marcelomendoncasoares: Suggestion: We can remove, since it is already clear from the method's doc comment.

```suggestion
```

marcelomendoncasoares:
Hey @craiglabenz! Finally had the time to think a little about this PR.

The main thing that worries me about the current design is that the whole `Idp` inheritance solution serves only to return a list of strings with the signed-in providers to the client. But then, users will still have to manually map these strings to the providers somehow. Since these strings are defined in server objects, they are arbitrary from the client perspective and it feels like we would be losing the type-safe experience they get when building the widgets.

Exploring the solution, one alternative that I thought of that is more aligned with how we implemented the `SignInWidget` is to only expose a `hasAccount` endpoint on each IDP. We can even use endpoint inheritance if we want to make all IDP endpoints extend a `BaseIdpEndpoint` that declares this method as abstract, and then iterate through all on the client.

In the end, users will still need to call specific widgets for each provider - as we do inside the `SignInWidget` -, so it is not a problem if they get responsible by checking each `has*` call to render the IDP on an account linking page.

What do you think of this approach?

craiglabenz:
Thanks for the feedback, @marcelomendoncasoares. I will explore that approach.

I'm not 100% sold on my current implementation either, and for the same reasons, but settled on strings because that would allow arbitrary IDPs that individual teams add to their own applications to seamlessly integrate into the system. However, it's possible that moving this logic to the Endpoint layer would improve things.

I will report back what I find after moving code around.

craiglabenz:
I've thought about this more, @marcelomendoncasoares and want to propose one last chance at a catch-all endpoint. If you don't find this approach sufficiently in line with existing Serverpod auth patterns then I will move forward with the conversion to individual endpoints as you suggested in your last comment.

The changes I made in this last patch introduce a new `IdentityProvider` enum with values for all first-class IDPs. This allows us to update the function signature of the catch-all endpoint from

```dart
Future<Set<String>> idpAccounts(final Session session);
```

to

```dart
Future<Set<IdentityProvider>> idpAccounts(final Session session);
```

This comes at the cost of slightly less flexibility for developers who add their own Idps to their projects. Such developers would have two options, both of which we can document:

1) They could only use the `IdpEndpoint.idpAccounts()` endpoint for Serverpod-supported providers and implement their own endpoints for any custom Idps they add. Or,

2) They could copy both the enum and the endpoint, adding any new fields to their enum and returning their enum from their copied endpoint. Then, they would merely ignore the Serverpod-provided endpoint and call their own.

In either case, developers adding their own Idps have a minor additional step; whereas developers only using Serverpod-supported Idps have the smoothest possible path.

---

WDYT about this approach? If you aren't happy with it, I will convert the whole thing to individual endpoints.

marcelomendoncasoares:
> I'm not 100% sold on my current implementation either, and for the same reasons, but settled on strings because that would allow arbitrary IDPs that individual teams add to their own applications to seamlessly integrate into the system. However, it's possible that moving this logic to the Endpoint layer would improve things.

This was precisely the same reason why we went with storing the `AuthStrategy` as a `String` on the `AuthSuccess` instead of the `enum`, so that developers could easily hook custom token managers without having to change the base implementation.

> The changes I made in this last patch introduce a new `IdentityProvider` enum with values for all first-class IDPs. This allows us to update the function signature of the catch-all endpoint from

> This comes at the cost of slightly less flexibility for developers who add their own Idps to their projects. Such developers would have two options, both of which we can document:

If we were designing the IDP classes from scratch, I think an approach closer to what you are proposing would make total sense. But, giving the current state, it feels like committing to permanent changes that make the public interface dirtier and harder for newcomers to comprehend "just" because we can't break the interface - like the `getMethod` that is redundant with the `static` name or the `identityProviders` getter that should be a list of `Idp`, but can't be.

And I'm failing to see which would be the use case for the provider list other than a debug log. On the app, we will need to call the widget for each provider individually, which makes a bool check required - as we do on the `SignInWidget`. And, on the server, each provider IDP is also individual, so we also need an individual bool to use each object. Is there a use case that I'm missing in which the provider identifier list (either of `String` or `enum`) is better?

craiglabenz:
The provider list is used by `IdpEndpoint.idpAccounts` which currently returns a `Set<IdentityProvider>` indicating all connected Idps in one request; but could also return `Map<IdentityProvider, bool>` to give a more complete picture.

If that endpoint is not compelling to you then I will throw this approach right in the garbage and convert to individual endpoints :)

marcelomendoncasoares:
> The provider list is used by `IdpEndpoint.idpAccounts` which currently returns a `Set<IdentityProvider>` indicating all connected Idps in one request; but could also return `Map<IdentityProvider, bool>` to give a more complete picture.

But how are you envisioning the usage of the endpoint return by developers? I get the usage internally in the implementation, my question is more of the experience we expect users to have when using the `Set` return on their app/server. Because I'm really failing to perceive a case where it is used differently than individual bool checks: 

If we only expect users to do:

```dart
final availableIdps = await client.modules.auth.idpAccounts();

if (availableIdps.contains(IdentityProvider.google) {
  return GoogleLinkWidget();
}
```

Then it feels clearer to offer a simpler approach with: 

```dart
if (await client.modules.auth.googleIdp.hasAccount()) {
  return GoogleLinkWidget();
}
```

Or is your concern with doing separate individual requests instead of only one?

> If that endpoint is not compelling to you then I will throw this approach right in the garbage and convert to individual endpoints :)

My question is genuine to understand if we have a use case that the list facilitates sufficiently more than the individual bools and justifies the commitment to the interface changes :)

craiglabenz:
Apologies, I understand the confusion now. Your code snippet is close, it's more like this:

```dart
// Returns the subset of all possible Idps which this user has
// already linked to their account
final linkedIdps = await client.modules.auth.idpAccounts();

// Show these in a Column, or some such
final linkWidgets = [
  if (!linkedIdps.contains(IdentityProvider.google) 
    GoogleLinkWidget(),
  if (!linkedIdps.contains(IdentityProvider.apple) 
    AppleLinkWidget(),
  // etc etc
];
```

instead of this:

```dart
final linkWidgets = [
  if (!await client.modules.auth.googleIdp.hasAccount()) 
    GoogleLinkWidget(),
  if (!await client.modules.auth.googleIdp.hasAccount())
    AppleLinkWidget(),
  // etc etc
];
```

So yes, it is all about making 1 total request on the page where existing users link additional Idps; rather that 1 request _per Idp_.

marcelomendoncasoares:
> So yes, it is all about making 1 total request on the page where existing users link additional Idps; rather that 1 request _per Idp_.

Great! I agree that awaiting several futures creates a poor UX when rendering the widgets. But I really wanted to get away from the `enum`/strings identifier tradeoff...

I made an experimentation on a branch and opened a PR for your branch for you to see. It creates the following experience - very similar to the good one you have proposed -, but implemented using individual endpoints. Built on top of the `AvailableProviders` class, it also exposes the convenient bool getters:

```dart
final connectedIdps = await client.auth.idp.getConnectedIdps();

// It is possible to check IDPs by name or with a bool getter.
debugPrint('Connected IDPs: ${connectedIdps.names}');
debugPrint('Has Google: ${client.auth.idp.hasGoogle}');
debugPrint('Has Email: ${client.auth.idp.hasEmail}');
debugPrint('Has Custom: ${client.auth.idp.has('customIdp')}');
```

Using the `has<Type>([String? name])` it is also ready for extensibility. Users just need to declare their IDP endpoints extending the `IdpBaseEndpoint` and it will be available here. What do you think?

craiglabenz:
Responded on that PR, @marcelomendoncasoares

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -728,8 +728,25 @@ abstract class EndpointCaller {
...
+  /// ```
   T getEndpointOfType<T extends EndpointRef>([String? name]) {
-    if (name != null) return endpointRefLookup[name] as T;
+    if (name != null) {
+      final endpoint = endpointRefLookup[name] as T?;
+      if (endpoint == null) {
+        throw ServerpodClientEndpointNotFound(T);
+      }
+      return endpoint;
+    }
```
marcelomendoncasoares, line 749: Suggestion: We should reflect this change on the tests at `serverpod_test_client/test/get_endpoint_of_type_test.dart`.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/apple/apple_idp_utils_test.dart
```diff
@@ -0,0 +1,99 @@
...
+        () async {
+          final session = sessionBuilder.build();
+          final account = await utils.getAccount(session);
+          expect(account, isNull);
+        },
+      );
+
+      test(
+        'when calling getAccount with authenticated session but no apple account then it returns null',
+        () async {
```
marcelomendoncasoares: Suggestion: As a general suggestion for all these tests, the `Given` is actually being described as "with" on the `when` part of the description. It would be more adherent to the test pattern to do:

- Given an unauthenticated session, when calling getAccount, then it returns null.
- Given an authenticated session that has no apple account, when calling getAccount, then it returns null.
- Given an authenticated session with existing apple account, when calling getAccount, then it returns the account.

To avoid duplicating the setup, you can simply move it to a helper function.

We should also replicate these tests to validate the correct behavior of the `hasAccount` calls to ensure that as much of the public API is covered.

examples/auth/auth_flutter/lib/main.dart
```diff
@@ -157,7 +174,28 @@ class ConnectedScreen extends StatelessWidget {
...
+              const SizedBox(height: 8),
+              Text('User has Apple: ${connectedIdps!.hasApple ? '✅' : '❌'}'),
+              const SizedBox(height: 8),
+              Text('User has GitHub: ${connectedIdps!.hasGitHub ? '✅' : '❌'}'),
+              const SizedBox(height: 8),
+              Text(
+                'User has Firebase: ${connectedIdps!.hasFirebase ? '✅' : '❌'}',
+              ),
+              const SizedBox(height: 8),
+            ],
```
marcelomendoncasoares: Suggestion(NIT): Wrap these inside a separate `Column` to escape from the `spacing: 16` of the outer `Column` and make the list tighter. By using `spacing` on such column, we can also remove the repeated `SizedBox`.

> The `map` below is only to get rid of Firebase breaking the line, but feel free to adopt as you prefer.

```suggestion
            connectedIdps == null
                ? const CircularProgressIndicator.adaptive()
                : Column(
                    spacing: 4,
                    children: [
                      'All connected Idps: ${connectedIdps!.names}',
                      'User has Google: ${connectedIdps!.hasGoogle ? '✅' : '❌'}',
                      'User has Email: ${connectedIdps!.hasEmail ? '✅' : '❌'}',
                      'User has Apple: ${connectedIdps!.hasApple ? '✅' : '❌'}',
                      'User has GitHub: ${connectedIdps!.hasGitHub ? '✅' : '❌'}',
                      'User has Firebase: ${connectedIdps!.hasFirebase ? '✅' : '❌'}',
                    ].map((e) => Text(e)).toList(),
                  ),
```

marcelomendoncasoares (approved):
Great work on this, @craiglabenz! LGTM! 🚀

Now only one more piece and we'll have the anonymous IDP fully complete :)

PR #4523, opened 2025-12-30
fix: Fix not being able to issue new verification codes for emails with pending registration

### LINKED ISSUE #4522: Not able to request a new verification code for email registration
### Describe the bug

This issue [was first reported on Discord](https://discord.com/channels/1155780586871468093/1451995630665994422) as the impossibility of starting a new registration if a user did not verified the previous request. However, it highlighted another underlying problem which is the impossibility of requesting a new verification code until the current one has expired.

### To reproduce

With the Email IDP configured, request the creation of an account. Without verifying, request again with the same email. An exception will be thrown on the server and a dummy ID will be returned to the client.

```console
2025-12-20 17:48:12.741611Z  806866049 LOG            DEBUG                    message=Failed to start account registration for example@example.com, reason: email account request already exists
2025-12-20 17:48:12.742416Z  806866049 METHOD         emailIdp.startRegistration user=null, queries=2, duration=36µs
```

### Expected behavior

The previous request should be replaced by the new one.

### Library version

3.0.0 onwards.

### Platform information

N/A

### Additional context

N/A

- [x] I want to work on this

marcelomendoncasoares:
Fixes #4522.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_idp_server_exceptions.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_idp_server_exceptions.dart
@@ -56,6 +56,10 @@ final class EmailAccountAlreadyRegisteredException
 final class EmailAccountRequestInvalidEmailException
     extends EmailAccountRequestServerException {}
 
+// NOTE: This exception is no longer being used after fixing the bug where
+// users could not request a new verification code until the previous one had
+// expired. The exception can be removed in the next major version if no other
+// use is found for it.
 /// Exception thrown when an email account request already exists.
 final class EmailAccountRequestAlreadyExistsException
     extends EmailAccountRequestServerException {}
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/utils/email_idp_account_creation_util.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/utils/email_idp_account_creation_util.dart
@@ -52,7 +52,6 @@ class EmailIdpAccountCreationUtil {
   /// subclasses:
   /// - [EmailAccountRequestInvalidEmailException] if the email address is not valid.
   /// - [EmailAccountAlreadyRegisteredException] if the email address is already registered.
-  /// - [EmailAccountRequestAlreadyExistsException] if an account request already exists for the email address.
   ///
   /// It is important that the caller does not leak the existence of the
   /// account request to the outside client.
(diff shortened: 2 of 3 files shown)


modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_idp_server_exceptions.dart
```diff
@@ -56,6 +56,10 @@ final class EmailAccountAlreadyRegisteredException
 final class EmailAccountRequestInvalidEmailException
     extends EmailAccountRequestServerException {}
 
+// NOTE: This exception is no longer being used after fixing the bug where
```
exaby73, line 59: Could we add a `TODO` comment here and possibly log an issue for it?

Suggestion to create an umbrella issue to track deprecated symbols so that we can easily remove them in the next major version

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/email/email_idp/email_idp_start_password_reset_test.dart
```diff
@@ -204,6 +204,82 @@ void main() {
     },
   );
 
+  withServerpod(
```
exaby73, line 207: Can we add a test case to check if the old (now deleted) `requestId` is used to verify, it throws an error

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/email/email_idp/email_idp_start_registration_test.dart
```diff
@@ -152,4 +152,67 @@ void main() {
       });
     },
   );
+
+  withServerpod(
```
exaby73, line 156: Same test case for this one too

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/email/email_idp/email_idp_start_password_reset_test.dart
marcelomendoncasoares: I don't think this should be the scope of this test, since here we are only interested that the behavior of the `startPassword` method works as expected. Given the request is correctly created, we validate that completing works as expected in the `email_idp_finish_password_reset_test.dart` file.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/email/email_idp/email_idp_start_registration_test.dart
marcelomendoncasoares: Same as above.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/email/business/email_idp_server_exceptions.dart
marcelomendoncasoares: I added a TODO comment at first, but then changed it to a NOTE since it is possible that we find a later usage for this exception.

I have intentionally avoided creating an issue to avoid bloating our backlog, but, since you are also suggesting it, we can have an issue marked as breaking for evaluating in the next major release. I would just have it separate per symbol instead of as an umbrella.

marcelomendoncasoares: @coderabbitai Please create an issue for the NOTE on the code so we can track it before the next release.

PR #4524, opened 2025-12-30
docs: Add missing docs to auth-related models for better UX

marcelomendoncasoares:
Add docs for models to help guide users on what models to use for what - specially regarding `AuthUser*` and `UserProfile*`.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/lib/src/protocol/auth_user/models/auth_user.dart
+++ b/modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/lib/src/protocol/auth_user/models/auth_user.dart
@@ -13,6 +13,10 @@
 import 'package:serverpod_client/serverpod_client.dart' as _i1;
 import 'package:serverpod_auth_core_client/src/protocol/protocol.dart' as _i2;
 
+/// Core database entity representing a user in the authentication system.
+///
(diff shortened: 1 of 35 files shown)


modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/models/auth_user.spy.yaml
```diff
@@ -1,3 +1,7 @@
+### Core database entity representing a user in the authentication system.
+###
+### This class is meant to be used only to interact with the database. To transfer
```
exaby73, line 3: This comment makes me think whether this should be marked `serverOnly`

exaby73 (approved):

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/models/auth_user.spy.yaml
marcelomendoncasoares: It was `serverOnly` previously, but it prevented users from creating relations with `AuthUsers` from their models that they didn't want to be `serverOnly`. Then we removed, both from `AuthUsers` and `UserProfle`.

Issue #4566, opened 2026-01-10
Firebase autoinitialising on ios


jakubgiminski:
### Describe the bug

Root Cause: Platform-Specific Firebase Auto-Initialization

iOS Behavior:
- Firebase Core auto-initializes natively when the Firebase plugin is registered via GeneratedPluginRegistrant.register(with: self) in AppDelegate.swift
- This happens before any Dart code runs - it's part of the native iOS plugin registration process
- The native iOS Firebase SDK strictly validates the GoogleService-Info.plist file during auto-initialization
- It requires specific fields (GOOGLE_APP_ID, GCM_SENDER_ID, PROJECT_ID, etc.) with proper formatting
- If validation fails, it throws an Objective-C exception and crashes the app immediately

Android Behavior:
- Firebase Core does NOT auto-initialize on Android by default
- Android requires explicit initialization in Dart code (Firebase.initializeApp()) OR the Firebase Gradle plugin + google-services.json 
- Your Android build has neither:
  - No com.google.gms.google-services plugin in build.gradle.kts
  - No google-services.json file
  - No explicit Firebase.initializeApp() call
- Since firebase_auth package on Android is more lenient and doesn't force initialization until actually used, the app launches successfully

Why This Matters
Serverpod 3.2.0 added firebase_auth as a dependency, which brought in firebase_core. On iOS, simply having these packages as dependencies triggers the native auto-initialization, even if you never call Firebase.initializeApp() in your Dart code. Android doesn't have this behavior.

### To reproduce

Upgrade serverpod  to 3.2.0 and run on ios.

### Expected behavior

Firebase not autoinitialising in ios.

### Library version

3.2.0

### Platform information

ios 26

### How experienced are you with this library?

Intermediate - Familiar with the basics or have used it in a few projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

marcelomendoncasoares:
Thanks for the report, @jakubgiminski! This is a major issue that we'll have fixed and released until later today.

nielsenko:
@jakubgiminski I'm having some trouble re-creating this. What version of `firebase_core` gets resolved in your project?  Could you paste the relevant section from `pubspec.yaml`?

Next, could you add the following to your `ios/Runner/Info.plist`?
```
<key>FirebaseAppDelegateProxyEnabled</key>
<false/>
```
and see if this resolves your problem?

jakubgiminski:
[...]

PR #4577, opened 2026-01-12
fix: Move Firebase IDP to a separate package

marcelomendoncasoares:
Move the just released Firebase IDP package out of the central IDP Flutter package to avoid pushing Firebase compilation issues to users that are not using the provider.

Closes: #4566
Closes: #4583

## Pre-launch Checklist

## Breaking changes

Since the Firebase IDP has just been released and is still not documented, this is not considered a breaking change. The package will also be released for version 3.2.0 so that the to-be-released documentation can reference it.


jakubgiminski (approved):
Nice one, thanks a lot!

Issue #4422, opened 2025-12-11
Add Prometheus/OpenMetrics Support to Serverpod


gitrema:
### Problem to Solve

Serverpod currently lacks built-in support for exposing application and infrastructure metrics in a standardized, machine-scrapable format such as **Prometheus/OpenMetrics**. As a result, operators must implement custom solutions for collecting runtime statistics in that format.

### Proposal

Introduce **native Prometheus/OpenMetrics support** in Serverpod by adding:
1. **A dedicated middleware** that:
   * Exposes a /metrics endpoint that outputs metrics in Prometheus/OpenMetrics format.
   * Automatically collects core server metrics (e.g., request counts, response times, error rates).
   * Plays nicely with existing Serverpod middleware and logging.
2. **A set of Prometheus metric primitives** for server developers:
   * Counter
   * Gauge
   * Histogram
   * Summary (optional depending on scope)
   * Ability to register metrics globally or within specific modules/endpoints.
3. **A simple and ergonomic API such as:**

```dart
 final requestCounter = Counter(
	  name: 'serverpod_requests_total',
	  help: 'Total number of requests handled by this Serverpod instance',
	);

	class ExampleEndpoint extends Endpoint {
	  Future<String> getHello(Session session) async {
	    requestCounter.inc();
	    return 'Hello world';
	  }
	}	
```

4. **Proper namespacing to avoid collisions between core metrics and user-defined ones.**

This feature would be implemented as a new optional package or a first-party plugin (e.g., serverpod_prometheus), depending on maintainers’ preferences.

### Use Case

* **Production monitoring:** real-time visibility into request throughput, latency, and error rates.
* **Autoscaling:** Kubernetes HPA or KEDA can scale Serverpod instances based on Prometheus metrics.
* **Performance tuning:** Developers can track endpoint-level load, bottlenecks, and resource utilization.
* **Business metrics:** Apps can expose domain-specific counters and gauges for dashboards without relying on external instrumentation libraries.

### Alternatives

* Build custom metrics endpoints (duplicated effort, poor standardization).
* Rely solely on logs and external log processors (less efficient, lacks structured metrics).
* Use third-party Dart packages for Prometheus, but they currently do not integrate cleanly with Serverpod’s architecture or middleware pipeline.

### Additional context

* OpenMetrics is becoming the industry-standard interchange format for metrics and integrates seamlessly with Prometheus, Grafana, and cloud-native tooling.
[...]

PR #4601, opened 2026-01-16
feat: Add OpenMetrics/Prometheus monitoring support for Serverpod

gitrema:
This PR adds comprehensive OpenMetrics and Prometheus monitoring support to
Serverpod through a new `serverpod_open_metrics_server` package.
The implementation provides automatic HTTP metrics collection via middleware, custom
metrics support, and a `/metrics` endpoint compatible with Prometheus scraping.

**Key highlights:**
- ✅ Zero-configuration setup with sensible defaults
- ✅ Automatic HTTP request tracking (count, duration, in-flight)
- ✅ Custom metrics support (Counter, Gauge, Histogram)
- ✅ Security-first design with localhost-only access by default
- ✅ Cardinality protection to prevent memory exhaustion
- ✅ Full OpenMetrics and Prometheus text format support

Security Considerations

⚠️ **Important:** The metrics endpoint is **restricted to localhost by default** for security. This prevents accidental exposure of potentially sensitive operational data.

To allow remote access:
1. Use the `allowedIps` parameter with specific IPs or CIDR ranges
2. Configure firewall rules to restrict access to monitoring infrastructure
3. Consider using mutual TLS or other authentication mechanisms in production

Setting `allowedIps: []` disables all access control (**not recommended for production**).

Closes: https://github.com/serverpod/serverpod/issues/4422

## Pre-launch Checklist

## Breaking changes

None - this is a new optional package that doesn't modify existing Serverpod behavior.

## Summary by CodeRabbit

* **New Features**
  * Standalone metrics module with Counter, Gauge, and Histogram metric types
  * HTTP middleware for automatic request tracking and duration measurement
  * Prometheus-compatible metrics endpoint with format negotiation
  * IP-based access control for metrics endpoint (localhost-only by default)
  * Configurable path normalization to manage metric cardinality
  * Support for labeled metrics and custom metric registration

* **Documentation**
  * Added README with setup instructions and configuration guide
  * Added CHANGELOG documenting v3.2.2 release

* **Tests**
  * Comprehensive test coverage for metrics, middleware, and access control

* **Chores**
  * Added configuration and build files

### DIFF
--- a/templates/pubspecs/tests/serverpod_test_nonvector/serverpod_test_nonvector_server/pubspec.yaml
+++ b/templates/pubspecs/tests/serverpod_test_nonvector/serverpod_test_nonvector_server/pubspec.yaml
@@ -36,8 +36,8 @@ dependency_overrides:
     path: ../../../integrations/serverpod_cloud_storage_s3
   serverpod_lints:
     path: ../../../packages/serverpod_lints
-  serverpod_open_metrics_server:
-    path: ../../../modules/serverpod_open_metrics/serverpod_open_metrics_server
+  serverpod_open_metrics:
+    path: ../../../integrations/serverpod_open_metrics
   serverpod_serialization:
     path: ../../../packages/serverpod_serialization
   serverpod_service_client:
--- a/templates/pubspecs/tests/serverpod_test_server/pubspec.yaml
+++ b/templates/pubspecs/tests/serverpod_test_server/pubspec.yaml
@@ -15,7 +15,7 @@ dependencies:
   serverpod: SERVERPOD_VERSION
   serverpod_auth_server: SERVERPOD_VERSION
   serverpod_cloud_storage_s3: SERVERPOD_VERSION
-  serverpod_open_metrics_server: SERVERPOD_VERSION
+  serverpod_open_metrics: SERVERPOD_VERSION
   serverpod_test_module_server:
     path: ../serverpod_test_module/serverpod_test_module_server
   serverpod_test_shared:
@@ -52,8 +52,8 @@ dependency_overrides:
     path: ../../integrations/serverpod_cloud_storage_s3
   serverpod_lints:
     path: ../../packages/serverpod_lints
-  serverpod_open_metrics_server:
-    path: ../../modules/serverpod_open_metrics/serverpod_open_metrics_server
+  serverpod_open_metrics:
+    path: ../../integrations/serverpod_open_metrics
   serverpod_serialization:
     path: ../../packages/serverpod_serialization
   serverpod_service_client:
--- a/tests/serverpod_test_nonvector/serverpod_test_nonvector_server/pubspec.yaml
+++ b/tests/serverpod_test_nonvector/serverpod_test_nonvector_server/pubspec.yaml
@@ -37,8 +37,8 @@ dependency_overrides:
     path: ../../../integrations/serverpod_cloud_storage_s3
   serverpod_lints:
     path: ../../../packages/serverpod_lints
-  serverpod_open_metrics_server:
-    path: ../../../modules/serverpod_open_metrics/serverpod_open_metrics_server
+  serverpod_open_metrics:
+    path: ../../../integrations/serverpod_open_metrics
   serverpod_serialization:
     path: ../../../packages/serverpod_serialization
   serverpod_service_client:


gitrema:
@coderabbitai review

integrations/serverpod_open_metrics/lib/src/util/text_format.dart
```diff
@@ -0,0 +1,217 @@
...
+  /// The content type for Prometheus text format 0.0.4.
+  static const String prometheusContentType =
+      'text/plain; version=0.0.4; charset=utf-8';
+
+  /// The content type for OpenMetrics format 1.0.0.
+  static const String openMetricsContentType =
+      'application/openmetrics-text; version=1.0.0; charset=utf-8';
+
+  /// The content type for Prometheus text format (default).
+  /// @deprecated Use prometheusContentType or openMetricsContentType instead
```
exaby73: This should use the `@Deprecated` annotation instead.

Though, why are we introducing a deprecated field in a new package? Could this be removed instead?

integrations/serverpod_open_metrics/lib/src/util/text_format.dart
```diff
@@ -0,0 +1,217 @@
...
+  static const String contentType = prometheusContentType;
+
+  /// Export a registry's metrics to OpenMetrics or Prometheus text format.
+  ///
+  /// [registry] The metric registry to export.
+  /// [format] The format to use (defaults to Prometheus 0.0.4 for backward compatibility).
+  ///
+  /// Returns a string in the specified text exposition format.
```
exaby73: Docs should follow the prose style from effective Dart. See https://dart.dev/effective-dart/documentation#do-use-prose-to-explain-parameters-return-values-and-exceptions

integrations/serverpod_open_metrics/lib/src/metrics/counter.dart
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  /// Get the current value of the counter.
+  double get value => _value;
+
+  /// Reset the counter to zero.
+  /// This should typically only be used for testing.
+  void reset() {
+    _value = 0;
+  }
```
exaby73: If this is used for testing only, could we move this into an extension that is not exported by the public API and instead imported directly by tests?

integrations/serverpod_open_metrics/lib/src/metrics/counter.dart
```diff
@@ -0,0 +1,223 @@
...
+import 'metric.dart';
+
+/// A Counter is a cumulative metric that represents a single monotonically
+/// increasing value. Counters can only go up (or be reset to zero on restart).
+///
+/// **Concurrency:** This implementation is designed for Dart's single-isolate
+/// async/await concurrency model. Counter increments are safe from race
+/// conditions within a single isolate because individual operations complete
+/// atomically between await points. For multi-isolate deployments, each isolate
+/// should maintain its own registry, or use an external metrics aggregator.
```
exaby73: Should we include this for every metric type? This would be better documented on docs.serverpod.dev in my opinion, instead of having it documented here. It should also be common knowledge for developers working with Dart to consider Dart's concurrency model.

integrations/serverpod_open_metrics/lib/src/metrics/metric_registry.dart
```diff
@@ -0,0 +1,172 @@
...
+  Counter counter(final String name, {required final String help}) {
+    final counter = Counter(name: name, help: help);
+    register(counter);
+    return counter;
+  }
+
+  /// Create and register a labeled counter.
+  ///
+  /// Throws [ArgumentError] if a metric with the same name already exists.
+  LabeledCounter labeledCounter(
```
exaby73: Instead of these methods creating the metric object, it should accept a metrics object instead. This allows better maintainability if we ever need to change the constructor signiture, we have to duplicate the constructor params in 2 places

exaby73: Same applied for others in this file

exaby73: There is a generic `register` method. We could just keep that one, and remove the others as well. I would like some discussion on this.

integrations/serverpod_open_metrics/lib/src/metrics/metric_registry.dart
```diff
@@ -0,0 +1,172 @@
...
+///
+/// final requestDuration = registry.histogram(
+///   'http_request_duration_seconds',
+///   help: 'HTTP request duration in seconds',
+/// );
+///
+/// // Collect all metrics
+/// final allSamples = registry.collectAll();
+/// ```
+class MetricRegistry {
```
exaby73, line 30: Would this make sense to be a singleton instead? Possibly also be attached to the `Serverpod.instance` via an extension since we currently do not have any dependency injection set up for Serverpod. Currently, users would have to either make this registry a global variable or create their own extensions or using a DI library to access this within endpoints or middleware.

As for it being a singleton, are there disadvantages to this being one?

integrations/serverpod_open_metrics/lib/src/metrics/metric.dart
```diff
@@ -0,0 +1,150 @@
...
+      case MetricType.counter:
+        return 'counter';
+      case MetricType.gauge:
+        return 'gauge';
+      case MetricType.histogram:
+        return 'histogram';
+      case MetricType.summary:
+        return 'summary';
+    }
+  }
```
exaby73: Could we just return `this.name` here?

integrations/serverpod_open_metrics/lib/src/metrics/metric.dart
```diff
@@ -0,0 +1,150 @@
...
+
+  /// Remove a child metric with the specified label values.
+  void remove(final List<String> labelValues);
+
+  /// Clear all child metrics.
+  void clear();
+}
+
+/// Helper class to validate label names and values.
+class LabelValidator {
```
exaby73: If this class is exported by the public API as it is now, we should have tests for it. This class could be an implementation detail and the validation tested through the public APIs

integrations/serverpod_open_metrics/lib/serverpod_open_metrics.dart
```diff
@@ -0,0 +1,43 @@
...
+// Metric types
+export 'src/metrics/counter.dart';
+export 'src/metrics/gauge.dart';
+export 'src/metrics/histogram.dart';
+export 'src/metrics/metric.dart';
+export 'src/metrics/metric_registry.dart';
+
+// IP access control (for advanced use cases)
+export 'src/middleware/ip_access_control.dart'
+    show IpAccessControl, IpAccessRule;
```
exaby73: `show` does nothing here as all members are exported. This could be removed

integrations/serverpod_open_metrics/lib/src/middleware/metrics_middleware.dart
```diff
@@ -0,0 +1,564 @@
...
+  0.1,
+  0.25,
+  0.5,
+  1.0,
+  2.5,
+  5.0,
+  10.0,
+];
+
+/// Creates an OpenMetrics middleware that collects HTTP metrics and
```
exaby73, line 42: The doc here, while it is very detailed, should be documented on our docs on docs.serverpod.dev. These doc comments should be simpler

integrations/serverpod_open_metrics/test/counter_test.dart
```diff
@@ -0,0 +1,436 @@
+import 'package:serverpod_open_metrics/src/metrics/counter.dart';
```
exaby73, line 1: Tests should follow the "Given When Then" pattern. This applies to all tests in this directory and `tests/serverpod_test_server/test/middleware/metrics_middleware_test.dart` as well

integrations/serverpod_open_metrics/lib/src/metrics/metric_registry.dart
gitrema: Thanks for raising it. It makes sene what you propose. I want to make sure we're aligned on the tradeoffs before proceeding.

I understand your concerns about the maintenance cost:
  - Constructor parameters are duplicated between the metric class and registry factory methods
  - Adding/changing parameters requires updates in multiple places
  - More code to maintain (6 factory methods)

On the other hand the factory methods provide a safety benefit:

Current approach (fail-fast):
```dart
final counter = registry.counter('http_requests_total', help: 'Help text');
// counter is created AND registered - can't forget
```

Proposed approach (requires discipline):
```dart
final counter = Counter(name: 'http_requests_total', help: 'Help text');
registry.register(counter);  // <- easy to forget this line
```

A developer who forgets to call register() will have a metric that:
  - Compiles fine
  - Runs fine locally
  - Never appears in /metrics endpoint
  - Creates a subtle production bug that's hard to debug

The factory pattern prevents this entire class of errors.

My assumption, that mitigate the maintenance cost, is that Agentic AI will make this class of refactoring easier.

So, with that said, what should we optimize for? Easier maintenance or DevEx?
Happy to follow your guidance.

integrations/serverpod_open_metrics/lib/src/metrics/counter.dart
gitrema: Good catch! A leftover of the initial prototype. I'll remove it completely

integrations/serverpod_open_metrics/lib/src/metrics/metric_registry.dart
exaby73: It being clearly documented about calling `register` and the fact that you could do:

```dart
final counter = registry.register(Counter(name: 'http_requests_total', help: 'Help text'));
```

I think is a worthy trade-off as it's a user error. Having a single `register` also benefits users when they implement their own metrics as they don't need to either implement their own factories or use different APIs. It's just one interface.

This pattern is also not uncommon in Dart like many DI libraries will have a register method to call, and even in Serverpod, our client requires you to call `client.initialize()` as well.

gitrema: it makes sense, lemme remove the factory methods

integrations/serverpod_open_metrics/lib/src/metrics/metric_registry.dart
gitrema: The main disadvantage I see is for testing and the usual issues that might arise from lifecycle management.

I understand the pain for a user though as I tried myself.
We can have a compromise solution here:

- MetricRegistry will have a static `shared` field (the singleton) along with the existing Constructor
- the shared instance of MetricRegistry will be exposed through a Serverpod extension (as you also suggested)

How does the plan sound?

exaby73: Yeah I think that makes sense to have. 

As for testing, as long as the registry is tested in isolation, we shouldn't have any issues with the actual singleton. I suppose when it comes time to test Serverpod specific metrics like query times, DB calls, etc. we might face an issue to test those in isolation. Maybe you could take inspiration from the singleton class we have for the new auth module AKA `AuthServices` and how that's tested?

PR #4610, opened 2026-01-19
docs: Add tech spike design doc for revamped health checking

nielsenko:
# Summary

Design proposal to replace Serverpod's single `GET /` health endpoint with Kubernetes-style probes (`/livez`, `/readyz`, `/startupz`) and an extensible `HealthIndicator` registry.

## Key Changes

- **Separate probe endpoints**: Liveness (process alive), Readiness (can handle traffic), Startup (initialization complete)
- **Extensibility**: Plugin architecture for custom health indicators (e.g., Stripe API, internal services)
- **Built-in indicators**: Database and Redis connectivity checks
- **Standardized response format**: JSON following the draft RFC for health check responses
- **Anti-cascade design**: Permissive liveness probes to prevent thundering herd restarts

## Motivation

The current `GET /` endpoint conflates liveness and readiness, causing unnecessary pod restarts when the database is slow (which restarting won't fix). This proposal enables orchestrators to make better lifecycle decisions.

### DIFF
--- a/docs/design/health_check.md
+++ b/docs/design/health_check.md
@@ -1,0 +1,360 @@
+# Tech Spike: Health Check Endpoints and Extensibility
+
+## Executive Summary
+
+This spike investigates the transition of Serverpod from its current internal health metrics system to a modular system. The goal is to support modern orchestration (Kubernetes/AWS) and provide extensibility for custom infrastructure checks.
+
+### Current State
+
+Serverpod currently has health check functionality that:
+
+1. Exposes a `GET /` endpoint (`server.dart`) that returns:
+   - `200 OK` with timestamp when all metrics are healthy
+   - `503 SADNESS` with details when any metric is unhealthy
+
+2. Collects metrics (`health_check.dart`) for:
+   - Database connection status and response time (via `testConnection()`)
+   - CPU load average
+   - Memory usage
+   - Connection pool information (active, closing, idle)
+
+3. Provides an Insights API (`InsightsEndpoint.checkHealth()`) for programmatic access to health data, plus `getHealthData()` for historical metrics.
+
+4. Periodically logs metrics to the database for historical analysis.
+
+While the current `GET /` endpoint can technically be used for orchestrator health checks, the design has limitations:
+- A single endpoint conflates liveness and readiness concerns
+- Database unavailability triggers an "unhealthy" response, which could cause unnecessary pod restarts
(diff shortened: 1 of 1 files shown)


docs/design/health_check.md
```diff
@@ -0,0 +1,360 @@
...
+
+This spike investigates the transition of Serverpod from its current internal health metrics system to a modular system. The goal is to support modern orchestration (Kubernetes/AWS) and provide extensibility for custom infrastructure checks.
+
+### Current State
+
+Serverpod currently has health check functionality that:
+
+1. Exposes a `GET /` endpoint (`server.dart`) that returns:
+   - `200 OK` with timestamp when all metrics are healthy
+   - `503 SADNESS` with details when any metric is unhealthy
```
exaby73, line 13: Sadness?? Isn't it Service Unavailable?

nielsenko: I'm not sure _why_ we use SADNESS today (and _who_ introduced it) . It doesn't matter, probes only depend on the status code, but yes it is unconventional.

We fill in metrics in the response as well (see `packages/serverpod/lib/src/server/server.dart:267`.

This is how it is today, not how the design suggest it will be

docs/design/health_check.md
```diff
@@ -0,0 +1,360 @@
...
+  args,
+  healthConfig: HealthConfig(
+    cacheTtl: Duration(seconds: 1),
+    readinessIndicators: [
+      DatabaseHealthIndicator(),    // Built-in: checks Postgres connectivity
+      RedisHealthIndicator(),       // Built-in: checks Redis connectivity
+      StripeApiIndicator(),         // Custom: checks Stripe API availability
+      InventoryServiceIndicator(),  // Custom: checks internal microservice      
+    ],
+    startupIndicators: [...],  // If different from readiness
```
marcelomendoncasoares, line 153: Suggestion: We can have a startup build-in check that awaits the conclusion of the `Serverpod.start` method to then return `pass`. Since the method will exit upon failure, checking that it has concluded ensures that migrations were correctly applied and all existing service checks passed.

About the question on whether `startupIndicators` should be different from `readinessIndicators`, I think it make sense to treat the `startupIndicators` parameter as additional checks to run on startup, but always run the `readinessIndicators` as well.

nielsenko:
This is now reflected in #4662.

PR #4618, opened 2026-01-20
fix: Session log duration unit display (#4617)

### LINKED ISSUE #4617: Session log duration displays wrong unit (ms instead of s)
### Describe the bug

The console log displays duration values with the wrong unit. For example, a FutureCall that took 47.6 seconds displays as duration=47.6ms instead of duration=47.6s.                                     
Image

The _printDuration method in log_writers.dart expects milliseconds (per its parameter name and logic), but all duration values in Serverpod are stored in seconds (as documented in LogSettings).

Root cause: In log_writers.dart:236, the code multiplies by Duration.microsecondsPerMillisecond (1000) instead of Duration.microsecondsPerSecond (1000000): var micros = (milliseconds * Duration.microsecondsPerMillisecond).round();

### To reproduce

1. Create a FutureCall that takes several seconds to execute                                                                                
2. Observe console log shows duration=Xs with wrong unit (ms instead of s)                                                                  
3. Compare with serverpod_session_log.duration in database (stored in seconds)                                                              

### Expected behavior

  A 47.6 seconds duration should display as 47.6s, not 47.6ms.                                                                                 

### Library version

serverpod: 3.2.3                                                                                                                          
Dart SDK version: 3.10.7 (stable)

### Platform information

macOS

### Additional context

_No response_

- [x] I want to work on this

Tokotuu:
The `_printDuration` method in `log_writers.dart` was treating duration values as milliseconds, but Serverpod stores all durations in seconds (as documented in `LogSettings`). This caused durations to display with incorrect units - e.g., a 47.6 second FutureCall displayed  as "47.6ms" instead of "47.6s".                                                                                                             
                                                                                                                                              
  **Changes:**                                                                                                                                
  - Updated doc comment from "milliseconds" to "seconds"                                                                                      
  - Renamed parameter from `milliseconds` to `seconds`                                                                                        
  - Changed multiplier from `Duration.microsecondsPerMillisecond` to `Duration.microsecondsPerSecond`                                         
                                                                                                                                              
  Fixes #4617                                                                                                                                 
                                                                                                                                              
  **Note on tests:** No unit test was added because `_printDuration` is a private method.

### DIFF
--- a/packages/serverpod/lib/src/server/log_manager/log_writers.dart
+++ b/packages/serverpod/lib/src/server/log_manager/log_writers.dart
@@ -228,12 +228,19 @@ class JsonStdOutLogWriter extends LogWriter {
 class TextStdOutLogWriter extends LogWriter {
   static bool headersWritten = false;
 
-  /// Formats a duration value expressed in seconds so that it is easy to
-  /// read in logs. Very short durations will be printed using microseconds
-  /// while longer ones will switch to milliseconds or seconds.
-  static String _printDuration(double? seconds) {
-    if (seconds == null) return 'n/a';
-    var micros = (seconds * Duration.microsecondsPerSecond).round();
+  /// Converts a duration value expressed in seconds to a [Duration].
+  static Duration? _secondsToDuration(double? seconds) {
+    if (seconds == null) return null;
+    var microseconds = (seconds * Duration.microsecondsPerSecond);
+    return Duration(microseconds: microseconds.round());
+  }
+
+  /// Formats a [Duration] so that it is easy to read in logs. Very short
+  /// durations will be printed using microseconds while longer ones will
+  /// switch to milliseconds or seconds.
+  static String _printDuration(Duration? duration) {
+    if (duration == null) return 'n/a';
+    var micros = duration.inMicroseconds;
     if (micros < 1000) {
       // Ignore required because dart does not understand that "µ" is a valid
       // character in a string.
@@ -317,7 +324,7 @@ class TextStdOutLogWriter extends LogWriter {
       id: _logId,
       fields: {
         'id': entry.id,
-        'duration': _printDuration(entry.duration),
+        'duration': _printDuration(_secondsToDuration(entry.duration)),
         'query': entry.query,
       },
       error: entry.error,
@@ -355,7 +362,7 @@ class TextStdOutLogWriter extends LogWriter {
           fields: {
             'user': entry.userId,
             'queries': entry.numQueries,
-            'duration': _printDuration(entry.duration),
+            'duration': _printDuration(_secondsToDuration(entry.duration)),
           },
           error: entry.error,
           stackTrace: entry.stackTrace,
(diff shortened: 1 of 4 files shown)


Tokotuu:

packages/serverpod/lib/src/server/log_manager/log_writers.dart
```diff
@@ -228,12 +228,12 @@ class JsonStdOutLogWriter extends LogWriter {
...
   static bool headersWritten = false;
 
-  /// Formats a duration value expressed in milliseconds so that it is easy to
+  /// Formats a duration value expressed in seconds so that it is easy to
   /// read in logs. Very short durations will be printed using microseconds
   /// while longer ones will switch to milliseconds or seconds.
-  static String _printDuration(double? milliseconds) {
-    if (milliseconds == null) return 'n/a';
-    var micros = (milliseconds * Duration.microsecondsPerMillisecond).round();
+  static String _printDuration(double? seconds) {
```
nielsenko: @Tokotuu Thanks for spotting and correcting this bug.

Have you considered changing the signature to take an actual Duration instead?

Tokotuu: @nielsenko Good point, using Duration would eliminate unit ambiguity. However, since all 7 call sites pass entry.duration as a double? (seconds), adopting Duration would require conversion logic at each call site for a private method. I'd prefer keeping this fix minimal, but happy to expand if you feel strongly about it.

marcelomendoncasoares: Although I agree that a minimal fix is usually the best to do, I would also argue that it is better for us to fix this for good and prevent further occurrences of misinterpretation of the `int` parameter. We could consider that the parameter name should be enough, but the issue itself proved that it isn't.

 @Tokotuu If you're ok with this, I would suggest changing the parameter to `Duration` as @nielsenko suggested.

nielsenko: @Tokotuu Also, are we certain that all 7 call sites pass seconds? Haven't checked, but I would be surprised.

Tokotuu: @nielsenko @marcelomendoncasoares 
I've updated the PR to use Duration as suggested. Added a helper method _secondsToDuration(double? seconds) to handle the conversion from the model's double to Duration.
I also noticed SessionLogEntry.duration was missing the "in seconds" clarification that QueryLogEntry has, so I've added that documentation fix as well.


@nielsenko 
Regarding your question about whether all 7 call sites pass seconds - yes, they all do. Here's the proof:

1. Code evidence in log_manager.dart:
const double _microNormalizer = 1000 * 1000;  // = 1,000,000

// Line 311 - SessionLogEntry
duration: duration.inMicroseconds / _microNormalizer  // microseconds / 1,000,000 = seconds

// Lines 153, 200 - QueryLogEntry
var executionTime = duration.inMicroseconds / _microNormalizer  // also seconds

2. Model documentation:
- QueryLogEntry (https://github.com/serverpod/serverpod/blob/main/packages/serverpod/lib/src/models/query_log_entry.spy.yaml#L19): "The time it took to execute the query, in seconds."
- LogSettings (https://github.com/serverpod/serverpod/blob/main/packages/serverpod/lib/src/models/log_settings.spy.yaml#L29-L32): "The duration in seconds for a session/query to be considered slow."

nielsenko: Sorry, I lost the ball on this one.

PR #4630, opened 2026-01-21
feat: Add PostHog analytics integration

ammar2666:
# PostHog Analytics Integration

## What We've Set Up

This PR adds PostHog analytics to track usage of the Serverpod CLI and user interactions on the Serverpod website. PostHog works alongside the existing MixPanel analytics - both systems run in parallel, so you get analytics data in both platforms without any conflicts.

### CLI Analytics
We're now tracking all CLI commands (`serverpod create`, `serverpod generate`, `serverpod run`, `serverpod version`, `serverpod upgrade`, etc.) and sending events to PostHog with details about what commands were run, whether they succeeded, and version information.

### Website Analytics
We've added PostHog tracking to the website for important user actions like newsletter signups, form submissions, button clicks, and navigation. When users provide their email, we identify them in PostHog so you can track their journey across sessions. There's a second PR that will be opened on the `serverpod_website` for this to work alongside the website.

## Setting Up the API Key

PostHog requires an API key to send events. You need to set this up as an environment variable. If the API key isn't set, analytics will simply be disabled - the CLI and website will work normally without it. The code that gets the API key from the environment variable is in this file: `tools/serverpod_cli/lib/src/analytics/posthog_config.dart`

### For Local Development and Testing

**Windows (PowerShell):**
```powershell
$env:POSTHOG_API_KEY = "phc_your_api_key_here"
```

**Windows (Command Prompt):**
```cmd
set POSTHOG_API_KEY=phc_your_api_key_here
```

**macOS/Linux:**
```bash
export POSTHOG_API_KEY="phc_your_api_key_here"
```

To make this permanent for your local machine, add it to your shell profile (`.bashrc`, `.zshrc`, or PowerShell profile).

### For Production

Set the `POSTHOG_API_KEY` environment variable in your production environment. How you do this depends on the deployment setup:

- **Docker:** Add to your `docker-compose.yaml` or Docker environment variables
- **Cloud platforms (AWS, GCP, Azure):** Set in your environment configuration or secrets manager
- **CI/CD:** Add as a secret in your CI/CD pipeline configuration

The same environment variable works for both:
- **CLI analytics** - tracks when developers use Serverpod commands
- **Server analytics** - tracks events from your Serverpod application (when you create a project with `serverpod create`, the server automatically uses the same environment variable)

## Breaking Changes

None. This is a purely additive feature. If you don't set the API key, everything works exactly as before - you just won't get PostHog analytics.

Also, if you're concerned about the high number of LOC modifications, it was mostly for getting PostHog and MixPanel to work under a single Analytics composite object and gathering data for adding into properties of PostHog events.

Also, please note that while we have made our efforts in maintaining MixPanel functionality, we haven't tested it since we don't have access to MixPanel. It's best to first verify functionality of MixPanel before merging it in main.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,6 @@ set POSTHOG_API_KEY=phc_your_api_key_here
 export POSTHOG_API_KEY="phc_your_api_key_here"
 ```
 
-
 To make this permanent, add it to your shell profile (`.bashrc`, `.zshrc`, or PowerShell profile).
 
 **For production:**


ammar2666:

vlidholt:
Implemented in #4655

PR #4631, opened 2026-01-22
feat: Implement livez, readyz, and startupz endpoints

### LINKED ISSUE #4662: Implement new Kubernetes style health endpoints
This issue details the transition of Serverpod from its current internal health metrics system to a modular system. The goal is to support modern orchestration (Kubernetes/AWS) and provide extensibility for custom infrastructure checks.

### Current State

Serverpod currently has health check functionality that:

1. Exposes a `GET /` endpoint (`server.dart`) that returns:
   - `200 OK` with timestamp when all metrics are healthy
   - `503 SADNESS` with details when any metric is unhealthy

2. Collects metrics (`health_check.dart`) for:
   - Database connection status and response time (via `testConnection()`)
   - CPU load average
   - Memory usage
   - Connection pool information (active, closing, idle)

3. Provides an Insights API (`InsightsEndpoint.checkHealth()`) for programmatic access to health data, plus `getHealthData()` for historical metrics.

4. Periodically logs metrics to the database for historical analysis.

While the current `GET /` endpoint can technically be used for orchestrator health checks, the design has limitations:
- A single endpoint conflates liveness and readiness concerns
- Database unavailability triggers an "unhealthy" response, which could cause unnecessary pod restarts
- No startup probe support for migration/initialization delays
- No extensibility for custom health indicators
- Redis connectivity is not checked

### Proposal

Add orchestrator-friendly HTTP endpoints that:
- Allow orchestrators to make informed decisions about container lifecycle and traffic routing
- Provide extensibility for custom health indicators
- Follow Kubernetes conventions for health probing

Modern container orchestration platforms like Kubernetes rely on health probes to make automated decisions about container lifecycle and traffic routing. A poorly designed health check system can cause more harm than good - leading to unnecessary restarts, cascading failures, or services that appear healthy but cannot actually serve requests. This document outlines an approach that tries to balance sensitivity (detecting real problems) with stability (avoiding false positives that destabilize the system).

### Separation of Concerns

The most critical insight in health check design is that _Is the process alive?_ and _Can the process handle requests?_ are fundamentally different questions with different consequences for a given answer.

Kubernetes and similar orchestrators provide distinct probe types precisely because conflating these concerns leads to pathological behavior. For example, if a single "health check" endpoint returns unhealthy whenever the database is slow, the orchestrator may restart the container - but restarting doesn't fix a slow database. Worse, the restart destroys all in-flight requests and causes a cold-start penalty, making the situation worse.

We therefore adopt distinct endpoints for:
- Liveness: Is the application fundamentally broken?
- Readiness: Can the application usefully handle new requests?
- Startup: Has the application completed its boot sequence?

### Architectural Decoupling

The new health endpoints should be decoupled from the `Server` core. By extracting health logic into a pluggable `HealthCheckManager`, we achieve:
- Extensibility: Users can add custom indicators
- Maintainability: Changes to health check behavior don't require changes to server lifecycle code
- Testability: Health logic can be unit tested in isolation

### Resiliency: The "Anti-Cascade" Rule

In distributed systems, health checks can inadvertently cause the failures they're meant to detect. Consider a scenario where a brief network partition causes all pods to fail their database health check simultaneously. If the health check is too aggressive, the orchestrator restarts all pods at once, creating a "thundering herd" that overwhelms the database when they all reconnect simultaneously.

The guiding principle is: **liveness probes must be permissive**. A restart is a "nuclear option" that should be reserved for truly unrecoverable states (deadlocks, memory corruption, infinite loops). Transient issues like slow queries, garbage collection pauses, or temporary network blips should not trigger restarts.

## Proposed Endpoints

We adopt the `z-suffix` convention (`/livez`, `/readyz`, `/startupz`) to align with Kubernetes API server conventions.

### `/livez`

"Should this container be killed and restarted?" This should only fail if the process is fundamentally broken.

Rationale: A restart destroys the entire process and all in-flight requests. This is only appropriate when the process is truly unrecoverable. If the server can respond to the health check, the process is fundamentally alive - even if it's experiencing temporary difficulties.

Common scenarios where we should NOT restart:
- A slow database query is blocking a request handler
- Garbage collection pause
- Processing an unusually large request
- Temporary network latency to an external service

In all these cases, the server will typically recover on its own. Restarting would destroy work in progress and potentially make the situation worse (cold start, reconnection storms, lost caches).

Implementation: The `/livez` endpoint should do minimal work - ideally just return `200 OK` immediately. It should NOT check database connectivity, external services, or perform any I/O that could block or fail due to external factors.

### `/readyz` 

"Should traffic be routed to this container?" This can fail for transient reasons like high load, database connectivity issues, or resource pressure.

While we're permissive about liveness, we can be more aggressive about readiness. If a pod is struggling, we want Kubernetes to stop sending it new traffic - but without killing it.

Implementation: The `/readyz` endpoint checks actual service capacity:
- Can we connect to the database?
- Can we connect to Redis (if configured)?
- Are custom health indicators healthy?

If any critical dependency is unavailable, the endpoint returns `503 Service Unavailable`, and Kubernetes stops routing traffic to this pod until it recovers.

### `/startupz`

"Has this container finished initializing?" Without a startup probe, Kubernetes might mark a still-initializing container as failed and restart it repeatedly. The startup probe tells Kubernetes to wait before even beginning liveness/readiness checks.

Questions: 
- How does the server signal that migrations have completed?
- What constitutes "warm-up"? Connection pool initialization? Cache priming?
- Should there be a configurable list of startup indicators similar to readiness indicators?

### Security

As metrics data can be highly sensitive, the metrics endpoints should run on a separate port for easy segregation.
Unauthenticated probes should only receive the status code, and no details.

### Summary

| Endpoint    | Type      | Responsibility                                            | Failure Outcome |
| ----------- | --------- | --------------------------------------------------------- | --------------- |
| `/livez`    | Liveness  | Check the server can respond.                             | Pod Restart     |
| `/readyz`   | Readiness | Checks dependencies like database and Redis connectivity. | Traffic Paused  |
| `/startupz` | Startup   | Checks if migrations and initial warm-up are complete.    | Delayed Ingress |

## Extensibility: The `HealthIndicator` Registry

Different deployments have different health requirements. A payment processing service might need to verify Stripe API connectivity; a media service might need to check CDN availability. Rather than hard-coding every possible check, we provide an extensible registry.

```dart
abstract class HealthIndicator {
  /// Human-readable name for logging and debugging
  String get name;
  
  /// Maximum time to wait for this check before considering it failed.
  /// This prevents a slow check from blocking the entire health endpoint.
  Duration get timeout => Duration(seconds: 5);
  
  /// Perform the health check and return the result.
  Future<HealthCheckResult> check();
}
```

Users register indicators when configuring Serverpod:

```dart
final pod = Serverpod(
  args,
  healthConfig: HealthConfig(
    cacheTtl: Duration(seconds: 1),
    readinessIndicators: [
      DatabaseHealthIndicator(),    // Built-in: checks Postgres connectivity
      RedisHealthIndicator(),       // Built-in: checks Redis connectivity
      StripeApiIndicator(),         // Custom: checks Stripe API availability
      InventoryServiceIndicator(),  // Custom: checks internal microservice      
    ],
    startupIndicators: [...],  // If different from readiness
  ),
);
```

Questions:
- How are timeouts configured per-indicator vs globally?
- What's the cache TTL default and how is it configured?
- Should there be separate `startupIndicators`?
- Is it an error to register indicators for unused services? Or should they just pass?

### Default Indicators

Serverpod will provide built-in indicators for common dependencies:

DatabaseHealthIndicator: Tests database connectivity using a lightweight query. This builds on the existing `testConnection()` functionality in `health_check.dart` but exposes it through the new indicator interface.

RedisHealthIndicator: Tests Redis connectivity with a simple PING command. This is a new addition - currently Serverpod does not probe Redis health.

### Design Considerations for Custom Indicators

Timeout Enforcement: The `HealthCheckManager` enforces the timeout specified by each indicator. If a check takes too long, it's considered failed. This prevents a hung external service from making the health endpoint itself unresponsive.

Failure Isolation: Each indicator runs independently. A failure in one check doesn't prevent other checks from completing. The final health status aggregates all results.

Caching: Health checks can be expensive (database queries, network calls). The manager caches results for a configurable duration (default: 1 second) to prevent a "thundering herd" on the health endpoint itself during high-frequency probing.

## Response Format

Health endpoints should return JSON following the draft RFC: _Health Check Response Format for HTTP APIs_ (https://datatracker.ietf.org/doc/html/draft-inadarei-api-health-check-06) for (hopefully) broad compatibility with monitoring tools:

```json
{
  "status": "pass",
  "checks": {
    "database:latency": [
      {
        "componentId": "primary-db",
        "componentType": "datastore",
        "observedValue": 12,
        "observedUnit": "ms",
        "status": "pass",
        "time": "2026-01-14T10:30:00Z"
      }
    ],
    "redis:latency": [
      {
        "componentId": "cache-cluster",
        "componentType": "datastore",
        "observedValue": 3,
        "observedUnit": "ms",
        "status": "pass",
        "time": "2026-01-14T10:30:00Z"
      }
    ]
  }
}
```

For unhealthy states:

```json
{
  "status": "fail",
  "notes": "Connection timeout while reaching the primary database cluster.",
  "output": "ConnectionError: ETIMEDOUT 10.0.1.5:5432",
  "checks": {
    "database:connection": [
      {
        "componentId": "primary-db",
        "componentType": "datastore",
        "status": "fail",
        "time": "2026-01-19T09:50:00Z",
        "output": "timeout after 5000ms"
      }
    ],
    "redis:latency": [
      {
        "componentId": "cache-cluster",
        "status": "pass",
        "observedValue": 2,
        "observedUnit": "ms",
        "time": "2026-01-19T09:50:00Z"
      }
    ]
  }
}
```

HTTP status codes:
- `200 OK` - All checks passed
- `503 Service Unavailable` - One or more critical checks failed

The `/livez` endpoint returns a minimal response since it performs no checks:

```json
{
  "status": "pass",
  "time": "2026-01-19T10:30:00Z"
}
```

Question: Should we support `"warn"` state as well? (Supported by RFC)

## Future Consideration: Multi-Isolate Architecture

> [!NOTE]
> This section can be disregarded for now. It has not been decided to use multiple isolates.

Serverpod currently runs in a single isolate. However, if Serverpod moves to a multi-isolate architecture in the future (for better CPU utilization on multi-core machines), the health check system will need to account for this. This section outlines how the design would extend.

### The Challenge

In a multi-isolate architecture, the main isolate handles incoming requests and coordinates worker isolates that execute application logic. This introduces complexity for health checks:

- A problem in one isolate doesn't necessarily indicate a problem with the entire pod
- We need to avoid killing the entire pod because one worker is temporarily slow
- The health endpoint runs on the main isolate but needs visibility into worker health

### Permissive Liveness in Multi-Isolate

The pod stays "Live" as long as the main isolate is responding.

If the main isolate can respond to the health check, the process is fundamentally alive - even if some worker isolates are slow or blocked. Common scenarios where we should NOT restart:

- A single worker is blocked on a slow database query
- Garbage collection pause affecting some workers
- One worker processing an unusually large request

In all these cases, the affected worker will typically recover on its own. Restarting would destroy work in progress across all isolates.

### Sensitive Readiness with Worker Awareness

While liveness stays permissive, readiness could incorporate worker health:

Workers would periodically measure their event loop responsiveness by scheduling a zero-delay timer and measuring actual execution delay:

```dart
// Conceptual implementation
final scheduledTime = DateTime.now();
Timer(Duration.zero, () {
  final actualTime = DateTime.now();
  final lag = actualTime.difference(scheduledTime);
  reportLagToMainIsolate(lag);
});
```

If the event loop is blocked (by synchronous code, GC, or I/O wait), the timer executes late, and this latency is detected.

Rather than failing readiness when any single worker is slow, we would use a quorum-based approach:

- If an isolate reports lag > 500ms (configurable), it is marked as "Busy"
- If more than a configurable ratio of isolates (e.g., 50%) are "Busy," `/readyz` returns `503`
- This means one slow worker out of eight doesn't affect traffic, but widespread slowness pauses traffic

Configuration additions for multi-isolate:

```yaml
health:
  readyz:
    # ... existing options ...
    isolate_lag_threshold_ms: 500
    isolate_unhealthy_ratio: 0.5
```

### Communication Pattern

Workers would report their status to the main isolate via `SendPort`/`ReceivePort`. The `HealthCheckManager` on the main isolate would aggregate these reports when responding to `/readyz` probes.

This architecture ensures that health decisions are made centrally with full visibility, while keeping the liveness check simple and permissive.

## Relationship to Existing Health System

The existing health system serves multiple purposes that should be preserved:

| Aspect    | Current `GET /`        | Current Insights API  | Current Periodic Logging | Proposed System        |
| --------- | ---------------------- | --------------------- | ------------------------ | ---------------------- |
| Purpose   | Quick health check     | Programmatic access   | Historical analysis      | Orchestrator lifecycle |
| Consumer  | Load balancers, humans | Serverpod Insights UI | Dashboards, alerting     | Kubernetes probes      |
| Frequency | On-demand              | On-demand             | Periodic (configurable)  | On-demand              |
| Output    | Plain text             | `ServerHealthResult`  | Database records         | JSON HTTP responses    |

### Migration Path

The proposed system should coexist with and eventually supersede the current `GET /` endpoint. Phases:

1. Add `/livez`, `/readyz`, `/startupz` alongside the existing `GET /` endpoint
2. Deprecate `GET /` for health checking (it may still serve other purposes like a welcome message)
3. Document migration path for users currently relying on `GET /`

The new indicators can reuse the existing `testConnection()` logic from `health_check.dart` for database checks. The periodic logging and Insights API should continue to use `performHealthChecks()` as they do today.

## Action Items

1. Create the core manager class with:
   - Timeout enforcement per indicator
   - Result caching with configurable TTL
   - Aggregation logic for determining overall status

2. Implement built-in health indicators:
   - `DatabaseHealthIndicator` - reuses existing `testConnection()` logic from `health_check.dart`
   - `RedisHealthIndicator` - new capability, tests Redis connectivity with PING

3. Register `/livez`, `/readyz`, and `/startupz` routes in `server.dart` alongside the existing `GET /` endpoint.

4. Implement logic to track when server initialization (including migrations) is complete for the `/startupz` endpoint.

5. Add deprecation notice to the `GET /` endpoint for health check purposes, pointing users to the new endpoints.

nielsenko:
## Desciption

Add Kubernetes-style health check endpoints (`/livez`, `/readyz`, `/startupz`)

- `/livez` (Liveness) - Returns 200 if the server can respond. Only fails if the process is fundamentally broken. A failed liveness check triggers a pod restart.
- `/readyz` (Readiness) - Checks dependencies (database, Redis, custom indicators). A failed readiness check stops traffic routing without restarting.
- `/startupz` (Startup) - Checks if initialization is complete. Prevents premature liveness/readiness checks during boot.
- Extensible `HealthIndicator` system - Users can implement custom health indicators for their specific dependencies (external APIs, microservices, etc.)
- Built-in indicators - `DatabaseHealthIndicator` and `RedisHealthIndicator` are automatically registered based on configuration
- Result caching - Prevents thundering herd during high-frequency probing (configurable TTL, default 1 second)
- Timeout enforcement - Each indicator specifies its own timeout to prevent slow checks from blocking the endpoint
- Security - Unauthenticated requests receive status codes only; authenticated requests get detailed JSON responses following [RFC draft for Health Check Response Format](https://datatracker.ietf.org/doc/html/draft-inadarei-api-health-check-06)

### Usage

```dart
final pod = Serverpod(
  args,
  Protocol(),
  Endpoints(),
  healthConfig: HealthConfig(
    cacheTtl: Duration(seconds: 2),
    readinessIndicators: [
      StripeApiIndicator(),      // Custom indicator
      InventoryServiceIndicator(),
    ],
    startupIndicators: [
      CacheWarmupIndicator(),
    ],
  ),
);
```

### Design Document

See `docs/design/health_check.md` for detailed rationale, architectural principles, and future considerations.

Fixes #4662

## Pre-launch Checklist

## Breaking changes

There are no breaking changes, but the `healthCheckHandler` parameter on `Serverpod` constructor is now deprecated. Users should migrate to `healthConfig` with custom `HealthIndicator` implementations instead. The old parameter continues to work but will be removed in a future version.

## Release Notes

* **New Features**
  * Introduced health check endpoints (/livez, /readyz, /startupz) following Kubernetes conventions
  * Added extensible health indicator system with built-in database and Redis connectivity checks
  * Health responses include per-indicator results, caching with configurable TTL, and per-indicator timeouts
  * Authentication controls health probe details visibility

### DIFF
--- a/packages/serverpod_serialization/lib/src/serialization.dart
+++ b/packages/serverpod_serialization/lib/src/serialization.dart
@@ -220,10 +220,10 @@ abstract class SerializationManager {
         List<dynamic> l => [
           for (final i in l) _toEncodable(i, encodeForProtocol),
         ],
-        Map<String, dynamic> m => Map.of({
+        Map<String, dynamic> m => {
           for (final i in m.entries)
             i.key: _toEncodable(i.value, encodeForProtocol),
-        }),
+        },
         DateTime() => object.toUtc().toIso8601String(),
         ByteData() => object.base64encodedString(),
         Duration() => object.inMilliseconds,


nielsenko:
@coderabbitai review

nielsenko:
@marcelomendoncasoares Since this is purely additive and the history is quite deep, I suggest reviewing the end-result instead of commit by commit.

packages/serverpod/lib/src/server/health/database_health_indicator.dart
```diff
@@ -0,0 +1,50 @@
...
+          output: 'Database connection test returned false',
+        );
+      }
+    } catch (e) {
+      stopwatch.stop();
+      return HealthCheckResult.fail(
+        name: name,
+        componentType: 'datastore',
+        output: 'Database connection failed: $e',
+      );
```
marcelomendoncasoares: Suggestion: At least for the ones we know and use, should we have these `componentType` constants somewhere and use them here? In order to avoid repeating the string, which is error-prone - and since it is the same in other health indicators as well.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -45,6 +45,11 @@ class Serverpod {
     return _startedTime!;
   }
 
+  /// Whether the server has completed its startup sequence.
+  ///
+  /// Returns `true` if [start] has been called and completed successfully.
+  bool get isRunning => _startedTime != null;
```
marcelomendoncasoares: Suggestion: Since we don't account for the `shutdown` process in this method, the name `isRunning` can be misleading. We can change to `isStartupComplete`, `hasStarted` or something similar that is only bound to the start process.

packages/serverpod/lib/src/server/health/serverpod_startup_indicator.dart
```diff
@@ -0,0 +1,37 @@
...
+  @override
+  Duration get timeout => const Duration(seconds: 1);
+
+  @override
+  Future<HealthCheckResult> check() async {
+    if (_pod.isRunning) {
+      return HealthCheckResult.pass(
+        name: name,
+        componentType: 'system',
+      );
```
marcelomendoncasoares: Question: Does it make sense to return the start time as `observedValue` for the `ServerpodStartupIndicator`?

packages/serverpod/lib/src/server/health/health_indicator.dart
```diff
@@ -0,0 +1,161 @@
...
+  ///
+  /// Useful when there are multiple instances of the same component type,
+  /// e.g., `primary-db`, `replica-db`.
+  final String? componentId;
+
+  /// Type of the component, e.g., `datastore`, `system`, `component`.
+  final String? componentType;
+
+  /// The observed value from the check, e.g., response time in milliseconds.
+  final double? observedValue;
```
marcelomendoncasoares: Question: As [per the RFC](https://datatracker.ietf.org/doc/html/draft-inadarei-api-health-check-06#name-observedvalue), shouldn't the `observedValue` be any valid JSON type? Having it as `double` limits our possibility of sending ISO datetimes, categorical values, etc.

packages/serverpod/lib/src/server/health/health_indicator.dart
```diff
@@ -0,0 +1,161 @@
...
+      name: name,
+      status: HealthStatus.pass,
+      componentId: componentId,
+      componentType: componentType,
+      observedValue: observedValue,
+      observedUnit: observedUnit,
+      output: output,
+      time: time ?? DateTime.now().toUtc(),
+    );
+  }
```
marcelomendoncasoares: Discussion: Given that one health indicator must return several different answers and users can build custom indicators, I only worry with the fact that the `name`, `componentId`, `componentType`, `observedValue` (if we change to support arbitrary JSON) and `observedUnit` can be set to different values on each return. Specially considering that our cloud service will receive checks from user-defined indicators, this can mess things up.

To solve this, I propose we change both the `HealthCheckResult` and `HealthIndicator` to be generic to the `observedValue` type (if we allow any valid JSON object) and add three more abstract parameters to the `HealthIndicator`: `componentId`, `componentType` and `observedUnit`. Then, the factory methods of the `HealthCheckResult` could replace several arguments by the very `HealthIndicator` that is creating it, making the creation of a `pass` and `fail` statuses type-safe and stable for the indicator:

```dart
abstract class HealthIndicator<T extends Object> {
  String get name;
  String? get componentId;
  String? get componentType;
  String? get observedUnit;
  Duration get timeout => const Duration(seconds: 5);
  Future<HealthCheckResult> check();
}

class HealthCheckResult<T extends Object> {
  ...

  /// Creates a passing health check result.
  factory HealthCheckResult.pass({
    required HealthIndicator<T> indicator,
    T? observedValue,
    String? output,
    DateTime? time,
  }) {
    return HealthCheckResult(
      name: indicator.name,
      status: HealthStatus.pass,
      componentId: indicator.componentId,
      componentType: indicator.componentType,
      observedValue: observedValue,
      observedUnit: indicator.observedUnit,
      output: output,
      time: time ?? DateTime.now().toUtc(),
    );
  }
  ...
```

Then, on the indicator, users would create the result by doing:

```dart
  return HealthCheckResult.pass(
    indicator: this,
    observedValue: stopwatch.elapsedMilliseconds.toDouble(),
  );
```

What do you think?

packages/serverpod/lib/src/server/health/health_config.dart
```diff
@@ -0,0 +1,50 @@
...
+  /// Use this for checks that should only run during startup, such as
+  /// cache warming or connection pool initialization.
+  final List<HealthIndicator> startupIndicators;
+
+  /// Creates a health configuration.
+  const HealthConfig({
+    this.cacheTtl = const Duration(seconds: 1),
+    this.readinessIndicators = const [],
+    this.startupIndicators = const [],
+  });
```
marcelomendoncasoares, line 49: Suggestion: With the parameters named `readinessIndicators` and `startupIndicators`, it feels to me as if setting them would mean to replace the default ones. I get that the names are cleaner like this, but maybe adding an `extra`, `additional` or even `custom` as prefix could make the intention clearer for users (like `additionalReadinessIndicators`, `extraReadinessIndicators` or `customReadinessIndicators`).

packages/serverpod/test/server/health/health_check_result_test.dart
```diff
@@ -0,0 +1,143 @@
...
+      final result = HealthCheckResult.pass(name: 'test:indicator');
+
+      expect(result.isHealthy, isTrue);
+    });
+
+    test('when toJson is called then status is "pass"', () {
+      final result = HealthCheckResult.pass(name: 'test:indicator');
+
+      expect(result.toJson()['status'], 'pass');
+    });
```
marcelomendoncasoares, line 43: One general comment on the tests: the `Given`/`when`/`then` pattern feels a bit off. The setup should be only on the `Given` part, with the `when` containing the action and the `then` containing the result. As they are, tests are presenting part of the setup on the when clause. Also, the description on the `Given` clause is not being accompanied by the code that it's describing.

A more adherent separation could be:

```dart
  group('Given a passing health check result', () {
    late final result = HealthCheckResult.pass(name: 'test:indicator');

    test('when getting the status then it is HealthStatus.pass', () {
      expect(result.status, HealthStatus.pass);
    });

    test('when checking isHealthy then it is true', () {
      expect(result.isHealthy, isTrue);
    });

    test('when calling toJson then serialized status is "pass"', () {
      expect(result.toJson()['status'], 'pass');
    });
    ...
```

packages/serverpod/test/server/health/health_check_result_test.dart
```diff
@@ -0,0 +1,143 @@
...
+    test('when created with explicit time then that time is used', () {
+      final explicitTime = DateTime.utc(2024, 1, 15, 10, 30, 0);
+      final result = HealthCheckResult.pass(
+        name: 'test:indicator',
+        time: explicitTime,
+      );
+
+      expect(result.time, explicitTime);
+      expect(result.toJson()['time'], '2024-01-15T10:30:00.000Z');
+    });
```
marcelomendoncasoares: Following up the last comment, these should all be separate `Given`:

- Given a passing health check result created with observedValue set
- Given a passing health check result created with componentId and componentType set
- Given a passing health check result created without optional fields

Since they are all groups of one test, the entire description could be on the test - with no need for a group. This comment can be considered for all subsequent tests.

marcelomendoncasoares (changes requested):
Implementation looks great! Besides other minor details, I have put up two comments for discussion:

1. Whether we should support arbitrary JSON objects as `observedValue` as the RFC states.
2. An improvement suggestion to make the creation of `HealthCheckResult` objects more type-safe and stable for a given `HealthIndicator`.

On the tests, I haven't validated all cases in-depth yet, but made one general comment on the structure to adhere more to the `Given`/`when`/`then` pattern.

packages/serverpod/lib/src/server/health/database_health_indicator.dart
nielsenko: Fixed in https://github.com/serverpod/serverpod/pull/4631/commits/56cb744847153a1e53adc5f1d32b1fad1ec5fa52

packages/serverpod/lib/src/server/serverpod.dart
nielsenko: Fixed in [6475e22](https://github.com/serverpod/serverpod/pull/4631/commits/6475e22cb49018234bd0bba215d801ac455c9ae9)

packages/serverpod/lib/src/server/health/health_config.dart
nielsenko: Fixed in e682a91

packages/serverpod/lib/src/server/health/health_indicator.dart
nielsenko: I have decided to make `HealthIndicator` generic, and make `HealthCheckResult.observedValue` an `Object?`.
`HealthCheckResult` ctor is now private, and can only be constructed via `HealthIndicator.pass/fail` which are generic on `observedValue`. This should prevent misuse. Test use `HealthCheckResultInternal.create`.

Perhaps we don't need to keep it `observedValue` nullable?

packages/serverpod/test/server/health/health_check_result_test.dart
nielsenko: Addressed in [8bf89e3](https://github.com/serverpod/serverpod/pull/4631/commits/8bf89e335c901ae24bff8254c7e8fe127b9e484c)

packages/serverpod/test/server/health/health_check_result_test.dart
nielsenko: Fixed

packages/serverpod/lib/src/server/health/health.dart
```diff
@@ -34,20 +34,25 @@
 /// Implement [HealthIndicator] to create custom checks:
 ///
 /// ```dart
-/// class StripeApiIndicator extends HealthIndicator {
+/// class StripeApiIndicator extends HealthIndicator<double> {
 ///   @override
 ///   String get name => 'stripe:api';
 ///
 ///   @override
+///   String get componentType => HealthComponentType.component;
```
marcelomendoncasoares:
```suggestion
///   String get componentType => HealthComponentType.component.name;
```

packages/serverpod/lib/src/server/health/health_indicator.dart
```diff
@@ -1,30 +1,55 @@
...
-///         output: 'Connection failed: $e',
-///       );
+///       return fail(output: 'Connection failed: $e');
 ///     }
 ///   }
 /// }
 /// ```
-abstract class HealthIndicator {
+abstract class HealthIndicator<T extends Object> {
   /// Human-readable name for this indicator.
```
marcelomendoncasoares, line 58: Suggestion: Is it worth adding a `const` constructor to allow subclasses to be `const`?

packages/serverpod/lib/src/server/health/health_indicator.dart
```diff
@@ -1,30 +1,55 @@
...
+///
 /// Example implementation:
 /// ```dart
-/// class MyServiceHealthIndicator extends HealthIndicator {
+/// class MyServiceHealthIndicator extends HealthIndicator<double> {
 ///   @override
 ///   String get name => 'myservice:connection';
 ///
 ///   @override
+///   String get componentType => HealthComponentType.component;
```
marcelomendoncasoares:
```suggestion
///   String get componentType => HealthComponentType.component.name;
```

packages/serverpod/lib/src/server/health/health_indicator.dart
```diff
@@ -151,11 +197,19 @@ class HealthCheckResult {
     return {
       if (componentId != null) 'componentId': componentId,
       if (componentType != null) 'componentType': componentType,
-      if (observedValue != null) 'observedValue': observedValue,
+      if (observedValue != null)
+        'observedValue': observedValue is DateTime
+            ? (observedValue as DateTime).toIso8601String()
+            : observedValue,
```
marcelomendoncasoares: Suggestion: Since we already have a powerful encoder and the RFC states that any valid JSON should be supported, it would be nice to reuse our encoder - which would enable support even to custom model classes.

```suggestion
      if (observedValue != null)
        'observedValue': SerializationManager.encode(observedValue),
```

packages/serverpod/lib/src/server/health/health_indicator.dart
marcelomendoncasoares: Nice! The helper methods make it even simpler to implement the indicators. On the question you made, I guess we can have the `observedValue` nullable only on the `fail` method - like not having a start time to show if it hasn't completed yet -, and make it non-nullable on `pass`.

packages/serverpod/lib/src/server/health/health_indicator.dart
```diff
@@ -0,0 +1,215 @@
...
+      if (componentId != null) 'componentId': componentId,
+      if (componentType != null) 'componentType': componentType,
+      if (observedValue != null)
+        'observedValue': observedValue is DateTime
+            ? (observedValue as DateTime).toIso8601String()
+            : observedValue,
+      if (observedUnit != null) 'observedUnit': observedUnit,
+      'status': status.name,
+      'time': time.toIso8601String(),
+      if (output != null) 'output': output,
```
marcelomendoncasoares: Question: Since you have changed all occurrences of `if (xyz != null)` for null-aware elements, is there any particular reason to keep these as `if` conditions instead of `?element`?

packages/serverpod/test/server/health/health_check_result_test.dart
```diff
@@ -0,0 +1,209 @@
...
+  group('Given HealthStatus enum', () {
+    test('then pass.name returns "pass"', () {
+      expect(HealthStatus.pass.name, 'pass');
+    });
+
+    test('then fail.name returns "fail"', () {
+      expect(HealthStatus.fail.name, 'fail');
+    });
+  });
+}
```
marcelomendoncasoares: Question: Is this test really needed? It seems to be an implementation detail that doesn't matter if the above tests of "pass" and "fail" serialization value are correct.

packages/serverpod/test/server/health/health_check_result_test.dart
```diff
@@ -0,0 +1,209 @@
...
+
+      expect(json['componentId'], 'primary-db');
+      expect(json['componentType'], 'datastore');
+      expect(json['observedValue'], 5000.0);
+      expect(json['observedUnit'], 'ms');
+      expect(json['status'], 'fail');
+      expect(json['time'], '2024-01-15T10:30:00.000Z');
+      expect(json['output'], 'Connection timeout after 5000ms');
+    });
+  });
```
marcelomendoncasoares: Suggestion: This test group is still a bit off on the pattern, since the test objects are being created inside the test, but grouped by a `Given` clause that does not specify the difference in the setup. It would be better to make each of these tests have the entire description:

- Given a passing HealthCheckResult when calling toJson() then status is serialized as "pass".
- Given a failing HealthCheckResult when calling toJson() then status is serialized as "fail".
- Given a HealthCheckResult with then null optional fields when calling toJson() then null fields are omitted.
- Given a HealthCheckResult with a DateTime observed value when calling toJson() then the observedValue is serialized to ISO8601.
- Given a HealthCheckResult with all fields set to non-nullable values when calling toJson() then all set fields are included.

packages/serverpod/test/server/health/health_config_test.dart
```diff
@@ -0,0 +1,96 @@
...
+      expect(
+        config.additionalReadinessIndicators,
+        same(additionalReadinessIndicators),
+      );
+      expect(
+        config.additionalStartupIndicators,
+        same(additionalStartupIndicators),
+      );
+    });
+  });
```
marcelomendoncasoares: Suggestion: Following the same comment on the above test, to better adhere to the test description pattern, each test of this group should have its own `Given` describing which custom value is being tested (setup) - instead of describing it on the `then`.

tests/serverpod_test_server/test/health_endpoints_test.dart
```diff
@@ -0,0 +1,502 @@
...
+        Uri.http('localhost:$port', '/livez'),
+        headers: {'Authorization': 'Bearer valid'},
+      );
+
+      expect(response.statusCode, 200);
+      final json = jsonDecode(response.body) as Map<String, dynamic>;
+      expect(json['status'], 'pass');
+      expect(json['time'], isA<String>());
+    });
+  });
```
marcelomendoncasoares: Suggestion: The `when` clause is repeating on the group description and after the `then`. Since the difference between these tests are not in the setup, but only in the parameter of the action, it wouldn't be a problem to group them with descriptions like:

- Given a running server with configured authentication
  - when calling GET /livez with valid authentication then response is JSON with status pass.
  - when calling GET /livez with no authentication then response has no body.

This comment is valid for all tests in this file.

One extra note for below tests is that the action (calling /GET) is being described separately from the actual code for the action - which is inside the test. We could either:

- Move the `when` description down from the group to the test. It repeats a little, but is simpler and easy to read. The example above follows this strucuture.
- Or, for the tests that have no difference in authentication, create an intermediary group that has the `when GET /endpointz is called` description and the code for this action on the setup.

This makes the tests read a lot more fluent, since we remove from the reader the responsibility from understanding what part of the code is setup and what part is action.

tests/serverpod_test_server/test/health_endpoints_test.dart
```diff
@@ -0,0 +1,502 @@
...
+      final response = await httpClient.get(
+        Uri.http('localhost:$port', '/startupz'),
+      );
+
+      expect(response.statusCode, 200);
+    });
+
+    test(
+      'then response includes startup indicator when authenticated',
+      () async {
```
marcelomendoncasoares, line 207: Suggestion: Besides the same comment as above on the description, it is also missing the case for calling `startupz` with no authentication.

packages/serverpod/lib/src/server/health/health_indicator.dart
nielsenko: 👍

packages/serverpod/lib/src/server/health/health_indicator.dart
nielsenko: Missed it. Will fix

tests/serverpod_test_server/test/health_endpoints_test.dart
nielsenko: Addressed in a771128

tests/serverpod_test_server/test/health_endpoints_test.dart
nielsenko: Addressed in  7577113

packages/serverpod/test/server/health/health_config_test.dart
nielsenko: Addressed in a771128

packages/serverpod/test/server/health/health_check_result_test.dart
nielsenko: Addressed in a771128

packages/serverpod/test/server/health/health_check_result_test.dart
nielsenko: Removed in 9199cc9

packages/serverpod_serialization/lib/src/serialization.dart
```diff
@@ -201,6 +201,42 @@ abstract class SerializationManager {
...
+          object.toJsonForProtocol(),
+        Map(keyType: != String) =>
+          object.entries.map((e) => {'k': e.key, 'v': e.value}).toList(),
+        Record() => throw Exception(
+          'Records are not supported. '
+          'They must be converted beforehand via `Protocol.mapRecordToJson` '
+          'or the enclosing `SerializableModel`.',
+        ),
+        _ => (object as dynamic).toJson(),
+      };
```
marcelomendoncasoares, line 256: Discussion: One issue with this function is that it will only encode non-encodable types to encodable, but will fail on already-encodable types. This will be very brittle, because a `Map` that contains both `int` and `DateTime` will fail with a `toJson` being called on `int`. If we are to detach this function, we might need to also skip encodable types (which at least aren't soo many). Either way, having this function also creates more test requirements.

Don't know what is the best strategy here... Should we couple this change in the current PR? Or we ship the feature quickly with the encode/decode hack and tackle this improvement soon afterwards, but without blocking the release?

nielsenko: You are right. I wanted this function to fallback to returning `object` like:
```dart
        ...
        SerializableModel() => object.toJson(),
        _ => object,
```
But we have one test where we expect a `CustomClass2` with a just a `toJson` to serialize correctly.

I'm not sure I'll ever forgive that `toJson` doesn't require an interface.

Anyway, I have made a small work-around:
```dart
        ...
        SerializableModel() => object.toJson(),
        _ => object.safeToJson(), // <-- workaround

// ..

extension on Object? {
  /// Returns a safe JSON representation of the object.
  /// If the object does not implement the [toJson] method,
  /// it will be returned as is.
  Object? safeToJson() {
    try {
      return (this as dynamic).toJson();
    } catch (_) {
      return this;
    }
  }
}
```

Or I can roll last two commits back and just use re-work the tests? What do you think?

marcelomendoncasoares: The reason for the `toJson` not requiring an interface is to allow people to [just plug-in their existing classes](https://docs.serverpod.dev/concepts/serialization) (such as Freezed) without imposing changes to it. I agree that the unsafe is bothering for us, but it is a tradeoff for users to have more flexibility. I think the way you have done with safe conversion is the best here.

PR #4639, opened 2026-01-22
fix: Preserve ORDER BY in nested includeList queries with LIMIT

### LINKED ISSUE #4110: Problems with nested queries when using OrderBy in combination with Limit
### Describe the bug

There seems to be some bug in the final SQL query when using LIMIT in combination with ORDER BY for nested includes.

I have a hiercy of tables that looks something like this:
Node
  -> Sensor
    -> Data

I want to retrieve all the nodes with their respective sensors and the latest data for each sensor.
So I thought something like this would work:
```dart
var nodes = Node.db.find(
  session,
  include: Node.include(
    sensors: Sensor.includeList(
      include: Sensor.include(
        rawDataList: RawData.includeList(
          orderBy: (t) => t.createdAt,
          orderDescending: true,
          limit: 1,
        ),
      ),
    ),
  ),
);
```

Using no limit results in a ordered rawDataList but when adding limit 1 it seems to result in whatever order is in the database.

I used the ServerPod insight to find out that this is the query for RawData (I increased limit to 3 to get further details)
```SQL
with _base_query_sorting_and_ordering as (
select
	"raw_data"."id" as "raw_data.id",
	"raw_data"."sensorId" as "raw_data.sensorId",
	"raw_data"."temperature" as "raw_data.temperature",
	"raw_data"."createdAt" as "raw_data.createdAt"
from
	"raw_data"
where
	"raw_data"."sensorId" in ('8f105f99-db92-4ef8-bb5d-92f8def82a19', 'a1875b3f-2c5c-4c28-8c9c-71e7631839fa')
order by
	"raw_data"."createdAt" desc),
_partitioned_list_by_parent_id as (
select
	*,
	row_number() over ( partition by _base_query_sorting_and_ordering."raw_data.sensorId")
from
	_base_query_sorting_and_ordering)
select
	*
from
	_partitioned_list_by_parent_id
where
	row_number between 1 and 3
```

The problem seems to be that I am losing the ordered list between first query and the second.
If I run the first portion of the query I get a ordered list by the date time.
Image

And then if I run the full query I get "ordered" list by row_number but the date_created order seems to be lost.
Image

### To reproduce

Run a nested join query with orderBy and limit.

Should result in the orderBy not working.

### Expected behavior

I expect the orderBy to still apply with the limit, in this case I only want the latest row.

### Library version

Serverpod: 2.9.1

### Platform information

Flutter 3.35.7 • channel stable •
https://github.com/flutter/flutter.git
Framework • revision adc9010625 (6 days ago) • 2025-10-21
14:16:03 -0400
Engine • hash 6b24e1b529bc46df7ff397667502719a2a8b6b72
(revision 035316565a) (6 days ago) • 2025-10-21
14:28:01.000Z
Tools • Dart 3.9.2 • DevTools 2.48.0

### Additional context

_No response_

None

- [ ] I want to work on this

FXschwartz:
The ROW_NUMBER() window function was missing an ORDER BY clause, causing row numbers to be assigned in arbitrary order within each partition. This meant that LIMIT would return incorrect rows even when orderBy was specified.

Fixes #4110 

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
+++ b/tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
@@ -1178,9 +1178,10 @@ void main() async {
       // where the CTE's materialized order is not preserved, which is the
       // root cause of issue #4110.
       await session.db.transaction((transaction) async {
-        await session.db.unsafeExecute(
-          'SET LOCAL enable_sort = off;',
-          transaction: transaction,
+        await transaction.setRuntimeParameters(
+          (params) => [
+            MapRuntimeParameters({'enable_sort': false}),
+          ],
         );
 
         // Query pattern matching the original issue:
--- a/tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
+++ b/tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
@@ -1029,8 +1029,7 @@ void main() async {
         city.id!,
         include: City.include(
           citizens: Person.includeList(
-            orderBy: (t) => t.name,
-            orderDescending: true,
+            orderBy: (t) => t.name.desc(),
             limit: 2,
           ),
         ),
--- a/tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
+++ b/tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
@@ -1045,4 +1045,187 @@ void main() async {
       );
     },
   );
+
+  test(
+    'Given a deeply nested includeList with orderBy and limit when querying with include then the innermost list is correctly ordered and limited.',
+    () async {
+      // Mirrors the original issue structure: Node -> Sensor -> Data
+      // Using: City -> Organization (includeList) -> include -> Person (includeList with orderBy + limit)
+      var city = await City.db.insertRow(session, City(name: 'Stockholm'));
+
+      var orgAlpha = await Organization.db.insertRow(
+        session,
+        Organization(name: 'Alpha'),
+      );
+      var orgBeta = await Organization.db.insertRow(
+        session,
+        Organization(name: 'Beta'),
+      );
+
+      await City.db.attach.organizations(session, city, [orgAlpha, orgBeta]);
+
+      // Create people and attach in non-alphabetical order
+      var alice = await Person.db.insertRow(session, Person(name: 'Alice'));
+      var bob = await Person.db.insertRow(session, Person(name: 'Bob'));
+      var charlie = await Person.db.insertRow(session, Person(name: 'Charlie'));
+      var diana = await Person.db.insertRow(session, Person(name: 'Diana'));
+
+      var eve = await Person.db.insertRow(session, Person(name: 'Eve'));
+      var frank = await Person.db.insertRow(session, Person(name: 'Frank'));
+      var grace = await Person.db.insertRow(session, Person(name: 'Grace'));
+
+      await Organization.db.attach.people(session, orgAlpha, [
+        diana, // Intentionally non-alphabetical
+        alice,
+        charlie,
+        bob,
+      ]);
+
+      await Organization.db.attach.people(session, orgBeta, [
+        grace,
+        eve,
+        frank,
+      ]);
+
+      // Query pattern matching the original issue:
+      // find -> includeList -> include -> includeList(orderBy, limit)
+      var result = await City.db.findById(
+        session,
+        city.id!,
+        include: City.include(
+          organizations: Organization.includeList(
+            orderBy: (t) => t.name,
+            include: Organization.include(
+              people: Person.includeList(
+                orderBy: (t) => t.name,
+                limit: 2,
+              ),
+            ),
+          ),
+        ),
+      );
+
+      expect(result?.organizations, hasLength(2));
+
+      // Alpha comes first alphabetically
+      var alphaResult = result?.organizations?.first;
+      expect(alphaResult?.name, 'Alpha');
+      expect(alphaResult?.people, hasLength(2));
+      expect(
+        alphaResult?.people?.map((e) => e.name).toList(),
+        ['Alice', 'Bob'],
+        reason: 'Expected first 2 people alphabetically: Alice, Bob',
+      );
+
+      // Beta comes second
+      var betaResult = result?.organizations?.last;
+      expect(betaResult?.name, 'Beta');
+      expect(betaResult?.people, hasLength(2));
+      expect(
+        betaResult?.people?.map((e) => e.name).toList(),
+        ['Eve', 'Frank'],
+        reason: 'Expected first 2 people alphabetically: Eve, Frank',
+      );
+    },
+  );
+
+  test(
+    'Given a list relation with orderBy and limit when sorting is disabled in PostgreSQL then the ORM query still returns correctly ordered rows.',
+    () async {
+      var city = await City.db.insertRow(session, City(name: 'TestCity'));
+
+      var org1 = await Organization.db.insertRow(
+        session,
+        Organization(name: 'Org1'),
+      );
+      var org2 = await Organization.db.insertRow(
+        session,
+        Organization(name: 'Org2'),
+      );
+
+      await City.db.attach.organizations(session, city, [org1, org2]);
+
+      // Insert many persons per organization in reverse alphabetical order.
+      // This creates heap storage where the physical order is opposite to
+      // the desired sort order.
+      var names1 = List.generate(
+        50,
+        (i) => 'Person_${(99 - i).toString().padLeft(2, '0')}',
+      );
+      var names2 = List.generate(
+        50,
+        (i) => 'Member_${(99 - i).toString().padLeft(2, '0')}',
+      );
+
+      var persons1 = <Person>[];
+      for (var name in names1) {
+        persons1.add(
+          await Person.db.insertRow(session, Person(name: name)),
+        );
+      }
+      var persons2 = <Person>[];
+      for (var name in names2) {
+        persons2.add(
+          await Person.db.insertRow(session, Person(name: name)),
+        );
+      }
+
+      await Organization.db.attach.people(session, org1, persons1);
+      await Organization.db.attach.people(session, org2, persons2);
+
+      // Disable sorting to force PostgreSQL to use hash-based operations
+      // for the window function's PARTITION BY. This simulates conditions
+      // where the CTE's materialized order is not preserved, which is the
+      // root cause of issue #4110.
+      await session.db.transaction((transaction) async {
+        await session.db.unsafeExecute(
+          'SET LOCAL enable_sort = off;',
+          transaction: transaction,
+        );
+
+        // Query pattern matching the original issue:
+        // find -> includeList -> include -> includeList(orderBy, limit)
+        var result = await City.db.findById(
+          session,
+          city.id!,
+          include: City.include(
+            organizations: Organization.includeList(
+              orderBy: (t) => t.name,
+              include: Organization.include(
+                people: Person.includeList(
+                  orderBy: (t) => t.name,
+                  limit: 3,
+                ),
+              ),
+            ),
+          ),
+          transaction: transaction,
+        );
+
+        expect(result?.organizations, hasLength(2));
+
+        var org1Result = result?.organizations?.firstWhere(
+          (o) => o.name == 'Org1',
+        );
+        expect(org1Result?.people, hasLength(3));
+        expect(
+          org1Result?.people?.map((e) => e.name).toList(),
+          ['Person_50', 'Person_51', 'Person_52'],
+          reason:
+              'Expected first 3 people alphabetically: Person_50, Person_51, Person_52',
+        );
+
+        var org2Result = result?.organizations?.firstWhere(
+          (o) => o.name == 'Org2',
+        );
+        expect(org2Result?.people, hasLength(3));
+        expect(
+          org2Result?.people?.map((e) => e.name).toList(),
+          ['Member_50', 'Member_51', 'Member_52'],
+          reason:
+              'Expected first 3 people alphabetically: Member_50, Member_51, Member_52',
+        );
+      });
+    },
+  );
 }
--- a/tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
+++ b/tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
@@ -932,4 +932,117 @@ void main() async {
       expect(city?.citizens?.map((e) => e.id), [person3.id, person2.id]);
     },
   );
+
+  test(
+    'Given a list relation with many rows when including the list with orderBy and a constraining limit then the result contains the correctly ordered first N rows.',
+    () async {
+      var stockholm = await City.db.insertRow(session, City(name: 'Stockholm'));
+      var gothenburg = await City.db.insertRow(
+        session,
+        City(name: 'Gothenburg'),
+      );
+
+      // Create persons with names that sort alphabetically: Alice, Bob, Charlie, Diana
+      var alice = await Person.db.insertRow(session, Person(name: 'Alice'));
+      var bob = await Person.db.insertRow(session, Person(name: 'Bob'));
+      var charlie = await Person.db.insertRow(session, Person(name: 'Charlie'));
+      var diana = await Person.db.insertRow(session, Person(name: 'Diana'));
+
+      // Create persons for second city: Eve, Frank, Grace
+      var eve = await Person.db.insertRow(session, Person(name: 'Eve'));
+      var frank = await Person.db.insertRow(session, Person(name: 'Frank'));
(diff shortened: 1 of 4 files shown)


packages/serverpod/lib/src/database/sql_query_builder.dart
```diff
@@ -143,6 +150,29 @@ class SelectQueryBuilder {
...
+          var clause = '$tableAlias."$alias"';
+          if (order.orderDescending) {
+            clause += ' DESC';
+          }
+          return clause;
+        })
+        .join(', ');
+
+    return ' ORDER BY $orderClauses';
+  }
```
marcelomendoncasoares, line 185: Question: Can this method be unified with `_buildOrderByQuery`? I understand that there are subtle differences (like the usage of table aliases to make it safer), but is it possible to make these subtleties configurable on the existing method?

marcelomendoncasoares (changes requested):
Another nice and clean PR, @FXschwartz! Only one question to reduce code duplication (if you think it make sense). Otherwise, I'll suggest adding an integration test that validates the order so we can have the issue scoped and covered for future changes.

packages/serverpod/lib/src/database/sql_query_builder.dart
FXschwartz: It's certainly possible but I would argue that the complexity tradeoff would not be worth it in my opinion. In addition to the usage of table aliases `_buildOrderByQuery` has special handling for ColumnCount columns that require subqueries and use `_formatOrderByCount` and `_buildWindowOrderByClause` doesn't need this. They also have different return values and output format.

Completely happy to merge them together if you'd still like me to though!

FXschwartz:
@marcelomendoncasoares Added integration tests for this! But I am seeing several failed tests that appear to be related to auth?

packages/serverpod/lib/src/database/sql_query_builder.dart
marcelomendoncasoares: That's a fair point. We can stick with them separate if it makes clearer and improves maintainability :)

marcelomendoncasoares:
> @marcelomendoncasoares Added integration tests for this! But I am seeing several failed tests that appear to be related to auth?

It seems [`file_picker` updated very recently](https://github.com/miguelpruivo/flutter_file_picker/issues/1946) with [a breaking change on the patch version](https://github.com/miguelpruivo/flutter_file_picker/issues/1951). Lets await a day to see if it gets fixed by the package, [since maintainers are already looking into it](https://github.com/miguelpruivo/flutter_file_picker/pull/1953#issuecomment-3807312570). Otherwise we'll lock the patch version on main for the CI to pass until it gets fixed.

tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
```diff
@@ -941,4 +941,117 @@ void main() async {
...
+
+      expect(result?.citizens, hasLength(2));
+      expect(
+        result?.citizens?.map((e) => e.name).toList(),
+        ['Diana', 'Charlie'],
+        reason:
+            'Expected first 2 citizens reverse-alphabetically: Diana, Charlie',
+      );
+    },
+  );
```
marcelomendoncasoares, line 1046: These tests are passing even without the fix, so they are not exactly scoping the problem. We should try to have something similar to what is reported on the issue and check that it breaks without the fix on this PR.

marcelomendoncasoares (changes requested):
Thanks for the quick iteration! But we still need to revisit the integration test, since it is not catching the problem.

tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
FXschwartz: Ok this is an odd one. I ran the tests several times and you are right they do pass without the fix. But I "think" the reason for this is Postgres row ordering without ORDER BY in the window function is undefined but deterministic. It could be that Postgres is returning the correct row ordering by itself in this specific instance which is causing the tests to pass.

If you have any ideas on how to validate this I'd love to try as I'm not really sure the best approach.

I will say however that the unit tests are validating here since they verify the generated SQL contains the ORDER BY clause. They fail immediately when the fix is disabled.

vicajilau:
@marcelomendoncasoares We apologize for any inconvenience caused. A new version 10.3.10 of file_picker was released a few minutes ago.

marcelomendoncasoares:
> @marcelomendoncasoares We apologize for any inconvenience caused. A new version 10.3.10 of file_picker was released a few minutes ago.

Thanks for the quick action on this, @vicajilau! No worries, we know these things happen on open-source. I'd just recommend an additional action to retract the `10.3.9` on `pub.dev` to prevent users from inadvertently download it - even though the chances are low with a greater patch version available.

tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
marcelomendoncasoares: The main concern here is that we ensure that this fix will resolve the issue that it was derived from. The unit tests assert that the introduced order by is correctly generated, but it does not answer the questions:

- Is this change needed?
- Does it fix the original issue?

Not questioning the value of the change, because it clearly makes the query safer. But just to justify the work on this correctly - and be able to close the issue with it.

From the original case, the problem happened with nested includes:

```dart
var nodes = Node.db.find(
  session,
  include: Node.include(
    sensors: Sensor.includeList(
      include: Sensor.include(
        rawDataList: RawData.includeList(
          orderBy: (t) => t.createdAt,
          orderDescending: true,
          limit: 1,
        ),
      ),
    ),
  ),
);
```

So maybe we can try to use a similar nested structure to reproduce?

marcelomendoncasoares:
@FXschwartz We are also having one test failure that is cause by the changes on this PR:

```console
serverpod_test_server-1  | SERVERPOD version: 3.2.3, dart: 3.8.0 (stable) (Wed May 14 09:07:14 2025 -0700) on "linux_x64", time: 2026-01-28 12:05:57.842646Z
serverpod_test_server-1  | runMode: production
serverpod_test_server-1  | serverId: default
serverpod_test_server-1  | role: monolith
serverpod_test_server-1  | loggingMode: normal
serverpod_test_server-1  | applyMigrations: false
serverpod_test_server-1  | applyRepairMigration: false
serverpod_test_server-1  | SERVERPOD initialized, time: 2026-01-28 12:05:57.957006Z
postgres-1               | 2026-01-28 12:05:58.870 UTC [417] ERROR:  column _base_query_sorting_and_ordering.person_organization_organization_people_person.id does not exist at character 1526
postgres-1               | 2026-01-28 12:05:58.870 UTC [417] STATEMENT:  WITH _base_query_sorting_and_ordering AS (WITH "order_by_person_organization_organization_people_person_0" AS (SELECT "organization"."id" AS "organization.id", COUNT("organization_people_person"."organizationId") AS "count" FROM "organization" LEFT JOIN "person" AS "organization_people_person" ON "organization"."id" = "organization_people_person"."organizationId" WHERE "organization_people_person"."name" ILIKE 'A%' GROUP BY "organization"."id") SELECT "person"."id" AS "person.id", "person"."name" AS "person.name", "person"."organizationId" AS "person.organizationId", "person"."_cityCitizensCityId" AS "person._cityCitizensCityId", "person_organization_organization"."id" AS "person_organization_organization.id", "person_organization_organization"."name" AS "person_organization_organization.name", "person_organization_organization"."cityId" AS "person_organization_organization.cityId" FROM "person" LEFT JOIN "organization" AS "person_organization_organization" ON "person"."organizationId" = "person_organization_organization"."id" LEFT JOIN "order_by_person_organization_organization_people_person_0" ON "person_organization_organization"."id" = "order_by_person_organization_organization_people_person_0"."organization.id" WHERE "person"."_cityCitizensCityId" IN (22) ORDER BY "order_by_person_organization_organization_people_person_0"."count" DESC NULLS LAST), _partitioned_list_by_parent_id AS (SELECT *, row_number() OVER ( PARTITION BY _base_query_sorting_and_ordering."person._cityCitizensCityId" ORDER BY _base_query_sorting_and_ordering."person_organization_organization_people_person.id" DESC) FROM _base_query_sorting_and_ordering) SELECT * FROM _partitioned_list_by_parent_id WHERE row_number BETWEEN 1 AND 2
serverpod_test_server-1  | +414 -1: test_integration/database_operations/changed_id_type_relations/one_to_many/find/list_include_test.dart: Given a list relation in a list relation when filtering the nested list on a count and limiting the result then only the selected rows are included and the size is the same as the limit. [E]
serverpod_test_server-1  |   DatabaseQueryException: { message: column _base_query_sorting_and_ordering.person_organization_organization_people_person.id does not exist, code: 42703, position: 1526 }
serverpod_test_server-1  |   package:postgres/src/v3/connection.dart 87:30                                                                  _PgSessionBase._sendAndWaitForQuery
serverpod_test_server-1  |   package:postgres/src/v3/connection.dart 195:11                                                                 _PgSessionBase._prepare
serverpod_test_server-1  |   package:postgres/src/v3/connection.dart 171:30                                                                 _PgSessionBase.execute
serverpod_test_server-1  |   package:postgres/src/pool/pool_impl.dart 298:24                                                                _PoolConnection.execute
serverpod_test_server-1  |   package:postgres/src/pool/pool_impl.dart 63:54                                                                 PoolImplementation.execute.<fn>
serverpod_test_server-1  |   package:postgres/src/pool/pool_impl.dart 153:24                                                                PoolImplementation.withConnection
serverpod_test_server-1  |   ===== asynchronous gap ===========================
serverpod_test_server-1  |   package:serverpod/src/database/adapters/postgres/database_connection.dart 527:20                               DatabaseConnection._query
serverpod_test_server-1  |   ===== asynchronous gap ===========================
serverpod_test_server-1  |   package:serverpod/src/database/adapters/postgres/database_connection.dart 626:18                               DatabaseConnection._mappedResultsQuery
serverpod_test_server-1  |   ===== asynchronous gap ===========================
serverpod_test_server-1  |   package:serverpod/src/database/adapters/postgres/database_connection.dart 786:33                               DatabaseConnection._queryIncludedLists
serverpod_test_server-1  |   ===== asynchronous gap ===========================
serverpod_test_server-1  |   package:serverpod/src/database/adapters/postgres/database_connection.dart 666:33                               DatabaseConnection._deserializedMappedQuery
serverpod_test_server-1  |   ===== asynchronous gap ===========================
serverpod_test_server-1  |   package:serverpod/src/database/adapters/postgres/database_connection.dart 99:16                                DatabaseConnection.findFirstRow
serverpod_test_server-1  |   ===== asynchronous gap ===========================
serverpod_test_server-1  |   package:serverpod/src/database/adapters/postgres/database_connection.dart 124:12                               DatabaseConnection.findById
serverpod_test_server-1  |   ===== asynchronous gap ===========================
serverpod_test_server-1  |   test_integration/database_operations/changed_id_type_relations/one_to_many/find/list_include_test.dart 919:18  main.<fn>
serverpod_test_server-1  |  
```

> Tip: When searching for errors on the integration tests, put "-1:" on the search box. This will lead to the failure directly without having to scan the logs. We still want to clean it up and make the experience with reading them smooth, but this is the best way to search until then.

tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
FXschwartz: Hmm so following your example above I created a new nested include test mirroring the original issue structure (City → Organization → Person with orderBy/limit on the innermost level), but it still passed without the fix.

The ORDER BY clause is missing from the window function without the fix. This is what I _**think**_ caused the original issue - without explicit ordering in the window function, PostgreSQL returns rows in an undefined order.

What if we modified the integration test to check the actual SQL that's used? Here is the output of that test

With fix:                                                                                                                            
```sql
row_number() OVER (PARTITION BY ... ORDER BY _base_query_sorting_and_ordering."person.name")
```                               
                                                                                                                                       
Without fix:                                                                                                                         
```sql
row_number() OVER (PARTITION BY _base_query_sorting_and_ordering."person.organizationId")
```

I totally get this is nice to validate the SQL is correct but doesn't directly validate that the original issue is fixed.

FXschwartz:
@marcelomendoncasoares All checks are now passing and ready for another review! I also left another comment above about not having a good solution for that test. Let me know your thoughts!

tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
marcelomendoncasoares: But the structure with the City/Organization/Person didn't follow the same types of relations as the example from the issue, right? With include -> includeList -> include -> includeList -> orderBy. I'd suggest trying that exact same setup so we can justify this change as the fix. Specially since this is a sensible change that touches the core of relation queries and could have unexpected side-effects (bugs, performance) if we're not doing it for a justified reason.

FXschwartz: @marcelomendoncasoares So I tried the exact pattern you suggested with include -> includeList -> include -> includeList -> orderBy using City → Organization → Person models, but the test still passes without the fix — even after 20+ runs. The problem I believe is that under normal conditions PostgreSQL happens to preserve the CTE's materialization order, so row_number() assigns numbers in the "right" order by coincidence. 

However, I did find a way to reliably reproduce the actual bug. It appears the root cause is that when PostgreSQL uses hash-based partitioning instead of sort-based (which can happen with larger datasets, different plans, etc.), the row order within each partition gets scrambled. By using `SET LOCAL enable_sort = off` in a transaction, I can force that behavior and simulate the conditions the bug reporter was hitting.

I wrote a new integration test that executes the same raw SQL both with and without ORDER BY in the window function and confirms they return different results:

  - With ORDER BY (fix): [Member_50, Member_51, Member_52, Person_50, Person_51, Person_52]
  - Without ORDER BY (bug): [Member_50, Member_51, Member_52, Person_50, Person_51, Person_99]

I know this uses raw SQL rather than going through the ORM which isn't ideal. One thing we could try is running the actual ORM query inside a transaction where we've set enable_sort = off — findById accepts a transaction parameter so it might propagate to the internal includeList sub-queries. Want me to give that a shot?

marcelomendoncasoares: This is a great finding! I think you have found the test we need, then. Go ahead with creating a test setting `enable_sort` to `off` on the transaction to compare before and after the fix. If it isolates the problem, even though the real cause is not this parameter but the data condition that leads to Postgres changing the sort strategy, it serves the purpose of ensuring the reproducibility.

FXschwartz: Test added!

FXschwartz:
@marcelomendoncasoares Raw SQL test added and all checks are passing. Ready for another review!

tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
```diff
@@ -938,4 +938,300 @@ void main() async {
...
+
+      // Disable sorting to force PostgreSQL to use hash-based operations
+      // for the window function's PARTITION BY. This simulates conditions
+      // where the CTE's materialized order is not preserved, which is the
+      // root cause of issue #4110.
+      await session.db.transaction((transaction) async {
+        await session.db.unsafeExecute(
+          'SET LOCAL enable_sort = off;',
+          transaction: transaction,
+        );
```
marcelomendoncasoares, line 1184: Suggestion: Expose this `enable_sort` as a runtime parameter and set through the official interface.

tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
marcelomendoncasoares, line 1: Suggestion: Replicate this same test file to the SQLite test project.

marcelomendoncasoares (changes requested):
After a long long time, two minor comments for us to proceed to merge :)

tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
FXschwartz: Done! I switched from the raw SET LOCAL enable_sort = off to the official interface. I also reconfirmed it still scopes the bug: with the fix reverted the test fails `['Person_99', 'Person_50', 'Person_51']` instead of `['Person_50', 'Person_51', 'Person_52']`, and passes with the fix in place.

tests/serverpod_test_server/test_integration/database_operations/entity_relations/one_to_many/find/list_include_test.dart
FXschwartz: Replicated the ordering tests to serverpod_test_sqlite! But just to note I had to leave the `enable_sort-disabled` reproduction test out of the SQLite suite since it has no equivalent planner GUC and theres no way to force it to not use the sorting.

marcelomendoncasoares (approved):
LGTM! 🚀

PR #4546, opened 2026-01-07
feat: Add GitHub identity provider support

vfiruz97:
This PR adds GitHub OAuth authentication support to the serverpod_auth module. It implements a complete GitHub identity provider (IDP) with PKCE (Proof Key for Code Exchange) flow for enhanced security across all platforms.

**Key features:**
- OAuth 2.0 authorization code flow with PKCE for secure authentication without client secrets
- Cross-platform support (iOS, Android, Web, macOS, Windows, Linux)
- Customizable sign-in button widget with GitHub branding
- Complete server-side token exchange, user authentication, and account management
- Integration with existing Serverpod auth system
- Comprehensive test coverage including integration tests

**Dependencies added:**

This PR adds two dependencies to `serverpod_auth_idp_flutter`:

- **`flutter_web_auth_2: ^4.1.0`** - Required for launching GitHub OAuth flow across all platforms. Well-maintained with permissive BSD-3-Clause license. It was difficult to find a Flutter package specifically for GitHub OAuth client implementation. This is a general-purpose OAuth package that works with any OAuth server supporting custom URL redirects, making it the best choice available. Other packages evaluated (github_signin_promax, github_oauth, github_sign_in_plus, oauth2_client) are either GitHub-specific but unmaintained, or lack proper cross-platform support.

- **`crypto: ^3.0.6`** - Official Dart SDK package required for PKCE implementation (SHA-256 hashing). Essential for secure OAuth 2.0 flow. Already used throughout Serverpod ecosystem.

Fixes #3924.

## Pre-launch Checklist

- [x] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [x] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
- [x] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.
- [x] I added new tests to check the change I am making.
- [x] All existing and new tests are passing.
- [x] Any breaking changes are documented below.

If you need help, consider asking for advice on the [discussion board](https://github.com/serverpod/serverpod/discussions).

## Breaking changes

[...]

PR #4642, opened 2026-01-24
feat(auth): Adds Twitch identity provider support

89pleasure:
This PR adds Twitch OAuth authentication support to the serverpod_auth module. It implements a complete Twitch identity provider (IDP) with PKCE (Proof Key for Code Exchange) flow for enhanced security across all platforms. This is possible to the great work of @vfiruz97

**Key features:**

Cross-platform support (iOS, Android, Web, macOS, Windows, Linux)
Customizable sign-in button widget with Twitch branding
Complete server-side token exchange, user authentication, and account management
Integration with existing Serverpod auth system
Comprehensive test coverage including integration tests

**No new dependencies added**

## Pre-launch Checklist

## Breaking changes

This is a new feature. I changed the `exchangeCodeForToken` to a generic method to support a different response schemas. Twitch is using a response token approach similar to Apple which won't work the way it is implemented right now. Since im new to dart development this needs further testing.


89pleasure:

89pleasure:
Still in progress, just wanted to save and push my work already :)

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/common/oauth2_pkce/oauth2_pkce_server_config.dart
```diff
@@ -0,0 +1,67 @@
...
+  header,
+
+  /// Send credentials as parameters in the request body.
+  ///
+  /// The client ID and secret are sent as form parameters in the token request.
+  body,
+}
+
+/// Signature for parsing the data from the provider's token response.
+typedef ParseTokenResponse<T> = T Function(Map<String, dynamic> responseBody);
```
89pleasure, line 21: @vfiruz97 What do you think about that? I needed to change something on that, since Twitch for example gives back more than an access_token String. Instead it has refresh_token and more in the response which is needed. 

Im not very used to generics, but I found it a good way to make it possible to define the success response body of a token exchange individually for each idp.

marcelomendoncasoares:
Thanks for this valuable contribution, @89pleasure! Since you had to base your work from #4546, the diff is carrying a lot of changes and will make the revision harder. We are still waiting for the docs, but I'll merge #4546 already so you can rebase from main!

89pleasure:
> Thanks for this valuable contribution, @89pleasure! Since you had to base your work from #4546, the diff is carrying a lot of changes and will make the revision harder. We are still waiting for the docs, but I'll merge #4546 already so you can rebase from main!

Thanks @marcelomendoncasoares ! Im on it. But I recognized that the merge into main has changes the PR does not have. Seems to be force pushed too. That makes a rebase on main pretty hard to distinguish the changes now.

marcelomendoncasoares:
> Thanks @marcelomendoncasoares ! Im on it. But I recognized that the merge into main has changes the PR does not have. Seems to be force pushed too. That makes a rebase on main pretty hard to distinguish the changes now.

You should do a rebase using the `--onto` option to only move your commits on top of latest main. Since all merges in Serverpod are done with squash, trying to rebase the actual branch will carry all commits of the last branch as well - down until the common ancestor of the main commit where the GitHub branch was based upon. Also, if your branch was not based on the last commit of the other PR, then you will indeed face conflicts on newer code.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/common/oauth2_pkce/oauth2_pkce_server_config.dart
vfiruz97: > @vfiruz97 What do you think about that? I needed to change something on that, since Twitch for example gives back more than an access_token String. Instead it has refresh_token and more in the response which is needed.

Hi @89pleasure,


Thanks for pointing this out, you're right. Providers often return more than just `access_token`, like `refresh_token` and `expires_in`, so it makes sense to leave room for that.

I looked into the generics approach you suggested. While it's very flexible, it adds some cascading complexity. We'd need to pass <T> through several classes (`OAuth2PkceServerConfig`, `OAuth2PkceUtil`, etc.), which makes the API harder to use and reason about.

```diff
-class OAuth2PkceServerConfig {
+class OAuth2PkceServerConfig<T> {
-  final ParseAccessToken parseAccessToken;
+  final ParseAccessToken<T> parseAccessToken;
-class OAuth2PkceUtil {
+class OAuth2PkceUtil<T> {
-  final OAuth2PkceServerConfig config;
+  final OAuth2PkceServerConfig<T> config;
-  Future<String> exchangeCodeForToken({
+  Future<T> exchangeCodeForToken({
-  late final OAuth2PkceServerConfig oauth2Config;
+  late final OAuth2PkceServerConfig<String> oauth2Config;
-    _oauth2Util = OAuth2PkceUtil(config: config.oauth2Config);
+    _oauth2Util = OAuth2PkceUtil<String>(config: config.oauth2Config);
```

**A simpler alternative**

Instead of generics, we could use a dedicated response object. This supports the Twitch use case while keeping the developer experience clean and straightforward:

```dart
class OAuth2PkceTokenResponse {
  final String accessToken;
  final String? refreshToken;
  final int? expiresIn;
  final String? tokenType;

  /// Raw parsed response for provider-specific fields
  final Map<String, dynamic> raw;

  const OAuth2PkceTokenResponse({
    required this.accessToken,
    this.refreshToken,
    this.expiresIn,
    this.tokenType,
    required this.raw,
  });
}

....

typedef ParseTokenResponse = OAuth2PkceTokenResponse Function(Map<String, dynamic> responseBody);
```

No generic pollution, extensible, better DX. What do you think about moving in this direction? 
I'm happy to hear @marcelomendoncasoares's thoughts on this as well.

vfiruz97:
> We are still waiting for the docs, but I'll merge #4546 already so you can rebase from main!

Thanks @marcelomendoncasoares, I am happy, finally to get my PR merged! The docs on its way. Needs your attention :) https://github.com/serverpod/serverpod_docs/pull/402

89pleasure:
> > Thanks @marcelomendoncasoares ! Im on it. But I recognized that the merge into main has changes the PR does not have. Seems to be force pushed too. That makes a rebase on main pretty hard to distinguish the changes now.
> 
> You should do a rebase using the `--onto` option to only move your commits on top of latest main. Since all merges in Serverpod are done with squash, trying to rebase the actual branch will carry all commits of the last branch as well - down until the common ancestor of the main commit where the GitHub branch was based upon. Also, if your branch was not based on the last commit of the other PR, then you will indeed face conflicts on newer code.

I look into it, I already did a rebase, but for some reason the commits of the GitHub idp are still in here. I'll redo it with the --onto flag. @marcelomendoncasoares

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/common/oauth2_pkce/oauth2_pkce_server_config.dart
marcelomendoncasoares: Agree with the direction, @vfiruz97! This is a very nice suggestion that keeps the API clean. It also allows users to extract other custom properties through the `raw` field. It is worth thinking if we need the `tokenType` on the interface or treat it as one of these custom fields that is not part of a broader standard.

vfiruz97: Thanks @marcelomendoncasoares ! I agree, `tokenType` is quite rare and not consistently needed across providers. Treating it as part of the `raw` field feels sufficient and keeps the interface simpler, while still allowing advanced use cases when required.

vfiruz97: Thanks both! I've implemented this along with some other improvements in the GitHub IDP and opened a #4676 here.

marcelomendoncasoares:
@89pleasure Now that we have merged the GitHub IDP with the changes you proposed, my suggestion will be for you to rebase your commits on latest main and open a new PR with a clean history. This will simplify the conflict resolution for you and also facilitate the revision a lot. What do you think?

89pleasure:
@marcelomendoncasoares Yes! That's great news. I was waiting for this. Also I'm currently very busy at work so I will take care of that on the weekend.

PR #4655, opened 2026-01-29
feat: Adds support for PostHog analytics.

vlidholt:
This PR adds support for anonymized PostHog analytics in the CLI.

## Pre-launch Checklist

## Breaking changes

No breaking changes.

### DIFF
--- a/tools/serverpod_cli/lib/src/util/browser_launcher.dart
+++ b/tools/serverpod_cli/lib/src/util/browser_launcher.dart
@@ -7,9 +7,12 @@ abstract final class BrowserLauncher {
   static Future<bool> openUrl(final Uri url) async {
     if (!canLaunchBrowser()) return false;
     try {
-      final result = await Process.run(_getPlatformBrowserCommand(), [
-        url.toString(),
-      ], runInShell: true);
+      final commandWithArgs = _getPlatformBrowserCommand();
+      final result = await Process.run(
+        commandWithArgs.first,
+        [...commandWithArgs.sublist(1), url.toString()],
+        runInShell: true,
+      );
       return result.exitCode == 0;
     } catch (_) {
       return false;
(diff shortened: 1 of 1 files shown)


tools/serverpod_cli/lib/src/util/browser_launcher.dart
```diff
@@ -0,0 +1,61 @@
...
+  static String _getPlatformBrowserCommand() {
+    if (Platform.isWindows) {
+      return 'start';
+    } else if (Platform.isLinux) {
+      return 'xdg-open';
+    } else if (Platform.isMacOS) {
+      return 'open';
+    }
+    throw Exception('Unsupported platform: ${Platform.operatingSystem}.');
+  }
```
marcelomendoncasoares: Testing this on the WSL worked (the URL launched), but it launched from within the WSL. This means that it won't really track anything, since the cookie will be in a browser users never use. Moreover, for users without GUI enabled on the WSL, it might not even launch.

I have tested a couple approaches and managed to launch the URL on the Windows machine from the WSL bash with the following commands:

```bash
/mnt/c/Windows/System32/cmd.exe /C start https://serverpod.dev/checkin
```
```bash
/mnt/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe -Command "Start-Process 'https://serverpod.dev/checkin'"
```

Since they use the `cmd` or `powershell` executable that comes with Windows, the path is safe to guess. Using only the alias `cmd.exe` would work if a user has the Windows PATHs in its WSL, but several users remove this to avoid conflicts - myself included.

To check that we are on the WSL and not regular Linux, we can use the very `cmd` or `powershell` path check.

```bash
[ -f /mnt/c/Windows/System32/cmd.exe ] && echo "exists" || echo "not found"
```

marcelomendoncasoares (commented):
Just one suggestion for improvement to account for WSL users! If you want, I can make the necessary changes, since I have the environment to properly test.

tools/serverpod_cli/lib/src/util/browser_launcher.dart
marcelomendoncasoares: Implemented at https://github.com/serverpod/serverpod/pull/4655/commits/f92f1eba03d17db0775a02ff7b42ae44ea5d4cff.

marcelomendoncasoares:
@vlidholt Leaving for you to merge in case you want to take a look on my changes first and adjust anything.

vlidholt:
Nice addition, thanks @marcelomendoncasoares .

Issue #3237, opened 2025-02-18
Support retention period and automated purging of log data in DB


christerswahn:
### Problem to Solve

Serverpod continuously adds log data to the database. These tables grow indefinitely unless manually truncated.

### Proposal

In order to guard against unchecked storage growth in the Serverpod database for non-business-data, we should be able to configure a log data retention period. Serverpod will automatically purge / truncate all data the log tables older than this retention period.

### Use Case

Based on a configuration / environment parameter, log data older than the duration specified in that parameter would be automatically deleted from the database. This will have no effect on running the server, except that the older log records will no longer be available.

### Alternatives

An alternative would be to add an instruction to the user documentation on how to purge / truncate these table manually, and provide a small SQL script example for this.

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

ghost:
This feature would be very useful. Our database is now extremely large and exceeds 30 GB after just a few months, while the actual data only amounts to a few hundred MB. We are therefore forced to implement a solution ourselves, which in itself should not be very complicated. But I think this feature is highly needed for a longer running production server and would be very handy to have out of the box.

vlidholt:
When working on this feature, we should also examine if there are database indexes that should be added to the session logs. It may improve performance by a lot if users have large logs.

marcelomendoncasoares:
We should implement this by combining two optional mechanisms:

- Time-based purge (older than 3 months by default).
- Count-based purge (keep only ~10_000 by default).

## Configuration

The purge mechanism have three configurations on the Serverpod `RuntimeSettings` class:

- A `Duration? logRetentionPeriod` parameter that will be used for time-based purge (default: Duration(days: 90)).
- A `int? logRetentionCount` parameter that will be used for count-based purge (default: 10_000).
- A `Duration? logCleanupInterval` parameter that will define how often to run the log cleanup routine.

All configs can be set to null to disable such mechanism. And must be also overridable through the `sessionLogs` key of the config file or environment variables. For the retention period, the config override must be `*_days`. For the interval, it must be `*_hours`.

If the `logCleanupInterval` is set to null, no regular cleanup will be made.

## Time-based purge

To implement the time-based purge efficiently, we must ensure that all required indexes for a good performance are in place.

## Count-based purge

To implement the count purge efficiently, knowing the log table uses autoincrement int ID, we can use a query like the following:

```dart
[...]

PR #4656, opened 2026-01-29
feat: Add configurable log retention and automated purging

marcelomendoncasoares:
Closes: #3237

Implements retention policies with automated purging to prevent unchecked storage growth in the log tables. Implementation follows the [design outlined in the issue](https://github.com/serverpod/serverpod/issues/3237#issuecomment-3773400812), with only the following changes:

- The defaults are:
  - retention period of 90 days
  - retention count of 100,000 entries (spec was 10,000)
  - cleanup interval of 24 hours

The cleanup is enabled by default, but can be disabled by setting the cleanup interval to null.

Three environment variables are added to configure the retention policies:
- `SERVERPOD_SESSION_LOG_RETENTION_PERIOD`
- `SERVERPOD_SESSION_LOG_RETENTION_COUNT`
- `SERVERPOD_SESSION_LOG_CLEANUP_INTERVAL`

The two time-based variables accept duration strings (retention period and cleanup interval). The duration format is the same we accept on the models.

**Example Configuration**
```yaml
sessionLogs:
  persistentEnabled: true
  retentionPeriod: 30d  # Keep 30 days
  retentionCount: 5000  # Keep max 5K entries
  cleanupInterval: 6h   # Run every 6 hours
```

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod/lib/src/server/log_manager/log_cleanup.dart
+++ b/packages/serverpod/lib/src/server/log_manager/log_cleanup.dart
@@ -11,12 +11,10 @@ class LogCleanupManager {
   final Duration? _retentionPeriod;
   final int? _retentionCount;
 
-  LogCleanupManager({
-    required Database database,
-    required SessionLogConfig config,
-  }) : _cleanupInterval = config.cleanupInterval,
-       _retentionPeriod = config.retentionPeriod,
-       _retentionCount = config.retentionCount;
+  LogCleanupManager(SessionLogConfig config)
+    : _cleanupInterval = config.cleanupInterval,
+      _retentionPeriod = config.retentionPeriod,
+      _retentionCount = config.retentionCount;
 
   DateTime? _lastCleanupTime;
   Future<void>? _activeCleanupTask;
@@ -34,7 +32,7 @@ class LogCleanupManager {
     final cleanupInterval = _cleanupInterval;
     if (cleanupInterval == null) return false;
 
-    return durationSinceLastCleanup < cleanupInterval;
+    return durationSinceLastCleanup > cleanupInterval;
   }
 
   Future<void> performCleanup(Session session) async {
--- a/packages/serverpod/lib/src/server/log_manager/log_manager.dart
+++ b/packages/serverpod/lib/src/server/log_manager/log_manager.dart
@@ -4,9 +4,8 @@ import 'dart:io';
 import 'package:meta/meta.dart';
 import 'package:serverpod/database.dart';
 import 'package:serverpod/src/server/log_manager/log_writers.dart';
-import 'package:serverpod/src/server/log_manager/log_cleanup.dart';
+import 'package:serverpod/src/server/serverpod.dart';
 import 'package:serverpod/src/server/session.dart';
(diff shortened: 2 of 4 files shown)


packages/serverpod/lib/src/server/log_manager/log_cleanup.dart
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  bool get shouldPerformCleanup {
+    final timeSinceLastCleanup = _timeSinceLastCleanup;
+    if (timeSinceLastCleanup == null) return true;
+
+    final cleanupInterval = _cleanupInterval;
+    if (cleanupInterval == null) return false;
+
+    return timeSinceLastCleanup < cleanupInterval;
```
nielsenko: This seems wrong. Should be inverted.

packages/serverpod/lib/src/server/log_manager/log_cleanup.dart
```diff
@@ -0,0 +1,105 @@
...
+
+  Duration? get _timeSinceLastCleanup {
+    final lastCleanup = _lastCleanupTime;
+    if (lastCleanup == null) return null;
+    return DateTime.now().difference(lastCleanup);
+  }
+
+  bool get shouldPerformCleanup {
+    final timeSinceLastCleanup = _timeSinceLastCleanup;
+    if (timeSinceLastCleanup == null) return true;
```
nielsenko: I guess this introduce jitter enough, that not all hosted pods run cleanup at the same time

packages/serverpod/lib/src/server/log_manager/log_cleanup.dart
```diff
@@ -0,0 +1,105 @@
...
+    required Database database,
+    required SessionLogConfig config,
+  }) : _cleanupInterval = config.cleanupInterval,
+       _retentionPeriod = config.retentionPeriod,
+       _retentionCount = config.retentionCount;
+
+  DateTime? _lastCleanupTime;
+  Future<void>? _activeCleanupTask;
+
+  Duration? get _timeSinceLastCleanup {
```
nielsenko: Using time in the name can be a bit confusing, what about `_durationSinceLastCleanup`

marcelomendoncasoares: Better indeed. Done at https://github.com/serverpod/serverpod/pull/4656/commits/1955658b71d68342480174dfb4f047d7de66d2e2.

packages/serverpod/lib/src/server/log_manager/log_cleanup.dart
marcelomendoncasoares: Thanks for catching this! The worst issue was actually that the test that should catch this was also inverted after a refactor and I didn't realize. Cleanup was always triggering due to living in the wrong place. Fixed together with moving the cleanup to the `Serverpod` object at https://github.com/serverpod/serverpod/pull/4656/commits/03c65002739d252c3794cd1b63416548d606e58f.

PR #4661, opened 2026-01-30
fix: Handle Android redirection for Apple Sign In

### LINKED ISSUE #4586: No redirect endpoint for Apple Sign In on Web/Android
### Describe the bug

Apple Sign In works natively on iOS, but requires a redirect endpoint for non-apple systems like Web or Android. 

Currently there is a placeholder route that throws a 500: https://github.com/serverpod/serverpod/blob/3620e9d295056fe38017a7b7c16ac6a82ff5210e/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/apple/business/routes/apple_server_notification_route.dart#L65

### To reproduce

1. Enable Apple IdP and call configureAppleIdpRoutes with default callback path.  
2. In Flutter, call initializeAppleSignIn with a valid redirectUri pointing at the callback route.  
3. Attempt Apple sign‑in on Android; callback returns 500 and the flow never completes.  

### Expected behavior

Callback endpoint should redirect to the app (custom scheme/intent) carrying the query params so the sign‑in flow completes on Android/Web.

### Library version

3.1.X, as well as 3.2.0

### Platform information

Web and Android

### Additional context

I've managed to work around this by implementing a custom endpoint. It would be great to have this included in the package for future use.

- [x] I want to work on this

jakubgiminski:
This PR adds support for Apple Sign In OAuth callbacks on Android. When Apple redirects back to the server after authentication on Android, the server now properly redirects to the app using an Android intent URI with the signinwithapple scheme.

**Changelog**
- Added optional androidPackageIdentifier field to AppleIdpConfig (can be set via `appleAndroidPackageIdentifier` password key)
- Updated AppleWebAuthenticationCallbackRoute to:
  - Detect Android clients via User-Agent header
  - Redirect Android clients to the app using intent URI: `intent://callback?{params}#Intent;package={pkg};scheme=signinwithapple;end`
  - Return 500 with helpful error message for web clients (not yet implemented) or when config is missing
- Added unit tests for the callback route

**Partially fixes #4586 (android only, no web)**

**Pre-launch Checklist**
- [x] I read the Contribute (https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [x] I read and followed the Dart Style Guide (https://dart.dev/guides/language/effective-dart/style) and formatted the code with dart format (https://dart.dev/tools/dart-format).
- [x] I updated/added relevant documentation (doc comments with ///), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.

**Breaking changes**
None. The `androidPackageIdentifier` configuration is optional and existing code will continue to work unchanged

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/apple/business/routes/apple_server_notification_route.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/apple/business/routes/apple_server_notification_route.dart
@@ -105,6 +105,10 @@ final class AppleWebAuthenticationCallbackRoute extends _SignInWithAppleRoute {
       h['Location'] = [intentUri];
     });
 
+    // 307 Temporary Redirect preserves the original POST method, unlike 302
+    // which may change it to GET. This is required because Apple sends the
+    // authentication callback as a POST request.
+    // See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/307
     return Response(
       307,
       headers: headers,
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/apple/routes/apple_web_authentication_callback_route_test.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/apple/routes/apple_web_authentication_callback_route_test.dart
@@ -48,7 +48,8 @@ void main() {
       });
 
       test(
-        'when Android User-Agent and POST with form body then returns 307 with correct intent URI.',
+        'when calling endpoint with Android User-Agent and POST with form body '
+        'then it returns 307 with correct intent URI.',
         () async {
           final request = _RequestFake(
             userAgent: androidUserAgent,
@@ -61,18 +62,19 @@ void main() {
           final response = result as Response;
           expect(response.statusCode, 307);
           final location = response.headers['Location']?.first;
-          expect(location, isNotNull);
-          expect(location, startsWith('intent://callback?'));
-          expect(location, contains('code=test_auth_code'));
-          expect(location, contains('state=test_state'));
-          expect(location, contains('package=$testAndroidPackageIdentifier'));
-          expect(location, contains('scheme=signinwithapple'));
-          expect(location, endsWith(';end'));
+          expect(
+            location,
+            equals(
+              'intent://callback?code=test_auth_code&state=test_state#Intent;'
+              'package=$testAndroidPackageIdentifier;scheme=signinwithapple;end',
+            ),
+          );
         },
       );
 
       test(
-        'when Android User-Agent and params have special characters then URL encodes correctly.',
+        'when calling endpoint with Android User-Agent and params have special '
+        'characters then it URL encodes correctly.',
         () async {
           final request = _RequestFake(
             userAgent: androidUserAgent,
@@ -85,14 +87,20 @@ void main() {
           final response = result as Response;
           expect(response.statusCode, 307);
           final location = response.headers['Location']?.first;
-          expect(location, isNotNull);
-          expect(location, contains('code=abc%2B123'));
-          expect(location, contains('user=%7B%22name%22%3A%22test%22%7D'));
+          expect(
+            location,
+            equals(
+              'intent://callback?code=abc%2B123&'
+              'user=%7B%22name%22%3A%22test%22%7D#Intent;'
+              'package=$testAndroidPackageIdentifier;scheme=signinwithapple;end',
+            ),
+          );
         },
       );
 
       test(
-        'when Android User-Agent and empty body then returns 307 with empty params.',
+        'when calling endpoint with Android User-Agent and empty body '
+        'then it returns 307 with empty params.',
         () async {
           final request = _RequestFake(
             userAgent: androidUserAgent,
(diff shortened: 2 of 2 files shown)


modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/apple/business/routes/apple_server_notification_route.dart
```diff
@@ -53,15 +53,85 @@ final class AppleRevokedNotificationRoute extends _SignInWithAppleRoute {
...
+
+    final headers = Headers.build((final h) {
+      h['Location'] = [intentUri];
+    });
+
+    return Response(
+      307,
+      headers: headers,
+      body: Body.empty(),
+    );
```
marcelomendoncasoares, line 116: Suggestion: Document on a comment for the reason to return 307 here (possibly with a link to the relevant documentation).

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/apple/business/apple_idp_config.dart
```diff
@@ -67,6 +75,9 @@ class AppleIdpConfigFromPasswords extends AppleIdpConfig {
         teamId: Serverpod.instance.getPasswordOrThrow('appleTeamId'),
         keyId: Serverpod.instance.getPasswordOrThrow('appleKeyId'),
         key: Serverpod.instance.getPasswordOrThrow('appleKey'),
+        androidPackageIdentifier: Serverpod.instance.getPassword(
+          'appleAndroidPackageIdentifier',
+        ),
```
marcelomendoncasoares, line 80: Suggestion: Use the `getPasswordOrThrow` method for a nice error message.

```suggestion
        androidPackageIdentifier: Serverpod.instance.getPasswordOrThrow(
          'appleAndroidPackageIdentifier',
        ),
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/apple/routes/apple_web_authentication_callback_route_test.dart
```diff
@@ -0,0 +1,220 @@
...
+      late AppleWebAuthenticationCallbackRoute route;
+
+      setUp(() {
+        route = createRoute(
+          androidPackageIdentifier: testAndroidPackageIdentifier,
+        );
+      });
+
+      test(
+        'when Android User-Agent and POST with form body then returns 307 with correct intent URI.',
```
marcelomendoncasoares: Suggestion: Clarify a bit the test action to improve readability. Same suggestion applies to other tests as well.

```suggestion
        'when calling endpoint with Android User-Agent and POST with form body '
        'then it returns 307 with correct intent URI.',
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/apple/routes/apple_web_authentication_callback_route_test.dart
```diff
@@ -0,0 +1,220 @@
...
+          final response = result as Response;
+          expect(response.statusCode, 307);
+          final location = response.headers['Location']?.first;
+          expect(location, isNotNull);
+          expect(location, startsWith('intent://callback?'));
+          expect(location, contains('code=test_auth_code'));
+          expect(location, contains('state=test_state'));
+          expect(location, contains('package=$testAndroidPackageIdentifier'));
+          expect(location, contains('scheme=signinwithapple'));
+          expect(location, endsWith(';end'));
```
marcelomendoncasoares: Suggestion: Replace by a single test that validates the entire string instead of using `contains`. Same suggestion applies to all other tests as well.

marcelomendoncasoares (changes requested):
Thanks for this contribution, @jakubgiminski!! Made only a few minor comments for improvement.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/apple/business/apple_idp_config.dart
jakubgiminski: I made `androidPackageIdentifier` not required here because the endpoint was intended to handle both web and android platforms. If the user only cares about apple sign in on web, we shouldn't force them to provide android id.
I think we can either leave this parameter as not required or split the endpoint into two - have one for web redirection and another for android. Let me know what you think would be the better approach @marcelomendoncasoares.

jakubgiminski:
Thanks for taking a look @marcelomendoncasoares. I've applied the suggested changes, except the one about requiring `androidPackageIdentifier`, provided with explanation in the comment. Looking forward to further feedback. 👍

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/apple/business/apple_idp_config.dart
marcelomendoncasoares: Thanks for clarifying! It make sense to leave it as optional indeed.

marcelomendoncasoares (approved):
Great work with this, @jakubgiminski! PR approved 🚀 

Would you mind opening a PR updating the docs on [serverpod_docs](https://github.com/serverpod/serverpod_docs) for this change?

jakubgiminski:
Thank you for the praise, it's much appreciated, @marcelomendoncasoares. 
Documentation update is ready for review here: https://github.com/serverpod/serverpod_docs/pull/422

PR #4676, opened 2026-02-02
fix: Improves GitHub IDP and OAuth2 implementations

### LINKED ISSUE #3924: Sign in with Github
Add support for signing in with Github.

vfiruz97:
This PR addresses issues found in #4546 by simplifying the GitHub IDP configuration and improving OAuth2 token handling:

**Changes:**

Removed `GitHubOAuthCredentials` class - Simplified configuration by accepting `clientId` and `clientSecret` directly in `GitHubIdpConfig` instead of wrapping them in a separate credentials class.

Added `OAuth2PkceTokenResponse` class - OAuth2 token endpoint responses. Renamed for clarity - `parseAccessToken` → `parseTokenResponse` to better reflect that it parses the complete token response, not just the access token string.

Fixes #3924

## Pre-launch Checklist

### For `GitHubIdpConfig`:

```dart
// Before
GitHubIdpConfig(
  oauthCredentials: GitHubOAuthCredentials.fromJson({
    'clientId': clientId,
    'clientSecret': clientSecret,
  }),
)

// After
GitHubIdpConfig(
  clientId: clientId,
  clientSecret: clientSecret,
)
```

### For custom OAuth2 provider implementations:

``` dart
// Before
final ParseAccessToken parseAccessToken;

// After
final ParseTokenResponse parseTokenResponse;
```

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/pubspec.yaml
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/pubspec.yaml
@@ -19,7 +19,7 @@ dependencies:
   flutter_localizations:
     sdk: flutter
   flutter_svg: ^2.2.1
-  flutter_web_auth_2: ^5.0.0
+  flutter_web_auth_2: ^5.0.1
   google_fonts: ">=6.3.2 <9.0.0"
   google_sign_in: ^7.2.0
   google_sign_in_web: ^1.1.0
--- a/templates/pubspecs/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/pubspec.yaml
+++ b/templates/pubspecs/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/pubspec.yaml
@@ -19,7 +19,7 @@ dependencies:
   flutter_localizations:
     sdk: flutter
   flutter_svg: ^2.2.1
-  flutter_web_auth_2: ^5.0.0
+  flutter_web_auth_2: ^5.0.1
   google_fonts: ">=6.3.2 <9.0.0"
   google_sign_in: ^7.2.0
   google_sign_in_web: ^1.1.0


vfiruz97:
Hi @marcelomendoncasoares, take a look please.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/common/oauth2_pkce/oauth2_pkce_token_response.dart
```diff
@@ -0,0 +1,22 @@
+/// OAuth2 token response containing access token and optional metadata.
+class OAuth2PkceTokenResponse {
+  /// The access token issued by the authorization server.
+  final String accessToken;
+
+  /// The refresh token used to obtain new access tokens.
+  final String? refreshToken;
+
+  /// The lifetime in seconds of the access token.
+  final int? expiresIn;
```
marcelomendoncasoares, line 10: Question: Should we expose it as a `Duration` instead? Integer parameters for durations are always prone to interpretation error if we can avoid them.

vfiruz97: My recommendation is to keep it as `int`. The `expiresIn` field matches the raw OAuth2 specification [RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749), where `expires_in` is always an integer representing seconds. If we were building a higher-level domain model that processes this data for application use, that would be the appropriate place to convert to Duration, but `OAuth2PkceTokenResponse` is intentionally a low-level response parser.
I'm good to convert it to Duration if you think that provides better DX

marcelomendoncasoares: If having as `int` is more compliant with the RFC, I'm ok with keeping it that way.

PR #4680, opened 2026-02-03
feat: Added PostgreSQL row-level locking functionality to Serverpod DB.

### LINKED ISSUE #4065: Add ability to lock a specific database row during transactions
### Problem to Solve

There is no built-in mechanism to prevent concurrent modifications of a single row, increasing risk for data races and potential overwrites in scenarios like financial transactions.

### Proposal

Introduce an API that enables developers to explicitly lock a specific row in the database during a transaction (such as with SQL's `SELECT ... FOR UPDATE`). This would ensure only one transaction can update the row until the lock is released, improving consistency and safety.

### Use Case

A developer wants to update a user's account balance in a financial application. Locking the row for the account during a transaction prevents double-spending or accidental overwrites.

### Alternatives

Application-level locks (e.g., Redis, in-memory structures) are less reliable than native row-level database locks. Optimistic concurrency control only partially mitigates the issue and does not fully prevent race conditions.

### Additional context

This feature would make Serverpod more suitable for transactional integrity in multi-user applications.

Some inspiration for this feature might be taken from the solution that ensures only a single migration is run at a time: https://github.com/serverpod/serverpod/blob/eb072797467f0c736de79a0e22a4bc430b1d5d5d/packages/serverpod/lib/src/database/migrations/migration_manager.dart#L221-L252

- [ ] I want to work on this

FXschwartz:
Adds PostgreSQL row-level locking support to prevent race conditions in concurrent transactions.

New features:                                                                                                                        
  - LockMode enum with PostgreSQL lock strengths: forUpdate, forNoKeyUpdate, forShare, forKeyShare
  - LockBehavior enum for lock wait behavior: wait (default), noWait, skipLocked
  - lockMode and lockBehavior parameters added to find() and findRow() methods
  - New lockRows() method for acquiring locks without fetching/deserializing data
  - Validation that throws ArgumentError when lock parameters are used without a transaction

Example usage:
```dart
await session.db.transaction((tx) async {
  await MyModel.db.lockRows(
    session,
    where: (t) => t.status.equals('pending'),
    lockMode: LockMode.forUpdate,
    lockBehavior: LockBehavior.skipLocked,
    transaction: tx,
  );
  // Locked rows are now safe to update
});
```
Fixes #4065 

## Pre-launch Checklist

## Breaking changes
None.

### DIFF
--- a/tests/serverpod_test_server/test_integration/database_operations/row_lock_test.dart
+++ b/tests/serverpod_test_server/test_integration/database_operations/row_lock_test.dart
@@ -22,240 +22,596 @@ void main() {
         );
       });
 
-      group('when finding rows with lock mode forUpdate', () {
-        test('then the query succeeds and returns the locked rows', () async {
-          await session.db.transaction((transaction) async {
-            final rows = await SimpleData.db.find(
-              session,
-              where: (t) => t.id.equals(insertedRow.id!),
-              lockMode: LockMode.forUpdate,
-              transaction: transaction,
-            );
+      test('when finding rows with lock mode forUpdate '
+          'then the query succeeds and returns the locked rows.', () async {
+        await session.db.transaction((transaction) async {
+          final rows = await SimpleData.db.find(
+            session,
+            where: (t) => t.id.equals(insertedRow.id!),
+            lockMode: LockMode.forUpdate,
+            transaction: transaction,
+          );
 
-            expect(rows.length, 1);
-            expect(rows.first.num, 42);
-          });
+          expect(rows.length, 1);
+          expect(rows.first.num, 42);
         });
       });
 
-      group('when finding rows with lock mode forShare', () {
-        test('then the query succeeds and returns the locked rows', () async {
-          await session.db.transaction((transaction) async {
-            final rows = await SimpleData.db.find(
-              session,
-              where: (t) => t.id.equals(insertedRow.id!),
-              lockMode: LockMode.forShare,
-              transaction: transaction,
-            );
+      test('when finding rows with lock mode forShare '
+          'then the query succeeds and returns the locked rows.', () async {
+        await session.db.transaction((transaction) async {
+          final rows = await SimpleData.db.find(
+            session,
+            where: (t) => t.id.equals(insertedRow.id!),
+            lockMode: LockMode.forShare,
+            transaction: transaction,
+          );
 
-            expect(rows.length, 1);
-            expect(rows.first.num, 42);
-          });
+          expect(rows.length, 1);
+          expect(rows.first.num, 42);
         });
       });
 
-      group('when finding a row by id with lock mode forUpdate', () {
-        test('then the query succeeds and returns the locked row', () async {
-          await session.db.transaction((transaction) async {
-            final row = await SimpleData.db.findById(
-              session,
-              insertedRow.id!,
-              lockMode: LockMode.forUpdate,
-              transaction: transaction,
-            );
+      test('when finding a row by id with lock mode forUpdate '
+          'then the query succeeds and returns the locked row.', () async {
+        await session.db.transaction((transaction) async {
+          final row = await SimpleData.db.findById(
+            session,
+            insertedRow.id!,
+            lockMode: LockMode.forUpdate,
+            transaction: transaction,
+          );
 
-            expect(row, isNotNull);
-            expect(row!.num, 42);
-          });
+          expect(row, isNotNull);
+          expect(row!.num, 42);
         });
       });
 
-      group('when finding first row with lock mode forUpdate', () {
-        test('then the query succeeds and returns the locked row', () async {
-          await session.db.transaction((transaction) async {
-            final row = await SimpleData.db.findFirstRow(
-              session,
-              where: (t) => t.num.equals(42),
-              lockMode: LockMode.forUpdate,
-              transaction: transaction,
-            );
+      test('when finding first row with lock mode forUpdate '
+          'then the query succeeds and returns the locked row.', () async {
+        await session.db.transaction((transaction) async {
+          final row = await SimpleData.db.findFirstRow(
+            session,
+            where: (t) => t.num.equals(42),
+            lockMode: LockMode.forUpdate,
+            transaction: transaction,
+          );
 
-            expect(row, isNotNull);
-            expect(row!.num, 42);
-          });
+          expect(row, isNotNull);
+          expect(row!.num, 42);
         });
       });
 
-      group('when locking rows without returning data', () {
-        test('then the lock is acquired successfully', () async {
-          await session.db.transaction((transaction) async {
-            await SimpleData.db.lockRows(
-              session,
-              where: (t) => t.id.equals(insertedRow.id!),
-              lockMode: LockMode.forUpdate,
-              transaction: transaction,
-            );
+      test('when locking rows without returning data '
+          'then the query succeeds and the row is still accessible.', () async {
+        await session.db.transaction((transaction) async {
+          await SimpleData.db.lockRows(
+            session,
+            where: (t) => t.id.equals(insertedRow.id!),
+            lockMode: LockMode.forUpdate,
+            transaction: transaction,
+          );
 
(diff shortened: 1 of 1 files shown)


marcelomendoncasoares:
Hi @FXschwartz! How is it going with this one? I'm just awaiting the CI turn green to review, but feel free to ping me if you need any guidance or help here!

FXschwartz:
> Hi @FXschwartz! How is it going with this one? I'm just awaiting the CI turn green to review, but feel free to ping me if you need any guidance or help here!

Ah sorry I've gotten these tests failing on previous PR's and assumed they were flaky, my bad.

I'll rebase and take a look at them today!

FXschwartz:
@marcelomendoncasoares Ok rebased branch, added missing generated files which broke CI tests. All tests pass and it's ready for a review!

packages/serverpod/lib/src/database/database.dart
marcelomendoncasoares, line 65: Suggestion: Add the documentation for the new lock behavior (for all methods that received the parameter).

packages/serverpod/test/database/sql_query_builder_test.dart
```diff
@@ -1025,4 +1025,163 @@ void main() {
...
+      var query = SelectQueryBuilder(
+        table: table,
+      ).withLockMode(LockMode.forShare, LockBehavior.skipLocked).build();
+
+      expect(
+        query,
+        'SELECT "citizen"."id" AS "citizen.id" FROM "citizen" FOR SHARE SKIP LOCKED',
+      );
+    });
+  });
```
marcelomendoncasoares: Suggestion: Improve the description to inform the behavior being tested (not the specific values of the test). For example: `when specifying both the mode and behavior then query is correct`.

packages/serverpod/test/database/sql_query_builder_test.dart
```diff
@@ -1025,4 +1025,163 @@ void main() {
...
     },
   );
+
+  group('Given SelectQueryBuilder with lock mode', () {
+    var table = Table<int?>(tableName: 'citizen');
+
+    test('when LockMode.forUpdate then query contains FOR UPDATE', () {
+      var query = SelectQueryBuilder(
+        table: table,
+      ).withLockMode(LockMode.forUpdate).build();
```
marcelomendoncasoares: Suggestion: We can improve a bit on the descriptions to make the test cases clearer. Something like:

- Given a SelectQueryBuilder object
  _(create the SelectQueryBuilder in the group to match the description)_
  - when building with lock mode `LockMode.forUpdate` then query contains FOR UPDATE.
    _(call the `withLockMode` and `build` on the test to match the description)_

tests/serverpod_test_server/test_integration/database_operations/row_lock_test.dart
```diff
@@ -0,0 +1,218 @@
...
+            );
+
+            // Use lock within transaction
+            await session.db.transaction((transaction) async {
+              final rows = await SimpleData.db.find(
+                session,
+                where: (t) => t.id.equals(inserted.id!),
+                lockMode: LockMode.forUpdate,
+                transaction: transaction,
+              );
```
marcelomendoncasoares: Suggestion: We can improve a bit on the organization of these tests to make them more adherent to the pattern. The sequence of the `Given`/`when`/`then` should follow a valid test description (avoiding intermediary "scope" groups like the `find with lockMode`).

Since the tests are having different setups, this means that we should separate them in different groups altogether, with their `Given` description matching the scenario. For example, this test case has test data that should be in the setup and not inside the test.

For each of these methods, we should also cover the attempt to run a concurrent transaction to ensure that the feature is behaving as expected. Below a suggestion for the reorganization and coverage improvement:

- Given a table with existing data
  - when finding rows with row lock for update
    - then the query succeeds.
- Given a table with existing data and a lock for update in place
  - when finding matching rows with no lock then the operation completes normally without awaiting.
  - when finding matching rows with no wait then the operation throws due to rows being locked.
  - when finding matching rows with skip locked then only not matching rows are returned.

tests/serverpod_test_server/test_integration/database_operations/row_lock_test.dart
```diff
@@ -0,0 +1,218 @@
...
+          expect(
+            () => SimpleData.db.find(
+              session,
+              where: (t) => Constant.bool(true),
+              lockMode: LockMode.forUpdate,
+            ),
+            throwsA(isA<ArgumentError>()),
+          );
+        },
+      );
```
marcelomendoncasoares: Suggestion: We should have this test for the other methods.

tests/serverpod_test_server/test_integration/database_operations/row_lock_test.dart
```diff
@@ -0,0 +1,218 @@
...
+              expect(rows.length, 2);
+            });
+          },
+        );
+      });
+    },
+  );
+
+  withServerpod(
+    'Given row locking without rollback',
```
marcelomendoncasoares: Question: Why the "without rollback"? If it's because we need to disable rollback on the test framework, there is no need for it to be in the description, since it is a particularity of the test wrapper and not part of the test scenario. A comment can be useful to explain the reason, though.

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
```diff
@@ -380,6 +403,8 @@ class BuildRepositoryClass {
                 'transaction': refer('transaction'),
                 if (objectRelationFields.isNotEmpty)
                   'include': refer('include'),
+                'lockMode': refer('lockMode'),
+                'lockBehavior': refer('lockBehavior'),
```
marcelomendoncasoares, line 407: Suggestion: We also need to test the changes to the code generator (new methods and parameters). These tests live in the `tools/serverpod_cli/test/generator/dart/server_code_generator/entity_class_test.dart` file.

packages/serverpod/lib/src/database/database.dart
FXschwartz: Done!

packages/serverpod/test/database/sql_query_builder_test.dart
FXschwartz: Updated!

packages/serverpod/test/database/sql_query_builder_test.dart
FXschwartz: Updated!

tests/serverpod_test_server/test_integration/database_operations/row_lock_test.dart
FXschwartz: Done!

tests/serverpod_test_server/test_integration/database_operations/row_lock_test.dart
FXschwartz: Added!

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
FXschwartz: Added!

tests/serverpod_test_server/test_integration/database_operations/row_lock_test.dart
FXschwartz: Yeah it's needed to be disabled on the test framework but you are right the description isn't needed. Removed it from the description and added a comment!

marcelomendoncasoares (approved):
Another excellent work with this one, @FXschwartz!! I have made just a minor improvement to the lock mode and behavior coverage/format.

PR approved! 🚀 

I'll just ask you to hold the merge for now because we'll release a patch version tomorrow with two minor fixes. While this, would you prepare the docs for the lock methods?

FXschwartz:
> I'll just ask you to hold the merge for now because we'll release a patch version tomorrow with two minor fixes. While this, would you prepare the docs for the lock methods?

Sounds great! Documentation PR created.

PR #4689, opened 2026-02-04
feat: Adds support to Facebook Identity Provider

### LINKED ISSUE #147: Sign in with Facebook
Add support for Sign in with Facebook for all platforms:

- [ ] iOS
- [ ] Android
- [ ] Web
- [ ] Mac
- [ ] Windows

vfiruz97:
This PR adds Facebook OAuth authentication support to the `serverpod_auth` module. It implements a complete Facebook identity provider (IDP) with access token validation for enhanced security across all platforms.

**Key features:**

- Facebook access token validation and authentication flow
- Cross-platform support (iOS, Android, Web, macOS)
- Customizable sign-in button widget with Facebook branding
- Complete server-side token verification, user authentication, and account management
- Integration with existing Serverpod auth system
- Comprehensive test coverage

**Dependencies added:**

This PR adds one dependency to `serverpod_auth_idp_flutter`:

- `flutter_facebook_auth: ^7.1.5` - Popular Flutter package for Facebook authentication. Well-maintained with cross-platform support. This is the standard Flutter package for Facebook OAuth implementation that handles the native platform SDKs.

**Note:** On iOS, Facebook may issue limited access tokens when App Tracking Transparency (ATT) permission is not granted. These limited tokens cannot be validated by the server or used to retrieve user profile data, which will cause authentication to fail. Documentation will include a dedicated section on requesting ATT permissions to ensure full Facebook authentication functionality on iOS.

**Demo:**

A short video demonstration of Facebook Sign-In working across different platforms is available here: [e2e test video](https://www.youtube.com/watch?v=-hmANE_RxRo)

Fixes #147.

## Pre-launch Checklist

## Breaking changes

None. This is a new feature that adds Facebook authentication support without modifying existing authentication methods or APIs.

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/facebook/business/facebook_idp_config.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/facebook/business/facebook_idp_config.dart
@@ -114,11 +114,13 @@ class FacebookIdpConfig extends IdentityProviderBuilder<FacebookIdp> {
 /// ```
 class FacebookIdpConfigFromPasswords extends FacebookIdpConfig {
   /// Creates a new [FacebookIdpConfigFromPasswords] instance.
-  FacebookIdpConfigFromPasswords()
-    : super(
-        appId: Serverpod.instance.getPasswordOrThrow('facebookAppId'),
-        appSecret: Serverpod.instance.getPasswordOrThrow('facebookAppSecret'),
-      );
+  FacebookIdpConfigFromPasswords({
+    super.facebookAccountDetailsValidation,
+    super.getExtraFacebookInfoCallback,
+  }) : super(
+         appId: Serverpod.instance.getPasswordOrThrow('facebookAppId'),
+         appSecret: Serverpod.instance.getPasswordOrThrow('facebookAppSecret'),
+       );
 }
 
 /// Exception thrown when Facebook user information is missing required data.


vfiruz97:
Hi @marcelomendoncasoares,
This PR is ready for review. I followed the same pattern as other IDPs and rebased it on upstream/main to ensure smooth merging. I will proceed with the docs next.

vfiruz97:
Hi @marcelomendoncasoares,

Just checking in on this one! I know you're super busy, but whenever you have a moment, could you take a look?

Since it's closely aligned with our previous work on other IDPs, I'm hoping it'll be a fairly straightforward review. Thanks again for your time!

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/pubspec.yaml
```diff
@@ -16,6 +16,7 @@ dependencies:
   email_validator: '>=2.1.17 <4.0.0'
   flutter:
     sdk: flutter
+  flutter_facebook_auth: ^7.1.5
```
marcelomendoncasoares: Discussion: Since this package pushes a lot of platform-specific code that can break users apps, it is better for us to spin it to its own package - such as with `serverpod_auth_firebase_flutter`. It it also important to avoid bloating users apps for nothing when this dependency is not used.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter_facebook/lib/src/facebook_sign_in_button.dart
```diff
@@ -0,0 +1,181 @@
...
+              width: buttonStyle.size.height * 0.5 + 8,
+            ),
+            Expanded(
+              child: Center(child: text),
+            ),
+          ],
+        ),
+      ],
+    );
+  }
```
marcelomendoncasoares, line 169: Suggestion: Looking on the Stories app, the logo feels a bit off on the left position (it should have the same spacing on the left as the other widgets). Look the comparison with Apple below:

- Image
- Image

vfiruz97: Yeah, good catch! I will fix this.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/facebook/business/facebook_idp_utils.dart
```diff
@@ -0,0 +1,319 @@
...
+    return await FacebookAccount.db.insertRow(
+      session,
+      facebookAccount,
+      transaction: transaction,
+    );
+  }
+}
+
+extension on Session {
+  Never logAndThrow(final String message) {
```
marcelomendoncasoares, line 328: Suggestion: One detail that I forgot on the revision is to add the `getAccount` and `hasAccount` utilities (with tests) that were introduced after you started working on this PR. It will also require making the endpoint extend the `EndpointIdpBase`.

vfiruz97: Yes, noticed too. Will fix that. Thanks for remaining.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/pubspec.yaml
vfiruz97: Thanks @marcelomendoncasoares for the thoughtful feedback!

I agree that `flutter_facebook_auth` pulls in the native Facebook SDK and is relatively heavy, so your point about avoiding unnecessary platform-specific dependencies in the main package makes sense.

At the same time, Apple and Google are currently not decoupled. I'm not fully sure whether they are considered lightweight, or common enough in usage, to justify keeping them in the main package.

From a technical perspective, I'm fine with decoupling Facebook if that's the preferred direction. The approach could be:

- Keep `serverpod_auth_idp_server` unchanged.
- Create a new package, e.g. `serverpod_auth_idp_flutter_facebook`.
- Move the 5 Facebook files from `serverpod_auth_idp_flutter/lib/src/facebook/` into that package.
- Remove the Facebook dependency from `serverpod_auth_idp_flutter`.

One nuance I'm thinking about is the UI layer. If we decouple it, we lose direct Facebook integration in `SignInWidget`, because that would again require adding it as a dependency. Maybe we can keep the Facebook button widgets integrated only in the `storybook` app by adding the new Facebook package as a dependency there. Since `storybook` is just for showcasing and testing widgets, it won't affect end users. What do you think?

If decoupling is the intended architectural direction, I'm happy to proceed. Just let me know once again your final thought.

marcelomendoncasoares: You guessed correctly: Google and Apple are very common and rather lightweight, so it is not an issue having them there. They are also very stable and permissive, allowing the dependency to exist without requiring configuration. Others, like Firebase (and I guess Facebook also), break the app if added as dependency and not configured.

But we need indeed to solve the extensibility of the `SingInWidget` to make this refactor. Decoupling is required, but the user must have the same UX of using a single `SignInWidget` to show all providers. Given the introduction of the `IdpBaseEndpoint`, we can explore an alternative of using this and a global registry to render external IDPs.

1. On the `SignInWidget`, we would get all endpoints of type `IdpBaseEndpoint` that are available on the `client`.
2. Then, it would search on a global `Map<IdpBaseEndpoint, WidgetBuilderFunction>` on how to render.
3. If found, the provider would be rendered after the known IDPs.

External packages (such as the Facebook-specific) can register themselves on this global map when the `initialize*` method is called. We wouldn't need to think of a disable mechanism since it would be easily achievable by not initializing the provider or removing the dependency completely.

Do you think something like this would work?

vfiruz97: Great suggestion! I think this approach works! Right now, `SignInWidget` is hardcoded to show specific providers. It fixes the decoupled providers plugin issue without having to include them as dependencies.

The logic is simple: all providers (both internal and external) register their builder in a global registry. The `SignInWidget` first identifies the available endpoints via `IdpBaseEndpoint`, and then searches the registry for the corresponding builder to show the UI. This way, we can filter endpoints dynamically without any hardcoding. Truth: even then we will hardcode them to order.

**Implementation Example:**
In `serverpod_auth_idp_flutter/src/common/`, we add the registry:

```dart
typedef IdpWidgetBuilder = Widget Function(BuildContext context, Client client);

class IdpWidgetRegistry {
  static final _registry = <Type, IdpWidgetBuilder>{};
  
  static void register<T extends IdpBaseEndpoint>(IdpWidgetBuilder builder) {
    _registry[T] = builder;
  }
  
  static IdpWidgetBuilder? getBuilder(Type type) => _registry[type];
}

```
Note: Decoupled providers will have the `serverpod_auth_idp_flutter` package in the deps.

Then, the decoupled providers just need to register themselves during initialization, before `SignInWidget` renders, mostly in the `main` function.

**Proposed Plan:**
Since this change will require significant updates to both the serverpod and the documentation, I suggest we merge the **Facebook** (decoupled) and **Microsoft** PRs first. I will then follow up with a fresh PR to implement this registry refactor across everything. This keeps the current work moving while we prepare for the shift.  What do you think?

marcelomendoncasoares: I wouldn't change to register the internal IDPs in the global registry, but only keep this as a feature targeted explicitly at external IDPs. Besides being simpler to register the internal IDPs directly, it is also required for us to control the order when needed (as with Google and Apple). And it would be a breaking change if we change the disable mechanism for the existing ones - or we would step outside of the pattern to disable them by boolean flags.

Keeping the feature targeted only to the external packages allows us to make the change on this PR without propagating to other providers or PRs - such as the Microsoft PR.

vfiruz97: That makes total sense. Keeping the internal IDPs as they are avoids breaking changes to the disable mechanism and preserves our control over the display order (Google/Apple). I will focus the global registry strictly on external/decoupled providers.

Plan:

- Implement IdpWidgetRegistry: I'll add the registry in `serverpod_auth_idp_flutter` for external use.
- Update `SignInWidget`: It will still handle internal providers directly, but will now also check the global registry for any externally registered builders.
 - Decouple Facebook: I will implement the register itself in the "initialize" method, allowing it to work without being a direct dependency of the core widget.

I'll update this PR with these changes now. I agree this is much cleaner and keeps the Microsoft PR and others unaffected.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/facebook/business/facebook_idp_utils.dart
vfiruz97: Thanks! Fixed in b7cb840.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter_facebook/lib/src/facebook_sign_in_button.dart
vfiruz97: Thanks! Fixed in b7cb840.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/pubspec.yaml
vfiruz97: Thanks! Fixed in 899f1e2.

vfiruz97:
Hi @marcelomendoncasoares, the comments are addressed.
There were some conflicts, which I resolved. As a result, I had to force-push the branch.
It’s now up to date with `upstream/main`. I also removed and recreated the Facebook migration to ensure it appears at the bottom in the correct order.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/common/external_idp_registry.dart
```diff
@@ -0,0 +1,66 @@
...
+  /// Throws [StateError] if type [T] is already registered.
+  void register<T extends EndpointIdpBase>(ExternalIdpWidgetBuilder builder) {
+    if (_registry.containsKey(T)) {
+      throw StateError(
+        'External IDP widget builder for type $T is already registered. '
+        'Each provider can only be registered once.',
+      );
+    }
+    _registry[T] = builder;
+  }
```
marcelomendoncasoares, line 36: Suggestion: We should have a unit test for the `ExternalIdpRegistry` methods and behavior.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/sign_in_widget.dart
marcelomendoncasoares, line 79: Suggestion: Given that the `hasFacebook` is still possible due to it depending only on the endpoint class (and not on the external package), I think we can bring back the disable bool for such provider. Although I have proposed previously that disabling is easy by just not-initializing the provider, this would be a shortcut in case we couldn't easily disable on the widget. But it actually creates a complication for developers that want to temporarily disable the provider without completely removing the dependencies - because omitting the initialize is not possible without removing the underlying Facebook SDK library.

Then, we should also return the disable story for the `SignInWidget` on the storybook app.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/common/external_idp_registry.dart
vfiruz97: Of course! Will fix in the next commit.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/sign_in_widget.dart
vfiruz97: Yes, I will fix this in the next commit.

vfiruz97: Thankns! Fixed in [a633f41](https://github.com/serverpod/serverpod/pull/4689/commits/a633f41d2f738fbe363c01cfeaa091c4e4625f99).

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/common/external_idp_registry.dart
vfiruz97: Thankns! Fixed in [a633f41](https://github.com/serverpod/serverpod/pull/4689/commits/a633f41d2f738fbe363c01cfeaa091c4e4625f99).

vfiruz97:
@marcelomendoncasoares, as always great suggestions! The comments are addressed. This is ready for another review.

marcelomendoncasoares (approved):
Great work with this provider, @vfiruz97! LGTM! 🚀

vfiruz97:
I'm addressing the issue discussed here: https://github.com/serverpod/serverpod_docs/pull/418#discussion_r2813272423
I made a small fix to `FirebaseIdpConfigFromPasswords` to accept additional optional parameters.

Since this PR is still open, I included the fix here. The remaining fixes will come in a separate PR from main.

Next, I'll review the Facebook docs according to the new guide style and request a review.

PR #4701, opened 2026-02-06
feat: Adds sending of pings in streaming connections.

### LINKED ISSUE #4684: Streaming methods quietly failing on MacOS when computer goes to sleep
Streaming methods can fail without an exception being thrown on MacOS. It seems to consistently happen when the computer goes to sleep for some time (probably longer than the timeout for the stream), very quick sleeps does keep the connection open.

The stream will seem to remain open, but no data is transmitted. Expected behavior would be to get an exception when the stream is no longer responsive or disconnected.

After some research with Cursor:

> The most likely explanation is “unclean” or half‑open WebSocket/TCP connections after sleep/hibernation. That’s common: the OS suspends networking, NATs/routers drop idle connections, but the local socket stays open and no FIN/RST is delivered to the app. In that case, no exception is raised until you try to write or a heartbeat detects it.

A proposed solution is to set a timeout for when the client received data. The server already sends ping/pong messages to the client, so no messages have been received for a while (1 minute?), then we should consider the connection to be failed and throw an exception.

vlidholt:
Adds keepalive pings and idle-timeout detection for method streams sent from the client. It surfaces a specific `MethodStreamIdleTimeoutException` when no inbound messages are received within the timeout.

Pings are sent every 20 seconds. If no message has been received in 40 seconds, the connection is considered idle. The timeouts are currently not configurable (we may want to do that in a future PR).

Fixes #4684

## Pre-launch Checklist

## Breaking changes

No breaking changes.

### DIFF
--- a/packages/serverpod_client/lib/src/client_method_stream_manager.dart
+++ b/packages/serverpod_client/lib/src/client_method_stream_manager.dart
@@ -603,7 +603,6 @@ final class ClientMethodStreamManager {
 
     _pingTimer = Timer.periodic(_pingInterval, (_) {
       if (_webSocket == null) return;
-      print('Sending ping');
       _addMessageToWebSocket(PingCommand.buildMessage());
     });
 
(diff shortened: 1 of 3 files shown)


marcelomendoncasoares (approved):
LGTM! 🚀

PR #4702, opened 2026-02-06
feat: Replace system_resources with system_resources_2

thiagoalmeidasa:
## Description

Upgrade from `system_resources` v1 (FFI-based) to `system_resources_2` which uses a pure Dart cgroup implementation on Linux with automatic cgroup v1/v2 detection. This enables Insights CPU and memory monitoring in containerized environments without precompiled native libraries.

- Add cgroup v1/v2 and gVisor support for resource monitoring
- Resolve glibc mismatch and .so bundling issues in containers
- Improve HealthCheckManager error logging on init failure
- Simplify startup_test.dart (init is now idempotent)

## Pre-launch Checklist

## Breaking changes

No breaking changes. The `SystemResources` API (`init()`, `cpuLoadAvg()`, `memUsage()`) is fully compatible between the old and new packages.

### DIFF
--- a/packages/serverpod/pubspec.yaml
+++ b/packages/serverpod/pubspec.yaml
@@ -34,7 +34,7 @@ dependencies:
   serverpod_serialization: 3.2.3
   serverpod_shared: 3.2.3
   synchronized: ^3.1.0
-  system_resources_2: ^2.2.1
+  system_resources_2: ^2.2.2
   uuid: ^4.1.0
   vm_service: '>=13.0.0 <16.0.0'
   web_socket: ^1.0.1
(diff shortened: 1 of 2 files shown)


thiagoalmeidasa:

packages/serverpod/lib/src/server/health_check_manager.dart
```diff
@@ -67,10 +67,12 @@ class HealthCheckManager {
 
     try {
       await SystemResources.init();
-    } catch (e) {
+    } catch (e, stackTrace) {
       stderr.writeln(
         'WARNING: CPU and memory usage metrics are not supported on this platform.',
       );
+      stderr.writeln('Error: $e');
+      stderr.writeln('Stack trace: $stackTrace');
```
marcelomendoncasoares, line 75: Question: What happens now when we start the service on Windows? Will it print the stacktrace or is it now supported? If it's still not supported, I think the best would be not to push the stacktrace on the user. Otherwise, users might open issues thinking this is a bug and not an expected behavior.

marcelomendoncasoares: This concern is actually not needed, since we abort the health checks on Windows right above.

PR #4711, opened 2026-02-10
feat(auth): Implement localization for sign in widget

### LINKED ISSUE #4496: SignInWidget internationalization
### Problem to Solve

SignInWidget only shows English texts.

Image

### Proposal

Add attributes to customize the labels for the whole authentication module.

### Use Case

Yes, the app can be used with different languages. 

Image

### Alternatives

No

### Additional context

_No response_

- [ ] I want to work on this

justlunix:
Implements localization for the sign in widgets for all current providers except firebase.

Uses an `InheritedWidget` to pass down localization texts with the current english texts as default.

Fixes #4496.

## Pre-launch Checklist

## Breaking changes

None. The widgets still work as they simply use the default texts if none are provided otherwise.

## Summary by CodeRabbit

* **New Features**
  * Introduced localization system for authentication UI, enabling customization of sign-in text across email, Apple, Google, GitHub, and anonymous authentication methods. Developers can now provide localized text strings for all authentication screens, including login, registration, password reset, and verification flows.

### DIFF
--- a/examples/auth/auth_flutter/lib/main.dart
+++ b/examples/auth/auth_flutter/lib/main.dart
@@ -107,7 +107,7 @@ class SignInScreen extends StatelessWidget {
       // See the `firebase.dart` file for more information.
       // body: FirebaseSignInScreen(client: client),
       body: SignInLocalizationProvider(
-        basic: BasicSignInTexts(
+        basic: const BasicSignInTexts(
           noAuthenticationProvidersConfigured: 'No auth providers configured',
           orContinueWith: 'or continue with',
         ),
@@ -129,12 +129,12 @@ class SignInScreen extends StatelessWidget {
           containsNumber: 'Contains at least one number',
           containsSpecialCharacter: 'Contains at least one special character',
         ),
-        apple: AppleSignInTexts(signInButton: 'Use Apple account'),
-        google: GoogleSignInTexts(signInButton: 'Use Google account'),
-        github: GitHubSignInTexts(signInButton: 'Use GitHub account'),
-        microsoft: MicrosoftSignInTexts(signInButton: 'Use Microsoft account'),
-        facebook: FacebookSignInTexts(signInButton: 'Use Facebook account'),
-        anonymous: AnonymousSignInTexts(signInButton: 'Continue as guest'),
+        apple: const AppleSignInTexts(signInButton: 'Apple'),
+        google: const GoogleSignInTexts(signInButton: 'Google'),
+        github: const GitHubSignInTexts(signInButton: 'GitHub'),
+        microsoft: const MicrosoftSignInTexts(signInButton: 'Microsoft'),
+        facebook: const FacebookSignInTexts(signInButton: 'Facebook'),
+        anonymous: const AnonymousSignInTexts(signInButton: 'Sign In as Guest'),
         child: SignInWidget(
           client: client,
           // NOTE: No need to call navigation here if it gets done on the


justlunix:

justlunix:
recheck

marcelomendoncasoares:
Thanks for your contribution with this issue, @justlunix! It will be a highly valuable addition to the framework! I'll first ask our AI agent for a quick review and then I'll review the PR manually.

@coderabbitai Please full review.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/email/forms/verification.dart
marcelomendoncasoares, line 45: Suggestion: Fix this hardcoded "Verify" on the doc comments.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/localization/sign_in_localization_provider.dart
```diff
@@ -0,0 +1,641 @@
...
+  /// Returns the nearest localization provider in the widget tree.
+  ///
+  /// Throws an assertion error in debug mode if no provider is found.
+  static SignInLocalizationProvider of(BuildContext context) {
+    final localizations = maybeOf(context);
+    assert(
+      localizations != null,
+      'No SignInLocalizationProvider found in context',
+    );
+    return localizations!;
```
marcelomendoncasoares: Suggestion: Replace the assert by a proper exception to avoid null check errors in release mode.


```suggestion
    if (localizations == null) {
      throw StateError('No SignInLocalizationProvider found in context');
    }
    return localizations;
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/localization/sign_in_localization_provider.dart
marcelomendoncasoares, line 1: Suggestion: Split this file to declare each localization class and extension getter in the provider folder.

justlunix: I've kept the naming of the getters to be specific, I don't know if there would be a time where a build context may get both extensions (maybe custom widgets) so I think it makes more sense like that.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/apple/apple_sign_in_texts.dart
```diff
@@ -0,0 +1,71 @@
...
+    required this.signIn,
+  });
+
+  /// Default English texts.
+  static const defaults = AppleSignInTexts(
+    signInWith: 'Sign in with Apple',
+    continueWith: 'Continue with Apple',
+    signUpWith: 'Sign up with Apple',
+    signIn: 'Sign in',
+  );
```
marcelomendoncasoares: Suggestion: Looking at the files now split, I just realized that we don't actually need to override for all possible options of single-text buttons such as Apple, Anonymous, Google, etc. For these, we could have one option only that is null by default and will be used if set. Otherwise, the provider will fallback to the configured value from the enum. This will simplify for users creating a translation, since they should only worry about the single text they want to use - instead of translating a map of options that they will never use.

justlunix: Hi @marcelomendoncasoares, sorry for the late reply! I think I don't quite understand what exactly you mean with the suggestion, could you elaborate further?

marcelomendoncasoares: Sure! Sorry for not being clear. My suggestion is to reduce the text options for the providers that consist only on a single button to something like:

```dart
class AppleSignInTexts {
  /// Creates a new [AppleSignInTexts] configuration.
  const AppleSignInTexts({
    required this.signInButton,
  });

  /// Default English texts.
  static const defaults = AppleSignInTexts(
    signInButton: null,
  );
  
  /// Text for the Sign-In with Apple button.
  final String? signInButton;  // <-- Change to a single option instead of all from the enum
```

Then, on the sign-in widget, instead of mapping all possible entries, override the text directly:

```dart
  Widget build(BuildContext context) {
    final texts = context.appleSignInTexts;
    final buttonStyle = AppleSignInStyle.fromConfiguration(
      shape: shape,
      size: size,
      width: minimumWidth,
    );

    return ConstrainedBox(
      constraints: BoxConstraints(
        minWidth: minimumWidth,
        maxWidth: 400,
        minHeight: buttonStyle.size.height,
        maxHeight: buttonStyle.size.height,
      ),
      child: SignInWithAppleButton(
        onPressed: isLoading || isDisabled ? null : onPressed ?? () {},
        text: texts.signInButton ?? _getButtonText(), // <-- Here only use the enum if the option is missing
        height: buttonStyle.size.height,
        style: style,
        borderRadius: buttonStyle.borderRadius,
        iconAlignment: logoAlignment,
      ),
    );
  }
```

This will greatly simplify for users translating/replacing the texts on their app since they will only use one option on the button anyways. This would be true for Anonymous, Apple, Google, GitHub, Microsoft, Facebook, which are all single-button providers.

justlunix: @marcelomendoncasoares I see, thanks for the quick clarification! You're right, that's much simpler. Will do!

PR #4712, opened 2026-02-10
feat: Add legacy client support for email auth migrations

exaby73:
## Pre-launch Checklist

## Breaking changes

None

## Summary by CodeRabbit

* **New Features**
  * Backward-compatible legacy client support with endpoint name mapping.
  * Legacy session token manager and builder for legacy sessions.
  * Legacy user profile operations (image, username, full name).
  * Legacy email, Google, Firebase, and admin authentication endpoints (stubs where unsupported).
  * Legacy session status, sign-out, and session revocation endpoints.
  * Public legacy data types and protocol exports.

* **Tests**
  * Comprehensive integration tests covering legacy client flows and migrations.

### DIFF
--- a/tests/serverpod_auth_test/serverpod_auth_test_server/test/integration/legacy_client_support_test.dart
+++ b/tests/serverpod_auth_test/serverpod_auth_test_server/test/integration/legacy_client_support_test.dart
@@ -189,9 +189,7 @@ void main() {
               );
           authKeyProvider.setToken('${authResult.keyId}:${authResult.key}');
 
-          final isValid = await legacy_auth.Caller(
-            legacyClient,
-          ).status.isSignedIn();
+          final isValid = await legacyClient.modules.auth.status.isSignedIn();
 
           expect(isValid, isTrue);
         },
@@ -212,9 +210,7 @@ void main() {
               );
           authKeyProvider.setToken('${authResult.keyId}:${authResult.key}');
 
-          final userInfo = await legacy_auth.Caller(
-            legacyClient,
-          ).status.getUserInfo();
+          final userInfo = await legacyClient.modules.auth.status.getUserInfo();
 
           expect(userInfo?.userIdentifier, equals(authUserId.toString()));
         },
@@ -235,12 +231,11 @@ void main() {
               );
           authKeyProvider.setToken('${authResult.keyId}:${authResult.key}');
 
-          final changed = await legacy_auth.Caller(
-            legacyClient,
-          ).user.changeFullName('Legacy Full Name');
-          final updatedUserInfo = await legacy_auth.Caller(
-            legacyClient,
-          ).status.getUserInfo();
+          final changed = await legacyClient.modules.auth.user.changeFullName(
+            'Legacy Full Name',
+          );
+          final updatedUserInfo = await legacyClient.modules.auth.status
+              .getUserInfo();
 
           expect(changed, isTrue);
           expect(updatedUserInfo?.fullName, equals('Legacy Full Name'));
@@ -264,9 +259,7 @@ void main() {
 
           await legacy_auth.Caller(legacyClient).status.signOutDevice();
 
-          final isValid = await legacy_auth.Caller(
-            legacyClient,
-          ).status.isSignedIn();
+          final isValid = await legacyClient.modules.auth.status.isSignedIn();
 
           expect(isValid, isFalse);
         },
@@ -469,9 +462,10 @@ void main() {
         'when calling changePassword then returns false.',
         () async {
           final legacyClient = _createLegacyClient(sessionBuilder);
-          final result = await legacy_auth.Caller(
-            legacyClient,
-          ).email.changePassword('old', 'new');
+          final result = await legacyClient.modules.auth.email.changePassword(
+            'old',
+            'new',
+          );
 
           expect(result, isFalse);
         },
@@ -516,9 +510,9 @@ void main() {
         'when calling Firebase authenticate then it is handled by non-legacy endpoint.',
         () async {
           final legacyClient = _createLegacyClient(sessionBuilder);
-          final result = await legacy_auth.Caller(
-            legacyClient,
-          ).firebase.authenticate('fake-token');
+          final result = await legacyClient.modules.auth.firebase.authenticate(
+            'fake-token',
+          );
 
           expect(result.success, isFalse);
           expect(
@@ -658,9 +652,11 @@ void main() {
           expect(blockedInfo!.blocked, isTrue);
 
           final targetLegacyClient = _createLegacyClient(sessionBuilder);
-          final authResult = await legacy_auth.Caller(
-            targetLegacyClient,
-          ).email.authenticate(targetEmail, password);
+          final authResult = await targetLegacyClient.modules.auth.email
+              .authenticate(
+                targetEmail,
+                password,
+              );
 
           expect(authResult.success, isFalse);
           expect(
@@ -709,20 +705,34 @@ _LegacyClient _createLegacyClient(
   return _LegacyClient(baseUrl)..authKeyProvider = authKeyProvider;
 }
 
-class _LegacyClient extends legacy_auth.ServerpodClientShared {
+class _LegacyClient extends ServerpodClientShared {
   _LegacyClient(final String host)
     : super(
         host,
         legacy_auth.Protocol(),
         streamingConnectionTimeout: null,
         connectionTimeout: null,
-      );
+      ) {
+    modules = _LegacyModules(this);
+  }
+
+  late final _LegacyModules modules;
 
   @override
-  Map<String, legacy_auth.EndpointRef> get endpointRefLookup => const {};
+  Map<String, EndpointRef> get endpointRefLookup => const {};
 
   @override
-  Map<String, legacy_auth.ModuleEndpointCaller> get moduleLookup => const {};
+  Map<String, ModuleEndpointCaller> get moduleLookup => {
+    'auth': modules.auth,
+  };
+}
+
+class _LegacyModules {
+  _LegacyModules(final _LegacyClient client) {
+    auth = legacy_auth.Caller(client);
+  }
+
+  late final legacy_auth.Caller auth;
 }
 
 Client _createClient(


exaby73:
@coderabbitai full review

tests/serverpod_auth_test/serverpod_auth_test_server/test/integration/legacy_client_support_test.dart
```diff
@@ -0,0 +1,515 @@
...
+}
+
+Client _createClient(
+  final TestSessionBuilder sessionBuilder, {
+  final ClientAuthKeyProvider? authKeyProvider,
+}) {
+  final session = sessionBuilder.build();
+  final baseUrl = 'http://localhost:${session.server.port}/';
+  return Client(baseUrl)..authKeyProvider = authKeyProvider;
+}
```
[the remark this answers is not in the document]
exaby73: Sessions are closed by the test harness. I don't want to break that contract here

tests/serverpod_auth_test/serverpod_auth_test_server/test/integration/legacy_client_support_test.dart
```diff
@@ -0,0 +1,515 @@
...
+Future<void> _cleanUpDatabase(final Session session) async {
+  await AuthUser.db.deleteWhere(
+    session,
+    where: (final _) => Constant.bool(true),
+  );
+
+  await RateLimitedRequestAttempt.db.deleteWhere(
+    session,
+    where: (final t) => t.domain.equals('email'),
+  );
```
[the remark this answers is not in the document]
exaby73: Due to a `ON DELETE CASCADE`, it will clean up the sessions

exaby73:
@coderabbitai review

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/business/legacy_session_token_manager.dart
```diff
@@ -0,0 +1,82 @@
...
+  @override
+  Future<List<TokenInfo>> listTokens(
+    final Session session, {
+    required final UuidValue? authUserId,
+    final String? method,
+    final String? tokenIssuer,
+    final Transaction? transaction,
+  }) async {
+    return [];
+  }
```
marcelomendoncasoares: Question: Why not return the list of `LegacySession` for compatibility of the API?

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/business/legacy_session_token_manager_builder.dart
```diff
@@ -0,0 +1,12 @@
...
+
+/// Builder that creates [LegacySessionTokenManager] instances for registration
+/// with [AuthServices].
+class LegacySessionTokenManagerBuilder
+    implements TokenManagerBuilder<LegacySessionTokenManager> {
+  @override
+  LegacySessionTokenManager build({required final AuthUsers authUsers}) {
+    return LegacySessionTokenManager();
+  }
+}
```
marcelomendoncasoares: Suggestion: Move the implementation of the `TokenManagerBuilder` to the `LegacySessionTokenManager` class to reduce the hierarchy.

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/endpoints/legacy_firebase_endpoint.dart
```diff
@@ -0,0 +1,17 @@
...
+  Future<LegacyAuthenticationResponse> authenticate(
+    final Session session,
+    final String idToken,
+  ) async {
+    return LegacyAuthenticationResponse(
+      success: false,
+      failReason: LegacyAuthenticationFailReason.internalError,
+    );
+  }
+}
```
marcelomendoncasoares: Discussion: Wouldn't adding these others here prevent users from migrating email while the others are not available yet? I think the best would be to make the middleware only route email for now and completely remove these others. This will allow the migration package to coexist with the legacy until we have migrations for all providers.

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/models/legacy_user_info.spy.yaml
```diff
@@ -0,0 +1,11 @@
...
+fields:
+  id: int?
+  userIdentifier: String
+  userName: String?
+  fullName: String?
+  email: String?
+  created: DateTime
+  imageUrl: String?
+  scopeNames: List<String>
+  blocked: bool
```
marcelomendoncasoares, line 21: Suggestion: Add doc comments to the class and all fields. This also applies to the other models.

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/endpoints/legacy_admin_endpoint.dart
```diff
@@ -0,0 +1,10 @@
+import 'package:serverpod/serverpod.dart';
+
+/// Stub endpoint for legacy admin operations. Requires admin scope.
+class LegacyAdminEndpoint extends Endpoint {
+  @override
+  bool get requireLogin => true;
+
+  @override
+  Set<Scope> get requiredScopes => {Scope.admin};
+}
```
marcelomendoncasoares, line 106: Question: Aren't the methods missing for this endpoint? The original has three: `getUserInfo`, `blockUser` and `unblockUser`.

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/endpoints/legacy_email_endpoint.dart
```diff
@@ -0,0 +1,156 @@
...
+      );
+    }
+
+    final authUser = await AuthUser.db.findById(session, authUserId);
+    if (authUser == null || authUser.blocked) {
+      return LegacyAuthenticationResponse(
+        success: false,
+        failReason: LegacyAuthenticationFailReason.blocked,
+      );
+    }
```
marcelomendoncasoares, line 64: Question: This is actually a flaw on the `EmailIdp` implementation, right? Because the `AuthUserBlockedException` should be thrown inside its authenticate, but it seems this exception is not being thrown in nearly any provider - only Passkey tests blocked to actually block the user.

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/endpoints/legacy_email_endpoint.dart
```diff
@@ -0,0 +1,156 @@
...
+    final legacyRow = legacyRows.first;
+
+    final profile = await AuthServices.instance.userProfiles
+        .maybeFindUserProfileByUserId(session, authUserId);
+
+    final secret = generateRandomString();
+    final salt = session.passwords['authKeySalt'] ?? 'salty';
+    final hash = sha256.convert(utf8.encode(salt + secret)).toString();
```
marcelomendoncasoares, line 86: Suggestion: Add a `NOTE` explaining why this is done like this for clarity to future developers.

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/endpoints/legacy_user_endpoint.dart
```diff
@@ -0,0 +1,75 @@
...
+
+/// Proxy endpoint for legacy user profile operations (image, name changes).
+class LegacyUserEndpoint extends Endpoint {
+  @override
+  bool get requireLogin => true;
+
+  /// Removes the user's profile image.
+  Future<bool> removeUserImage(final Session session) async {
+    try {
+      final uuid = UuidValue.fromString(session.authenticated!.userIdentifier);
```
marcelomendoncasoares: Suggestion: Make this an extension somewhere, since this operation is repeated a few times on different files.

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/endpoints/legacy_user_endpoint.dart
```diff
@@ -0,0 +1,75 @@
...
+      await AuthServices.instance.userProfiles.setUserImageFromBytes(
+        session,
+        uuid,
+        image.buffer.asUint8List(),
+      );
+      return true;
+    } catch (_) {
+      return false;
+    }
+  }
```
marcelomendoncasoares, line 51: Question: All these methods are swallowing the exceptions completely. Should they at least log errors to the session for inspection?

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/business/enable_legacy_client_support.dart
```diff
@@ -0,0 +1,42 @@
...
+      final forwardedRequest = request.copyWith(
+        url: request.url.replace(
+          pathSegments: [forwardedEndpoint, ...pathSegments.skip(1)],
+        ),
+      );
+
+      return request.forwardTo(forwardedRequest);
+    };
+  });
+}
```
marcelomendoncasoares: Suggestion: Make this a `Serverpod` extension method instead to be called `pod.enableLegacyClientSupport()`.

tests/serverpod_auth_test/serverpod_auth_test_server/test/integration/legacy_client_support_test.dart
```diff
@@ -0,0 +1,575 @@
...
+    );
+  });
+
+  withServerpod(
+    'Given a migrated legacy user,',
+    testGroupTagsOverride: TestTags.concurrencyOneTestTags,
+    rollbackDatabase: RollbackDatabase.disabled,
+    (final sessionBuilder, final endpoints) {
+      // Must run before withServerpod starts the server; doing this in setUp
+      // is too late for the server auth handler wiring used by legacy routes.
```
marcelomendoncasoares: Question: Is this still true? Or just left from the previous implementation?

tests/serverpod_auth_test/serverpod_auth_test_server/test/integration/legacy_client_support_test.dart
```diff
@@ -0,0 +1,575 @@
...
+      _configurePublicLegacySupport(sessionBuilder);
```
marcelomendoncasoares, line 32: Suggestion: It is a bit more copy-paste, but, since there are only few lines inside this function and it is critical to the behavior of the package, I'd suggest replacing the function with its contents. As a private function it seems like something that is for the test only.

tests/serverpod_auth_test/serverpod_auth_test_server/test/integration/legacy_client_support_test.dart
```diff
@@ -0,0 +1,575 @@
...
+  final String password,
+) {
+  return client.callServerEndpoint<bridge_auth.LegacyAuthenticationResponse>(
+    'serverpod_auth.email',
+    'authenticate',
+    {
+      'email': email,
+      'password': password,
+    },
+  );
```
marcelomendoncasoares: Question: Can't we use the legacy client instead of simulating it?

tests/serverpod_auth_test/serverpod_auth_test_server/test/integration/legacy_client_support_test.dart
```diff
@@ -0,0 +1,575 @@
...
+
+      test(
+        'when authenticating with correct credentials then returns success with session key.',
+        () async {
+          final client = _createClient(sessionBuilder);
+          final result = await _legacyAuthenticate(
+            client,
+            email,
+            password,
+          );
```
marcelomendoncasoares: Suggestion: The same comment above is also valid for these functions, that are crucial to understand the test and are better inline than abstracted.

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/endpoints/legacy_admin_endpoint.dart
exaby73: This isn’t a bug in the `EmailIdp.login()` flow itself. Blocked enforcement currently happens at token issuance, which is where AuthUserBlockedException is thrown.
In this bridge endpoint we intentionally call the lower-level `utils.authentication.authenticate(...)` and then create a legacy session manually, so we bypass that token-manager check

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/endpoints/legacy_admin_endpoint.dart
```diff
@@ -1,10 +1,106 @@
...
+
+  /// Unblocks a user so that they can log in again.
+  Future<void> unblockUser(final Session session, final int userId) async {
+    final authUser = await _findAuthUserByLegacyUserId(session, userId);
+    if (authUser == null) {
+      throw StateError('userId $userId not found');
+    }
+    if (!authUser.blocked) {
+      throw StateError('userId $userId already unblocked');
+    }
```
marcelomendoncasoares, line 70: Suggestion: Throw a proper exception here instead of `StateError`, since this is a valid exception that can happen at runtime - and not a developer error.

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/endpoints/legacy_admin_endpoint.dart
marcelomendoncasoares: I'm understanding that your answer is for the question below, right? But understood, I'll resolve it.

tests/serverpod_auth_test/serverpod_auth_test_server/test/integration/legacy_client_support_test.dart
```diff
@@ -65,12 +65,14 @@ void main() {
...
-            password,
-          );
+          final legacyClient = _createLegacyClient(sessionBuilder);
+          final result =
+              await legacy_auth.Caller(
+                legacyClient,
+              ).email.authenticate(
+                email,
+                password,
+              );
```
marcelomendoncasoares, line 75: Question: Kind of a repeated question, but can't this be the usage of the actual legacy client? Using the legacy package as a dev dependency to do:

```dart
client.modules.auth.email.authenticate(
        email,
        password,
      );
```

marcelomendoncasoares (changes requested):
Nice improvement! A few last comments and we're good to go.

tests/serverpod_auth_test/serverpod_auth_test_server/test/integration/legacy_client_support_test.dart
exaby73: This is a generated API. We need to use the `Caller` directly and we can't generate the client with a newer version. To keep true to a pre-migrated client, I am calling it directly

modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_server/lib/src/endpoints/legacy_admin_endpoint.dart
exaby73: Fixed in 40c4af91

tests/serverpod_auth_test/serverpod_auth_test_server/test/integration/legacy_client_support_test.dart
marcelomendoncasoares: But the client is generated for this test project, with the module alias "auth". You can do `client.modules.auth.email.authenticate` and you'll be calling the legacy endpoint.

exaby73: Ahh I didn't realize it that same :) I assumed the legacy code gen would have had some differences but seems I was wrong. Updated in 4d3b0e8c

exaby73: So I remember why I did it. We only fallback on deserialization when `__className__` isn't there on `FormatException`. But because the bridge classes are duplicates and not sub classes of the original ones from legacy auth, it throws a type mismatch when it tries to cast. Casting works when it's an inheritance chain for forwards compatibility, but not when the class is completely different. I reverted this change

marcelomendoncasoares: The cause is actually because the project's `Client` knows the new class, so the deserialization returns it leads to a cast failure on the endpoint. But it can be solved if we create a legacy client that uses the legacy `Protocol`, which will then fallback to the expected class. Done at https://github.com/serverpod/serverpod/pull/4712/commits/253e48643a6adae915cdd99ae4049aa75822265f.

Issue #4722, opened 2026-02-12
cache should supporting stroring premitive data types


Hanibachi:
### Problem to Solve

currently, `session.cache.put()` only supports `SerializableModel`.
it should support the basic data types like string, int, double, datetime, bool .

### Proposal

broaden data type support to be the same as with endpoints

### Use Case

`session.cache.put('key','value');`

### Alternatives

creating a 'cached...' class for every basic data type.
```yaml
class: CachedString
fields:
  value: String
```
```yaml
class: Cachedint
fields:
  value: int
```
```yaml
class: CachedDateTime
fields:
  value: DateTime
```
... etc .

### How experienced are you with this library?

Beginner - Just getting started with this library

### Are you interested in working on a PR for this?

- [ ] I want to work on this

Issue #3493, opened 2025-05-01
Support caching of lists and maps of serializable models without manual wrapper classes


dkbast:
### Problem to Solve

Currently, `List<T extends SerializableModel>` does not conform to `SerializableModel`, which prevents lists of serializable models from being cached directly. Although lists are supported when sending data over endpoints, caching requires a workaround by wrapping the list in a custom object. This inconsistency makes caching cumbersome and adds unnecessary boilerplate.
The same is true for primitives which can be serialized for endpoints but not for caches or future calls.

### Proposal

Extend the caching system to support storing collections (Lists and Maps) of `SerializableModel` directly without requiring a wrapper class. This could be implemented by introducing serialization logic similar to the one used for endpoints, which already support collections and primitive types. This improvement would align caching capabilities with the endpoint serialization features.

### Use Case

An application wants to cache a list of fetched records (e.g. `List<User>`) for performance reasons. Instead of creating a `CachedUserList` wrapper implementing `SerializableModel`, developers should be able to cache the list directly:

```dart
await cache.put('users', userList); // userList is List<User>
```

This avoids repetitive and error-prone wrapper creation.

### Alternatives

- Manually wrap the list in a custom class implementing `SerializableModel` (current approach).
- Serialize the list to JSON manually and cache the raw JSON (loses type safety and is verbose).

### Additional context

Similar limitation exists when using future calls, as described in `Generate typed interface for future calls via code generation #3418`. Addressing this could provide a unified and more ergonomic developer experience across Serverpod.

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

SandPod:
One significant difference between endpoints/future calls and the cache is that the content added to the cache is not defined in code. It could simply be a list of "anything" added at runtime.

In order for us to generate valid serialization logic, the user's intents must be expressed somewhere.

Floris29:
Let the users define the class of the list?
 I would like to cache a list of chats. I can just call my Model Class where is defined what the contents are.

PR #4724, opened 2026-02-12
feat: Allow caching any type of object to the local/Redis cache

marcelomendoncasoares:
Fixes: #3493
Fixes: #4722

## Pre-launch Checklist

## Breaking changes

None, since the change relaxes the requirement on the type of the object being cached from `SerializableModel` only to any `Object`.

### DIFF
--- a/packages/serverpod/lib/src/cache/cache.dart
+++ b/packages/serverpod/lib/src/cache/cache.dart
@@ -14,7 +14,7 @@ abstract class Cache {
   /// Creates a new [Cache] of the specified size.
   Cache(this.maxLocalEntries, this.serializationManager);
 
-  /// Stores a copy of a [SerializableModel] in the cache using the specified
+  /// Stores a copy of the [Object] in the cache using the specified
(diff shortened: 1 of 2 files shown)


marcelomendoncasoares:
> LGTM. One suggestion though if we add type-safe accessors like `getString`, `getList<T extends SerializableModel>`, etc. instead so we remain type-safe, while also documenting which types we support. `get<T extends Object>` could remain.
> 
> Extending this, as a user, how do I know whether an object can be saved to cache?

All types that can be serialized with endpoints or the database are now supported. If an object is not supported, an informative error will occur. It won't be a compilation error, but it will be easy enough to spot. To ensure this behavior, a test was added on https://github.com/serverpod/serverpod/pull/4724/commits/af4fb3fc26cfc09155c48d7b5bbb8c477190d7df.

The only concern that could be raised with the current interface is the possibility of a user to serialize one class and request another. But this was already possible prior to this change between serializable models, so it is not an issue for the merge.

PR #4729, opened 2026-02-13
feat: Add `ignoreConflicts` support for insert queries

### LINKED ISSUE #3963: Support for ON CONFLICT DO NOTHING in Inserts
### Problem to Solve

When inserting data into tables with unique constraints, inserts may fail if another process or user has already inserted the same row. For example, in a junction table with a unique (campaignId, contactId) constraint, an "insert all" fails if a duplicate already exists. This creates race conditions and complicates bulk insert operations.

Currently, the only way to handle this is by writing raw SQL with ON CONFLICT DO NOTHING, which breaks Serverpod’s abstraction and makes the code harder to maintain.

### Proposal

Introduce support for ON CONFLICT DO NOTHING in Serverpod’s insert APIs.
This could be exposed as an option or flag when calling an insert method, e.g.:

```
await MyModel.db.insert(
      session,
      rows,
      transaction: transaction,
      onConflict: OnConflict.doNothing,
    );
```

This way, conflicts are gracefully ignored and the operation continues without errors.

### Use Case

Entity: Campaign

Entity: Contact

Junction table with unique constraint (campaignId, contactId)

A user selects 100 contacts in the UI and adds them to a campaign.
If some contacts were already added by another user or process in the meantime, those should simply be skipped while inserting the rest successfully.

Without ON CONFLICT DO NOTHING, the entire insert operation fails.

### Alternatives

Raw SQL queries

### Additional context

_No response_

- [ ] I want to work on this

FXschwartz:
- Adds `ignoreConflicts` parameter to `insert` that appends `ON CONFLICT DO NOTHING` to the SQL, silently skipping rows that violate unique constraints and returning only the successfully inserted rows.

Closes #3963 

## Pre-launch Checklist

## Breaking changes
none

### DIFF
--- a/packages/serverpod/lib/src/server/session.dart
+++ b/packages/serverpod/lib/src/server/session.dart
@@ -253,6 +253,7 @@ abstract class Session implements DatabaseSession {
 
   /// Logs a message. Default [LogLevel] is [LogLevel.info]. The log is written
   /// to the database when the session is closed.
+  @override
   void log(
     String message, {
     LogLevel? level,


packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -179,15 +180,26 @@ class DatabaseConnection {
...
+      transaction: transaction,
+    );
+
+    // When ignoring conflicts, RETURNING only yields actually inserted rows,
+    // so we can't match them back to input rows by index. Skip the merge step.
+    if (ignoreConflicts) {
+      return dbResults
+          .map(_poolManager.serializationManager.deserialize<T>)
+          .toList();
+    }
```
marcelomendoncasoares: Question: Won't this break the deserialization if the model has non-persistent non-nullable fields (such as [the example from the docs](https://docs.serverpod.dev/concepts/database/models#non-persistent-fields)? Or silently return wrong results if the non-persistent fields are nullable.

The `_mergeResultsWithNonPersistedFields` is already very brittle since [Postgres provides no guarantees regarding the order of the returning clause](https://stackoverflow.com/questions/5439293/is-insert-returning-guaranteed-to-return-things-in-the-right-order). It seems to return in order, but [we're relying on implementation details rather than on documented behaviors](https://www.postgresql.org/message-id/CAMsr+YEn5TOuhv_tTwY70S1zXf8jCQ-uixU8aOs4OQs7kojf6Q@mail.gmail.com).

That being said, we do have it working today. So how should we handle the `ignoreConflicts` with models that have non-persistent fields? I can think of a few approaches:

- Forbid the flag for models with non-persistent fields to prevent deserialization issues.
- Try a best effort to find matching rows (e.g. for immutable models or models that define the ID on the application) - which is probably worst than throwing.
- Fallback internally to a batch of single inserts on these cases so we can safely return all inserted models mapped.

If we don't have another creative alternative to make it work safely, maybe the fallback to single inserts on these cases will be our safest bet - with a performance warning if the batch is too big. What do you think?

packages/serverpod/test/database/database_query/insert_sql_query_test.dart
```diff
@@ -224,4 +224,99 @@ SELECT * FROM insertWithIdNotNull
...
+
+        expect(
+          query,
+          'INSERT INTO "person" ("name", "age") VALUES (\'Alex\', 33) ON CONFLICT DO NOTHING RETURNING *',
+        );
+      },
+    );
+
+    test(
+      'when building insert query with a row with id then ON CONFLICT DO NOTHING is appended.',
```
marcelomendoncasoares: Suggestion: Same general comment regarding test organization (setup only on Given, acting on when, then with results) applies here.

tests/serverpod_test_server/test_integration/database_operations/crud/insert_ignore_conflicts_test.dart
```diff
@@ -0,0 +1,127 @@
...
+  tearDown(() async {
+    await UniqueData.db.deleteWhere(
+      session,
+      where: (t) => Constant.bool(true),
+    );
+  });
+
+  group('Given an empty database', () {
+    test(
+      'when inserting with ignoreConflicts then all rows are inserted and returned.',
```
marcelomendoncasoares, line 27: Suggestion: We also need to cover insert of models with non-persistent fields here.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
FXschwartz: Ah you are right, I totally missed that! 

I would lean with the fallback option as well. It's the safest, preserves correctness, and the performance cost would only affect models with non-persistent fields.

I'll get to fixing this!

FXschwartz: Done!

tests/serverpod_test_server/test_integration/database_operations/crud/insert_ignore_conflicts_test.dart
FXschwartz: Added!

packages/serverpod/test/database/database_query/insert_sql_query_test.dart
FXschwartz: Fixed!

FXschwartz:
@marcelomendoncasoares All comments addressed and all checks are passing. Ready for another review!

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -171,23 +171,46 @@ class DatabaseConnection {
...
+
+    return _mergeResultsWithNonPersistedFields(rows)(
+      dbResults,
+    ).map(_poolManager.serializationManager.deserialize<T>).toList();
```
marcelomendoncasoares: Suggestion: Revert this change, since the case for non-persistent fields is already handled separately above.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -962,6 +985,55 @@ class DatabaseConnection {
...
+      if (dbResults.isEmpty) continue;
+
+      var merged = _mergeResultsWithNonPersistedFields([row])(dbResults);
+      results.add(
+        _poolManager.serializationManager.deserialize<T>(merged.first),
+      );
+    }
+
+    return results;
+  }
```
marcelomendoncasoares: Suggestion: Instead of duplicating the logic, we can simply loop over rows calling the `insert` function recursively for each individual row.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -171,23 +171,46 @@ class DatabaseConnection {
...
+    // When ignoring conflicts AND the model has non-persistent fields,
+    // fall back to single-row inserts so we can safely merge non-persistent
+    // fields with each successfully inserted row.
+    if (ignoreConflicts && _hasNonPersistedFields(rows.first)) {
+      return _insertRowsIndividuallyIgnoringConflicts<T>(
+        session,
+        rows,
+        transaction: transaction,
+      );
+    }
```
marcelomendoncasoares: Suggestion: To make the recursive call possible, we should only delegate to individual inserts if the insert is not individual already. For an insert of 1 row, it is safe to use the existing code, because the return will either be the list with the record or an empty list - in which case the `_mergeResultsWithNonPersistedFields` will do nothing.

To keep the recursion clear, I think we can even dispense a private method for this and just loop directly:

```suggestion
    if (ignoreConflicts &&
        rows.length > 1 &&
        _hasNonPersistedFields(rows.first)) {
      return [
        for (var row in rows)
          await insert<T>(
            session,
            [row],
            transaction: transaction,
            ignoreConflicts: ignoreConflicts,
          ).then((results) => results.firstOrNull),
      ].whereType<T>().toList();
    }
```

tests/serverpod_test_server/test_integration/database_operations/crud/insert_ignore_conflicts_with_non_persisted_fields_test.dart
marcelomendoncasoares, line 1: Suggestion: Merge these tests in the `insert_ignore_conflicts_test.dart` file to avoid fragmentation of test cases that are expected to live together. Then, differentiate them in the `Given` of the description. For example: "Given an empty database and a table model with non persistent fields"...

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -962,6 +985,55 @@ class DatabaseConnection {
...
+  /// represented in the table's columns (i.e., non-persistent fields).
+  bool _hasNonPersistedFields(TableRow row) {
+    var json = row.toJson();
+    if (json is! Map<String, dynamic>) return false;
+
+    var columnFieldNames = row.table.columns.map((c) => c.fieldName).toSet();
+    return json.keys
+        .where((k) => k != '__className__')
+        .any((key) => !columnFieldNames.contains(key));
+  }
```
marcelomendoncasoares: Suggestion: Make this function receive all rows instead of just one to also log a warning in case the number of rows can cause performance issues. Then cover it on a test that ensure the expected warning message.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
FXschwartz: Reverted!

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
FXschwartz: Done!

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
FXschwartz: Nice yeah this is way cleaner. Done!

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
FXschwartz: Done! For the warning I followed the example/pattern of using `stderr.writeln` in `health_check_manager.dart` but let me know if you had something else in mind.

tests/serverpod_test_server/test_integration/database_operations/crud/insert_ignore_conflicts_with_non_persisted_fields_test.dart
FXschwartz: Merged!

FXschwartz:
@marcelomendoncasoares All comments addressed and CI is passing. Ready for another review!

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -962,6 +980,32 @@ class DatabaseConnection {
...
+        .where((k) => k != '__className__')
+        .any((key) => !columnFieldNames.contains(key));
+
+    if (hasNonPersisted && rows.length > 100) {
+      stderr.writeln(
+        'WARNING: Inserting ${rows.length} rows with ignoreConflicts on a '
+        'model with non-persistent fields. This requires individual inserts '
+        'and may cause performance issues. Consider removing non-persistent '
+        'fields or inserting in smaller batches.',
+      );
```
marcelomendoncasoares: Suggestion: Use the `session.log` instead of writing to the `stderr` so that it can be inspected together with the queries logged on the session. It is also nice to add the table name to the message.

tests/serverpod_test_server/test_integration/database_operations/crud/insert_ignore_conflicts_test.dart
```diff
@@ -0,0 +1,337 @@
...
+
+        expect(
+          output.toString(),
+          contains(
+            'WARNING: Inserting 101 rows with ignoreConflicts on a model '
+            'with non-persistent fields.',
+          ),
+        );
+      },
+    );
```
marcelomendoncasoares: Suggestion: Detach this into a test in the bottom of the file, since it is an edge case, like:
- Given model with non-persistent fields and a batch of more than 100 entries
  - when inserting all rows with ignoreConflicts then a performance warning is logged.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
FXschwartz: Done!

tests/serverpod_test_server/test_integration/database_operations/crud/insert_ignore_conflicts_test.dart
FXschwartz: Done!

FXschwartz:
@marcelomendoncasoares Comments addressed and is ready for another review!

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -975,6 +996,38 @@ class PostgresDatabaseConnection
...
+      if (s is Session) {
+        s.log(
+          'WARNING: Inserting ${rows.length} rows with ignoreConflicts on '
+          'table "${table.tableName}" with non-persistent fields. This '
+          'requires individual inserts and may cause performance issues. '
+          'Consider removing non-persistent fields or inserting in smaller '
+          'batches.',
+          level: LogLevel.warning,
+        );
+      }
```
marcelomendoncasoares: Suggestion: Instead of pushing the dependency on the server's `Session` here, a better alternative is to push the `log` function down to the `DatabaseSession` interface. Otherwise we loose part of the work done on #4748 to isolate the database code from the server as a preparation to move it to a separate package.

```dart
/// Interface for accessing the database.
abstract interface class DatabaseSession {
  /// The database to access.
  Database get db;

  /// Optional transaction to use for all database queries.
  Transaction? get transaction;

  /// Optional function to log a query.
  LogQueryFunction? get logQuery;

  /// Optional function to log a message as part of the session.
  void log(
    String message, {
    LogLevel? level,
    dynamic exception,
    StackTrace? stackTrace,
  });
}
```

tests/serverpod_test_server/test_integration/database_operations/crud/insert_ignore_conflicts_test.dart
```diff
@@ -0,0 +1,337 @@
...
+            expect(inserted.first.extra, 'transacted-extra');
+          });
+
+          var allRows = await UniqueDataWithNonPersist.db.find(session);
+          expect(allRows, hasLength(2));
+        },
+      );
+
+      test(
+        'when all rows conflict with ignoreConflicts then an empty list is returned.',
```
marcelomendoncasoares: Suggestion: Clarify the acting on the description.

```suggestion
        'when inserting only conflicting rows with ignoreConflicts then an empty list is returned.',
```

tests/serverpod_test_server/test_integration/database_operations/crud/insert_ignore_conflicts_test.dart
```diff
@@ -0,0 +1,337 @@
...
+
+          var logEntries = await protocol.LogEntry.db.find(
+            session,
+            where: (t) => t.message.like(
+              '%Inserting 101 rows with ignoreConflicts%',
+            ),
+          );
+
+          expect(logEntries, hasLength(1));
+          expect(logEntries.first.logLevel, LogLevel.warning);
```
marcelomendoncasoares, line 332: Suggestion: Also add an `expect` to the full string with the message so that we ensure the UX that it provides.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
FXschwartz: Ah yeah my bad on this. Fixed!

tests/serverpod_test_server/test_integration/database_operations/crud/insert_ignore_conflicts_test.dart
FXschwartz: Done!

tests/serverpod_test_server/test_integration/database_operations/crud/insert_ignore_conflicts_test.dart
FXschwartz: Fixed!

Issue #3463, opened 2025-04-24
Support generating value-based models in Flutter from Serverpod YAML files


dkbast:
### Problem to Solve

Serverpod currently generates models primarily for use with backend communication. However, in Flutter apps, especially in offline or view-model contexts, developers often need value-equatable, serializable models for local state management and UI logic. Without code generation support on the Flutter side, users will have to use another package for achieving pretty much the same what is already built in to Serverpod (eg use freezed, dart mappable etc etc)

### Proposal

Enable model generation in Flutter apps directly from any `serverpod.yaml` file found in the workspace. This generation should:
- Produce Dart classes that support value equality (e.g., `==` and `hashCode`).
- Include JSON serialization methods.
- Work without requiring network dependencies (i.e., models are purely structural/data-focused).
- Be usable for offline apps, widget state management, or view models.

The codegen should be runnable anywhere in the workspace, aligning with Serverpod’s direction of broader workspace flexibility.

### Use Case

A developer working on a Flutter app with offline-first capabilities or complex UI state where the user usually would resort to a generate state class. With this feature, they can use Serverpod’s YAML files to generate these models directly within the app, ensuring consistency and eliminating boilerplate, without needing to connect to the backend or run a server.

### Alternatives

- Using third-party model generators (e.g., Freezed or json_serializable) outside of Serverpod's toolchain.

### How experienced are you with this library?

None

### Are you interested in working on a PR for this?

- [ ] I want to work on this

vlidholt:
This could potentially be nice, but it's definitely doable today. You just need to place the models somewhere in your server package. Having them generated on the server isn't really a problem (they will even be tree-shaken on server compilation). Perhaps this should be mentioned in the documentation as a starter, though. Created a documentation issue here: #3670

Issue #3339, opened 2025-03-22
server should depend on client to fix duplicate classes


vanlooverenkoen:
### Describe the bug

Right now all models are generated in server & client. This results in having duplicated classes if you want to use a shared package that can be used by server & frontend. (like extensions) 

Why is server just no using client for all the models? 

### To reproduce

Just any serverpod project

### Expected behavior

I would assume that server depends on client

### Library version

2.4.0

### Platform information

all platforms

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

SandPod:
This is discussed in part here with a possible workaround (quite advanced): https://github.com/serverpod/serverpod/issues/1256

The client versions of the models can't be used directly because there might be `ServerOnly` fields.

With the introduction of interfaces, https://github.com/serverpod/serverpod/issues/2842, it would be possible to use the same definition if we were to exclude that feature.

vanlooverenkoen:
@SandPod perfect!

Indeed the ServerOnly fields will still only be generated in the `server` package. but everything else should be the exact same class between `server` & `frontend`

Issue #993, opened 2023-06-03
Need way of marking server class for inclusion in client


lukehutch:
**Is your feature request related to a problem? Please describe.**

Often a client and server need to share code. Currently with Serverpod, as far as I can tell you have to manually keep in sync two copies of any class that you need on both the client and server.

**Describe the solution you'd like**

There needs to be an annotation (e.g. `@clientAndServer`) that can be added to a server class to cause that class to be copied over verbatim into the client when `serverpod generate` is called.

Issue #3670, opened 2025-06-18
Mention creating models that are used client-side only


vlidholt:
### Summary of the issue

It's currently not possible to create models that are client-side only. However, there is no harm in creating models that are only used by the client. We should mention this in the documentation.

### How experienced are you with this library?

None

### Are you interested in working on a PR for this?

- [ ] I want to work on this

abdulawalarif:
I opened a [PR](https://github.com/serverpod/serverpod_docs/pull/739) that documents using models on the client only.
There’s no `clientOnly` flag, but it’s fine to define models you only use in Flutter.

PR #4737, opened 2026-02-17
feat: Allow generating shared models between server and client

marcelomendoncasoares:
This PR introduces the ability to create shared packages that contains only models and the protocol file. This shared package depends exclusively on the `serverpod_serialization` package and is safe to be imported in both the server and client code.

Closes: https://github.com/serverpod/serverpod/issues/3339
Closes: https://github.com/serverpod/serverpod/issues/993
Closes: https://github.com/serverpod/serverpod/issues/3463

The shared package is tied to the project through the `shared_packages` field in the `config/generator.yaml` file, but the models and protocol file are generated in the package's own directory. This means that the code on the shared package is generated as part of the `serverpod generate` command called for the project.

```yaml
# config/generator.yaml
shared_packages:
  - ../shared_package
  - ../another_shared_package
```

After merge, we can work on having this shared package as available in the template to close the original issue: https://github.com/serverpod/serverpod/issues/1256.

Related:
- https://github.com/serverpod/serverpod/issues/3670

## Restrictions

Nearly all features are supported on shared models, with the exception of:

- Specifying the `table` property, since the models are not tied to a database.
- Setting the `scope` property to `serverOnly` on model fields, since models are shared.
- Setting the `serverOnly` property on the model class, since models are shared.

## Notes for reviewers

This PR can be reviewed commit by commit for a better comprehension of the changes. The main changes are:

1. Addition of the configuration on the `generator.yaml` file.
2. Support on the `ModelSource` and `SerializableModelDefinition` to identify shared models.
3. Enhancement of the `TypeDefinition` to generate correct references to shared models.
4. Filtering of the shared models on the `DartServerCodeGenerator` and `DartClientCodeGenerator`.
5. Implementation of the `DartSharedCodeGenerator` as a copy of the `DartClientCodeGenerator`, but targeting the shared models.

## Regarding tests

Since this feature introduces a new `DartSharedCodeGenerator`, the applying tests for the `DartServerCodeGenerator` and `DartClientCodeGenerator` were copied and adapted to the new generator. There are only two test files that cover functionality specific to the shared package generator and are worth reviewing:

- `shared_code_generator_test.dart`
- `shared_models_extends_test.dart`
- `shared_models_restrictions_test.dart`

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/config/config.dart
+++ b/tools/serverpod_cli/lib/src/config/config.dart
@@ -554,12 +554,12 @@ Map<String, List<String>> _extractSharedPackages(
   String serverRootDir,
   YamlMap generatorConfig,
 ) {
-  var sharedModelPackagesPathParts = <String, List<String>>{};
-
   var sharedPackages = generatorConfig['shared_packages'];
   if (sharedPackages == null) {
-    return sharedModelPackagesPathParts;
+    return {};
   }
+
+  var sharedModelPackagesPathParts = <String, List<String>>{};
 
   if (sharedPackages is! YamlList) {
     throw SourceSpanFormatException(
@@ -569,7 +569,7 @@ Map<String, List<String>> _extractSharedPackages(
   }
 
   for (var path in sharedPackages) {
-    if (path is! String || !path.startsWith('.')) {
+    if (path is! String || p.isAbsolute(path)) {
       throw SourceSpanFormatException(
         'The path for the shared package must be a string path relative to the '
         'server package. Current path: $path',
@@ -584,8 +584,8 @@ Map<String, List<String>> _extractSharedPackages(
       sharedModelPackagesPathParts[pubspec.name] = p.split(path);
     } catch (_) {
       throw const ServerpodProjectNotFoundException(
-        'Failed to load client pubspec.yaml. If you are using a none default '
-        'path it has to be specified in the config/generator.yaml file!',
+        'Failed to load shared package pubspec.yaml. Make sure the path is '
+        'correctly specified in the config/generator.yaml file.',
       );
     }
   }
--- a/tools/serverpod_cli/test/integration/generator_config/shared_packages_config_test.dart
+++ b/tools/serverpod_cli/test/integration/generator_config/shared_packages_config_test.dart
@@ -159,7 +159,8 @@ shared_packages:
               isA<SourceSpanFormatException>().having(
                 (e) => e.message,
                 'message',
-                contains('relative to the'),
+                'The path for the shared package must be a string path relative '
+                    'to the server package. Current path: 123',
               ),
             ),
           );
@@ -169,14 +170,14 @@ shared_packages:
   );
 
   group(
-    'Given a generator.yaml with shared_packages path not starting with "."',
+    'Given a generator.yaml with an absolute shared_packages path',
     () {
       setUpAll(() async {
         var projectDir = _createProject(
           generatorYamlContent: '''
 type: server
 shared_packages:
-  - packages/shared
+  - /packages/shared
 ''',
         );
         await projectDir.create();
@@ -194,7 +195,8 @@ shared_packages:
               isA<SourceSpanFormatException>().having(
                 (e) => e.message,
                 'message',
-                contains('must be a string path relative to the'),
+                'The path for the shared package must be a string path relative '
+                    'to the server package. Current path: /packages/shared',
               ),
             ),
           );
@@ -230,7 +232,8 @@ shared_packages:
               isA<ServerpodProjectNotFoundException>().having(
                 (e) => e.message,
                 'message',
-                contains('Failed to load'),
+                'Failed to load shared package pubspec.yaml. Make sure the path '
+                    'is correctly specified in the config/generator.yaml file.',
               ),
             ),
           );


tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -538,6 +550,49 @@ generatedServerModel: ${p.joinAll(generatedServeModelPathParts)}
...
 
+Map<String, List<String>> _extractSharedPackages(
+  String serverRootDir,
+  YamlMap generatorConfig,
+) {
+  var sharedModelPackagesPathParts = <String, List<String>>{};
+
+  var sharedPackages = generatorConfig['shared_packages'];
+  if (sharedPackages == null) {
+    return sharedModelPackagesPathParts;
```
nielsenko: Slightly clearer to return `const {}`

tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -538,6 +550,49 @@ generatedServerModel: ${p.joinAll(generatedServeModelPathParts)}
...
+
+  if (sharedPackages is! YamlList) {
+    throw SourceSpanFormatException(
+      'The "shared_packages" property must be a list of package paths.',
+      sharedPackages is YamlNode ? sharedPackages.span : null,
+    );
+  }
+
+  for (var path in sharedPackages) {
+    if (path is! String || !path.startsWith('.')) {
```
nielsenko: Surprised we don't check if path is absolute instead?

tools/serverpod_cli/lib/src/config/config.dart
```diff
@@ -538,6 +550,49 @@ generatedServerModel: ${p.joinAll(generatedServeModelPathParts)}
...
+        'The path for the shared package must be a string path relative to the '
+        'server package. Current path: $path',
+        sharedPackages.span,
+      );
+    }
+
+    try {
+      var pubspecFile = File(p.join(serverRootDir, path, 'pubspec.yaml'));
+      var yamlStr = pubspecFile.readAsStringSync();
+      var pubspec = Pubspec.parse(yamlStr);
```
nielsenko, line 583: `PubspecPlus` gives much better error messages

tools/serverpod_cli/test/integration/generator_config/shared_packages_config_test.dart
```diff
@@ -0,0 +1,314 @@
...
+
+  group(
+    'Given a generator.yaml with shared_packages pointing to a valid package',
+    () {
+      setUpAll(() async {
+        var projectDir = _createProject(
+          generatorYamlContent: '''
+type: server
+shared_packages:
+  - ./packages/shared
```
nielsenko, line 76: Why not allow `- "packages/shared"` (see pervious comment on validation)

tools/serverpod_cli/test/integration/generator_config/shared_packages_config_test.dart
```diff
@@ -0,0 +1,314 @@
...
+              ),
+            ),
+          );
+        },
+      );
+    },
+  );
+
+  group(
+    'Given a generator.yaml with shared_packages path not starting with "."',
```
nielsenko: I don't get it? Why is '.' required? Why not just insist it is relative? There are better ways to check that.

tools/serverpod_cli/lib/src/generator/generator_continuous.dart
nielsenko, line 54: Not related to this PR, but something I realized in another setting. If a directory is created/deleted you won't see events for the files inside (at least not on all platforms).

We should consider traversing the subtree on directory add, and prune the full subtree on delete. We cannot depend on individual file events.

tools/serverpod_cli/lib/src/language_server/language_server.dart
```diff
@@ -236,6 +244,20 @@ List<DiagnosticTag>? _convertToDiagnosticTags(List<SourceSpanTag>? tags) {
...
+  Uri uri,
+  Uri serverUri,
+  GeneratorConfig config,
+) {
+  var path = uri.path;
+  for (var e in config.sharedModelsSourcePathsParts.entries) {
+    if (path.contains(e.value.join('/'))) {
+      return true;
+    }
+  }
```
nielsenko, line 257: This makes me think we are using the wrong data-structure. Are we ever doing lookup by package name?

tools/serverpod_cli/lib/src/util/model_helper.dart
```diff
@@ -172,7 +221,76 @@ class ModelHelper {
...
+  ) {
+    var serverLibPath = absolute(
+      joinAll([...config.serverPackageDirectoryPathParts, 'lib']),
+    );
+    if (isWithin(serverLibPath, absolutePath)) return null;
+    for (var e in config.sharedModelsSourcePathsParts.entries) {
+      var rootParts = [
+        ...config.serverPackageDirectoryPathParts,
+        ...e.value,
+      ];
```
nielsenko, line 261: again..

tools/serverpod_cli/lib/src/generator/dart/library_generators/util/model_generators_util.dart
```diff
@@ -122,8 +122,23 @@ extension SerializableModelPath on SerializableModelDefinition {
   ]);
 
   /// Returns a String with the full server or client path followed by
-  /// `filename.dart`.
+  /// `filename.dart`. When the model is a shared model, returns the path inside
+  /// the shared package's generated directory.
   String getFullFilePath(GeneratorConfig config, {required bool serverCode}) {
+    if (isSharedModel && sharedPackageName != null) {
+      var pathParts = config.sharedModelsSourcePathsParts[sharedPackageName];
```
nielsenko, line 129: Ahh.. here we use `operator[]` 👍

tools/serverpod_cli/lib/src/generator/types.dart
```diff
@@ -350,19 +357,35 @@ class TypeDefinition {
               'package:'
               '${serverCode ? config.serverPackage : config.dartClientPackage}'
               '/${split[1]}';
-        } else if (url == defaultModuleAlias) {
-          // protocol: reference
+        } else if (url == defaultModuleAlias ||
+            config.sharedModelsSourcePathsParts.containsKey(url)) {
```
nielsenko, line 361: 👍

tools/serverpod_cli/test/test_util/builders/exception_class_definition_builder.dart
```diff
@@ -84,4 +90,16 @@ class ExceptionClassDefinitionBuilder {
...
   }
+
+  ExceptionClassDefinitionBuilder withSharedPackageName(
+    String? sharedPackageName,
+  ) {
+    _sharedPackageName = sharedPackageName;
+    return this;
+  }
+
+  ExceptionClassDefinitionBuilder withTypeUrl(String? url) {
```
nielsenko, line 101: Why is `url` a `String` and not a `Uri`?

packages/serverpod_serialization/pubspec.yaml
```diff
@@ -12,6 +12,7 @@ environment:
   sdk: '^3.8.0'
 
 dependencies:
+  meta: ^1.15.0
```
nielsenko, line 15: We should tighten this (we don't really support ^1.15.0).

tests/serverpod_test_server/lib/src/generated/shared_model_subclass.dart
```diff
@@ -0,0 +1,156 @@
...
+// ignore_for_file: public_member_api_docs
+// ignore_for_file: type_literal_in_constant_pattern
+// ignore_for_file: use_super_parameters
+// ignore_for_file: invalid_use_of_internal_member
+
+// ignore_for_file: no_leading_underscores_for_library_prefixes
+import 'package:serverpod_test_shared/serverpod_test_shared.dart' as _i1;
+import 'package:serverpod/serverpod.dart' as _i2;
+
+abstract class SharedModelSubclass extends _i1.SharedModel
```
nielsenko, line 16: Why is `SharedModelSubclass` defined here and also in the [shared project](https://github.com/serverpod/serverpod/pull/4737/changes/5d2f46e2584b7a80d28c9f5619187e26633725e7#diff-aafefbfde731b1d1663c0732b6e32a03d719d5f1af0f3be64fb55cb2e2eb2627?)?

nielsenko (commented):
Mostly questions..

FXschwartz:
@marcelomendoncasoares Would this allow us to define a `.spy.yaml` table definition using a separately defined model in this shared package?

For instance if I defined a `Car` model in this shared package could I then go over to the server and create a `car.spy.yaml` and reference this shared `Car` model but then define the table as well?

Let me know if this is off topic and I can create a feature request to move the discussions to.

marcelomendoncasoares:
> Would this allow us to define a `.spy.yaml` table definition using a separately defined model in this shared package?
> 
> For instance if I defined a `Car` model in this shared package could I then go over to the server and create a `car.spy.yaml` and reference this shared `Car` model but then define the table as well?
> 
> Let me know if this is off topic and I can create a feature request to move the discussions to.

@FXschwartz Precisely! Shared packages won't allow the definition of table models, but you'll be able to declare a base model on your shared package and extend it on the server to add a `table` key.

tests/serverpod_test_server/lib/src/generated/shared_model_subclass.dart
marcelomendoncasoares: There are two subclasses: the `SharedSubclass` on the `serverpod_test_shared` package and the `SharedModelSubclass` on the `serverpod_test_server` that extends the `SharedModel` from the `serverpod_test_shared`. Maybe the names are a bit confusing indeed.

packages/serverpod_serialization/pubspec.yaml
marcelomendoncasoares: I have only copied the dependency from the `serverpod_client`, but you're right that we should revisit our dependencies as a whole. Since this same dependency is repeated in several packages, I would only change it in a separate PR to avoid creating an inconsistency or bundling unrelated changes.

tools/serverpod_cli/lib/src/config/config.dart
marcelomendoncasoares: This is indeed better. The look for `.` was just a shortcut with the idea of enforcing to not be in the same directory of the server package, but it is indeed a poor implementation. I'll change to an absolute check and leave the placement free.

tools/serverpod_cli/lib/src/config/config.dart
marcelomendoncasoares: Agree with you, although I wouldn't add it without replacing everywhere we interact with a `pubspec` file for consistency.

tools/serverpod_cli/lib/src/generator/generator_continuous.dart
marcelomendoncasoares: Surprising we don't have an issue for it until now. Maybe few people use the `watch` mode?

tools/serverpod_cli/test/test_util/builders/exception_class_definition_builder.dart
marcelomendoncasoares: This is a possible enhancement for the `TypeDefinition` class. The parameter is `String` here to follow the same type used on the `TypeDefinition` (which maps to 3 years ago):

```dart
TypeDefinition({
  required String className,
  List<TypeDefinition> generics = const [],
  required bool nullable,
  String? url,
  DartType? dartType,
  bool customClass = false,
  EnumDefinition? enumDefinition,
  SerializableModelDefinition? projectModelDefinition,
  String? recordFieldName,
  int? vectorDimension,
})
```

It was just not exposed on the `ExceptionClassDefinitionBuilder`, so I added it to use on the tests.

packages/serverpod_serialization/pubspec.yaml
marcelomendoncasoares: Mapped this task on issue https://github.com/serverpod/serverpod/issues/4745.

tools/serverpod_cli/test/integration/generator_config/shared_packages_config_test.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/4737/commits/157b59ab68b1c2f4f32053e7b27669fc228bc0d4.

tools/serverpod_cli/test/integration/generator_config/shared_packages_config_test.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/4737/commits/157b59ab68b1c2f4f32053e7b27669fc228bc0d4.

tools/serverpod_cli/lib/src/config/config.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/4737/commits/157b59ab68b1c2f4f32053e7b27669fc228bc0d4.

tools/serverpod_cli/lib/src/config/config.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/4737/commits/157b59ab68b1c2f4f32053e7b27669fc228bc0d4.

tools/serverpod_cli/test/test_util/builders/exception_class_definition_builder.dart
nielsenko: @coderabbitai Create a follow up issue for this

tools/serverpod_cli/lib/src/generator/generator_continuous.dart
nielsenko: @coderabbitai Create a follow-up issue for this

tests/serverpod_test_server/lib/src/generated/shared_model_subclass.dart
nielsenko: My bad 😊

PR #4741, opened 2026-02-17
feat: Allow attaching custom metadata to SSS/JWT tokens

### LINKED ISSUE #3165: Support additional token data in AuthenticationInfo
### Problem to Solve

When authenticating user requests, I would like the possibility to retrieve additional data from the authentication token and pass that on to my application. Currently, the ´AuthenticationInfo` only allows `userId`, `scopes`, and `authId` to be passed back from the `AuthenticationHandler` method.

### Solve the problem in code

Add an additional field or getter to `AuthenticationInfo` that can be used to retrieve user-defined data from the class.

### Document a suggested way around

Add documentation to explain how to get around this limitation if we would like to keep the interface at a minimum.

### Use Case

If I issue tokens with additional information used by the application's business logic.
If I interact with tokens from other services that provide additional information used for the application's business logic.

### Workarounds

1. I can subclass `AuthenticationInfo` and add an extension that retrieves the desired data or either throws or returns `null` depending on my application's needs.
2. I can store any additional data in the dynamic field `userObject` on the Session object.

### Additional context

_No response_

- [ ] I want to work on this

### LINKED ISSUE #4738: Add a user-defined 'name' field to auth sessions / tokens
### Problem to Solve

In order to use `ServerSideSession.createSession()` / `Jwt.createTokens()` for automation and scripting purposes, it currently becomes unwieldy as soon as more than one is created since it is hard to keep track of which is for what. This is especially detrimental considering old tokens ought to be cleaned out and it is easy to remove the wrong one, leading to pipelines or scripts failing.

### Proposal

It ought to be possible to assign an (optional, nullable) user-defined `name` to each token. The auth framework does not need to care about its value other than maybe a reasonable length limitation, not even uniqueness - just pass it back to API users in the `ServerSideSessionInfo` / `JwtTokenInfo` objects.

### Use Case

When creating personal access / CLI / API tokens this would be very valuable, for observability of which tokens are issued and for what.

### Alternatives

Create one's own database table with a reference to the sessions/tokens and store the name there, adding significant complexity for storing just one field.

### Additional context

The `method` field could in theory be used for this, but that would mix up separate concerns - how the session/token was created, and its purpose.

- [ ] I want to work on this

marcelomendoncasoares:
This PR closes #3165 by introducing a callback on each of the default token manager configs to allow attaching custom metadata to tokens. Since with the metadata it is possible to add any custom information, it also closes #4738 with no need for changes on the `ServerSideSessionInfo` / `JwtTokenInfo` classes.

## Using the callback

Considering the following schema for the metadata:

```yaml
class: SessionMetadata
serverOnly: true
table: session_metadata
fields:
  ### The [ServerSideSession] this metadata belongs to
  serverSideSession: module:serverpod_auth_core:ServerSideSession?, relation(onDelete=Cascade)
  
  ### Device information for the session
  deviceName: String
  
  ### IP address from which the session was created
  ipAddress: String?
  
  ### User agent string
  userAgent: String?
  
  ### Additional metadata stored as JSON
  metadata: String?

indexes:
  server_side_session_id_unique_idx:
    fields: serverSideSessionId
    unique: true
```

To configure the callback, do:

```dart
pod.initializeAuthServices(
  tokenManagerBuilders: [
    ServerSideSessionsConfigFromPasswords(
      onSessionCreated:
          (
            final session, {
            required final authUserId,
            required final serverSideSessionId,
            required final transaction,
          }) async {
            await SessionMetadata.db.insertRow(
              session,
              SessionMetadata(
                serverSideSessionId: serverSideSessionId,
                deviceName: 'Test Device',
                ipAddress: session.request?.connectionInfo.remote.address.toString(),
                userAgent: 'TestAgent/1.0',
                metadata: null,
              ),
              transaction: transaction,
            );
          },
    ),
  ],
  identityProviderBuilders: [],
);
```

## Revoking tokens from custom metadata

To revoke tokens from custom metadata, simply filter the ID of the tokens to revoke from the metadata table and call the `revokeToken` method:

```dart
final tokenMetadata = await SessionMetadata.db.find(
  session,
  where: (final row) => row.deviceName.equals('Test Device'),
);

await AuthServices.instance.tokenManager.revokeToken(
  session,
  tokenId: tokenMetadata.first.serverSideSessionId.toString(),
);
```

## Pre-launch Checklist

## Breaking changes

None.


christerswahn (commented):
A question, how would an endpoint method implementation which creates a token (e.g. a PAT / CLI token), forward its contextual information to the callback? Since the callback is configured globally, how will it receive such information, e.g. a name that was passed as parameter to that specific endpoint method call?

marcelomendoncasoares:
> A question, how would an endpoint method implementation which creates a token (e.g. a PAT / CLI token), forward its contextual information to the callback? Since the callback is configured globally, how will it receive such information, e.g. a name that was passed as parameter to that specific endpoint method call?

@christerswahn It could be done using the `dynamic` `session.userObject` property. Below is the example for using the `userObejct`:

1. Configure the callback to extract the metadata from the user object:
    ```dart
    ServerSideSessionsConfig(
      sessionKeyHashPepper: 'test-pepper',
      onSessionCreated:
          (
            final session, {
            required final authUserId,
            required final serverSideSessionId,
            required final transaction,
          }) async {
            // Extract the metadata here
            final object = session.userObject as Map<String, dynamic>;
            final deviceName = object['deviceName'] as String;
            final userAgent = object['userAgent'] as String;
    
            await SessionMetadata.db.insertRow(
              session,
              SessionMetadata(
                serverSideSessionId: serverSideSessionId,
                deviceName: deviceName,
                userAgent: userAgent,
                metadata: null,
              ),
              transaction: transaction,
            );
          },
    ),
    ```

2. Then simply set the metadata before issuing the token:
    ```dart
    // Set the metadata here before issuing the token
    session.userObject = {
      'deviceName': 'Test Device',
      'userAgent': 'TestAgent/1.0',
    };
    
    await AuthServices.instance.tokenManager.issueToken(
      session,
      authUserId: authUserId,
      method: 'test',
      scopes: {},
    );
    ```
 
---

But, having said that, it is better to be able to do it in a type-safe way. So I have added two new getters to the `AuthSuccess` that you can use after issuing the token to attach the metadata (one for SSS, other for JWT):

```dart
final authSuccess = await AuthServices.instance.tokenManager.issueToken(
  session,
  authUserId: authUserId,
  method: 'test',
  scopes: {},
);

await SessionMetadata.db.insertRow(
  session,
  SessionMetadata(
    serverSideSessionId: authSuccess.serverSideSessionId,
    deviceName: 'Test Device',
    ipAddress: '192.168.1.1',
    userAgent: 'TestAgent/1.0',
  ),
);
```

Then you can use the parameters from your endpoint to add the metadata after the token is created.

christerswahn:
Ok thank you for the thorough answer!

Suggestion 1, add to the test case this passing of a data point via the session object and picking it up in the callback.

Suggestion 2, add a version of your good answer above to the auth docs.

marcelomendoncasoares:
> Ok thank you for the thorough answer!
> 
> Suggestion 1, add to the test case this passing of a data point via the session object and picking it up in the callback.
> 
> Suggestion 2, add a version of your good answer above to the auth docs.

You're welcome! Suggestion 1 is already covered at https://github.com/serverpod/serverpod/pull/4741/commits/de78418655c27cb407d5c84ca3d41106923f1337 and suggestion 2 will be done right away after merging.

christerswahn (approved):
Awesome work! LGTM!

Issue #3485, opened 2025-04-28
Not registering a FutureCall results in FutureCallEntry being silently discarded


inf0rmatix:
### Describe the bug

Not registering a FutureCall results in FutureCallEntry being silently discarded since it will be removed from the database but not executed without warning.

### To reproduce

- Schedule any FutureCall with a name not registered
- See the FutureCallEntry being removed from DB once it is due
- No error or warning is thrown or logged about the absence of a FutureCall matching the name

### Expected behavior

At least a warning should be logged when finding a non-resolvable FutureCall name

### Library version

2.6.0

### Platform information

All platforms

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

Crazelu:
This occurrence is now highly unlikely with the introduction of typed interface generation for future calls [(3418)](https://github.com/serverpod/serverpod/issues/3418).

However, it is still possible to experience this with the legacy approach to future calls.
My proposal would be to log a warning and recommend migrating to the typed interface.

Issue #4550, opened 2026-01-08
Lack of tools to prevent runtime errors due to changes in future calls


marcelomendoncasoares:
### Problem to Solve

Since future calls are stored references to be invoked later, they can lead to runtime errors in case future call classes or parameter models are changed in a non-backwards compatible way between scheduling and executing.

A few ways that a scheduled future call run can lead to a runtime error is if, before it runs, the server is restarted changing:

* The method of a future call spec class, leading to the removal of the previous generated future call execution class.
* The signature of a future call method that will lead to a generated model that fails to deserialized the stored JSON.
* The model that is used as parameter to a future call method, also leading to failure in the stored JSON deserialization.

Although ensuring backwards compatibility is responsibility of the developer, we should provide tools to facilitate the check that stored future calls are still valid before starting the server.

### Proposal

When starting the server, users should be warned that there are broken future calls by validating that:

1. All `FutureCall` classes referenced in the `FutureCallEntry` table are registered (see also #3485).
2. All strings stored as `serializedObject` for `FutureCallEntry` entries can be deserialized with success.

It should also be possible to run the check from maintenance mode to abort with error to bundle in CI pipelines.

### Use Case

Prevent starting the server or at least be promptly informed that existing future calls will break due to server changes before receiving runtime errors.

### Alternatives

Manually check the `FutureCallEntry` entries or leave runtime errors happen.

### Additional context

This type of error is now even more latent after #4375 that introduced generated typed-interfaces for future calls.

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

Crazelu:
@marcelomendoncasoares some questions to gather more clarity:

- If we must report broken future calls and abort in maintenance mode, is there a way for users to delete the broken future calls? Migrations? Because this will probably need to be documented.
- Should broken future calls only be reported in maintenance mode? If no, are we comfortable with the costs associated with scanning the future call table every time the server starts? Perhaps we want users to control this with a feature flag?

marcelomendoncasoares:
Great questions!

> * If we must report broken future calls and abort in maintenance mode, is there a way for users to delete the broken future calls? Migrations? Because this will probably need to be documented.

This is a topic for exploration in the current issue. Maybe a command-line flag to cleanup the broken ones and proceed?

[...]

PR #4744, opened 2026-02-18
chore: Adds warning when server tries to run unregistered future calls

Crazelu:
This PR adds a warning when the server tries to run an unregistered future call. The warning also recommends migrating to the typed interface for future calls since this [issue](https://github.com/serverpod/serverpod/issues/3485) is more likely with the legacy approach where manual registration of future calls is required.

Fixes [3485](https://github.com/serverpod/serverpod/issues/3485)

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
+++ b/packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
@@ -200,15 +200,11 @@ class FutureCallManager {
 
       if (futureCall == null) {
         _internalSession.log(
-          'Failed to run unregistered FutureCall. '
-          'Likely causes include: '
-          '1. Existing FutureCall method code was updated causing a name change '
-          'in the generated code and when server restarted, '
-          'the FutureCall was registered with a different name. '
-          '2. If you are using the legacy approach, '
-          'you may have forgotten to register the FutureCall. '
-          'Consider migrating to the typed interface for future calls in this case. '
-          '$entry',
+          'Attempted to run a FutureCall that was not registered. This is likely due '
+          'to changing a FutureCall method after it was scheduled, leading to an '
+          'entry that no longer has a matching method. For legacy future calls, '
+          'make sure they are registered in the server start. Entry: $entry',
+          level: LogLevel.error,
         );
         return null;
       }
--- a/tests/serverpod_test_server/test_integration/future_calls/future_call_manager_test.dart
+++ b/tests/serverpod_test_server/test_integration/future_calls/future_call_manager_test.dart
@@ -741,18 +741,8 @@ void main() async {
 
       test(
         'when executing all scheduled FutureCalls '
-        'then a warning is logged for the unregistered FutureCall',
+        'then a message is logged for the unregistered FutureCall with error level',
         () async {
-          final expectedMessage =
-              'Failed to run unregistered FutureCall. '
-              'Likely causes include: '
-              '1. Existing FutureCall method code was updated causing a name change '
-              'in the generated code and when server restarted, '
-              'the FutureCall was registered with a different name. '
-              '2. If you are using the legacy approach, '
-              'you may have forgotten to register the FutureCall. '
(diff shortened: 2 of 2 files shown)


packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -199,8 +199,11 @@ class FutureCallManager {
       final futureCall = _futureCalls[entry.name];
 
       if (futureCall == null) {
-        // ISSUE(https://github.com/serverpod/serverpod/issues/3485):
-        // This should be logged or caught otherwise.
+        stdout.writeln(
+          'Failed to execute future call (${entry.name}) which was scheduled but not registered. '
+          'This is likely due to using the legacy approach. '
+          'Consider migrating to the typed interface for future calls.',
+        );
```
marcelomendoncasoares, line 208: Discussion: On the issue, you have mentioned:

> This occurrence is now highly unlikely with the introduction of typed interface generation for future calls.
> 
> However, it is still possible to experience this with the legacy approach to future calls. My proposal would be to log a warning and recommend migrating to the typed interface.

But what if a future call is registered to a future time and the code for it is changed before the execution, leading to the future call not being registered when the execution comes? This is one of the cases I listed at https://github.com/serverpod/serverpod/issues/4550 and would still be possible with the code generation.

It does not change our possibilities for this fix, which is only to log the runtime failure, but might change the message we present.

It would also be interesting to log the future call entry in detail and as a persistent log (using `session.log`) instead of just a print. Then, we can add a test for this to ensure the correct message, since it is part of the UX.

marcelomendoncasoares (changes requested):
Just one topic for discussion and coverage on the tests.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: I have updated the message and also used `session.log`.
I am not very sure about the warning message I currently have. Specifically, I do not know if we want to include likely causes in that warning.
What do you think?

marcelomendoncasoares: I think having the cause suggestion is good, maybe just a little bit more resumed, like:

```
Attempted to run a FutureCall that was not registered. This is likely due
to changing a FutureCall method after it was scheduled, leading to an
entry that no longer has a matching method. For legacy future calls,
make sure they are registered in the server start. Entry: $entry
```

Then we should also log it as error level.

tests/serverpod_test_server/test_integration/future_calls/future_call_manager_test.dart
```diff
@@ -714,4 +717,70 @@ void main() async {
...
+        () async {
+          final expectedMessage =
+              'Failed to run unregistered FutureCall. '
+              'Likely causes include: '
+              '1. Existing FutureCall method code was updated causing a name change '
+              'in the generated code and when server restarted, '
+              'the FutureCall was registered with a different name. '
+              '2. If you are using the legacy approach, '
+              'you may have forgotten to register the FutureCall. '
+              'Consider migrating to the typed interface for future calls in this case. ';
```
marcelomendoncasoares: Suggestion: Move to the `expect` and use a complete equality check instead of `contains`.

Crazelu: Resolved in [0e4a589](https://github.com/serverpod/serverpod/pull/4744/commits/0e4a5897311f7c48c6b6b22de6c1157ebdabdded)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Resolved in [0e4a589](https://github.com/serverpod/serverpod/pull/4744/commits/0e4a5897311f7c48c6b6b22de6c1157ebdabdded)

Issue #4604, opened 2026-01-16
RFC: New migrations experience


marcelomendoncasoares:
<img width="2400" height="1792" alt="Image" src="https://github.com/user-attachments/assets/fedbd97e-17f1-47c0-99e3-d67e55c05fbc" />

## Summary

One of the core routines of the development experience when using Serverpod is creating and applying migrations to the database. It is also where users face most common conflicts, like local state that diverges from the database and merge conflicts on migrations of different branches. Our tooling in this area is currently very limited and pushes a lot of manual work to developers when such conflicts occur.

The purpose of this RFC is to outline a new set of tools to provide a smooth developer experience when dealing with migrations.

## Current pain points

- Merge conflicts are frequent and require the developer to manually delete everything.  
- Empty folders left by discarding git changes break the create command.  
- Applying migrations requires starting the server.  
- Too many generated SQL files that bloat the repository.  
- Migration folders pile over time, and cleaning up is manual.  
- Lack of a way to abort a pipeline with an error if a migration is missing on a given project.  
- The command to apply migrations is not easily discoverable from the tools for newcomers.  
- Rolling migrations back is done by creating a repair migration, which is not intuitive.  
- No easy way to inspect which migrations were applied to the database (and when).

\> Not all of the above pain points will be addressed with the effort we’re going to dedicate now, but this RFC will also serve to collect feedback and suggestions on how to improve every aspect regarding migrations on Serverpod.

## Out of scope

The current RFC focuses on the experience when interacting with migrations from the user perspective, but does not aim to change how they are executed when applying. This means that changing the definition files, the migration registry, or the migration SQL generator is out of scope. It is open for improvement on when to generate and how to store them, but without changing their formats.

For reference, a few pain points that are not covered in this RFC are:

- Hooking custom transformation into migrations requires manually editing generated SQL files.  
- No way to attach custom Dart scripts/logic to run before/after specific migrations.  
- Migrations being registered as a linear history creates frequent conflicts on branch merges.

## Proposal

To provide a complete \- and centralized \- experience with migrations, we could introduce a new `migration` subcommand on the Serverpod CLI that would contain all commands related to this area. The goal would be to create, apply, rollback, and squash migrations as a part of the same experience. Discovery also improves a lot, since users can always call `--help` and see all available commands.

### Generating SQL

[...]

Issue #4564, opened 2026-01-09
Tech spike: Evaluate the best way to decouple the CLI from the strict Serverpod version


marcelomendoncasoares:
### Problem to Solve

The CLI version is currently tied to the Serverpod version due to dependencies for the code generation. This creates some frequent problems:

- Developers need to reinstall the CLI when switching to work on projects with different Serverpod versions.
- We are constrained to only release breaking changes to the CLI with major Serverpod upgrades.
- Users need to upgrade their projects just to receive the improvements to the CLI.

The second point is specially painful now that we intend to improve the UX of using Serverpod - which largely resumes to the interaction with the CLI.

### Proposal

If we move the generator part of the CLI into a `serverpod_generator` package, we could have this package depend on the `serverpod` version and have the CLI compatible with a range of the `serverpod_generator` (according to interface changes, just as with any normal packages).

The `generate` command (and others that depend on Serverpod) would delegate the execution to the `serverpod_generator` that is installed in the project environment - if compatible with the current CLI support range.

### Use Case

1. Allow users to run `serverpod` commands on projects with different package versions.
2. Allow releasing UX improvements in a faster pace without pushing the need for users to upgrade their projects.

### Alternatives

Continue as it is.

### Additional context

This issue should evaluate the feasibility of such separation and which problems we might run into during the implementation.

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

marcelomendoncasoares:
The investigation highlighted a great complexity on the complete separation that can be made in steps. The first one would be to implement the cached compiled binaries to dispatch the execution of commands matching the version of the current project - a similar strategy to `melos`, that uses a project-local binary instead of the global version.

Such strategy will achieve mostly the same results as the complete separation of the CLI internals into another package, since current changes to the CLI are still a lot coupled to the other packages. And the greatest pain point would be solved: using Serverpod on projects with different versions without having to reinstall the CLI every time.

PR #4748, opened 2026-02-19
refactor: Decouple the database implementation from the Serverpod package

### LINKED ISSUE #4721: Tech spike: Evaluate the changes needed to adapt Serverpod to support SQLite and other databases
### Problem to Solve

Serverpod currently rely only on Postgres, but support for other databases has been extensively requested in the past (specially SQLite and MySQL/MariaDB, as shown in below references). Internally, Serverpod is mostly structured to support other databases, with several database concepts abstracted in `serverpod/lib/src/database/concepts`. But some parts of the implementation ended up scattered through the codebase, making it hard for contributors unfamiliar with the codebase to add another engine.

This issue aims to document the changes needed to support other database engines and serve as a guide for future implementations or community contributions.

### Proposal

Analyze the Serverpod codebase to map the changes needed to implement other database engines. The main packages involved are:

- `packages/serverpod`: Implementation of the database concepts, adapter, connection and analyzer.
- `packages/serverpod_shared`: Definition of the database config that is specified on the `config/*.yaml` files.
- `tools/serverpod_cli`: Dialect-specific SQL generation for migrations.

This issue should result in a guide on the changes required to add a new database engine.

### Use Case

Allow extension of Serverpod to support other database engines in the future.

### Alternatives

Continue using only Postgres.

### Additional context

This work begun as part of https://github.com/serverpod/serverpod/issues/4564. Since the decoupling between the Serverpod version and CLI will require moving part of the generator from the CLI to another package, this should be done in a way that gets the codebase closer to support other engines - instead of making it even harder.

Related issues and discussions:

- https://github.com/serverpod/serverpod/issues/201
- https://github.com/serverpod/serverpod/discussions/178
- https://github.com/serverpod/serverpod/discussions/612
- https://github.com/serverpod/serverpod/discussions/1280
- https://github.com/serverpod/serverpod/discussions/2084

None

- [x] I want to work on this

marcelomendoncasoares:
This PR is part of the #4721 tech spike and serves two purposes:

1. Allow an easy implementation of new database adapters.
2. Prepare the database layer for being separated from the Serverpod package.

The second purpose is part of #4564 and the #4604 RFC since we'll need access to the database from the CLI to run migrations and we can't push the `serverpod` package as a CLI dependency if we want to decouple them.

**One core premise of this refactor is that Serverpod will always be Postgres-first**. Which means that our interfaces are allowed to contain methods/properties that are not supported by all databases - like the `lockRows` method for row-level locking or certain runtime parameters. Specific methods/properties for features that Postgres does not support are also possible in the future.

## Core changes

The changes in this PR can be grouped into four main categories:

1. Update to the `DatabaseConfig` to support new dialects.
2. Centralization of the SQL generation (DDL) for migrations in the `serverpod_cli`.
3. Centralization of the database adapter in the `serverpod` package.
4. Decoupling from all server-specific logic inside `serverpod/src/database`.

The commits follow this order to facilitate the review process, since a lot of the code is being moved around and it's easier to review the changes in small chunks.

### Implementing new database adapters

The central piece to facilitate the implementation of new database adapters is the `DatabaseDialect` class. This is a simple enum that contains all the supported dialects. By adding a new entry to this enum, all the places that need to be updated to support the new adapter will be highlighted. This change closes #4721.

### Update to the `DatabaseConfig`

The ideal way to implement this would be to have a base `DatabaseConfig` class and implement subclasses for each dialect. However, this would be a breaking change, since this class is exposed in the public API of the `serverpod_shared` package and is part of the `ServerpodConfig` class.

Instead, this PR adds only an optional `dialect` parameter to the `DatabaseConfig` constructor that is not exposed in the public API and will be set according to the specific database configuration being parsed from the configuration file.

For the `serverpod_cli` to be able to generate the correct SQL for the chosen dialect, a new `databaseDialect` key was added to the `generator.yaml` file that can accept any entry from the `DatabaseDialect` enum.

### Centralization of the SQL generation for migrations

Inside the `serverpod_cli` package, all SQL generation is under a `src/database/dialects` folder with a central `SqlGenerator` interface that must be implemented by each dialect. This interface is also a factory that returns the correct implementation for the chosen dialect.

### Centralization of the database adapter in the `serverpod` package

Despite already having a `src/database/adapters` folder in the `serverpod` package, a lot of the Postgres-specific logic was scattered across the package. Now, it is centralized in the `src/database/adapters/postgres` folder and completely abstracted by interfaces declared in the `src/database/interface` folder.

A new `DatabaseProvider` interface was created to abstract the creation of the database elements (pool manager, connection, analyzer, migration runner). This interface is also a factory that returns the correct implementation for the chosen dialect.

### Decoupling from all server-specific logic inside `serverpod/src/database`

As a last step, all database interface classes were completely decoupled from the rest of the package by abstracting a few pieces:

- The database methods now receive a `DatabaseSession` instead of a `Session` to allow the database layer to be used without a Serverpod instance. This is a replacement for the previous `DatabaseAccessor` with very few changes. **The generated repository methods were not changed** and still depend on the `Session` class, since this would be a major breaking change.
- The `SerializationManagerServer` was moved to the `src/database/interface` folder and is now a dependency of the `DatabaseProvider` interface. Since it only contains database-related getter methods, it makes sense to declare it as parte of the database implementation.
- A new `ValueEncoder` interface was created to abstract the encoding of values for the database. This interface is also a singleton that returns the encoder for the current database pool. This is a replacement for the previous access to the Postgres-specific encoder as a static attribute.

## Regarding testing

Few tests were needed for this PR since most of the changes are internal. The changes to public interfaces (such as the `DatabaseConfig` and the `ValueEncoder`) classes were covered by new tests. All of the rest is covered by existing tests.

## Next steps

The only dependency on the `serverpod` package left on the database layer is due to the database-related models (`TableDefinition`, `ColumnDefinition`, `DatabaseMigration`, etc.). Now that we can have [packages with shared models](https://github.com/serverpod/serverpod/pull/4737), such models will be moved to a shared `serverpod_database` - together with all files from `serverpod/lib/src/database`. This package will depend only on the `serverpod_serialization` package, making the database safe to be used on the CLI and even in the client.

## Pre-launch Checklist

## Breaking changes

None, since all the changes are in the internal implementation of the database layer.

### DIFF
--- a/packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
+++ b/packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
@@ -40,14 +40,11 @@ class PostgresDatabaseConnection
   /// to access the database.
   PostgresDatabaseConnection(super.poolManager);
 
-  /// Tests the database connection.
-  /// Throws an exception if the connection is not working.
   @override
   Future<bool> testConnection() async {
     return poolManager.testConnection();
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<List<T>> find<T extends TableRow>(
     DatabaseSession session, {
@@ -85,7 +82,6 @@ class PostgresDatabaseConnection
     );
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<T?> findFirstRow<T extends TableRow>(
     DatabaseSession session, {
@@ -119,7 +115,6 @@ class PostgresDatabaseConnection
     return rows.first;
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<T?> findById<T extends TableRow>(
     DatabaseSession session,
@@ -140,8 +135,6 @@ class PostgresDatabaseConnection
     );
   }
 
-  /// Acquires row-level locks without returning row data.
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<void> lockRows<T extends TableRow>(
     DatabaseSession session, {
@@ -166,7 +159,6 @@ class PostgresDatabaseConnection
     );
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<List<T>> insert<T extends TableRow>(
     DatabaseSession session,
@@ -191,7 +183,6 @@ class PostgresDatabaseConnection
         .toList();
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<T> insertRow<T extends TableRow>(
     DatabaseSession session,
@@ -213,7 +204,6 @@ class PostgresDatabaseConnection
     return result.first;
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<List<T>> update<T extends TableRow>(
     DatabaseSession session,
@@ -262,7 +252,6 @@ class PostgresDatabaseConnection
         .toList();
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<T> updateRow<T extends TableRow>(
     DatabaseSession session,
@@ -286,12 +275,6 @@ class PostgresDatabaseConnection
     return updated.first;
   }
 
-  /// Updates a single row by its ID with the specified column values.
-  ///
-  /// Returns the updated row or null if no row with the given ID exists.
-  /// Throws [ArgumentError] if [columnValues] is empty.
-  ///
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<T> updateById<T extends TableRow>(
     DatabaseSession session,
@@ -334,15 +317,6 @@ class PostgresDatabaseConnection
     );
   }
 
-  /// Updates all rows matching the WHERE expression with the specified column values.
-  ///
-  /// Returns a list of all updated rows. Returns an empty list if no rows match.
-  /// Throws [ArgumentError] if [columnValues] is empty.
-  ///
-  /// When [limit], [offset], [orderBy], [orderByList], or [orderDescending] are provided,
-  /// only the rows selected by these parameters will be updated.
-  ///
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<List<T>> updateWhere<T extends TableRow>(
     DatabaseSession session, {
@@ -419,7 +393,6 @@ class PostgresDatabaseConnection
     return result.map(poolManager.serializationManager.deserialize<T>).toList();
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<List<T>> delete<T extends TableRow>(
     DatabaseSession session,
@@ -440,7 +413,6 @@ class PostgresDatabaseConnection
     );
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<T> deleteRow<T extends TableRow>(
     DatabaseSession session,
@@ -462,7 +434,6 @@ class PostgresDatabaseConnection
     return result.first;
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<List<T>> deleteWhere<T extends TableRow>(
     DatabaseSession session,
@@ -483,7 +454,6 @@ class PostgresDatabaseConnection
     );
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<int> count<T extends TableRow>(
     DatabaseSession session, {
@@ -511,7 +481,6 @@ class PostgresDatabaseConnection
     return rows.first;
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<PostgresDatabaseResult> simpleQuery(
     DatabaseSession session,
@@ -530,7 +499,6 @@ class PostgresDatabaseConnection
     return PostgresDatabaseResult(result);
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<PostgresDatabaseResult> query(
     DatabaseSession session,
@@ -624,7 +592,6 @@ class PostgresDatabaseConnection
     }
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<int> execute(
     DatabaseSession session,
@@ -645,7 +612,6 @@ class PostgresDatabaseConnection
     return result.affectedRows;
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<int> simpleExecute(
     DatabaseSession session,
@@ -665,7 +631,6 @@ class PostgresDatabaseConnection
     return result.affectedRows;
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   Future<Iterable<Map<String, dynamic>>> _mappedResultsQuery(
     DatabaseSession session,
     String query, {
@@ -755,7 +720,6 @@ class PostgresDatabaseConnection
     );
   }
 
-  /// For most cases use the corresponding method in [Database] instead.
   @override
   Future<R> transaction<R>(
     TransactionFunction<R> transactionFunction, {
--- a/packages/serverpod/lib/src/database/adapters/postgres/postgres_pool_manager.dart
+++ b/packages/serverpod/lib/src/database/adapters/postgres/postgres_pool_manager.dart
@@ -11,12 +11,9 @@ import 'value_encoder.dart';
 /// Configuration for connecting to the Postgresql database.
 @internal
 class PostgresPoolManager implements DatabasePoolManager {
-  /// The dialect of the database pool manager.
   @override
   DatabaseDialect get dialect => DatabaseDialect.postgres;
 
-  /// The last time a database operation was performed. This can be used to
-  /// determine if the database is sleeping.
   @override
   DateTime? lastDatabaseOperationTime;
 
@@ -25,7 +22,6 @@ class PostgresPoolManager implements DatabasePoolManager {
 
   late SerializationManagerServer _serializationManager;
 
-  /// Access to the serialization manager.
   @override
   SerializationManagerServer get serializationManager => _serializationManager;
 
@@ -45,7 +41,6 @@ class PostgresPoolManager implements DatabasePoolManager {
     return pgPool;
   }
 
-  /// The encoder used to encode objects for storing in the database.
   @override
   PostgresValueEncoder get encoder => PostgresValueEncoder();
 
@@ -85,7 +80,6 @@ class PostgresPoolManager implements DatabasePoolManager {
     _serializationManager = serializationManager;
   }
 
-  /// Starts the database connection pool.
   @override
   void start() {
     // Setup database connection pool
@@ -104,14 +98,12 @@ class PostgresPoolManager implements DatabasePoolManager {
     );
   }
 
-  /// Closes the database connection pool.
   @override
   Future<void> stop() async {
     await _pgPool?.close();
     _pgPool = null;
   }
 
-  /// Tests the database connection.
   @override
   Future<bool> testConnection() async {
     await pool.execute(


packages/serverpod_shared/lib/src/config.dart
```diff
@@ -477,6 +486,7 @@ class DatabaseConfig {
     this.isUnixSocket = false,
     this.searchPaths,
     this.maxConnectionCount = defaultMaxConnectionCount,
+    this.dialect = DatabaseDialect.postgres,
   });
```
nielsenko: This could be a factory ctor returning subclasses of `DatabaseConfig` by `dialect`. This would not be a breaking change.

packages/serverpod_shared/lib/src/config.dart
nielsenko, line 609: should this not read `dialect` from then json?

nielsenko: Add `dialect`?

packages/serverpod/lib/src/database/adapters/postgres/sql_query_builder.dart
```diff
@@ -243,7 +242,7 @@ class SelectQueryBuilder {
 
     var relationFieldName = tableRelation.foreignFieldBaseQuery;
 
-    var strIds = ids.map(DatabasePoolManager.encoder.convert).join(', ');
+    var strIds = ids.map(ValueEncoder.instance.convert).join(', ');
```
nielsenko, line 245: Would it be better to have access to the `DatabasePoolManager` instance here, so it could be `_poolManager.encoder.convert` instead?

Same elsewhere where we use `ValueEncoder.instance`.

packages/serverpod/lib/src/database/interface/value_encoder.dart
```diff
@@ -0,0 +1,32 @@
...
+    }
+    return _instance!;
+  }
+
+  static ValueEncoder? _instance;
+
+  /// Sets the singleton [instance] to the given [encoder].
+  static void set(ValueEncoder encoder) {
+    _instance = encoder;
+  }
```
nielsenko, line 20: Should we avoid this singleton?

tests/serverpod_test_server/test_integration/database_operations/value_encoder_set_test.dart
```diff
@@ -0,0 +1,99 @@
+import 'package:serverpod/serverpod.dart';
+import 'package:serverpod/src/generated/protocol.dart';
+import 'package:test/test.dart';
+
+void main() {
```
nielsenko, line 5: All tests basically become redundant if we avoid the singleton pattern..

Also, these tests are groups with single test calls.. why not just do setup inside the test itself then?

packages/serverpod/lib/src/database/migrations/table_comparison_warning.dart
```diff
@@ -100,14 +100,14 @@ abstract class ComparisonWarning<T extends ComparisonWarning<T>> {
     } else if (isMismatch) {
       // If this is the top-level mismatch, provide context about the mismatch.
       if (parent == null) {
-        buffer.write('$type "$name" mismatch: ');
+        buffer.write('$type "$name" mismatch:');
```
nielsenko, line 103: Is this intended?

packages/serverpod/test/database/columns/column_bigint_test.dart
```diff
@@ -2,6 +2,8 @@ import 'package:serverpod/database.dart';
 import 'package:test/test.dart';
 
 void main() {
+  ValueEncoder.set(PostgresValueEncoder());
```
nielsenko, line 5: I don't like this `ValueEncoder` singleton pattern.. But I'll stop commenting more

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -771,6 +771,39 @@ class DatabaseConnection {
...
+  /// Migrations on Postgres use a transaction to ensure that the advisory lock
+  /// is retained until the transaction is completed.
+  ///
+  /// The transaction ensures that the session used for acquiring the lock is
+  /// kept alive in the underlying connection pool, and that we can later use
+  /// that exact same session for releasing the lock. The transaction is thus
+  /// only used to get the desired behavior from the database driver, and does
+  /// not have any effect on the Postgres level.
+  ///
+  /// This ensures that we are only running migrations one at a time.
```
nielsenko, line 761: We already discussed this in meeting, but we should run the entire migration under the transaction.

I know this is old inherited code, and even the addition of the advisory lock is old, but we should not let this linger.

packages/serverpod/lib/src/database/adapters/postgres/postgres_analyzer.dart
```diff
@@ -1,385 +1,13 @@
...
-
-import '../util/column_type_extension.dart';
+import 'package:serverpod/database.dart';
+import 'package:serverpod/src/database/adapters/postgres/postgres_analyzer.dart';
 
 /// Analyzes the structure of [Database]s.
 class DatabaseAnalyzer {
   /// Analyze the structure of the [database].
   static Future<DatabaseDefinition> analyze(Database database) async {
-    var currentDb = await database.unsafeQuery('SELECT current_database();');
```
nielsenko, line 13: Should we get rid of this now. It simply creates postgress specific analyzer, and call analyze. I think we should bite the bullet and thread an abtract database analyzer through. This seems a bit half cooked.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -21,37 +23,32 @@ import 'package:uuid/uuid.dart';
...
   Future<bool> testConnection() async {
-    await _postgresConnection.execute(
-      'SELECT 1;',
-      timeout: const Duration(seconds: 2),
-    );
-    return true;
+    return poolManager.testConnection();
   }
 
   /// For most cases use the corresponding method in [Database] instead.
```
nielsenko: Move comment to base. It will be inherited. This seems to be a common issue.

packages/serverpod/lib/src/database/interface/provider.dart
```diff
@@ -0,0 +1,34 @@
...
+
+/// Abstract interface for database providers.
+/// Provides a unified interface for any database dialect implementation.
+@internal
+abstract interface class DatabaseProvider {
+  /// Creates a new [DatabaseProvider] for the given [dialect].
+  factory DatabaseProvider.forDialect(DatabaseDialect dialect) =>
+      switch (dialect) {
+        DatabaseDialect.postgres => PostgresDatabaseProvider(),
+      };
```
nielsenko, line 21: So here you are using the pattern already.. As mentioned I think we should use the same for `DatabaseConfig`.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -20,24 +20,24 @@ import 'package:serverpod/src/database/postgres_error_codes.dart';
...
+import '../../interface/database_session.dart';
 import '../../concepts/expressions.dart';
 import '../../concepts/table.dart';
 import '../../query_parameters.dart';
 
 part 'postgres_exceptions.dart';
 
 /// A connection to the database. In most cases the [Database] db object in
-/// the [Session] object should be used when connecting with the database.
+/// the [DatabaseSession] object should be used when connecting with the database.
```
nielsenko, line 31: Is `DatabaseSession` the right name for this?

packages/serverpod/lib/src/database/database_accessor.dart
```diff
@@ -1,10 +0,0 @@
-import 'package:serverpod/serverpod.dart';
-
-/// Interface for accessing the database.
-abstract interface class DatabaseAccessor {
```
nielsenko, line 4: Why not just use this instead of `DatabaseSession`? Not sure I follow?

marcelomendoncasoares: Agree that `DatabaseSession` is not the best name, but I don't think `DatabaseAccessor` is also right since it carries an optional `transaction` parameter - which is somewhat scoped, as a "session". The `logQuery` parameter is also bound to a session because the queries are logged as part of a same session. And the name of the parameter is `session` everywhere in the codebase, which would be inconsistent if we kept the old name (e.g. `DatabaseAccessor session`) or would be a massive rewrite. Even in this case, it would create some inconsistencies due to the impossibility of changing the name of the `session` parameter in the generated repository classes, leading to a `session` being passed to an `accessor` parameter.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
marcelomendoncasoares: Justified on the comment below.

packages/serverpod/lib/src/database/interface/provider.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4748/commits/b9b968cf7e567e85dbf41730decefc9e2a72bf97.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4748/commits/8594e189342b9534f199d0f3e0a6170d5f96e31e.

packages/serverpod/lib/src/database/migrations/table_comparison_warning.dart
marcelomendoncasoares: Yes! This trailing space was always followed by a newline, which created strings on the tests with a trailing space that the editor automatically stripes when saving the file. The changes on https://github.com/serverpod/serverpod/pull/4748/commits/3cb5cd02099ae9c6ec0afe45d7257ea97e67b512 led to this removal, so I fixed it by removing the trailing space altogether.

packages/serverpod_shared/lib/src/config.dart
marcelomendoncasoares: I thought about this ideia, but ended up not implementing due to the impossibility of removing Postgres-specific parameters from the base class. But I agree that having the subclasses poses us closer to the ideal state, enabling us to eventually deprecate the constructor of the base class in favor of the subclasses directly. Done at https://github.com/serverpod/serverpod/pull/4748/commits/b9b968cf7e567e85dbf41730decefc9e2a72bf97.

packages/serverpod_shared/lib/src/config.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/4748/commits/b9b968cf7e567e85dbf41730decefc9e2a72bf97.

marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/4748/commits/b9b968cf7e567e85dbf41730decefc9e2a72bf97.

packages/serverpod/test/database/columns/column_bigint_test.dart
marcelomendoncasoares: I also dislike and tried to remove it, but it is a debt that we can't easily fix. Since the encoder was accessed as the static method `DatabasePoolManager.encode`, it is used in several places where we can't easily reach to inject the pool manager - like inside the `toString` methods of the `Expression` classes. The singleton was the best way I found to preserve the current structure while allowing the encoder to be bound to the database.

packages/serverpod/lib/src/database/adapters/postgres/postgres_analyzer.dart
marcelomendoncasoares: Agreed. I kept the facade to avoid exposing the `DatabaseAnalyzer`, but it became too thin to be justified. Fixed at https://github.com/serverpod/serverpod/pull/4748/commits/501c8b3fecc21a6076857f5470665966be447cee by exposing the abstract `DatabaseAnalyzer` as the late variable `Database.analyzer`. This way we also expose this feature to users building more advanced use cases that might require database introspection.

packages/serverpod/lib/src/database/adapters/postgres/sql_query_builder.dart
marcelomendoncasoares: Since we can't get rid of the singleton (as justified below on https://github.com/serverpod/serverpod/pull/4748#discussion_r2842568186), I preferred to keep the same access pattern and avoid pushing extra dependencies on the tests - like initializing a database pool manager on all SQL query builder tests (which are many).

packages/serverpod/lib/src/database/interface/value_encoder.dart
marcelomendoncasoares: Justified below on https://github.com/serverpod/serverpod/pull/4748#discussion_r2842568186.

tests/serverpod_test_server/test_integration/database_operations/value_encoder_set_test.dart
marcelomendoncasoares: The purpose of the separation on the groups is to have a proper `setUp` and `tearDown` (in case any failure occurs) and to make the test itself easily readable. By separating the setup in the group together with the "Given", it gets clear to read what is setup and what is the actual acting/result.

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
marcelomendoncasoares: Agreed, although I think we should do it separately, since it might require changes on tests that will be unrelated to the refactor. @coderabbitai, please spin up an issue to ensure migrations run inside a transaction on Postgres.

nielsenko: @oderabbitai Try creating the issue again.

nielsenko: 🙄 @coderabbitai create the requested issue

PR #4752, opened 2026-02-19
feat: Add account merging mechanics to auth module

craiglabenz:
Adds `AuthUser` merging logic to be called in certain scenarios while linking new Idps to users. When to call this code in that linking process is outside the scope of this PR.

The main entry point of this PR's changes is in [AuthUsers.merge], which makes use of a new [UserMergeConfig] object to handle the complicated tasks of merging two accounts.

The PR provides two merging functions to get users started: `defaultIdpMergeHook` which merges all Idp `<Provider>Account` records and `mergeCoreServerpodData` to, well, merge core Serverpod data.

Developers are able to use these methods and provide their own additional method for application-specific logic. They can also opt out of Serverpod's default account merging offerings and handle everything themselves (though that would likely imply some necessary changes in the default offerings).

* [Note]: To power the `defaultIdpMergeHook` method, this PR also adds `migrate` methods to each Idp class which knows how to move a record from one user to another.

Contributes to #4379

## Pre-launch Checklist

## Breaking changes

No breaking changes.


modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/business/auth_users.dart
```diff
@@ -179,6 +180,169 @@ class AuthUsers {
...
+      final userToKeepEntity = users[0];
+      final userToRemoveEntity = users[1];
+
+      if (userToKeepEntity == null || userToRemoveEntity == null) {
+        throw AuthUserNotFoundException();
+      }
+
+      // Upgrade to non-nullable types.
+      AuthUserModel userToKeep = userToKeepEntity.toModel();
+      final userToRemove = userToRemoveEntity.toModel();
```
marcelomendoncasoares: Suggestion(NIT): Small suggestion to follow the same pattern.

```suggestion
      var userToKeep = userToKeepEntity.toModel();
      final userToRemove = userToRemoveEntity.toModel();
```

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/business/auth_users_config.dart
```diff
@@ -10,13 +13,58 @@ class AuthUsersConfig {
...
   /// Called when an auth user is about to be created.
   final BeforeAuthUserCreatedHandler? onBeforeAuthUserCreated;
 
   /// Called when an auth user has been created.
   final AfterAuthUserCreatedHandler? onAfterAuthUserCreated;
+
+  /// Configuration for merging two auth users.
+  // TODO (craiglabenz): Should this be a nested field here or should it only be
+  // a parameter to initializeAuthServices?
+  final UserMergeConfig? userMergeConfig;
```
marcelomendoncasoares: Discussion: I think we can go with a detached parameter in the `AuthServices` (and the `initializeAuthServices`). Since it also specifies merge of `UserProfiles` and other data, it has a broader scope that just `AuthUsers`. This will also contribute to a cleaner addition, as you wrote in other places.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/business/auth_users_config.dart
```diff
@@ -10,13 +13,58 @@ class AuthUsersConfig {
...
+/// linked to a different [AuthUser]. Methods registered here will then be
+/// invoked in [AuthUsers.merge].
+class UserMergeConfig {
+  /// Creates a new [UserMergeConfig] instance.
+  const UserMergeConfig({
+    required this.mergeHandler,
+    this.mergeHooks = const [],
+    this.shouldDeleteUser = false,
+    this.shouldMergeCoreData = true,
+  });
```
marcelomendoncasoares: Discussion: A few considerations about this class:

1. Instead of being a required parameter, could we make the `mergeHandler` throw by default? This would allow the config to be always present (just as the others in the `AuthServices` class) while preventing unplanned merges that could result in loss of data.
2. Should this class expose a method to allow adding merge hooks officially? It would then be usable by our IDPs to add their specific merges instead of having a single `defaultIdpMergeHook` with all the providers. The hook could be done inside the initialization of the provider.
3. Should `mergeHooks` be `UserMergeHandler` or a different type that returns `Future<void>`? At a first glance as a user, I would expect to have one callback that merges the `AuthUser`/`UserProfile` and other merge hooks that are only going to propagate the merge on specific domains. Returning the auth user on each hook opens the possibility of one hook returning the new and the other the old, which would produce a lot of inconsistencies.
4. I'd suggest having `shouldDeleteUser` as true by default. Because the expected result of a merge of 2 accounts is having only 1 left, and not having to manually cleanup afterwards. The option would be more for custom merge implementations that require preserving the old account.
5. Instead of a boolean `shouldMergeCoreData`, why not have it as a separate merge handler in the config? This makes the logic extensible/replaceable and very easy for users to inspect - since the function will be there in the signature and not be some behavior deep inside the implementation changing with a top-level flag.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/business/auth_users_config.dart
```diff
@@ -10,13 +13,58 @@ class AuthUsersConfig {
...
+
+  /// Callbacks to merge data from additional modules or custom tables.
+  ///
+  /// This is where you should implement logic to move data from IDP-specific
+  /// tables (e.g. `EmailAuth`, `GoogleAuth`) or your own app's tables from
+  /// the [userToRemove] to the [userToKeep].
+  ///
+  /// Third-party Serverpod authors should consider writing a [UserMergeHandler]
+  /// for their module which developers can include in this list.
+  final List<UserMergeHandler> mergeHooks;
```
marcelomendoncasoares: Suggestion: Explain the order of execution in the doc comments, so that users can easily understand when each callback will be invoked during the merge process.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/util/auth_user_scopes_extension.dart
```diff
@@ -8,4 +8,12 @@ extension AuthUserScopes on AuthUserModel {
...
     return scopeNames.map(Scope.new).toSet();
   }
+
+  /// Converts this [AuthUserModel] to an [AuthUser] entity.
+  AuthUser toEntity() => AuthUser(
+    id: id,
+    blocked: blocked,
+    scopeNames: scopeNames,
+    createdAt: createdAt,
+  );
```
marcelomendoncasoares: Suggestion: Since this is used only inside the `auth_users.dart` file, it could be declared there as a private extension. Ideally, we don't want users to be manipulating `AuthUser` objects directly since it bypasses the user management hooks that the official methods of the `AuthUsers` class implements.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/business/auth_services.dart
```diff
@@ -157,13 +157,25 @@ extension AuthServicesInit on Serverpod {
...
   }) {
+    if (authUsersConfig != null && userMergeConfig != null) {
+      // TODO: This goes away if we don't nest UserMergeConfig as a field on
+      // AuthUsersConfig.
+      throw Exception(
+        'Cannot provide both authUsersConfig and userMergeConfig. '
+        'userMergeConfig is a part of authUsersConfig and is exposed only as a '
+        'convenience parameter to be used with the default AuthUsersConfig.',
+      );
+    }
```
marcelomendoncasoares: Comment: By removing the `UserMergeConfig` from the `AuthUsersConfig`, we avoid this inconvenience and are able to follow the previous pattern of all config classes being initialized as const objects in this method and the `AuthServices` constructor.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/business/auth_users.dart
```diff
@@ -179,6 +180,169 @@ class AuthUsers {
...
+          transaction: transaction,
+        );
+      }
+
+      userToKeep = await userMergeConfig.mergeHandler(
+        session,
+        userToKeep,
+        userToRemove,
+        transaction: transaction,
+      );
```
marcelomendoncasoares: Comment: Replacing the `userToKeep` at every call of the `mergeHandler` is dangerous, as pointed in the discussion in the config class, since a poor handler implementation can lead to the next receiving both user to keep and to remove as the same object.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/business/auth_users.dart
```diff
@@ -179,6 +180,169 @@ class AuthUsers {
...
+          session,
+          removeProfile.copyWith(authUserId: userToKeep.id),
+          columns: (final t) => [t.authUserId],
+          transaction: transaction,
+        );
+      }
+    }
+
+    return userToKeep;
+  }
```
marcelomendoncasoares: Discussion: As suggested in the config comment, instead of having this as a single function, we could have a separate object on the config class to register each individual handler for the core data. It would allow users to do granular changes - if they want a custom profile merging, but without opting-out from all the rest or having to copy it - and also a easy comprehension on everything that is merged by only looking at the config class.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/providers/apple/business/apple_idp.dart
```diff
@@ -133,6 +133,43 @@ class AppleIdp {
...
+  /// [userToRemove] will be deleted.
+  ///
+  /// If [userToKeep] does not have an [AppleAccount], the [AppleAccount] of
+  /// [userToRemove] will be transferred to [userToKeep].
+  static Future<void> migrate(
+    final Session session,
+    final AuthUserModel userToKeep,
+    final AuthUserModel userToRemove, {
+    required final Transaction transaction,
+  }) async {
```
marcelomendoncasoares: Suggestion: Since they have the same type, the `userToKeep` and `userToRemove` are better as named parameters to help prevent users switching the order. Same is valid for all similar signatures.

marcelomendoncasoares (changes requested):
Great work with this next piece of the anonymous auth feature, @craiglabenz! I have focused this review more on the API and design decisions. On a next review after all the discussions I'll get to the tests in detail.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/business/auth_users_config.dart
craiglabenz: 1. RE the default method that throws: I think that's reasonable. The trade-offs I was weighing were around the complexity to reach a minimum viable configuration. And, while a required callback parameter seems to add to complexity rather than diminish it, I think it is more self-explanatory than an optional parameter which could be ignored and throws by default. This scenario might only reveal itself later on (especially for new Serverpod developers who haven't yet internalized all of the auth systems). 

    I don't think either option is clearly superior, so if it's more idiomatic to Serverpod's existing patterns to go with a default function which throws, I am happy to make that change.

2. An official callback registration function would be a nice improvement. I'll convert that.

3. I was probably being un-idiomatic in writing methods that each returned a user. My goal was to have each method accept the latest user model and not have to continuously run queries to reload the same user again and again and again, but in reflecting, I think the reloading pattern is actually closer to how Serverpod handles this kind of thing elsewhere; favoring cleanliness and correctness over sweating every last query.

    I can convert this.

4. Fair. I'm just always so nervous about deleting data. What if a developer forgets some portion of their application merge? I'd feel bad having written the code to delete the unwanted `AuthUser` for them, thus taking down that extra data with it. Still, you're right.

5. I simultaneously like your suggestion to convert the core data merge parameter from a boolean into a default callback and an wondering when a developer would override this. Surely any such situation is actually just a bug in Serverpod?

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/business/auth_services.dart
craiglabenz: Yep, as I was fleshing out all of the implications of the nested config, I kept thinking, _I think the nesting is the problem_ 😆

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/business/auth_users.dart
craiglabenz: I'll convert these return types to `Future<void>` and pass in Ids to every handler.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/business/auth_users_config.dart
craiglabenz: > An official callback registration function would be a nice improvement. I'll convert that.

I'm having second thoughts about the callback registration function, as it would require the other Idps to become aware of the active account merging utility (or its `config` object). This likely requires a two-phase initialization process, as the account merging utility must first be registered before the other Idps could find it. Conversely, in the current setup, the developer would simply register static functions from all of their Idps in their initial setup.

---

On a separate thought, it would be nice if the auth system could opaquely allow each Idp to register its merge function so that developers don't have to; but that would likely require a base class which informs the existence of such a merge method (currently named `migrate` in this PR). Without this additional layer to power said automatic registration, I am no longer seeing much benefit in the registration function.

---

I will implement the rest of this refactor but hold on this part for now until we discuss further.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/business/auth_users.dart
craiglabenz: I have applied the other refactoring suggestions, but as stated above am pausing on this suggestion until further discussion.

craiglabenz:
I forgot to pull the commit you proposed within GitHub, which created a tangled mess in the branch; so I reset everything. My two separate commits and your suggestion are now just 1 blob.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/business/auth_users_config.dart
marcelomendoncasoares: > 1. RE the default method that throws: I think that's reasonable. The trade-offs I was weighing were around the complexity to reach a minimum viable configuration. And, while a required callback parameter seems to add to complexity rather than diminish it, I think it is more self-explanatory than an optional parameter which could be ignored and throws by default. This scenario might only reveal itself later on (especially for new Serverpod developers who haven't yet internalized all of the auth systems).
>
>    I don't think either option is clearly superior, so if it's more idiomatic to Serverpod's existing patterns to go with a default function which throws, I am happy to make that change.

It would be indeed more idiomatic, since all configs are constant and always present. In the end, we would just be moving the place to throw if a merge is attempted when no configuration was made.

> 5. I simultaneously like your suggestion to convert the core data merge parameter from a boolean into a default callback and an wondering when a developer would override this. Surely any such situation is actually just a bug in Serverpod?

A few use cases I can think of are if users want to do custom merging when one of the users possesses certain scopes (like `admin`), filter which tokens are kept valid, hook custom profile relations merging, wrap the operation in audit logging, etc.

---

> I'm having second thoughts about the callback registration function, as it would require the other Idps to become aware of the active account merging utility (or its `config` object). This likely requires a two-phase initialization process, as the account merging utility must first be registered before the other Idps could find it. Conversely, in the current setup, the developer would simply register static functions from all of their Idps in their initial setup.
> 
> On a separate thought, it would be nice if the auth system could opaquely allow each Idp to register its merge function so that developers don't have to; but that would likely require a base class which informs the existence of such a merge method (currently named `migrate` in this PR). Without this additional layer to power said automatic registration, I am no longer seeing much benefit in the registration function.
> 
> I will implement the rest of this refactor but hold on this part for now until we discuss further.

Maybe a middle-groud here is to merge the IDPs from other mechanism that is evaluated when the request comes instead of when the config is created. For example, by exposing a `mergeIdentityProviders` method on the `AuthServices` class that iterates over existing providers calling their respective merge callback. Then, the `mergeHooks` would be only for custom hooks from users, and not for internal usage on the package.

Such approach would achieve both the seamless "just works" with no coupling on the IDP level. It would also prevent unexpected behavior in case users forget to add the callback. What do you think?

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/business/auth_users.dart
marcelomendoncasoares: As mentioned on the use cases above, having the possibility of overriding core data merge handlers individually would make very easy to deal with custom use cases. The most common example might be if I have a custom profile class that is a relation towards the `AuthUser` that I want to merge before the old one is deleted.

craiglabenz:

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/auth_user/util/auth_user_scopes_extension.dart
craiglabenz: Looks like I actually removed whatever call site was originally using this, so I'm now deleting.

marcelomendoncasoares (approved):
Thanks a lot for the work on the next step of this feature, @craiglabenz! Since I took too long to review, I preferred pushing some changes directly so we can jump right off to the account linking step.

Most of my comments would be around the consolidation of the `IdentityProvider` interface - that we can do now because of the release of Serverpod 4.0 - and tests. So I did the consolidation and worked a bit on the tests to broaden the coverage and adjust the pattern.

I'll leave the PR opened for you to review my changes and also commit other changes you might feel adequate. Otherwise, on Monday I'll solve the CLA issue for us to get it merged! 🚀

PR #4771, opened 2026-02-25
feat: Cloud storage improvements

### LINKED ISSUE #1625: Incorrect URL Format in AWS S3 File Uploads
**Describe the bug**
There is a critical bug in Serverpod (version 1.1.1) related to file uploads with AWS S3. The issue stems from an incorrect format in the "description" URL, where the S3 domain and the region are incorrectly concatenated using a dash ("-") instead of a dot ("."). This formatting error leads to unsuccessful file uploads, as the expected format should be "....s3.[region]..." but it incorrectly appears as "....s3-[region]...".

**To Reproduce**
1. Set up Serverpod (version 1.1.1) for file uploads to AWS S3.
2. Initiate a file upload to S3 using the Serverpod framework.
3. Observe the error during the upload process, specifically focusing on the URL format used to access the S3 endpoint.
4. The URL is incorrectly formatted, using a dash to separate the S3 domain and region, resulting in a failed upload.

**Expected behavior**
The expected behavior is a seamless and successful file upload to AWS S3 when using Serverpod 1.1.1. The URL should be formatted correctly, chaining the S3 domain with the region using a dot (.), i.e., "....s3.[region]...".

**Serverpod versions**
 -  _CLI Version:_ 1.1.1
 - _Library version:_ 1.1.1

**Platform information**
Windows, Mac

**Additional context**
This issue is critical as it hinders basic file upload functionality with AWS S3, a common use case for Serverpod applications. Prompt attention and a fix would be greatly appreciated to ensure smooth operation and integration with AWS services.

**It's crazy that nobody ever found this bug in months.**

### LINKED ISSUE #3701: Add comprehensive tests for FileUploader multipart uploads with S3 and GCP integration
## Context

The FileUploader class supports two upload types:
- Binary uploads (used for Postgres backend) 
- Multipart uploads (used for S3 and GCP)

Currently, the multipart upload functionality may not have adequate test coverage, particularly E2E tests that exercise the actual integration with cloud storage services.

## Requested Work

Add more comprehensive tests for FileUploader that cover:
- Multipart upload type functionality 
- Integration testing with S3 and GCP services
- Both upload types to ensure full coverage
- Edge cases and error scenarios

## Additional Context

This was identified during the FileUploader refactoring in PR #3695 where StreamedRequest was introduced to reduce memory pressure during uploads.

## References

- Related PR: https://github.com/serverpod/serverpod/pull/3695
- Comment discussion: https://github.com/serverpod/serverpod/pull/3695#discussion_r2158695734

### LINKED ISSUE #4588: All Google Cloud Storage upload descriptions create publicly accessible files/uploads
### Describe the bug

The `createDirectFileUploadDescription` implementation for Google Cloud Storage does not pass the public/private cloud storage configuration flag to `GCPS3Uploader` when generating upload descriptions, resulting in all upload descriptions being generated as public. Files subsequently uploaded with this description are uploaded with the Cloud Stoarge `acl` set to `public-read`.

This behaviour is not expected and represents a security risk if access to the files is not further restricted by other Google Cloud permissions/ACLs.

This is resolved by including `public` when calling `GCPS3Uploader.getDirectUploadDescription(...)`. PR submitted.

### To reproduce

- Create a Serverpod Project
- Add a `GoogleCloudStorage` `CloudStorage` provider to the server with `public` set to `false`
- Generate an upload description by calling `session.storage.createDirectFileUploadDescription(...)` using the storage set up prior
- Note the `acl` key is set to `public-read`
- Upload a file from the client utilising this upload description
- Note the file being publicly accessible

### Expected behavior

- Upload description `acl` set to `private`
- Files uploaded with the description are not publicly accessible

### Library version

Serverpod 3.2.1 _however_ present since initial commit in March 2023

### Platform information

N/A

### Additional context

_No response_

- [x] I want to work on this

### LINKED ISSUE #4623: Cloud storage improvements (light)
The current cloud storage integration works, but it's not well documented and non-trivial to set up with Google Cloud. Let's patch it up and make play well with GCP and other cloud storage providers other than GCP and AWS.

Image

We currently have integrations for GCP and AWS which lives in the `serverpod_cloud_storage_gcp` and `serverpod_cloud_storage_aws` packages. They are virtually the same, as the GCP integration uses the AWS compatibility layer provided by Google Cloud.

## New GCP integration
Build a new integration specifically for GCP, which is using Google Cloud's native APIs. There is working code used by Serverpod Cloud, which can be used as a starting point. We can most likely use the very same APIs, but check for another set of keys for accessing the Google APIs. This would keep the implementation completely backwards compatible.

## Harmonize the AWS-compatible code
Almost all the code (except configuration) is exactly the same between the old GCP and AWS. We should move the similar code out into its own package (e.g., `serverpod_cloud_storage_s3compat`). This base package can be used with any S3 compatible service. However, we keep the AWS and GCP packages for backwards compatibility.

## Good docs
This is the largest part that is missing. Good documentation for how to set up GCP buckets to be correctly configured with Serverpod (and in the extension, Serverpod Cloud).

Future goals are outlined in the more comprehensive RFC:
https://github.com/serverpod/serverpod/issues/3128

### LINKED ISSUE #4733: `session.storage.createDirectFileUploadDescription()` should expose the options for max file size and expiration
### Problem to Solve

why is it that in `CloudStorage` ,  `createDirectFileUploadDescription()` includes parameters for max file size and expiration , but `StorageAccess` which is the getter on session ... does not expose those options.

`session.storage.createDirectFileUploadDescription()`
vs
`session.server.serverpod.storage[storageId].createDirectFileUploadDescription()`

### Proposal

add the parameters to the method.

having them as global settings for the storage is also good

### Use Case

self explanatory

### Alternatives

/

### Additional context

_No response_

- [ ] I want to work on this

nielsenko:
## Refactor cloud storage with shared S3-compatible base package

Introduces a layered architecture for cloud storage integrations, eliminating code duplication between S3 and GCP packages while adding support for new providers.

### Shared base package (`serverpod_cloud_storage_s3_compat`)
- `S3CompatCloudStorage` base implementation using configurable abstractions
- `S3EndpointConfig` interface with implementations for AWS, GCP, R2, and custom endpoints
- `S3UploadStrategy` interface with multipart POST (AWS/GCP/MinIO) and presigned PUT (R2) strategies
- `S3Client` for signed S3 API requests with proper resource lifecycle (`close()`)
- Proper error handling: distinguishes 404 from server errors instead of silently swallowing failures

### Native Google Cloud Storage (`NativeGoogleCloudStorage`)
- Full GCS JSON API implementation with service account credentials
- **Application Default Credentials (ADC)** support via IAM `signBlob` API and GCE metadata server
- Local RSA-SHA256 signing via `pointycastle` (avoids `googleapis_auth` ^2.1.0 dependency chain that forces Flutter 3.38.1+)
- Async factory constructors: `create()`, `fromServiceAccountJson()`, `fromApplicationDefaultCredentials()`

### New features
- **`preventOverwrite`** parameter on `storeFile` and `createDirectFileUploadDescription` — prevents uploading to an existing path (GCS native: `ifGenerationMatch: '0'`; S3 presigned PUT: `If-None-Match: *`; S3 multipart POST: not supported)
- **`maxFileSize`, `expirationDuration`, and `contentLength`** exposed on `StorageAccess.createDirectFileUploadDescription`. When `contentLength` is provided it is validated against `maxFileSize` and signed into presigned PUT URLs, enabling server-side enforcement of exact file size
- New `serverpod_cloud_storage_r2` package for Cloudflare R2 support
- `FileUploader` updated to support PUT method and custom headers for R2-style uploads

### Hardening & cleanup
- Fix silent error swallowing: `retrieveFile`, `fileExists`, and `deleteFile` now properly distinguish 404 from 5xx errors
- Throw exceptions on upload failure instead of silent null return
- Fix HTTP client leak in `PresignedPutUploadStrategy.uploadData`
- Fix `ByteData` buffer aliasing in `retrieveFile` using `ByteData.sublistView`
- Fix `Policy` expiration to use proper ISO 8601 formatting
- Add `close()` to `S3Client` and `S3CompatCloudStorage` for resource cleanup
- Replace hand-rolled MIME type lookup with `mime` package
- Replace `Completer` pattern with async factory constructors (eliminates race between `isCompleted` checks and async init)
- Improve GCP HMAC credential error messages

Fixes #4623
Fixes #4588
Fixes #4733
Fixes #1625
Closes: #3701

Related:
- Issue #3128
- PR #4200
- PR #3938
- PR #4525

## Pre-launch Checklist

## Breaking changes

None. Refactored packages maintain full backwards compatibility with same constructor parameters and environment variable names.


nielsenko:
Rebased on main

integrations/cloud_storage_integration_test_support/lib/src/run_s3_compat_integration_tests.dart
```diff
@@ -0,0 +1,653 @@
...
+  required S3CompatTestConfig? config,
+  String? skipReason,
+}) {
+  final effectiveSkipReason = config == null
+      ? (skipReason ?? 'Credentials not configured in environment')
+      : null;
+
+  group(
+    'Given a real ${config?.providerName ?? 'S3-compatible'} bucket',
+    skip: effectiveSkipReason,
```
marcelomendoncasoares, line 28: Suggestion: This top level group bearing a "Given" blocks some of the underlying tests to properly set their test scenarios using the full description. Maybe we could have a generic group here to allow using the full Given/when/then on the test cases?

integrations/cloud_storage_integration_test_support/lib/src/run_s3_compat_integration_tests.dart
```diff
@@ -0,0 +1,653 @@
...
+        final path =
+            'integration-test/${DateTime.now().millisecondsSinceEpoch}-$name';
+        testFiles.add(path);
+        return path;
+      }
+
+      group('when performing basic file operations', () {
+        test(
+          'then uploading a file via the upload strategy succeeds '
+          'and the file can be retrieved',
```
marcelomendoncasoares: Suggestion: This is an example of my previous comment. We're having part of the test scenario in the "then" because the "Given" is "occupied" by the top-level setup. If we transform it into a generic setup group, we could rewrite the descriptions like:

- Given no file in the storage when uploading a file via the upload strategy then the operation succeeds.
- Given a successfully uploaded file to the storage.
  - when getting the object then the file can be retrieved.
  - when getting the headObject then it returns 200.
  - when deleting the object then the operation succeeds and the file no longer exists.
  - when overwriting the file then the new content replaces the old content.
- Given a file path that does not exist in the storage
  - when getting the headObject then it returns 400.
  - when retrieving the file then it returns 404.
  - when deleting the file then nothing occurs and not error is returned.
- etc...

util/update_pubspecs
marcelomendoncasoares, line 1: It's also missing the new package entry on:

- `util/all_packages`
- `util/run_tests_analyze`
- `PUBLISHABLE_PACKAGES`

We should definitely unify these all... I had one ongoing branch for it a while ago, but stopped mid-work.

packages/serverpod/lib/src/cloud_storage/database_cloud_storage.dart
```diff
@@ -153,6 +153,46 @@ class DatabaseCloudStorage extends CloudStorage {
...
+  Future<void> storeFileWithOptions({
+    required Session session,
+    required String path,
+    required ByteData byteData,
+    DateTime? expiration,
+    bool verified = true,
+    required CloudStorageOptions options,
+  }) async {
+    // preventOverwrite is not supported for database storage.
+    await storeFile(
```
marcelomendoncasoares, line 174: Question: Shouldn't `preventOvewrite` be just a throw if `fileExists` returns true? If so, we could add support for it to make the API complete.

integrations/serverpod_cloud_storage_gcp/test/native/native_google_cloud_storage_test.dart
```diff
@@ -0,0 +1,994 @@
...
+    test(
+      'when accessing public '
+      'then it returns the configured value',
+      () {
+        expect(storage.public, isTrue);
+      },
+    );
+
+    group('when checking if a file exists', () {
+      test('then it returns true when the object exists', () async {
```
marcelomendoncasoares: Suggestion: The same suggestion on the the test descriptions is valid for the native tests as well to avoid the test scenario being defined in the "when" or "then" (in this last example, the existing file is part of the test scenario). The same is also valid for the files:

- integrations/serverpod_cloud_storage_gcp/test/integration/native_gcp_integration_test.dart
- integrations/serverpod_cloud_storage_s3_compat/test/s3_compat_cloud_storage_test.dart

integrations/serverpod_cloud_storage_s3/lib/src/cloud_storage/s3_cloud_storage.dart
```diff
@@ -0,0 +1,81 @@
...
+///
+/// ```dart
+/// pod.addCloudStorage(S3CloudStorage(
+///   serverpod: pod,
+///   storageId: 'public',
+///   public: true,
+///   region: 'us-east-1',
+///   bucket: 'my-bucket',
+/// ));
+/// ```
```
marcelomendoncasoares, line 28: Suggestion(NIT): Fix the formatting on the docs to follow the style of the current formatter.

```suggestion
/// ```dart
/// pod.addCloudStorage(
///   S3CloudStorage(
///     serverpod: pod,
///     storageId: 'public',
///     public: true,
///     region: 'us-east-1',
///     bucket: 'my-bucket',
///   ),
/// );
/// ```
```

integrations/serverpod_cloud_storage_s3_compat/lib/src/endpoints/r2_endpoint_config.dart
```diff
@@ -0,0 +1,52 @@
+import 's3_endpoint_config.dart';
+
+/// Cloudflare R2 endpoint configuration.
+///
+/// R2 uses a different URL pattern than AWS S3:
+/// - API endpoint: `accountId.r2.cloudflarestorage.com`
+/// - Public URLs: `bucket.accountId.r2.dev` (or custom domain)
+///
+/// R2 uses 'auto' as the region for all operations.
+class R2EndpointConfig extends S3EndpointConfig {
```
marcelomendoncasoares: Question: Can't these provider-specific `S3EndpointConfig` classes be implemented in their specific packages? Since they are only actually used there.

integrations/serverpod_cloud_storage_s3_compat/test/s3_compat_cloud_storage_test.dart
```diff
@@ -0,0 +1,538 @@
...
+    );
+  }
+
+  /// Test helper for verify direct upload without a real Session.
+  Future<bool> testVerifyDirectFileUpload(String path) async {
+    return testFileExists(path);
+  }
+}
+
+void main() {
```
marcelomendoncasoares: Suggestion(NIT): Move these test classes to the end of the file to make it easy to find and read the tests without having to first dig in details of the test object implementations.

marcelomendoncasoares (changes requested):
Great work with this big refactor! Only a few suggestions and questions on the comments.

integrations/cloud_storage_integration_test_support/lib/src/run_s3_compat_integration_tests.dart
nielsenko: Fixed in 5924855

integrations/cloud_storage_integration_test_support/lib/src/run_s3_compat_integration_tests.dart
nielsenko: Fixed in 5924855

packages/serverpod/lib/src/cloud_storage/database_cloud_storage.dart
nielsenko: Fixed in b562b19

util/update_pubspecs
nielsenko: Fixed in 0e43461b

integrations/serverpod_cloud_storage_s3/lib/src/cloud_storage/s3_cloud_storage.dart
nielsenko: Fixed in b562b19

integrations/serverpod_cloud_storage_s3_compat/lib/src/endpoints/r2_endpoint_config.dart
nielsenko: ~~This will create a circular dependency. Dart handles, but do we want that?~~

integrations/serverpod_cloud_storage_s3_compat/test/s3_compat_cloud_storage_test.dart
nielsenko: Done

integrations/serverpod_cloud_storage_gcp/test/native/native_google_cloud_storage_test.dart
nielsenko: Fixed in 5924855

integrations/serverpod_cloud_storage_s3_compat/lib/src/endpoints/r2_endpoint_config.dart
nielsenko: Fixed in 32cfe04

integrations/cloud_storage_integration_test_support/lib/src/run_s3_compat_integration_tests.dart
```diff
@@ -0,0 +1,666 @@
...
+                    preventOverwrite: true,
+                  );
+
+              expect(description, isNotNull);
+
+              final duplicateData = ByteData.view(
+                Uint8List.fromList('duplicate'.codeUnits).buffer,
+              );
+
+              // Upload should fail (returns false or throws)
```
marcelomendoncasoares: Not accurate

integrations/cloud_storage_integration_test_support/lib/src/run_s3_compat_integration_tests.dart
```diff
@@ -0,0 +1,666 @@
...
+            await client.deleteObject(path);
+
+            // Verify file no longer exists
+            final response = await client.headObject(path);
+            expect(response.statusCode, 404);
+
+            // Remove from cleanup list since already deleted
+            testFiles.remove(path);
+          },
+        );
```
marcelomendoncasoares, line 146: Suggestion(NIT): Delete these redundant comments.

```suggestion
        test(
          'when deleting the file then it no longer exists',
          () async {
            await client.deleteObject(path);

            final response = await client.headObject(path);
            expect(response.statusCode, 404);

            testFiles.remove(path);
          },
        );
```

integrations/cloud_storage_integration_test_support/lib/src/run_s3_compat_integration_tests.dart
```diff
@@ -0,0 +1,666 @@
...
+
+        test(
+          'when deleting the file then no error is returned',
+          () async {
+            // Should complete without error
+            await client.deleteObject(
+              'non-existent-file-${DateTime.now().millisecondsSinceEpoch}.txt',
+            );
```
marcelomendoncasoares, line 202: Suggestion(NIT): Same here, since the test description and code is already clear.

```suggestion
        test(
          'when deleting the file then no error is returned',
          () async {
            await client.deleteObject(
              'non-existent-file-${DateTime.now().millisecondsSinceEpoch}.txt',
            );
```

integrations/cloud_storage_integration_test_support/lib/src/run_s3_compat_integration_tests.dart
```diff
@@ -0,0 +1,666 @@
...
+            final response = await client.getObject(path);
+
+            expect(response.statusCode, 200);
+            expect(response.bodyBytes, bytes);
+          },
+        );
+
+        test(
+          'when uploading a larger file (1MB) then the upload succeeds',
+          () async {
```
marcelomendoncasoares, line 237: Suggestion: Make it a different "Given":

- Given a large binary data to upload (1MB) when uploading to the storage then the upload succeeds.

integrations/cloud_storage_integration_test_support/lib/src/run_s3_compat_integration_tests.dart
```diff
@@ -0,0 +1,666 @@
...
+
+              expect(description, isNotNull);
+
+              final duplicateData = ByteData.view(
+                Uint8List.fromList('duplicate'.codeUnits).buffer,
+              );
+
+              // Upload should fail (returns false or throws)
+              final uploader = FileUploader(description!);
+              final success = await uploader.uploadByteData(duplicateData);
```
marcelomendoncasoares: Question: Is this comment accurate? If the call throws, the test will break and not reach the final expects.

integrations/cloud_storage_integration_test_support/lib/src/run_s3_compat_integration_tests.dart
```diff
@@ -0,0 +1,666 @@
...
+
+            expect(response.statusCode, 200);
+            expect(response.body, content);
+          },
+        );
+
+        test(
+          'when uploading with unicode characters in the path '
+          'then the file is handled correctly',
+          () async {
```
marcelomendoncasoares: Suggestion: Break these tests in two separate scenarios, since the group is not actually sharing anything:
- Given a path with spaces when uploading the file then it is handled correctly.
- Given a path with unicode characters when uploading the file then it is handled correctly.

packages/serverpod/lib/src/cloud_storage/database_cloud_storage.dart
marcelomendoncasoares: Question: Should we add tests for it?

integrations/serverpod_cloud_storage_gcp/lib/src/config/gcp_endpoint_config.dart
marcelomendoncasoares, line 1: Discussion: These `S3EndpointConfig` classes are in `src/endpoints/*` files, but they are endpoint **config** classes. Would naming the folder `config` instead of `endpoint` be more semantic? The main point of the distinction for me is that we use `endpoints` on modules and projects to store actual `Endpoint` classes, which could cause confusion in this case.

marcelomendoncasoares (changes requested):
Just a few more comments on the tests and one point for discussion regarding the folder name for the endpoint config classes.

marcelomendoncasoares (approved):
Only one minor suggestion to replicate the changes done on the `run_s3_compat_integration_tests.dart` to the other similar storage tests (like the `native_gcp_integration_test.dart`). Apart from this, LGTM! 🚀 

I'll leave it already approved, so feel free to merge directly after synchronizing the tests.

PR #4779, opened 2026-02-27
feat: Add support for finding and deleting broken future calls on startup

Crazelu:
This PR adds support for scanning and deleting broken future calls on startup. Broken future calls may lead to runtime exceptions when the server attempts to run them. The scanning and deletion behaviours are enabled/disabled using YAML configs and environment variable overrides.

Fixes [4550](https://github.com/serverpod/serverpod/issues/4550)

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
+++ b/packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
@@ -272,8 +272,8 @@ class FutureCallManager {
 
     if (count >= threshold) {
       _logSession.log(
-        'Skipping check for broken future calls. '
-        'Enable FutureCallConfig.checkBrokenCalls to perform the check. '
+        'Skipping automatic check for broken future calls due to high number of future calls in the database. '
+        'Enable FutureCallConfig.checkBrokenCalls to always perform the check, regardless of the number of future calls. '
         'Optionally enable FutureCallConfig.deleteBrokenCalls to automatically delete broken future calls.',
         level: LogLevel.warning,
       );
--- a/tests/serverpod_test_server/test_integration/future_calls/broken_future_calls_test.dart
+++ b/tests/serverpod_test_server/test_integration/future_calls/broken_future_calls_test.dart
@@ -112,51 +112,53 @@ void main() {
         await server.shutdown(exitProcess: false);
       });
 
-      test(
-        'when starting Serverpod, then unregistered and broken future calls are not logged',
-        () async {
+      group('when starting Serverpod', () {
+        setUp(() async {
           await server.start();
-          await server.internalLoggingSession.close();
+        });
 
-          final logs = await LoggingUtil.findAllLogs(session);
-          final logMessages =
-              logs.lastOrNull?.logs.map((e) => e.message).toList() ?? [];
+        test(
+          'then unregistered and broken future calls are not logged',
+          () async {
+            await server.internalLoggingSession.close();
 
-          expect(
-            logMessages.where((message) => message.contains(r'TestCall\d')),
-            isEmpty,
-          );
-        },
-      );
+            final logs = await LoggingUtil.findAllLogs(session);
+            final logMessages =
+                logs.lastOrNull?.logs.map((e) => e.message).toList() ?? [];
 
-      test(
-        'when starting Serverpod, then unregistered and broken future calls are not deleted from the database',
-        () async {
-          await server.start();
-          final entries = await FutureCallEntry.db.find(session);
-          expect(entries, hasLength(futureCallsCount));
-        },
-      );
+            expect(
+              logMessages.where((message) => message.contains(r'TestCall\d')),
+              isEmpty,
+            );
+          },
+        );
 
-      test(
-        'when starting Serverpod, then a warning is logged about skipping the check for broken future calls',
-        () async {
-          await server.start();
-          await server.internalLoggingSession.close();
+        test(
+          'then unregistered and broken future calls are not deleted from the database',
+          () async {
+            final entries = await FutureCallEntry.db.find(session);
+            expect(entries, hasLength(futureCallsCount));
+          },
+        );
 
-          final logs = await LoggingUtil.findAllLogs(session);
-          final logEntry = logs.last.logs.first;
+        test(
+          'then a warning is logged about skipping the check for broken future calls',
+          () async {
+            await server.internalLoggingSession.close();
+            final logs = await LoggingUtil.findAllLogs(session);
+            final logEntry = logs.last.logs.first;
 
-          expect(
-            logEntry.message,
-            matches(
-              'Skipping check for broken future calls. '
-              'Enable FutureCallConfig.checkBrokenCalls to perform the check. '
-              'Optionally enable FutureCallConfig.deleteBrokenCalls to automatically delete broken future calls.',
-            ),
-          );
-        },
-      );
+            expect(
+              logEntry.message,
+              matches(
+                'Skipping automatic check for broken future calls due to high number of future calls in the database. '
+                'Enable FutureCallConfig.checkBrokenCalls to always perform the check, regardless of the number of future calls. '
+                'Optionally enable FutureCallConfig.deleteBrokenCalls to automatically delete broken future calls.',
+              ),
+            );
+          },
+        );
+      });
     },
   );
 
@@ -198,50 +200,51 @@ void main() {
         await server.shutdown(exitProcess: false);
       });
 
-      test(
-        'when starting Serverpod, then unregistered future calls are logged',
-        () async {
+      group('when starting Serverpod', () {
+        setUp(() async {
           await server.start();
-          await server.internalLoggingSession.close();
+        });
 
-          final logs = await LoggingUtil.findAllLogs(session);
-          final logEntry = logs.last.logs.first;
+        test(
+          'then unregistered future calls are logged',
+          () async {
+            await server.internalLoggingSession.close();
+            final logs = await LoggingUtil.findAllLogs(session);
+            final logEntry = logs.last.logs.first;
 
-          expect(
-            logEntry.message,
-            matches(
-              r'Unregistered future call: \{.*\"name\":\s*\"TestCall0\".*\}\n.*',
-            ),
-          );
-        },
-      );
+            expect(
+              logEntry.message,
+              matches(
+                r'Unregistered future call: \{.*\"name\":\s*\"TestCall0\".*\}\n.*',
+              ),
+            );
+          },
+        );
 
-      test(
-        'when starting Serverpod, then broken future calls are logged',
-        () async {
-          await server.start();
-          await server.internalLoggingSession.close();
+        test(
+          'then broken future calls are logged',
+          () async {
+            await server.internalLoggingSession.close();
+            final logs = await LoggingUtil.findAllLogs(session);
+            final logEntry = logs.last.logs.first;
 
-          final logs = await LoggingUtil.findAllLogs(session);
-          final logEntry = logs.last.logs.first;
+            expect(
+              logEntry.message,
+              matches(
+                r'.*Future call failed deserialization. Error: .* Entry: \{.*\"name\":\s*\"TestCall1\".*\}\n',
+              ),
+            );
+          },
+        );
 
-          expect(
-            logEntry.message,
-            matches(
-              r'.*Future call failed deserialization. Error: .* Entry: \{.*\"name\":\s*\"TestCall1\".*\}\n',
-            ),
-          );
-        },
-      );
-
-      test(
-        'when starting Serverpod, then unregistered and broken future calls are not deleted from the database',
-        () async {
-          await server.start();
-          final entries = await FutureCallEntry.db.find(session);
-          expect(entries, hasLength(futureCallsCount));
-        },
-      );
+        test(
+          'then unregistered and broken future calls are not deleted from the database',
+          () async {
+            final entries = await FutureCallEntry.db.find(session);
+            expect(entries, hasLength(futureCallsCount));
+          },
+        );
+      });
     },
   );
 
@@ -287,45 +290,47 @@ void main() {
         await server.shutdown(exitProcess: false);
       });
 
-      test(
-        'when starting Serverpod, then unregistered future calls are logged',
-        () async {
+      group('when starting Serverpod', () {
+        setUp(() async {
           await server.start();
-          await server.internalLoggingSession.close();
+        });
 
-          final logs = await LoggingUtil.findAllLogs(session);
-          final logEntry = logs.last.logs.first;
+        test(
+          'then unregistered future calls are logged',
+          () async {
+            await server.internalLoggingSession.close();
+            final logs = await LoggingUtil.findAllLogs(session);
+            final logEntry = logs.last.logs.first;
 
-          expect(logEntry.logLevel, LogLevel.warning);
-          expect(
-            logEntry.message,
-            matches(
-              r'Unregistered future call: \{.*\"name\":\s*\"TestCall0\".*\}\n'
-              r'Unregistered future call: \{.*\"name\":\s*\"TestCall1\".*\}\n.*',
-            ),
-          );
-        },
-      );
+            expect(logEntry.logLevel, LogLevel.warning);
+            expect(
+              logEntry.message,
+              matches(
+                r'Unregistered future call: \{.*\"name\":\s*\"TestCall0\".*\}\n'
+                r'Unregistered future call: \{.*\"name\":\s*\"TestCall1\".*\}\n.*',
+              ),
+            );
+          },
+        );
 
-      test(
-        'when starting Serverpod, then broken future calls are logged',
-        () async {
-          await server.start();
-          await server.internalLoggingSession.close();
+        test(
+          'then broken future calls are logged',
+          () async {
+            await server.internalLoggingSession.close();
+            final logs = await LoggingUtil.findAllLogs(session);
+            final logEntry = logs.last.logs.first;
 
-          final logs = await LoggingUtil.findAllLogs(session);
-          final logEntry = logs.last.logs.first;
-
-          expect(logEntry.logLevel, LogLevel.warning);
-          expect(
-            logEntry.message,
-            matches(
-              r'.*Future call failed deserialization. Error: .* Entry: \{.*\"name\":\s*\"TestCall2\".*\}\n'
-              r'Future call failed deserialization. Error: .* Entry: \{.*\"name\":\s*\"TestCall3\".*\}\n',
-            ),
-          );
-        },
-      );
+            expect(logEntry.logLevel, LogLevel.warning);
+            expect(
+              logEntry.message,
+              matches(
+                r'.*Future call failed deserialization. Error: .* Entry: \{.*\"name\":\s*\"TestCall2\".*\}\n'
+                r'Future call failed deserialization. Error: .* Entry: \{.*\"name\":\s*\"TestCall3\".*\}\n',
+              ),
+            );
+          },
+        );
+      });
     },
   );
 


packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -246,4 +246,84 @@ class FutureCallManager {
...
+      final deletedEntries = await FutureCallEntry.db.delete(
+        _internalSession,
+        allCalls,
+      );
+      _internalSession.log(
+        'Deleted ${deletedEntries.length}/${allCalls.length} broken future calls.',
+      );
+    }
+
+    return unregisteredCalls.isNotEmpty || brokenCalls.isNotEmpty;
```
marcelomendoncasoares: Suggestion: Reuse the `allCalls` on the check:

```suggestion
    return allCalls.isNotEmpty;
```

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -246,4 +246,84 @@ class FutureCallManager {
...
+        _serializationManager.decode(
+          futureCallEntry.serializedObject!,
+          futureCall.dataType,
+        );
+      }
+      return true;
+    } catch (e) {
+      return false;
+    }
+  }
```
marcelomendoncasoares: Discussion: Instead of returning a bool, would it be worth returning the exception string to log the actual failure? So users can inspect why the model deserialization was broken.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -773,16 +781,46 @@ class Serverpod {
...
           ),
         );
       } else {
+        if (canScanBrokenFutureCalls) {
+          unawaited(
+            _futureCallManager?.scanBrokenFutureCalls(
+              deleteBrokenCalls: shouldDeleteBrokenFutureCalls,
+            ),
+          );
+        }
```
marcelomendoncasoares: Suggestion: Would it be possible to move this to a private method inside the `FutureCallManager` and call it as part of the `start` method? If not possible to bundle on the start, at least move the body to a method to be invoked explicitly before the start. It would be good to avoid pushing more feature-specific code into the `Serverpod` start method, which is already too long.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -695,12 +695,20 @@ class FutureCallConfig {
   /// How long to wait before checking the queue again.
   final Duration scanInterval;
 
+  /// If true, the server will scan for broken future calls on startup.
+  final bool? scanBrokenFutureCalls;
```
marcelomendoncasoares: Suggestion: Call it "validate" or "check" since it is a one-time only operation. The word "scan" can give a false impression of a regular process, as we have with the `scanInterval` config.

It is worth documenting the behavior, since it is not clear at a glance on why it is nullable and the `deleteBrokenFutureCalls` is not.

tests/serverpod_test_server/test_integration/future_calls/future_call_manager_test.dart
```diff
@@ -783,4 +783,222 @@ void main() async {
...
+    () {
+      late Serverpod server;
+      late Session session;
+      late FutureCallManager futureCallManager;
+      List<FutureCallEntry> futureCallEntries = [];
+      int futureCallsCount = 3;
+
+      setUp(() async {
+        server = IntegrationTestServer.create();
+        await server.start();
```
marcelomendoncasoares: Question: Do we need to use the `IntegrationTestServer` here or can we use the `withServerpod` from test framework?

tests/serverpod_test_server/test_integration/future_calls/future_call_manager_test.dart
```diff
@@ -783,4 +783,222 @@ void main() async {
       );
     },
   );
+
+  group(
+    'Given FutureCallManager with scheduled FutureCalls',
```
marcelomendoncasoares: Suggestion: Separate these test cases in different groups - describing each use case on the "Given" part of the description. It might also be worth separating these tests to a different test file, since they are scoped around broken future calls.

tests/serverpod_test_server/test_integration/future_calls/future_call_manager_test.dart
```diff
@@ -783,4 +783,222 @@ void main() async {
...
+            ),
+            completion(true),
+          );
+        },
+      );
+
+      test(
+        'when there are fewer more calls in the database than the threshold'
+        'then canPerformDefaultScanForBrokenFutureCalls returns false',
+        () async {
```
marcelomendoncasoares: Suggestion: Clarify the acting here on the "when" description.

- Given a FutureCallManager with scheduled FutureCalls
  - when calling canPerformDefaultScanForBrokenFutureCalls with a threshold greater than the existing count
    - then it returns true
  - when calling canPerformDefaultScanForBrokenFutureCalls with a threshold lower than the existing count
    - then it returns false

tests/serverpod_test_server/test_integration/future_calls/future_call_manager_test.dart
```diff
@@ -783,4 +783,222 @@ void main() async {
...
+
+          expect(logEntry.logLevel, LogLevel.warning);
+          expect(
+            logEntry.message,
+            matches(
+              r'Unregistered future call: \{.*\"name\":\s*\"TestCall0\".*\}\n'
+              r'Unregistered future call: \{.*\"name\":\s*\"TestCall2\".*\}\n',
+            ),
+          );
+        });
```
marcelomendoncasoares: Suggestion: This should actually be a different "Given" (for this and the next group), like:
- Given a FutureCallManager with broken future calls
  - when calling `validateBrokenFutureCalls` without deleteBrokenCalls set
    - then broken calls are logged.
    - then broken future calls are not deleted from the database.
  - when calling `validateBrokenFutureCalls` with deleteBrokenCalls set
    - then broken calls are logged.
    - then broken calls are deleted from the database.

tests/serverpod_test_server/test_integration/future_calls/future_call_manager_test.dart
Crazelu: `IntegrationTestServer` is necessary here to capture and verify logged output.
Specifically, it supports creating a different session which has logging enabled.

I can't get `withServerpod` to work well with logs even when `enableSessionLogging` is set to true.
And other existing tests for logs use this pattern with `IntegrationTestServer`.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Done in [a2dc0a7](https://github.com/serverpod/serverpod/pull/4779/commits/a2dc0a7909e42f419552f512385395ec1128e718)

packages/serverpod/lib/src/server/serverpod.dart
Crazelu: Done in [a2dc0a7](https://github.com/serverpod/serverpod/pull/4779/commits/a2dc0a7909e42f419552f512385395ec1128e718)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -246,4 +251,97 @@ class FutureCallManager {
       await futureCallSession.close(error: error, stackTrace: stackTrace);
     }
   }
+
+  /// Returns `true` if the server is configured to check for broken future calls
+  /// or if the server can perform a default check.
+  ///
+  /// The server can perform a default check for broken future calls
+  /// if the amount of future calls in the database
+  /// is less than [threshold].
```
marcelomendoncasoares: Suggestion: On docs, break the line at 80 chars to maintain (mostly) the standard.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -246,4 +251,97 @@ class FutureCallManager {
...
+
+  /// Checks the database for broken future calls.
+  /// Broken future calls include unregistered calls and
+  /// those with stored inputs that cannot be deserialized.
+  ///
+  /// If [FutureCallConfig.deleteBrokenFutureCalls] is enabled,
+  /// the broken future calls are deleted.
+  ///
+  /// The check will not happen if [FutureCallConfig.checkBrokenFutureCalls]
+  /// is disabled.
```
marcelomendoncasoares: Suggestion: Same here.

packages/serverpod_shared/lib/src/environment_variables.dart
```diff
@@ -181,7 +181,13 @@ enum ServerpodEnv {
...
 
   /// The interval in seconds between websocket ping messages.
-  websocketPingInterval;
+  websocketPingInterval,
+
+  /// If true, the server will check for broken future calls on startup.
+  checkBrokenFutureCalls,
+
+  /// If true, the server will delete broken future calls on startup.
+  deleteBrokenFutureCalls;
```
marcelomendoncasoares: Suggestion: Name them prefixing `futureCall`, as we do for the others (`futureCallConcurrencyLimit`, `futureCallScanInterval`, ...), and place them next to the existing future call configs.

packages/serverpod_shared/lib/src/environment_variables.dart
```diff
@@ -233,6 +239,8 @@ enum ServerpodEnv {
       (ServerpodEnv.applyRepairMigration) => 'applyRepairMigration',
       (ServerpodEnv.validateHeaders) => 'validateHeaders',
       (ServerpodEnv.websocketPingInterval) => 'websocketPingInterval',
+      (ServerpodEnv.checkBrokenFutureCalls) => 'checkBrokenFutureCalls',
+      (ServerpodEnv.deleteBrokenFutureCalls) => 'deleteBrokenFutureCalls',
```
marcelomendoncasoares: Suggestion(NIT): Follow the order of the configs here and in the other places where similar keys exist after changing their names, to keep the pattern.

packages/serverpod/lib/src/server/session.dart
```diff
@@ -259,12 +259,6 @@ abstract class Session implements DatabaseSession {
     dynamic exception,
     StackTrace? stackTrace,
   }) {
-    if (_closed) {
-      throw StateError(
-        'Session is closed, and logging can no longer be performed.',
-      );
-    }
-
```
marcelomendoncasoares, line 268: Comment: Let's keep an eye for flaky tests related to logs entry after removing this. Seems a harmless change, as we discussed, but it's the kind that has the potential to introduce flakiness.

tests/serverpod_test_server/test_integration/future_calls/broken_future_calls_test.dart
```diff
@@ -0,0 +1,878 @@
...
+
+void main() {
+  group(
+    'Given a FutureCallManager with default config and scheduled FutureCalls',
+    () {
+      late Serverpod server;
+      late Session session;
+      late FutureCallManager futureCallManager;
+      List<FutureCallEntry> futureCallEntries = [];
+      int futureCallsCount = 3;
```
marcelomendoncasoares: Suggestion: I think we can still improve a bit on the organization of these tests. The "Given" (test scenario) is describing "scheduled FutureCalls", but there are future calls being scheduled inside the test call - which is the "when" and should only be the acting. We're also mixing broken future calls with healthy ones, which results in non-isolated test cases.

A better separation would be to have each type of wrong `FutureCall` in a different "Given", with the future calls being created on the `setUp` of the group and only the `start` being called inside the test. The first groups should be the success-path test cases ("Given ... valid future calls..."), followed by the ones with error and lastly, if needed, a few with mixed entries. This would follow the principle of each test focusing on a single responsibility (from [the unmerged doc on testing guidelines](https://github.com/serverpod/serverpod_docs/pull/229/changes#diff-d56063bac00c632f16ede12fc89d9e0cb8bd8131d723429fea9ad4bc3be1648c)).

An example of such separation would be:

- Given a `FutureCallManager` with `checkBrokenFutureCalls` enabled and only valid future calls
  - when calling `start` then no errors are logged.
- Given a `FutureCallManager` with `checkBrokenFutureCalls` enabled and unregistered future calls
  - when calling `start`
    - then unregistered future calls are logged.
    - then unregistered are not deleted from the database.
- Given a `FutureCallManager` with `deleteBrokenFutureCalls` enabled and unregistered future calls
  - when calling `start` then unregistered are deleted from the database.
    > Note that there is no need to repeat the "logged" test case from before, since it is already tested.
- Given a `FutureCallManager` with `checkBrokenFutureCalls` enabled and broken future calls
  - when calling `start`
    - then broken future calls are logged.
    - then broken are not deleted from the database.
- Given a `FutureCallManager` with `deleteBrokenFutureCalls` enabled and broken  future calls
  - when calling `start` then broken are deleted from the database.
Etc...

packages/serverpod/lib/src/server/session.dart
Crazelu: I've not noticed any flaky test related to this so far.

packages/serverpod_shared/lib/src/environment_variables.dart
Crazelu: Done in [082e444](https://github.com/serverpod/serverpod/pull/4779/commits/082e4444997aa62a323a8edd46af0835948b1ef9)

tests/serverpod_test_server/test_integration/future_calls/broken_future_calls_test.dart
Crazelu: Done in [feccfa1](https://github.com/serverpod/serverpod/pull/4779/commits/feccfa1572698464a6407c7d5c89898a4e68f99e)

tests/serverpod_test_server/test_integration/future_calls/broken_future_calls_test.dart
```diff
@@ -40,21 +39,113 @@ class SimpleFutureCall extends FutureCall<SimpleData> {
...
+        );
+
+        expect(
+          logEntry.message,
+          notMatches(
+            r'Future call failed deserialization. Error: .* Entry: \{.*\"name\":\s*\"TestCall0\".*\}\n'
+            r'Future call failed deserialization. Error: .* Entry: \{.*\"name\":\s*\"TestCall1\".*\}\n',
+          ),
+        );
+      });
```
marcelomendoncasoares: Suggestion: Instead of trying to filter from the messages with `noMatches`, filter the `logs` by messages that contain the name of the classes and expect it to be empty. This will be more robust in case we change the warning message or introduce other failure modes to search for in the future.

tests/serverpod_test_server/test_integration/future_calls/broken_future_calls_test.dart
```diff
@@ -0,0 +1,1238 @@
...
+            logEntry?.message,
+            notMatches(
+              r'Future call failed deserialization. Error: .* Entry: \{.*\"name\":\s*\"TestCall0\".*\}\n'
+              r'Future call failed deserialization. Error: .* Entry: \{.*\"name\":\s*\"TestCall1\".*\}\n',
+            ),
+          );
+        },
+      );
+    },
+  );
```
marcelomendoncasoares: Suggestion: This test case can be removed, since it is redundant with the first one. With only valid future calls, changing configs wouldn't provoke a different behavior.

tests/serverpod_test_server/test_integration/future_calls/broken_future_calls_test.dart
```diff
@@ -0,0 +1,1238 @@
...
+      );
+
+      test(
+        'when starting Serverpod, then unregistered future calls are not deleted from the database',
+        () async {
+          await server.start();
+          final entries = await FutureCallEntry.db.find(session);
+          expect(entries, hasLength(futureCallsCount));
+        },
+      );
```
marcelomendoncasoares: Discussion: While reviewing this test, I just noticed we don't flag a warning in case the check is skipped, but not configured to false. But this would be good to make users aware of the safety mechanism and that it is disabled. What do you think? If we add the warning, we should also add a test to expect it.

tests/serverpod_test_server/test_integration/future_calls/broken_future_calls_test.dart
```diff
@@ -0,0 +1,1238 @@
...
+    },
+  );
+
+  group(
+    'Given a Serverpod server instance with default config and less than 1000'
+    'future calls in the database with at least 1 unregistered future call',
```
marcelomendoncasoares: Suggestion: To reduce the cartesian of tests, for the ones testing combined behavior, we can mix unregistered with broken like "...and less  than 1000 future calls in the database containing unregistered and broken future calls". Then, each test will expect its target group - as they already do.

tests/serverpod_test_server/test_integration/future_calls/broken_future_calls_test.dart
```diff
@@ -0,0 +1,1238 @@
...
+          final entries = await FutureCallEntry.db.find(session);
+          expect(entries, hasLength(futureCallsCount));
+        },
+      );
+    },
+  );
+
+  group(
+    'Given a Serverpod server instance with checkBrokenCalls enabled'
+    'in future call config and unregistered future calls',
```
marcelomendoncasoares: Question: Aren't this and the following tests redundant with the ones above? This test cases seems the same as the second one, isn't it? Or am I missing some detail?

marcelomendoncasoares (changes requested):
Great improvement to the tests! One more round of comments.

tests/serverpod_test_server/test_integration/future_calls/broken_future_calls_test.dart
Crazelu: Some of the tests were duplicated as I targeted `FutureCallManager` and also `Serverpod.start` separately.
But since `Serverpod.start` uses `FutureCallManager`, the redundancy is not necessary.

tests/serverpod_test_server/test_integration/future_calls/broken_future_calls_test.dart
Crazelu: It is a good-to-have.
I will add a warning.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -246,4 +251,106 @@ class FutureCallManager {
...
+    final count = await FutureCallEntry.db.count(
+      _internalSession,
+      limit: threshold,
+    );
+
+    if (count >= threshold) {
+      _logSession.log(
+        'Skipping check for broken future calls. '
+        'Enable FutureCallConfig.checkBrokenCalls to perform the check. '
+        'Optionally enable FutureCallConfig.deleteBrokenCalls to automatically delete broken future calls.',
```
marcelomendoncasoares: Suggestion: Also explain that the **automatic check** is being skipped due to the high number of stored future calls. Then highlight "Enable ... to always perform the check, regardless of the number of future calls". It's also worth mentioning the performance tradeoff.

tests/serverpod_test_server/test_integration/future_calls/broken_future_calls_test.dart
```diff
@@ -0,0 +1,590 @@
...
+        await FutureCallEntry.db.delete(session, futureCallEntries);
+        await session.close();
+        await server.shutdown(exitProcess: false);
+      });
+
+      test(
+        'when starting Serverpod, then unregistered and broken future calls are not logged',
+        () async {
+          await server.start();
+          await server.internalLoggingSession.close();
```
marcelomendoncasoares: Suggestion: Move this to a group inside this test group to avoid the repetition of the description and action.

```dart
  group('when starting Serverpod', () {
    setUp(() async {
      await server.start();
      await server.internalLoggingSession.close();
    });
    
    test('then...', () async {
      ...
    });
  });
```

Crazelu: Done in [6b559d5](https://github.com/serverpod/serverpod/pull/4779/commits/6b559d5a74734c55ca38c31e042a6020a48fbb9b)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Done in [6b559d5](https://github.com/serverpod/serverpod/pull/4779/commits/6b559d5a74734c55ca38c31e042a6020a48fbb9b)

Issue #4796, opened 2026-03-05
Shared model not found when referencing as field on project models without module alias


marcelomendoncasoares:
While fixing issue #4795, I've bumped into another issue when trying to reference a shared model on fields without the module alias:

```yaml
class: SharedModelContainer
fields:
  sharedModel: SharedModel # This is not working
  sharedModelWithModuleAlias: serverpod_test_shared:SharedModel # This works
```

However, the reference works as expected when extending a shared model:

```yaml
class: SharedModelSubclass
extends: SharedModel # This works as expected
fields:
  sharedModelSubclassField: String
  sharedEnumField: SharedEnum
  serverOnlyField: String?, scope=serverOnly
```

Since shared models are considered part of the same project that has the server models, it should be possible to reference them with and without the module alias. Uniqueness on the class name must also be enforced between server and shared packages.

Issue #4795, opened 2026-03-04
Server code generator calls `toJsonForProtocol` on shared models used as fields on models with non-persist fields


dipsy:
## Bug description

When a server model has a field of a shared package class type (defined via `shared_packages` in `generator.yaml`), the server code generator emits `toJsonForProtocol()` calls on those fields in the generated `toJsonForProtocol()` method. However, shared models don't implement `ProtocolSerialization` (they only implement `SerializableModel`), so `toJsonForProtocol()` doesn't exist on them.

This causes a compilation error unless the user manually adds an extension method as a workaround.

**Enums from shared packages are not affected** — they serialize via `.name`/`.index`.

## Steps to reproduce

1. Create a shared package with a class model:
```yaml
# shared/lib/src/protocol/my_shared_model.spy.yaml
class: MySharedModel
fields:
  id: UuidValue
  name: String
```

2. Reference it as a `!persist` field in a server model with `table:`:
```yaml
# server model
class: MyServerModel
table: my_server_model
fields:
  # ...
  sharedRef: shared_package:MySharedModel?, !persist
```

3. Run `serverpod generate` — succeeds
4. Compile — **fails**:
```
The method 'toJsonForProtocol' isn't defined for the type 'MySharedModel'
```

## Root cause

In `tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart`, the method `_toJsonCallConversionMethod` (around line 1577) determines which serialization method to call:

```dart
var toJson = fieldType.isSerializedByExtension || fieldType.isEnumType
    ? _toJsonMethodName
    : methodName;
```

When `methodName` is `toJsonForProtocol`, shared class types fall through to `toJson = methodName` (i.e. `toJsonForProtocol`). The generator doesn't check whether the field type is a shared model.

The `TypeDefinition` already has `projectModelDefinition` which links to `SerializableModelDefinition`, and that has `isSharedModel` (`sharedPackageName != null`). So all the information needed is available.

## Suggested fix

```dart
var isSharedClass = fieldType.projectModelDefinition?.isSharedModel ?? false;
var toJson = fieldType.isSerializedByExtension || fieldType.isEnumType || isSharedClass
    ? _toJsonMethodName
    : methodName;
```

Since shared models have no `serverOnly` fields, `toJson()` and `toJsonForProtocol()` are semantically identical for them.

[...]

PR #4797, opened 2026-03-05
fix: Fix shared models reference and `toJsonForProtocol` usage

marcelomendoncasoares:
Fixes: #4795
Fixes: #4796 (had to be fixed together for the test model to pass)

With this PR and the fix for #4796, class names are now properly enforced to be unique between models on shared packages and server/client models.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/shared_model_name_conflict_test.dart
+++ b/tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/shared_model_name_conflict_test.dart
@@ -92,15 +92,184 @@ fields:
         onErrorsCollector(collector),
       ).validateAll();
 
+      expect(collector.errors, isEmpty);
+    },
+  );
+
+  test(
+    'Given a shared enum and a server enum with the same name '
+    'when analyzing models '
+    'then an error is collected that server enums cannot share names with shared package enums.',
+    () {
+      var models = <ModelSource>[
+        ModelSourceBuilder()
+            .withIsSharedModel(true)
+            .withModuleAlias('shared')
+            .withFileName('shared_enum')
+            .withYaml(
+              '''
+enum: SharedEnum
+values:
+  - value1
+  - value2
+''',
+            )
+            .build(),
+        ModelSourceBuilder().withFileName('shared_enum').withYaml(
+          '''
+enum: SharedEnum
+values:
+  - other1
+  - other2
+''',
+        ).build(),
+      ];
+
+      var collector = CodeGenerationCollector();
+      StatefulAnalyzer(
+        config,
+        models,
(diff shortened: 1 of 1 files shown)


tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -211,6 +211,23 @@ class Restrictions {
       ];
     }
 
+    if (documentDefinition?.isSharedModel != true) {
```
nielsenko:
```suggestion
    if (!(documentDefinition?.isSharedModel ?? false)) {
```
I know it is longer, but find it easier to reason about.

tools/serverpod_cli/lib/src/generator/types.dart
```diff
@@ -358,7 +358,8 @@ class TypeDefinition {
               '${serverCode ? config.serverPackage : config.dartClientPackage}'
               '/${split[1]}';
         } else if (url == defaultModuleAlias ||
-            config.sharedModelsSourcePathsParts.containsKey(url)) {
+            config.sharedModelsSourcePathsParts.containsKey(url) ||
+            projectModelDefinition?.isSharedModel == true) {
```
nielsenko: I prefer `?? false` this `null != true` feels hacky.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/4797/commits/18bc42b3b1e5740694fc77e14a2a588a7f063725.

tools/serverpod_cli/lib/src/generator/types.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/4797/commits/18bc42b3b1e5740694fc77e14a2a588a7f063725.

marcelomendoncasoares:
> Could perhaps use more tests regarding:
> 
> * blocks client models from conflicting with shared models
> * `validateClassName` restriction applies to enums too, but only class-vs-class conflict is tested.
> 
> Overall LGTM 🚀

Good suggestion on the test coverage. Client models are already assured, since they share the same declaration as server models. But the test was misplaced, leading to this confusion. I moved it to the proper location at https://github.com/serverpod/serverpod/pull/4797/commits/093e47cb99af7c8f17662b369536343c26e445eb and added a few more test cases regarding enums and exceptions at https://github.com/serverpod/serverpod/pull/4797/commits/bfb5d53bea744de20655ed614579318f3eca3ed5.

Issue #3417, opened 2025-04-14
Future calls are lost if server shuts down during execution


SandPod:
### Describe the bug

When a `FutureCall` task is picked up for execution, it is immediately removed from the database. If the server crashes or is terminated before completing the task, the task is lost and never retried.

### To reproduce

1. Schedule a future call.
2. Let the server pick it up for execution.
3. Forcefully shut down the server before the task is completed.
4. Observe that the task is not executed and does not exist in the database anymore.

### Expected behavior

Tasks should only be removed from the database once they have completed successfully. If the server crashes mid-execution, the task should be retained and retried later.

### Library version

Serverpod version: latest main branch 

### Platform information

Affects all deployment targets with future call scheduling.

### Additional context

~A proposed solution:~
- ~When a server picks up a future call, it should **update the execution time** in the database with a retry offset, not delete the task.~
- ~The retry offset should be configurable for each future call type (perhaps by adding a new property to the FutureCall class).~
- ~The task should only be removed from the database **after successful completion**.~
- ~If a server crashes mid-execution, the task remains and will be picked up again by another server at the new execution time.~

~This ensures fault tolerance in distributed environments and prevents task loss.~

---------

Edit: New proposal after feedback:

 **Proposed solution:**
- When a server picks up a future call, it **does not immediately remove** it from the database.
- Instead, when execution is about to begin, the server **attempts to delete** the future call record as a signal that it is claiming execution.
- If the deletion **succeeds**, the task proceeds with execution.
- If the deletion **fails** (e.g., another server already deleted the row), the task is **not executed**.

This guarantees that only one server executes the task, and that it is only removed from the database if execution is guaranteed to begin.

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

inf0rmatix:
I propose changing this slightly to only guarantee that a future call has been invoked. Anything else should be left to the user to implement according to their business logic needs.

Using a delete statements are atomic and right before invoking the actual future call, trying to delete the entry ensures it has not been run already.

SandPod:
@inf0rmatix - I think your reasoning makes sense and will greatly simplify the mental model and allow users to put that guarantee in place if needed.

Will update the issue accordingly.

craiglabenz:
[...]

PR #4799, opened 2026-03-05
feat: Improving reliability of future call execution using claim-based model

Crazelu:
This PR introduces a claim-based model to improve the reliability of future call execution. It guarantees that future calls are executed at-least once even if server unexpectedly crashes just as a due future call is scheduled for execution.

Fixes [3417](https://github.com/serverpod/serverpod/issues/3417)

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
+++ b/packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
@@ -19,10 +19,6 @@ typedef InitializeFutureCall =
       String name,
     );
 
-/// A timer used to update heartbeat for a future call claim.
-@visibleForTesting
-typedef FutureCallClaimTimer = ({Timer timer, int claimId});
-
 /// The [FutureCallManager] is responsible for managing and executing
 /// [FutureCall]s in the [Serverpod] framework. A [FutureCall] is a task
 /// that is scheduled to run at a specific time in the future. These tasks
@@ -54,7 +50,7 @@ class FutureCallManager {
   bool _hasPendingStart = false;
 
   /// Collection of active claim heartbeat timers.
-  final List<FutureCallClaimTimer> _heartbeatTimers = [];
+  final List<Timer> _heartbeatTimers = [];
 
   /// Creates a new [FutureCallManager]. Typically, this is instantiated
   /// internally by the [Serverpod].
@@ -206,10 +202,7 @@ class FutureCallManager {
     await _scanner.stop();
     await _scheduler.drain();
     if (unregisterAll) _futureCalls.clear();
-    for (final entry in _heartbeatTimers) {
-      entry.timer.cancel();
-    }
-    _heartbeatTimers.clear();
+    _heartbeatTimers.forEach(_cancelHeartbeatTimer);
   }
 
   /// Internal method to dispatch a list of [FutureCallEntry] objects to
@@ -240,29 +233,19 @@ class FutureCallManager {
 
   /// Collection of active claim heartbeat timers used for testing purposes.
   @visibleForTesting
-  List<FutureCallClaimTimer> get heartbeatTimers => _heartbeatTimers;
+  List<Timer> get heartbeatTimers => _heartbeatTimers;
 
-  /// Cancels any active timer associated with [claimId].
-  void _cancelTimerForClaim(int claimId) {
-    final entry = _heartbeatTimers
-        .where((t) => t.claimId == claimId)
-        .firstOrNull;
-
-    entry?.timer.cancel();
-    _heartbeatTimers.remove(entry);
-  }
-
-  /// Cancels [timer].
-  void _cancelTimer(Timer timer) {
-    final entry = _heartbeatTimers.where((t) => t.timer == timer).firstOrNull;
-    entry?.timer.cancel();
-    _heartbeatTimers.remove(entry);
+  void _cancelHeartbeatTimer(Timer timer) {
+    _heartbeatTimers.remove(timer..cancel());
   }
 
   /// Updates heartbeat for a future call claim to keep it alive.
   /// Heartbeat is only updated if the claim is held by the
   /// running process.
-  Future<void> _updateHeartbeat(int claimId) async {
+  Future<void> _updateHeartbeat({
+    required int claimId,
+    required Timer timer,
+  }) async {
     try {
       final rows = await FutureCallClaimEntry.db.updateWhere(
         _internalSession,
@@ -270,7 +253,9 @@ class FutureCallManager {
         columnValues: (t) => [t.lastHeartbeatTime(DateTime.now().toUtc())],
       );
 
-      if (rows.isEmpty) _cancelTimerForClaim(claimId);
+      if (rows.isEmpty) {
+        _cancelHeartbeatTimer(timer);
+      }
     } catch (error, stackTrace) {
       _diagnosticsService.submitFrameworkException(error, stackTrace);
     }
@@ -332,9 +317,9 @@ class FutureCallManager {
       if (claimId != null) {
         final timer = Timer.periodic(
           _heartbeatInterval,
-          (timer) => _updateHeartbeat(claimId),
+          (timer) => _updateHeartbeat(claimId: claimId, timer: timer),
         );
-        _heartbeatTimers.add((timer: timer, claimId: claimId));
+        _heartbeatTimers.add(timer);
         return timer;
       }
     } catch (error, stackTrace) {
@@ -382,7 +367,7 @@ class FutureCallManager {
 
       await futureCallSession.close(error: error, stackTrace: stackTrace);
     } finally {
-      _cancelTimer(heartbeatTimer);
+      _cancelHeartbeatTimer(heartbeatTimer);
       await FutureCallEntry.db.deleteWhere(
         _internalSession,
         where: (t) => t.id.equals(futureCallEntry.id),
--- a/tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
+++ b/tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
@@ -430,6 +430,9 @@ void main() {
       });
 
       tearDown(() async {
+        if (!testCall.completer.isCompleted) {
+          testCall.completer.complete();
+        }
         await degradedFutureCallManager.stop(unregisterAll: true);
         await FutureCallEntry.db.deleteWhere(
           session,


docs/design/reliable_future_calls.md
```diff
@@ -0,0 +1,107 @@
...
+
+The server attempts to claim execution of a future call by inserting a record into `serverpod_future_call_claim` table.
+
+```dart
+final claim = FutureCallClaimEntry(
+  futureCallId: futureCallEntry.id,
+  serverId: _serverId,
+  time: DateTime.now().toUtc(),
+);
+await FutureCallClaimEntry.db.insertRow(_internalSession, claim);
```
marcelomendoncasoares: Comment: We can use the new `ignoreConflicts` from the `insert` method for a seamless operation. If a row already exists, it will return an empty list.

docs/design/reliable_future_calls.md
```diff
@@ -0,0 +1,107 @@
...
+Server crashes mid execution may allow multiple executions of a future call. Developers should be encouraged to make future calls idempotent.
+
+#### 2. Orphaned claims
+
+Claims may accumulate if:
+
+- Cleanup fails on startup
+- A server with claim to a future call becomes unavailable in a distributed environment
+
+A periodic maintenance job will be introduced delete claim entries that are older than a configured TTL. Developers will also be able to configure how often this maintenance job runs.
```
marcelomendoncasoares: Discussion: Do we need a separate scan for the cleanup? Can't the dispatcher be responsible by dealing with the orphaned/expired future calls? If the relation is exposed as an object relation on the `FutureCallEntry`, fetching existing claims would be feasible on a single query using includes.

One of the main benefits for this would be avoiding the introduction of another delay in the execution. We already have the delay caused by the scan interval, and having another (larger) one for recovery would introduce more uncertainty.

docs/design/reliable_future_calls.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Configurations
+
+Two new configurations will be introduced:
+
+- **futureCall.claimCleanupInterval**: Configures how often to run the future call claim cleanup job. Defaults to 30 minutes.
+- **futureCall.claimTTL**: Configures how long to keep future call claims before they become eligible for cleanup. Defaults to 10 minutes.
```
marcelomendoncasoares: Discussion: If a legitimate call takes more than 10 minutes, it will be considered expired and the claim can be removed mid-execution, allowing another server to claim and re-execute it. A TTL is indeed required, but we might need to expose it as an optional timeout parameter when creating the future call, so that users can tweak it according to each call.

There's also no distinction between "claim expired because server died" and "claim expired because the call is taking long". We might need to introduce a heartbeat mechanism to prevent this confusion. With the heartbeat:

- While running the future call, the server would have to constantly update the heartbeat column. Each update should be done only if the instance still possesses the claim for it. A low frequency (~1 min) would be enough without putting pressure on the database.
- When a running future call takes 3x the expected interval to update the heartbeat, its execution is considered stale (service degraded or server died). Then, another instance can replace the claim and start the execution again.
- If the update took too long because the service degraded and it restores after the claim has been replaced, the next update will verify that the claim has been replaced and cancel the ongoing execution in favor of the other that replaced it.

This does not replace the need for calls to be idempotent, as you have stated, but would make the lock mechanism more robust. It also does not replace the timeout, as a stale call might never attempt to do the next heartbeat to cancel itself - making the external timeout the only solution.

docs/design/reliable_future_calls.md
```diff
@@ -0,0 +1,107 @@
...
+fields:
+    ### The id of the future call this claim entry is associated with
+    futureCallId: int?, relation(parent=serverpod_future_call,onDelete=Cascade)
+
+    ### Timestamp of this claim entry
+    time: DateTime
+
+    ### The id of the server where the claim was created.
+    serverId: String
+```
```
marcelomendoncasoares, line 43: Question: Why add a 1:1 relation instead of just adding more columns to the `FutureCallEntry`? It is a non-breaking operation and reduces the need for an additional relation + cascade operation. Then we can use the new lock mechanism with `skipLocked` to do the operations on the entries.

docs/design/reliable_future_calls.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### 4. Crash recovery
+
+If a server crashes, the future calls and claims remain. A cleanup job removes all pending claims created by that server on start up so that the claimed future calls become eligible for execution again.
+
+### Potential Issues
+
+#### 1. Multiple execution
+
+Server crashes mid execution may allow multiple executions of a future call. Developers should be encouraged to make future calls idempotent.
```
marcelomendoncasoares, line 106: Comment: Cleaning up based on `serverId` is brittle because we can't assure that the server will start with the same ID - which could lead to orphaned entries of IDs that are never seen anymore. [My proposal with the heartbeat](https://github.com/serverpod/serverpod/pull/4799/changes#r2895917886) solves this as a side-effect, since the cleanup will be made by whichever instance finds the stale/expired future call first.

docs/design/reliable_future_calls.md
```diff
@@ -0,0 +1,107 @@
...
+The server executes the future calls if the insert is successful. The claim acts as a lease lock during execution. If the insert fails due to the unique constraint, another server has already claimed the call.
+
+#### 3. On successful execution
+
+The future call is permanently deleted from `serverpod_future_call` and `serverpod_future_call_claim` tables.
+
+```dart
+// deletion will cascade to the serverpod_future_call_claim table
+await FutureCallEntry.db.delete(_internalSession, futureCallEntry);
+```
```
marcelomendoncasoares, line 82: Question: Are we considering handling failed future calls as part of the scope? Would a "successful execution" in this context mean that the execution finished, regardless of whether it threw an error internally?

Crazelu: Successful execution in this context means that the execution finished regardless of whether it threw an error.
Handling of failed future calls is not part of the scope and that remains within the purview of developers.

packages/serverpod/lib/src/models/future_call_claim_entry.spy.yaml
```diff
@@ -0,0 +1,16 @@
...
+class: FutureCallClaimEntry
+table: serverpod_future_call_claim
+
+fields:
+  ### The id of the future call this claim entry is associated with
+  futureCallId: int?, relation(parent=serverpod_future_call, onDelete=Cascade)
+
+  ### Heartbeat timestamp for this claim entry.
+  ### Used to detect stale claims that should be cleaned up.
+  heartbeat: DateTime
```
marcelomendoncasoares: Suggestion: Call it `lastHeartbeatTime` to make the content clear.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -67,11 +68,13 @@ class FutureCallManager {
...
     required FutureCallSessionBuilder sessionProvider,
     required InitializeFutureCall initializeFutureCall,
+    Duration? heartbeatInterval,
   }) : _diagnosticsService = diagnosticsService,
        _internalSession = internalSession,
        _logSession = logSession,
        _sessionBuilder = sessionProvider,
-       _initializeFutureCall = initializeFutureCall {
+       _initializeFutureCall = initializeFutureCall,
+       _heartbeatInterval = heartbeatInterval ?? const Duration(seconds: 30) {
```
marcelomendoncasoares: Discussion: Can we use a more tolerant (and lighter) interval like 1 minute? This will reduce the weight of the heartbeat on the database and can make a different under heavy load with too many future calls.

packages/serverpod/lib/src/server/future_call_manager/future_call_scanner.dart
```diff
@@ -68,13 +74,26 @@ class FutureCallScanner {
...
       );
 
       entries.sort((a, b) => a.time.compareTo(b.time));
 
+      if (entries.isNotEmpty) {
+        // Delete future call claims with heartbeats that are behind by
+        // at least 2x the normal interval.
+        final staleClaimThreshold = DateTime.now().toUtc().subtract(
+          _heartbeatInterval * 2,
+        );
```
marcelomendoncasoares: Suggestion: Consider 3x the interval instead of 2x to make it more tolerant to punctual degradations. Since the interval is already small, 3x will still fast enough to prevent locking an orphaned call.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -229,6 +233,24 @@ class FutureCallManager {
...
+          _internalSession,
+          where: (t) => t.id.equals(claimId!),
+          columnValues: (t) => [t.heartbeat(DateTime.now().toUtc())],
+        );
+
+        if (rows.isEmpty) heartbeatTimer?.cancel();
+      } catch (_) {
+        heartbeatTimer?.cancel();
+      }
+    }
```
marcelomendoncasoares: Suggestion: Move this out of the body of `_runFutureCall` as a private function that receives the `claimId`, since the complexity of the run method is already very high. On the doc comment, it is worth documenting that the heartbeat is only updated if the claim is held by the running process.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -239,6 +261,31 @@ class FutureCallManager {
...
+        deleteFutureCallEntry = false;
+        return;
+      }
+
+      claimId = insertedClaims.first.id;
+
+      heartbeatTimer = Timer.periodic(
+        _heartbeatInterval,
+        (_) => updateHeartbeat(),
+      );
```
marcelomendoncasoares: Suggestion: Move this to a separate private method to reduce the complexity of the `_runFutureCall` method - and since it is a well scoped action.

packages/serverpod/lib/src/models/future_call_claim_entry.spy.yaml
```diff
@@ -0,0 +1,16 @@
+### Bindings to a future call claim entry in the database.
+class: FutureCallClaimEntry
+table: serverpod_future_call_claim
+
+fields:
+  ### The id of the future call this claim entry is associated with
+  futureCallId: int?, relation(parent=serverpod_future_call, onDelete=Cascade)
```
marcelomendoncasoares, line 7: Question: Why haven't you preferred an object relation, that facilitates extraction of related entities? It could also be good to add a backlink on the `FutureCallEntry` to allow extracting the claim together with the entries using an `includes` clause.

packages/serverpod/lib/src/server/future_call_manager/future_call_scanner.dart
```diff
@@ -68,13 +74,26 @@ class FutureCallScanner {
...
+
+        await FutureCallClaimEntry.db.deleteWhere(
+          _internalSession,
+          where: (t) => t.heartbeat < staleClaimThreshold,
+        );
```
marcelomendoncasoares: Suggestion: Deleting the claims eagerly can prevent a degraded server that recovers to continue the execution if the worker of the reclaiming node still hasn't started. The best would be to test and remove the stale claim inside the run method, right before inserting the reclaim. To make the operation atomic, the check-insert can use a transaction with lock-for-update / skip-locked.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -239,6 +261,31 @@ class FutureCallManager {
...
+        [claim],
+        ignoreConflicts: true,
+      );
+
+      if (insertedClaims.isEmpty) {
+        // Claim already exists. The instance with the
+        // claim will run the call and clean up after running.
+        deleteFutureCallEntry = false;
+        return;
+      }
```
marcelomendoncasoares: Suggestion: As [commented on the `deleteWhere` of the scanner](https://github.com/serverpod/serverpod/pull/4799/changes#r2978628537), this operation could be the responsible by removing the claim and inserting another inside a lock-for-update / skip-locked transaction.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -229,6 +233,24 @@ class FutureCallManager {
...
+    /// Update hearbeat column for the future call claim to keep claim alive.
+    Future<void> updateHeartbeat() async {
+      try {
+        final rows = await FutureCallClaimEntry.db.updateWhere(
+          _internalSession,
+          where: (t) => t.id.equals(claimId!),
+          columnValues: (t) => [t.heartbeat(DateTime.now().toUtc())],
+        );
+
+        if (rows.isEmpty) heartbeatTimer?.cancel();
```
marcelomendoncasoares: Suggestion: If we fail to update the heartbeat, we should try to abort the current execution. We can use @nielsenko suggestion of the [cancellation_token](https://pub.dev/packages/) to make the current task cancellable.

Crazelu: We won't be able to get the benefit of task cancellation from that library because the future will always start running before we need to cancel. Cancelling in that case will just throw an error and return early but the future call will still complete its run.

I didn't add this dependency because it doesn't offer much benefits here. If we really wanted to, we could use Completers to roll our own. But I don't think even that is necessary because the future will still complete.

Screenshot 2026-03-24 at 09 19 12

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: I chose a smaller interval to allow for quicker recovery. In this case, the max time future calls can be delayed if a worker dies is 1 minute (30 secs x 2). If we use 1 minute and 3x, that delay jumps to 3 minutes.

I agree with your suggestion to use a 1 minute interval to reduce the frequency of update queries for long running calls. However a potential 3 minute delay seems excessive. What do you think?

packages/serverpod/lib/src/server/future_call_manager/future_call_scanner.dart
Crazelu: Started a discussion for this in a related [conversation](https://github.com/serverpod/serverpod/pull/4799#discussion_r2979844285)

packages/serverpod/lib/src/models/future_call_claim_entry.spy.yaml
Crazelu: I do not see a need for extracting related claim with a FutureCallEntry. All of the interactions with the claim table are `insert` and `deleteWhere`. Fetching related claims with future call entries would have been preferred if that removed the need for additional queries.

packages/serverpod/lib/src/server/future_call_manager/future_call_scanner.dart
Crazelu: I agree on moving the deletion to the run method.
> To make the operation atomic, the check-insert can use a transaction with lock-for-update / skip-locked.

We can't use transactions because of concurrent future call execution. It'll throw on concurrent access to the database without the same transaction.

packages/serverpod/lib/src/models/future_call_claim_entry.spy.yaml
Crazelu: Done in [7a8bf98](https://github.com/serverpod/serverpod/pull/4799/commits/7a8bf989a33b90279252bd32c0e7a62da48b8101)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Done in [af777d7](https://github.com/serverpod/serverpod/pull/4799/commits/af777d77929b0cd715106165e3506cf04c115d60)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Done in [af777d7](https://github.com/serverpod/serverpod/pull/4799/commits/af777d77929b0cd715106165e3506cf04c115d60)

packages/serverpod/lib/src/server/future_call_manager/future_call_scanner.dart
Crazelu: Done in [af777d7](https://github.com/serverpod/serverpod/pull/4799/commits/af777d77929b0cd715106165e3506cf04c115d60)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Done in [af777d7](https://github.com/serverpod/serverpod/pull/4799/commits/af777d77929b0cd715106165e3506cf04c115d60) without transaction because that throws on concurrent database access when running multiple future calls concurrently.

packages/serverpod/lib/src/models/future_call_claim_entry.spy.yaml
marcelomendoncasoares: Ok, fair enough. My thought was that using the objects make the modeling clear - understand the claim model just by looking at the entry class -, but it is indeed not necessary.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -223,12 +226,93 @@ class FutureCallManager {
...
+    FutureCallEntry futureCallEntry,
+  ) async {
+    final entries = await _internalSession.db.transaction((transaction) async {
+      await FutureCallClaimEntry.db.lockRows(
+        _internalSession,
+        where: (t) => t.futureCallId.equals(futureCallEntry.id),
+        lockMode: LockMode.forUpdate,
+        lockBehavior: LockBehavior.skipLocked,
+        transaction: transaction,
+      );
```
marcelomendoncasoares: Discussion: Using `lockRows` with `skipLocked` on the `FutureCallClaimEntry` will result in the method proceeding even if the lock could not be acquired (server A goes stale, server B acquires lock for reclaim, server C also tries and skips). It would be valid to use `skipLocked` if locking `FutureCallEntry` with the `find` method, but it would lead to an edge case of preventing a stale-then-recovered server to delete the future call if it finishes while another server started reclaiming.

If we were to lock the rows explicitly, the best would be to use `noWait` to throw immediately if the lock could not be acquired. It would be a no-op if no claim exist and abort the operation automatically if another node already locked for claiming.

However, while reviewing this I realized that the `lockRows` is redundant and can be removed [due to Postgres already locking rows that match the `WHERE` when we do a `DELETE` operation](https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS:~:text=The%20commands%20UPDATE%2C%20DELETE%2C%20INSERT%2C%20and%20MERGE%20acquire%20this%20lock%20mode%20on%20the%20target%20table%20(in%20addition%20to%20ACCESS%20SHARE%20locks%20on%20any%20other%20referenced%20tables).%20In%20general%2C%20this%20lock%20mode%20will%20be%20acquired%20by%20any%20command%20that%20modifies%20data%20in%20a%20table.). Since the first operation on the transaction is a `deleteWhere`, it will provide the lock implicitly.

Then, if server B calls `deleteWhere` first on the stale claim row, server C's `deleteWhere` will block until B's transaction commits. After B commits, the stale row is gone, C's `deleteWhere` condition no longer matches, and C's `insert` conflicts - which is handled by `ignoreConflicts` to return empty.

We can further optimize this workflow by changing the `transaction` isolation level to `repeatableRead`. This will make C's `deleteWhere` to immediately throw a serialization error instead of waiting for B to commit - failing faster and releasing resources earlier. We can catch this exception to return null without ever trying the second insert.

It is worth adding a comment above each of these steps to clarify which mechanism is preventing which scenario for future developers - like the implicit lock that prevents concurrent reclaims and the choice for repeatable read to fail fast when locked by another server.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -223,12 +226,93 @@ class FutureCallManager {
...
+  }
+
+  /// Attempts to claim the execution for a [FutureCallEntry].
+  /// Any stale claim for the [FutureCallEntry] is deleted to allow
+  /// re-claiming.
+  ///
+  /// Returns a [Future] that:
+  /// - resolves to a heartbeat timer if a claim was successfully inserted
+  /// - resolves to null otherwise
+  Future<Timer?> _claimFutureCallAndStartHearbeatTimer(
```
marcelomendoncasoares: Small typo.

```suggestion
  Future<Timer?> _claimFutureCallAndStartHeartbeatTimer(
```

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -223,12 +226,93 @@ class FutureCallManager {
...
+      final rows = await FutureCallClaimEntry.db.updateWhere(
+        _internalSession,
+        where: (t) => t.id.equals(claimId),
+        columnValues: (t) => [t.lastHeartbeatTime(DateTime.now().toUtc())],
+      );
+
+      if (rows.isEmpty) timer.cancel();
+    } catch (_) {
+      timer.cancel();
+    }
```
marcelomendoncasoares: Question: With the cancel on the catch, any issue when trying to update the heartbeat - like a punctual connection issue to the database - would immediately make the future call stale. The cancel on the empty update is necessary (claim was lost scenario), but what are we preventing with the cancel on the catch? Should it be a no-op instead, just ignoring the exception and leaving the update to the next interval?

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -239,6 +323,17 @@ class FutureCallManager {
...
+      heartbeatTimer = await _claimFutureCallAndStartHearbeatTimer(
+        futureCallEntry,
+      );
+
+      if (heartbeatTimer == null) {
+        // Claim already exists. The instance with the
+        // claim will run the call and clean up after running.
+        deleteFutureCallEntry = false;
+        return;
+      }
```
marcelomendoncasoares: Suggestion: Move this block out of the try-catch block and as the first operation on the method. This will avoid creating/closing the `futureCallSession` and paying the deserialization cost if the future call is already claimed by other server. It will also render the `deleteFutureCallEntry` unnecessary and return eagerly.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
marcelomendoncasoares: I'd say that a 3 minute delay is still better because it allows a missed heartbeat or recovering from a punctual degradation without immediately losing claims to other nodes - duplicating the execution. Even though a quick recovery is good, frequently running calls twice under periods of heavy load might harm more than waiting a minute more to retry.

docs/design/reliable_future_calls.md
```diff
@@ -0,0 +1,116 @@
...
+
+#### 1. Multiple execution
+
+Server crashes mid execution may allow multiple executions of a future call. Developers should be encouraged to make future calls idempotent.
+
+#### 2. Orphaned claims
+
+A claim becomes orphaned when a server with claim to a future call becomes unavailable before the future call execution completes.
+
+The heartbeat mechanism ensures that orphaned claims are freed. However there will be a delay of 2x the heartbeat interval before orphaned claims can be freed for available servers to reclaim.
```
marcelomendoncasoares: Suggestion: Just a reminder to update the design docs with the latest changes after review.

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
```diff
@@ -0,0 +1,335 @@
...
+
+        test('then the FutureCall is executed', () async {
+          expect(testCall.counter, equals(1));
+        });
+
+        test('then the claim is deleted', () async {
+          final claims = await FutureCallClaimEntry.db.find(session);
+          expect(claims, isEmpty);
+        });
+      });
```
marcelomendoncasoares, line 172: Suggestion: Merge this group into the first top-level group, since they share the same "Given" and setup.

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
```diff
@@ -0,0 +1,335 @@
...
+    },
+  );
+
+  withServerpod('Given FutureCallManager with scheduled FutureCall that is due '
+      'and existing stale claim in the database for the FutureCall', (
+    sessionBuilder,
+    _,
+  ) {
```
marcelomendoncasoares: Suggestion(NIT): Remove this comma to follow the pattern with the other tests.

```suggestion
  withServerpod(
      'Given FutureCallManager with scheduled FutureCall that is due '
      'and existing stale claim in the database for the FutureCall',
      (sessionBuilder, _,) {
```

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
```diff
@@ -0,0 +1,335 @@
...
+        session,
+        where: (entry) => entry.name.equals(testCallName),
+      );
+      await session.close();
+    });
+
+    group('when running scheduled FutureCalls', () {
+      setUp(() async {
+        await futureCallManager.runScheduledFutureCalls();
+      });
```
marcelomendoncasoares: Suggestion: Also add a group similar to the first to test that, when start is called, the claim for the future call is replaced.

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
```diff
@@ -0,0 +1,335 @@
...
+          await futureCallManager.start();
+        });
+
+        tearDown(() async {
+          await futureCallManager.stop();
+        });
+
+        test('then a claim is inserted for the FutureCall', () async {
+          // Wait for future call execution to be scheduled
+          await Future.delayed(const Duration(seconds: 1));
```
marcelomendoncasoares: Suggestion: With the long running call taking 2 seconds and this delay of 1 second, it feels to me that this has the potential to become a flaky test very soon. A slightly greater delay between `setUp` and the `find` method can lead to the call having already finished and the claim removed.

Instead of relying on timers, it would be better to use a `Completer` inside the future call so we can control when it finishes from within the test.

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
```diff
@@ -0,0 +1,335 @@
...
+          final initialClaim = initialClaims.first;
+
+          expect(updatedClaim.id, equals(initialClaim.id));
+          expect(
+            updatedClaim.lastHeartbeatTime.isAfter(
+              initialClaim.lastHeartbeatTime,
+            ),
+            isTrue,
+          );
+        });
```
marcelomendoncasoares, line 351: Suggestion: This test also has potential for flakiness, since the timers are started on the group `setUp` and the delay between initial and update comparison is the exact heartbeat interval. If the second timer takes a bit to start, both can fall in the same window and be the same value. To mitigate this, wait at least 2x the heartbeat interval before finding the updated claim.

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
```diff
@@ -0,0 +1,335 @@
...
+      tearDown(() async {
+        await FutureCallEntry.db.deleteWhere(
+          session,
+          where: (entry) => entry.name.equals(testCallName),
+        );
+        await session.close();
+      });
+
+      group('when start is called', () {
+        setUp(() async {
```
marcelomendoncasoares, line 316: Suggestion: Also add test coverage for the heartbeat no longer being updated after the call is complete.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Good catch
It should be a no-op

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
Crazelu: The current formatting was done by Dart formatter.

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
Crazelu: After the future call is executed, the claim is deleted (and this is already covered) so no heartbeat update is possible.

docs/design/reliable_future_calls.md
Crazelu: Updated in [420fc6a](https://github.com/serverpod/serverpod/pull/4799/commits/420fc6adfe8d25aee13fdeeb259a507b23b44d9d)

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
Crazelu: Updated in [90276ca](https://github.com/serverpod/serverpod/pull/4799/commits/90276ca3b97c6ba37937c678407a6d69cad65ce9)

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
Crazelu: Updated in [90276ca](https://github.com/serverpod/serverpod/pull/4799/commits/90276ca3b97c6ba37937c678407a6d69cad65ce9)

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
Crazelu: Updated in [90276ca](https://github.com/serverpod/serverpod/pull/4799/commits/90276ca3b97c6ba37937c678407a6d69cad65ce9)

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
Crazelu: Updated in [90276ca](https://github.com/serverpod/serverpod/pull/4799/commits/90276ca3b97c6ba37937c678407a6d69cad65ce9)

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
Crazelu: Updated in [90276ca](https://github.com/serverpod/serverpod/pull/4799/commits/90276ca3b97c6ba37937c678407a6d69cad65ce9)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Updated in [90276ca](https://github.com/serverpod/serverpod/pull/4799/commits/90276ca3b97c6ba37937c678407a6d69cad65ce9)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Updated in [90276ca](https://github.com/serverpod/serverpod/pull/4799/commits/90276ca3b97c6ba37937c678407a6d69cad65ce9)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Updated in [90276ca](https://github.com/serverpod/serverpod/pull/4799/commits/90276ca3b97c6ba37937c678407a6d69cad65ce9)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -223,11 +226,101 @@ class FutureCallManager {
...
+      final claimId = entries.firstOrNull?.id;
+      if (claimId != null) {
+        return Timer.periodic(
+          _heartbeatInterval,
+          (timer) => _updateHeartbeat(claimId: claimId, timer: timer),
+        );
+      }
+    } catch (error, stackTrace) {
+      _diagnosticsService.submitFrameworkException(error, stackTrace);
+    }
```
marcelomendoncasoares: Suggestion: Explicitly catch the serialization exception when the repeatable read leads to a failure and return null. Otherwise we'll be dispatching error events to the server that should never reach the user.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -223,11 +226,101 @@ class FutureCallManager {
...
+    return null;
+  }
+
   /// Runs a [FutureCallEntry] and completes when the future call is completed.
   Future<void> _runFutureCall({
     required FutureCallEntry futureCallEntry,
     required FutureCall<SerializableModel> futureCall,
   }) async {
+    Timer? heartbeatTimer = await _claimFutureCallAndStartHeartbeatTimer(
```
marcelomendoncasoares: Suggestion(NIT): Declare as `final` since the variable is not mutated.

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
marcelomendoncasoares: Sorry, I didn't make myself clear on the suggestion. I meant about ensuring that the timer is disposed and no queries are being timely executed again. Because the update will do nothing, but no test ensures that timers are note being accumulated. Is it testable somehow?

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
```diff
@@ -0,0 +1,338 @@
...
+        );
+      });
+
+      tearDown(() async {
+        await FutureCallEntry.db.deleteWhere(
+          session,
+          where: (entry) => entry.name.equals(testCallName),
+        );
+        await session.close();
+      });
```
marcelomendoncasoares, line 313: Suggestion: On the `tearDown` of all tests with the completer, do a:

```dart
if (!testCall.completer.isCompleted) {
  testCall.completer.complete();
}
```

This will prevent tests from hanging eternally if any change on the code ever breaks them. Otherwise, if a test aborts on an expect, it will never reach the complete and prevent the teardown.

marcelomendoncasoares (changes requested):
Think we're very close now! Only a few small suggestions after the last review.

One last test that I realized we're missing is for reclaimed future calls from the perspective of a recovering server - server A lost the claim but recovered and tried to update the heartbeat after server B already reclaimed it. We test that server B manages to replace a stale lock, but we should also assert that server A aborts the run and does not mess with server B claim. Such test will require two concurrent managers to simulate real world concurrency.

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
Crazelu: I see
That can be tested by allowing tests to optionally provide a Timer that should be used.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Is there a specific exception class for this kind of error?
It seems to be generalized under `DatabaseDeleteRowException` which can represent other kinds of reasons why a delete operation failed. In that case, should all instances of that exception not be reported to the framework when trying to claim a future call?

marcelomendoncasoares: You can capture the `DatabaseQueryException` and test that the `code` matches `PgErrorCode.serializationFailure` (which maps to [`40001`, Postgres exception for serialization failure](https://www.postgresql.org/docs/18/mvcc-serialization-failure-handling.html#MVCC-SERIALIZATION-FAILURE-HANDLING)). This will prevent silently ignoring unrelated errors.

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
marcelomendoncasoares: Hmm... We should always try to avoid test-only overrides because then we're not actually testing the production implementation.

What if we add a list that holds active heartbeat timers and expose this list as visible for testing? Each timer should be added when created and removed when cancelled, with the list serving as a proof that we're cleaning up timers.

Crazelu: Done in [299cf51](https://github.com/serverpod/serverpod/pull/4799/commits/299cf51cddf5cdd233f242d5b88dffe0a230477a)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -223,11 +238,128 @@ class FutureCallManager {
...
+  /// Cancels [timer].
+  void _cancelTimer(Timer timer) {
+    final entry = _heartbeatTimers.where((t) => t.timer == timer).firstOrNull;
+    entry?.timer.cancel();
+    _heartbeatTimers.remove(entry);
+  }
+
+  /// Updates hearbeat for a future call claim to keep it alive.
+  /// Hearbeat is only updated if the claim is held by the
+  /// running process.
```
marcelomendoncasoares: Small typo.

```suggestion
  /// Updates heartbeat for a future call claim to keep it alive.
  /// Heartbeat is only updated if the claim is held by the
  /// running process.
```

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -50,7 +54,7 @@ class FutureCallManager {
   bool _hasPendingStart = false;
 
   /// Collection of active claim heartbeat timers.
-  final List<Timer> _heartbeatTimers = [];
+  final List<FutureCallClaimTimer> _heartbeatTimers = [];
```
marcelomendoncasoares: Question: Is this change really needed? Why can't we delete the timer directly as we were doing before?

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -232,23 +240,37 @@ class FutureCallManager {
...
+    entry?.timer.cancel();
+    _heartbeatTimers.remove(entry);
+  }
+
+  /// Cancels [timer].
+  void _cancelTimer(Timer timer) {
+    final entry = _heartbeatTimers.where((t) => t.timer == timer).firstOrNull;
+    entry?.timer.cancel();
+    _heartbeatTimers.remove(entry);
+  }
```
marcelomendoncasoares: Question: Can't we have only one method for cancelling the timer? If we manage to revert storing the `claimId` together, this wouldn't be needed.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -223,11 +238,128 @@ class FutureCallManager {
...
+        _heartbeatTimers.add((timer: timer, claimId: claimId));
+        return timer;
+      }
+    } on DatabaseQueryException catch (error) {
+      if (error.code == PgErrorCode.serializationFailure) {
+        return null;
+      }
+    } catch (error, stackTrace) {
+      _diagnosticsService.submitFrameworkException(error, stackTrace);
+    }
```
marcelomendoncasoares: Suggestion: Missing a `_diagnosticsService.submitFrameworkException` in case the database exception is for something else. Might be easier to merge both blocks:

```suggestion
    } catch (error, stackTrace) {
      if (error is DatabaseQueryException &&
          error.code == PgErrorCode.serializationFailure) {
        return null;
      }
      _diagnosticsService.submitFrameworkException(error, stackTrace);
    }
```

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
```diff
@@ -0,0 +1,488 @@
...
+        );
+      });
+    },
+    rollbackDatabase: RollbackDatabase.disabled,
+  );
+
+  withServerpod(
+    'Given two FutureCallManager instances with different heartbeat intervals'
+    'and a FutureCall that is due',
+    (sessionBuilder, _) {
```
marcelomendoncasoares, line 378: Suggestion: Describe the scenario we're testing, not the mocked parameters that we use to simulate it:

```suggestion
  withServerpod(
    'Given a future call manager running a FutureCall that gets stale and a concurrent future call manager representing another server node ',
    (sessionBuilder, _) {
```

Then, add a comment on the `futureCallManagerA` and `futureCallManagerB` creation indicating that we're using the heartbeat interval to simulate the call getting stale from the perspective of the other instance.

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
```diff
@@ -0,0 +1,488 @@
...
+        test('then futureCallManagerA claims execution', () async {
+          // Wait for future call execution to be scheduled
+          await Future.delayed(const Duration(milliseconds: 100));
+
+          // ignore: invalid_use_of_visible_for_testing_member
+          expect(futureCallManagerA.heartbeatTimers, hasLength(1));
+          // ignore: invalid_use_of_visible_for_testing_member
+          expect(futureCallManagerB.heartbeatTimers, isEmpty);
+          testCall.completer.complete();
+        });
```
marcelomendoncasoares: Suggestion: This has already been tested before and does not need to be part of this test. It is also contributing to make the descriptions fail the Given/when/then pattern (duplicated when).

You can call the `futureCallManagerA.start()` in the group `setUp` directly and leave `futureCallManagerB` to start on the acting.

tests/serverpod_test_server/test_integration/future_calls/future_call_claim_test.dart
```diff
@@ -0,0 +1,488 @@
...
+
+        group("when futureCallManagerA's claim gets stale", () {
+          test(
+            'then futureCallManagerB reclaims execution',
```
marcelomendoncasoares: Suggestion: This would then be the only acting, with the call to `futureCallManagerB.start()`: 

- when scanning due future calls on the concurrent node
  - then the concurrent node detects the stale heartbeat and reclaims the execution.
  - then the heartbeat update is aborted on the original node due to losing the claim.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: It is needed because the callback that is executed by Timer.periodic gets passed a `Timer` object that is not equatable to the one returned by `Timer.periodic` constructor. Even though both timers act on the same callback, they can't be used as equatable objects to remove elements from a list.

So associating an id with a timer makes it possible to cancel that timer from inside the `_updateHeartbeat` since it has access to the claim id.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: We can't for the same reason described [here](https://github.com/serverpod/serverpod/pull/4799#discussion_r2998180050)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
marcelomendoncasoares: Hmm... This does not seem entirely correct, as the `Timer` object passed around is the same and the `remove` will default to `identical` comparison when a type does not override `==` or `hashCode`.

From the changes, it seems that the real issue was a missing "remove from the list" after cancelling the timer whan no rows return from the heartbeat. From the previous code:

```dart
  Future<void> _updateHeartbeat({
    required int claimId,
    required Timer timer,
  }) async {
    try {
      final rows = await FutureCallClaimEntry.db.updateWhere(
        _internalSession,
        where: (t) => t.id.equals(claimId),
        columnValues: (t) => [t.lastHeartbeatTime(DateTime.now().toUtc())],
      );

      if (rows.isEmpty) timer.cancel(); // <-- Here was missing a remove from the list.
    } catch (error, stackTrace) {
      _diagnosticsService.submitFrameworkException(error, stackTrace);
    }
  }
```

And you added this remove together with the change to store the ID, fixing the issue as a colateral change:

```dart
  Future<void> _updateHeartbeat(int claimId) async {
    try {
      final rows = await FutureCallClaimEntry.db.updateWhere(
        _internalSession,
        where: (t) => t.id.equals(claimId),
        columnValues: (t) => [t.lastHeartbeatTime(DateTime.now().toUtc())],
      );

      if (rows.isEmpty) _cancelTimerForClaim(claimId); // <-- Cancel and remove after the change
    } catch (error, stackTrace) {
      _diagnosticsService.submitFrameworkException(error, stackTrace);
    }
  }
```

I'll revert and patch this quickly.

Issue #4551, opened 2026-01-08
Flaky `serverpod_cli` future call method test failing due to cross-contamination with other tests


marcelomendoncasoares:
It seems we have now a [new flaky test that is happening on Windows](https://github.com/serverpod/serverpod/actions/runs/20820749380/job/59808320099?pr=4548) after #4375, probably as a side-effect of #4378:

```console
==== Running tests in ./tools/serverpod_cli ====
+3408 -1: test\integration\analyzer\dart\future_call_validation\future_call_method_test.dart: Given a future call method without a first positional `Session` parameterand the first parameter instead contains an optional `Session` parameter when analyzed then validation error is reported. [E]
  Expected: an object with length of <1>
    Actual: []
     Which: has length of <0>
  
  package:matcher                                                                           expect
  test\integration\analyzer\dart\future_call_validation\future_call_method_test.dart 430:9  main.<fn>.<fn>
  
+3408 -2: test\integration\analyzer\dart\future_call_validation\future_call_method_test.dart: Given a future call method without a first positional `Session` parameterand the first parameter instead contains an optional `Session` parameter when analyzed then future call definition is created. [E]
  Expected: an object with length of <1>
    Actual: []
     Which: has length of <0>
  
  package:matcher                                                                           expect
  test\integration\analyzer\dart\future_call_validation\future_call_method_test.dart 438:9  main.<fn>.<fn>
  
+3408 -3: test\integration\analyzer\dart\future_call_validation\future_call_method_test.dart: Given a future call method without a first positional `Session` parameterand the first parameter instead contains an optional `Session` parameter when analyzed then no future call method definition is created. [E]
  Expected: empty
    Actual: <null>
  NoSuchMethodError: The getter 'isEmpty' was called on null.
  Receiver: null
  Tried calling: isEmpty at #0      Object.noSuchMethod (dart:core-patch/object_patch.dart:38:5)
  #1      _Empty.matches (package:matcher/src/core_matchers.dart:18:25)
  #2      _expect (package:matcher/src/expect/expect.dart:176:30)
  #3      expect (package:matcher/src/expect/expect.dart:65:3)
  #4      main.<anonymous closure>.<anonymous closure> (file:///D:/a/serverpod/serverpod/tools/serverpod_cli/test/integration/analyzer/dart/future_call_validation/future_call_method_test.dart:443:9)
[...]

PR #4803, opened 2026-03-05
fix: Fix flaky `serverpod_cli` integration test failing only on CI

marcelomendoncasoares:
Fixes: #4551

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/nested_template_macro_warning_test.dart
+++ b/tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/nested_template_macro_warning_test.dart
@@ -29,13 +29,13 @@ void main() {
     'Given an endpoint with nested template in class documentation when analyzed',
     () {
       var collector = CodeGenerationCollector();
-      late final testDirectory = Directory(
-        path.join(testProjectDirectory.path, const Uuid().v4()),
-      );
-
+      late Directory testDirectory;
       late List<EndpointDefinition> endpointDefinitions;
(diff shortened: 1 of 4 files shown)


tools/serverpod_cli/test/integration/analyzer/dart/endpoint_validation/nested_template_macro_warning_test.dart
```diff
@@ -213,7 +207,7 @@ class ExampleEndpoint extends Endpoint {
     'Given an endpoint with both nested template and nested macro when analyzed',
     () {
       var collector = CodeGenerationCollector();
-      var testDirectory = Directory(
+      late final testDirectory = Directory(
         path.join(testProjectDirectory.path, const Uuid().v4()),
       );
```
nielsenko: This is better than what was, but it would be even better if it happened inside a setUp (or setUpAll call - if you keep `final`). As is - this setup happens during test discovery, not during the test-run. Making test discovery slower and more brittle than it needs to be.

nielsenko: Same pattern in most other test groups in this PR.

marcelomendoncasoares: You're right on this. I tried to avoid a larger diff, but better make it right. Done at https://github.com/serverpod/serverpod/pull/4803/commits/5ce02a691b3dac792e57c2af8ccff7ed4d35a5d4.

PR #4806, opened 2026-03-06
feat: Implement presigned URL generation for S3 and native cloud storage

mathis6787:
### Description

Adds a `getPresignedUrl` method to the `CloudStorage` abstract interface and implements it across all storage backends. Presigned URLs allow temporary authenticated access to files without requiring the client to have direct credentials — useful for secure file downloads, browser-direct uploads, and short-lived sharing links.

### Changes

**Core** (`packages/serverpod`)
- Added abstract `getPresignedUrl` to `CloudStorage`
- `DatabaseCloudStorage` throws `UnsupportedError` (database storage has no concept of signed URLs)
- `StorageAccess` (on `Session`) exposes `getPresignedUrl` and `getPresignedUrls` for bulk lookups

**S3-compatible** (`integrations/serverpod_cloud_storage_s3_compat`)
- Added `getPresignedUrl()` to `S3Client` using AWS SigV4 query-string signing with `UNSIGNED-PAYLOAD`
- `S3CompatCloudStorage` delegates to `S3Client` — covers **AWS S3, GCP (HMAC), Cloudflare R2, MinIO, LocalStack** with a single implementation

**GCP native** (`integrations/serverpod_cloud_storage_gcp`)
- `NativeGoogleCloudStorage` implements `getPresignedUrl` using the existing GOOG4-RSA-SHA256 signing path

### Testing

- 12 unit tests for `S3Client.getPresignedUrl` (query params, expiration, methods, determinism, custom hosts)
- 2 tests for `S3CompatCloudStorage.getPresignedUrl` (file exists → URL returned, file missing → `null`)
- 2 tests for `DatabaseCloudStorage.getPresignedUrl` (throws `UnsupportedError` with correct message)

### Pre-launch Checklist

- [x] I have run `dart format .` on all changed files
- [x] I have run `dart analyze --fatal-infos` with no issues
- [x] I have written or updated tests for all changed functionality

### Breaking changes

None. This is an additive change only.

### DIFF
--- a/integrations/serverpod_cloud_storage_gcp/lib/src/native/native_google_cloud_storage.dart
+++ b/integrations/serverpod_cloud_storage_gcp/lib/src/native/native_google_cloud_storage.dart
@@ -307,6 +307,34 @@ class NativeGoogleCloudStorage extends CloudStorage
   }
 
   @override
+  Future<Uri?> getPresignedUrl({
+    required Session session,
+    required String path,
+    String method = 'GET',
+    Duration expiration = const Duration(minutes: 15),
+  }) async {
+    final signingContext = _signingContext;
+    if (signingContext == null) {
+      throw CloudStorageException(
+        'Cannot create presigned URL without signing credentials. '
+        'Ensure the storage was created with service account credentials.',
+      );
+    }
+
+    if (!await fileExists(session: session, path: path)) return null;
+
+    final signedUrl = await _createSignedUrl(
+      signingContext: signingContext,
+      bucket: bucket,
+      path: path,
+      expiration: expiration,
+      method: method,
+    );
+
+    return Uri.parse(signedUrl);
+  }
+
+  @override
   Future<bool> fileExists({
     required Session session,
     required String path,
--- a/integrations/serverpod_cloud_storage_s3_compat/lib/src/client/s3_client.dart
+++ b/integrations/serverpod_cloud_storage_s3_compat/lib/src/client/s3_client.dart
@@ -64,6 +64,79 @@ class S3Client {
     return _doSignedRequest(key: key, method: 'DELETE');
   }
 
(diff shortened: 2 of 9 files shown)


nielsenko:
Hi @mathis6787 

Generating temporary download URLs for private files is a reasonable feature request. However, this PR has several issues that would need to be addressed:

- Adding an abstract method to `CloudStorage` breaks all third-party implementations. The PR description says "no breaking changes" - this is incorrect.

- Supporting arbitrary HTTP methods (`PUT`, `DELETE`) via presigned URLs bypasses the existing upload lifecycle - no `verifyDirectFileUpload`, no `maxFileSize`, no `preventOverwrite`. If the need is download URLs, this should be scoped to GET only.

- The S3 implementation reimplements SigV4 signing from scratch instead of reusing the existing `_buildSignedRequestParams` infrastructure. On the GCP side `_createSignedUrl` is reused, but inconsistently.

- `DatabaseCloudStorage` throws `UnsupportedError`. The rest of the API returns `null` for unsupported operations.

- Checking file existence before generating the URL is a race condition that adds latency without real safety - the client will get a 404 from S3/GCS regardless (TOCTOU).

Before we invest more time on the API design here - what's your concrete use case?

marcelomendoncasoares:
> Before we invest more time on the API design here - what's your concrete use case?

@mathis6787 Following @nielsenko question, the best workflow here [following the contributing guidelines](https://github.com/serverpod/serverpod/blob/main/CONTRIBUTING.md#contributing-code) is to first open an issue with the request, describing your use case, so we can discuss on the best approach for a solution. This prevents a lot of back-and-forth during implementation.

mathis6787:
You're right I did the pr very quickly and I haven't review it before pushing it , I will rework that as soon as I can

Issue #4793, opened 2026-03-04
Adapt the migration system to generate and run client-side migrations


marcelomendoncasoares:
## Problem to Solve

To support client-side databases, the migration system must be adapted to analyze and generate migrations for the server and client databases independently. The current system runs based on files, but this is not suitable for the client without a complex asset management on Flutter apps.

The proposal for this feature touches several parts of the migration system, as detailed below.

## Create migration

Adapt the create migrations CLI commands to evaluate the server and the client in parallel. Migrations would be generated at once, sharing the same version if both succeeds. If one fails (due to requiring a `--force` parameter, for example), the other should still be generated successfully, mirroring the same experience with models and endpoints.

```
Creating migration (0.7s)
✓ Server migration created: migrations/20260302051713422
✗ Client migration skipped: No changes detected.
```

The flags `--server` and `--client` could be introduced later to allow running the command with only one database as target.

## Migration assets

To make migration artifacts easily available on Flutter, we could generate them as `.dart` files instead of `.txt`/`.sql`. This would make them importable with no asset management complexity and would work seamlessly in both Dart-only clients or Flutter apps.

A good strategy for the implementation would be to abstract reading and writing migrations, so that server and client can use their specific formats without changing the command logic.

### Migration registry

The `migration_registry.txt` would be replaced by a generated `migration_registry.dart` similar to:

```dart
// migrations/migration_registry.dart (generated)
library client_migrations;

part 'v20231205080937028/migration.dart';
part 'v20231205081959122/migration.dart';

abstract interface class ClientMigration {
  String get version;
  String get migrationSql;
  String get definitionSql;
}

/// Ordered list of all client migrations.
/// New migrations are always appended at the end.
class MigrationRegistry {
  static final List<ClientMigration> migrations = [
    _Migration20231205080937028(),
    _Migration20231205081959122(),
  ];

  static List<String> get versions =>
      migrations.map((m) => m.version).toList();
}
```

Since the `MigrationRegistry.migrations` would be a list where migrations are appended, conflicts would be detected in the same way as with the text file when merging conflicting changes.

### Migration version files

The migrations exposed on the `MigrationRegistry` would be generated as `part` files that implements the `ClientMigration` class:

```dart
// migrations/v20231205080937028/migration.dart (generated)
part of client_migrations;

class _Migration20231205080937028 implements ClientMigration {
  @override
  String get version => '20231205080937028';

  @override
  String get migrationSql => '''
[...]

Issue #4784, opened 2026-03-02
Detach the database from `serverpod` and into a `serverpod_database` package


marcelomendoncasoares:
To be able to generate a client-side database, Serverpod database classes and models must be accessible on the client. Currently, they live inside the `serverpod` package, which is not compatible with importing on Flutter.

Since #4748 isolated all database classes inside `serverpod/src/database` and #4737 introduced the possibility of having shared models, we can now move all database-related modules and classes into a `serverpod_package` as a shared package of `serverpod` itself. The package will depend only on `serverpod_serialization`, making it safe to be used on both the server and the client.

It is important to have platform-aware exports on this package to make it importable anywhere. Postgres, for example, should never be exported for web, since it imports `dart:io` and it would break Flutter build for web apps.

PR #4816, opened 2026-03-10
refactor: Decouple the migration artifacts read/write from the migration execution

marcelomendoncasoares:
Refactor the migration manager to decouple the way migration artifacts are read/written from the migration execution logic. This work serves two purposes:

- #4784: Will merge the database implementation that lives on `serverpod` with the one that lives on `serverpod_cli` into the new `serverpod_database`. This PR already lays the groundwork for the merge of the migration read from `serverpod` with the migration write from `serverpod_cli`.
- #4793: Will reuse the migration execution logic to implement client-side migrations with a completely different storage mechanism. The current PR enables this with ease - it will only require adding another migration artifact store implementation.

> NOTE: The write methods on this PR are not used yet and have no tests since they are meant to be used/tested as part of #4784 when moving the database code from the CLI to the new `serverpod_database` package.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod/lib/src/database/migrations/migration_artifacts_store/file_system.dart
+++ b/packages/serverpod/lib/src/database/migrations/migration_artifacts_store/file_system.dart
@@ -36,8 +36,7 @@ class FileSystemMigrationArtifactStore implements MigrationArtifactStore {
 
     return await migrationsDirectory
           .list()
-          .where((entity) => entity is Directory)
-          .cast<Directory>()
+          .whereType<Directory>()
           .map((directory) => path.basename(directory.path))
           .toList()
       ..sort();
@@ -225,4 +224,8 @@ class FileSystemMigrationArtifactStore implements MigrationArtifactStore {
   static String _encodeProtocolObject<T>(T value) {
     return SerializationManager.encode(value, formatted: true);
   }
+}
+
+extension<T> on Stream<T> {
+  Stream<U> whereType<U>() => where((t) => t is U).cast<U>();
 }


packages/serverpod/lib/src/database/migrations/migration_artifacts_store/file_system.dart
```diff
@@ -0,0 +1,234 @@
...
+    );
+
+    if (!await migrationsDirectory.exists()) {
+      return [];
+    }
+
+    return await migrationsDirectory
+          .list()
+          .where((entity) => entity is Directory)
+          .cast<Directory>()
```
nielsenko: Use `whereType<Directory>`

packages/serverpod/lib/src/database/migrations/migration_artifacts_store/file_system.dart
```diff
@@ -0,0 +1,234 @@
...
+    }
+  }
+
+  static Future<String?> _readFileIfExists(File file) async {
+    if (!await file.exists()) {
+      return null;
+    }
+
+    return file.readAsString();
+  }
```
nielsenko: This is only used on place. Seems a little excessive to split into separate function.

packages/serverpod/lib/src/database/migrations/migration_artifacts_store/file_system.dart
marcelomendoncasoares: I went directly for the same, but unfortunately there isn't. The `Directory.list` method returns a `Stream<FileSystemEntity>` (confusing, no?), which lacks the `whereType` convenience method that `Iterable`s have.

nielsenko: Ahh yes.. I should know.. I hit the same in the snip tool 😄 .

Perhaps put
```dart
extension<T> on Stream<T> {
  Stream<U> whereType<U>() => where((t) => t is U).cast<U>();
}
```
somewhere

packages/serverpod/lib/src/database/migrations/migration_artifacts_store/file_system.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4816/commits/ae00cd6c6e1af7114fe9cc896e1ecf4122c6da31.

packages/serverpod/lib/src/database/migrations/migration_artifacts_store/file_system.dart
marcelomendoncasoares: Good suggestion. I thought about doing and decided against it because it is only used once. But since you also passed through the same, I'll use the extension. Done at https://github.com/serverpod/serverpod/pull/4816/commits/52e2357b7997cdc881d1adcc2fe44ef9db7e45af.

PR #4832, opened 2026-03-12
feat: Upsert

### LINKED ISSUE #232: UPSERT support?
Postgresql has a cool ability to insert, but if there's a conflict then just update instead, known as Upsert.

https://wiki.postgresql.org/wiki/UPSERT
https://stackoverflow.com/questions/17267417/how-to-upsert-merge-insert-on-duplicate-update-in-postgresql

I see a function for `.insert` and `.update` but no `.upsert` and I went and looked at insert and it doesn't work like upsert by default.

I think it'd be a cool feature to have `.upsert` so I don't have to query first to see if it exists, then query again to put it in.

It would be cool but for now I'll do it the long way.

sedobrengocce:
## Description
Add upsert and upsertRow methods to the database layer, enabling insert-or-update operations based on unique column 
  conflicts.                                                                                                          
                                                  
###  Core implementation:
  - Add upsert (batch) and upsertRow (single) methods to Database and DatabaseConnection
  - Implement ON CONFLICT ... DO UPDATE SQL query generation in SqlQueryBuilder
  - Add PostgreSQL-specific exception handling for upsert operations
  - Add TestDatabaseProxy overrides so serverpod_test compiles correctly with the updated Database interface

###  Code generation:
  - Update RepositoryClasses generator to emit upsert/upsertRow for all models
  - Regenerate all test server model files with the new repository methods

###  Tests:
  - Add unit tests for upsert SQL query generation
  - Add integration tests covering single row upsert, batch upsert, conflict resolution, and error cases

  Fixes #232
  
  ## Pre-launch Checklist

  - [x] I read the Contribute page and followed the process outlined there for submitting PRs.
  - [x] I read and followed the Dart Style Guide and formatted the code with dart format.
  - [x] I updated/added relevant documentation (doc comments with ///), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.

##  Breaking changes

  None.

## Release Notes

**New Features**
- Added `upsert()` method to bulk insert or update multiple rows based on specified unique column constraints
- Added `upsertRow()` method for upserting a single row with optional transaction support
- Upserted rows return with IDs populated and non-persisted fields preserved

**Tests**
- Added comprehensive test coverage for upsert operations including batch and single-row scenarios, transaction handling, and conflict resolution


marcelomendoncasoares:
Thanks for this great contribution, @sedobrengocce!

We have just merged a refactor to the database layer introducing a `serverpod_database` package that is now decoupled from the `serverpod` package itself. Sorry for the pushing the merge conflicts, but would you be able to resolve them on this PR?

Then I'll ask our AI to do a first round of review and, after you addressing its relevant suggestions, I'll follow up with a manual review myself.

sedobrengocce:
@marcelomendoncasoares conflict solved

marcelomendoncasoares:
@coderabbitai Full review.

sedobrengocce:
@marcelomendoncasoares  reviwed codeRabbit suggestions

packages/serverpod_database/lib/src/interface/database_connection.dart
```diff
@@ -91,6 +91,22 @@ abstract class DatabaseConnection<D extends DatabasePoolManager> {
...
+    Transaction? transaction,
+  });
+
+  /// For most cases use the corresponding method in [Database] instead.
+  Future<T> upsertRow<T extends TableRow>(
+    DatabaseSession session,
+    T row, {
+    required List<Column> uniqueColumns,
+    Transaction? transaction,
+  });
```
marcelomendoncasoares, line 113: Suggestion: The current design does not cover all features that the `ON CONFLICT` syntax support and leaves out important mechanisms:

- Users might want to update only a subset of columns instead of all non-conflicting columns.
- It is frequent to use `WHERE` to filter when the conflict resolution should update values.
- The `uniqueColumns` name might imply that all unique columns must be passed.

To address the above topics, I propose we change the contract to the following:

```dart
Future<List<T>> upsert<T extends TableRow>(
  DatabaseSession session,
  List<T> rows, {
  required List<Column> conflictColumns, // Rename from `uniqueColumns` to be more idiomatic.
  List<Column>? updateColumns, // Allow specifying a non-empty subset of columns to update.
  Expression? conflictWhere, // Builds "ON CONFLICT DO UPDATE SET … WHERE …"
  Transaction? transaction,
})

Future<T> upsertRow<T extends TableRow>(
  DatabaseSession session,
  T row, {
  required List<Column> conflictColumns,
  List<Column>? updateColumns,
  Expression? conflictWhere,
  Transaction? transaction,
})
```

A null `updateColumns` should still resolve to all non-conflicting columns, but users will be able to restrict the updated columns if desired.

marcelomendoncasoares (changes requested):
Nice work! Before digging down on the tests, I think we still need to improve the design a bit to cover all important features of `ON CONFLICT DO UPDATE` operations.

marcelomendoncasoares:
@sedobrengocce The proposed interface is not fully implemented yet (missing the code generation).

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
```diff
@@ -210,6 +210,89 @@ class PostgresDatabaseConnection
...
+      if (transaction != null) {
+        return runSequentialUpserts(transaction);
+      }
+
+      return this.transaction<List<T>>(
+        (tx) => runSequentialUpserts(tx),
+        settings: const TransactionSettings(),
+        session: session,
+      );
+    }
```
marcelomendoncasoares, line 252: Use `DatabaseUtil.runInTransactionOrSavepoint` to simplify (and make it more robust). See https://github.com/serverpod/serverpod/pull/4909 for reference (also copy the test from that PR, since it is an important validation).

tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart
```diff
@@ -801,6 +803,151 @@ class BuildRepositoryClass {
...
+/// updated. Otherwise, a new row is inserted.
+///
+/// The returned [$className] will have its `id` field set.''')
+        ..name = 'upsertRow'
+        ..returns = TypeReference(
+          (r) => r
+            ..symbol = 'Future'
+            ..types.add(refer(className)),
+        )
+        ..requiredParameters.addAll([
```
marcelomendoncasoares, line 954: Missing all parameter of the new interface on the code generation.

marcelomendoncasoares (changes requested):

templates/serverpod_templates/projectname_server_upgrade/lib/src/greetings/greeting.spy.yaml
marcelomendoncasoares, line 1: Question: Why have this project changed?

packages/serverpod_database/lib/src/adapters/postgres/sql_query_builder.dart
```diff
@@ -420,6 +420,198 @@ SELECT * FROM insertWithIdNotNull
...
+  /// [updateColumns] optionally specifies which columns to update on conflict.
+  /// If null, all non-conflict, non-id columns are updated.
+  /// [conflictWhere] optionally adds a WHERE clause to the conflict resolution.
+  UpsertQueryBuilder({
+    required Table table,
+    required List<TableRow> rows,
+    required List<Column> conflictColumns,
+    List<Column>? updateColumns,
+    Expression? conflictWhere,
+  }) : _table = table,
```
marcelomendoncasoares: Suggestion: This class is largely redundant with the `InsertQueryBuilder`, since the upsert operation is just a special `onConflict` clause on the `INSERT`. Merge it with the `InsertQueryBuilder` using optional parameters to avoid unnecessary code duplication.

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
```diff
@@ -219,6 +219,89 @@ class PostgresDatabaseConnection
...
+    );
+
+    if (result.length != 1) {
+      throw _PgDatabaseUpsertRowException(
+        'Failed to upsert row, updated number of rows is ${result.length} != 1',
+      );
+    }
+
+    return result.first;
+  }
```
marcelomendoncasoares, line 304: Question: Shouldn't this be `Future<T?>` instead of throwing? If the `updateWhere` (replacing `conflictWhere`) is specified and the row does not match, the operation will skip all rows, leading to an empty list return. In this case, the user should receive `null` instead of an exception.

sedobrengocce:
The two failing CI checks on this PR are pre-existing CI flakiness, not regressions introduced by this change.

**Job 1 — `Serverpod integration tests (server) (3.32.0)`** (6 failed tests)

All 6 failures match exactly the pattern tracked in #4935 ("Flaky broken future call integration test failing with `Bad state: No element`"):

- `test_integration/future_calls/broken_future_calls_test.dart` lines 147, 210, 226, 299, 317
- `test_integration/future_calls/future_call_manager_test.dart` line 789

**Job 2 — `Serverpod concurrent integration tests (server) (3.32.0)`** (2 failed tests)

- `test_integration/future_calls/future_call_claim_test.dart:106` — `expect(claims, isEmpty)` after only a 100ms delay for async cleanup; race-sensitive on slow CI runners.
- `test_integration/test_tools/start_timeout_test/start_timeout_test.dart` "should timeout immediately" — `TimeoutException after 0:00:30` while waiting for a `dart test` subprocess to terminate.

No dedicated issue exists for the two concurrent failures, but the same matrix variant on Flutter 3.41.5 passed against the exact same code, which strongly suggests environmental flakiness rather than a code defect.

**Why this PR is unaffected:**

- The only `future_call_*` files modified are `packages/serverpod/lib/src/generated/future_call_claim_entry.dart` and `future_call_entry.dart`, which are **regenerated models** from the `regenerate all packages` commit — no future-call logic changes.
- Concurrent tests on Flutter 3.41.5 pass; only 3.32.0 fails.

A rerun of the failed jobs should clear them.

sedobrengocce:
@marcelomendoncasoares waiting for a new review

marcelomendoncasoares:
> @marcelomendoncasoares waiting for a new review

Hi @sedobrengocce! Sorry for the long time since last review... We were heads down an important milestone on the next release, but I'll get to your PR today.

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
```diff
@@ -219,6 +219,84 @@ class PostgresDatabaseConnection
     return result.first;
   }
 
+  @override
+  Future<List<T>> upsert<T extends TableRow>(
+    DatabaseSession session,
+    List<T> rows, {
+    required List<Column> conflictColumns,
+    List<Column>? updateColumns,
+    Expression? conflictWhere,
```
marcelomendoncasoares: Suggestion: During this review I realized that we can't call this parameter `conflictWhere` with the purpose it has right now. The more suitable name might actually be `updateWhere`, since it will filter the rows to update.

```suggestion
    required List<Column> conflictColumns,
    List<Column>? updateColumns,
    Expression? updateWhere,
```

A `conflictWhere` would be to generate a where query on the `ON CONFLICT` clause, like:

```sql
INSERT INTO upsert(key, val) VALUES(10, 'value')
ON CONFLICT (key WHERE is_active)
DO UPDATE SET val = EXCLUDED.val;
```

But we should not add this `conflictWhere` right now because it requires partial indexes, which Serverpod still does not support.

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
```diff
@@ -219,6 +219,84 @@ class PostgresDatabaseConnection
...
+        .toList();
+  }
+
+  @override
+  Future<T> upsertRow<T extends TableRow>(
+    DatabaseSession session,
+    T row, {
+    required List<Column> conflictColumns,
+    List<Column>? updateColumns,
+    Expression? conflictWhere,
```
marcelomendoncasoares:
```suggestion
    Expression? updateWhere,
```

packages/serverpod_database/lib/src/concepts/exceptions.dart
```diff
@@ -64,8 +64,16 @@ abstract base class DatabaseDeleteRowException implements DatabaseException {
...
+abstract base class DatabaseUpsertRowException implements DatabaseException {
+  @override
+  String toString() => 'DatabaseUpsertRowException: $message';
+
+  /// Thrown when SQLite [PRAGMA foreign_key_check](https://www.sqlite.org/pragma.html#pragma_foreign_key_check)
+  /// reports one or more rows that violate foreign key constraints.
+}
+
+/// Exception thrown when SQLite foreign key constraints are violated.
 final class SqliteForeignKeyViolationException implements DatabaseException {
```
marcelomendoncasoares, line 75: Suggestion: This change messed up the doc comment of the `SqliteForeignKeyViolationException`.

packages/serverpod_database/test/database_query/upsert_sql_query_test.dart
marcelomendoncasoares, line 1: Suggestion: After merging the `UpsertQueryBuilder` with the insert, remove this file and move the upsert-related tests to the `insert_sql_query_test.dart`.

tests/serverpod_test_server/test_integration/database_operations/crud/upsert_test.dart
```diff
@@ -0,0 +1,233 @@
...
+    });
+
+    test(
+      'when upserting a row with the same unique value then the existing row is updated.',
+      () async {
+        var result = await UniqueData.db.upsertRow(
+          session,
+          UniqueData(number: 42, email: 'existing@serverpod.dev'),
+          conflictColumns: (t) => [t.email],
+        );
```
marcelomendoncasoares: Suggestion: We need to improve the coverage on this feature a bit. Also add tests for:

- Using `conflictTarget` on column without unique index.
- Using multiple columns on `conflictTarget`.
- Choosing the `conflictTarget` on a model with more than one unique index.
- Passing an invalid `conflictTarget` (empty list, id column, column from related model).
- Upserting does not change columns that are not part of the `updateColumns` list.
- Upserting with `upsertWhere` (replaces `conflictWhere`) does not change rows that do not meet the where clause.

marcelomendoncasoares (changes requested):
Sorry again for the long delay and thank you for your patience, @sedobrengocce! Have a few comments for improvement, specially on the test coverage, and a few questions.

marcelomendoncasoares:
@sedobrengocce I took the liberty to make some small changes after your last review to get the branch ready, hope you don't mind. Let's just wait for CI to finish and we're good to merge! While this, would you open a PR on the [serverpod_docs](https://github.com/serverpod/serverpod_docs) repository documenting the new `upsert` and `upsertRow` methods?

marcelomendoncasoares (approved):
Thank you for your contribution with this much requested feature, @sedobrengocce! 🚀

Issue #4786, opened 2026-03-03
Remove Postgres-specific elements from the definition JSON files


marcelomendoncasoares:
## Problem to Solve

Even though the `definition.json`/`migration.json` files are mostly agnostic to the database, there are two elements that are specific to Postgres and create a blocker to properly implement other database engines:

- The `defautlValue` field of `ColumnDefinition` classes is stored with the actual value that goes in the SQL. We already have abstract defaults on the `.spy` models, but **they are translated to Postgres values before being written** to the definition files. Examples:
  - For int PK: `nextval('table_name'::regclass)`
  - For BigInt: `'12345678901234567890'::text`
  - For DateTime: `'2024-05-24 22:00:00'::timestamp without time zone`.
  - etc.
- The entry for the automatic `*_pkey` index that Postgres creates for every table. This is actually an inconsistency, since such entries are not declared by the user, are skipped on the SQL generation and exist only for the analyzer to be happy in runtime.

The main issue is that such values are also saved inside the `Protocol.targetTableDefinitions` of the modules, and this field is imported as part of the `Protocol.targetTableDefinitions` of projects. If the project uses a different database dialect, suddenly its definition JSON files and `Protocol.targetTableDefinitions` contains mixed default values and index representations from Postgres dialect - likely used on the module. The same can also happen if a module developer generates the definitions using other database dialect.

## Proposal

The key principle of the solution is that **nothing dialect-specific should ever be deserialized from JSON** - only serialized *into* the final SQL files.

### Default values

The fix is to push the abstract defaults that we already have on the `.spy` files down to the definition JSON files and translate them to SQL while writing the SQL files. The same translation will be required while analyzing the database, which will temporarily mean a code duplication until the database is made a shared package with #4784.

### Primary key indexes

Since such indexes are:

1. Automatically created by the database engine (`*_pkey` on Postgres, `sqlite_autoindex_*` on SQLite).
2. Not declared by the user on models.
3. Not included and executed in the SQL files.

They could be completely removed from the files and be skipped in the analyzer. This will align with the definition files being a JSON representation of the elements expressed in the models and present in the SQL files.

### Backwards compatibility

To introduce this change without breaking the migration system, we can add a `schemaVersion` field to the definition files. Files with no `schemaVersion` would pass through an in-memory decoding of the default values and removal of the automatic index definitions. Files with the key matching the expected version would be parsed directly.

Issue #4785, opened 2026-03-03
Implement SQLite database support


marcelomendoncasoares:
The main purpose of #4748 was to allow an easy addition of new database engines - with the most demanded being SQLite. As the database is already abstracted behind interfaces, it is only required to add a new `DatabaseDialect` entry and then implement the highlighted interfaces.

## Pool manager

For the interaction with SQLite, we can use the battle-tested [sqlite_async](https://pub.dev/packages/sqlite_async) package that already does all the heavy-lifting on the database management level. If query introspection is needed, the [sqlparser](https://pub.dev/packages/sqlparser) package is also worth using - both a remarkable work from [Simon Binder](https://github.com/simolus3) with SQLite3 support on Dart/Flutter.

## Restrictions

We should aim to support all column types on SQLite, even when advanced query operation might not be available - like with `Vector` columns. This will provide a predictable experience for users, specially considering the objective of using SQLite for offline-first support. It is acceptable for operations that are not supported to throw in runtime - such as trying to filter a `Vector` column based on the distance.

For indexes, on the other hand, we should forbid the usage of any other type than `btree` - the only one supported by SQLite. Otherwise it would be impossible for the analyzer to correctly match expected indexes. This is easy to implement because the `config` object has access to the `databaseDialect` while evaluating restrictions.

## Migrations

Since SQLite3 has only few basic columns types available, the implementation must have special care to ensure that migrations take into consideration the model type changes and not only the column type. Otherwise, a field type change that is stored using the same database type would not be migrated and would produce deserialization issues in runtime. The premise is to preserve the type-safety of the models, and not the column types only.

## Query builder

Given that "SQLite implementation strives to be compatible with PostgreSQL" [[1]](https://sqlite.org/lang_update.html#:~:text=SQLite%20implementation%20strives%20to%20be%20compatible%20with%20PostgreSQL) [[2]](https://sqlite.org/forum/forumpost/41017a8da1?t=h&unf#:~:text=%23%23%20Why%20SQLite%20chooses%20the%20PG%20syntax%0A%0A%20%201.%20%20SQLite%20has%20traditionally%20followed%20PG%20syntax%20whenever%20practical.), we might be able to vastly reuse the Postgres SQL query builder. Most of the syntax we use is valid SQLite syntax, with only a few exceptions that can be worked around (like the writable CTEs when mixing rows without and with ID). The work on #4764 was already a preparation for this.

## Testing

[...]

Issue #4783, opened 2026-03-02
☂️ Generate a client-side SQLite database using Serverpod models


marcelomendoncasoares:
### Problem to Solve

When working with Serverpod on the server, developers that need offline support have to rely on other packages, like Drift, which introduces asymmetry in the interaction with the same relational data between server and client, duplicated models, more documentation context, different migration systems, etc.

### Proposal

Offer a fully integrated developer experience by exposing the same `Model.db` on the client to interact with a local SQLite database. The `Client` would expose a way to create a long-living session and migrations should share the same workflow as migrations on the server. The implementation can be broken out into several steps that will be discussed individually as sub-issues covering:

1. Moving the database to a package that can be used on the client.
2. Implementing SQLite database support.
3. Adapting the CLI to generate client-side repositories.
4. Exposing a `DatabaseSession` on the client.
5. Adapting the migration system to generate and run client-side migrations.

### Use Case

A developer wants to build an offline-first application with the same Serverpod ORM experience across the client and server.

### Alternatives

Use other packages on the client and deal with the difference in the semantics.

### Additional context

Related issues and discussions:

- https://github.com/serverpod/serverpod/issues/160
- https://github.com/serverpod/serverpod/issues/1719
- https://github.com/serverpod/serverpod/discussions/1977
- https://github.com/serverpod/serverpod/discussions/2084

This issue is one big step towards seamlessly synchronizing data between client and server database for a true offline-first experience.

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [x] I want to work on this

PR #4836, opened 2026-03-12
feat: Move all database-related code from `serverpod` to the new `serverpod_database`

marcelomendoncasoares:
Closes: #4784

This PR moves all database-related code from the `serverpod` package to the new `serverpod_database` shared package. This means that the database API will be usable anywhere, including client side (as part of #4783) and the CLI (enables some features from #4604).

The `serverpod_package` is safe to import on any platform or web. Currently, the `DatabaseProvider` throws `UnsupportedError` for all dialects until SQLite is introduced in #4785.

## Pending work

Although part of the database-related code from the CLI was already merged with this PR, the migration SQL generation is still left due to the dependency on the internal definition classes for the column default values. This will be fixed after #4786 is merged, since it will remove such default representation from the definition classes.

## Pre-launch Checklist

## Breaking changes

None, since all public classes are re-exported from the `serverpod` and `serverpod_service_client` packages.

### DIFF
--- a/packages/serverpod_database/lib/src/migrations/migration_artifacts_store/file_system.dart
+++ b/packages/serverpod_database/lib/src/migrations/migration_artifacts_store/file_system.dart
@@ -204,11 +204,10 @@ class FileSystemMigrationArtifactStore implements MigrationArtifactStore {
   }
 
   static Future<String?> _readFileIfExists(File file) async {
-    if (!await file.exists()) {
-      return null;
-    }
-
-    return file.readAsString();
+    try {
+      return await file.readAsString();
+    } on PathNotFoundException catch (_) {} // ignore
+    return null;
   }
 
   static Future<String> _readRequiredFile(File file) async {


packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
```diff
@@ -1011,13 +1010,12 @@ class PostgresDatabaseConnection
...
 
     if (hasNonPersisted && rows.length > 100) {
-      session.log(
+      session.logWarning?.call(
         'WARNING: Inserting ${rows.length} rows with ignoreConflicts on '
         'table "${table.tableName}" with non-persistent fields. This '
         'requires individual inserts and may cause performance issues. '
         'Consider removing non-persistent fields or inserting in smaller '
         'batches.',
-        level: LogLevel.warning,
```
nielsenko, line 1020: There is a subtle change in semantics here we should be aware of. Now `rows.length` and `table.tableName` will only be evaluated if `session.logWarning` is not `null`.  

This is fine and preferable for sure, but it does make the codepath change depending on logging. 

Just something to be aware of.

packages/serverpod/lib/database.dart
```diff
@@ -1,24 +1,40 @@
...
+export 'package:serverpod_database/serverpod_database.dart'
+    hide
+        BulkData,
+        BulkDataException,
+        BulkQueryColumnDescription,
+        BulkQueryResult,
+        ColumnDefinition,
+        ColumnMigration,
+        ColumnType,
+        DatabaseBulkData,
```
nielsenko, line 10: Why all these hides?

packages/serverpod_database/lib/src/migrations/migration_artifacts_store/file_system.dart
```diff
@@ -202,12 +202,18 @@ class FileSystemMigrationArtifactStore implements MigrationArtifactStore {
     }
   }
 
+  static Future<String?> _readFileIfExists(File file) async {
+    if (!await file.exists()) {
+      return null;
+    }
+
+    return file.readAsString();
+  }
```
nielsenko, line 211: To protect against TOCTOU you may want to do:
```dart
    try {
      return await file.readAsString();
    } on PathNotFoundException catch (_) {} // ignore
    return null;
```

packages/serverpod_database/analysis_options.yaml
```diff
@@ -4,8 +4,13 @@ linter:
   rules:
     # Temporary disable the following so that each can be fixed one by one.
     avoid_escaping_inner_quotes: false
-    directives_ordering: false
+    directives_ordering: true
```
nielsenko, line 7: 👍

packages/serverpod_database/analysis_options.yaml
```diff
@@ -4,8 +4,13 @@ linter:
...
+    prefer_relative_imports: true
     prefer_final_in_for_each: false
     prefer_final_locals: false
     prefer_final_parameters: false
-    sort_pub_dependencies: false
+    sort_pub_dependencies: true
```
nielsenko, line 12: 👍

packages/serverpod_database/analysis_options.yaml
```diff
@@ -4,8 +4,13 @@ linter:
   rules:
     # Temporary disable the following so that each can be fixed one by one.
     avoid_escaping_inner_quotes: false
-    directives_ordering: false
+    directives_ordering: true
+    prefer_relative_imports: true
     prefer_final_in_for_each: false
     prefer_final_locals: false
```
nielsenko, line 10: Should we switch these as well?

packages/serverpod/lib/src/database/adapters/postgres/database_connection.dart
marcelomendoncasoares: The `rows.length` are actually always evaluated since the condition to log uses it. For the `table.tableName`, it is actually not a problem, since `tableName` is a field and not a getter.

```dart
    if (hasNonPersisted && rows.length > 100) {
      session.logWarning?.call(
        'WARNING: Inserting ${rows.length} rows with ignoreConflicts on '
        'table "${table.tableName}" with non-persistent fields. This '
        'requires individual inserts and may cause performance issues. '
        'Consider removing non-persistent fields or inserting in smaller '
        'batches.',
      );
    }
```

packages/serverpod/lib/database.dart
marcelomendoncasoares: Just to maintain the same exposure as previously. When they were part of the `serverpod` package, these classes were not exported as public interfaces. Since `serverpod_database` is declared as an internal package, by hiding these classes from the public export of the `serverpod` package, we're free to change them without worrying about breaking changes.

packages/serverpod_database/analysis_options.yaml
marcelomendoncasoares: The directives sorting was a no-brainer since I had to change imports completely, but I preferred avoiding these others since they will result in a way larger diff on this PR - and also push a lot of conflicts on ongoing contributions to the database layer. Maybe a better timing to do it is on the next formatting changes.

packages/serverpod_database/lib/src/migrations/migration_artifacts_store/file_system.dart
marcelomendoncasoares: Good catch! Done at https://github.com/serverpod/serverpod/pull/4836/commits/8f6c89314b46a1966d92fb91ef1055b6e55f17c9.

PR #4842, opened 2026-03-13
fix: Use no-cache for all Flutter web assets

### LINKED ISSUE #4709: Enhancement: Include main.dart.js in FlutterRoute default no-cache list for safer Flutter web updates
## Problem to Solve

`FlutterRoute` already avoids caching critical Flutter web files, but in production deployments there can still be temporary stale-asset mix cases where icons or font glyphs do not render correctly after updates.

A representative console warning is:

`Could not find a set of Noto fonts to display all missing characters.`

## Proposal

Please consider including `main.dart.js` in the default no-cache list used by `FlutterRoute`, alongside the current critical files.

If changing defaults is not preferred, please document an official recommended `cacheControlFactory` override pattern for production deployments that want stricter update safety.

## Use Case

In real deployments (with existing service worker state), users may refresh around release time and briefly get mixed old/new assets. This can cause missing icons or incorrect font behavior until cache state fully converges.

Making `main.dart.js` default no-cache (or clearly documenting the override) would reduce this edge case significantly for teams using standard `FlutterRoute`.

## Alternatives

- Keep current default behavior and let users provide custom `cacheControlFactory` manually.
- Disable PWA/service worker completely (not always desirable).
- Rely only on deployment-level cache headers and atomic deploy procedures.

## Additional context

- Serverpod version used: `3.2.3`
- Flutter app served under `/app` via `FlutterRoute`
- Related previous work:
  - Issue: https://github.com/serverpod/serverpod/issues/4452
  - PR: https://github.com/serverpod/serverpod/pull/4454

I am reporting this as an enhancement request, not a regression bug.

### LINKED ISSUE #5077: The `FlutterRoute` does not correctly handle caching and cache busting
Currently, the Flutter app is cached for one day and does not update with the latest version of the app. There are two issues with this behavior:

1. Users won't get the latest version when the app is updated.
2. Caching for only one day is not very helpful as the full app will be frequently loaded anyway.

**Suggested quick fix**
Do not cache the files at all. This will solve the first issue, but not the second.

**Proposal for a real fix**
Serve the files from a sub-route (this can potentially be done through a middleware or through building with different targets). The app page stays the same, but it points to different Flutter-related content based on the version of the Flutter app and its build number.

For instance, instead of serving `/app/flutter.js`, we serve `/app/1.2.0+23/flutter.js`. The `/app` route (the index file) has no caching, so we always get a fresh one.

Set the cache time to a very long time (like a year).

nielsenko:
Default to `private, no-cache` (ETag revalidation) for all files served by `FlutterRoute`.

Flutter's service worker has been deprecated since 3.32 ([flutter#156910](https://github.com/flutter/flutter/issues/156910)) and is now a self-destructing stub that provides no caching for either WASM or JS builds. Since Flutter's build output uses fixed filenames (`main.dart.wasm`, `main.dart.js`) with no content hashing, ETag revalidation is the only safe default to prevent serving stale applications after a rebuild.

Hence the default cache policy for `FlutterRoute` changed from selective `private, no-cache`  (critical files only) + `public, max-age=1day` (all other files) to `private, no-cache` for all files. Users who relied on the previous aggressive caching default should pass a custom `cacheControlFactory`.

The `cacheControlFactory` and `cacheBustingConfig` parameters are retained so users who post-process build output to embed content hashes can opt into aggressive caching.

Fixes #4709
Fixes #5077
## Pre-launch Checklist

## Breaking changes

This is **NOT** considered a breaking change as it will at most cause a few more roundtrips. Users can add their own `cacheControlFactory` to change the behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated default caching behavior for static file serving to use more conservative cache control settings (private, no-cache). Custom cache strategies remain configurable.
  * Removed `noCacheFiles` constant from public API.

[Review Change Stack](https://app.coderabbit.ai/change-stack/serverpod/serverpod/pull/4842)

### DIFF
--- a/packages/serverpod/lib/src/web_server/routes/flutter_route.dart
+++ b/packages/serverpod/lib/src/web_server/routes/flutter_route.dart
@@ -9,40 +9,56 @@ import 'package:serverpod/serverpod.dart';
 /// Combines static file serving with automatic fallback to index.html for
 /// client-side routing, plus WASM headers (COOP/COEP) for multi-threading.
 ///
-/// By default, critical Flutter files (index.html, flutter_service_worker.js,
-/// flutter_bootstrap.js, manifest.json, version.json) are served with no-cache
-/// headers to prevent stale app manifests and service workers. All other files
-/// can be cached according to browser heuristics or custom cache control.
-///
-/// To invalidate the cache, change the version in your pubspec.yaml file and
-/// build your Flutter project.
-///
 /// ```dart
 /// pod.webServer.addRoute(
 ///   FlutterRoute(Directory('web/flutter_app')),
 /// );
 /// ```
+///
+/// ## About Caching ..
+///
+/// By default, all files are served with `private, no-cache` headers so the
+/// browser always revalidates with the server via ETag. This is the only safe
+/// default because:
+///
+/// - Flutter's service worker is deprecated and will be removed
+///   (see [flutter#156910](https://github.com/flutter/flutter/issues/156910)).
+///   Since Flutter 3.32, the generated service worker is a self-destructing
+///   stub that provides no caching for either WASM or JS builds.
+/// - Flutter's build output uses fixed filenames (`main.dart.wasm`,
+///   `main.dart.js`, `main.dart.mjs`) with no content hashing, so aggressive
+///   `max-age` caching risks serving stale applications after a rebuild.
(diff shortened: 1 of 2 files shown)


nielsenko:
@coderabbit review

marcelomendoncasoares (approved):
👍

PR #4846, opened 2026-03-13
feat: Add Tencent COS cloud storage foundation

Metaildust:
## Summary
- Add the new `serverpod_cloud_storage_cos` package, template entry, and publishable package registration so Tencent COS can be reviewed as a first-class provider in the monorepo.
- Implement `CosCloudStorage` with COS-native signing and server-side `storeFile`, `storeFileWithOptions`, `retrieveFile`, `fileExists`, `deleteFile`, and `getPublicUrl` support.
- Map `preventOverwrite` to `x-cos-forbid-overwrite`, validate `contentLength` on the server upload path, and add tests covering the signer, public URL rules, path normalization, and the intentionally deferred direct upload APIs.

## Overall Rollout
- This integration is intentionally split into two phases to keep the provider contract reviewable before the direct upload flow is added.
- Phase 1 establishes the official package shape, native COS signing, and the server-side CRUD baseline.
- Phase 2 will add direct upload descriptions, `FileUploader` / upload verification closure, broader integration coverage, and the final user-facing README details.

## Phase 1 Scope
- Use COS-native XML API signing instead of `serverpod_cloud_storage_s3_compat`, since COS requires its own signature model and COS-specific overwrite semantics.
- Support official-style credential loading via `SERVERPOD_COS_ACCESS_KEY_ID` / `COSAccessKeyId` and `SERVERPOD_COS_SECRET_KEY` / `COSSecretKey`, while keeping the legacy aliases as compatibility fallbacks.
- Leave `createDirectFileUploadDescription` and `createDirectFileUploadDescriptionWithOptions` intentionally unimplemented for now; both return `null` and are covered by tests so the current boundary is explicit.

## Test plan
- [x] `dart format lib/ test/`
- [x] `dart analyze`
- [x] `dart test`

### DIFF
--- a/integrations/serverpod_cloud_storage_cos/pubspec.yaml
+++ b/integrations/serverpod_cloud_storage_cos/pubspec.yaml
@@ -3,7 +3,7 @@
 
 name: serverpod_cloud_storage_cos
 description: Serverpod integration for Tencent Cloud Object Storage (COS)
-version: 3.4.3
+version: 3.4.4
 repository: https://github.com/serverpod/serverpod
 homepage: https://serverpod.dev
 issue_tracker: https://github.com/serverpod/serverpod/issues
@@ -14,12 +14,12 @@ environment:
 dependencies:
   crypto: ^3.0.2
   http: '>=1.1.0 <2.0.0'
-  serverpod: 3.4.3
+  serverpod: 3.4.4
 
 dev_dependencies:
   lints: ">=3.0.0 <7.0.0"
   mocktail: ^1.0.0
-  serverpod_client: 3.4.3
+  serverpod_client: 3.4.4
   test: ^1.25.5
 
 dependency_overrides:


Metaildust:
I took a look at the remaining failing checks after the `LICENSE` fix landed.

`Publish check` is now passing. The remaining red jobs appear to be unrelated to this PR and are all failing in the Firefox-based E2E setup:
- `Serverpod E2E web tests (3.32.0)`
- `Serverpod E2E IO tests (3.32.0)`
- `Serverpod E2E web tests (3.38.4)`
- `Serverpod E2E IO tests (3.38.4)`

They fail while building the test Docker image in `util/run_tests_e2e`, specifically at:

```sh
RUN apt-get install -y --no-install-recommends firefox
```

The failing logs report:

```text
target tests: failed to solve: process "/bin/sh -c apt-get install -y --no-install-recommends firefox" did not complete successfully: exit code: 100
```

Since the failure happens before the new COS package is exercised, this looks like a repository / CI environment issue rather than a regression introduced by this PR.

Metaildust:
## Phase 2 summary

This phase closes the remaining Tencent COS integration gaps without expanding the scope to multipart uploads, STS credentials, or stronger verify semantics.

- fixed `publicHost` leakage so all server-side `PUT/GET/HEAD/DELETE` operations and direct-upload descriptions always target the COS bucket endpoint, while `getPublicUrl()` remains the only `publicHost` consumer
- added native COS integration tests covering CRUD, byte preservation, 1 MB uploads, direct upload via `FileUploader`, `verifyDirectFileUpload`, `contentLength`, `preventOverwrite`, and special paths
- wired COS into `INTEGRATION_TESTING.md`, `util/run_tests_integration_cloud`, and the GitHub Actions cloud storage integration jobs
- introduced an injectable `http.Client` seam, expanded unit coverage for CRUD / error mapping / host routing, and added a provider-level `close()` that only releases internally owned clients
- documented credential fallback, bucket naming, `publicHost` semantics, and the versioning / CORS / CAM constraints that are specific to COS

## Verification

- `dart analyze`
- `dart format --output=none --set-exit-if-changed .`
- `dart test`
- native COS integration tests are now wired for CI and run when the COS test secrets are available

Metaildust:
@nielsenko Ready for reviewing

Metaildust:
One documentation-related note: I do have more detailed usage documentation in my community package docs, and I think parts of that can be useful as source material for future official docs.

The current package name there is [`tencent_cos_cloud_storage_serverpod`](https://pub.dev/packages/tencent_cos_cloud_storage_serverpod), previously published as [`serverpod_cloud_storage_cos`](https://pub.dev/packages/serverpod_cloud_storage_cos). The repository is still here as well: [`Metaildust/serverpod_cloud_storage_cos`](https://github.com/Metaildust/serverpod_cloud_storage_cos#readme).

That said, I do not think it should be copied over directly as-is, because there are a few important differences from the official implementation in this PR:

- the old community package used a custom Serverpod upload endpoint flow, while this official package now uses the standard `FileUploader` + presigned PUT flow
- some configuration names differ (`customDomain` vs `publicHost`, and the official env / password alias naming)
- in the main Serverpod repo, detailed package usage seems better suited for dartdoc / centralized docs than for the generated package README template

Metaildust:
There is one remaining package-ownership issue we need to resolve before the official `serverpod_cloud_storage_cos` package can be published on pub.dev.

I have already sent an email from the original package-owner address `themetaildust@gmail.com` to `info@serverpod.dev` to formally document the transfer discussion.

At the moment, I see two possible solutions:

1. Temporarily grant my account admin access to `serverpod.dev` so I can complete the one-time transfer, and then remove that access immediately afterward.
2. Contact `support@pub.dev` together and ask whether they can help facilitate a consensual transfer to the `serverpod.dev` publisher.

If there is any trust concern around temporary publisher admin access, I recommend option 2. In practice, a jointly-sent email is usually the safer path for completing this kind of transfer.

Please reply by email as soon as possible. We should keep the discussion on email because, whether the package is transferred via pub.dev support or via a temporary admin-permission workflow, it is best to preserve a traceable written record to reduce legal and ownership risk.

Metaildust:
hi @nielsenko. If it's busy, please let me know when will this review be handled, so that I can do my schedule. Thanks!

PR #4856, opened 2026-03-18
refactor: Remove database-specific default values from the definition files

marcelomendoncasoares:
Closes: #4786

With this PR, default values are now stored with nearly the same format as they are defined in the models. The only difference is for types that will be stored as text on all databases, such as `BitInt` and enums with values serialized by name. For such, the default value is stored as a quoted text literal to match database introspection.

A normalization is done on previous migration files to remove the database-specific default values when comparing database definitions, so that the change is backwards compatible.

## Pre-launch Checklist

## Breaking changes

None, the refactor is backwards compatible.

### DIFF
--- a/examples/legacy/chat/chat_server/lib/src/generated/protocol.dart
+++ b/examples/legacy/chat/chat_server/lib/src/generated/protocol.dart
@@ -37,7 +37,7 @@ class Protocol extends _i1.SerializationManagerServer {
           columnType: _i2.ColumnType.bigint,
           isNullable: false,
           dartType: 'int?',
-          columnDefault: 'nextval(\'channel_id_seq\'::regclass)',
+          columnDefault: 'serial',
         ),
         _i2.ColumnDefinition(
           name: 'name',
@@ -53,21 +53,7 @@ class Protocol extends _i1.SerializationManagerServer {
         ),
       ],
       foreignKeys: [],
-      indexes: [
-        _i2.IndexDefinition(
-          indexName: 'channel_pkey',
-          tableSpace: null,
-          elements: [
-            _i2.IndexElementDefinition(
-              type: _i2.IndexElementDefinitionType.column,
-              definition: 'id',
-            ),
-          ],
-          type: 'btree',
-          isUnique: true,
-          isPrimary: true,
-        ),
-      ],
+      indexes: [],
       managed: true,
     ),
     ..._i3.Protocol.targetTableDefinitions,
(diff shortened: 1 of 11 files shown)


packages/serverpod_database/lib/src/extensions.dart
```diff
@@ -3,8 +3,8 @@ import 'package:collection/collection.dart';
 import '../serverpod_database.dart';
 import 'migrations/table_comparison_warning.dart';
 
-/// Comparison methods for [DatabaseDefinition].
-extension DatabaseComparisons on DatabaseDefinition {
+/// Utility methods for [DatabaseDefinition].
+extension DatabaseDefinitionUtils on DatabaseDefinition {
```
nielsenko, line 7: Is `Utils` a slightly unconventional suffix for an `extension`? I have seen `Ex` and `Extension` used other places, but we are actually a bit all over the place.

packages/serverpod_database/lib/src/adapters/postgres/postgres_analyzer.dart
```diff
@@ -82,18 +83,18 @@ ORDER BY ordinal_position;
...
-            columnDefault: e[1],
-            columnType: ExtendedColumnType.fromSqlType(e[3]),
-            // SQL outputs YES or NO. So we have to convert it to a bool manually.
-            isNullable: e[2] == 'YES',
-            vectorDimension: e[4],
-          ),
-        )
-        .toList();
+    return queryResult.map((e) {
+      var columnType = ExtendedColumnType.fromSqlType(e[3]);
```
nielsenko: Not sure I prefer the block body syntax here, but these kind of changes makes it a bit noisy.

packages/serverpod_database/lib/src/adapters/postgres/postgres_default_value.dart
```diff
@@ -0,0 +1,68 @@
...
+
+  final literal = StringBuffer("'");
+  for (var i = 1; i < sql.length; i++) {
+    final char = sql[i];
+    if (char == "'") {
+      if (i + 1 < sql.length && sql[i + 1] == "'") {
+        literal.write("''");
+        i++;
+        continue;
+      }
```
nielsenko, line 54: This handles '' escaping (SQL92), what about double-dollar (`SELECT $$It's a "sunny" day, isn't it?$$;`) or escaped string (`E'foo\nbar'`) - given this is postgres specific? Do we need to handle that?

packages/serverpod_database/lib/src/extensions.dart
marcelomendoncasoares: My idea with the `Utils` was intentional to indicate that it contains only helpers (like the find and contains methods). The goal is to keep it more scoped than a generic `Ex` or `Extension` that would accept anything related to the class, not only helper/utilitiy methods.

packages/serverpod_database/lib/src/adapters/postgres/postgres_default_value.dart
marcelomendoncasoares: We don't need to handle those because we only allow users to specify single quoted strings on the models. Double-dollar or escaped strings would only be possible by manually creating default values with raw SQL, which is very far from expected - would break the analyzer either way.

Issue #4862, opened 2026-03-19
Relations on id fields are discarded


Crazelu:
### Describe the bug

Relations defined on id fields are correctly parsed by the analyzer but are removed by `ModelDependencyResolver` before database migrations are generated resulting in migrations that don't include the defined relations.

### To reproduce

Create a model definition:

```yaml
class: FutureCallClaimEntry
table: serverpod_future_call_claim

fields:
  ### The id of the future call this claim entry is associated with
  id: int?, relation(parent=serverpod_future_call, onDelete=Cascade)

  ### Heartbeat timestamp for this claim entry.
  ### Used to detect stale claims that should be cleaned up.
  heartbeat: DateTime
```

Create migrations:

```sh
serverpod create-migration
```

Generated `migrations.sql` file does not create a foreign relationship.

### Expected behavior

The defined relation should be created

### Library version

3.4.4

### Platform information

All platforms

### How experienced are you with this library?

None

### Are you interested in working on a PR for this?

- [x] I want to work on this

marcelomendoncasoares:
As we have discussed, to move this work onward, we'll first need to remove the existing premise that the `id` field always require a default value - which does not make sense when it assumes the value of the PK of another table.

PR #4863, opened 2026-03-19
fix: Add parsed relations to id fields

Crazelu:
This PR fixes an issue where relations defined on id fields are discarded by `ModelDependencyResolver`.

Fixes [4862](https://github.com/serverpod/serverpod/issues/4862)

## Pre-launch Checklist

## Breaking changes

None


tools/serverpod_cli/test/analyzer/models/stateful_analyzer/stateful_analyzer_test.dart
```diff
@@ -424,4 +425,50 @@ and neither is this line
       );
     },
   );
+
+  test(
+    'Given valid yaml models with one-to-one relationship on the id field, when validating all,'
+    'then the class with the relation is serialized containing an id field with valid relation.',
```
marcelomendoncasoares, line 668: Suggestion(NIT): Separate the sentence on the clauses to follow the pattern of other tests (with a trailing space for the sentence to form correctly):

```suggestion
  test(
    'Given valid yaml models with one-to-one relationship on the id field, '
    'when validating, '
    'then the class with the relation is serialized containing an id field with valid relation.',
```

Crazelu:
> it is also worth adding more tests on the different types of relations (like object relations using the id field as the field parameter, named bidirectional relationships, etc.) 

It is currently possible to add object and named bidirectional relationships using the id field as the field parameter without this fix and they won't be discarded. That works because the relations are not directly defined on the id field.
Existing tests (that don't use id field as field parameter) already cover this behaviour.

This fix is specifically for relations directly added on the id field, so they are not discarded. And that can only be foreign relations.

Other kinds of relations cannot be defined directly on the id field like a foreign key relation. If users attempt to define them directly on id field, the analyzer throws expected errors. We don't have tests for these error cases specific to id field relations so that is a good area to increase coverage.

Hanibachi:
been waiting for this.

tests/serverpod_test_server/test_integration/database_operations/id_field_one_to_one_relation/create_test.dart
marcelomendoncasoares, line 1: Suggestion: Combine both tests in one file directly under `database_operation` since they are very small.

tests/serverpod_test_server/test_integration/database_operations/id_field_one_to_one_relation/create_test.dart
```diff
@@ -0,0 +1,40 @@
...
+      session,
+      where: (_) => Constant.bool(true),
+    );
+    await Task.db.deleteWhere(
+      session,
+      where: (_) => Constant.bool(true),
+    );
+  });
+
+  group(
```
marcelomendoncasoares, line 20: Suggestion: Use the test framework (`withServerpod`) instead, since the test does not require more advanced control over the test server.

I know this breaks the pattern with surrounding tests, but this is exclusively due to them being old and using a pattern that we need to gradually replace.

tests/serverpod_test_server/test_integration/database_operations/id_field_one_to_one_relation/create_test.dart
```diff
@@ -0,0 +1,40 @@
...
+    'Given models with a 1:1 relation on id field',
+    () {
+      test(
+        'when inserting an entry in the related table with an id that is not in the parent table, '
+        'then the insertion fails',
```
marcelomendoncasoares, line 25: Suggestion: Also add a test that validates the `id` of the related class to be the same as the parent class.

tests/serverpod_test_server/lib/src/models/id_field_one_to_one_relation/task.spy.yaml
```diff
@@ -0,0 +1,5 @@
+class: Task
+table: task
+fields:
+  name: String
+  time: DateTime
```
marcelomendoncasoares, line 5: Suggestion: Also add a test model with `UuidValue` `id` field and an integration test for it - to ensure it works for native `int` field and for changed `id` types as well.

PR #4867, opened 2026-03-20
perf: Improve incremental analysis speed

nielsenko:
WIP


nielsenko:
This is now included in #4788.

PR #4889, opened 2026-03-25
feat: Add dedicated support for recurring future calls

### LINKED ISSUE #4728: Introduce a dedicated method for creating recurrent future calls
### Problem to Solve

The current scheduling system only truly supports one-time future task execution. There is [a documented way to create recurring tasks by scheduling the next on the beginning of the current](https://docs.serverpod.dev/3.2.0/concepts/scheduling/recurring-task), but it is kind of a hack that requires multiple steps to be implemented.

Other problems of this approach are:

- **Exact periodicity can't be ensured with duration-based intervals:** since the start of each task can have a few seconds of delay (due to the scanning interval) and the stacked delay will compound into a considerable drift over time.
- **Complex custom logic is necessary to implement non-duration based periodic scheduling:** It is possible to use the `callAtTime` to schedule to exact intervals (like at every first day of the month), but it requires complex custom logic.
- **The recurrence can be lost if one future call is cancelled before scheduling the other:** Since the recurrence is created inside the future call, if it is picked up and cancelled before registering the next, the entire recurrence just vanishes silently (as reported at https://github.com/serverpod/serverpod/issues/3417).

### Proposal

Introduce a dedicated interface and mechanism to schedule recurrent future calls. The API must be well thought to support multiple interval configuration strategies (duration based, pinned to day/month/week/etc) to achieve a cron-like experience - but with a human-centered design:

- Every N minutes/hours/days
- At specific times (e.g. every day at midnight)
- On specific days of the week or month

Special care must be taken regarding missed executions (e.g. that should have run when the server was down). Maybe a way to handle it is by exposing a configuration on whether the server should run them immediately or drop them completely.

### Use Case

A developer wants to:

- Send daily summary emails to users every day at midnight.
- Clean up expired sessions every hour.
- Sync data from external services every 5 minutes.

Currently, each of these requires complex custom logic or external schedulers. A built-in solution would make these use cases easy and reliable.

### Alternatives

Use the current documented alternative.

### Additional context

This issue is a detailed specification derived from #899.

- [ ] I want to work on this

Crazelu:
This PR adds dedicated support for recurring future calls. Recurring future calls can be scheduled using cron expressions or Dart's `Duration` objects.

Closes [4728](https://github.com/serverpod/serverpod/issues/4728)

## Pre-launch Checklist

## Breaking changes

None


docs/design/recurring_future_call.md
```diff
@@ -0,0 +1,113 @@
...
+
+```yaml
+### Represents different future call types.
+enum: FutureCallType
+serialized: byIndex
+default: oneOff
+values:
+  - oneOff
+  - recurring
+```
```
marcelomendoncasoares: Suggestion: We don't need the `FutureCallType` if the `cron` field is nullable and will only have a value for recurrent future calls. Then we also prevent the null-unsafe `cron!` call if the `FutureCallType` is recurrent.

docs/design/recurring_future_call.md
```diff
@@ -0,0 +1,113 @@
...
+### Database Schema
+
+The `serverpod_future_call` table schema will be extended to include two new fields.
+
+```yaml
+class: FutureCallEntry
+table: serverpod_future_call
+fields:
+  ### Type of future call
+  type: FutureCallType
```
marcelomendoncasoares: Comment: Just a note (for the future, since we might not need this field): when adding fields to existing models in a backwards-compatible way, they must either be nullable or use `default=`. Otherwise, we break deserialization of older objects and incur in destructive migrations to recreate the tables.

docs/design/recurring_future_call.md
```diff
@@ -0,0 +1,113 @@
...
+FutureCallSchedule.everyHour(12); // * 12 * * *
+FutureCallSchedule.daily(); // 0 0 * * *
+FutureCallSchedule.annually(); // * * 1 1 *
+```
+
+These utility methods will return valid cron expressions that interoperate with the `recurring` method.
+
+```dart
+await pod.futureCalls.recurring(FutureCallSchedule.daily()).example.runTask();
+```
```
marcelomendoncasoares: Discussion: Although supporting cron is definitely needed (and robust for complex cases), it can also be restrictive due to cron only supporting intervals that divide evenly. Given that we are in Dart, it would be nice to also support specifying the recurrency using `Duration` objects - just as the `callWithDelay`.

Also, instead of adding another object, we could expose the type-safe convenience methods directly to the user as we do with the one-off calls. Combining both suggestions, I'd propose something like:

```dart
await pod.futureCalls
  .callRecurring()
  .cron("0 * * * *")
  .example
  .runTask();

// Positive side-effect: one method covers all possible intervals instead of one per unit.
await pod.futureCalls
  .callRecurring()
  .every(const Duration(minutes: 15))
  .example
  .runTask();

// An optional `shift` parameter would allow controlling the start time.
await pod.futureCalls
  .callRecurring()
  .every(const Duration(minutes: 15), shift: const Duration(seconds: 10))
  .example
  .runTask();
```
> Changed `recurring` to `callRecurring` to follow the `callAtTime`/`callWithDelay` pattern, but it's also up for discussion.

To store either the cron or the interval-based in a type-safe way, we could change the new `cron` field to scheduling and store a [sealed `FutureCallScheduling` class](https://docs.serverpod.dev/concepts/models#sealed-classes) that is subclassed into `CronFutureCallScheduling` and `IntervalFutureCallScheduling`. With [polymorphism support](https://docs.serverpod.dev/concepts/models#polymorphism-support), this leaves room for introducing other scheduling syntaxes in the future while allowing us to switch on the options to calculate the next run.

docs/design/recurring_future_call.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Execution Flow
+
+Existing logic for scanning and scheduling future calls will remain unchanged and only add a special case handling for running recurring future calls.
+
+After running a recurring future call, the entry's `time` must be updated with the next run time derived from parsing the cron expression. The next run time is computed relative to the current time and not the last scheduled time for the `FutureCallEntry`. This is to ensure the next time is always in the future.
+
+Unlike one-off calls, recurring future calls will not be deleted after the execution is completed.
```
marcelomendoncasoares: Discussion: Not removing the future calls upon completion would change the concept of the `FutureCallEntry` from one-per-run to either be a run or a recurrency spec that is also the current run. Besides mixing the concepts, this seems to introduce more changes and complexity than just creating a new entry for the next execution. Instead, we simplify a lot if we just create another `FutureCallEntry` using `copyWith` to persist the `cron`/`scheduling`. Then, testing will be as simple as ensuring that the entries are created, without worrying about deletions, handling claims, etc.

An important topic for discussion here is whether we should compute the next run **before** or **after** executing the current. If a given call takes longer than the scheduling interval (either due to punctual fluctuations or a misconfiguration), we will be choosing between:

- **after**: we skip runs that would trigger while the current is running. This is a simple model and effectively mean that we restrict the concurrency for the same recurring call to 1, but it leaves users with no guarantee that the interval will be honored. If we go this way, we need to document it clearly and ensure that the next call is created before the current is deleted.
- **before**: this is today's recommended model and a more robust one, which ensures no run is lost, but can also result in overlapping executions if calls take longer than they should. For this model to be safe, we need to expose per-call configurations to handle the overlap (concurrency limit, wait, abort).

Since we're aiming to replace the current recurrency model, I think we should aim for **creating the next entry before running** and use the `callRecurring` method suggested above to expose how to handle overlapping executions.

marcelomendoncasoares (changes requested):
Nice design, @Crazelu! A few topics for discussion to make the feature more robust.

docs/design/recurring_future_call.md
Crazelu: This is a great suggestion.
I was originally skeptical about this because of how `Duration` would not correctly serialize to a cron expression. The `sealed FutureCallScheduling class` idea is nice for this.

I'm biased to cron because this is the standard across other frameworks and languages so I expect developers to be more familiar with it. Supporting `Duration` would probably be a new and niche experience for developers who have worked with recurring tasks before.

My primary reason for not designing for `Duration` is the serialization but we can use polymorphism indeed.
I still think that it'll be useful to have utility methods that produce commonly used cron expressions.

marcelomendoncasoares: I agree that many that are already familiar with cron will prefer it, but I'd argue kind of the opposite: `Duration` feels a lot more native and type-safe than cron for most of the basic scheduling. And, since we have a lot of users that are not that familiar with back-end development, it might be even more used. Either way, supporting both is the way to go to make everyone happy.

Regarding helpers to build cron, I just think that there are already a ton out there - either [Dart packages (hron)](https://pub.dev/packages/hron) or online tools - and we can leave users choose what fits best on their workflow.

docs/design/recurring_future_call.md
Crazelu: > Not removing the future calls upon completion would change the concept of the FutureCallEntry from one-per-run to either be a run or a recurrency spec that is also the current run.

I agree. Deleting will prevent some edge cases especially with claims.

For the `before` or `after` options for computing the next run time, I preferred the `after` for its simplicity. 

The `before` model is definitely more robust and can perhaps be a subsequent improvement especially because the overlap handling will require some additional changes to the scheduler such that it also knows the name associated with the tasks it runs.

Alternatively, we can go for the `before` model with the default overlap handling and then improve on this subsequently.

docs/design/recurring_future_call.md
Crazelu: Right.
To refine the suggestion, I find an optional `DateTime startAt` to be more intuitive:

```dart
final date = DateTime.now().add(Duration(hour: 1));
final start = DateTime(date.year, date.month, date.day, date.hour);

// An optional `startAt` parameter would allow controlling the start time.
await pod.futureCalls
  .callRecurring()
  .every(const Duration(hours: 1), startAt: start)
  .example
  .runTask();
```

This will allow for developers to explicitly set the start time and be closer to cron-like evenly divided time if they prefer.

docs/design/recurring_future_call.md
marcelomendoncasoares: As discussed, we can implement the next future call scheduling `before` executing, but postpone handling the overlapping calls as an enhancement - stating on the docs that it is currently on the user to handle it, as they already have to with the current way to do recurring tasks.

docs/design/recurring_future_call.md
marcelomendoncasoares: Agreed, the `startAt` parameter is more intuitive and more robust, since it allows users to schedule the first run of a recurring sequence for a given point in the future.

If a given `startAt` happen to be in the past (due date of first run passed or the server was down when it was due), we should run immediately and calculate the next run to be the first one of the sequence that would fall in the future. Since future calls do not receive a scheduled parameter, it would be pointless to trigger N runs (at the same time, all with same parameter) just to comply with the number of calls that would have happened in the interval that the server was off.

packages/serverpod/test/server/cron_test.dart
```diff
@@ -0,0 +1,120 @@
...
+        () => Cron.parse('invalid'),
+        throwsA(isA<CronFormatException>()),
+      );
+
+      expect(
+        () => Cron.parse('* * *'),
+        throwsA(isA<CronFormatException>()),
+      );
+    },
+  );
```
marcelomendoncasoares: Suggestion: Missing a `Given` for this group of tests.

packages/serverpod/lib/src/server/future_call_manager/cron.dart
```diff
@@ -0,0 +1,246 @@
...
+// ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+// WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+// DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
+// DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+// (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+// LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+// ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
```
marcelomendoncasoares: Suggestion: Follow the same format used on `serverpod_serialization/lib/src/pgvector.dart` to add a disclaimer first that the current file only is under a different license:

```dart
/*
 * This file and all contents under `src/pgvector` and `test/pgvector` folders
 * are cloned from the original pgvector package and licensed under the MIT
 * License. Source: https://github.com/pgvector/pgvector-dart
 *
 * The scope of the below license ("Software") is limited to the files referred
 * above, which is a derivative work of the original package. The license does
 * not apply to any other part of the codebase.
 *
 *
 * The MIT License (MIT)
 *
 * Copyright (c) 2023-2024 Andrew Kane
 * ...
```

packages/serverpod/lib/src/server/future_call_manager/cron.dart
```diff
@@ -0,0 +1,246 @@
...
+          .map(parseField)
+          .expand((list) => list!)
+          .toSet()
+          .toList();
+      items.sort();
+      return items;
+    }
+
+    final singleValue = int.tryParse(constraint);
+    if (singleValue != null) return [singleValue];
```
marcelomendoncasoares: Suggestion: This method seems a little convoluted. I know it is from the cron package, but we can improve it now that we'll maintain it. Maybe splitting this function in smaller private parts and adding a doc comment might be enough to improve the readability.

packages/serverpod/lib/src/server/future_call_manager/cron.dart
```diff
@@ -0,0 +1,246 @@
...
+
+    var intervalPart = '';
+    if (constraint.contains('/')) {
+      final parts = constraint.split('/');
+      if (parts.length > 2) {
+        throw CronFormatException(
+          'Invalid cron expression. More than one `/` for intervals.',
+        );
+      }
+      // ignore: parameter_assignments
```
marcelomendoncasoares: Question: Do we need this comment? I think we don't enforce this rule, right?

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -372,9 +377,58 @@ class FutureCallManager {
...
+          final now = DateTime.now().toUtc();
+          if (start != null && start.isBefore(now)) {
+            final difference = now.difference(start);
+            final count = (difference.inMilliseconds / interval.inMilliseconds)
+                .ceil();
+            return now.add(interval * count);
+          }
+          return now.add(interval);
+        }(),
+    };
```
marcelomendoncasoares: Suggestion: Use a regular `switch` with an inline `return` on each block to avoid the IIFE for interval.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -372,9 +377,58 @@ class FutureCallManager {
...
+      IntervalFutureCallScheduling(interval: var interval, start: var start) =>
+        () {
+          final now = DateTime.now().toUtc();
+          if (start != null && start.isBefore(now)) {
+            final difference = now.difference(start);
+            final count = (difference.inMilliseconds / interval.inMilliseconds)
+                .ceil();
+            return now.add(interval * count);
+          }
+          return now.add(interval);
```
marcelomendoncasoares: Suggestion: For interval-based recurrent future calls with no `start`, if we schedule the next based on `now`, we'll be introducing a drift that will compound over time - specially due to the scan delay. The safest path is to use `start ?? futureCallEntry.time` and always apply the same diff/count algorithm.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -372,9 +377,58 @@ class FutureCallManager {
...
+  Future<void> _scheduleNextRecurringCall(
+    FutureCallEntry futureCallEntry,
+    DateTime? nextRunTime,
+  ) async {
+    var scheduling = futureCallEntry.scheduling;
+    if (scheduling == null || nextRunTime == null) return;
+
+    if (scheduling is IntervalFutureCallScheduling) {
+      scheduling = scheduling.copyWith(start: null);
+    }
```
marcelomendoncasoares: Question: Why remove the `start` value? It still serves as the reference point for the interval (for example, the exact second used to run a minute-based interval). After changing the next schedule to use `start ?? futureCallEntry.time`, it won’t be strictly necessary, but it still carries information about when the recurrence began.

packages/serverpod/lib/src/server/future_call_dispatch.dart
```diff
@@ -12,7 +12,23 @@ abstract class FutureCallDispatch<T> {
   /// Calls a [FutureCall] after the specified [delay], optionally passing a [String] identifier.
   T callWithDelay(Duration delay, {String? identifier});
 
+  /// Calls a [FutureCall] at a recurring interval, optionally passing a [String] identifier.
+  RecurringFutureCallDispatch callRecurring({String? identifier});
```
marcelomendoncasoares: Question: Doesn't this need to be generic as well?

```suggestion
  /// Calls a [FutureCall] at a recurring interval, optionally passing a [String] identifier.
  RecurringFutureCallDispatch<T> callRecurring({String? identifier});
```

packages/serverpod/lib/src/server/future_call_dispatch.dart
```diff
@@ -12,7 +12,23 @@ abstract class FutureCallDispatch<T> {
...
+/// Typically, this class is overriden by a generated class.
+abstract class RecurringFutureCallDispatch<T> {
+  /// Calls a [FutureCall] at a recurring interval defined by [cronExpression].
+  T cron(String cronExpression);
+
+  /// Calls a [FutureCall] at a recurring interval defined by [interval],
+  /// optionally passing a [start] time.
+  /// If [start] is in the past, the [FutureCall] is first called immediately
+  /// and then subsequently called recurrently every [interval].
+  T every(Duration interval, {DateTime? start});
```
marcelomendoncasoares, line 39: Suggestion: Enrich the doc comment to make clear that the `start` can be used to define the exact moment in which the job will run on every interval - with concise examples.

packages/serverpod/test/server/cron_test.dart
```diff
@@ -0,0 +1,120 @@
...
+
+  group('Given a valid cron expression', () {
+    test(
+      'when parsing a 5-field expression, '
+      'then all fields are parsed correctly',
+      () {
+        final schedule = Cron.parse('1 13 */2 3-6 *');
```
marcelomendoncasoares: Suggestion: The type of the cron is actually part of the Given (test scenario), not the when (acting). We can remove the group and transform all tests in individual tests.

- Given a valid 5-field cron expression, when parsing, then all fields are parsed correctly.
- Given a valid 5-field cron expression with extra space, when parsing, then all fields are parsed correctly.
- Given a valid 6-field cron expression, when parsing, then all fields are parsed correctly.

These tests also require a lot more cases for completeness. The [original tests of the cron package](https://github.com/agilord/cron/blob/master/test/parser_test.dart) have a few more cases, but the parser has a lot of branches that are not covered here. And we also need to test what happens on the wrong expressions, since it is part of the UX for our users.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
```diff
@@ -372,9 +377,58 @@ class FutureCallManager {
         _internalSession,
         where: (t) => t.id.equals(futureCallEntry.id),
       );
+
+      // Handle recurring future calls
+      await _scheduleNextRecurringCall(futureCallEntry, nextRunTime);
```
marcelomendoncasoares: Question: Haven't we decided to schedule the next recurring future call before invoking? So that we mirror the behavior of how they are done today, leaving up to users to handle concurrent executions in case of overlap - to be improved as a future follow up.

packages/serverpod/lib/src/server/future_call_manager/cron.dart
```diff
@@ -0,0 +1,246 @@
...
+    )?.where((x) => x >= 0 && x <= 59).toList();
+    final parsedHours = parseField(
+      hours,
+    )?.where((x) => x >= 0 && x <= 23).toList();
+    final parsedDays = parseField(
+      days,
+    )?.where((x) => x >= 1 && x <= 31).toList();
+    final parsedMonths = parseField(
+      months,
+    )?.where((x) => x >= 1 && x <= 12).toList();
```
marcelomendoncasoares: Suggestion: Abstract this repeated pattern into a helper function.

packages/serverpod/test/server/cron_test.dart
```diff
@@ -0,0 +1,120 @@
+import 'package:serverpod/src/server/future_call_manager/cron.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test(
+    'when parsing a fixed interval, '
+    'then the field is parsed correctly',
+    () {
+      expect(parseField('12,25'), [12, 25]);
```
marcelomendoncasoares: Suggestion: All tests on this file should test the `Cron.parse` function and not the `parseField`, since it is an implementation detail.

packages/serverpod/lib/src/server/future_call_manager/cron.dart
```diff
@@ -0,0 +1,246 @@
...
+        continue;
+      }
+      return currentDate;
+    }
+  }
+}
+
+@visibleForTesting
+/// Parses a cron field and returns a list of allowed values.
+List<int>? parseField(dynamic constraint) {
```
marcelomendoncasoares: Suggestion: As [mentioned on the test file](https://github.com/serverpod/serverpod/pull/4889/changes#r3054900585), we should make this function private and only test the `Cron.parse` factory constructor.

packages/serverpod/lib/src/server/future_call_manager/cron.dart
```diff
@@ -0,0 +1,246 @@
...
+  /// Parses the [cronExpression].
+  factory Cron.parse(String cronExpression) {
+    List<String?> fields = cronExpression
+        .split(RegExp(r'\s+'))
+        .where((p) => p.isNotEmpty)
+        .toList();
+
+    if (fields.length < 5 || fields.length > 6) {
+      throw CronFormatException('Invalid cron expression: $cronExpression');
+    }
```
marcelomendoncasoares, line 82: Suggestion: Mention the length as cause for the expression being invalid.

tests/serverpod_test_server/test_integration/future_calls/generated_future_calls_test.dart
```diff
@@ -157,6 +158,190 @@ void main() {
...
+          expect(futureCallEntries, isEmpty);
+        },
+      );
+
+      test(
+        'when scheduling a recurring future call with interval, '
+        'then a FutureCallEntry is added to the database with IntervalFutureCallScheduling',
+        () async {
+          final interval = Duration(minutes: 5);
+          final start = DateTime.now().toUtc();
```
marcelomendoncasoares: Suggestion: Invert with the next test and make the `no start` the first test of interval. The idea is to always test the simpler cases first and grow after. Then you don't even need to specify `no start` on the "when", since not passing the parameter is the default.

tests/serverpod_test_server/test_integration/future_calls/generated_future_calls_test.dart
```diff
@@ -157,6 +158,190 @@ void main() {
...
+            session,
+            where: (entry) => entry.name.equals(testCallName),
+          );
+
+          expect(futureCallEntries, hasLength(1));
+          expect(
+            futureCallEntries.first.time.difference(DateTime.now().toUtc()),
+            greaterThan(Duration(minutes: 4, seconds: 55)),
+          );
+        },
```
marcelomendoncasoares: Suggestion: Use the `clock` to pin the current time and match the `interval` exactly.

marcelomendoncasoares (changes requested):
Implementation looks very good, with only a few minor improvements to make it 100% accurate with the spec. 

On the `Cron` implementation, given that it is now our code, we can improve a bit - and specially on the tests coverage, to guarantee the UX we'll be pushing to our users.

packages/serverpod/lib/src/server/future_call_manager/cron.dart
Crazelu: We don't need it actually

packages/serverpod/lib/src/server/future_call_dispatch.dart
Crazelu: It is inferred but better to be explicit.

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: You're right. I mixed that up

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Removing it won't be necessary anymore with your suggestion. Originally, I intended to use it for the first next time computation and then use the current time for subsequent ones. But as you pointed out, there will be time drift issues waiting to happen.

tests/serverpod_test_server/test_integration/future_calls/generated_future_calls_test.dart
Crazelu: Done in [9aeea56](https://github.com/serverpod/serverpod/pull/4889/commits/9aeea568d12390d8e2aecbe3826d36f33f585a0f)

packages/serverpod/test/server/cron_test.dart
Crazelu: Done in [404894b](https://github.com/serverpod/serverpod/pull/4889/commits/404894b0d38cbd98f41116a76049cc527f90925e)

packages/serverpod/lib/src/server/future_call_manager/cron.dart
Crazelu: Done in [57e8c15](https://github.com/serverpod/serverpod/pull/4889/commits/57e8c1598863e52e4727e0eba04a1343bd4e26ef)

packages/serverpod/lib/src/server/future_call_manager/cron.dart
Crazelu: Done in [57e8c15](https://github.com/serverpod/serverpod/pull/4889/commits/57e8c1598863e52e4727e0eba04a1343bd4e26ef)

packages/serverpod/lib/src/server/future_call_manager/cron.dart
Crazelu: Done in [57e8c15](https://github.com/serverpod/serverpod/pull/4889/commits/57e8c1598863e52e4727e0eba04a1343bd4e26ef)

packages/serverpod/lib/src/server/future_call_manager/cron.dart
Crazelu: Done in [57e8c15](https://github.com/serverpod/serverpod/pull/4889/commits/57e8c1598863e52e4727e0eba04a1343bd4e26ef)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Done in [9aeea56](https://github.com/serverpod/serverpod/pull/4889/commits/9aeea568d12390d8e2aecbe3826d36f33f585a0f)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Done in [9aeea56](https://github.com/serverpod/serverpod/pull/4889/commits/9aeea568d12390d8e2aecbe3826d36f33f585a0f)

packages/serverpod/lib/src/server/future_call_manager/future_call_manager.dart
Crazelu: Done in [9aeea56](https://github.com/serverpod/serverpod/pull/4889/commits/9aeea568d12390d8e2aecbe3826d36f33f585a0f)

packages/serverpod/test/server/cron_test.dart
```diff
@@ -0,0 +1,323 @@
+import 'package:serverpod/src/server/future_call_manager/cron.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test(
+    'Given a cron expression with less than 5 fields, '
+    'then a CronFormatException is thrown',
```
marcelomendoncasoares: Suggestion: Missing the "when".

```suggestion
  test(
    'Given a cron expression with less than 5 fields, '
    'when parsing, '
    'then a CronFormatException is thrown',
```

packages/serverpod/test/server/cron_test.dart
```diff
@@ -0,0 +1,323 @@
...
+      expect(
+        () => Cron.parse('* * * *'),
+        throwsA(
+          isA<CronFormatException>().having(
+            (e) => e.message,
+            'message',
+            'Invalid cron expression: * * * *. Only 5-field and 6-field formats are supported.',
+          ),
+        ),
+      );
```
marcelomendoncasoares: Suggestion: Leave only one expect, since they are testing the same.

```suggestion
      expect(
        () => Cron.parse('* * * *'),
        throwsA(
          isA<CronFormatException>().having(
            (e) => e.message,
            'message',
            'Invalid cron expression: * * * *. Only 5-field and 6-field formats are supported.',
          ),
        ),
      );
```

packages/serverpod/test/server/cron_test.dart
```diff
@@ -0,0 +1,323 @@
...
+      expect(
+        () => Cron.parse('* * * * * * * *'),
+        throwsA(
+          isA<CronFormatException>().having(
+            (e) => e.message,
+            'message',
+            'Invalid cron expression: * * * * * * * *. Only 5-field and 6-field formats are supported.',
+          ),
+        ),
+      );
```
marcelomendoncasoares: Suggestion: Same as above.

```suggestion
      expect(
        () => Cron.parse('* * * * * * *'),
        throwsA(
          isA<CronFormatException>().having(
            (e) => e.message,
            'message',
            'Invalid cron expression: * * * * * * *. Only 5-field and 6-field formats are supported.',
          ),
        ),
      );
```

packages/serverpod/test/server/cron_test.dart
```diff
@@ -0,0 +1,323 @@
...
+    'when parsing, '
+    'then a CronFormatException is thrown',
+    () {
+      expect(
+        () => Cron.parse('*/* * * * *'),
+        throwsA(
+          isA<CronFormatException>().having(
+            (e) => e.message,
+            'message',
+            'Invalid cron expression. Invalid interval value: *',
```
marcelomendoncasoares: Question: Is it easy to say in which position the parse error happened? The message as is does not seem much helpful.

packages/serverpod/test/server/cron_test.dart
```diff
@@ -0,0 +1,323 @@
+import 'package:serverpod/src/server/future_call_manager/cron.dart';
+import 'package:test/test.dart';
+
+void main() {
```
marcelomendoncasoares, line 4: Suggestion: General suggestion for this file: put the success cases first and the errors cases last. The philosophy for all tests is to first start by the happy path and grow in complexity then error cases gradually.

tools/serverpod_cli/lib/src/generator/dart/library_generators/future_calls_library_generator.dart
```diff
@@ -247,6 +273,161 @@ extension FutureCallsLibraryGenerator on LibraryGenerator {
...
+                  return '''
+                    final now = $clock.now().toUtc();
+                    DateTime effectiveStart = now.add(interval);
+
+                    if (start != null) {
+                      if (start.isBefore(now))
+                        effectiveStart = now;
+                      else
+                        effectiveStart = start.toUtc();
+                    }
```
marcelomendoncasoares: Suggestion: There is no need to do this, since an entry created on the past is automatically due and will be picked up - there is no expiration of future calls. We can use just the `start` directly.

marcelomendoncasoares (changes requested):
Just a few more small comments and we're ready to go!

packages/serverpod/test/server/cron_test.dart
Crazelu: Yup
CronFormatException extends FormatException which already has a nice implementation for that.

tools/serverpod_cli/lib/src/generator/dart/library_generators/future_calls_library_generator.dart
Crazelu: Done in [c9310d0](https://github.com/serverpod/serverpod/pull/4889/commits/c9310d01e73fc8cff22fd1ff7e923fbc6741282e)

packages/serverpod/test/server/cron_test.dart
Crazelu: Done in [ddf7a1f](https://github.com/serverpod/serverpod/pull/4889/commits/ddf7a1f91ee55ccea6fc2cd9be3a11fb67a8f389)

packages/serverpod/test/server/cron_test.dart
Crazelu: Done in [20e22c4](https://github.com/serverpod/serverpod/pull/4889/commits/20e22c4181d384786694a99ae648cfd8e440c244)

packages/serverpod/test/server/cron_test.dart
Crazelu: Done in [20e22c4](https://github.com/serverpod/serverpod/pull/4889/commits/20e22c4181d384786694a99ae648cfd8e440c244)

packages/serverpod/test/server/cron_test.dart
Crazelu: Done in [42cb440](https://github.com/serverpod/serverpod/pull/4889/commits/42cb440d1026181995a64ee8f12bfc7bdab4ef41)

packages/serverpod/test/server/cron_test.dart
Crazelu: Done in [0c1476b](https://github.com/serverpod/serverpod/pull/4889/commits/0c1476b4fefe822e2bd93535a40e4889aff48382)

packages/serverpod/lib/src/server/future_call_manager/cron.dart
```diff
@@ -86,12 +93,47 @@ class Cron {
...
+      months,
+      offset,
+    )?.where((x) => x >= 1 && x <= 12).toList();
+
+    offset += (months?.length ?? 0) + 1;
+
+    final parsedWeekdays = _parseWeekdayField(
+      weekdays,
+      offset,
+    );
```
marcelomendoncasoares: Suggestion: Restore the cleaner previous pattern and also return the offset from the `_parseField` helper to avoid the duplicated code.

tests/serverpod_test_server/lib/src/generated/future_calls.dart
```diff
@@ -170,15 +170,7 @@ class _RecurringFutureCallDispatchImpl
...
-    DateTime effectiveStart = now.add(interval);
-
-    if (start != null) {
-      if (start.isBefore(now))
-        effectiveStart = now;
-      else
-        effectiveStart = start.toUtc();
-    }
-
+    DateTime effectiveStart = start ?? now.add(interval);
```
marcelomendoncasoares: Suggestion(NIT): Use `final` instead of the type for consistency. Might as well declare the variable inline on the `scheduleFutureCall`.

tools/serverpod_cli/test/generator/dart/server_code_generator/future_calls_test.dart
```diff
@@ -293,14 +293,7 @@ void main() {
...
-                          r'    DateTime effectiveStart = now.add\(interval\);\n'
-                          r'\n'
-                          r'    if \(start != null\) \{\n'
-                          r'      if \(start.isBefore\(now\)\)\n'
-                          r'        effectiveStart = now;\n'
-                          r'      else\n'
-                          r'        effectiveStart = start.toUtc\(\);\n'
-                          r'    \}\n',
+                          r'    DateTime effectiveStart = start \?\? now.add\(interval\);\n',
                         ),
```
marcelomendoncasoares: Suggestion: Also add an expect that the start is used on the `scheduleFutureCall`.

marcelomendoncasoares (approved):
LGTM! 🚀 

Have only three small suggestions for consideration, but feel free to merge when you're ready.

PR #4897, opened 2026-03-26
feat: Add first-class Decimal support for NUMERIC/DECIMAL columns

### LINKED ISSUE #1046: support for NUMERIC and DECIMAL column types on postgresql databases
Hi, by now i guess the only floating type that serverpod support when dealing with database column types is 'double precision'
also i guess there are some scenarios that NUMERIC and DECIMAL needs to be supported.
At the moment if you try to parse a NUMERIC field with a double type, you will get the exception: 
type 'String' is not a subtype of type 'num?'
i guess the engine treats NUMERIC database fields as STRING data type

mohn93:
Adds first-class support for PostgreSQL's `NUMERIC`/`DECIMAL` column type using a custom `Decimal` Dart type in `serverpod_serialization`, with no external dependencies.

## Changes

**Core Type (`serverpod_serialization`):**
- New immutable `Decimal` class wrapping `String` for arbitrary precision
- `@immutable`, implements `Comparable<Decimal>`
- Factory constructors: `parse`, `tryParse`, `fromInt`, `fromDouble`
- Input normalization (leading zeros, trailing zeros, `-0`, `+` sign)
- String-based `compareTo` for arbitrary precision correctness
- `DecimalJsonExtension` for JSON serialization (String-based transport)

**ORM (`serverpod_database`):**
- `ColumnDecimal` extending `ColumnComparable` — supports `<`, `>`, `<=`, `>=`, `between` (unlike `ColumnBigInt` which is equality-only)
- New `ColumnType.numeric` enum entry
- PostgreSQL native `numeric` type mapping (not `text` like `BigInt`)
- Optional precision/scale: `Decimal(10,2)` syntax following `Vector(dim)` pattern
- `Decimal` without parentheses is also valid (unbounded precision)
- Value encoding and filter constraint support

**Code Generation (`serverpod_cli`):**
- Full type system registration: `ValueType.decimal`, `databaseType`, `columnType`, `databaseTypeEnum`
- Model YAML parsing for `Decimal`, `Decimal(10,2)`, `Decimal(19)`, nullable variants
- Default value support: `default=`, `defaultModel=`, `defaultPersist=`
- Default value validation (`_decimalValueValidation`)
- Precision/scale validation (precision > 0, scale >= 0, scale <= precision)
- Code generation outputs `ColumnDecimal` with optional `precision`/`scale` parameters

**Migrations:**
- `numeric` and `numeric(p,s)` SQL generation in Postgres dialect
- Precision/scale change detection triggers column DROP + ADD migration
- Database introspection reads `numeric_precision`/`numeric_scale` from `information_schema`
- `ColumnDefinition` stores `decimalPrecision` and `decimalScale`

**Test Coverage (163 tests):**
- Decimal class: 22 tests (parsing, normalization, equality, compareTo, edge cases)
- Serialization: 9 tests (JSON encode/decode, round-trip, className)
- ORM Column: 32 tests (all comparison operators, NULL, inSet, between)
- CLI type parsing: 7 tests (all Decimal variants)
- CLI code generation: 10 tests (ColumnDecimal with/without precision)
- Field validation: 13 tests (valid + invalid precision/scale)
- SQL fragments: 9 tests (numeric, numeric(p,s), defaults, nullable)
- Migration changes: 21 tests (precision up/down, scale change, unbounded ↔ bounded)
- Column comparison: 5 tests (migration compatibility)
- Model defaults: 11 tests (default/defaultModel/defaultPersist/mix + copyWith)
- Integration - unbounded: 13 tests (real DB CRUD with all filter operations)
- Integration - precision: 10 tests (Decimal(10,2) with rounding, max precision, NULL)
- Integration - defaults: 3 tests (defaultPersist, unsafe query, override)
- E2E migration: 2 tests (create + apply migration lifecycle)

Fixes #1046

## Pre-launch Checklist

## Breaking changes

None — this is purely additive. No existing APIs or behavior are modified.


mohn93:

mohn93:
Hi team — just checking in on this PR. All CI checks pass except for `CLI e2e tests (3.32.0, shard 2)`, which fails on the `start_watch_test.dart` ("serverpod start --watch --no-fes") due to a race condition in `ServerProcess.connectToVmService` (`server_process.dart:192`). This same test also fails on `main` (e.g. [run #23629323972](https://github.com/serverpod/serverpod/actions/runs/23629323972)), and passes locally — so it appears to be a pre-existing flaky test in CI.

This PR is ready for review. Happy to address any feedback!

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
```diff
@@ -947,6 +947,12 @@ class PostgresDatabaseConnection
     if (column is ColumnUuid) return 'uuid';
     if (column is ColumnUri) return 'text';
     if (column is ColumnBigInt) return 'text';
+    if (column is ColumnDecimal) {
+      if (column.precision != null) {
+        return 'numeric(${column.precision},${column.scale ?? 0})';
+      }
+      return 'numeric';
+    }
```
marcelomendoncasoares, line 949: Question: Why go with `numeric` instead of `decimal` to keep consistency? Given that they are the same, but the user will define `Decimal`.

packages/serverpod_serialization/lib/src/decimal.dart
```diff
@@ -0,0 +1,142 @@
...
+/// This class wraps a [String] representation internally to preserve
+/// exact precision, avoiding the rounding errors inherent in
+/// floating-point types like [double].
+///
+/// Maps to PostgreSQL's `numeric` / `decimal` column type.
+@immutable
+class Decimal implements Comparable<Decimal> {
+  final String _value;
+
+  const Decimal._(this._value);
```
marcelomendoncasoares: Question: Why go with a custom `Decimal` implementation instead of using the [decimal](https://pub.dev/packages/decimal) package?

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -322,28 +323,18 @@ class ModelParser {
...
       node,
       Keyword.defaultPersistKey,
     );
 
-    RelationDefinition? relation = _parseRelation(
-      fieldName,
-      typeResult,
-      node,
-    );
+    RelationDefinition? relation = _parseRelation(fieldName, typeResult, node);
```
marcelomendoncasoares: Suggestion: Avoid generating diff on unrelated code just to change the style, as it pollutes the review process. If such changes were done by the formatter, we use `trailing_commas: preserve` on the `analysis_option.yaml` to avoid such changes. Please revert all unrelated changes to make the diff cleaner.

marcelomendoncasoares (changes requested):
Hi @mohn93! Sorry for the delay in getting back to this PR. Before I dive into a more detailed review, I have a couple questions and one suggestion to discuss on the design.

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
mohn93: Good question! I went with `numeric` because PostgreSQL internally normalizes `DECIMAL` to `NUMERIC` — they're aliases, but `information_schema.columns.data_type` always reports `numeric`. Since the migration analyzer in `postgres_analyzer.dart` reads column types from the database and matches them against our definitions, using `decimal` in CREATE TABLE would create a mismatch (we'd write `decimal` but read back `numeric`).

That said, if the team prefers `decimal` for user-facing consistency, I can add the normalization in the analyzer to map both to the same type. Let me know which approach you'd prefer!

packages/serverpod_serialization/lib/src/decimal.dart
mohn93: A few reasons for the custom implementation:

1. **No external dependencies policy** — Serverpod's contribution guidelines state not to add new dependencies. The `decimal` pub package would introduce a transitive dependency chain (`rational` → `decimal`).

2. **BigInt precedent** — The existing `BigInt` support in Serverpod follows the same pattern: a thin Dart-native wrapper, not an external package. `Decimal` mirrors that approach for consistency.

3. **Minimal scope** — Our `Decimal` only needs to be a transport/storage type: parse, serialize, compare. It doesn't need arithmetic operations since the heavy lifting happens in PostgreSQL. The `decimal` package brings much more than needed.

That said, if the team prefers using the `decimal` package for richer functionality (arithmetic, formatting), I'm happy to switch — just wanted to align with the existing patterns first.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
mohn93: Good catch, sorry about that. I'll revert all the unrelated formatting changes in the next push.

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
marcelomendoncasoares: Interesting that Postgres treat numeric as the first-class then. But I'll suggest we aim for consistency on the codebase, even at the expense of replacing `numeric` by `decimal` on the analyzer. Make sure to leave a concise comment on why the replacement is necessary for future developers to know.

packages/serverpod_serialization/lib/src/decimal.dart
marcelomendoncasoares: I'd argue that `Decimal` is pretty different than `BigInt` because `BigInt` is a complete type on Dart that has support for all operations. If we use a custom approach that only serve as storage, we would be pushing a lot of boilerplate for users to convert the `Decimal` to something else to perform operations on Dart, and back again to save it. Using the `decimal` package is better.

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
mohn93: Done! Renamed `ColumnType.numeric` to `ColumnType.decimal` throughout the codebase. SQL output now uses `decimal` everywhere (DDL, query type annotations). Added normalization in `fromSqlType()` to map PostgreSQL's reported `'numeric'` back to `ColumnType.decimal`, with a comment explaining why.

packages/serverpod_serialization/lib/src/decimal.dart
mohn93: Done! Replaced the custom `Decimal` class with a re-export of the [`decimal`](https://pub.dev/packages/decimal) package (`^3.2.4`). Users now get full arithmetic support out of the box.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
mohn93: Done — all unrelated formatting changes have been reverted. The diff now only contains Decimal-specific additions.

mohn93:
@marcelomendoncasoares The only failing check is **Serverpod integration tests (server)** — 6 tests in `broken_future_calls_test.dart` and `future_call_manager_test.dart` fail with `Bad state: No element` on `List.first` (line 147), a race condition where log entries haven't flushed before the assertion runs.

These tests were added by #4779 and #4799 on main — our PR doesn't touch any future_call files. The same 6 failures also appear on unrelated PRs:
- `fix/stream-controller-isclosed-guard` ([run](https://github.com/serverpod/serverpod/actions/runs/24173832343))
- `fix/issue-925-s3-ownership-controls` ([run](https://github.com/serverpod/serverpod/actions/runs/24155849052))

All other checks pass.

packages/serverpod_database/lib/src/adapters/postgres/value_encoder.dart
```diff
@@ -34,6 +34,8 @@ class PostgresValueEncoder extends PostgresTextEncoder implements ValueEncoder {
       return "'${input.toString()}'";
     } else if (input is BigInt) {
       return '\'${input.toString()}\'';
+    } else if (input is Decimal) {
+      return input.toString();
     } else if (input is String &&
```
marcelomendoncasoares, line 42: Suggestion: After merging main, we'll also need to cover SQLite's value encoder.

packages/serverpod_database/lib/src/adapters/postgres/postgres_analyzer.dart
```diff
@@ -72,7 +72,9 @@ WHERE schemaname != 'pg_catalog' AND schemaname != 'information_schema';
       '''
 SELECT column_name, column_default, is_nullable,
        CASE WHEN (data_type = 'USER-DEFINED') THEN udt_name ELSE data_type END as data_type,
-       CASE WHEN (udt_name IN ($vectorTypes)) THEN a.atttypmod ELSE NULL END as vector_size
+       CASE WHEN (udt_name IN ($vectorTypes)) THEN a.atttypmod ELSE NULL END as vector_size,
+       CASE WHEN (data_type = 'numeric') THEN numeric_precision ELSE NULL END as decimal_precision,
+       CASE WHEN (data_type = 'numeric') THEN numeric_scale ELSE NULL END as decimal_scale
 FROM information_schema.columns
```
marcelomendoncasoares, line 78: Suggestion: For SQLite, you'll see that we have an internal table that stores additional information (since `Decimal`, just as `Vector`, are not officially supported). This is for the migration system to be able to properly detect changes to the models and generate the correct migration actions. For `Decimal` we'll need to follow the same pattern.

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
```diff
@@ -975,6 +975,12 @@ class PostgresDatabaseConnection
     if (column is ColumnUuid) return 'uuid';
     if (column is ColumnUri) return 'text';
     if (column is ColumnBigInt) return 'text';
+    if (column is ColumnDecimal) {
+      if (column.precision != null) {
+        return 'decimal(${column.precision},${column.scale ?? 0})';
+      }
+      return 'decimal';
+    }
```
marcelomendoncasoares, line 949: Suggestion: For SQLite, we'll need to store the data as text, since it does not support `Decimal` and storing as integer would easily overflow for larger numbers/precisions.

packages/serverpod_serialization/lib/src/decimal.dart
marcelomendoncasoares, line 1: Suggestion: No need for this intermediary file. We can export the `decimal` package on the `serverpod_serialization.dart` directly (as we do with `uuid`).

tests/serverpod_test_server/lib/src/models/object_with_decimal_precision.spy.yaml
```diff
@@ -0,0 +1,7 @@
+class: ObjectWithDecimalPrecision
+table: object_with_decimal_precision
+fields:
+  price: Decimal(10,2)
+  priceNullable: Decimal(10,2)?
+  quantity: Decimal(19,4)
+  unbounded: Decimal
```
marcelomendoncasoares, line 7: Suggestion: Add more fields with with precision and scale set with default value.

tests/serverpod_test_server/migrations/migration_registry.txt
```diff
@@ -22,3 +22,5 @@
 20260219020242500
 20260305002030445
 20260324085914798
+20260326165627860
+20260326195725300
```
marcelomendoncasoares: Suggestion: Once the feature is finished, squash all migrations to ensure that the PR only adds one migration folder. It also helps review that the expected migration is being generated.

tools/serverpod_cli/test/generator/types_test.dart
```diff
@@ -668,4 +668,72 @@ void main() {
...
+  test(
+    'Given a Decimal(10,2) when databaseType is computed then it returns decimal(10,2).',
+    () {
+      var type = parseType('Decimal(10,2)', extraClasses: []);
+      expect(type.databaseType, 'decimal(10,2)');
+    },
+  );
+
+  test(
+    'Given a Decimal(19) when databaseType is computed then it returns decimal(19,0).',
```
marcelomendoncasoares: Suggestion: Describe the scenario under test rather than the object that represents such scenario. For example:
- "Given a field type of Decimal(10,2)" -> "Given a decimal field with both scale and precision set".

tools/serverpod_cli/test/generator/types_test.dart
```diff
@@ -668,4 +668,72 @@ void main() {
...
+    () {
+      var type = parseType('Decimal(19)', extraClasses: []);
+      expect(type.databaseType, 'decimal(19,0)');
+    },
+  );
```
marcelomendoncasoares: Suggestion: This test scenario (Given) is the same as the third test. Prefer grouping tests that share the same scenario but have only a different acting (when).

tests/serverpod_test_server/test/defaults/decimal/decimal_default_persist_test.dart
```diff
@@ -0,0 +1,14 @@
...
+void main() {
+  group('Given a class with "defaultPersist" Decimal fields,', () {
+    test(
+      'when an object is created, then the "decimalDefaultPersist" field should be null.',
+      () {
+        var object = DecimalDefaultPersist();
+        expect(object.decimalDefaultPersist, isNull);
+      },
+    );
+  });
```
marcelomendoncasoares, line 24: Suggestion: Missing one test case of "'when an object of the class is created with a specific value for..."

tests/serverpod_test_server/test/defaults/decimal/decimal_default_model_test.dart
```diff
@@ -0,0 +1,29 @@
...
+      'when an object is created, then the nullable "decimalDefaultModelStrNull" field should be Decimal(20.5).',
+      () {
+        var object = DecimalDefaultModel();
+        expect(
+          object.decimalDefaultModelStrNull,
+          equals(Decimal.parse('20.5')),
+        );
+      },
+    );
+  });
```
marcelomendoncasoares, line 54: Suggestion: Missing test cases for the nullable field.

tests/serverpod_test_server/test/defaults/decimal/decimal_default_mix_test.dart
```diff
@@ -0,0 +1,40 @@
...
+void main() {
+  group('Given a class with mixed Decimal default fields,', () {
+    test(
+      'when "default" and "defaultModel" are set, then "defaultModel" takes precedence.',
+      () {
+        var object = DecimalDefaultMix();
+        expect(
+          object.decimalDefaultAndDefaultModel,
+          equals(Decimal.parse('20.5')),
+        );
```
marcelomendoncasoares, line 14: Suggestion: Also missing a few test cases in comparison to the default mix test for other column types.

marcelomendoncasoares (changes requested):
Did a first pass on the implementation and partially on the tests. Implementation seems solid, although we'll need to expand a bit to cover SQLite (now merged on main).

After SQLite support is done (requires duplicating the test server models and test files to the sqlite test server project), I'll dive in the tests review.

mohn93:
Hi @marcelomendoncasoares — just pushed a round of commits addressing all the Round-2 feedback. Summary of what changed and where:

**SQLite support** (full mirror of the Postgres Decimal implementation, following the Vector metadata pattern):
- `baf3aafa3` — CLI dialect: `ColumnType.decimal → TEXT`; `serverpod_sqlite_schema` gains `column_decimal_precision` and `column_decimal_scale`.
- `6210c1a3b` — `SqliteValueEncoder` encodes `Decimal` as a quoted TEXT literal (preserves precision vs integer/REAL).
- `376514e58` — `SqliteDatabaseAnalyzer` reads precision/scale from the metadata table and surfaces them on `ColumnDefinition`.
- `c2ea1f985` — Default values for Decimal are quoted (`DEFAULT ('10.5')`) to satisfy STRICT TEXT columns; inverse strips quotes on read-back.

**Code organization:**
- `46b902321` — Dropped `packages/serverpod_serialization/lib/src/decimal.dart`; `decimal` is now re-exported directly from the barrel (`export 'package:decimal/decimal.dart' show Decimal;`), matching the `uuid` pattern.

**Test expansion:**
- `a53b183c7` — Added `priceWithDefault`, `priceWithDefaultNullable`, `quantityWithDefault` fields to `object_with_decimal_precision.spy.yaml`.
- `86f3481ac` — Rewrote `types_test.dart` decimal cases as scenario-centric groups ("Given a decimal field with both precision and scale set," etc.), grouping tests that share the same Given.
- `f8d6bd86f` — Added missing test cases: custom-value override for `decimalDefaultPersist`; nullable/custom-value cases for `decimalDefaultModel`; custom-value cases for every combination in `decimalDefaultMix`.
- `c05532303` — Duplicated the decimal spy.yaml models and test suite into `serverpod_test_sqlite_server`, regenerated the protocol, created the SQLite migration. All 17 decimal tests pass on the SQLite side too.

**Housekeeping:**
- `d0c9d110f` — Squashed three per-step decimal migrations into one clean migration `20260419095621370-decimal`; also fixed the `prefer_const_constructors` lint on `column_decimal_test.dart`.
- `6b5e2a129` — Merge of `origin/main` to pull in SQLite adapter, reconciled `getPgColumnDefault` (moved to `postgres_default_value.dart`) and added the decimal case there.

Full mapping of your review comments → commits:

| Review comment | Commit |
|---|---|
| `value_encoder.dart:39` — cover SQLite value encoder | `6210c1a3b` |
| `postgres_analyzer.dart:78` — follow Vector metadata table pattern | `baf3aafa3` + `376514e58` |
| `database_connection.dart:983` — store as TEXT in SQLite | `baf3aafa3` |
| `decimal.dart:1` — no intermediary, export pub package directly | `46b902321` |
| `object_with_decimal_precision.spy.yaml:7` — add precision/scale defaults | `a53b183c7` |
| `migration_registry.txt:26` — squash into one migration | `d0c9d110f` |
| `types_test.dart:733` — rephrase as scenario | `86f3481ac` |
| `types_test.dart:738` — regroup tests sharing the same Given | `86f3481ac` |
| `decimal_default_persist_test.dart:13` — missing "specific value" case | `f8d6bd86f` |
| `decimal_default_model_test.dart:28` — missing nullable cases | `f8d6bd86f` |
| `decimal_default_mix_test.dart:14` — missing cases vs other types | `f8d6bd86f` |
| Duplicate models/tests to sqlite test server | `c05532303` |

Local verification: `melos`-level analyze clean on `serverpod_database`, `serverpod_test_server`, `serverpod_test_sqlite_server`; `serverpod_cli` shows only pre-existing `deprecated_member_use` infos. 712 CLI DB tests, 780 `serverpod_database` tests, and the decimal test suites on both the Postgres and SQLite test servers (17 each) all pass. Ready for the tests review whenever you are!

PR #4852, opened 2026-03-16
feat: Add support for orderBy and orderByList in delete and deleteWhere

henycave:
This PR adds support for ordering the returned rows when using delete and deleteWhere.

Fixes https://github.com/serverpod/serverpod/issues/3386

## Pre-launch Checklist

- [X] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [X] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
- [X] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).
- [X] I listed at least one issue that this PR fixes in the description above.
- [X] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.
- [X] I added new tests to check the change I am making.
- [X] All existing and new tests are passing.
- [ ] Any breaking changes are documented below.

## Summary by CodeRabbit

* **New Features**
  * Delete operations now support optional ordering parameters to control how deleted rows are returned from the database, including both direct deletion and batch operations.

Issue #633, opened 2023-02-09
Custom fields in config files ?


dufguix:
Hello,

Is it possible to use development.yaml, production.yaml, staging.yaml to add custom config ?
I see it's possible for passwords.yaml.

I think about smtp config for now.

Have a good day.

vlidholt:
That would be a nice addition! I guess it's technically possible to use the passwords file for this until this feature is implemented.

dufguix:
Yes for sure. I'll create a passwords.yaml template without passwords to be able to git push.

vlidholt:
An option is also to do the configuration in code, rather than having it in a config file. Or just load a different config file depending on you run mode. You can access the run mode from `Serverpod.instance!.runmode`. It can be set to one of: `ServerpodRunMode.development`, `ServerpodRunMode.staging`, or `ServerpodRunMode.production`.

dufguix:
Thanks. Good to know `Serverpod.instance!.runmode`.

egyleader:
I can see this is already implemented in serverpod 1.1.0 , I think this issue should be closed 
<img width="282" alt="image" src="https://user-images.githubusercontent.com/31688162/235862311-8df01de7-dfdb-4163-bb77-9b9407ec5973.png">

migalv:
It seems it's not yet implemented if I'm not mistaken. I think this would be very cool to have.

SandPod:
I will update the title of this issue to cover some additional requirements.

## Description
To support advanced use cases and plugins, it should be possible for developers to extend the `<runMode>.yaml` configuration file with their own custom configuration options. These custom configurations should integrate seamlessly with Serverpod’s existing configuration system.

## Requirements

- Developers should be able to register additional configuration fields under custom namespaces.
- It must be possible to specify validation rules and metadata for each configuration field:
  - Whether the field is required or optional.
  - Default values for optional fields.
- Serverpod should produce a **typed configuration object** for the custom configuration:
  - This type can either be supplied by the user at registration or generated by Serverpod.
- The resolution order for custom configuration values must match the standard Serverpod configuration system:
  - Values from environment variables should override values from the configuration file (`<runMode>.yaml`).
- Clear error messages when required fields are missing or have invalid types.

PR #4912, opened 2026-04-01
feat: Implement SQLite database support

### LINKED ISSUE #160: sqlite3 & drift database support
would be nice if have [sqlite3](https://pub.dev/packages/sqlite3) based databases support instead of relying on postgresql and docker  for small projects ,, this will make it possible to integrate orm solutions like [drift](https://pub.dev/packages/drift) for easy creation and minpulation of the database .

### LINKED ISSUE #4785: Implement SQLite database support

marcelomendoncasoares:
Closes: #4785
Closes: #160

Implementation followed closely the spec on the issue (worth reading before reviewing this PR).

The only deviation was to ignore unsupported indexes instead of forbidding. This was preferred because it allows the same project definition to be used for different dialects - which is specially important for modules, that will be used in different projects with different dialects.

## About batch operations

The original idea was to completely reuse all SQL query builders from Postgres, since SQLite is mostly compatible on the syntax. However, for `insert` and `update` operations, Postgres rely on the `DEFAULT` keyword that can be used inside the `VALUES` clause, and this is not supported by SQLite. On SQLite, we need to use the `DEFAULT` keyword outside the `VALUES` clause and omit the columns that have a default value for it to take effect.

Because of this, batch inserts on SQLite were simplified to a transaction/savepoint wrapper around multiple single row inserts. An improvement could be to group rows by the columns that are present, but this would introduce a lot of complexity and mess with the order of the rows. If a better solution is found in the future, this should be improved.

## Migrations

As specified, SQLite has way less column types than Postgres, which requires us to store the target column type during migrations for a proper analysis when inspecting the database schema. This is specially important for repair migrations. **If the `ColumnDefinition` class receive new fields, such fields must also be added to the stored schema table.**

## Testing

This PR introduces the `serverpod_test_sqlite` package, which contains the integration and migrations e2e tests for the SQLite dialect. These tests will run in the CI in parallel with the other tests.

Tests were copied from the `serverpod_test_server` package "as is" on commit https://github.com/serverpod/serverpod/pull/4912/commits/91006c7e3209f648f570abccf41c42329b187570 and adapted to the SQLite dialect later on commit https://github.com/serverpod/serverpod/pull/4912/commits/32cb49f6091bb920f7063d0ed989360c44d7b8e5 so we can track exactly what is supported and how the behavior differs from the Postgres dialect.

## Pre-launch Checklist

## Breaking changes

None.


packages/serverpod_database/lib/src/interface/definition_restrictions.dart
```diff
@@ -0,0 +1,72 @@
...
+    final unsupportedIndexes = map(
+      (t) => t.indexes
+          .where((i) => !supportedIndexTypes.contains(i.type))
+          .toList(),
+    ).expand((t) => t).toList();
+
+    if (unsupportedIndexes.isEmpty) return this;
+
+    logWarnings?.call(
+      'The following indexes will be skipped since its type is not supported '
```
Crazelu: NIT: Would this not better read as:

`The following indexes will be skipped due to unsupported types `

packages/serverpod_database/test/sqlite_migration_runner_test.dart
```diff
@@ -0,0 +1,147 @@
...
+              'Foreign key integrity check failed: 1 violation.\n'
+              '  1. table=child, rowid=1, parent=parent, fkid=0',
+            ),
+          ),
+        ),
+      );
+    },
+  );
+
+  test(
```
Crazelu, line 76: Question: This test and the one below validate that migrations complete successfully in an implicit fashion (no errors thrown). Wouldn't it be better to validate that the migration action completed successfully with a find?

packages/serverpod_database/test/sqlite_migration_runner_test.dart
```diff
@@ -0,0 +1,147 @@
...
+    }
+  });
+
+  test(
+    'Given schema migrations that violate foreign key constraints in a development environment, '
+    'when running migrations, '
+    'then it fails and reports each foreign key constraint violation.',
+    () async {
+      const runner = SqliteDatabaseMigrationRunner(runMode: 'development');
+      await expectLater(
```
Crazelu, line 46: Suggestion: Add a separate expectation to validate that the insertion is not committed.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -610,6 +647,49 @@ class PostgresDatabaseConfig extends DatabaseConfig {
...
+  factory SqliteDatabaseConfig._fromJson(
+    Map dbSetup,
+    Map passwords,
+    String name,
+  ) {
+    final path = dbSetup[ServerpodEnv.databaseFilePath.configKey];
+    if (path == null || path is! String || path.trim().isEmpty) {
+      throw ArgumentError('Invalid SQLite database configuration: $dbSetup');
+    }
+    return SqliteDatabaseConfig(filePath: path);
```
Crazelu: Question: The max connection count config is not extracted from `dbSetup` for `_fromJson`. Why is the default always used for sqlite?

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -516,14 +527,40 @@ abstract class DatabaseConfig {
...
+    var dialect = dbSetup[ServerpodEnv.databaseDialect.configKey];
+    if (dialect != null && dialect is String && dialect.trim().isNotEmpty) {
+      final parser = _getParser(DatabaseDialect.values.byName(dialect));
+      return parser(dbSetup, passwords, name);
+    }
+
+    Object? error;
+    for (var parser in DatabaseDialect.values.map(_getParser)) {
+      try {
+        return parser(dbSetup, passwords, name);
```
Crazelu, line 558: Suggestion: We can improve UX slightly with an info log to communicate a fallback selection when dialect is not specified. It could be quick feedback to nudge users about a missing dialect.

Crazelu (commented):
I've made a first pass and have a few questions and suggestions.
I'll go through the tests next.

packages/serverpod_shared/lib/src/config.dart
marcelomendoncasoares: Good catch! This was indeed missing here. Fixed at https://github.com/serverpod/serverpod/pull/4912/commits/dda5c942f0ac94326142ba01dadd75fbc74af0d3.

We can't use it yet - [as commented on the `SqlitePoolManager` implementation](https://github.com/marcelomendoncasoares/serverpod/blob/fb6b9152bb2deb687a7246a555ba708020e5fc7b/packages/serverpod_database/lib/src/adapters/sqlite/sqlite_pool_manager.dart#L54-L64) -, but it should be correctly parsed from the config for when we update to `sqlite_async: ^0.14` (currently on beta and will require Dart 3.10).

packages/serverpod_database/lib/src/interface/definition_restrictions.dart
marcelomendoncasoares: Good suggestion. Done at https://github.com/serverpod/serverpod/pull/4912/commits/feb555fe06e12c1859a7aabc85cfd067d5191dc1.

packages/serverpod_database/test/sqlite_migration_runner_test.dart
marcelomendoncasoares: Added the expect on the tables not being created at https://github.com/serverpod/serverpod/pull/4912/commits/fb6b9152bb2deb687a7246a555ba708020e5fc7b.

packages/serverpod_database/test/sqlite_migration_runner_test.dart
marcelomendoncasoares: Better indeed with the explicit validation. Done at https://github.com/serverpod/serverpod/pull/4912/commits/fb6b9152bb2deb687a7246a555ba708020e5fc7b.

packages/serverpod_shared/lib/src/config.dart
marcelomendoncasoares: This is actually intentional, since we don't want users to be manually specifying the dialect. Specifying the dialect directly is only expected once we have dialects that share same setup (as a future `mariadb` sharing same configs as `postgres`). Given that we don't even have this scenario (`sqlite` has a different config set), a warning message would always be printed and just pollute the context for users.

packages/serverpod_database/lib/src/adapters/sqlite/database_provider.dart
```diff
@@ -0,0 +1,55 @@
...
+  @override
+  SqlitePoolManager createPoolManager(
+    SerializationManagerServer serializationManager,
+    RuntimeParametersListBuilder? runtimeParametersBuilder,
+    DatabaseConfig config,
+  ) {
+    if (runtimeParametersBuilder != null) {
+      throw UnsupportedError('SQLite does not support runtime parameters.');
+    }
+    if (config is! SqliteDatabaseConfig) {
```
Crazelu: Suggestion: We could get type safety guarantees with generics on `DatabaseProvider`.
```dart
abstract interface class DatabaseProvider<A extends DatabaseConfig, B extends DatabasePoolManager>
```

That would remove the need for a check like this.

packages/serverpod_database/lib/src/adapters/sqlite/database_provider.dart
```diff
@@ -0,0 +1,55 @@
...
+    }
+    return SqlitePoolManager(
+      serializationManager,
+      config,
+    );
+  }
+
+  @override
+  SqliteDatabaseConnection createConnection(DatabasePoolManager poolManager) {
+    if (poolManager is! SqlitePoolManager) {
```
Crazelu: Suggestion: We could get type safety guarantees with generics on `DatabaseProvider`.
```dart
abstract interface class DatabaseProvider<A extends DatabaseConfig, B extends DatabasePoolManager>
```

That would remove the need for a check like this.

packages/serverpod_database/lib/src/adapters/sqlite/database_provider.dart
marcelomendoncasoares: Unfortunately we can't, because the `DatabaseProvider` class is meant to be created from the factory `forDialect` that only receives the `DatabaseDialect` and knows nothing about the types of the config and pool manager that a given dialect require.

image

packages/serverpod_database/lib/src/adapters/sqlite/database_provider.dart
marcelomendoncasoares: See https://github.com/serverpod/serverpod/pull/4912#discussion_r3047495963.

packages/serverpod_database/lib/src/adapters/sqlite/database_provider.dart
Crazelu: I see
This looks like a good place to use [covariant parameters](https://dart.dev/language/type-system#covariant-parameters) to tighten the types and still get the runtime check implicitly.

For instance, in the image below, the if block is no longer necessary

Screenshot 2026-04-07 at 22 09 27

If a non-conformant type is passed, this kind of exception is thrown at runtime:

```
Unhandled exception:
type 'SqlitePoolManager' is not a subtype of type 'PostgresPoolManager' of 'poolManager'
```

marcelomendoncasoares: Indeed! I completely forgot about `covariant`. I'm more used to declaring it on the base class, though. So I added it to the parameters on the base `DatabaseProvider` so that each implementation is free to override the type without having to add it again. Done at https://github.com/serverpod/serverpod/pull/4912/commits/52469590b4fe830f1f682e6774f0a3db2fb81afb.

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -516,14 +527,40 @@ abstract class DatabaseConfig {
...
   /// This method will try to parse the database configuration for all available
-  /// database dialects.
+  /// database dialects. If parsing for all dialects fails, throws the first
+  /// error (PostgreSQL by default). If the [dialect] parameter is set, only
+  /// that dialect will be tried.
   factory DatabaseConfig._fromJson(Map dbSetup, Map passwords, String name) {
-    return PostgresDatabaseConfig._fromJson(dbSetup, passwords, name);
+    var dialect = dbSetup[ServerpodEnv.databaseDialect.configKey];
+    if (dialect != null && dialect is String && dialect.trim().isNotEmpty) {
+      final parser = _getParser(DatabaseDialect.values.byName(dialect));
```
nielsenko, line 551: Should we use the trimmed value when calling `byName`?  Or does the json decoding make trimming pointless?

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -610,6 +647,49 @@ class PostgresDatabaseConfig extends DatabaseConfig {
...
+       );
+
+  /// The file path to the SQLite database.
+  String get filePath => host;
+
+  factory SqliteDatabaseConfig._fromJson(
+    Map dbSetup,
+    Map passwords,
+    String name,
+  ) {
```
nielsenko, line 675: This sis a slightly odd signature for a _fromJson method, ie. taking two maps and a name. Should it be named differently?

packages/serverpod_shared/lib/src/config.dart
```diff
@@ -610,6 +647,49 @@ class PostgresDatabaseConfig extends DatabaseConfig {
...
+    final path = dbSetup[ServerpodEnv.databaseFilePath.configKey];
+    if (path == null || path is! String || path.trim().isEmpty) {
+      throw ArgumentError('Invalid SQLite database configuration: $dbSetup');
+    }
```
nielsenko, line 679: Should we normalize and check for existence here? Should we pass the trimmed (or normalized) path to `SqliteDatabaseConfig` ctor

packages/serverpod_shared/test/database_config_test.dart
```diff
@@ -649,4 +649,261 @@ database:
...
+    () {
+      var serverpodConfig = '''
+apiServer:
+  port: 8080
+  publicHost: localhost
+  publicPort: 8080
+  publicScheme: http
+database:
+  dialect: sqlite
+  filePath: /tmp/serverpod.sqlite
```
nielsenko, line 693: Curious.. how is `filePath:     /tmp/serverpod.sqlite` decoded? I would suppose json decoder trims.

tests/serverpod_test_sqlite/serverpod_test_sqlite_server/test_integration/database_operations/crud/update_test.dart
```diff
@@ -578,7 +578,11 @@ void main() async {
 
         var updated = await Types.db.updateRow(session, value);
 
-        expect(updated.aDateTime, equals(now));
+        // SQLite stores timestamps with millisecond precision only.
+        expect(
+          updated.aDateTime?.millisecondsSinceEpoch,
+          equals(now.millisecondsSinceEpoch),
```
nielsenko, line 584: Should we have extension method for convinience?

tests/serverpod_test_sqlite/serverpod_test_sqlite_server/test_integration/database_operations/transactions/find_test.dart
```diff
@@ -96,7 +96,8 @@ void main() async {
     );
 
     test(
-      'when calling `find` without transaction then does not find the object',
+      'when calling `find` without transaction '
+      'then an exception is thrown because SQLite does not allow recursive locks',
```
nielsenko, line 100: Something about this _then_ clause rubs me the wrong way 🤔

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -0,0 +1,1409 @@
...
+  Future<void> lockRows<T extends TableRow>(
+    DatabaseSession session, {
+    required Expression where,
+    required LockMode lockMode,
+    required Transaction transaction,
+    LockBehavior lockBehavior = LockBehavior.wait,
+  }) async {
+    // SQLite has no row-level locking; using a transaction is sufficient.
+    // No-op so callers (e.g. lockRows then find) still work.
+  }
```
nielsenko, line 182: Should we override the inherited doc comment to give inline documentation in IDE of this important caveat? Or make this throw on sqlite and add extension method `lockRowsIfSupported` that just don't call `lockRows` on SQLite?

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -0,0 +1,1409 @@
...
+    return result ?? ResultSet([], null, []);
+  }
+
+  Future<ResultSet> _runSingleStatementQuery(
+    DatabaseSession session,
+    _ParsedSqlStatement parsed, {
+    List<Object?>? parameters,
+    _SqliteTransaction? sqliteTx,
+  }) async {
+    var startTime = DateTime.now();
```
nielsenko, line 870: This is probably fine here, but `DateTime.now()` is heavier than most people think. If all you need is duration, then prefer to use a `StopWatch` instead.  A quick glance at `_logQuery` tells me that is all it is used for.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -0,0 +1,1409 @@
...
+  }
+
+  /// Maps a row from database column names to protocol field names when the
+  /// table has explicit column names. Required so RETURNING * deserialization
+  /// matches what Protocol expects (field names).
+  static Map<String, dynamic> _rowDbColumnNamesToFieldNames(
+    Table table,
+    Map<String, dynamic> row,
+  ) {
+    if (!table.hasColumnMapping) return row;
```
nielsenko, line 917: 👍 good

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -0,0 +1,1409 @@
...
+
+  void _normalizePrefixedColumns(
+    Map<String, dynamic> row,
+    Table table, {
+    Include? include,
+  }) {
+    for (final column in table.columns) {
+      final key = truncateIdentifier(
+        column.fieldQueryAlias,
+        DatabaseConstants.pgsqlMaxNameLimitation,
```
nielsenko, line 961: Is this an oversight?

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -0,0 +1,1409 @@
...
+
+  static void _logQuery(
+    DatabaseSession session,
+    String query,
+    DateTime startTime, {
+    int? numRowsAffected,
+    dynamic exception,
+    StackTrace? trace,
+  }) {
+    var duration = DateTime.now().difference(startTime);
```
nielsenko: Why not pass `Duration`? Then caller can use `StopWatch` and avoid two expensive `DateTime.now()` calls just to measure a duration.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -0,0 +1,1409 @@
...
+      return await _db.writeTransaction<R>((tx) async {
+        var transaction = _SqliteTransaction(tx, session);
+        return await transactionFunction(transaction);
+      });
+    } on TransactionCancelledException catch (_) {
+      Type typeOf<X>() => X;
+
+      if (R == Null || R == typeOf<void>()) {
+        return Future.value();
+      }
```
nielsenko: Not sure I understand why we don't _rethrow_ here? 🤔 

If result type is `Null` or `void` we consider a cancelled transaction as succesfull.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -0,0 +1,1409 @@
...
+      final p = script[i];
+      if (!p.isIgnoredForExecuteRowCount) return p;
+    }
+    return null;
+  }
+
+  static String _stripTrailingSemicolon(String sql) {
+    var t = sql.trim();
+    if (!t.endsWith(';')) return t;
+    return t.substring(0, t.length - 1).trim();
```
nielsenko: What if it ends with multiple ';'?

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -0,0 +1,1409 @@
...
+  }
+}
+
+/// Single SQL statement with optional [ast] for [SqliteDatabaseConnection].
+class _ParsedSqlStatement {
+  _ParsedSqlStatement({required String text, required this.ast})
+    : text = text.trim();
+
+  final String text;
```
nielsenko, line 1298: Isn't this available from `ast.span.text`?

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -0,0 +1,1409 @@
...
+    List<Object?> parameters = const [],
+  ]) {
+    return _ctx.execute(query, parameters);
+  }
+
+  Future<void> _execute(
+    String sql, [
+    List<Object?> parameters = const [],
+  ]) async {
+    try {
```
nielsenko, line 1453: Could add an `assert` here that validated the `sql` with the parser in debug.

packages/serverpod_database/lib/src/adapters/sqlite/sqlite_analyzer.dart
```diff
@@ -0,0 +1,235 @@
...
+            : sqliteSqlToAbstractDefault(e[4] as String?, columnType),
+        columnType: columnType,
+        isNullable: !isIdColumn && (e[3] as int?) != 1,
+        vectorDimension: columnTypes[columnName]?.vectorDimension,
+      );
+    }).toList();
+  }
+
+  @override
+  Future<List<IndexDefinition>> getIndexDefinitions({
```
nielsenko, line 97: General commnet. Sometime in the future we should consider using `Stream<T>` instead of `Future<List<T>>` on interfaces like this.

packages/serverpod_database/lib/src/interface/definition_restrictions.dart
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    return [
+      for (var t in this)
+        t.copyWith(
+          indexes: t.indexes
+              .where((i) => supportedIndexTypes.contains(i.type))
+              .toList(),
+        ),
+    ];
```
nielsenko, line 70: Should we cache this?

packages/serverpod_database/lib/src/adapters/sqlite/sqlite_migration_runner.dart
```diff
@@ -17,11 +17,28 @@ class SqliteDatabaseMigrationRunner implements MigrationRunner {
   ) async {
     await session.db.unsafeExecute('PRAGMA foreign_keys=OFF');
     try {
-      await session.db.transaction(
-        (transaction) => action(transaction),
-      );
+      await session.db.transaction((transaction) async {
+        await action(transaction);
+        if (runMode == 'development') {
```
nielsenko, line 22: Should we make `runMode` an `enum` instead of a `String`?

tests/docker/tests_sqlite_integration/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Dart SDK from the official image.
+FROM dart:3.8.0 AS dart-sdk
+FROM ubuntu:24.04 AS build
+
+# Copy SQLite >= 3.41.0 from Ubuntu 24.04.
```
nielsenko: misleading comment

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -532,11 +532,43 @@ class SqliteDatabaseConnection extends DatabaseConnection<SqlitePoolManager> {
...
-    ).withSelectFields([table.id]).withWhere(where).build();
+    var selectIds = SelectQueryBuilder(table: table)
+        .withSelectFields([table.id])
+        .withWhere(where)
+        .withOrderBy(orderByList)
+        .build();
+
+    // It is not possible to use CTEs with delete on SQLite, so we need to first
+    // select the ids and then delete the rows. For the operation to be atomic,
+    // it runs inside a transaction or savepoint.
```
nielsenko: I don't think this is completely correct. If the `WITH` clause comes first, like:
```sql
WITH to_die AS (
    SELECT id 
    FROM table 
    WHERE predicate
)
DELETE FROM table 
WHERE id IN (SELECT id FROM to_die);
```
I think it would work.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -651,28 +651,49 @@ class SqliteDatabaseConnection extends DatabaseConnection<SqlitePoolManager> {
...
+        var updatedRows = 0;
+        for (final statement in script) {
+          db.execute(statement.text, params);
+          updatedRows += db.updatedRows;
+        }
+        return updatedRows;
+      });
+    }
+
     var result = await _runQuery(
```
nielsenko: We already have a parsed script here.. a bit sad we parse it again in `_runQuery`.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -734,17 +730,32 @@ class SqliteDatabaseConnection extends DatabaseConnection<SqlitePoolManager> {
...
+  }) {
+    return _runQueryWithScript(
+      session,
+      _parseSqlScriptForBatch(query),
+      parameters: parameters,
+      transaction: transaction,
+    );
+  }
 
+  Future<ResultSet> _runQueryWithScript(
```
nielsenko, line 814: 😄 👍 .. disregard my earlier comment

nielsenko (commented):
This was a big one. Looks good, but got some comments/questions.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -0,0 +1,1452 @@
...
+            session,
+            table.id.inSet(orderedIds.castToIdType().toSet()),
+            transaction: tx,
+          );
+
+          return deletedRows.toList()..sort((a, b) {
+            final aId = orderedIds.indexOf(a.id);
+            final bId = orderedIds.indexOf(b.id);
+            return aId.compareTo(bId);
+          });
```
nielsenko: `indexOf` is O(n) per call, and sort calls the comparator O(n log n) times, making this O(n^2 log n).
I suggest building a Map<id, index> first (same pattern already used in `_restoreUpdateWhereSelectionOrder`).

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: My comment was a bit poorly explained here. You are right that regular CTEs can be used to filter like this, but the kind that we need for the recently introduced order by to work would be a [data-modifying CTE (as we do on Postgres), which is not supported on SQLite](https://sqlite.org/lang_returning.html#:~:text=PostgreSQL%20has%20the%20ability%20to%20use%20a%20DML%20statement%20with%20a%20RETURNING%20clause%20like%20a%20view%20in%20a%20common%20table%20expressions.%20SQLite%20does%20not%20currently%20have%20that%20ability%2C%20though%20that%20is%20something%20that%20might%20be%20added%20in%20a%20future%20release.).

```sql
WITH to_die AS (
  SELECT id 
  FROM "table"
  WHERE predicate
), deleted_rows AS (
  DELETE FROM "table"
  WHERE "id" IN (SELECT "id" FROM to_die)
  RETURNING *  --> Data-modifying CTE
)
SELECT *
FROM "deleted_rows"
ORDER BY "id" ASC NULLS LAST;
```

[The reason is because SQLite does not ensure order on `RETURNING` from a delete operation](https://sqlite.org/lang_returning.html#:~:text=The%20rows%20emitted%20by%20the%20RETURNING%20clause%20appear%20in%20an%20arbitrary%20order.). So we need to first query the data in the requested order and then delete it separately.

packages/serverpod_database/lib/src/adapters/sqlite/sqlite_migration_runner.dart
marcelomendoncasoares: It would sure be better on the comparison, but also a breaking change, since this parameter comes from the `Serverpod` class. And it would also prevent the [support for arbitrary user-defined environments](https://github.com/serverpod/serverpod/issues/633#issuecomment-2938983574).

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: You caught me 😆 

I'll trim extra semicolons.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
nielsenko: Is the ordering important? Otherwise we have:
```sql
DELETE FROM "table"
WHERE "id" IN (
    SELECT "id" 
    FROM "table" 
    WHERE predicate
)
RETURNING *; 
```
as of [3.35.0](https://sqlite.org/releaselog/3_35_0.html). Or sort the result in dart?

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: This is intentional. We're reusing the Postgres query builder and query result parser, which truncates identifiers to Postgres max limit, so we need to also truncate it to be compatible.

tests/serverpod_test_sqlite/serverpod_test_sqlite_server/test_integration/database_operations/transactions/find_test.dart
marcelomendoncasoares: Probably because I baked an explanation on the clause to make it clear why it differs from Postgres. I didn't see an issue because it is not violating the clause purpose, only adding extra info. But I can transform the explanation in a comment if you think it will be cleaner.

packages/serverpod_shared/lib/src/config.dart
marcelomendoncasoares: The name here is only following the already existing convention for all other configs. But I agree that it feels odd.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -0,0 +1,1452 @@
...
+        ...await _runInsert(
+          session,
+          rows,
+          ignoreConflicts,
+          withIdNull,
+          transaction,
+        ),
+    ];
+
+    var merged = _mergeResultsWithNonPersistedFields(rows)(allMaps);
```
nielsenko, line 219: `_mergeResultsWithNonPersistedFields` assumes index alignment. If `ignoreConflicts: true` is used, some rows may not produce results (conflict implies no row). The merge then indexes into the original rows list by position, which won't match.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: The order wasn't important until https://github.com/serverpod/serverpod/pull/4852.

I don't like much the idea of sorting in Dart as we'll have to implement a parser of the order by and this alone will have a lot of edge cases.

packages/serverpod_database/lib/src/adapters/sqlite/sqlite_analyzer.dart
```diff
@@ -0,0 +1,235 @@
...
+      'PRAGMA table_info($quotedTable)',
+    );
+
+    final columnTypes = {
+      for (var row in await database.unsafeQuery('''
+        SELECT "column_name",
+               "column_type",
+               "column_vector_dimension"
+        FROM serverpod_sqlite_schema
+        WHERE (table_name = '$tableName');'''))
```
nielsenko, line 69: Should this use `quotedTable`?

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -0,0 +1,1452 @@
...
+    String query,
+    QueryParameters? parameters,
+  ) {
+    if (parameters == null) return (query, const []);
+
+    if (parameters is QueryParametersPositional) {
+      var list = parameters.parameters;
+      var sql = query;
+      for (var i = 0; i < list.length; i++) {
+        sql = sql.replaceFirst(RegExp(r'\$' + (i + 1).toString()), '?');
```
nielsenko: Brittle.. what if sql contains `'Price is $10'`

packages/serverpod_database/lib/src/adapters/sqlite/sqlite_pool_manager.dart
```diff
@@ -0,0 +1,80 @@
...
+    var db = _db;
+    if (db == null) {
+      throw StateError('Database not started.');
+    }
+    return db;
+  }
+
+  /// The encoder used to encode objects for storing in the database.
+  @override
+  SqliteValueEncoder get encoder => SqliteValueEncoder();
```
nielsenko:
```suggestion
  @override
  late final SqliteValueEncoder encoder = SqliteValueEncoder();
```
or make it `const` constructed

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: You're right, but this is solved because all inserts with more than 1 row are converted into a transaction or savepoint that inserts individually. So the merge will only apply for a single row.

```dart
  Future<List<T>> insert<T extends TableRow>(
    DatabaseSession session,
    List<T> rows, {
    Transaction? transaction,
    bool ignoreConflicts = false,
  }) async {
    if (rows.isEmpty) return [];
    if (rows.length > 1) {
      return DatabaseUtil.runInTransactionOrSavepoint(
        session.db,
        transaction,
        (tx) async => [
          for (var row in rows)
            await insert<T>(
              session,
              [row],
              transaction: tx,
              ignoreConflicts: ignoreConflicts,
            ).then((results) => results.firstOrNull),
        ].whereType<T>().toList(),
      );
    }
    ...
```

This has performance impacts, but I could not work around because of one major SQLite limitation: we can't use `DEFAULT` as a placeholder for `VALUES` when the column has default value and the model field is `null`. On Postgres, this is how multiple rows are combined in a single statement, because the columns list is the same.

```sql
-- On Postgres, we do
INSERT INTO "table" ("column1", "column2", "columnWithDefault") VALUES (1, 2, DEFAULT) RETURNING *

-- On SQLite, we need to omit the columns with default values
INSERT INTO "table" ("column1", "column2") VALUES (1, 2) RETURNING *
```

We could try to group rows that set the same columns, or evaluate which rows have default values and which not, but this would increase the complexity a lot, specially to ensure the order. It can be an optimization for the future.

As a side note, the same problem with non-persistent fields would also happen on Postgres, and there we also use the same individual row insert strategy, but only for this particular use case.

packages/serverpod_database/lib/src/adapters/sqlite/sqlite_analyzer.dart
```diff
@@ -0,0 +1,235 @@
...
+      case 'SET DEFAULT':
+        return ForeignKeyAction.setDefault;
+      case 'RESTRICT':
+        return ForeignKeyAction.restrict;
+      case 'NO ACTION':
+        return ForeignKeyAction.noAction;
+      case 'CASCADE':
+        return ForeignKeyAction.cascade;
+      default:
+        return null;
```
nielsenko, line 204: Should this throw? I believe `ForeignKeyAction.noAction` is default for SQLite. So random `String` would behave like that, or?

packages/serverpod_database/lib/src/adapters/sqlite/sqlite_analyzer.dart
marcelomendoncasoares: Can't, because the `quotedTable` is with double quotes (as an identifier) and here we need it as a value (single quotes). I've also done it during the implementation and realized it when tests failed 😆

packages/serverpod_database/lib/src/adapters/sqlite/sqlite_pool_manager.dart
marcelomendoncasoares: Funny... I though I had made it const 🤔

packages/serverpod_database/lib/src/adapters/sqlite/sqlite_analyzer.dart
nielsenko: What if `tableName` contains a single quote?

marcelomendoncasoares: Can it? I didn't worry about this because we do not allow single quotes on table names on the model validation. It will break only if some user manually creates a table with a single quote, since SQLite supports it, but the migration system would break anyway. I can escape it to make 100% safe.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: I like the idea of overriding the comment, even though it might not reach the end-user since the `DatabaseConnection` is internal - the user sees only the generic `Database` class, which wraps the `DatabaseConnection`. Throwing or adding a similar method I don't think is a good UX as we want the code to be compatible with multiple engines. But we can make it safer by issuing a warning like:

```dart
  /// No-op for SQLite.
  ///
  /// Since SQLite allow only one write transaction at a time, locking specific
  /// rows is not necessary - nor supported by the SQLite engine. Given that
  /// [lockRows] require a transaction, the lock will be acquired implicitly.
  ///
  /// This method is no-op to maintain compatibility with code that can also
  /// be run against other databases.
  @override
  Future<void> lockRows<T extends TableRow>(
    DatabaseSession session, {
    required Expression where,
    required LockMode lockMode,
    required Transaction transaction,
    LockBehavior lockBehavior = LockBehavior.wait,
  }) async {
    if (lockBehavior != LockBehavior.wait) {
      await session.logWarning?.call(
        'Calling "lockRows" on SQLite is a no-op since it does not support '
        'concurrent writes and the transaction will acquire a lock implicitly. '
        'Note that the lock behavior "$lockBehavior" will be ignored. Use '
        '"LockBehavior.wait" to suppress this warning.',
      );
    }
  }
```

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: Good point. Reuse the already existing sorting method at https://github.com/serverpod/serverpod/pull/4912/commits/f5075318081881eec403c7eed04e77460284ba77.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: Improved the comment to be clearer at https://github.com/serverpod/serverpod/pull/4912/commits/f5075318081881eec403c7eed04e77460284ba77.

tests/docker/tests_sqlite_integration/Dockerfile
marcelomendoncasoares: This is leftover from previous attempts. I'll fix the comment.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: This was a dirty workaround to the limited SQLite API on how to deal with transactions. Differently from the `postgres` package, SQLite does not expose a `rollback` or `cancel` method to abort a transaction and proceed. So we're left with having to call `await _ctx.execute('ROLLBACK');` on the transaction write context directly. But SQLite will still try to `COMMIT` the transaction when it reaches the end, leading to the exception:

```console
00:36 +441 -7: test_integration/database_operations/transactions/insert_test.dart: Given a transaction that is cancelled when inserting row after transaction is cancelled then insertion has no effect [E]
  SqliteException(1): while selecting from statement, cannot commit - no transaction is active, SQL logic error (code 1)
    Causing statement: COMMIT
  package:sqlite3/src/implementation/exception.dart 95:3       throwException
  package:sqlite3/src/implementation/statement.dart 140:7      StatementImplementation._selectResults
```

This is specially problematic when calling `transaction.cancel()` like:

```dart
  test(
    'Given inserting an object inside a transaction that is cancelled when calling `count`'
    'inside the transaction then should return 1 but outside the transaction should return 0.',
    () async {
      await session.db.transaction(
        (transaction) async {
          await UniqueData.db.insertRow(
            session,
            UniqueData(number: 111, email: 'test@serverpod.dev'),
            transaction: transaction,
          );

          var count = await UniqueData.db.count(
            session,
            where: (_) => Constant.bool(true),
            transaction: transaction,
          );

          expect(count, 1);

          await transaction.cancel(); // <-- Here
        },
      );
```

The way that I found to avoid the driver exception while also rolling changes back was throwing an exception inside the transaction function and capturing outside, before reaching the user. But we could only throw inside cancel if the user was not returning any value, which is the type check you have highlighted. Since we had zero tests or docs for the case of returning a value after cancelling a transaction, I considered this as not being part of the API and settled with the hack for the known path.

BUT, while writing this answer, I came up with a way better solution - and, now, a lot more obvious: move the throw to the outer wrapper of the transaction, so that we can allow any return. This is now done at https://github.com/serverpod/serverpod/pull/4912/commits/51c40a66aad9d14e22d317e11761c6927ee9c567.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: Not for `InvalidStatement`. And, since the `text` is always present, I preferred to leave it explicit.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: Unfortunately no, because it will introduce an asymmetry on the exceptions received on development and production. For example, a query with the vector operator `<->` would throw an assert error in development, and a real exception in production, making hard to develop consistent guardrails for the exceptions.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/4912/commits/c8723c76292030edf1d11e4420aa492220b810c2 by creating a proper query parameter conversion utility (with a robust set of tests).

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/4912/commits/c8723c76292030edf1d11e4420aa492220b810c2.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: We actually need it to be `DateTime` because `poolManager.lastDatabaseOperationTime` is receiving it. We then reuse to calculate the duration, but we would not be able to get rid of the first call to `now`.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: We can't get rid of the first `DateTime.now()`, but I've replaced the duration count to use the `Stopwatch` at https://github.com/serverpod/serverpod/pull/4912/commits/ee31c8464bab1a73ef93e2be8c94891b4c59676a.

packages/serverpod_database/lib/src/interface/definition_restrictions.dart
marcelomendoncasoares: I don't think it's needed, because we only call it once inside the `DatabaseAnalyzer`.

packages/serverpod_database/lib/src/adapters/sqlite/sqlite_pool_manager.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/4912/commits/65bf9c00674f4e7e610be449d62b6766e244479e for both SQLite and Postgres.

packages/serverpod_database/lib/src/adapters/sqlite/sqlite_analyzer.dart
marcelomendoncasoares: No need to throw, since this is on the analyzer and we're switching over the SQL that comes from the SQLite tables (`PRAGMA foreign_key_list(...)`), with all possible strings covered.

packages/serverpod_shared/test/database_config_test.dart
marcelomendoncasoares: I suppose so as well. Only if the spaces are within quotes they should be kept.
Like `filePath: "    /tmp/serverpod.sqlite".`

packages/serverpod_shared/lib/src/config.dart
marcelomendoncasoares: We can't check for existence, since it is allowed for the database to be created on the first interaction. I would prefer leaving the path as specified by the user and only normalized as needed when it is used by the database.

packages/serverpod_shared/lib/src/config.dart
marcelomendoncasoares: The trim on the `byName` was only to skip a `" "`, but not to defend against `"sqlite "` (with quotes). In this case, I think we should throw instead of fix.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
nielsenko: Nicer indeed

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
nielsenko: Nice

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
nielsenko: 🤔

nielsenko (approved):
LGTM 🎸 

We have a recurring issue with:

> serverpod_test_server-1  | +1654 ~6 -1: test_integration/future_calls/broken_future_calls_test.dart: Given a Serverpod server instance with default config and at least 1000future calls in the database containing unregistered and broken future calls when starting Serverpod then a warning is logged about skipping the check for broken future calls [E]

etc.

But not related to this PR.

tests/docker/tests_sqlite_integration/Dockerfile
marcelomendoncasoares: Missed this! Fixed the comment at https://github.com/serverpod/serverpod/pull/4912/commits/caa3b8e35860918516c86e4af487c01b52dfd8e3.

PR #4936, opened 2026-04-08
feat: Implement the `unique` keyword feature for simplified unique indexes

### LINKED ISSUE #1141: Simplified unique fields in protocol
**Feature: unique fields**
Allow the simplified unique syntax to create unique indexes on fields.

Using the default index name of `table_name__field_name__unique_idx` should guarantee that the index name it self is unique in the db. 

If the index name `table_name__field_name__unique_idx` is manually used at the same time as we generate the same name we need to give an error.

Example:
```yaml
class: Example
table: example
fields:
  exampleField: String, unique
```

Would have the same result as the current way:
```yaml
class: Example
table: example
fields:
  exampleField: String
indexes:
  example_example_field_unique_idx
    fields: exampleField
    unique: true
```

This would save a lot of typing for the programmer.

marcelomendoncasoares:
Closes: #1141

## Pre-launch Checklist

## Breaking changes

None.


tools/serverpod_cli/lib/src/analyzer/models/validation/model_relations.dart
```diff
@@ -105,6 +123,18 @@ class ParsedModelsCollection {
...
   }
 
+  Map<SerializableModelIndexDefinition, ModelClassDefinition>
+  _createAutoGeneratedUniqueIndexesMap(
+    List<SerializableModelDefinition> models,
+  ) {
+    return {
+      for (var model in models)
+        if (model is ModelClassDefinition && model.tableName is String)
+          for (var field in model.fields)
```
nielsenko, line 133: Could use `model.tableName case String name` to avoid the `!` later.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -377,6 +381,20 @@ class SerializableModelFieldDefinition {
         shouldPersist &&
         scope == ModelFieldScopeDefinition.none;
   }
+
+  /// When [shouldCreateUniqueIndex] is true, the btree unique index that is
+  /// auto-created for this field on [tableName]; otherwise `null`.
+  SerializableModelIndexDefinition? autoGeneratedUniqueIndexDefinition(
+    String tableName,
```
nielsenko, line 388: Not related to this PR, but isn't it a bit weird that you cannot get the model from the field (and then table -> tableName, if persisted).  This I guess is what forces you to pass `tableName`?

nielsenko (approved):
LGTM, one comment, and a question, but not critical.

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
marcelomendoncasoares: Totally agree. I miss it every time I work with the serialized model fields. Having such access would reduce a lot of parameters and loops to search for classes/indexes.

PR #4947, opened 2026-04-10
fix: Skip empty migration folders in listVersions

### LINKED ISSUE #3985: Create migration should not throw error when the latest migration folder is empty
### Problem to Solve

When the latest migration version folder is empty - usually to a discard changes operation -, the `create-migration` command will end with an error, requesting the user to manually delete the folder. This workflow is inefficient, since it is frequent that developers discard their changes and end up with the empty folder to be manually removed in order to continue.

```
⠙ Creating migration... (1ms)
Unable to determine latest database definition due to a corrupted
migration. Please re-create or remove the migration version and try
again. Migration version: "20250918013903164".
PathNotFoundException: Cannot open file, path =
'/.../tests/serverpod_test_server/migrations/20250918013903164/definition_project.json'
(OS Error: No such file or directory, errno = 2)
✗ Creating migration (9ms)
```

This message was introduced on PR #1537 as a fix to the crash reported on #1512. It can be improved to not consider an empty folder as a "corrupted state".

### Proposal

Empty folders should be ignored by the versioning evaluation, since they are not really a corrupted state. As the `migrations` folder is managed by the CLI, it could safely remove these empty migration folders. An efficient implementation would be to remove the last folder if empty until it finds one that has files. Only then return this version as the last existing migration.

If such folder has files, but the `definition.json` file is missing, then we throw the exception, since this would be an actual corrupted state.

### Use Case

When developing, it is frequent that we create a migration but end up discarding it to create again after more changes. The discard operation lead to an empty folder that will break the second migration generation, only for the developer to have to manually delete it and call the command again.

### Alternatives

Continue removing the empty folder manually after the error in the command.

### Additional context

N/A.

- [ ] I want to work on this

mvanhorn:
## Summary

When a migration version folder is empty (common after discarding changes), `create-migration` throws a "corrupted migration" error requiring manual folder deletion. Now empty folders are silently skipped.

## Changes

`packages/serverpod_database/lib/src/migrations/migration_artifacts_store/file_system.dart`:

In `listVersions()`, changed the directory listing to check for the presence of `definition_project.json` before including a version in the list. Directories without this file (empty or incomplete) are filtered out.

## Why

When developers discard changes that included a migration, the empty version folder remains. The existing code lists it as a valid version, then fails trying to read files from it. The fix treats an empty folder the same as a non-existent one.

Fixes #3985

This contribution was developed with AI assistance (Claude Code).


mvanhorn:

mvanhorn:
Rebased onto latest main (740b99a). Picked up #4945 (dart-lang/setup-dart 1.7.1 to 1.7.2) which may help normalize CI.

Also applied dart format to the test file in 740b99a. dart format wanted to collapse a few unnecessarily wrapped lines in the tests I added. The core migration fix in file_system.dart is unchanged.

The integration test WebSocket errors on previous runs looked unrelated to this change (logging endpoint timing issues). CI running now.

mvanhorn:
Investigated the three failing checks after 740b99a:

`Serverpod integration tests (server) (3.32.0)` is a pre-existing main flake. It fails on the last 5 merged PRs (#4963, #4959, #4957, #4955, #4950) with the same ~15min timeout pattern.

`Unit tests IO (3.41.5, ubuntu-latest)` had a real bug in this branch. One of the two new tests in `migration_generator_exceptions_test.dart` (the `repairMigration` case) expected a `null` return, but after the `listVersions` fix returns `[]`, `_validateRepairMigrationVersion(null)` threw `MigrationRepairTargetNotFoundException`. Fixed in 6c59157 by returning `null` early when no versions exist, before validation. An explicit `targetMigrationVersion` not in the list still throws. Test passes locally.

`Serverpod SQLite integration tests (server) (3.41.5)` fails `concurrent_migration_test.dart` with 'DB has migration version ... registered but not found in the project files'. This passes on main. Possibly related to the listVersions filter. Waiting on this CI run to see if it persists.

One remaining issue: the `createMigration` test in the same file still fails on Linux CI with `PathAccessException: '/lib/ssl/private'`. That failure originates in `_loadMigrationVersionsFromModules`, past the `listVersions` return, so a separate fix. Happy to look at that next if this push turns the other checks green.

marcelomendoncasoares:
Thank you for your initiative in fixing this issue, @mvanhorn! I was going to nudge this to a different direction, but I see #5044 already goes this way. So I'll close this PR in favor of it.

PR #4952, opened 2026-04-11
fix(cloud-storage): Handle binary uploads correctly on Flutter Web (issue #141)

### LINKED ISSUE #141: Binary file uploads to database storage is not working on Flutter web
The file upload to database cloud storage is not working on web.

- There is a failing test for it.(`cloud_storage_test.dart` with the `-p chrome` option).
- Uploading to S3 seem to work correctly.
- Just gets an `XMLHttpRequest error.` No further information from the exception.
- It may be possible that Flutter web doesn't support 'Content-Type: application/octet-stream' and only multipart files. Needs further investigation.

abdulawalarif:
This PR fixes an issue where **binary file uploads to database cloud storage fail on Flutter Web**.

The failure occurs during web-based upload requests using `Stream<List<int>>`, where the browser throws an `XMLHttpRequest failed` error (`ClientException: Failed to fetch`, statusCode = -1). This prevents successful uploads in Flutter Web while the same functionality works correctly in other platforms.

The root cause appears to be related to how Flutter Web handles raw binary uploads using `application/octet-stream`, which may not be properly supported or may trigger browser-level restrictions (such as CORS preflight or unsupported request formatting).

This fix ensures that cloud storage uploads on Flutter Web are handled correctly and no longer fail during test execution.

### Fixed issues

* Fixes [https://github.com/serverpod/serverpod/issues/141](https://github.com/serverpod/serverpod/issues/141)

### Testing

* Ran `dart test -p chrome test_e2e/cloud_storage_test.dart`
* Verified that previously failing tests now pass successfully
* Ensured no regression in existing cloud storage functionality

---

## Pre-launch Checklist

* [x] I read the Contribute page and followed the process outlined there for submitting PRs.
* [x] This update contains only one single feature or bug fix and nothing else.
* [x] I read and followed the Dart Style Guide and formatted the code with `dart format`.
* [x] I listed at least one issue that this PR fixes in the description above.
* [x] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation.
* [x] I added new tests to check the change I am making.
* [x] All existing and new tests are passing.
* [x] Any breaking changes are documented below.

---

## Breaking changes

None


abdulawalarif:

abdulawalarif:
recheck

packages/serverpod_client/lib/src/file_uploader.dart
```diff
@@ -44,21 +44,33 @@ class FileUploader {
...
+          const isWeb =
+              bool.fromEnvironment('dart.library.js_interop') ||
+              identical(0, 0.0);
+          http.StreamedResponse response;
+          if (isWeb) {
+            var request = http.Request(method, _uploadDescription.url);
+            request.headers.addAll(headers);
+            request.bodyBytes = await stream.toBytes();
+            response = await request.send();
+          } else {
```
nielsenko: Not fan. This prevents streaming uploads on web, forcing all bytes to be read into buffer before starting upload. Binary uploads tend to be big.

Could we instead figure out what is causing streaming to break?

nielsenko (changes requested):

nielsenko:
Thank you for looking into this.

Could you elaborate on your findings? I would like to avoid the switch away from `StreamedRequest`

abdulawalarif:
Hi, @nielsenko, please take a look at the solution now.

abdulawalarif:
Hi, @nielsenko 
The failure was caused by the #4382 change from unawaited(pipe) + send() to (pipe, send).wait, which races on web when uploading a stream without a known length. Reverted to unawaited(stream.pipe(request.sink)) while keeping StreamedRequest. No full-file buffering added, on web package:http already buffers via BrowserClient.finalize().toBytes() before fetch().

nielsenko:
Thanks for digging into this. I don't think the analysis holds up, though.

`(stream.pipe(request.sink), request.send()).wait` evaluates left to right, so the pipe starts before `send()`, exactly as with `unawaited(pipe); await send()`. The only difference is that the pipe's future is now awaited. `StreamedRequest` is backed by a non-broadcast `StreamController`, which buffers events until someone listens, so ordering doesn't matter anyway. On web, `BrowserClient` does `await request.finalize().toBytes()` before calling `fetch`, and that completes once the pipe closes the sink. Both versions behave the same there.

Also, #141 predates #4382 by several years, so #4382 can't be the cause.

The new tests are skipped on the browser (`@OnPlatform({'browser': Skip(...)})`), so they don't exercise the web path. On the VM they should pass with or without the change.

Reverting also has a cost. With `.wait`, an error in the source stream is caught by the surrounding try/catch and upload returns false. With `unawaited`, it becomes an unhandled async error in the caller's zone. Avoiding that was part of the reason for #4382.

To move forward, I'd need a test that runs under `-p chrome` against a real binary upload endpoint, fails on main, and passes with your change. Without that I can't tell whether there's a web issue left to fix, or whether this change addresses it.

nielsenko:
Did a bit of digging myself. This is tested in _cloud_storage_test.dart_ which does run as part of the e2e web tests.

Browser results on main, with the `.wait` code this PR reverts:
- Chrome, dart2js (local run): all 16 tests passed, including both uploads.
- Chrome, dart2wasm (local run): all 16 passed.
- Firefox, CI: the "Serverpod E2E web tests" job already runs this file on every build.

PR #4957, opened 2026-04-13
feat: Infer database dialect from the CLI without manual config

### LINKED ISSUE #4956: Allow the CLI to infer the SQL dialect with no user intervention
### Problem to Solve

Users can now use SQLite after #4785, but a manual action is needed to instruct the `databaseDialect` on the `config/generator.yaml` to `sqlite`. Otheriwse, the CLI will generate SQL for Postgres. The manual step is brittle and easy to forget, both when configuring to use SQLite and when switching for Postgres later.

### Proposal

The ideal UX is that users should just update the `<mode>.yaml` files changing the `database` to have only the `filePath` and the CLI would generate the correct dialect.

### Use Case

Users want to experiment with SQLite and later switch to Postgres without having to configure implementation details.

### Alternatives

Leave the manual workflow as is.

### Additional context

N/A.

- [x] I want to work on this

marcelomendoncasoares:
Closes: #4956

## Pre-launch Checklist

## Breaking changes

None, the `databaseDialect` keyword had not yet been released to the public.

### DIFF
--- a/packages/serverpod_shared/lib/src/config.dart
+++ b/packages/serverpod_shared/lib/src/config.dart
@@ -1073,6 +1073,27 @@ Map? _databaseConfigMap(Map configMap, Map<String, String> environment) {
   ]);
 }
 
+/// Infer the database dialect from one run-mode config map (the body of
+/// `config/<runMode>.yaml`), using the same `database` merging rules as
+/// [ServerpodConfig.loadFromMap].
+///
+/// Returns `null` when there is no database section or it is empty after
+/// merging environment variables. Uses a placeholder password so PostgreSQL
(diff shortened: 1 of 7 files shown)


nielsenko (approved):
LGTM 👍  .. I can imagine that it might not be easily inferable from shape when we add more db types, but let us not get ahead of ourselves.

marcelomendoncasoares:
> LGTM 👍 .. I can imagine that it might not be easily inferable from shape when we add more db types, but let us not get ahead of ourselves.

The function added here is not actually using the config shape, it is using the `DatabaseConfig._fromJson` method to properly load the configs - which can infer from the shape, as for SQLite, but might require an explicit `dialect` in case of conflicting shapes, like for a future `mariadb`.

PR #4959, opened 2026-04-14
chore: Upgrade `sqlparser` and lock the lib version to avoid breaking changes

marcelomendoncasoares:
A new minor version of `sqlparser` was just released [with the removal of the `parseMultiple` method](https://github.com/simolus3/drift/commit/fbc733ed07538640f3de961e6de757d626d04aca#diff-ebeae2424ee28d30fde515969f50fe09663ef7c4a75997559f666dbf0b62012b) that we were using. This PR fixes the upgrade.

Replaces PR #4958.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
+++ b/packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
@@ -6,7 +6,6 @@ import 'package:serverpod_shared/serverpod_shared.dart';
 // We only use the `source_span` package to generated expected inputs for the
 // `sqlparser` package, which depends on it for public interfaces.
 // ignore: depend_on_referenced_packages
-import 'package:source_span/source_span.dart';
 import 'package:sqlite3/common.dart' hide Session;
 import 'package:sqlite_async/sqlite_async.dart';
 import 'package:sqlparser/sqlparser.dart'
@@ -16,8 +15,7 @@ import 'package:sqlparser/sqlparser.dart'
         CommitStatement,
         DeleteStatement,
         InsertStatement,
-        InvalidStatement,
-        SemicolonSeparatedStatements,
+        ParserEntrypoint,
         SqlEngine,
         Statement,
         UpdateStatement;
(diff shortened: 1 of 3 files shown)


packages/serverpod_database/pubspec.yaml
```diff
@@ -20,7 +20,7 @@ dependencies:
   serverpod_shared: 3.5.0-beta.3
   sqlite3: ^2.9.4
   sqlite_async: ^0.13.1
-  sqlparser: ^0.43.1
+  sqlparser: '>=0.44.3 <0.45.0'
```
nielsenko:
```suggestion
  sqlparser: ^0.44.3
```
is literally the same

templates/pubspecs/packages/serverpod_database/pubspec.yaml
```diff
@@ -19,7 +19,7 @@ dependencies:
   serverpod_shared: SERVERPOD_VERSION
   sqlite3: ^2.9.4
   sqlite_async: ^0.13.1
-  sqlparser: ^0.43.1
+  sqlparser: '>=0.44.3 <0.45.0'
```
nielsenko:
```suggestion
  sqlparser: ^0.44.3
```

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -16,8 +15,7 @@ import 'package:sqlparser/sqlparser.dart'
         CommitStatement,
         DeleteStatement,
         InsertStatement,
-        InvalidStatement,
-        SemicolonSeparatedStatements,
+        ParserEntrypoint,
         SqlEngine,
         Statement,
         UpdateStatement;
```
nielsenko, line 21: Why limit the shown symbols? This is an internal used package. We are not exporting these. If there is a conflict I would rather see us import it with a prefix, or hide the conflicting symbol.

nielsenko (changes requested):

packages/serverpod_database/pubspec.yaml
marcelomendoncasoares: You're right, fox 0.x versions they are the same... I was surprised by the "breaking" change on dependabot PR, but failed to realize that there is no breaking at all since it is below stable 🤦 

Fixed at https://github.com/serverpod/serverpod/pull/4959/commits/61ff6728f52a9eafedab67a5265726c804cc7e65.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
marcelomendoncasoares: In this particular case I preferred importing explicit symbols due to the conflicting nature of the library with our package. There are a lot of common symbols (`Column`, `Expression`, `Table`, `ResultSet`) and more that might come in the future that can lead confuse developers. So I preferred being explicit on what we're using than using a prefix or just hide the current conflicts.

nielsenko (approved):
LGTM

PR #4960, opened 2026-04-14
feat: Add jsonb column support and GIN index operator classes

### LINKED ISSUE #3919: Support for jsonb indexable columns

developerjamiu:
Adds JSONB column support for serializable fields, configurable at project/class/field level, with GIN indexes and operator class support.

Carries forward @borutjures's original [PR #3934](https://github.com/serverpod/serverpod/pull/3934) (32 commits preserved) and resolves all 7 review items from the Sep 23 review. Based on alignment call with @marcelomendoncasoares on 2026-04-08.

**Main architectural change:** refactored serialization config resolution from top-down to bottom-up three-pass (field → class → project), each pass isolated to its own level. Also adds SQLite dialect support (merged after original PR stalled), GIN-on-non-jsonb validation, and a `copyWithPrefix` bug fix that was silently dropping operator classes on inherited indexes.

**Usage:**

```yaml
# model - field-level
class: Product
table: product
fields:
  tags: List<String>, serializationDataType=jsonb
indexes:
  tags_gin:
    fields: tags
    type: gin # optional - defaults to gin for jsonb-only indexes
    operatorClass: jsonbPathOps # optional - defaults to jsonbOps

# model - class-level (applies to all serializable fields)
class: Product
table: product
serializationDataType: jsonb
fields:
  tags: List<String>
  metadata: Map<String, String>

# project-level (config/generator.yaml - applies to all models)
serialize_as_jsonb_by_default: true
```

**Known limitation:** migrating an existing `json` column to `jsonb` triggers a destructive warning and requires `--force` or a manual `ALTER COLUMN TYPE`. Matches existing `integer ↔ bigint` precedent; auto-generating the `ALTER` is a reasonable follow-up.

Closes serverpod/serverpod#3919, Closes serverpod/serverpod#3934

## Pre-launch Checklist

## Breaking changes

None.


developerjamiu:

packages/serverpod_database/lib/src/concepts/columns.dart
```diff
@@ -62,13 +62,18 @@ class ColumnByteData extends Column<ByteData> {
...
 }
 
-/// A [Column] holding an [SerializableModel]. The entity will be stored in the
-/// database as a json column.
+/// A [Column] holding a [SerializableModel]. The entity is stored in the
+/// database as a `json` or `jsonb` column, controlled by [serializationDataType].
 class ColumnSerializable<T> extends Column<T> {
+  /// The PostgreSQL storage type used for this column. Defaults to `json` when
+  /// not set.
+  final SerializationDataType? serializationDataType;
```
marcelomendoncasoares: Suggestion: Avoid Postgres-related comments outside of the adapter layer, except if it is related to something only Postgres support and other dialects will throw.

packages/serverpod_database/lib/src/models/gin_operator_class.spy.yaml
```diff
@@ -0,0 +1,9 @@
+### Operator classes used by GIN indexes.
+### See: https://www.postgresql.org/docs/current/gin.html#GIN-BUILTIN-OPCLASSES
+enum: GinOperatorClass
```
marcelomendoncasoares, line 3: Comment: This is an example where it is fine to have a Postgres reference on comments, since GIN indexes are Postgres-specific and will be a no-op in other dialects.

tools/serverpod_cli/test/database/migration/column_types_test.dart
```diff
@@ -48,6 +48,72 @@ void main() {
...
+          expect(
+            migration.actions.last.type,
+            DatabaseMigrationActionType.createTable,
+          );
+          expect(migration.warnings, isNotEmpty);
+          expect(migration.warnings.every((w) => w.destructive), isTrue);
+        },
+      );
+    },
+  );
```
marcelomendoncasoares, line 83: Suggestion: Actually, both Postgres and SQLite support migrating JSON <-> JSONB by casting the type of the column. 

```sql
--postgres
ALTER TABLE your_table
ALTER COLUMN your_column
TYPE jsonb
USING your_column::jsonb;  -- OR `your_column::json` if JSONB -> JSON

-- sqlite (while recreating the table for alter)
INSERT INTO new_my_table (..., json_col)
SELECT ..., jsonb(json_col) FROM my_table;  -- OR `json(json_col)` if JSONB -> JSON
```

We should support such alter table to make the feature useful for existing projects as well.

tools/serverpod_cli/lib/src/database/dialects/sqlite.dart
marcelomendoncasoares: Suggestion: For SQLite, we need to store `jsonb` as `BLOB` using the `jsonb` function instead of plain `TEXT`. We'll also need tests on the `serverpod_test_sqlite_server` package (unfortunately we need to copy from the `serverpod_test_server`, both the models and test files).

tools/serverpod_cli/test/database/extentions/pgsql_code_generation_test.dart
```diff
@@ -656,4 +656,82 @@ END
...
+        expect(
+          sql,
+          'CREATE INDEX "$indexName" ON "${tableDefinition.name}" '
+          'USING gin ("$fieldName");\n',
+        );
+      },
+    );
+
+    test(
+      'when creating a gin index with the jsonbPathOps operator class, then the SQL uses USING gin with the jsonb_path_ops suffix.',
```
marcelomendoncasoares: Suggestion: Prefer describing the objective of the test rather than the objects that represent it. In the current case, it is better to describe as "when creating a gin index with an operator class set, then the SQL uses USING gin with the correct suffix."

tools/serverpod_cli/test/database/extentions/pgsql_code_generation_test.dart
marcelomendoncasoares, line 593: Suggestion: Also mirror this test for the GIN OPS to ensure that all OPS are being correctly mapped to their prefix.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/project_serialization_data_type_test.dart
```diff
@@ -0,0 +1,127 @@
...
+import '../../../../test_util/builders/model_source_builder.dart';
+
+void main() {
+  var config = GeneratorConfigBuilder()
+      .withEnabledSerializeAsJsonbByDefault()
+      .build();
+
+  test(
+    'Given `serializeAsJsonbByDefault` enabled and a class without `serializationDataType` and a field without `serializationDataType`, when validating, then the field `serializationDataType` is jsonb.',
+    () {
```
marcelomendoncasoares, line 19: Suggestion(NIT): Prefer breaking long titles on the description "clauses" to improve readability:


```suggestion
  test(
    'Given `serializeAsJsonbByDefault` enabled and a class without `serializationDataType` and a field without `serializationDataType`, '
    'when validating, '
    'then the field `serializationDataType` is jsonb.',
    () {
```

But avoid breaking a clause mid-sentence, since it gets harder to track the action and result.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/project_serialization_data_type_test.dart
```diff
@@ -0,0 +1,127 @@
...
+      expect(
+        definition.fields.last.type.serializationDataType,
+        SerializationDataType.jsonb,
+      );
+    },
+  );
+
+  test(
+    'Given `serializeAsJsonbByDefault` enabled and a class with `serializationDataType` set to json, when validating, then the field `serializationDataType` is json.',
+    () {
```
marcelomendoncasoares, line 51: Suggestion: Missing only one more test case for project set, class not set and only field set (interaction between project config and field override).

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/indexes/index_type_test.dart
```diff
@@ -14,7 +14,6 @@ void main() {
     'hash',
     'gist',
     'spgist',
-    'gin',
     'brin',
   ];
```
marcelomendoncasoares, line 19: Question: Why remove the `gin` from here?

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/indexes/index_operator_class_test.dart
```diff
@@ -0,0 +1,366 @@
...
+        collector.errors,
+        isNotEmpty,
+        reason: 'Expected an error, but none was collected.',
+      );
+      expect(
+        collector.errors.first.message,
+        'GIN indexes are not supported in SQLite.',
+      );
+    },
+  );
```
marcelomendoncasoares: Suggestion: The philosophy for SQLite on Serverpod is that it should ignore unsupported specifications and only show a warning while creating the migration. The reason for this is because a definition can be part of a module that is developed/tested using SQLite, but will be used in Postgres projects - or vice-versa. As a rule of thumb, you can consider that the `GeneratorConfig.databaseDialect` is only for SQL generation, not for model analysis.

For indexes, we already have the warning in place (see the [DatabaseDefinitionRestrictions class](https://github.com/serverpod/serverpod/blob/b6caa820488ab74f88dec511bae431770c151e41/packages/serverpod_database/lib/src/interface/definition_restrictions.dart#L3-L12) and [this test](https://github.com/serverpod/serverpod/blob/b6caa820488ab74f88dec511bae431770c151e41/tests/serverpod_test_sqlite/serverpod_test_sqlite_server/test_e2e_migrations/unsupported_elements_migration_test.dart#L86-L101) for reference). Since [SQLite already ignores non-btree indexes on SQL generation](https://github.com/serverpod/serverpod/blob/b6caa820488ab74f88dec511bae431770c151e41/tools/serverpod_cli/lib/src/database/dialects/sqlite.dart#L194-L198), there is no need to add validation for this PR.

packages/serverpod_database/lib/src/concepts/columns.dart
```diff
@@ -62,13 +62,18 @@ class ColumnByteData extends Column<ByteData> {
...
+
   /// Creates a new [Column], this is typically done in generated code only.
   ColumnSerializable(
     super.columnName,
     super.table, {
+    this.serializationDataType,
     super.hasDefault,
     super.fieldName,
   });
```
marcelomendoncasoares, line 87: Discussion: After giving some thought, I'll suggest we shift from storing the `jsonb` information from a `serializationDataType` on the `ColumnSerializable` to create a separate `Column` class. It is a complex choice because it breaks the current pattern of `Column*` classes mapping 1:1 with **Dart** types, but it's also the only way for us to later introduce type-safe operations that will only work with `jsonb` columns.

As for the name, to keep the column class abstract from the database dialect, we could call it `ColumnStructured`, since the difference between standard `json` and `jsonb` is that the latter is parsed into a tree structure at the storage level - indexed and queryable on Postgres. For MySQL/MariaDB, for example, this type would be the only used, since the `JSON` type is already binary and there is no text JSON storage.

This does not change the CLI layer, only the way we register the column. The choice between `json` and `jsonb` is mostly for Postgres/SQLite and make sense to expose such clear names to the developer.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -2767,6 +2767,12 @@ class SerializableModelLibraryGenerator {
               : [],
         ),
     ).call(constructorArgs, {
+      if (field.type.isColumnSerializable &&
+          field.type.serializationDataType != null)
+        'serializationDataType': refer(
+          'SerializationDataType.${field.type.serializationDataType!.name}',
+          serverpodProtocolUrl(serverCode),
+        ),
```
marcelomendoncasoares: Comment: With the [change to have a new `ColumnStructured` discussed above](https://github.com/serverpod/serverpod/pull/4960/changes#r3112584668), this will no longer be necessary, since the type of the column will already encode the chosen serialization type.

packages/serverpod_database/lib/src/models/serialization_data_type.spy.yaml
```diff
@@ -0,0 +1,16 @@
...
+values:
+  ### The `json` data type stores an exact copy of the input text,
+  ### which processing functions must reparse on each execution
+  - json
+
+  ### The `jsonb` data is stored in a decomposed binary format
+  ### that makes it slightly slower to input due to added conversion overhead,
+  ### but significantly faster to process, since no reparsing is needed.
+  ### `jsonb` also supports indexing, which can be a significant advantage.
+  - jsonb
```
marcelomendoncasoares: Question: By [changing to the `ColumnStructured` as discussed above](https://github.com/serverpod/serverpod/pull/4960/changes#r3112584668), can this class be demoted to a local enum on the CLI package only? Since the `SerializationDataType` will be only for parsing the column type from the models.

tools/serverpod_cli/lib/src/analyzer/models/stateful_analyzer.dart
```diff
@@ -123,14 +124,31 @@ class StatefulAnalyzer {
...
+      var model = state.model;
+      if (model is! ModelClassDefinition) continue;
+      for (var field in model.fields) {
+        if (field.type.isColumnSerializable &&
+            field.type.serializationDataType == null) {
+          field.type.serializationDataType = SerializationDataType.jsonb;
+        }
+      }
+    }
+  }
```
marcelomendoncasoares: Suggestion: This is not the correct place to resolve the project-level `serializationDataType`. The resolution should be centralized evaluating all three levels at once (project, class, field) while parsing the model on the `serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart`.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -154,10 +159,22 @@ class ModelParser {
...
+      documentContents,
+    );
+    if (modelSerializationDataType != null) {
+      for (var field in fields) {
+        if (field.type.isColumnSerializable &&
+            field.type.serializationDataType == null) {
+          field.type.serializationDataType = modelSerializationDataType;
+        }
+      }
+    }
```
marcelomendoncasoares: Suggestion: [As commented below](https://github.com/serverpod/serverpod/pull/4960/changes#r3113387124), here we should also evaluate the project-level serialization data type.

tests/serverpod_test_server/test_integration/column_operations/column_jsonb_test.dart
```diff
@@ -0,0 +1,167 @@
...
+      jsonbIndexedGin: ['x'],
+      jsonbIndexedGinJsonbPath: ['x'],
+    ),
+  ]);
+
+  _nonEmptyRowId = rows[0].id!;
+  _emptyRowId = rows[1].id!;
+  _nullNullableRowId = rows[2].id!;
+  _emptyNullableRowId = rows[3].id!;
+  _specialCharactersRowId = rows[4].id!;
```
marcelomendoncasoares: Suggestion: This global mutable can be removed in favor for just getting the row of the expected index on the tests - since the insert order is preserved for both Postgres and SQLite.

tests/serverpod_test_server/lib/src/models/object_with_jsonb.spy.yaml
```diff
@@ -0,0 +1,17 @@
+class: ObjectWithJsonb
+table: object_with_jsonb
+fields:
+  notJsonb: List<String>
+  jsonb: List<String>, serializationDataType=jsonb
+  jsonbIndexed: List<String>, serializationDataType=jsonb
+  jsonbIndexedGin: List<String>, serializationDataType=jsonb
+  jsonbIndexedGinJsonbPath: List<String>, serializationDataType=jsonb
+  nullableJsonb: List<String>?, serializationDataType=jsonb
```
marcelomendoncasoares, line 12: Suggestion: Also nice to add a field with a `Map` type and one with an object like `SimpleData`.

tests/serverpod_test_server/lib/src/models/object_with_jsonb.spy.yaml
```diff
@@ -0,0 +1,17 @@
...
+indexes:
+  jsonb_index_gin:
+    fields: jsonbIndexedGin
+    type: gin
+  jsonb_index_gin_with_operator_class:
+    fields: jsonbIndexedGinJsonbPath
+    type: gin
+    operatorClass: jsonbPathOps
```
marcelomendoncasoares, line 20: Suggestion: Missing an index with implicit (undeclared) `gin` type.

tests/serverpod_test_server/test_integration/index_test.dart
```diff
@@ -342,4 +342,104 @@ void main() async {
...
+
+        indexes = table.indexes;
+      });
+
+      test(
+        'then the implicitly declared gin index exists with default jsonb operator class.',
+        () {
+          final index = indexes.firstWhere(
+            (idx) => idx.indexName == 'jsonb_index_gin',
+          );
```
marcelomendoncasoares, line 366: Suggestion: Here says "implicitly declared", but the index is being explicitly declared on the model - even though the operator class is implicit.

tests/serverpod_test_server/test_integration/index_test.dart
```diff
@@ -342,4 +342,104 @@ void main() async {
...
+        () {
+          final column = columns.firstWhere(
+            (idx) => idx.name == 'json',
+          );
+
+          expect(column.columnType, ColumnType.json);
+        },
+      );
+    },
+  );
```
marcelomendoncasoares: Suggestion: This test group is misplaced - and might be unnecessary once we change from `serializationDataType` to `ColumnStructured`.

marcelomendoncasoares (changes requested):
Great work with the PR, @developerjamiu!! All comments from the other PR were correctly addressed. Besides comments for improvement on the implementation and tests, there are two main points that are still not completely addressed:

- Proper support for JSONB as BLOB on SQLite (with integration tests).
- Migration support between JSON <-> JSONB (with migration e2e tests).

There is also one small change in the direction of the original implementation that I put for discussion below to create a new `ColumnStructured` instead of storing a `serializationDataType` field on the `ColumnSerializable`.

About the migration support, although it must be in for the feature to be complete, if you're ok with it, I'd suggest we create an issue and split the work to a follow-up PR to avoid bloating this one too much.

packages/serverpod_database/lib/src/concepts/columns.dart
developerjamiu: Noted. Will make the comment dialect-agnostic as part of the ColumnStructured change.

tools/serverpod_cli/test/database/migration/column_types_test.dart
developerjamiu: Agreed. Verified that both Postgres and SQLite support lossless casting. Per your suggestion, I'll create an issue and split this to a follow-up PR.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/indexes/index_type_test.dart
developerjamiu: I removed it because the "GIN requires jsonb" validation rejects gin on the String field this test uses. In hindsight, the better approach is to keep gin in the loop with a jsonb-compatible field (the same pattern the vector type tests use). I'll restore it.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/indexes/index_operator_class_test.dart
developerjamiu: Understood

packages/serverpod_database/lib/src/concepts/columns.dart
developerjamiu: Agreed. To confirm: `ColumnSerializable` stays for json, `ColumnStructured` for jsonb; the column class itself encodes the storage type, no extra field needed?

packages/serverpod_database/lib/src/models/serialization_data_type.spy.yaml
developerjamiu: Yes. With `ColumnStructured` the runtime no longer needs this. I'll remove it.

developerjamiu:
> Great work with the PR, @developerjamiu!! All comments from the other PR were correctly addressed. Besides comments for improvement on the implementation and tests, there are two main points that are still not completely addressed:
> 
> * Proper support for JSONB as BLOB on SQLite (with integration tests).
> * Migration support between JSON <-> JSONB (with migration e2e tests).
> 
> There is also one small change in the direction of the original implementation that I put for discussion below to create a new `ColumnStructured` instead of storing a `serializationDataType` field on the `ColumnSerializable`.
> 
> About the migration support, although it must be in for the feature to be complete, if you're ok with it, I'd suggest we create an issue and split the work to a follow-up PR to avoid bloating this one too much.

Thanks for the thorough review! Agreed on all three main points:

- SQLite JSONB as BLOB: will implement with integration tests
- ColumnStructured: agreed, separate column class works and could simplify stuffs
- JSON <-> JSONB migration: yes, happy to split to a follow-up PR. I've created issue #4985 for it.

For the remaining comments: I'll push changes and resolve each conversation as addressed.

packages/serverpod_database/lib/src/concepts/columns.dart
marcelomendoncasoares: Exactly, we won't need the extra field on the `ColumnSerializable` anymore.

packages/serverpod_database/lib/src/concepts/columns.dart
```diff
@@ -62,18 +62,25 @@ class ColumnByteData extends Column<ByteData> {
...
     super.columnName,
     super.table, {
-    this.serializationDataType,
+    super.hasDefault,
+    super.fieldName,
+  });
+}
+
+/// A [Column] holding an [SerializableModel]. The entity will be stored in the
+/// database as a structured binary column that supports indexing.
```
marcelomendoncasoares: Suggestion(NIT): Describe more in terms of the structure, since not all dialects support indexing. Here it is useful to go to a lower level and offer more details on what it means for each dialect. Something like:

```suggestion
/// A [Column] holding a [SerializableModel]. The entity will be stored in the
/// database as a binary-encoded object that retains the JSON structure.
/// Stored as JSONB on Postgres (with indexing support) and BLOB on SQLite.
```

tests/serverpod_test_server/test_integration/index_test.dart
```diff
@@ -359,7 +359,7 @@ void main() async {
       });
 
       test(
-        'then the implicitly declared gin index exists with default jsonb operator class.',
+        'then the explicitly declared gin index exists with default jsonb operator class.',
         () {
```
marcelomendoncasoares, line 363: Suggestion: A new implicit index was added to the test class, but the implicit index test is missing.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -1864,6 +1915,45 @@ class Restrictions {
...
+        .firstOrNull;
+    if (field != null &&
+        !field.type.isColumnSerializable &&
+        !field.type.isColumnStructured) {
+      errors.add(
+        SourceSpanSeverityException(
+          'The "${Keyword.serializationDataType}" key is only valid on '
+          'serializable field types (e.g. lists, maps, or custom classes).',
+          span,
+        ),
```
marcelomendoncasoares, line 1943: Suggestion: Missing a reference for serializable models on the list of valid types. Also missing tests for the restrictions on this function.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -1532,7 +1556,34 @@ class Restrictions {
...
+        : <String>[];
+
+    var ginErrors = [
+      if (nonJsonbFields.isNotEmpty)
+        SourceSpanSeverityException(
+          'GIN indexes require the indexed field to use '
+          '"${Keyword.serializationDataType}: jsonb" (fields: ${nonJsonbFields.join(', ')}).',
+          span,
+        ),
+    ];
```
marcelomendoncasoares: Suggestion: This validation is a bit inverted: we should validate that the index type is allowed for its fields, and not that the fields must have different types for the index. The best place to apply the type restriction is on the `validateIndexType`, next to the Vector index validation. We also need tests for these restrictions.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -154,10 +160,23 @@ class ModelParser {
...
+
+    if (effectiveDefault != null) {
+      for (var field in fields) {
+        if (field.type.isColumnSerializable &&
+            field.type.serializationDataType == null) {
+          field.type.serializationDataType = effectiveDefault;
+        }
+      }
+    }
+
```
marcelomendoncasoares: Suggestion: Resolving the project level here is still not the best place, because we're iterating through fields after they are already parsed. The best way is to resolve the `defaultSerializationDataType` inside the `_parseClassFields` and pass the resolved value to the `_parseModelFieldDefinition` so it does not need to be mutated afterwards.

tests/serverpod_test_server/migrations/migration_registry.txt
```diff
@@ -24,3 +24,5 @@
 20260324085914798
 20260407154802260
 20260417182403425
+20260420092904087
+20260421192857921
```
marcelomendoncasoares: Suggestion: Before merging, we should squash these migrations to ensure that only one migration is added per PR. This helps tracking database changes for each feature.

marcelomendoncasoares (changes requested):
Good job with the changes, @developerjamiu! A few more comments and we should be ready to merge.

packages/serverpod_database/lib/src/concepts/columns.dart
developerjamiu: I was avoiding this initially based on your [comment here](https://github.com/serverpod/serverpod/pull/4960#discussion_r3093086853). As you suggested, can make an exception for `ColumnStructured`

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
developerjamiu: Could you please clarify this?

marcelomendoncasoares: Sorry for not being so clear. The "e.g." list on the message skips "serializable models", but they are the primary case for stored JSON/JSONB:

```suggestion
        SourceSpanSeverityException(
          'The "${Keyword.serializationDataType}" key is only valid on serializable '
          'field types (e.g. lists, maps, serializable models or custom classes).',
          span,
        ),
```

developerjamiu: Thanks for the clarity.

tools/serverpod_cli/lib/src/analyzer/models/model_parser/model_parser.dart
```diff
@@ -145,17 +149,24 @@ class ModelParser {
...
+        (config.serializeAsJsonbByDefault ? SerializationDataType.jsonb : null);
+
     var fields = _parseClassFields(
       documentContents,
       docsExtractor,
-      extraClasses,
+      config,
       serverOnly,
+      defaultSerializationDataType,
     );
```
marcelomendoncasoares: Suggestion(NIT): Since the `_parseClassFields` also have access to the `config`, this default resolution can be made inside the `_parseClassFields`. This avoids resolving outside just to pass as a new parameter to the `_parseClassFields` class. Then, only the `_parseModelFieldDefinition` would need the new parameter.

marcelomendoncasoares (approved):
Great work with this PR, @developerjamiu! LGTM 🚀 

I have only one nit comment about the placement of the default resolution to get even closer to the fields resolution, but feel free to address or merge directly.

PR #4961, opened 2026-04-14
feat: Logging revamp

nielsenko:
The TUI needs to display framework log output as structured entries, not just what's already routed to session logs. This PR migrates framework output through a unified LogWriter chain that the VM service subscribes to, so the TUI sees everything.


docs/design/logging.md
```diff
@@ -0,0 +1,317 @@
...
+- No serverpod-specific types (generated `TableRow` classes) in the logging interface.
+
+## Implementation status
+
+- Generic types (`LogLevel`, `LogScope`, `LogEntry`, `LogWriter`), the `Log` class, and shared writers live in `serverpod_shared/lib/src/log/`.
+- `Serverpod` exposes two singletons: `log` (framework) and `sessionLog` (session), each backed by its own writer chain. Session events flow as typed `SessionOpen` / `SessionEntry` / `SessionClose` records, not as generic `LogEntry` values with metadata.
+- The CLI bridges `cli_tools.Logger` to `Log` via `ServerpodCliLogger`.
+- `DatabaseLogWriter`, `SessionTextLogWriter`, and `SessionJsonLogWriter` persist / echo session events; `VmServiceSessionLogWriter` surfaces them to the CLI on the same `ext.serverpod.log` wire channel as framework events.
+- Cross-chain correlation is **not yet wired**: session events go through `sessionLog` correctly, but framework `log.info(...)` calls made from inside an endpoint body have no link back to the enclosing session. They resolve to the synthetic root `LogScope` on the framework chain, so consumers see them as unrelated to the session handling the request.
```
marcelomendoncasoares: (ignore this comment, GitHub is not allowing me to delete)

nielsenko:
Rebased on main

tests/serverpod_test_sqlite/serverpod_test_sqlite_server/test_integration/database_operations/last_database_operation_test.dart
```diff
@@ -9,6 +9,10 @@ void main() async {
   withServerpod(
     'Given a running server',
     testGroupTagsOverride: [TestTags.concurrencyOneTestTag],
+    // Disable health checks to avoid the health check updating the last
+    // database operation time and causing flakiness on the test.
+    configOverride: (config) =>
+        config.copyWith(healthCheckInterval: Duration.zero),
```
marcelomendoncasoares, line 15: Comment: Let's see [who lands first](https://github.com/serverpod/serverpod/pull/4988/commits/613e516be66c7132dba72786beda587f4628cee6) 😆

templates/pubspecs/templates/serverpod_templates/modulename/pubspec.yaml
```diff
@@ -11,6 +11,7 @@ environment:
 dependency_overrides: #--CONDITIONALLY_REMOVE_LINE--#
   serverpod: #--CONDITIONALLY_REMOVE_LINE--#
     path: ../../../packages/serverpod #--CONDITIONALLY_REMOVE_LINE--#
+
   serverpod_database: #--CONDITIONALLY_REMOVE_LINE--#
     path: ../../../packages/serverpod_database #--CONDITIONALLY_REMOVE_LINE--#
   serverpod_serialization: #--CONDITIONALLY_REMOVE_LINE--#
```
marcelomendoncasoares, line 16: Question: Any particular reason for this newline on the templates? It breaks the pattern with the other dependencies.

packages/serverpod_shared/lib/serverpod_shared.dart
```diff
@@ -6,6 +6,10 @@ export 'src/enums.dart';
 export 'src/environment_variables.dart';
 export 'src/exceptions/exit_exception.dart';
 export 'src/exceptions/password_missing_exceptions.dart';
+export 'src/log/global_log.dart';
+export 'src/log/log.dart';
+export 'src/log/log_types.dart';
+export 'src/log/test_log_writer.dart';
```
marcelomendoncasoares: Discussion: Since `log`, `LogLevel` (and related names) are pretty common, exporting them from the main `serverpod_shared` might create a lot of conflicts on user projects. What if we create a second `serverpod_shared/log.dart` barrel file just for the log objects? Then we won't need to "hide" so much classes as we did (and surely users won't too). 

This change will also prevent `serverpod` to forward the logger automatically and require users to opt-in to use it - instead of finding by accident. Which, later, might facilitate a replacement when we move the logger to the `log` package as we plan.

tests/serverpod_test_server/lib/test_util/test_serverpod.dart
```diff
@@ -41,6 +45,38 @@ class IntegrationTestServer extends TestServerpod {
...
+          await server.internalSession.db.unsafeExecute(
+            'DELETE FROM serverpod_runtime_settings',
+          );
+        } catch (_) {
+          // Server or pool may already be gone; nothing to clean up.
+        }
+      });
+    } on StateError {
+      // Called outside a running test (group declaration time). Skip.
+    }
```
marcelomendoncasoares: Suggestion: Instead of blindly cleaning the table after the test, I suggest we move this fix to prevent configs from being persisted. The persistence is only used for Insights: the app can change the `RuntimeSettings` remotely, and the `HealthCheckManager` continuously syncs the database config with the running instance.

During tests, this mechanism is never used, so we can either:

- Set a singleton on tests (maybe on the `Features` singleton) that will disable the store/reload of the settings.
- Expose a parameter on the `Serverpod` class to disable settings store/reload.

https://github.com/serverpod/serverpod/blob/613e516be66c7132dba72786beda587f4628cee6/packages/serverpod/lib/src/server/serverpod.dart#L302-L317

Since this change is unrelated and might require more work, I'll suggest we decouple to another PR.

packages/serverpod_shared/lib/src/utils/isolated_object.dart
marcelomendoncasoares, line 1: Suggestion: Since this is a general utility, is it worth moving to a `src/utils` folder instead of on the `src/log`?

packages/serverpod/lib/src/server/log_manager/log_writers/database_session_log_writer.dart
```diff
@@ -0,0 +1,276 @@
...
+    );
+  }
+
+  protocol.LogLevel _toProtocolLogLevel(LogLevel level) => switch (level) {
+    LogLevel.debug => protocol.LogLevel.debug,
+    LogLevel.info => protocol.LogLevel.info,
+    LogLevel.warning => protocol.LogLevel.warning,
+    LogLevel.error => protocol.LogLevel.error,
+    LogLevel.fatal => protocol.LogLevel.fatal,
+  };
```
marcelomendoncasoares, line 260: Discussion: The duplicated `LogLevel` class comes from the fact that the `protocol.LogLevel` is declared on the `serverpod` package, but the `serverpod_shared` can't depend on it. But what if we transform the `serverpod_shared` in a shared package of `serverpod` - similar to the `serverpod_database` and move the `log_level.spy.yaml` (and other relevant log-related models) to the `serverpod_shared` instead? This will allow us to use the protocol model everywhere, without duplicating it anymore.

tools/serverpod_cli/test/test_util/endpoint_validation_helpers.dart
```diff
@@ -1,11 +1,31 @@
...
+    path.join(uri.toFilePath(), '..', '..', '..', '..'),
+  );
+}
+
 Future createTestEnvironment(
   Directory testProjectDirectory,
   String pathToServerpodRoot,
 ) async {
+  // Ignore the caller-provided value; resolve a stable absolute path here.
+  pathToServerpodRoot = await _resolveServerpodRoot();
```
marcelomendoncasoares: Question: If we're no longer going to use the `pathToServerpodRoot` parameter, shouldn't we remove it? This is indeed a much better approach.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -429,12 +431,15 @@ class Serverpod {
...
     this.runtimeParametersBuilder,
+    ServerpodLogSetup? loggingSetup,
   }) : httpResponseHeaders = httpResponseHeaders ?? _defaultHttpResponseHeaders,
        httpOptionsResponseHeaders =
            httpOptionsResponseHeaders ?? _defaultHttpOptionsResponseHeaders,
        _configOverride = configOverride,
        _securityContextConfig = securityContextConfig,
        _healthConfig = healthConfig ?? const HealthConfig(),
+       _loggingSetup = loggingSetup ?? ServerpodLogSetup.installDefaults(),
+       _ownsLoggingSetup = loggingSetup == null,
```
marcelomendoncasoares: Suggestion: As we discussed, remove `loggingSetup` the parameter and treat it like a private property.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -445,20 +450,55 @@ class Serverpod {
...
+  }
+
   void _initializeServerpod(
     List<String> args, {
     ServerpodConfig? config,
   }) {
+    // Set the singleton early so anything Serverpod's own constructor
+    // touches (e.g. CommandLineArgs warning on parse failure) can reach
+    // [Serverpod.instance].
+    _instance = this;
```
marcelomendoncasoares: Question: Now that we no longer use the `Serverpod.instance` for the logs, can we revert this?

packages/serverpod_shared/lib/src/log/global_log.dart
```diff
@@ -0,0 +1,20 @@
...
+/// logWriter.add(MyLogWriter());
+/// log.info('Server started');
+/// ```
+final MultiLogWriter logWriter = MultiLogWriter([]);
+
+/// Global [Log] that forwards to [logWriter]. Identity is stable: the
+/// instance is constructed at library init and never reassigned. Entry
+/// points configure logging by mutating [logWriter], not by replacing
+/// [log].
+final Log log = Log(logWriter, logLevel: LogLevel.info);
```
marcelomendoncasoares, line 20: Question: To avoid silently overriding instance `log` properties, given how common this name is for methods and properties, should we call them `globalLog` and `globalLogWriter` instead?

packages/serverpod/lib/src/server/log_manager/log_writers/database_session_log_writer.dart
```diff
@@ -0,0 +1,276 @@
...
+    _sessions[event.sessionId] = _SessionState(event, future);
+
+    // record / close observe insert failures via the stored future;
+    // silence here so open itself stays best-effort.
+    await _track(future).catchError((_) => 0);
+  }
+
+  Future<int> _insertOpenRow(Session session, SessionOpen event) async {
+    final row = _buildOpenRow(event);
+    final inserted = await session.db.insertRow<protocol.SessionLogEntry>(row);
```
marcelomendoncasoares: Question: Why not use the default way of inserting through the static `db` of the model?

```suggestion
    final inserted = await protocol.SessionLogEntry.db.insertRow(session, row);
```

packages/serverpod/lib/src/server/log_manager/log_writers/database_session_log_writer.dart
```diff
@@ -0,0 +1,276 @@
...
+        );
+      case SessionQueryEntry e:
+        state.queryCount++;
+        await session.db.insertRow<protocol.QueryLogEntry>(
+          _buildQueryRow(e, sessionLogId, state.open),
+        );
+      case SessionMessageEntry e:
+        await session.db.insertRow<protocol.MessageLogEntry>(
+          _buildMessageRow(e, sessionLogId, state.open),
+        );
```
marcelomendoncasoares, line 120: Question: Same question as the above about the usage of the `db` static property.

```suggestion
      case SessionLogEntry e:
        await protocol.LogEntry.db.insertRow(
          session,
          _buildLogRow(e, sessionLogId, state.open),
        );
      case SessionQueryEntry e:
        state.queryCount++;
        await protocol.QueryLogEntry.db.insertRow(
          session,
          _buildQueryRow(e, sessionLogId, state.open),
        );
      case SessionMessageEntry e:
        await protocol.MessageLogEntry.db.insertRow(
          session,
          _buildMessageRow(e, sessionLogId, state.open),
        );
```

packages/serverpod/lib/src/server/log_manager/log_writers/database_session_log_writer.dart
```diff
@@ -0,0 +1,276 @@
...
+  ) async {
+    int sessionLogId;
+    try {
+      sessionLogId = await state.sessionLogId;
+    } catch (_) {
+      return;
+    }
+
+    final row = _buildCloseRow(state.open, event, sessionLogId);
+    await session.db.updateRow<protocol.SessionLogEntry>(row);
```
marcelomendoncasoares:
```suggestion
    await protocol.SessionLogEntry.db.updateRow(session, row);
```

marcelomendoncasoares (changes requested):
Nice work! This is a much needed refactor of the logs structure. I didn't review in all the log writers to avoid holding this review longer, but I get their structure is similar.

templates/pubspecs/templates/serverpod_templates/modulename/pubspec.yaml
nielsenko: No. Fixed in e9889c65

packages/serverpod_shared/lib/serverpod_shared.dart
nielsenko: Agreed. Done in 1478390f

packages/serverpod_shared/lib/src/utils/isolated_object.dart
nielsenko: Much better. Done in e9889c65

packages/serverpod/lib/src/server/log_manager/log_writers/database_session_log_writer.dart
nielsenko: I would like logging not to depend on serverpod

tools/serverpod_cli/test/test_util/endpoint_validation_helpers.dart
nielsenko: Fixed in c069f644

packages/serverpod/lib/src/server/serverpod.dart
nielsenko: Done (616b3b80)

packages/serverpod/lib/src/server/serverpod.dart
nielsenko: c6e28915

packages/serverpod_shared/lib/src/log/global_log.dart
nielsenko: Not sure I like that ..

packages/serverpod/lib/src/server/log_manager/log_writers/database_session_log_writer.dart
nielsenko: Fixed in 0636dae3

packages/serverpod/lib/src/server/log_manager/log_writers/database_session_log_writer.dart
nielsenko: Fixed in 0636dae3

packages/serverpod/lib/src/server/log_manager/log_writers/database_session_log_writer.dart
nielsenko: 0636dae3

tests/serverpod_test_server/lib/test_util/test_serverpod.dart
nielsenko: We do use `LogSettingsBuilder().withLoggingTurnedDown()...)` in tests.. 

Anyway I agree this can be moved to a separate PR

@coderabbit Create a follow-up issue

nielsenko: Reverted in 5eb62810. This will likely cause test failure.

tests/serverpod_test_sqlite/serverpod_test_sqlite_server/test_integration/database_operations/last_database_operation_test.dart
nielsenko: You did. Merged in main in 56cb538

packages/serverpod/lib/src/server/log_manager/log_writers/database_session_log_writer.dart
nielsenko: As discussed I have removed `log_level.spy.yaml` and made `LogLevel` from `serverpod_shared` a custom class instead in `generator.yaml`

nielsenko: Done in 75211c0

marcelomendoncasoares (approved):
As discussed, since the implementation is complete, lets merge as is to focus on the other Jetstream features and leave the `LogLevel` consolidation and runtime settings fix on tests as follow-ups.

packages/serverpod/lib/src/server/log_manager/log_writers/database_session_log_writer.dart
nielsenko: This has been rolled back.

@coderabbit Can you create a follow up issue to reintroduce 75211c0, ie. consolidate `LogLevel` so only the one in `serverpod_shared` is used.

tests/serverpod_test_server/lib/test_util/test_serverpod.dart
nielsenko: re-added in 02573dd (alternative fix is delayed)

PR #4973, opened 2026-04-15
feat: Implement template enhancement with mustache syntax for conditional rendering

### LINKED ISSUE #4965: Refactor the project creation templates to prepare for interactive component selection
Currently, our templates are basically split into three:

- Serverpod mini
- Serverpod server
- Serverpod module

However, for #1643 to be possible, we need to be able to apply components more selectively. So a refactor is needed in the templates to break the upgrade packages in a way that we can offer the following options for users:

- Which database to use (Postgres, SQLite, None).
- With or without Redis.
- Whether to add and configure Auth with Email (requires a database).
- Whether to add and compile a Web app (serving included).

Crazelu:
This PR introduces enhancements to the templates to support conditional rendering using mustache template directives. This requires a template context that will be built from the create command. For now, the context is hardcoded with true values to maintain the existing behaviour where no conditional rendering happens.

Closes [4965](https://github.com/serverpod/serverpod/issues/4965)

## Pre-launch Checklist

## Breaking changes

None


templates/serverpod_templates/projectname_server_upgrade/config/passwords.yaml
```diff
@@ -16,7 +16,9 @@ shared:
 # These are passwords used when running the server locally in development mode
 development:
   database: 'DB_PASSWORD'
```
marcelomendoncasoares, line 19: Suggestion: The database password is only for Postgres.

templates/serverpod_templates/projectname_server_upgrade/config/passwords.yaml
marcelomendoncasoares, line 33: Suggestion: These are only for auth.

marcelomendoncasoares, line 1: Discussion: Missing the markers on the `modulename` as template as well. Alos, for module development with SQLite, there is no need entirely for this file. Should we mark it in a way we can remove it for this case?

templates/serverpod_templates/projectname_server_upgrade/lib/server.dart
```diff
@@ -1,19 +1,26 @@
+// {{#web}}
 import 'dart:io';
+// {{/web}}
```
marcelomendoncasoares, line 4: Question: Should the marker also include the empty line afterwards? Or we'll apply a dart format regardless?

templates/serverpod_templates/projectname_server_upgrade/{{#docker}}docker-compose{{!docker}}.yaml
marcelomendoncasoares, line 1: Question: Should we also annotate the file with a tag that allow us to completely remove it if neither Postgres nor Redis are required?

templates/serverpod_templates/projectname_server_upgrade/docker-compose.yaml
marcelomendoncasoares, line 53: Suggestion: Also annotate the volumes as postgres.

tools/serverpod_cli/lib/src/commands/create.dart
```diff
@@ -123,11 +123,20 @@ class CreateCommand extends ServerpodCommand<CreateOption> {
       }
     }
 
+    // TODO(Lucky): Build context from interactive user options/sane defaults.
```
marcelomendoncasoares: Suggestion: No need for the `TODO`, since we already have an issue for it.

tools/serverpod_cli/lib/src/commands/create.dart
```diff
@@ -123,11 +123,20 @@ class CreateCommand extends ServerpodCommand<CreateOption> {
...
+    final templateContext = <String, Object?>{
+      'redis': true,
+      'postgres': true,
+      'auth': true,
+      'web': true,
+    };
```
marcelomendoncasoares: Suggestion: Make the context a real Dart object, so we can have type-safety on the options. The conversion to Map for mustache can be a method of such object. Then we can also use this object to provide augmented tags, like a getter `bool get docker => postgres || redis;` for the file or broader section filters.

templates/serverpod_templates/projectname_server_upgrade/lib/server.dart
marcelomendoncasoares: Already saw that we'll format it, so no need to include the space.

tools/serverpod_cli/lib/src/create/create.dart
```diff
@@ -151,10 +157,20 @@ Future<bool> performCreate(
...
+  }
+
   success &= await log.progress('Getting workspace dependencies.', () {
     return CommandLineTools.dartPubGet(serverpodDirs.projectDir);
   });
 
+  success &= await log.progress('Formatting server file.', () {
+    return CommandLineTools.dartFormat(
+      File(p.join(serverpodDirs.serverDir.path, 'lib', 'server.dart')),
+    );
```
marcelomendoncasoares: Question: Should we rather format the whole server, since we might have more conditionals on other files in the future?

tools/serverpod_cli/lib/src/create/create.dart
```diff
@@ -323,6 +342,17 @@ Future<bool> _performUpgrade(
   return success;
 }
 
+/// Parses and renders the template files in the given directory.
+Future<bool> _renderTemplates(
+  Directory dir,
+  Map<String, Object?> context,
+) async {
+  return await log.progress('Finalizing project files', () async {
```
marcelomendoncasoares: Suggestion(NIT): Finalizing can pass the impression that the command is finishing.

```suggestion
  return await log.progress('Applying template options', () async {
```

tools/serverpod_cli/lib/src/create/create.dart
marcelomendoncasoares: Or better, move the format to the `TemplateRenderer` in case the rendered file is a `.dart` file. Then we only format the files that are actually touched.

tools/serverpod_cli/test/commands/create/template_renderer_test.dart
```diff
@@ -0,0 +1,431 @@
...
+      templateRenderer = TemplateRenderer(dir: testDir);
+    });
+
+    tearDown(() async {
+      await testDir.delete(recursive: true);
+    });
+
+    test(
+      'when rendering a directory with conditional naming, '
+      'and a true context value, '
```
marcelomendoncasoares: Suggestion(NIT): Prefer leaving each clause in one line, so it gets easier to spot the "when" and the "then" of the test (even if it makes the line longer; many tests do and it's fine).

marcelomendoncasoares: Despite the above comment, this is actually the given (test scenario), not the acting.
- Given a directory with conditional naming and a true context value, 
  - when rendering the template, 
    - then the directory name is formatted to remove the template directive.

If the goal to group them under "Given a TemplateRenderer" was to share the temp dir and `TemplateRenderer`, just move them to the top-level under `main` and keep the tests separate - since the same comment applies to all below tests.

tools/serverpod_cli/test/commands/create/template_renderer_test.dart
marcelomendoncasoares, line 1: Suggestion: Separate the tests in two files: file_name and file_contents.

marcelomendoncasoares (changes requested):
Great work with this refactor, @Crazelu!! A true QoL improvement with developing the templates. In the future, we might even get rid of the `*upgrade` projects completely and have only one mustache-annotated template for both module and server. Test coverage is also great for the template renderer!

Additionally to the comments, a few pieces that are yet missing on this PR:
- SQLite database configuration: must change all `database` sections on the run mode files.
- Removing Redis on the run mode files if not selected.
- Template markers on the `modulename` template.

For discussion: Should we add integration tests for the `performCreate` function to validate the templates here or leave it to the next PR with bootstrap tests? Currently, `performCreate` is not directly tested, but testing the interactive options on the bootstrap tests might also be way harder than just testing it here.

templates/serverpod_templates/projectname_server_upgrade/config/passwords.yaml
Crazelu: We can put the entire content of the file in a conditional directive. When the file is rendered, it'll become empty. TemplateRender deletes empty files upon rendering.

Crazelu: Alternatively, the file name can also include a computed conditional directive derived from the values for postgres and redis.

templates/serverpod_templates/projectname_server_upgrade/{{#docker}}docker-compose{{!docker}}.yaml
Crazelu: Sure. This can work with a context value computed from redis and postgres values.

tools/serverpod_cli/lib/src/create/create.dart
Crazelu: From my tests, formatting the whole server takes up significantly more time and increases the total time for creating a Serverpod project. I opted for formatting the `server.dart` file alone as it was the only one to need it. Moving the formatting to `TemplateRenderer` to reduce the scope of the formatting is good.

tools/serverpod_cli/test/commands/create/template_renderer_test.dart
Crazelu: I think a better separation is directory vs file.
Where tests for file combines tests for file contents and file names.

Crazelu:
> Should we add integration tests for the performCreate function to validate the templates here or leave it to the next PR with bootstrap tests? Currently, performCreate is not directly tested, but testing the interactive options on the bootstrap tests might also be way harder than just testing it here.

@marcelomendoncasoares We should leave it to the next PR with the bootstrap tests. Even if `performCreate` is tested separately, we will still need tests to validate that the correct context object is built from the options selected through the TUI. If that is possible, then bootstrap tests is the best place.

tools/serverpod_cli/lib/src/commands/create.dart
Crazelu: Done in [b23deea](https://github.com/serverpod/serverpod/pull/4973/commits/b23deea2a3a72bdd336892930491e90f1df46abe)

tools/serverpod_cli/lib/src/create/create.dart
Crazelu: Formatting now moved into `TemplateRenderer` only for dart files that are changed.
[251e85c](https://github.com/serverpod/serverpod/pull/4973/commits/251e85cb43e01883c7ca99d32d2016a17a78c48b)

tools/serverpod_cli/test/commands/create/template_renderer_test.dart
Crazelu: Directory vs file separation done in [e5d0530](https://github.com/serverpod/serverpod/pull/4973/commits/e5d053020c69a62183d70d14a322524d441ef787)

templates/serverpod_templates/projectname_server_upgrade/{{#docker}}docker-compose{{!docker}}.yaml
Crazelu: Done in [f46624e](https://github.com/serverpod/serverpod/pull/4973/commits/f46624e36c7ce476b908e07a14106c8dd5be468e)

templates/serverpod_templates/projectname_server_upgrade/config/passwords.yaml
Crazelu: File name now has the template directive in [f46624e](https://github.com/serverpod/serverpod/pull/4973/commits/f46624e36c7ce476b908e07a14106c8dd5be468e)

templates/serverpod_templates/modulename_server/{{#docker}}docker-compose{{!docker}}.yaml
```diff
@@ -10,13 +11,17 @@ services:
...
     image: redis:6.2.6
     ports:
       - '9091:6379'
     command: redis-server --requirepass "REDIS_TEST_PASSWORD"
     environment:
       - REDIS_REPLICATION_MODE=master
-
+  # {{/redis}}
+# {{#postgres}}
 volumes:
```
marcelomendoncasoares, line 26: Suggestion: Place the newline inside the `postgres` group so that the `volumes` key gets separated from the prior when rendering.

templates/serverpod_templates/projectname_server_upgrade/config/passwords.yaml
marcelomendoncasoares, line 1: Suggestion: This file can't actually be removed on the project without `docker` because it has the `serviceSecret` for each run mode that Insights will need to access the server. Even though SQLite does not support logging for Insights to read, the CLI uses the Insights endpoint to analyze the database (as of yet) and generate repair migrations. Without a service secret, repair migrations can't work. This is only for project, since modules have no service secret (nor real running server) to use repair migrations.

templates/serverpod_templates/projectname_server_upgrade/{{#docker}}docker-compose{{!docker}}.yaml
```diff
@@ -46,7 +46,9 @@ services:
     environment:
       - REDIS_REPLICATION_MODE=master
   # {{/redis}}
-
+  
+# {{#postgres}}
 volumes:
```
marcelomendoncasoares, line 51: Suggestion: Same as above to embed the newline in the `postgres` tag.

tools/serverpod_cli/lib/src/create/template_context.dart
```diff
@@ -0,0 +1,43 @@
+/// Context containing values for rendering templates.
+class TemplateContext {
+  TemplateContext({
+    this.auth = false,
+    this.redis = false,
+    this.postgres = false,
+    this.sqlite = false,
+    this.web = false,
+  });
```
marcelomendoncasoares, line 9: Question: Should we call it "context" or is it clearer "TemplateOptions"?

tools/serverpod_cli/lib/src/create/template_context.dart
```diff
@@ -0,0 +1,43 @@
...
+
+  /// True if docker is enabled.
+  bool get docker => postgres || redis;
+
+  /// True if a database is enabled.
+  bool get database => postgres || sqlite;
+
+  Map<String, bool> toJson() {
+    return {
+      'auth': auth & database, // auth requires a database
```
marcelomendoncasoares: Suggestion: Auth requires postgres, as SQLite will fail for most (if not all) IDPs due to the lack of concurrent transaction support. It may be possible to use SQLite for some providers by disabling rate limiting, but it is brittle. We can revisit in the future, but it is best to disable for now.

marcelomendoncasoares (changes requested):
Did not reviewed the new perform create tests, but a few follow-ups on the previous state.

tools/serverpod_cli/lib/src/create/template_context.dart
Crazelu: `Context` is a well known name when it comes to template rendering. I think that is clear and portable even if we move away from mustache to another syntax.

templates/serverpod_templates/projectname_server_upgrade/config/passwords.yaml
Crazelu: I see

tools/serverpod_cli/lib/src/create/template_context.dart
Crazelu: Done in [3d29fc9](https://github.com/serverpod/serverpod/pull/4973/commits/3d29fc9d2b62fad338070ac99f7bacfd4fed993b)

templates/serverpod_templates/projectname_server_upgrade/config/passwords.yaml
Crazelu: Done in [a3e594c](https://github.com/serverpod/serverpod/pull/4973/commits/a3e594ce8bfd82fab72d046e9b5add1b9d9f9e1b)

.github/workflows/dart-tests.yaml
```diff
@@ -426,7 +426,10 @@ jobs:
 
           colima start
       - name: Run unit tests
-        run: dart test --concurrency=1 --reporter=failures-only
+        shell: bash
+        run: |
+          export SERVERPOD_HOME="$(cd ../.. && pwd)"
+          dart test --concurrency=1 --reporter=failures-only
```
marcelomendoncasoares, line 432: Question: Why do we need this change on the CI?

templates/pubspecs/templates/serverpod_templates/projectname_server/pubspec.yaml
```diff
@@ -10,9 +10,6 @@ environment:
 
 dependencies:
   serverpod: SERVERPOD_VERSION
-  # {{#auth}}
-  serverpod_auth_idp_server: SERVERPOD_VERSION
-  # {{/auth}}
```
marcelomendoncasoares: Suggestion: Open an issue to refactor what is needed for the dependencies to also use the mustache template.

templates/serverpod_templates/projectname_server_upgrade/config/development.yaml
```diff
@@ -36,19 +39,24 @@ database:
...
   #user: # defaults to empty
   #requireSsl: true # defaults to false
+# {{/redis}}
 
 maxRequestSize: 524288 # The maximum size of requests allowed in bytes
 
 sessionLogs:
+  # {{#database}}
   persistentEnabled: true
+  # {{/database}}
```
marcelomendoncasoares: Suggestion: Escaped from me on the last review... Persistent logging is actually only supported on Postgres, because SQLite does not support concurrent writes.

templates/serverpod_templates/projectname_server_upgrade/config/passwords.yaml
```diff
@@ -15,44 +15,64 @@ shared:
...
 # Passwords used in your staging environment if you use one. The default setup
 # use a password for Redis.
 staging:
+  # {{#postgres}}
   database: 'DB_STAGING_PASSWORD'
+  # {{/postgres}}
   serviceSecret: 'SERVICE_SECRET_STAGING'
 
+  # {{#auth}}
   emailSecretHashPepper: 'EMAIL_SECRET_HASH_PEPPER_STAGING'
```
marcelomendoncasoares, line 60: Suggestion(NIT): Make the empty line part of the `auth` flat to avoid a space of two empty lines between the `serviceSecret` and the `production` key if `auth` is not used. The same is valid for the `development` and `test` keys.

tools/serverpod_cli/lib/src/analyzer/dart/endpoints_analyzer.dart
```diff
@@ -277,12 +277,14 @@ class EndpointsAnalyzer {
...
+      for (var context in collection.contexts) {
+        var result = await context.currentSession.getResolvedLibrary(filePath);
+        if (result is ResolvedLibraryResult) {
+          return result;
+        }
       }
-    }
+    } catch (_) {}
     return null;
   }
```
marcelomendoncasoares, line 290: Question: Why did we need this change (and the one on the future calls)?

tests/bootstrap_project/test/perform_create/database_test.dart
```diff
@@ -0,0 +1,698 @@
...
+      'get',
+    ], workingDirectory: cliProjectPath);
+    assert(await pubGetProcess.exitCode == 0);
+  });
+
+  group(
+    'Given a TemplateContext with postgres enabled and redis disabled, '
+    'when performCreate is called with the context and a server template type',
+    () {
+      final projectName =
```
marcelomendoncasoares: Suggestion: I'd suggest separating these tests a bit:

1. Make one `redis_test.dart` that tests only redis present/absent.
2. Make one `postgres_test.dart` that tests only postgres present/absent.
3. The same for a `sqlite_test.dart` (currently missing on the tests).

This will allow us to isolate what appears when the option is true, regardless of others. It will also reduce the "combined" tests that can hide flaws (see the comment on the `docker_test.dart`).

For the tests order, might be better to always start with the full scenario and then proceed with the reduced ones.

tests/bootstrap_project/test/perform_create/docker_test.dart
```diff
@@ -0,0 +1,227 @@
...
+      );
+
+      setUpAll(() async {
+        setupForPerformCreateTest();
+        await performCreate(
+          projectName,
+          ServerpodTemplateType.server,
+          false,
+          interactive: false,
+          context: TemplateContext(redis: true),
```
marcelomendoncasoares: Suggestion: When splitting the `database_test.dart` in one file per component, these tests should be part of the component test to ensure that we cover all configurations. Otherwise, relying on a test that says "if one or other" but uses only "one" option mean that we're not validating that the "other" also produces the result.

marcelomendoncasoares: Also, aren't these tests duplicated with similar ones on the `database_test.dart`?

marcelomendoncasoares (changes requested):
Nice work with the tests! The structure is very clear to read. Only a few questions/suggestions for improvement.

.github/workflows/dart-tests.yaml
Crazelu: `SERVERPOD_HOME` needs to be set for the performCreate tests.
It wasn't required for the existing bootstrap tests because they start Process where the environment variable can be configured per test.

templates/serverpod_templates/projectname_server_upgrade/config/development.yaml
Crazelu: There are loads of SQLite restrictions... wish that wasn't the case

templates/serverpod_templates/projectname_server_upgrade/config/passwords.yaml
Crazelu: I'm not too excited about these arbitrary whitespace placements. I will look at proper YAML formatting subsequently.

tests/bootstrap_project/test/perform_create/database_test.dart
Crazelu: I left out sqlite tests because there are no configurations in the templates for it yet. Seems right to do that in the quickstart PR.

tools/serverpod_cli/lib/src/analyzer/dart/endpoints_analyzer.dart
Crazelu: The performCreate tests produce non-normalized file paths that cause endpoint and future call analyzers to throw.

I have made that change clearer in [8096594](https://github.com/serverpod/serverpod/pull/4973/commits/8096594f605ed263fc630a8fba8228d8b65ad62c) by normalizing paths first.

tests/bootstrap_project/test/perform_create/docker_test.dart
Crazelu: Indeed.
The separation is done now and that has removed the duplication

templates/pubspecs/templates/serverpod_templates/projectname_server/pubspec.yaml
Crazelu: @coderabbitai please create this issue

templates/serverpod_templates/projectname_server_upgrade/config/passwords.yaml
marcelomendoncasoares: I agree that a YAML format would be better, but might also be too opinionated and end up worst than this arbitrary spacing. To avoid increasing the scope, I'd solve this now just adjusting the tags and leave a better formatting to when we revisit all templates.

The `development` and `test` keys are already correct, only `staging` and `production` are missing the newline inside the `auth` tag.

tests/bootstrap_project/test/perform_create/sqlite_test.dart
```diff
@@ -0,0 +1,244 @@
...
+      );
+
+      test(
+        'then the server production config file does not contain sqlite configurations',
+        () async {
+          final file = File(p.join(serverDir, 'config', 'production.yaml'));
+          final content = await file.readAsString();
+          expect(content, isNot(contains('filePath: ${projectName}_prod.db')));
+        },
+      );
```
marcelomendoncasoares, line 156: Suggestion: To make these tests more robust to refactor of the `filePath` pattern, change the expect to `isNot(contains('filePath:'))`. The happy path can still use the full `contains` with the pattern, it's only the negative ones that are safer without the actual path. This is also valid for the module with SQLite disabled test.

marcelomendoncasoares (approved):
LGTM! 🚀 

Just two minor comments, but feel free to merge when finished.

PR #4980, opened 2026-04-17
fix: Fix concurrent reads without transaction not being allowed from within a transaction zone on SQLite

marcelomendoncasoares:
For the SQLite implementation, we skipped the support of reading without a transaction from inside a transaction due to `sqlite_async` not supporting this natively. But, this is only forbidden in SQLite if the read comes from the same zone as the transaction.

From [`sqlite_async` tests](https://github.com/powersync-ja/sqlite_async.dart/blob/399dbf47df07fe10bef5582de6de362e35ea309c/packages/sqlite_async/test/native/basic_test.dart#L208-L237), the way to bypass this is to run the query in a forked zone of the transaction parent zone.

```dart
    test(
        'should allow read-only db calls within transaction callback in separate zone',
        () async {
      final db = await testUtils.setupDatabase(path: path);
      await createTables(db);

      // Get a reference to the parent zone (outside the transaction).
      final zone = Zone.current;

      // Each of these are fine, since it could use a separate connection.
      // Note: In highly concurrent cases, it could exhaust the connection pool and cause a deadlock.

      await db.writeTransaction((tx) async {
        // Use the parent zone to avoid the "recursive lock" error.
        await zone.fork().run(() async {
          await db.getAll('SELECT * FROM test_data');
        });
      });

      await db.readTransaction((tx) async {
        await zone.fork().run(() async {
          await db.getAll('SELECT * FROM test_data');
        });
      });

      await db.readTransaction((tx) async {
        await zone.fork().run(() async {
          await db.execute('SELECT * FROM test_data');
        });
      });
```

This PR applies this approach to the SQLite implementation.

For testing, the original `find` tests were restored (equally to the `serverpod_test_server` package), but I preferred also adding a separate one targeting concurrent transactions to add more details on this fix, in case the implementation needs to be revisited in the future.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
+++ b/packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
@@ -38,6 +38,8 @@ class SqliteDatabaseConnection extends DatabaseConnection<SqlitePoolManager> {
 
   static final _sqlEngine = SqlEngine();
 
+  Zone? _currentTransactionParentZone;
+
   SqliteDatabase get _db => poolManager.database;
 
   @override
@@ -776,6 +778,10 @@ class SqliteDatabaseConnection extends DatabaseConnection<SqlitePoolManager> {
     required DatabaseSession session,
   }) async {
     try {
+      // Set a reference to the parent zone (outside the transaction) to allow
+      // for concurrent reads to operate while a write lock is held. Will skip
+      // assigning if already set to preserve the highest parent zone.
+      _currentTransactionParentZone ??= Zone.current;
       return await _db.writeTransaction<R>((tx) async {
         var transaction = _SqliteTransaction(tx, session);
         final result = await transactionFunction(transaction);
@@ -786,6 +792,8 @@ class SqliteDatabaseConnection extends DatabaseConnection<SqlitePoolManager> {
       });
     } on _TransactionCancelledException catch (e) {
       return e.result;
+    } finally {
+      _currentTransactionParentZone = null;
     }
   }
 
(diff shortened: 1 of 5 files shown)


nielsenko (approved):
In sqlite in WAL mode readers are never blocked by each other or by writers. Only writers block each other. 

And reading within the writer transaction should read whatever the transaction would commit. 

This feels like a somewhat odd way to get the natural behavior from sqlite, needed due to some internal bookkeeping in sqlite_async.

Is this correctly understood? 

Anyway, reading the tests, and seeing them pass is convincing. I just wish the code would read more natural.

Crazelu:
A little hesitant about this disclaimer:
> Note: In highly concurrent cases, it could exhaust the connection pool and cause a deadlock.

Could we possibly track the number of forked zones to ensure they are within the range allowed by the connection pool?

marcelomendoncasoares:
> In sqlite in WAL mode readers are never blocked by each other or by writers. Only writers block each other.
> 
> And reading within the writer transaction should read whatever the transaction would commit.
> 
> This feels like a somewhat odd way to get the natural behavior from sqlite, needed due to some internal bookkeeping in sqlite_async.
> 
> Is this correctly understood?
> 
> Anyway, reading the tests, and seeing them pass is convincing. I just wish the code would read more natural.

I was also surprised by this, because, as you said, WAL mode on SQLite is precisely to allow concurrent reads. The constraint is on the `sqlite_async` implementation, that enforces this lock per `Zone` - reading concurrently worked already if the call did not come from inside the `transaction` zone, even with the transaction held open.

My guess is that this behavior is somewhat borrowed from Drift, in which operations are run in a transaction if they run inside a `transaction` zone (there is no `transaction` parameter on the operations).

marcelomendoncasoares:
> A little hesitant about this disclaimer:
> 
> > Note: In highly concurrent cases, it could exhaust the connection pool and cause a deadlock.
> 
> Could we possibly track the number of forked zones to ensure they are within the range allowed by the connection pool?

We can create an issue with an improvement to await a free connection instead of leaving the deadlock to happen - although I guess in Postgres we also do not prevent the user from exhausting the connection pool. [By default, `sqlite_async` will provide 5 connections](https://github.com/powersync-ja/sqlite_async.dart/blob/399dbf47df07fe10bef5582de6de362e35ea309c/packages/sqlite_async/lib/src/sqlite_options.dart#L83-L86), so it's not hard to track how many are available.

marcelomendoncasoares:
> > A little hesitant about this disclaimer:
> > > Note: In highly concurrent cases, it could exhaust the connection pool and cause a deadlock.
> > 
> > 
> > Could we possibly track the number of forked zones to ensure they are within the range allowed by the connection pool?
> 
> We can create an issue with an improvement to await a free connection instead of leaving the deadlock to happen - although I guess in Postgres we also do not prevent the user from exhausting the connection pool. [By default, `sqlite_async` will provide 5 connections](https://github.com/powersync-ja/sqlite_async.dart/blob/399dbf47df07fe10bef5582de6de362e35ea309c/packages/sqlite_async/lib/src/sqlite_options.dart#L83-L86), so it's not hard to track how many are available.

@coderabbitai Please create this issue from the discussion.

PR #4981, opened 2026-04-17
perf: Improves `Insights` queries performance by 60x

marcelomendoncasoares:
Drastically improves Insights logs performance by changing the way the logs are queried in the database. Previously, we would do 1+3N queries (301 queries per 100 lines request) to join the information from the different log tables. This PR introduces object relations on the log tables to allow fetching all logs with a single query.

It also adds a one index to a FK column and the order column to existing indexes to improve performance of the logs queries.

On the `serverpod_test_server`, this change reduced the time to fetch 100 lines of logs from 2-3 seconds to 0.05 seconds (60x faster).

## Pre-launch Checklist

## Breaking changes

None. A migration will be required to recreate the Insights indexes, but this is acceptable even between patch versions.


packages/serverpod/lib/src/endpoints/insights.dart
```diff
@@ -92,37 +92,37 @@ class InsightsEndpoint extends Endpoint {
...
+        logs: LogEntry.includeList(
+          orderByList: (t) => [t.order.asc()],
+        ),
+        queries: QueryLogEntry.includeList(
+          orderByList: (t) => [t.order.asc()],
+        ),
+        messages: MessageLogEntry.includeList(
+          orderByList: (t) => [t.order.asc()],
+        ),
+      ),
```
nielsenko, line 105: 😄 🚀

nielsenko (approved):
Nice!

PR #4984, opened 2026-04-17
feat: Long lived MCP server

### LINKED ISSUE #4968: The `serverpod start` MCP server should work even without a server running
When no `serverpod start` have been run, the MCP fails, which creates the wrong idea that something is not right. Since this is expected behavior, we should change to make the MCP green even when no server is running.

Image

nielsenko:
This PR replaces the project-local, runner-bound MCP server with a **long-lived stdio bridge** that survives runner restarts and discovers `serverpod start --watch` instances at runtime.

Previously each `serverpod start --watch` instance owned an MCP socket under its project's `.dart_tool/serverpod/mcp.sock`, and `serverpod mcp` was a thin stdio-to-socket pipe. Stopping or restarting the runner tore down the AI client's MCP server and Claude Code does not auto-reconnect, so the integration broke on every restart. It also required the AI client to be launched from the right project directory.

`serverpod mcp` is now a persistent bridge. It scans a shared per-machine socket directory, lets the client pick an instance via a `connect` tool, and forwards tool/resource calls. Runners are ephemeral; the bridge stays up across restarts and the client just calls `connect` again.

### Highlights

- **Two-process design.** Runner-side `ServerpodMcpServer` exposes per-instance dev tools; bridge-side `BridgeMcpServer` discovers instances and forwards. Single-instance bridge - one connected runner at a time keeps tool semantics unambiguous.
- **Shared socket directory** at `<systemTemp>/serverpod/serverpod-<pid>-<project>.sock`, created with mode 0700 on POSIX (via `mkdtemp(3)` + rename) so other local users on Linux can't reach the sockets in `/tmp`. Stale sockets for dead PIDs are cleaned up opportunistically on scan. Liveness checks via `kill -0` (POSIX) and `tasklist` (Windows). Windows compatibility for AF_UNIX reparse points is handled in directory listing.
- **Native bridge tools/resource:** `connect`, `disconnect`, `spawn`, `stop`, plus `serverpod://instances`. Auto-picks the runner if exactly one is alive.
- **Auto-forwarding.** On `connect`, the bridge calls `listTools()`/`listResources()` upstream and registers thin forwarders for each item not in the bridge's native deny set, then unregisters them on `disconnect`. New runner-side tools appear in the AI client automatically. Subscribes to forwarded resources to re-emit `resources/updated` upstream.
- **Expanded runner surface.** `apply_migrations` (existing), plus new `create_migration`, `hot_reload`, `tail_logs` tools and the `serverpod://vm-service` resource. `create_migration` shares a new `createMigrationAction` helper with the `serverpod create-migration` CLI command and the TUI "Create Migration" button.

### ~Supporting~ Accidental changes (stuff that should really have been in separate PRs 😬 ) 

- **`serverpod_test`:** `withServerpod` now survives mid-setup failures - if `createTransaction`/`addSavepoint` throws, the transaction stack is unlocked so the next test group starts clean instead of inheriting a half-built transaction. `tearDownAll` runs DB-touching cleanup in parallel and is robust to start-up failure.
- **TUI improvements:** rebuilds throttled to 30 FPS; spinner animation shared via `SpinnerScope` to avoid N independent timers; loading screen falls back to a plain layout in small terminals and pads on black for light terminals; fixes for bright-terminal color handling.
- **Watch session:** skips code generation when a changed `.dart` file cannot affect the protocol (no `extends *Endpoint` or `FutureCall` declaration). Conservative on missing files / I/O errors.
- **Design doc** (`docs/design/mcp_server.md`) rewritten around the two-process bridge model, including a migration note for existing AI client configs.

### Tests

- `bridge_mcp_server_test.dart` covers `connect`/`disconnect`/`spawn`/`stop`, instance auto-pick, switching, auto-forwarding registration/unregistration, and disconnect-on-runner-death.
- `socket_directory_test.dart` covers path construction, project name sanitization, dir-creation race, stale-PID cleanup, and Windows AF_UNIX listing.
- `mcp_server_test.dart` extended for `create_migration`, `hot_reload`, `tail_logs`, and `serverpod://vm-service`.
- `watch_session_test.dart` adds coverage for the protocol-change classifier.

### Issues fixed

_(Please add the relevant issue number(s) - none referenced in commits.)_

## Pre-launch Checklist

## Breaking changes

Note these changes are breaking only in relation to previously release _beta_ versions.

- **Project-local MCP socket removed.** `<server>/.dart_tool/serverpod/mcp.sock` is no longer created or used. Anything that pointed an MCP client directly at that path will stop working - point at `serverpod mcp` instead, which now discovers instances independently of working directory.
- **`serverpod mcp --directory` / `-d` removed.** The bridge does not need a server directory; it scans the shared socket dir. Existing AI client configs that just invoke `serverpod mcp` continue to work; configs that pass `-d` should drop the flag.
- **Bumped `dart_mcp` to `^0.5.x`** (was `^0.4.x`). Adds `stream_transform` as a direct dependency for socket-watcher event coalescing.
- The `serverpod mcp` command no longer fails fast with "no running serverpod watch instance found" - it stays up waiting for one. Scripts that relied on the old non-zero-exit-on-no-instance behavior will need to change.

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -914,25 +914,46 @@ Future<void> _runTuiBackend({
   }
 }
 
+/// Maps a [CreateMigrationOutcome] to a `(message, isError)` pair shared by
+/// the TUI and MCP wrappers. [forceHint] is the surface-specific instruction
+/// for retrying past warnings (e.g. `--force` for the CLI/TUI, `force: true`
+/// for the MCP tool).
+({String message, bool isError}) _describeCreateMigration(
+  CreateMigrationOutcome outcome, {
+  required String forceHint,
+}) {
+  return switch (outcome) {
+    CreateMigrationCreated(:final versionName, :final migrationDirectory) => (
+      message: 'Migration "$versionName" created at $migrationDirectory.',
+      isError: false,
+    ),
+    CreateMigrationNoChanges() => (
+      message: 'No schema changes detected; no migration created.',
+      isError: false,
+    ),
+    CreateMigrationAborted() => (
+      message: 'Migration aborted due to warnings. $forceHint',
+      isError: true,
+    ),
+    CreateMigrationFailed(:final message) => (
+      message: message,
+      isError: true,
+    ),
+  };
+}
+
 /// Runs `create-migration` for the TUI's Create Migration button.
 ///
 /// Logs the outcome; throws on failure so [runTrackedAction] marks the
 /// operation red.
 Future<void> _runCreateMigrationForTui(GeneratorConfig config) async {
   final outcome = await createMigrationAction(config: config);
-  switch (outcome) {
-    case CreateMigrationCreated(:final versionName):
-      log.info('Migration created: $versionName');
-    case CreateMigrationNoChanges():
-      log.info('No schema changes detected; no migration created.');
-    case CreateMigrationAborted():
-      throw Exception(
-        'Migration aborted due to warnings. Run `serverpod create-migration '
-        '--force` to create it anyway.',
-      );
-    case CreateMigrationFailed(:final message):
-      throw Exception(message);
-  }
+  final result = _describeCreateMigration(
+    outcome,
+    forceHint: 'Run `serverpod create-migration --force` to create it anyway.',
+  );
+  if (result.isError) throw Exception(result.message);
+  log.info(result.message);
 }
 
 /// Runs `create-migration` for the MCP `create_migration` tool. Returns a
@@ -947,25 +968,15 @@ Future<CreateMigrationMcpResult> _createMigrationForMcp(
     tag: tag,
     force: force,
   );
-  return switch (outcome) {
-    CreateMigrationCreated(:final versionName, :final migrationDirectory) =>
-      CreateMigrationMcpResult(
-        message:
-            'Migration "$versionName" created at $migrationDirectory. '
-            'Call `apply_migrations` to run it against the database.',
-      ),
-    CreateMigrationNoChanges() => const CreateMigrationMcpResult(
-      message: 'No schema changes detected; no migration created.',
-    ),
-    CreateMigrationAborted() => const CreateMigrationMcpResult(
-      message:
-          'Migration aborted due to warnings. Call again with `force: true` '
-          'to create it anyway.',
-      isError: true,
-    ),
-    CreateMigrationFailed(:final message) => CreateMigrationMcpResult(
-      message: message,
-      isError: true,
-    ),
-  };
+  final result = _describeCreateMigration(
+    outcome,
+    forceHint: 'Call again with `force: true` to create it anyway.',
+  );
+  final followUp = outcome is CreateMigrationCreated
+      ? ' Call `apply_migrations` to run it against the database.'
+      : '';
+  return CreateMigrationMcpResult(
+    message: result.message + followUp,
+    isError: result.isError,
+  );
 }
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -377,6 +377,7 @@ Future<int> _runWatchMode({
     serverArgs: serverArgs,
     serverpodToolDir: serverpodToolDir,
     vmServiceInfoFile: vmServiceInfoFile,
+    project: config.name,
     shutdownSignal: shutdownSignal,
     watcher: FileWatcher(
       watchPaths: {
@@ -415,6 +416,7 @@ Future<int> _startWatchSession({
   required List<String> serverArgs,
   required String serverpodToolDir,
   required String vmServiceInfoFile,
+  required String project,
   required Future<int> shutdownSignal,
   required FileWatcher watcher,
   required Set<String> generatedDirPaths,
@@ -504,9 +506,7 @@ Future<int> _startWatchSession({
   // current MCP tools require the compiler and process lifecycle.
   McpSocketServer? mcpSocket;
   if (!noFes) {
-    mcpSocket = McpSocketServer(
-      socketPath: p.join(serverpodToolDir, 'mcp.sock'),
-    );
+    mcpSocket = McpSocketServer(project: project);
     try {
       await mcpSocket.start();
       mcpSocket.connect(onApplyMigration: session.applyMigration);
@@ -817,9 +817,7 @@ Future<void> _runTuiBackend({
 
     // Start MCP socket server.
     McpSocketServer? mcpSocket;
-    mcpSocket = McpSocketServer(
-      socketPath: p.join(serverpodToolDir, 'mcp.sock'),
-    );
+    mcpSocket = McpSocketServer(project: config.name);
     try {
       await mcpSocket.start();
       mcpSocket.connect(onApplyMigration: session.applyMigration);
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -820,7 +820,10 @@ Future<void> _runTuiBackend({
     mcpSocket = McpSocketServer(project: config.name);
     try {
       await mcpSocket.start();
-      mcpSocket.connect(onApplyMigration: session.applyMigration);
+      mcpSocket.connect(
+        onApplyMigration: session.applyMigration,
+        getLogHistory: () => holder.state.logHistory.toList(),
+      );
       log.info('MCP server listening on ${mcpSocket.socketPath}');
     } on SocketException catch (e) {
       log.warning('Failed to start MCP server: $e');
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -512,6 +512,8 @@ Future<int> _startWatchSession({
       mcpSocket.connect(
         onApplyMigration: session.applyMigration,
         onHotReload: session.forceReload,
+        getVmServiceUri: () => session.vmServiceUri,
+        vmServiceUriChanges: session.vmServiceUriChanges,
       );
       log.info('MCP server listening on ${mcpSocket.socketPath}');
     } on SocketException catch (e) {
@@ -827,6 +829,8 @@ Future<void> _runTuiBackend({
         onApplyMigration: session.applyMigration,
         onHotReload: session.forceReload,
         getLogHistory: () => holder.state.logHistory.toList(),
+        getVmServiceUri: () => session.vmServiceUri,
+        vmServiceUriChanges: session.vmServiceUriChanges,
       );
       log.info('MCP server listening on ${mcpSocket.socketPath}');
     } on SocketException catch (e) {
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -509,7 +509,10 @@ Future<int> _startWatchSession({
     mcpSocket = McpSocketServer(project: project);
     try {
       await mcpSocket.start();
-      mcpSocket.connect(onApplyMigration: session.applyMigration);
+      mcpSocket.connect(
+        onApplyMigration: session.applyMigration,
+        onHotReload: session.forceReload,
+      );
       log.info('MCP server listening on ${mcpSocket.socketPath}');
     } on SocketException catch (e) {
       log.warning('Failed to start MCP server: $e');
@@ -822,6 +825,7 @@ Future<void> _runTuiBackend({
       await mcpSocket.start();
       mcpSocket.connect(
         onApplyMigration: session.applyMigration,
+        onHotReload: session.forceReload,
         getLogHistory: () => holder.state.logHistory.toList(),
       );
       log.info('MCP server listening on ${mcpSocket.socketPath}');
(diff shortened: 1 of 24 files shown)


packages/serverpod_test/lib/src/with_serverpod.dart
marcelomendoncasoares, line 1: Question: Are these changes related to the current MCP work? Asking to understand the motivation behind, if they were to fix a flaky test or some consequence of the MCP changes.

tools/serverpod_cli/lib/src/mcp/socket_directory.dart
```diff
@@ -0,0 +1,151 @@
...
+///
+/// Lower-cases, replaces non-alphanumeric (other than hyphens) with hyphens,
+/// collapses runs of hyphens, and strips leading/trailing hyphens.
+String sanitizeProjectName(String name) {
+  return name
+      .toLowerCase()
+      .replaceAll(RegExp(r'[^a-z0-9-]'), '-')
+      .replaceAll(RegExp(r'-+'), '-')
+      .replaceAll(RegExp(r'^-|-$'), '');
+}
```
marcelomendoncasoares, line 78: Question: Can't these replacements of duplicated chars (like "___" becoming "-") lead to probable conflicts? We could hash the name instead, to avoid a "serverpod_mcp" and "serverpod-mcp" mapping to the same name.

tools/serverpod_cli/test/mcp/bridge_mcp_server_test.dart
```diff
@@ -0,0 +1,284 @@
...
+              arguments: {'instanceId': project},
+            ),
+          );
+          expect(
+            connectResult.isError,
+            anyOf(isNull, isFalse),
+            reason:
+                'connect should succeed: '
+                '${(connectResult.content.first as TextContent).text}',
+          );
```
marcelomendoncasoares: Suggestion: For the "connected" tests, split in a separate group with the "connected" as part of the "Given" - since the server connected is part of the test condition. This will also reduce a lot of code duplication on the tests.

tools/serverpod_cli/test/mcp/bridge_mcp_server_test.dart
```diff
@@ -0,0 +1,284 @@
...
+            ),
+          );
+        },
+      );
+
+      test(
+        'when disconnect is called after connect, '
+        'then subsequent apply_migrations errors',
+        () async {
+          runner.connect(onApplyMigration: () async {});
```
marcelomendoncasoares: Suggestion: Here the "disconnected" could also be a separate group ("Given a disconnected server...").

tools/serverpod_cli/test/commands/start/watch_session_test.dart
```diff
@@ -722,4 +730,203 @@ void main() {
...
+        initialServer: classifierServer,
+        generatedDirPaths: {'/generated'},
+        classifyProtocolChange: (_) async => false,
+      );
+    });
+
+    test(
+      'when a non-protocol dart file changes, '
+      'then generation is skipped and only compile + reload run',
+      () async {
```
marcelomendoncasoares: Question: What does it mean when a protocol-change classifier returns false or true? Feels like the test scenarios added in this file are being described from the implementation perspective rather from the outside observed perspective.

tools/serverpod_cli/test/commands/start/mcp_socket_test.dart
```diff
@@ -1,38 +1,39 @@
...
 
     setUp(() async {
-      tmpDir = await Directory.systemTemp.createTemp('mcp_socket_test_');
-      final socketPath = p.join(tmpDir.path, 'mcp.sock');
-      server = McpSocketServer(socketPath: socketPath);
+      // Project name kept short: macOS limits Unix socket paths to ~104 chars
+      // and the systemTemp prefix can use ~50 of them on its own.
+      server = McpSocketServer(
+        project: 'mst${DateTime.now().microsecondsSinceEpoch % 100000}',
+      );
```
marcelomendoncasoares, line 19: Question: Can this be a problem for real usage on MacOS? Shouldn't we validate/prevent this when creating the socket file?

tools/serverpod_cli/test/commands/start/mcp_server_test.dart
```diff
@@ -56,12 +56,20 @@ void main() {
 
     test(
       'when listing tools, '
-      'then apply_migrations is available',
+      'then apply_migrations, create_migration, hot_reload, and tail_logs are '
+      'available',
```
marcelomendoncasoares: Suggestion(NIT): The overall pattern that we have on the descriptions is to break on the "clauses" only, not to keep the 80 char paging. Breaking otherwise makes it harder to detect each "clause" when the "Given" or "when" exceed the chars.

tools/serverpod_cli/lib/src/mcp/socket_directory.dart
```diff
@@ -0,0 +1,151 @@
...
+  final dir = Directory(serverpodMcpSocketDir);
+  if (dir.existsSync()) return;
+
+  final temp = Directory.systemTemp.createTempSync('serverpod-mcp-init-');
+  try {
+    temp.renameSync(serverpodMcpSocketDir);
+  } on FileSystemException {
+    // Lost the race against another runner that created the dir first.
+    temp.deleteSync();
+  }
```
marcelomendoncasoares, line 48: Question: Why create and rename instead of just create inside a try/catch?

tools/serverpod_cli/lib/src/mcp/bridge_mcp_server.dart
```diff
@@ -0,0 +1,537 @@
...
+    addResource(_instancesResource, _readInstances);
+
+    bridge.onSocketsChanged = () {
+      if (ready) updateResource(_instancesResource);
+    };
+  }
+
+  // ---------------------------------------------------------------------------
+  // Connection state
+  // ---------------------------------------------------------------------------
```
marcelomendoncasoares: Question: Do we need the flower boxes? They are a bit off on the pattern of the repo. I'd be more favorable to add doc comments on the methods, even the private ones, if we feel that they require explanation.

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -884,3 +913,59 @@ Future<void> _runTuiBackend({
...
+          'Migration aborted due to warnings. Call again with `force: true` '
+          'to create it anyway.',
+      isError: true,
+    ),
+    CreateMigrationFailed(:final message) => CreateMigrationMcpResult(
+      message: message,
+      isError: true,
+    ),
+  };
+}
```
marcelomendoncasoares, line 982: Question: Can this be unified with the create migration action to avoid introducing more copies of the migration creation?

marcelomendoncasoares (changes requested):
Nice! Just a few small questions/comments.

tools/serverpod_cli/lib/src/mcp/socket_directory.dart
nielsenko: No, each .sock also includes the pid, so this is not important for uniqueness.

packages/serverpod_test/lib/src/with_serverpod.dart
nielsenko: Flaky tests. Not CI, but would leak connections. Hit while iterating on this. Should have created a separate PR.

tools/serverpod_cli/test/commands/start/watch_session_test.dart
nielsenko: Fair. Will rework.

tools/serverpod_cli/test/commands/start/mcp_socket_test.dart
nielsenko: It is an unfortunate fact of life. macOS / BSD `sockaddr_un.sun_path` is 104 bytes; on macOS `Directory.systemTemp` is something like `/var/folders/xy/abc.../T/` which alone is ~50 chars, leaving ~50 for `serverpod-<pid>-<project>.sock`. A long project name (say >35 chars) would push it over.

I think the limit is slightly higher on Linux and Windows, but since they are less wasteful with the size of he `Directory.systemTemp` path, it is less likely to happen.

We do throw if it happens, but perhaps not the best of error messages.

tools/serverpod_cli/lib/src/mcp/socket_directory.dart
nielsenko: We want a known name. The reason to use  `createTempSync` and not just `createSync` is that it uses `mkdtemp` under the hood on Linux, which ensures that it is only readable to the user creating it. That is not the case for `Directory.systemTemp` which often just points to `/tmp`.

This is not a problem on MacOS and Windows, but harmless.

tools/serverpod_cli/lib/src/commands/start.dart
nielsenko: Both wrappers already share `createMigrationAction`. I can collapse them into one switch that returns `({String message, bool isError})`, with the TUI wrapper throwing on `isError == true`. Will refactor.

nielsenko: fixed in c869e04

tools/serverpod_cli/lib/src/mcp/bridge_mcp_server.dart
nielsenko: [3ee3dad](https://github.com/serverpod/serverpod/pull/4984/commits/3ee3dad83c7ae97f7b1b3fd8f59a59d0b58bcb8f)

tools/serverpod_cli/test/commands/start/mcp_socket_test.dart
nielsenko: Better error handling in [3957b67](https://github.com/serverpod/serverpod/pull/4984/commits/3957b6761a14ffce3148ed3be56bbfb559dcdc02)

tools/serverpod_cli/lib/src/mcp/socket_directory.dart
nielsenko: Explained better in edd5da9

tools/serverpod_cli/test/commands/start/watch_session_test.dart
nielsenko: Done in [4890fb5](https://github.com/serverpod/serverpod/pull/4984/commits/4890fb5e458bb709bf91381703e7a6510833dc69)

tools/serverpod_cli/test/mcp/bridge_mcp_server_test.dart
nielsenko: [cb4ef45](https://github.com/serverpod/serverpod/pull/4984/commits/cb4ef451d0366a4034e3a80fb2e9c01c939f6df6)

tools/serverpod_cli/test/mcp/bridge_mcp_server_test.dart
nielsenko: [cb4ef45](https://github.com/serverpod/serverpod/pull/4984/commits/cb4ef451d0366a4034e3a80fb2e9c01c939f6df6)

packages/serverpod_test/lib/src/with_serverpod.dart
nielsenko: Should I split it out?

tools/serverpod_cli/test/commands/start/mcp_socket_test.dart
```diff
@@ -94,6 +96,41 @@ void main() {
...
+          await expectLater(
+            server.start,
+            throwsA(
+              isA<SocketException>().having(
+                (e) => e.message,
+                'message',
+                allOf(
+                  contains('Unix socket path'),
+                  contains('too long'),
+                ),
```
marcelomendoncasoares: Suggestion: Expect the entire message instead of multiple partial blocks, since it is part of the UX.

tools/serverpod_cli/test/mcp/bridge_mcp_server_test.dart
```diff
@@ -210,75 +166,100 @@ void main() {
...
 
-          final applyResult = await client.callTool(
-            CallToolRequest(name: 'apply_migrations'),
+        group('Given the bridge is then disconnected', () {
+          setUp(() async {
+            final result = await client.callTool(
+              CallToolRequest(name: 'disconnect'),
+            );
+            expect(result.isError, anyOf(isNull, isFalse));
+          });
```
marcelomendoncasoares: Suggestion: This given is nested inside another given, forming an incorrect test description. It needs to be moved out to the same level as the parent.

packages/serverpod_test/lib/src/with_serverpod.dart
marcelomendoncasoares: If the fix is because of the way we're handling connections on this PR, I think it's fine to keep it here. And maybe just add an issue describing the previously existing problem so we can track the reasons behind the change in the future.

marcelomendoncasoares (approved):
LGTM! 🤖 

Have two more small suggestions for the test, but I'll already leave it approved.

packages/serverpod_test/lib/src/with_serverpod.dart
nielsenko: Not really. This PR doesn't fiddle with database connections

nielsenko: Moved out of PR

PR #4988, opened 2026-04-22
refactor: Only store column information on SQLite for non-basic types

marcelomendoncasoares:
The SQLite implementation uses an untracked table (created on the migration file directly) to store additional information about the columns. This is required because SQLite only has 4 basic column types (`INTEGER`, `REAL`, `TEXT`, `BLOB`) that maps all `ColumnType`s.

As a result, each `definition.sql` and `migration.sql` end up with a lot of lines just for inserting the column types into the table. This PR reduces the amount of lines by skipping the basic column types (`ColumnType.integer`/`ColumnType.bigint`, `ColumnType.doublePrecision`, `ColumnType.bytea`, `ColumnType.text`) that map 1:1 with the SQLite basic types.

The only side-effect of this change is the loss of the `int`/`bigint` distinction on SQLite. Most of the reduction on the rows come from skipping the `ColumnType.bigint` type, since it is the one used for `int` ID columns. However, since on SQLite it is already the same as `int` and the migration system already skips migrations between them, this does not really matter.

## Pre-launch Checklist

## Breaking changes

None.


nielsenko (approved):
Small and tidy 👍 LGTM

nielsenko:
image
nice

PR #4994, opened 2026-04-24
feat: Generate a client-side database from the models

### LINKED ISSUE #4783: ☂️ Generate a client-side SQLite database using Serverpod models

### LINKED ISSUE #4791: Allow specifying client-side tables on models
To use a client-side database, users will need a way to distinguish whether models with the `table` keyword should generate the repository methods on the server, client or both.

## Proposal

To allow this, we could introduce a new `database` keyword that would accept the values:

- `server` (default)
- `client`
- `all` (mirrored table between server and client).

If the `database` keyword is omitted when a `table` keyword is present, the default would be to generate the repository only on the server, preserving the current behavior.

To create client-side tables, users would then do:

```yaml
class: Greeting
table: greeting
database: client # or server, all
fields:
  message: String
  author: String
  timestamp: DateTime

indexes:
  author_index:
    type: btree # Only type supported due to SQLite restrictions
```

The result of the above model would be the dual of what happen today: a model with no repository on the server and a model with repository on the client. If `database` is set to `all`, both sides would receive the repository methods.

## Restrictions

For tables targeting `client` or `all`, model restrictions would be applied considering the most restrictive dialect - which will almost always be SQLite. For example, a model with `database` set to `all` would only allow `btree` indexes, since it is the only type supported by SQLite.

In the future, it is possible to introduce `clientOnly` and `serverOnly` keyword for elements such as indexes to allow a more flexible declaration according to each dialect.

### LINKED ISSUE #4792: Expose a method to create a `DatabaseSession` on the client
Once the repository methods are generated on the client (#4791), a `DatabaseSession` will be needed to interact with them - following the same pattern as on the server.

For this, a new `createSession` method could be generated on the `Client` class to be used as:

```dart
final session = client.createSession();
final models = await Model.db.find(session);
```

Differently from the server, it would be a single session for the lifetime of the application (passed around using providers). This is similar to the experience with other packages that provide a local database, with the difference that usually it is the database that is passed around.

The `createSession` method would be generated only if at least one client-side table is declared on the models.

### LINKED ISSUE #4793: Adapt the migration system to generate and run client-side migrations

### LINKED ISSUE #5021: Migrations for SQLite is sometimes generating invalid syntax
### Describe the bug

After https://github.com/serverpod/serverpod/pull/4988, a regression was introduced where the generated `INSERT` statement for the `serverpod_sqlite_schema` table would be incomplete, leading to an `INSERT` error when trying to apply migrations.

### To reproduce

Create a project with one model having a table that sorts last (like "z_table_name"), one column with a non-basic type (like UUID) and the last column with a basic type. Like:

```yaml
class: MyTable
table: z_table_name
fields:
  name: String
```

The generated migration SQL will be broken.

### Expected behavior

Migrations should work.

### Library version

3.5.0-beta.4

### Platform information

N/A

### Additional context

_No response_

- [x] I want to work on this

marcelomendoncasoares:
This PR contain all pieces left to allow the creation of client-side databases. Besides having separate issues, it would be impossible to test any of the pieces in isolation. To avoid carrying bugs from PRs that would look complete, but not tested in runtime, I decided to open a single PR that validates the feature end-to-end.

For an easier review, this PR is best reviewed commit by commit.

Closes: #4791
Closes: #4792
Closes: #4793
Closes: #4783
Fixes: #5021

## About testing

One third of the changes of this PR come from the tests added on commit https://github.com/serverpod/serverpod/pull/4994/commits/701092201a66bed69c0069fb94531a063fcc345b to validate the runtime behavior of the client-side database. Such tests do not need review as they are a mere copy of the `serverpod_test_sqlite_server` relation tests - only adapted to create the client session.

The tests are a small subset of the integration because the database is the same SQLite database that runs on the server - the same logic applies for migration tests.

The only area that showed unexpected behaviors was CRUD involving relations, because they depend on private/implicit properties of the models that only existed on the server-side. All other database features that are not tied to the model internals can be considered tested by the existing SQLite integration tests.

## Pre-launch Checklist

## Breaking changes

None.


packages/serverpod_database/lib/src/migrations/migration_manager.dart
```diff
@@ -330,3 +334,57 @@ abstract class MigrationManager {
...
+    // are available for shared package models. Currently, only the server and
+    // client packages have access to the [DatabaseMigrationVersion] model.
+    late DatabaseResult result;
+    try {
+      result = await session.db.unsafeQuery(
+        'SELECT * FROM "serverpod_migrations";',
+        transaction: transaction,
+      );
+    } catch (_) {
+      return [];
```
nielsenko: Why not bubble here? Base class `_updateState` (where this is called) already swallows any exception (also without any trace btw).

packages/serverpod_database/lib/src/migrations/migration_artifacts.dart
```diff
@@ -56,10 +95,17 @@ abstract interface class MigrationArtifactStore {
...
+  /// Reads the stored migration SQL for running migrations.
+  ///
+  /// Does not include the project definition. Returns `null` if the version
+  /// cannot be found.
+  Future<MigrationVersionSql?> readVersionSql(String version);
+
+  /// Reads the stored definition for a migration version.
   ///
-  /// Returns `null` if the version cannot be found.
-  Future<MigrationVersionArtifacts?> readVersion(String version);
```
nielsenko, line 62: ~Hmm.. is this a breaking change?~

packages/serverpod_database/lib/src/migrations/migration_artifacts_store/runtime_list.dart
```diff
@@ -0,0 +1,68 @@
...
+class RuntimeListMigrationArtifactStore implements MigrationArtifactStore {
+  /// Creates a store from the given [migrations] (same [moduleName] for all).
+  RuntimeListMigrationArtifactStore(
+    this.migrations, {
+    required this.moduleName,
+  }) : _byVersion = {for (final m in migrations) m.version: m},
+       assert(
+         migrations.isEmpty ||
+             migrations.every((m) => m.moduleName == moduleName),
+         'All migrations must be from the "$moduleName" module',
```
nielsenko, line 15: Should be also assert they are sorted?

packages/serverpod_database/lib/src/migrations/migration_manager.dart
```diff
@@ -94,7 +94,9 @@ abstract class MigrationManager {
     if (availableVersions.isEmpty) return null;
 
     var latestVersion = availableVersions.last;
```
nielsenko, line 96: This assumes `availableVersions` are sorted...

I have zeroed in on this.

I guess maintaining a `SplayTreeSet` so `values` are sorted and you can lookup by version without duplicating a list and a map would be safest.

packages/serverpod_database/lib/src/migrations/migration_artifacts_store/runtime_list.dart
nielsenko: I think I prefer a `SplayTreeSet<MigrationVersionSql> _migrations` with a proper compare function.

tools/serverpod_cli/lib/src/migrations/generator.dart
nielsenko, line 37: seems brittle. This flows all the way from command line, or?
```dart
      versionName += '-$tag';
```

tools/serverpod_cli/lib/src/migrations/client_side/client_migration_dart_emitter.dart
```diff
@@ -0,0 +1,194 @@
...
+    ).format(source);
+  }
+}
+
+String migrationClassSuffix(String version) {
+  return version.replaceAllMapped(RegExp(r'[^0-9a-zA-Z]'), (_) => '_');
+}
+
+String _partUri(String version) {
+  if (version.contains('..') || version.contains(r'\')) {
```
nielsenko, line 151: What about ' ', '\', '"', "'"? This should santizied in `createVersionName` I guess.
```dart
  static String createVersionName(String? tag) {
    var now = DateTime.now().toUtc();
    var fmt = DateFormat('yyyyMMddHHmmssSSS');
    var versionName = fmt.format(now);
    if (tag != null) {
      versionName += '-$tag';
    }
    return versionName;
  }
```
in particular given how `tag` is just appended.

Would be nice if version was a richer object than `String`.

packages/serverpod_database/lib/src/migrations/migration_manager.dart
```diff
@@ -330,3 +334,57 @@ abstract class MigrationManager {
...
+    DatabaseSession session, {
+    Transaction? transaction,
+  }) async {
+    return null;
+  }
+
+  @override
+  Future<String?> _loadLatestDefinitionModuleName() async {
+    return (_artifactStore as RuntimeListMigrationArtifactStore).moduleName;
+  }
```
nielsenko: Defeats the abstraction doesn't it?

tools/serverpod_cli/lib/src/migrations/client_side/client_migration_artifact_store.dart
```diff
@@ -0,0 +1,188 @@
...
+          : migration,
+    );
+  }
+
+  @override
+  Future<MigrationVersionSql?> readVersionSql(String version) async {
+    throw UnsupportedError(
+      'Reading version SQL from the migration artifact store is not supported '
+      'for client migrations. The SQL is only available at runtime.',
+    );
```
nielsenko: There are a number of these `UnsupportedError` thrown, which smells like the store abstraction is slightly off. Would it be an idea to split in multiple interfaces (like read vs write?).

```dart
abstract class MigrationArtifactReader { 
  Future<...> readVersionSql(...); 
  //... 
}
abstract class MigrationArtifactStore extends MigrationArtifactReader {
    Future<void> writeVersion(...);  // only on the writable subtype
  }
}
```

nielsenko (changes requested):

packages/serverpod_database/lib/src/migrations/migration_manager.dart
marcelomendoncasoares: I did this because it's expected that the table does not exist when running the migration manager for a new database, but you're right that bubbling up is more symmetric with the `ServerMigrationManager` implementation. Done at https://github.com/serverpod/serverpod/pull/4994/commits/0b1ed747b6cea4ca4578a447e07ef65dd5473689.

tools/serverpod_cli/lib/src/migrations/generator.dart
marcelomendoncasoares: Yes, it does, and with no formatting to restrict chars or whatever. Don't know if this was intentional (spaces maybe, chars that could break a path certainly not), but this functions is been like this for 2 years.

image

tools/serverpod_cli/lib/src/migrations/client_side/client_migration_dart_emitter.dart
marcelomendoncasoares: @coderabbitai Please create an issue to sanitize the `tag` when creating a migration name to avoid chars that could break a path.

packages/serverpod_database/lib/src/migrations/migration_manager.dart
marcelomendoncasoares: Not exactly, because the `artifactStore` is always `RuntimeListMigrationArtifactStore` for the `ClientMigrationManager`:

```dart
/// Handles migrations for client-side databases.
class ClientMigrationManager extends MigrationManager {
  /// Creates a manager for the given in-memory [migrations] and [moduleName].
  ClientMigrationManager({
    required super.runMode,
    required List<MigrationVersionSql> migrations,
    required String moduleName,
  }) : super(
         RuntimeListMigrationArtifactStore(
           migrations,
           moduleName: moduleName,
         ),
       );
```

So we can safely cast it on the `_loadLatestDefinitionModuleName` to use the `moduleName` property.

tools/serverpod_cli/lib/src/migrations/generator.dart
nielsenko: Tracked in https://github.com/serverpod/serverpod/issues/5029

tools/serverpod_cli/lib/src/migrations/client_side/client_migration_artifact_store.dart
marcelomendoncasoares: Good idea! I did split the artifacts and methods, but should have gone one level deeper to split the interfaces as well. Done at https://github.com/serverpod/serverpod/pull/4994/commits/9a6245a013d0aa75535545eb7d565d4976157d72.

packages/serverpod_database/lib/src/migrations/migration_manager.dart
marcelomendoncasoares: With https://github.com/serverpod/serverpod/commit/9a6245a013d0aa75535545eb7d565d4976157d72 it is no longer needed.

packages/serverpod_database/lib/src/migrations/migration_artifacts_store/runtime_list.dart
marcelomendoncasoares: Agree with this improvement, but, since we're tight on time, I'll suggest we leave for the next pass on migrations.

nielsenko: @coderabbit Create a follow-up issue for this

packages/serverpod_database/lib/src/migrations/migration_manager.dart
nielsenko: tracked in #5031

nielsenko (commented):
LGTM 🥇

nielsenko (approved):

PR #4998, opened 2026-04-27
refactor: Move the column rename from `TableMigration` to `ColumnMigration`

marcelomendoncasoares:
This PR fixes a misplacement of the column rename migration that was on the `TableMigration` instead of in the more suitable `ColumnMigration`. This fixes the inconsistency of having to compare the tables by mocking the name to avoid a rename diff.

## Pre-launch Checklist

## Breaking changes

This PR does change a public interface, but it is not a breaking change since the column rename was only released on a beta version.

### DIFF
--- a/tools/serverpod_cli/lib/src/database/dialects/sqlite.dart
+++ b/tools/serverpod_cli/lib/src/database/dialects/sqlite.dart
@@ -331,7 +331,7 @@ extension SqliteTableMigrationSqlGeneration on TableMigration {
     );
 
     final needsRebuild =
-        modifyColumns.isNotEmpty ||
+        modifyColumns.hasColumnMigrationThatRequiresRebuild ||
         deleteForeignKeys.isNotEmpty ||
         addForeignKeys.isNotEmpty ||
         addColumns.any((c) => c.addColumnNeedsRebuild(targetTable));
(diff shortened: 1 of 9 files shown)


tools/serverpod_cli/lib/src/database/dialects/sqlite.dart
```diff
@@ -523,19 +542,16 @@ extension on ColumnType {
...
+extension on List<ColumnMigration> {
+  /// Whether any of the column migrations require a table rebuild. The only
+  /// operation that does not require a table rebuild is a column rename.
+  bool get hasColumnMigrationThatRequiresRebuild => any(
+    (m) =>
+        m.addNullable ||
+        m.removeNullable ||
+        m.changeDefault ||
+        m.newType != null,
+  );
```
nielsenko, line 554: maybe a `m.pureRename`?

nielsenko (approved):
LGTM.. a small suggestion - not critical

tools/serverpod_cli/lib/src/database/dialects/sqlite.dart
marcelomendoncasoares: Since it is only used once, I didn't think the abstraction was really necessary. And then reading the conditions is close to when it matters to branch on it.

PR #5001, opened 2026-04-27
perf: Replace Protocol.deserialize<T> if-chain with static Map dispatch

### LINKED ISSUE #5000: AOT compile time grows non-linearly with model count due to monolithic Protocol.deserialize<T> if-chain
## Problem

`gen_snapshot` (the AOT step of `dart compile exe`) takes 30–40 minutes to compile a Serverpod server with a few hundred models. A clean Serverpod project compiles in ~12 seconds, so the cost is proportional to project size — but non-linearly.

On a 291-model project (~3,250 deserializer entries), `gen_snapshot --print_precompiler_timings` shows:

```
[100.00%] CompileAll                         1334.67 s
  [ 99.98%] CompileFunction                  1328.68 s
    [ 56.56%] TypePropagation                  751.47 s
    [ 19.31%] AllocateRegisters                256.52 s
    [ 11.97%] OptimizeBranches                 159.06 s
    [ 11.72%] ConstantPropagation              155.68 s
    [  0.24%] Inlining                           3.16 s
```

Almost the entire compile is spent in **per-function** optimization passes on a single method.

## Root cause

The generated `Protocol.deserialize<T>` is one method whose body is an if-chain over every registered type:

```dart
T deserialize<T>(dynamic data, [Type? t]) {
  t ??= T;
  // ... prologue ...
  if (t == _i5.Foo)  { return _i5.Foo.fromJson(data) as T; }
  if (t == _i6.Bar)  { return _i6.Bar.fromJson(data) as T; }
  // ... 3,251 more branches ...
  return super.deserialize<T>(data, t);
}
```

`gen_snapshot`'s per-function passes (`TypePropagation`, `AllocateRegisters`, `OptimizeBranches`, `ConstantPropagation`) iterate over the function's CFG until a fixed point. Each `if (t == X)` refines the type lattice for `t` on the true-branch and that refinement has to be propagated through all subsequent blocks. The cost grows roughly quadratically with branch count.

This affects **server-side AOT** (`dart compile exe bin/main.dart`) and **client-side desktop AOT** (Flutter macOS/Linux/Windows) equally. It does not affect dart2js / Flutter web.

## Empirical proof

Stubbing the bodies of the five dispatch methods in `Protocol` (`deserialize`, `deserializeByClassName`, `getClassNameForObject`, `getClassNameForType`, `getTableForType`) with `throw UnimplementedError()` drops `gen_snapshot` from **1,334 s → 3.5 s** on the same project. So the cost is fully attributable to those methods, not to anything else in the model graph.

Replacing the if-chain in `deserialize<T>` alone with a static `Map<Type, ...>` lookup (where each branch becomes its own tiny closure — moving the work out of one giant function and into many trivial ones) drops gen_snapshot to **~12 s on a minimal probe** and **~35 s on a full server** entrypoint that previously took 30–40 min.

## Fix

Replace the if-chain with a static `Map<Type, dynamic Function(dynamic, Protocol)> _deserializers`, populated by a `_buildDeserializers()` static method (function-body population avoids `equal_keys_in_map` analyzer warnings caused by the same Dart class being imported under multiple module aliases — harmless at runtime since duplicates have identical bodies).

`deserialize<T>` becomes:

```dart
T deserialize<T>(dynamic data, [Type? t]) {
  t ??= T;
  // ... prologue ...
  final fn = _deserializers[t];
  if (fn != null) return fn(data, this) as T;
  // module fallbacks ...
  return super.deserialize<T>(data, t);
}
```

I have a working patch against `serverpod_cli` that does this for both server and client `Protocol` classes; happy to open a PR if there's interest. All existing CLI tests pass after a small update to two golden-tests that searched for the old `if (t == ...) { return ...; }` shape.

## Environment

Dart SDK: stable, AOT compile via `dart compile exe` / `flutter build`.
Serverpod: 3.5.0-beta.4 (also affects current main).

dipsy:
Replaces the `Protocol.deserialize<T>` if-chain with a static `Map<Type, dynamic Function(dynamic, Protocol)> _deserializers` lookup. Each former branch becomes its own closure, so dispatch is O(1) and the compiler no longer has to optimize one method holding thousands of `Type`-equality comparisons.

Fixes #5000.

## Why the if-chain is expensive to compile

The cost is not method size. In the same generated `protocol.dart`, compiled into the same snapshot:

| Method | Shape | Size | `gen_snapshot` cost |
|---|---|---:|---|
| `deserialize` | if-chain over `Type` constants | 4,342 branches | 2,122 s |
| `deserializeByClassName` | if-chain over `String` constants | 5,971 lines, ~1,850 branches | negligible |
| `getClassNameForType` | `switch` expression over `Type` | 1,854 arms | negligible |

A 5,971-line if-chain compiles fine; 4,342 `t == <Type constant>` comparisons do not. 73% of the time goes to `TypePropagation`.

A `switch` is not an alternative here: 1,907 of the 4,342 keys (43.9%) are `getType<X?>()` calls, and `case` patterns require constant expressions. Nullable types cannot be written as type literals in Dart, which is what that helper exists for. Converting only the const-expressible keys would leave ~1,907 comparisons in the chain.

## Measured impact

Apple M5 Max, Dart 3.12.2 (Flutter 3.44.6 SDK), a 4,367-entry project protocol. AOT built as `gen_kernel` + `gen_snapshot` so pass timings are visible; runtime measured with `dartaotruntime`.

### Compile

Minimal probe — Serverpod + Database + the project protocol, no endpoints:

| | Before | After |
|---|---:|---:|
| `gen_kernel` real | 15.34 s | 53.45 s |
| `gen_snapshot` real | 2,122.67 s | 12.61 s |
| └ `TypePropagation` | 1,551.09 s | 0.53 s |
| └ `ConstantPropagation` | 277.84 s | 0.24 s |
| └ `OptimizeBranches` | 274.95 s | 0.21 s |
| **total AOT** | **2,138.0 s** | **66.1 s** |

`gen_kernel` gets slower — 4,387 closures are more work for the front end — and that is included in the total above.

### Runtime

Enum models deserialized at spread positions of the chain; enums keep `fromJson` trivial so the measurement reflects dispatch. 300k iterations per case after 20k warmup.

| Position in chain | Before | After |
|---:|---:|---:|
| 1 | 17.4 ns | 18.1 ns |
| 464 | 976.2 ns | 17.8 ns |
| 918 | 1,883.7 ns | 17.7 ns |
| 1,389 | 2,974.9 ns | 17.0 ns |
| 1,832 | 4,943.4 ns | 17.6 ns |
| 4,108 (`List<T>` variant) | 39,887.7 ns | 73.2 ns |
| round-robin across all of the above | 8,704.8 ns | 29.7 ns |

Dispatch becomes flat in position. The chain is faster only for the type that happens to sit first, by 0.7 ns.

For models with many fields the relative gain is smaller, since `fromJson` dominates; the absolute saving per call is unchanged.

### Memory and startup

| | Before | After |
|---|---:|---:|
| first `deserialize` call | 13 µs | 2,617 µs |
| RSS after first call | 29.8 MB | 33.2 MB |
| snapshot `Total(HeapSize)` | 11.28 MB | 12.99 MB |
| snapshot `Total(CodeSize)` | 13.25 MB | 13.61 MB |

The map is built lazily on the first `deserialize`, not at process start, and pays for itself after roughly 300 calls.

## Implementation notes

* Map population uses a function body (`map[K] = (data, protocol) => ...;`) rather than a map literal so that duplicate `Type` keys — caused by the same Dart class being imported under multiple module aliases — do not trigger the `equal_keys_in_map` analyzer warning. Last-write-wins is harmless because duplicates have identical bodies.
* Closures have no access to `this`, so everything a body used to reach through the enclosing instance method is routed through the closure's `protocol` parameter: recursive `deserialize<X>` calls inside container types (`List<X>`, `Map<K,V>`, `Set<X>`, nullable variants, records), and `protocol.deserializeDynamicFieldValue(data)` for `dynamic` fields. The single `as T` cast is applied once at the call site in `deserialize<T>` rather than in each body.
* `deserialize<T>` is now a lookup followed by the same module-fallback chain, so module composition semantics are preserved.

## Pre-launch Checklist

## Breaking changes

No semantic breaking changes. The generated `protocol.dart` shape changes — the if-chain in `Protocol.deserialize<T>` becomes a static `_deserializers` map with a `_buildDeserializers()` method — but this is internal generated code, not a stable API.

CLI test updates come with it:

* One `protocol_test.dart` case matched a regex against the `if (t == ... List<Person>?...)` shape; it now matches `map[...] = ...`.
* `inheritance_protocol_test.dart` and `inheritance_exception_protocol_test.dart` (server / client / shared each) inspected `deserializeMethod.toSource()` for `X.fromJson` to verify top-node alias resolution. Those calls now live in `_buildDeserializers`, so the tests target the protocol class instead — the same check, independent of which method holds the call, and it makes the negative assertion about sealed parents meaningful rather than vacuous.

### DIFF
--- a/examples/auth/auth_client/lib/src/protocol/protocol.dart
+++ b/examples/auth/auth_client/lib/src/protocol/protocol.dart
@@ -27,6 +27,24 @@ class Protocol extends _i1.SerializationManager {
 
   static final Protocol _instance = Protocol._().._registerHostProtocols();
 
+  static final Map<Type, dynamic Function(dynamic, Protocol)> _deserializers =
+      _buildDeserializers();
+
+  static Map<Type, dynamic Function(dynamic, Protocol)> _buildDeserializers() {
+    final map = <Type, dynamic Function(dynamic, Protocol)>{};
+    map[_i2.Greeting] = (data, protocol) => _i2.Greeting.fromJson(data);
+    map[_i1.getType<_i2.Greeting?>()] = (data, protocol) =>
+        (data != null ? _i2.Greeting.fromJson(data) : null);
+    map[_i1.getType<({_i3.ByteData challenge, _i1.UuidValue id})>()] =
+        (data, protocol) => (
+          challenge: protocol.deserialize<_i3.ByteData>(
+            ((data as Map)['n'] as Map)['challenge'],
+          ),
+          id: protocol.deserialize<_i1.UuidValue>(data['n']['id']),
+        );
+    return map;
+  }
+
   static String? getClassNameFromObjectJson(dynamic data) {
     if (data is! Map) return null;
     final className = data['__className__'] as String?;
@@ -54,20 +72,9 @@ class Protocol extends _i1.SerializationManager {
       }
     }
 
-    if (t == _i2.Greeting) {
-      return _i2.Greeting.fromJson(data) as T;
-    }
-    if (t == _i1.getType<_i2.Greeting?>()) {
-      return (data != null ? _i2.Greeting.fromJson(data) : null) as T;
-    }
-    if (t == _i1.getType<({_i3.ByteData challenge, _i1.UuidValue id})>()) {
-      return (
-            challenge: deserialize<_i3.ByteData>(
-              ((data as Map)['n'] as Map)['challenge'],
-            ),
-            id: deserialize<_i1.UuidValue>(data['n']['id']),
-          )
-          as T;
+    final fn = _deserializers[t];
+    if (fn != null) {
+      return fn(data, this) as T;
     }
     try {
       return _i4.Protocol().deserialize<T>(data, t);
--- a/examples/auth/auth_client/lib/src/protocol/protocol.dart
+++ b/examples/auth/auth_client/lib/src/protocol/protocol.dart
@@ -12,12 +12,12 @@
 
 // ignore_for_file: no_leading_underscores_for_library_prefixes
 import 'package:serverpod_client/serverpod_client.dart' as _i1;
-import 'greeting.dart' as _i2;
-import 'dart:typed_data' as _i3;
 import 'package:serverpod_auth_idp_client/serverpod_auth_idp_client.dart'
-    as _i4;
+    as _i2;
 import 'package:serverpod_auth_core_client/serverpod_auth_core_client.dart'
-    as _i5;
+    as _i3;
+import 'greeting.dart' as _i4;
+import 'dart:typed_data' as _i5;
 export 'greeting.dart';
 export 'client.dart';
 
@@ -27,6 +27,9 @@ class Protocol extends _i1.SerializationManager {
   factory Protocol() => _instance;
 
   static final Protocol _instance = Protocol._().._registerHostProtocols();
+
+  static final Map<Type, dynamic Function(dynamic, Protocol)> _deserializers =
+      _buildDeserializers();
 
   static String? getClassNameFromObjectJson(dynamic data) {
     if (data is! Map) return null;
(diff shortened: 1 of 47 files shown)


dipsy:
***I have read the CLA Document and I hereby sign the CLA***

dipsy:
recheck

dipsy:
Rebased onto current `main` (4.0.0-beta.2) — the branch is mergeable again and the description's numbers are unchanged.

One conflict, in `library_generator.dart`: `targetTableDefinitions` moved from the fields list into the methods list upstream. The deserializer map is now emitted after that method rather than before it, which keeps the import prefix allocation identical to `main` — otherwise every `_iN` reference in all 31 generated protocols shifts and the diff balloons for no semantic reason.

Three files joined the diff: the generated protocols of `serverpod_test_shared_module`, a test package added after this PR was opened.

Verification on the rebased tree: `util/generate_all` regenerates clean across all 15 server packages and the templates, `dart format` is clean, and the `serverpod_cli` suite passes (6346 tests, 2 skipped).

PR #5007, opened 2026-04-28
test: Avoid withServerpod doing work on test declare time

nielsenko:
This PR reworks `withServerpod` so that it defers setup. It is important in dart test to defer setup so that it doesn't happen during the _declare_ phase. This allows quick test discovery in the IDE, and select by pattern to work fast.

### DIFF
--- a/packages/serverpod_test/lib/src/with_serverpod.dart
+++ b/packages/serverpod_test/lib/src/with_serverpod.dart
@@ -72,9 +72,13 @@ typedef TestClosure<T> =
 /// The default integration test tag used by `withServerpod`.
 const String defaultIntegrationTestTag = 'integration';
 
-/// Builds the `withServerpod` test helper.
-/// Used by the generated code.
-/// Note: The [testGroupName] parameter is needed to enable IDE support.
+/// Builds the `withServerpod` test helper. Used by generated code.
+///
+/// Registers a `group` with `setUpAll`, `tearDown`, and `tearDownAll`
+/// hooks and nothing else. Session, transaction manager, endpoint
+/// wrappers, and Serverpod construction are all deferred into
+/// `setUpAll`, so test-tree enumeration stays cheap and sendable
+/// across isolates.
 void Function(TestClosure<T>)
 buildWithServerpod<T extends InternalTestEndpoints>(
   String testGroupName,
@@ -95,41 +99,7 @@ buildWithServerpod<T extends InternalTestEndpoints>(
   }
 
   var startTimeout = maybeServerpodStartTimeout ?? const Duration(seconds: 30);
-
-  var mainServerpodSession = testServerpod.createSession(
-    rollbackDatabase: rollbackDatabase,
-  );
-
-  TransactionManager? transactionManager;
-  if (testServerpod.isDatabaseEnabled) {
-    transactionManager = mainServerpodSession.transactionManager;
-    if (transactionManager == null) {
-      throw InitializationException(
-        'The transaction manager is null but database is enabled.',
-      );
-    }
-  }
-
-  TransactionManager getTransactionManager() {
-    var localTransactionManager = transactionManager;
-    if (localTransactionManager == null) {
-      throw StateError(
-        'The transaction manager is null.',
-      );
-    }
-
-    return localTransactionManager;
-  }
-
-  List<InternalServerpodSession> allTestSessions = [];
-
-  InternalTestSessionBuilder mainTestSessionBuilder =
-      InternalTestSessionBuilder(
-        testServerpod,
-        allTestSessions: allTestSessions,
-        enableLogging: maybeEnableSessionLogging ?? false,
-        mainServerpodSession: mainServerpodSession,
-      );
+  var enableLogging = maybeEnableSessionLogging ?? false;
 
   bool startServerpodFailed = false;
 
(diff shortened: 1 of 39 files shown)


packages/serverpod_test/lib/src/with_serverpod.dart
```diff
@@ -95,114 +99,133 @@ buildWithServerpod<T extends InternalTestEndpoints>(
...
-            await testServerpod.start().timeout(
-              startTimeout,
-              onTimeout: () {
-                throw InitializationException(
-                  'Serverpod did not start within the timeout of $startTimeout. '
-                  'This might indicate that Serverpod cannot connect to the database. '
-                  'Ensure that you have run `docker compose up` and check the logs for more information.',
-                );
-              },
+        final sessionBuilder = InternalTestSessionBuilder();
```
marcelomendoncasoares, line 114: Question: Shouldn't this be `late` as well to defer to when accessed?

tests/serverpod_auth_test/serverpod_auth_test_server/test/integration/legacy_client_support_test.dart
```diff
@@ -23,673 +24,644 @@ void main() {
...
-          identityProviderBuilders: [config],
-          tokenManagerBuilders: [
-            tokenManagerConfig,
-            const LegacySessionTokenManager(),
-          ],
+      setUpAll(() => _configurePublicLegacySupport(sessionBuilder));
+
+      group('Given a migrated legacy user,', () {
+        const email = 'legacy-client-test@serverpod.dev';
+        const password = 'LegacyPass123!';
```
marcelomendoncasoares, line 33: Suggestion: This is now an incorrect nesting of two "Given" descriptions.

tests/serverpod_test_server/test_integration/test_tools/rollback_database_test.dart
```diff
@@ -151,9 +156,12 @@ void main() {
...
-        var session = sessionBuilder.build();
-        var newSessionBuilder = sessionBuilder.copyWith();
-        var newSession = newSessionBuilder.build();
+        late Session session;
+        late Session newSession;
+        setUp(() {
+          session = sessionBuilder.build();
+          newSession = sessionBuilder.copyWith().build();
+        });
         setUp(() async {
```
marcelomendoncasoares: Question: Why two `setUp` groups?

tests/serverpod_test_server/test_integration/test_tools/rollback_database_test.dart
```diff
@@ -253,7 +263,8 @@ void main() {
       withServerpod(
         '',
         (sessionBuilder, endpoints) {
-          var session = sessionBuilder.build();
+          late Session session;
+          setUp(() => session = sessionBuilder.build());
           setUp(() async {
```
marcelomendoncasoares, line 263: Same here.

tests/serverpod_test_sqlite/serverpod_test_sqlite_server/test_integration/test_tools/database_operations_test.dart
```diff
@@ -326,7 +333,8 @@ void main() {
       withServerpod(
         '',
         (sessionBuilder, endpoints) {
-          var session = sessionBuilder.build();
+          late Session session;
+          setUp(() => session = sessionBuilder.build());
           setUpAll(() async {
```
marcelomendoncasoares, line 334: Question: Shouldn't these cases be created inside a `setUpAll` instead of `setUp` to match the previous behavior?

marcelomendoncasoares (changes requested):
Aside from some small comments on the tests' replacements, my only concern with this change is that it is massively breaking for users. The fix is indeed important, but all tests using `withServerpod` on users' projects will break with a considerable cost to fix. Can we make this change in a configurable way so that users can migrate intentionally?

tests/serverpod_test_server/test_integration/test_tools/rollback_database_test.dart
nielsenko: 😯 .. will fix

packages/serverpod_test/lib/src/with_serverpod.dart
nielsenko: No. The `late` is inside `InternalTestSessionBuilder`. We need to pass it in the callback during declaration. At least if we want to keep the current interface.

tests/serverpod_auth_test/serverpod_auth_test_server/test/integration/legacy_client_support_test.dart
nielsenko: Given Given When Then - what is not to like.  I will fix

nielsenko:
Converted to draft. Despite fixing a bug this will be too big a change for users for 3.5 release. We have discussed a way to land this as opt-in, but it will be deferred to after Jetstream.

packages/serverpod_test/lib/src/with_serverpod.dart
```diff
@@ -95,114 +99,133 @@ buildWithServerpod<T extends InternalTestEndpoints>(
...
+          await testServerpod.start().timeout(
+            startTimeout,
+            onTimeout: () {
+              throw InitializationException(
+                'Serverpod did not start within the timeout of $startTimeout. '
+                'This might indicate that Serverpod cannot connect to the database. '
+                'Ensure that you have run `docker compose up` and check the logs for more information.',
+              );
+            },
+          );
```
marcelomendoncasoares, line 130: Comment: Another thing to have in mind with this work is that, if you want to add the `authenticationHandler` to the `Serverpod` instance that will be started by `withServerpod`, it is only possible if getting the instance from the `sessionBuilder.build().serverpod` before entering the `setUpAll`. Otherwise the `authenticationHandler` is set too late and is not picked up by the `serverpod.server`.

nielsenko:
Re-opening after rebasing on main (4.0-beta.1 soon .2)

PR #5008, opened 2026-04-28
feat: Ensure serverpod start works with build hook deps

nielsenko:
Adds support for `package:native_assets` build hooks to the Serverpod CLI so that server packages depending on FFI libraries with `hook/build.dart` scripts (fx `package:sqlite3`) actually build and load their native assets when running `serverpod start`.

`dart compile` refuses build hooks and `frontend_server` does not know about them, so we drive `package:hooks_runner` ourselves and feed the resulting `native_assets.yaml` to `frontend_server` via the existing `--native-assets` flag. 

`--no-fes` mode is unaffected - it already goes through `dart run`/dartdev which handles hooks natively.

Highlights:
- New `NativeAssetsBuilder` that wraps `hooks_runner`, bundles assets, and writes `.dart_tool/serverpod/native_assets.yaml` (only when content changes).
- `WatchSession` reruns hooks per reload cycle and restarts the frontend server when the manifest changes; migrations and IDE `reloadSources` callbacks rerun hooks too.
- `KernelCompiler` / `FrontendServerClient` plumbed with a mutable `nativeAssetsPath`.
- Workspace-aware package_config discovery: walks up from the server dir to the first pubspec with `resolution != 'workspace'`.
- Vendored `bundleNativeAssets`, `writeNativeAssetsYaml`, and the macOS install-name rewriter from dartdev (BSD-attributed) since they are not exported by `package:hooks_runner` and dartdev itself is unpublishable.
- New deps: `hooks_runner`, `hooks`, `code_assets`, `data_assets`, `file`, `logging`; `pubspec_parse` bumped for `PubSpec.resolution`.

Fixes: #?

## Pre-launch Checklist

## Breaking changes

None. `--native-assets` plumbing in `KernelCompiler`/`FrontendServerClient` is opt-in via an optional parameter, and behavior is unchanged when no server package declares build hooks.

### DIFF
--- a/tools/serverpod_cli/test/commands/start/native_assets_builder_test.dart
+++ b/tools/serverpod_cli/test/commands/start/native_assets_builder_test.dart
@@ -88,11 +88,31 @@ void main() {
     });
 
     test(
-      'when build is called from the member dir, '
-      'then path discovery walks up to the workspace root',
+      'when discoverPaths is called from the member dir, '
+      'then it returns the workspace root pubspec and package_config',
       () async {
-        final outcome = await builder.build();
-        expect(outcome, isA<NativeAssetsBuildSkipped>());
+        final paths = await builder.discoverPaths();
+
+        expect(
+          p.equals(
+            paths.workspacePubspecPath,
+            p.join(tempDir.path, 'pubspec.yaml'),
+          ),
+          isTrue,
+          reason:
+              'workspacePubspecPath should be the root pubspec, not the '
+              'member pubspec. Got: ${paths.workspacePubspecPath}',
+        );
+        expect(
+          p.equals(
+            paths.packageConfigPath,
+            p.join(tempDir.path, '.dart_tool', 'package_config.json'),
+          ),
+          isTrue,
+          reason:
+              'packageConfigPath should be at the workspace root .dart_tool/. '
+              'Got: ${paths.packageConfigPath}',
+        );
       },
       timeout: const Timeout(Duration(seconds: 60)),
     );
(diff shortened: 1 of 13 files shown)


tools/serverpod_cli/test/commands/start/native_assets_builder_test.dart
```diff
@@ -0,0 +1,219 @@
...
+
+    test(
+      'when build is called from the member dir, '
+      'then path discovery walks up to the workspace root',
+      () async {
+        final outcome = await builder.build();
+        expect(outcome, isA<NativeAssetsBuildSkipped>());
+      },
+      timeout: const Timeout(Duration(seconds: 60)),
+    );
```
marcelomendoncasoares, line 104: Question: Shouldn't the outcome be different? The test description also seem different than what is being asserted.

marcelomendoncasoares (changes requested):
Just one question on one test that validate the fix.

tools/serverpod_cli/test/commands/start/native_assets_builder_test.dart
nielsenko: The workspace member dir has no `.dart_tool/`, only the root does. If `_discoverPaths()` stopped at the member, `_loadPackageLayout()` would throw a `StateError`. So not throwing implies discovery walked to the root. 😄 

I'll make `_discoverPaths()` visible for testing, and rework.

nielsenko: Fixed in [db523fc](https://github.com/serverpod/serverpod/pull/5008/commits/db523fc6f8ffe9a681612627c90b52b04aacc524)

nielsenko: Added extra tests in a452a9b

nielsenko: Thinking more about this I'll get rid of `_ResolvedPaths`

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5011, opened 2026-04-28
fix: Remove the requirement of `SERVERPOD_HOME` for CLI commands that doesn't touch it

marcelomendoncasoares:
This PR prevents the CLI from throwing when running `generate` or `create-migration` with a locally installed CLI - commands that don't use the `serverpodHome`. This is a small QoL improvement for testing on projects outside of the Serverpod monorepo while using the locally installed CLI.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/shared/environment.dart
+++ b/tools/serverpod_cli/lib/src/shared/environment.dart
@@ -1,22 +1,47 @@
 import 'dart:io';
 
+import 'package:cli_tools/cli_tools.dart';
 import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart';
 
 import '../generated/version.dart';
 
-String serverpodHome = '';
+String get serverpodHome {
+  if (_serverpodHome == null) {
+    loadEnvironmentVars();
+  }
+  return _serverpodHome!;
+}
 
+String? _serverpodHome;
+
+/// Loads the environment variables for the serverpod CLI.
+///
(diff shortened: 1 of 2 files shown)


tools/serverpod_cli/lib/src/shared/environment.dart
```diff
@@ -1,22 +1,47 @@
...
+  return _serverpodHome!;
+}
+
+String? _serverpodHome;
 
+/// Loads the environment variables for the serverpod CLI.
+///
+/// If in dev mode and the SERVERPOD_HOME environment variable is not set,
+/// it will throw an ExitException.error().
 bool loadEnvironmentVars() {
```
nielsenko: Can this ever return `false`?
```suggestion
void loadEnvironmentVars() {
```

nielsenko (changes requested):
Now we throw instead of returning false we should update signature to match

tools/serverpod_cli/lib/src/shared/environment.dart
marcelomendoncasoares: Not anymore, the return type should indeed be void. Fixed at https://github.com/serverpod/serverpod/pull/5011/commits/5a022b81c5d9deb7d39b628a31604f02a2022ba1.

nielsenko (approved):
LGTM 💯

PR #5012, opened 2026-04-28
fix: Fix Postgres throwing when using row-lock with includes

### LINKED ISSUE #5005: Bug: `findById` with `lockMode` and `include` fails when includes generate outer joins
### Summary

When using `findById` with a `lockMode` together with `include`, Serverpod generates a query where `FOR UPDATE` is applied in a way that also affects joined tables.

If the include contains nullable relations / outer joins, PostgreSQL throws the following error:

```text
databaseExceptionMessage: FOR UPDATE cannot be applied to the nullable side of an outer join

### To reproduce

```dart
final company = await Company.db.findById(
  session,
  X,
  transaction: transaction,
  lockMode: LockMode.forUpdate,
  lockBehavior: LockBehavior.wait,
  include: someIncudes,
);
```
This fails because it adds relations that are joined via outer joins.

### Expected behavior

When lockMode is passed to findById, Serverpod should only lock the primary row that is being queried by ID.

In this example, only the Company row should be locked.

The included relations should still be joined and returned, but they should not be part of the FOR UPDATE lock target.

Actual behavior

The query fails with:
`FOR UPDATE cannot be applied to the nullable side of an outer join`

This makes it impossible to use findById with both:

- `lockMode: LockMode.forUpdate`
- `include: ...`

when the include contains nullable relations.

### Why this matters

A common use case is:

- Lock the main row for update.
- Load the full entity including relations.
- Perform business logic safely within the same transaction.

Currently, this is not possible in a single findById call when includes are involved.

### Expected behavior

Serverpod should generate SQL that applies the lock only to the primary table / primary row.

For PostgreSQL, this could likely be done by using a table-specific lock target, for example conceptually:

`FOR UPDATE OF company`

instead of applying FOR UPDATE to the full result including all joined tables.

### Expected benefit

- findById works with lockMode and include
- The intended row is locked safely
- Relations are still loaded correctly
- No need for a two-query workaround
- More intuitive behavior for transactional entity updates

### Library version

3.4.7

### Platform information

All Platforms

### Additional context

_No response_

- [ ] I want to work on this

marcelomendoncasoares:
Fixes: #5005

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod_database/lib/src/adapters/postgres/sql_query_builder.dart
+++ b/packages/serverpod_database/lib/src/adapters/postgres/sql_query_builder.dart
@@ -106,7 +106,7 @@ class SelectQueryBuilder {
     if (groupBy != null) query += ' GROUP BY $groupBy';
     if (_having != null) query += ' HAVING $_having';
     if (orderBy != null) query += ' ORDER BY $orderBy';
-    var lockClause = _buildLockClause();
+    var lockClause = _buildLockClause(restrictLockToBaseTable: join != null);
     if (lockClause != null) query += ' $lockClause';
     if (limit != null) query += ' LIMIT $limit';
     if (offset != null) query += ' OFFSET $_offset';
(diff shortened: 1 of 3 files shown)


nielsenko (approved):
LGTM.. perhaps testing all versions of find is not necessary here (not really the important part)

PR #5013, opened 2026-04-28
feat: Add support for `dynamic` fields on models and endpoints

### LINKED ISSUE #3231: Map<String, dynamic> as an Endpoint Parameter
### Problem to Solve

Currently, Serverpod restricts Map parameters in endpoint methods to strictly typed maps (e.g., Map<String, int?>, Map<String, String?>, etc.). This limitation prevents the direct use of dynamic JSON objects—such as those produced by jsonDecode()—which makes it challenging to handle cases where payloads may vary or be invalid. As a result, when a client sends an incorrect payload, the system returns a generic Serverpod log error rather than a tailored response that could improve error handling and client feedback.

### Proposal

Add support for passing a Map<String, dynamic> as an endpoint parameter.

### Use Case

If a client sends a payload with an unexpected structure or incorrect data types, I currently only receive a generic Serverpod log error. With the ability to pass Map<String, dynamic>, I could:

Directly pass JSON-decoded maps to my endpoints.
Validate the payload within the endpoint.
Return custom, meaningful error messages to the client, improving the overall API experience.

### Alternatives

No

### Additional context

_No response_

- [ ] I want to work on this

### LINKED ISSUE #4996: Asymmetric handling of Map<String, dynamic>: serialization works, deserialization throws
### Summary

Serverpod's stance is "`dynamic` is not supported as a type" (per #2722, #2600). That rule is **enforced inconsistently**: `Map<String, dynamic>` round-trips partially through the serialization layer, then throws on the deserialize side. Either the rule should be enforced uniformly at codegen time, or the runtime should handle the case it already half-supports.

### Reproduction

Endpoint signature:

```dart
Future<Map<String, dynamic>> foo(Session session) async => {
  'nested': {'a': 1},  // any List or Map value
};
```

- **Serialization** (server → wire): succeeds. JSON-encodes fine.
- **Deserialization** (wire → server, e.g. as a parameter, or client-side decoding the response): throws `DeserializationTypeNotFoundException: No deserialization found for type dynamic` at `SerializationManager.deserialize` when a value is a nested `List`/`Map`.

The stack trace originates in generated `Protocol.deserialize`, called from `MapBase.map` over each entry of the `Map<String, dynamic>`, dispatching `deserialize<dynamic>(value)` — which has no matching branch.

### Why this is a bug, not a feature request

The cited maintainer guidance (#2722, #2600) is sound for the **wire-format / type-preservation** argument. But the current implementation:

1. Lets `Map<String, dynamic>` pass CLI codegen without error.
2. Generates `fromJson` code that calls `deserialize<dynamic>(...)`.
3. Throws at runtime when that code path is exercised on nested containers.

That's three inconsistent layers. Pick one:

- **Strict**: reject `dynamic` at the CLI / spy.yaml validator. Fail fast with a clear error pointing users to typed models. (Aligns with the stated design.)
- **Permissive**: handle `t == dynamic` in `SerializationManager.deserialize` by returning already-decoded `List`/`Map` containers as-is. (What #4874 did.)

The current state — accept it everywhere except at the one runtime point where it matters — is the worst of both.

### Use cases that the typed-model workaround doesn't cover

The standard advice ("define a typed class") works when **you** own the schema. It doesn't cover:

- **User-defined / schema-on-read content** (CMS document bodies, headless content APIs). The shape is defined by end users at runtime, not by the Serverpod developer.
- **Webhook payloads** from third parties whose schemas evolve outside your release cycle.
- **Passthrough / gateway endpoints** that re-emit external JSON without needing to model it.

For these, the only escape hatches are (a) `String` + manual `jsonEncode`/`jsonDecode` at every callsite, or (b) hand-rolled custom serializers. Both defeat the point of having a typed RPC framework.

### Real-world context

We hit this building [`dart_desk_be`](https://github.com/ThangVuNguyenViet/dart_desk_be), an open-source Serverpod-based headless CMS backend. The `Document.data` field is user-defined JSON (schema controlled by end users, not us). We tried every supported route:

- `Map<String, dynamic>?` → blocked by this issue (CLI accepts; deserialize throws).
- Forking Serverpod to add the `dynamic` branch → works, but the consuming Flutter library is published to pub.dev, and `dependency_overrides` aren't transitive — consumers would silently break.
- `String?` text column with manual `jsonEncode`/`jsonDecode` → works but loses any chance of typed jsonb queries.

We ended up adding a Postgres-side **generated jsonb column** (`data_jsonb`) derived from the text column, with hand-written `unsafeQuery` for containment lookups (commits [`4c75373`](https://github.com/ThangVuNguyenViet/dart_desk_be/commit/4c75373), [`03a4a71`](https://github.com/ThangVuNguyenViet/dart_desk_be/commit/03a4a71)). That works, but it's a workaround for a framework-level inconsistency. We documented the tradeoffs in the repo's `CLAUDE.md` for other contributors.

### Relationship to PR #4960 (jsonb as first-class storage)

#4960 makes `jsonb` storage available for `List<T>`, `Map<String, T>` (concrete `T`), and named classes — all cases where `T` is statically known. It does not relax the "must know the type" rule, nor should it. This issue is **orthogonal**: it's about the deserializer's behavior when the user has explicitly opted out of type info via `dynamic`.

### Proposal

**Option A** (preferred — keeps existing intent, fixes the inconsistency): handle `t == dynamic` in `SerializationManager.deserialize`:

- `data == null` → return null
- `data is List` or `data is Map` → return as-is (already decoded by `jsonDecode`)
- Otherwise → re-dispatch on runtime type

This is a non-breaking change (previously-throwing cases now succeed). PR #4874 implemented exactly this with tests; happy to revive it if this issue lands favorably.

**Option B** (stricter): reject `dynamic` at CLI codegen time. Breaking change for anyone currently relying on the partially-working serialization side, but internally consistent.

Either is better than the status quo.

### Environment

- Serverpod: 3.4.x (also reproducible on 2.x per linked discussions)
- Affects: any endpoint or model with `Map<String, dynamic>` fields where values are nested containers

### Related

- Discussions: #2722, #2600, #1048, #444
- Closed PR with implementation + tests: #4874

marcelomendoncasoares:
Implements the support for `dynamic` fields in models and endpoints.

Under the hood, it uses the same strategy that we apply for streaming endpoints to encode the value as a `Map<String, dynamic>`. with the type on the `className` key.

On the database, the value is stored in a `ColumnSerializable` or `ColumnStructured` column - according to the JSON/JSONB serialization data type.

Closes: #4996
Closes: #3231

Related discussions:

- #2722
- #2600

## Pre-launch Checklist

## Breaking changes

None.


packages/serverpod_serialization/lib/src/serialization.dart
```diff
@@ -57,6 +57,19 @@ abstract class SerializationManager {
     return deserializeByClassName(jsonDecode(data));
   }
 
+  /// Decodes a value for a `dynamic` model field from protocol or database
+  /// shapes: a JSON [String] from [encodeWithType], or a [Map] when JSON was
+  /// parsed before deserialization.
+  dynamic decodeDynamicFieldValue(Object? value) {
+    if (value == null) return null;
+    if (value is String) return decodeWithType(value);
```
nielsenko, line 65: ~what about `int` and `bool`?~

nielsenko (approved):
👍

tests/serverpod_test_server/test_integration/database_operations/dynamic_roundtrip_test.dart
```diff
@@ -0,0 +1,121 @@
...
+    payload: 42,
+    jsonbPayload: 10.23,
+    payloadList: [1, 'b', SimpleData(num: 7)],
+    payloadMap: {'a': 1, 'b': 2, 'c': SimpleData(num: 3)},
+    payloadSet: {1, 2, 3, 'd'},
+    payloadMapWithDynamicKeys: {'a': 1, 2: SimpleData(num: 1)},
+  );
+
+  /// Small abstraction to keep expectations close to the source model.
+  void _expectDynamicFields(ObjectWithDynamic retrieved) {
```
nielsenko: Defined inside another function. Marking it private with `_` makes no sense

tests/serverpod_test_sqlite/serverpod_test_sqlite_server/test_integration/database_operations/dynamic_roundtrip_test.dart
nielsenko, line 1: It would be nice if we could abstract these tests into contract tests (avoid the duplication between pg and sqlite).. not for now, but just saying

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
```diff
@@ -145,13 +145,6 @@ abstract class EndpointMethodAnalyzer {
       return null;
     }
 
-    if (innerType is DynamicType && !dartType.isDartAsyncStream) {
-      return SourceSpanSeverityException(
-        'Return generic must have a type defined. E.g. ${dartType.element.name}<String>.',
-        dartElement.span,
-      );
-    }
-
```
nielsenko, line 154: Should we log an info/warning? If a user forgets the generic argument they now implicitly get `dynamic`, ie. `Future` becomes `Future<dynamic>`.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -1332,14 +1332,14 @@ class Restrictions {
 
     var errors = <SourceSpanSeverityException>[];
 
-    if (_isUnsupportedType(fieldType)) {
+    if (typeText != null && typeText.contains('dynamic?')) {
```
nielsenko: A bit loose.. what if somebody uses `Mydynamic?` 😄

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -892,19 +892,30 @@ class SerializableModelLibraryGenerator {
...
-            );
+        if (field.type.className == 'dynamic') {
+          // Since `dynamic` also covers `_Undefined`, the check for `param`
+          // must be inverted to explicitly not be `_Undefined`.
+          valueDefinition = refer(field.name)
+              .isNotA(refer('_Undefined'))
+              .conditional(
+                refer(field.name),
+                assignment,
+              );
```
nielsenko: Isn't this impossible? The `types.dart` change forces `isNullable=false` for `dynamic`. After parsing, valid `dynamic` fields always have `field.type.nullable == false,? so they fall through to
  the else branch that emits `payload ?? this.payload`.

In particular: `copyWith(payload: null)` cannot actually overwrite a non-null dynamic field with `null`?! 

if I read this correctly.

nielsenko (changes requested):
A few nits.. and a potential problem

tests/serverpod_test_sqlite/serverpod_test_sqlite_server/test_integration/database_operations/dynamic_roundtrip_test.dart
marcelomendoncasoares: Agree... I started writing an issue about this, but wished to think a bit more on the problem to already suggest an alternative.

tests/serverpod_test_server/test_integration/database_operations/dynamic_roundtrip_test.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5013/commits/636d245b345769e52090ef32a097656cd71dc260.

tools/serverpod_cli/lib/src/analyzer/dart/endpoint_analyzers/endpoint_method_analyzer.dart
marcelomendoncasoares: At first I removed to avoid pushing more complexity (since we need to get the declaration for this), but you are right that the experience is better. Fixed at https://github.com/serverpod/serverpod/pull/5013/commits/43530252a46c909c58ce521a98ee7442a3cb2916.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
marcelomendoncasoares: Good catch! Fixed at https://github.com/serverpod/serverpod/pull/5013/commits/b4ef5d2fb0066dcf0f68be1b61e7a6e174bb85d7 (with tests).

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
marcelomendoncasoares: The `copyWith(payload: null)` is intended behavior, since the default (should be) `payload= _Undefined)` as we do for all nullable fields. It was not, and it is now fixed at https://github.com/serverpod/serverpod/pull/5013/commits/a90b1a21a58c613639cb63739a5742996cef8758 (together with the wrong placement of the condition).

nielsenko (approved):
LGTM 👍

PR #5022, opened 2026-04-30
feat: VmServiceProxy (for intercepting reloadSources)

### LINKED ISSUE #4969: Replace the IDE server launch command to use `serverpod start`
With #4509, new projects receive a `.vscode/launch.json` that adds the server, docker and web app (if there is one) for a seamless start on IDEs based on VS Code OSS by just hitting F5.

Now that we have `serverpod start`, we should replace the launch by a call to `serverpod start` on the templater as well - as done during development at https://github.com/serverpod/serverpod/pull/4788/changes/a2ea8aacae619a81485df5cd1a4ea432b7e57285, which was later reverted.

nielsenko:
This PR is an alternative to #dap. It achieves the same intercept on `reloadSources` that is needed to work-around https://github.com/dart-lang/sdk/issues/62822.

### Custom adapter wins on
- Hot restart UX (no reconnect blip)
- DAP-level extensibility (room to add custom requests later)

### Proxy wins on
- Code volume (~3-4× less)
- Editor portability (no per-editor adapter wiring)
- Failure-mode count (smaller transport surface, fewer races)
- Conceptual cleanliness (the FES owner is the reload responder; no relay through the pod)
- Long-term maintenance (no `package:dds/src/...` implementation import, no stdio fallback, no announcement protocol, no vendored sdk code)

##  Questions

1. Is hot restart seamless-ness worth the architectural cost? The custom adapter's only material advantage is keeping the DAP session alive across restart.
2. Is a `serverpod` debug adapter a feature or a tax?


templates/serverpod_templates/vscode/tasks.json
```diff
@@ -0,0 +1,51 @@
...
+    {
+      "label": "serverpod_start",
+      "type": "shell",
+      "command": "serverpod",
+      "args": [
+        "start",
+        // {{#database}}
+        "--",
+        "--apply-migrations"
+        // {{/database}}
```
marcelomendoncasoares: Question: Do we still need this after https://github.com/serverpod/serverpod/pull/5028?

marcelomendoncasoares (approved):
LGTM! 🚀 

Only one question regarding a related PR.

templates/serverpod_templates/vscode/tasks.json
nielsenko: Yes, but it is not needed after this PR (removed in 7d704ba)

PR #5023, opened 2026-04-30
feat: Interactive configuration for serverpod create

### LINKED ISSUE #1643: Interactive configuration during serverpod create
Often times the user would like to set a different configuration than the default one for a Serverpod project.
Things such as database user, hostname and port could be configurable through an interactive mode when creating the project.

Running Serverpod create (without a `-y`) starts an interactive mode where the user has the possibility to provide configuration for the project. The information is then stored in the `passwords` / `config` file of the project.

**Related discussions**
- https://github.com/serverpod/serverpod/discussions/521

Crazelu:
`serverpod create` command by default now renders a TUI component for interactive project configuration. Opt out is available with the `no-interactive` flag.

Interfaces have been extracted from the existing TUI feature for `serverpod start` and reused here to achieve a cohesive UX and feel. New TUI features in the Serverpod CLI should be recommended to use this approach.

Closes [1643](https://github.com/serverpod/serverpod/issues/1643)

## Pre-launch Checklist

## Breaking changes

None


tools/serverpod_cli/lib/src/create/create.dart
```diff
@@ -251,41 +289,43 @@ Future<bool> performCreate(
...
+
+    return (success: success, relativeServerPath: relativeServerPath);
   }
 
-  return success;
+  return (success: success, relativeServerPath: '');
 }
 
-Future<bool> _performUpgrade(
+Future<CreateResult> _performUpgrade(
```
marcelomendoncasoares: Suggestion: Instead of returning a record, we could simplify to just return `String?` and avoid having to pass `relativeServerPath: ''` on all failure cases.

tools/serverpod_cli/lib/src/create/create.dart
```diff
@@ -48,13 +50,49 @@ enum ServerpodTemplateType {
...
+  return success;
+}
+
+Future<CreateResult> performCreateWithResult(
+  String name,
+  ServerpodTemplateType template,
+  bool force, {
+  required bool? interactive,
+  required TemplateContext context,
+}) async {
```
marcelomendoncasoares: Suggestion: If we unify the return to be `String?` instead of `bool` or a `CreateResult`, we don't even need two methods anymore.

tools/serverpod_cli/lib/src/commands/create.dart
```diff
@@ -119,12 +133,24 @@ class CreateCommand extends ServerpodCommand<CreateOption> {
...
 
+    if (useTui && !template.isMini) {
+      await _performCreateWithTui(
+        name,
+        template,
+        force,
+        interactive: interactive,
+      );
+      return;
+    }
```
marcelomendoncasoares, line 171: Question: Should the template be exposed as options on the app? Feels like a partial experience that you have to chose the template before and only the other options within.

tools/serverpod_cli/lib/src/commands/create/tui/config.dart
```diff
@@ -0,0 +1,112 @@
...
+/// Configuration for [ServerpodCreateApp].
+/// The enum values are mapped to the configurable features
+/// for the `serverpod create` command, typically held by [TemplateContext].
+enum ServerpodCreateConfig<T extends ConfigOption> {
+  database<DatabaseConfigOption>(
+    label: 'Database',
+    options: DatabaseConfigOption.values,
+    defaultOption: DatabaseConfigOption.postgres,
+    templates: [ServerpodTemplateType.server, ServerpodTemplateType.module],
+  ),
```
marcelomendoncasoares, line 22: Suggestion: Still missing the `skills` option (on/off only, since we'll use the `.agents` strategy). But I'm ok with leaving it for a follow-up.

tools/serverpod_cli/lib/src/commands/create.dart
```diff
@@ -55,6 +61,13 @@ enum CreateOption<V> implements OptionDefinition<V> {
       mandatory: true,
     ),
   ),
+  tui(
+    FlagOption(
+      argName: 'tui',
+      defaultsTo: true,
+      helpText: 'Show interactive terminal UI.',
+    ),
+  )
```
marcelomendoncasoares: Suggestion: Should we call it `--interactive` (`-i`) instead? Feels like a more natural name and bound to what the user gets, instead of how it's implemented.

tools/serverpod_cli/test/commands/create/tui/state_test.dart
```diff
@@ -0,0 +1,388 @@
...
+          state.updateFocusedConfig(-1);
+          expect(state.focusedConfigIndex, 0);
+        },
+      );
+
+      test(
+        'when updating the focused config with negative delta, '
+        'and the current focused config index is 0,'
+        'then the focused config index wraps to the max config index',
+        () {
```
marcelomendoncasoares, line 204: Suggestion: Describe these tests from the behavior perspective instead of from the implementation (i.e. instead of mentioning a positive/negative delta, refer navigating up/down), Valid for all tests in this file.

tools/serverpod_cli/test/commands/create/tui/state_test.dart
```diff
@@ -0,0 +1,388 @@
...
+              );
+            },
+          );
+        },
+      );
+
+      test(
+        'when selecting non-postgres database config option, '
+        'then config requirements are evaluated for auth config',
+        () {
```
marcelomendoncasoares: Suggestion: Also add another that auth is not possible to enable when Postgres is disabled.

marcelomendoncasoares (changes requested):
Great work with this feature, @Crazelu! Also nice integration with the existing assets for `serverpod start`.

tools/serverpod_cli/lib/src/commands/create.dart
Crazelu: We could do that.
But I think the template flag should be kept for the command as well to maintain backwards compatibility and for use cases where tui is not needed for create.

The specified template from the command will be automatically selected in the TUI app and users can change that.

tools/serverpod_cli/lib/src/commands/create/tui/config.dart
Crazelu: A follow up PR will handle this.

tools/serverpod_cli/lib/src/commands/create.dart
Crazelu: This was the original idea but there is an existing `interactive` flag and I did not want that to have two meanings for the create command.

tools/serverpod_cli/lib/src/create/create.dart
Crazelu: There won't be a need for a new method or updating the return type if the start instructions mention `serverpod start` instead of `cd $relativeServerPath` and `dart run bin/main.dart`

tools/serverpod_cli/test/commands/create/tui/state_test.dart
Crazelu: I think that this will make more sense if we were testing the nocterm component directly and using the arrow up and down keys. Then the described behaviour would be accurate.

But for the unit test, that behaviour (navigating up/down) is not inferred.

tools/serverpod_cli/lib/src/create/create.dart
Crazelu: The start instructions now mention `serverpod start`. No need for the modifications.
[0252d85](https://github.com/serverpod/serverpod/pull/5023/commits/0252d853c823f228ce5c2a232b1e132e9bec7b6a)

tools/serverpod_cli/lib/src/create/create.dart
Crazelu: The start instructions now mention `serverpod start`. No need for the modifications.
[0252d85](https://github.com/serverpod/serverpod/pull/5023/commits/0252d853c823f228ce5c2a232b1e132e9bec7b6a)

tools/serverpod_cli/lib/src/commands/create.dart
Crazelu: Done in [1cc2564](https://github.com/serverpod/serverpod/pull/5023/commits/1cc25646e213c443e89481dbf4aa9c0ecf397cd5).

Running `serverpod create someProject --mini` does not show the tui as there are no configurations to be made for mini. But `server` and `module` templates will show the `Project Type` option and users can still switch to mini (all configurations disappear from the TUI).

tools/serverpod_cli/test/commands/create/tui/state_test.dart
Crazelu: Done [here](https://github.com/serverpod/serverpod/pull/5023/changes/1cc25646e213c443e89481dbf4aa9c0ecf397cd5#diff-c49b54bf36c200332d03d540165d29081446e2b59c846de5ad6a5d39a3a12c5eR400)

tools/serverpod_cli/lib/src/commands/create/tui/config.dart
marcelomendoncasoares: @coderabbitai Please create an issue for exposing the option to add `skills` to the `create` interactive command.

tools/serverpod_cli/lib/src/create/create.dart
```diff
@@ -396,36 +390,11 @@ void _logStartInstructions(String relativeServerPath) {
...
-    log.info(
-      'dart run bin/main.dart --apply-migrations',
-      type: TextLogType.command,
-    );
-  }
+  log.info(
+    'serverpod start',
+    type: TextLogType.command,
+    newParagraph: true,
+  );
```
marcelomendoncasoares, line 456: Suggestion: Regardless of using the `serverpod start`, it is still good to show the `cd <projectdir>` before suggesting the start command. This prevent conflicts when there are other projects on the same parent dir - [which is now broken for `serverpod start`](https://github.com/serverpod/serverpod/issues/5039).

tools/serverpod_cli/lib/src/commands/create.dart
marcelomendoncasoares: There is just an issue that I found while manually testing: if we launch the `create` with a `--template`, it will exit the process without showing the start instructions.

image

tools/serverpod_cli/lib/src/commands/create.dart
marcelomendoncasoares: I would argue that reusing the `interactive` is actually optimal, because it has no meaning for the `create` command. The `interactive` flag only affects the commands that depend on a server directory (`generate`, `create-migration`, `start`, etc.), and its description is perfect for what it means on `create`:

https://github.com/serverpod/serverpod/blob/37d6f91b9343e9fe6f908f389ff92872b77a1c16/tools/serverpod_cli/lib/src/runner/serverpod_command_runner.dart#L131-L137

The only use of it on create is to forward to `generate` and `create-migration` in the case of an upgrade.

And, above all, it is the natural flag that users would expect to control such behavior. Otherwise, we could have users trying to disable the TUI using the `interactive` flag - that will be there because of the command runner.

tools/serverpod_cli/lib/src/create/create.dart
marcelomendoncasoares, line 237: Suggestion: This is now wrong on TUI. If we press CTRL+C, it will exit the app instead of skipping the Flutter build. The best experience would be to say "press S to skip" and listen to this key during Flutter build to abort.

tools/serverpod_cli/lib/src/create/create.dart
```diff
@@ -87,7 +108,7 @@ Future<bool> performCreate(
   );
   var pubspecFile = File(p.join(serverpodDirs.projectDir.path, 'pubspec.yaml'));
   if (pubspecFile.existsSync()) {
-    log.error('Project $name already exists.');
+    _logError('Project $name already exists.');
     return false;
   }
```
marcelomendoncasoares, line 123: Suggestion: This is happening somewhat late on the process, which leads to a bad experience. We should fail even before launching the TUI, not after the user has already explored and chosen all options.

marcelomendoncasoares (changes requested):
After manually testing, I think we still need to polish a bit more before releasing the interactive create. Besides the comments, what I'm missing the most is the possibility to see what happened after the command finishes.

Specially when something goes wrong after the project exits - like if the pub get or Flutter build fails -, the TUI flashes the error, exits and shows the green check with start instructions.

A much nicer UX would be if we:
- Expect the user to press Q instead of instantly leaving, so they can calmly read the output.
- Also print the end message of each step before the instruction when quitting.

The ideal is that, even after reading the output and quitting the app, the user can still go back to a persistent result as we have today - allowing inclusively to read and copy errors again, if needed.

Image

tools/serverpod_cli/lib/src/commands/create.dart
Crazelu: That's because there are no start instructions for the module template.
We only have start instructions for server and mini

tools/serverpod_cli/lib/src/create/create.dart
Crazelu: Dry run added in [de3134f](https://github.com/serverpod/serverpod/pull/5023/commits/de3134f257d9b1e7efc6e290e11893779d802cfe)

tools/serverpod_cli/lib/src/create/create.dart
Crazelu: "press S to skip" added in [ceb9872](https://github.com/serverpod/serverpod/pull/5023/commits/ceb98722ba1a66d618fe6b13a53e483194daaecb)

tools/serverpod_cli/lib/src/create/create.dart
Crazelu: Done in [df6781f](https://github.com/serverpod/serverpod/pull/5023/commits/df6781fd671ece6d8cc582e3fbae43bd7a7958d4)

tools/serverpod_cli/lib/src/create/create.dart
marcelomendoncasoares: Nice elegant solution! We just skip waiting, but the project still gets built.

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5027, opened 2026-05-01
feat(serverpod_cli): Show progress spinners in non-TUI start path

nielsenko:
Spit and shine..

Wrap more phases of `serverpod start ` in log.progress() so they get the same braille spinners and per-step timings the TUI path already produces. 

Extra phases covered:
- loading project configuration, 
- starting Docker services,
- initializing analyzers (other path that actually awaits them before generation),
- and starting the server (both --no-fes and --fes branches).


marcelomendoncasoares (approved):
Nice UX touch!

PR #5028, opened 2026-05-01
feat: Apply migrations directly via cli (no server start)

### LINKED ISSUE #3119: Apply DB migration via the serverpod CLI command
### Problem to Solve

Today DB migrations are applied using the server runtime. But DB migrations are often better to run as a deployment step that is handled separately from actual server start.

### Proposal

Refactor the apply migrations logic in Serverpod server for reusability from the serverpod command, and add support for running it there.

(Issue #3118 is not strictly a prerequisite but would make using this command easier.)

### Use Case

As a user, I'd like to be able to run it as an operation that stands alone from the server, via the serverpod CLI command.

### Alternatives

An alternative could be to add a distinct subcommand to Serverpod server that effectively corresponds to the command line:

`$ dart run bin/main.dart --role maintenance --apply-migrations`

### Additional context

_No response_

- [ ] I want to work on this

### LINKED ISSUE #4967: The `serverpod start` command should have `--watch` and `--apply-migrations` by default
Currently, the `serverpod start` does not watch for changes or applies migrations automatically by default. Users must to pass `--watch` to watch for changes and hot reload the server and `-- --apply-migrations` to automatically apply migrations. We should flip the default to have both behaviors on by default, with a way for users to opt-out if they prefer.

nielsenko:
This PR allows migrations to be applied by cli without restarting pod. This plays well together with the `VmServiceProxy` (from #5022) as it avoids detaching any potential debugger when applying migration from TUI (or if an agent does it via the MCP).

Also fixes a long standing request.

Fixes: #3119


packages/serverpod_database/lib/src/migrations/migration_manager.dart
```diff
@@ -32,23 +40,70 @@ abstract class MigrationManager {
...
+       );
+
   /// Loads the installed versions of the migrations from the database.
   ///
-  /// This method depends on the table model that will be available only in the
-  /// server/client package.
+  /// Only `module` and `version` are read - the manager doesn't use the
+  /// stored timestamp, and skipping it side-steps dialect differences in
+  /// how the timestamp is encoded (sqlite stores ms-since-epoch integers,
+  /// postgres stores native timestamps).
```
marcelomendoncasoares: Suggestion: This note is now wrong since the `fromJson` correctly deserializes the datetime for each dialect.

tools/serverpod_cli/lib/src/commands/start/watch_session.dart
```diff
@@ -394,17 +405,31 @@ class WatchSession {
...
+  /// migrations are applied via that callback (typically a CLI-side
+  /// runner that connects to the database directly), and the running
+  /// pod is left in place - hot reload covers any model code changes.
   ///
-  /// If another restart or migration is in progress, this call waits for it
-  /// to finish before proceeding. Throws a [StateError] if the session has
-  /// been disposed or the compiler is not available (--no-fes mode). Throws
-  /// on compilation failure so the caller (MCP server) can report the error.
+  /// When no callback is supplied (legacy path), the pod is recompiled
+  /// and restarted with `--apply-migrations`.
```
marcelomendoncasoares: Question: Do we even need a legacy path, since it was only available on beta versions? The more branches we can remove, the simpler to test and maintain.

tools/serverpod_cli/lib/src/migrations/cli_migration_runner.dart
```diff
@@ -0,0 +1,154 @@
...
+
+    final manager = MigrationManager.fromDirectory(
+      Directory(serverDir),
+      runMode: runMode,
+    );
+    final applied = await manager.migrateToLatest(session);
+    return applied ?? const [];
+  } finally {
+    await pool.stop();
+  }
```
marcelomendoncasoares, line 57: Discussion: This code is not a direct replacement for the current apply on restart because it does not validate schema drift after migrations are applied - which is critical for feedback during development. From the original code:

```dart
  Future<void> _applyMigrations({
    required bool applyRepairMigration,
    required bool applyMigrations,
  }) async {
    bool verified;

    try {
      _internalLogVerbose('Initializing migration manager.');
      var migrationManager = ServerMigrationManager(
        Directory.current,
        runMode: runMode,
      );

      /* ... apply repair migration ... */

      if (applyMigrations) {
        _internalLogVerbose('Applying database migrations.');
        var migrationsApplied = await migrationManager.migrateToLatest(
          internalSession,
        );

        if (migrationsApplied == null) {
          _writeLifecycleMessage('Latest database migration already applied.');
        } else {
          _writeLifecycleMessage(
            'Applied database migration${migrationsApplied.length > 1 ? 's' : ''}:',
          );
          for (var migration in migrationsApplied) {
            _writeLifecycleMessage(' - $migration');
          }
        }
      }

      // Verify migrations will always run on server start, regardless of applying migrations
      // because this current method is always called on the `unguardedStart` of Serverpod.
      _internalLogVerbose('Verifying database integrity.');
      verified = await ServerMigrationManager.verifyDatabaseIntegrity(
        internalSession,
      );
    } catch (e, stackTrace) {
      verified = false;

      const message = 'Failed to apply database migrations.';
      _reportException(e, stackTrace, message: message);
    }

    if (!verified) {
      _internalLogVerbose('Database integrity verification failed.');
      if (config.runMode == ServerpodRunMode.development) {
        throw ExitException(1);
      }
    }
  }
```

However, as stated below on the `_CliSerializationManager`, it does not have access to the runtime serialization manager of the project, which makes impossible to detect schema drift with the running models. To solve this, we can use the Insights endpoint to get the target table definitions and pass to the stub serialization manager to run the integrity check. Similar to how repair migrations are calculated.

```dart
import 'package:serverpod_cli/src/config_info/config_info.dart';

final client = ConfigInfo(runMode).createServiceClient();
client.insights.getTargetTableDefinitions();
```

tools/serverpod_cli/lib/src/migrations/cli_migration_runner.dart
```diff
@@ -0,0 +1,154 @@
...
+}) async {
+  final originalCwd = Directory.current;
+  Directory.current = Directory(serverDir);
+  try {
+    final passwords = PasswordManager(runMode: runMode).loadPasswords();
+    return ServerpodConfig.load(runMode, null, passwords);
+  } finally {
+    Directory.current = originalCwd;
+  }
+}
```
marcelomendoncasoares: Suggestion: Reuse the existing `ConfigInfo` class from `serverpod_cli/src/config_info/config_info.dart`.

tools/serverpod_cli/lib/src/migrations/cli_migration_runner.dart
```diff
@@ -0,0 +1,154 @@
...
+
+  try {
+    // Session and Database have a circular construction dependency:
+    // DatabaseConstructor.create needs a session, the session needs a
+    // db. We break the cycle by initializing _db post-construction.
+    final session = _CliDatabaseSession();
+    session._db = DatabaseConstructor.create(
+      session: session,
+      poolManager: pool,
+    );
```
marcelomendoncasoares, line 47: Question: Should we better expose a natural way for getting the database from configs on the `serverpod_database` instead?

tools/serverpod_cli/test/commands/start/watch_session_test.dart
```diff
@@ -703,6 +703,154 @@ void main() {
...
+          compiler: compiler,
+          generate: (affectedPaths, requirements) async =>
+              (success: true, generatedFiles: <String>{}),
+          createServer:
+              (String dillPath, {List<String> extraArgs = const []}) =>
+                  throw StateError('not used'),
+          initialServer: server,
+          generatedDirPaths: {'/generated'},
+          applyMigrationsAction: () async => throw StateError('boom'),
+        );
```
marcelomendoncasoares: Suggestion: Instead of using a `appliedVersions` as a `List<String>` that is configured on each test, replace it by a `List<String> Function()` so you can configure to throw without having to rewrite the `WatchSession`.

tools/serverpod_cli/test/commands/start/watch_session_test.dart
```diff
@@ -703,6 +703,154 @@ void main() {
...
+          createServer:
+              (String dillPath, {List<String> extraArgs = const []}) =>
+                  throw StateError('not used'),
+          initialServer: server,
+          generatedDirPaths: {'/generated'},
+          applyMigrationsAction: () async {
+            followUpRan = true;
+            return const <String>[];
+          },
+        );
```
marcelomendoncasoares: Comment: Then you won't need to redeclare the `WatchSession` here also, just change the value of the `applyMigrationsAction` callback.

marcelomendoncasoares (changes requested):
Few suggestions for improvement. On the integrity check, we can spin to fix as a follow-up, as discussed.

tools/serverpod_cli/lib/src/migrations/cli_migration_runner.dart
nielsenko: @coderabbit create follow-up issue

tools/serverpod_cli/lib/src/migrations/cli_migration_runner.dart
nielsenko: yes, serverpod_database should expose this.

@coderabbit Create an issue

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5030, opened 2026-05-01
refactor: Default to using --watch

### LINKED ISSUE #4967: The `serverpod start` command should have `--watch` and `--apply-migrations` by default

nielsenko:


marcelomendoncasoares (approved):
LGTM! Surgical change 🪡

PR #5034, opened 2026-05-01
fix: Skip `flutter` requirement if running `serverpod_cli` on CI

### LINKED ISSUE #5035: Flutter should not be required by `serverpod_cli` when running from CI
### Problem to Solve

Currently, the `serverpod_cli` throws if running any command from an environment without `flutter` on the `PATH`. Although this make sense for local development because we expect users to call the `create` command (which requires `flutter` to build), on CI we can have users developing Dart-only modules and they should not need to use Flutter (which is slow to setup) just to run unrelated commands.

### Proposal

Remove the `flutter` requirement when running `serverpod_cli` from the CI.

### Use Case

Users developing Dart-only modules that want light CI pipelines without Flutter.

### Alternatives

Install Flutter and have longer-running pipelines.

### Additional context

_No response_

- [x] I want to work on this

marcelomendoncasoares:
Closes: #5035

## Pre-launch Checklist

## Breaking changes

None.


nielsenko (approved):
.. but why not just skip the check altogether. People will realize when they need flutter

marcelomendoncasoares:
> .. but why not just skip the check altogether. People will realize when they need flutter

I think it is ok to prevent a "runtime failure" that would be a bad experience for first time users (throw in the middle of the `create` due to only Flutter not installed). But, as the issue suggests, this can only happen locally.

marcelomendoncasoares:
Fixed a weird behavior on https://github.com/serverpod/serverpod/pull/5034/commits/c5907acc9ab9630a535f79431c3fcf6fd67f97c6 that all three CI pipelines that generate pubspecs were hanging stuck after `closeLogger` (inside `_preExit`).

We were never explicitly ending the process with `exit(0) - which fixed the issue -, but seems weird that just now it started to consistently hang at the same spot.

PR #5044, opened 2026-05-04
fix: Prune trailing empty migration dirs (#3985)

### LINKED ISSUE #3985: Create migration should not throw error when the latest migration folder is empty

Moe1211:
When the newest migration version directory exists but is completely empty (common after discarding generated files), `create-migration` used to treat that as a corrupted migration and fail when reading `definition_project.json`.

This PR aligns behavior with [issue #3985](https://github.com/serverpod/serverpod/issues/3985):

- **`FileSystemMigrationArtifactStore.listVersions`** (server `migrations/`): from the end of the sorted version list, remove directories that contain **no files or subdirectories**. Stop as soon as a non-empty directory is found, so a folder that still has files but is missing real migration artifacts continues to surface the existing “missing artifact” / corrupted path.
- If any directory was removed and **`migration_registry.txt` already exists**, rewrite it so it matches the remaining versions.
- **`ClientMigrationArtifactStore.listVersions`** applies the same rule under `lib/migrations/` and refreshes **`migration_registry.dart`** when pruning occurred and that file exists (so `part` directives do not point at removed folders).
- **Tests:** `packages/serverpod_database/test/migration_artifacts_store/file_system_migration_artifact_store_test.dart` and an extra case in `tools/serverpod_cli/test/integration/migrations/client_migration_artifact_store_test.dart`.

Fixes: #3985

---

## Pre-launch Checklist

## Breaking changes

_None._


Moe1211:

packages/serverpod_database/lib/src/migrations/migration_artifacts_store/file_system.dart
```diff
@@ -247,6 +261,44 @@ class FileSystemMigrationArtifactStore
...
+        continue;
+      }
+      if (!await _directoryHasNoEntities(versionDir)) {
+        break;
+      }
+      await versionDir.delete(recursive: false);
+      versions.removeLast();
+      prunedAny = true;
+    }
+    return prunedAny;
```
marcelomendoncasoares: Suggestion: We should log a warning with all versions that were pruned.

packages/serverpod_database/lib/src/migrations/migration_artifacts_store/file_system.dart
```diff
@@ -247,6 +261,44 @@ class FileSystemMigrationArtifactStore
...
+  }
+
+  static Future<bool> _directoryHasNoEntities(Directory dir) async {
+    await for (final _ in dir.list(followLinks: false)) {
+      return false;
+    }
+    return true;
+  }
```
marcelomendoncasoares: Suggestion: Extract this to a helper to reuse on both server and client without duplicating the logic.

packages/serverpod_database/test/migration_artifacts_store/file_system_migration_artifact_store_test.dart
```diff
@@ -0,0 +1,122 @@
...
+
+    tearDown(() async {
+      if (await projectRoot.exists()) {
+        await projectRoot.delete(recursive: true);
+      }
+    });
+
+    test(
+      'removes trailing migration version directories that contain no files',
+      () async {
```
marcelomendoncasoares: Suggestion: Tests should follow the Given/when/then pattern. The "Given" is the test scenario and usually matches what the `setUp` of the group prepares. The "when" is the acting and should only contain conditions that are related to the action, such as "with" parameters. And the "then" is the result expected with the test.

The current test could be:

- Given a FileSystemMigrationArtifactStore and the last migration version directory empty,
  - when listing migration versions,
    - then the empty trailing migration version is removed.
    - then the migration registry is updated removing the trailing empty version.
- Given a FileSystemMigrationArtifactStore and no empty trailing migration version directory...
- Given a FileSystemMigrationArtifactStore with two empty trailing migration version directories...

Also use the `test_descriptor` package to create the structure of the test directories (on the "Given" `setUp`) so that it gets easy to read what to expect.

packages/serverpod_database/test/migration_artifacts_store/file_system_migration_artifact_store_test.dart
```diff
@@ -0,0 +1,122 @@
...
+        final versions = await store.listVersions();
+
+        expect(versions, ['10000000000000']);
+        expect(await second.exists(), isFalse);
+      },
+    );
+
+    test(
+      'rewrites migration_registry.txt when trailing empty directories are pruned',
+      () async {
```
marcelomendoncasoares: Comment: This is a secondary "then" on the same "when" group: "when listing migration versions".

tools/serverpod_cli/test/integration/migrations/client_migration_artifact_store_test.dart
```diff
@@ -159,4 +159,38 @@ void main() {
       expect(registryContents, contains('_Migration$v2()'));
     },
   );
+
+  test(
+    'when the last client migration directory is completely empty, '
+    'listVersions removes it and refreshes migration_registry.dart',
+    () async {
```
marcelomendoncasoares: Suggestion: Following the explanation of the Given/when/then above, this is a different "Given" group, similar to the one suggested on the previous comment.

marcelomendoncasoares (changes requested):
Thank you for your contribution with this fix, @Moe1211! And sorry for the long delay on the review, we have been heads down on a big release in the past month.

Your approach is very good! It fixes the issue with negligible performance penalty - given that we just check the trailing directory and abort on the first found file. I made only a few suggestions to reduce code duplication and make the tests more adherent to the pattern.

I know it has been a while since you opened the PR, so please signal if you do not intend to work on it anymore - in which case we'll get it ready for merge.

marcelomendoncasoares (approved):

PR #5048, opened 2026-05-04
feat: Restore terminal settings when exiting tui process

### LINKED ISSUE #5024: The new `start` and `create` commands with `nocterm` leave the terminal unusable on Linux
### Describe the bug

After running the new `serverpod start` or `serverpod create --name <projectname>`, the terminal on Linux becomes unusable. No typed char is printed until blindly typing `stty sane` and hitting enter.

The cause seems to be that [Dart does not restore the terminal input mode automatically after a process exists](https://dart-review.googlesource.com/c/sdk/+/190484), and neither does Linux - differently from MacOS. Since `nocterm` changes the input to raw, the process ends and the terminal is still in raw mode.

### To reproduce

Simply call one of the two mentioned commands on Linux and exit (either normally or with error).

### Expected behavior

The terminal should maintain its original state - even regardless of terminating successfully, with error or through a signal.

### Library version

3.5.0-beta.5 (unreleased, main)

### Platform information

Ubuntu 24.04 LTS (WSL)

### Additional context

The solution might be something like:

```dart
import 'dart:io';

void main() {
  final originalLineMode = stdin.lineMode;
  final originalEchoMode = stdin.echoMode;

  try {
    // Run the TUI
  } finally {
    stdin.lineMode = originalLineMode;
    stdin.echoMode = originalEchoMode;
  }
}
```

We should also capture OS signals to ensure that we revert for a CTRL+C as well. This should be a generic wrapper for everywhere we use `nocterm` (possibly in a centralized app widget that we reuse, with the style baked in).

- [ ] I want to work on this

Crazelu:
`echoMode` and `lineMode` are restored when exiting nocterm. Restoration also happens on SIGINT and SIGTERM.

Closes [5024](https://github.com/serverpod/serverpod/issues/5024)

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/create.dart
+++ b/tools/serverpod_cli/lib/src/commands/create.dart
@@ -7,6 +7,7 @@ import 'package:nocterm/nocterm.dart';
 import 'package:serverpod_cli/src/commands/create/tui/app.dart';
 import 'package:serverpod_cli/src/commands/create/tui/state.dart';
 import 'package:serverpod_cli/src/commands/create/tui/state_holder.dart';
+import 'package:serverpod_cli/src/commands/tui/run_app.dart';
 import 'package:serverpod_cli/src/commands/tui/terminal_backend.dart';
 import 'package:serverpod_cli/src/commands/tui/tui_log_writer.dart';
 import 'package:serverpod_cli/src/create/create.dart';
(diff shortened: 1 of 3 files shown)


nielsenko (approved):
👍

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5049, opened 2026-05-04
fix(cli): Serialize handleFileChange through _pending

nielsenko:
Route handleFileChange through _pending so all entry points (file watcher, vm-service proxy, MCP/TUI buttons) serialize.

Previously, if using `serverpod start --watch` (now default) and _hot-reload on save_ in you IDE (common), when saving a file would trigger two codepaths concurrently. One via the file change, and the other via the requested hot-reload. This would compete for accessing the `FES` causing a crash.

This PR ensures file handling is serialized on the session with all other operations.


marcelomendoncasoares (approved):
🪡

PR #5052, opened 2026-05-05
feat: Add skills configuration to create command

### LINKED ISSUE #5038: Expose 'skills' option in the 'serverpod create' interactive TUI command
## Summary

The `serverpod create` interactive TUI command ([`ServerpodCreateConfig`](https://github.com/serverpod/serverpod/pull/5023/files#diff-...)) is missing an option to enable/disable `skills` support when creating a new project.

## Context

During the review of #5023, it was noted that the `skills` configuration option is not yet exposed in the interactive `create` command. Since the `.agents` strategy will be used, this should be a simple **on/off toggle**.

A follow-up PR was agreed upon to handle this (see comment: https://github.com/serverpod/serverpod/pull/5023#discussion_r3170369359).

## Expected Behavior

- Add a `skills` entry to `ServerpodCreateConfig` (or equivalent) as an on/off option.
- The option should integrate with the existing `TemplateContext` and follow the `.agents` strategy.

## Requested by

@marcelomendoncasoares

Crazelu:
Exposed skills configuration in the create command's TUI. When enabled, skills are installed for Claude and other IDEs using the `skills` package. Skills are enabled by default for `quickstart` and `create --no-interactive`.

Closes [5038](https://github.com/serverpod/serverpod/issues/5038)

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tests/bootstrap_project/test/module_quickstart_test.dart
+++ b/tests/bootstrap_project/test/module_quickstart_test.dart
@@ -339,7 +339,7 @@ void main() {
         test('has agent skills installed', () {
           expect(
             Directory(
-              path.join(tempPath, projectName, '.agent', 'skills'),
+              path.join(tempPath, projectName, '.agents', 'skills'),
             ).existsSync(),
             isTrue,
           );
--- a/tests/bootstrap_project/test/module_quickstart_test.dart
+++ b/tests/bootstrap_project/test/module_quickstart_test.dart
@@ -335,6 +335,21 @@ void main() {
             isTrue,
           );
         });
+
+        test('has agent skills installed', () {
+          expect(
+            Directory(
+              path.join(tempPath, projectName, '.agent', 'skills'),
+            ).existsSync(),
+            isTrue,
+          );
+          expect(
+            Directory(
+              path.join(tempPath, projectName, '.claude', 'skills'),
+            ).existsSync(),
+            isTrue,
+          );
+        });
       });
     });
   });
--- a/tests/bootstrap_project/test/module_test.dart
+++ b/tests/bootstrap_project/test/module_test.dart
@@ -350,7 +350,7 @@ void main() {
         test('has agent skills installed', () {
           expect(
             Directory(
-              path.join(tempPath, projectName, '.agent', 'skills'),
+              path.join(tempPath, projectName, '.agents', 'skills'),
             ).existsSync(),
             isTrue,
           );
--- a/tests/bootstrap_project/test/module_test.dart
+++ b/tests/bootstrap_project/test/module_test.dart
@@ -346,6 +346,21 @@ void main() {
             reason: 'Root pubspec.lock file does not exist.',
           );
         });
+
+        test('has agent skills installed', () {
+          expect(
+            Directory(
+              path.join(tempPath, projectName, '.agent', 'skills'),
+            ).existsSync(),
+            isTrue,
+          );
+          expect(
+            Directory(
+              path.join(tempPath, projectName, '.claude', 'skills'),
+            ).existsSync(),
+            isTrue,
+          );
+        });
       });
     });
   });
--- a/tools/serverpod_cli/pubspec.yaml
+++ b/tools/serverpod_cli/pubspec.yaml
@@ -52,6 +52,8 @@ dependencies:
   serverpod_serialization: 3.5.0-beta.5
   serverpod_service_client: 3.5.0-beta.5
   serverpod_shared: 3.5.0-beta.5
+  # skills must be pinned to 0.2.2 to avoid breaking changes
+  # as the package receives external contribution
   skills: 0.2.2
   source_span: ^1.10.0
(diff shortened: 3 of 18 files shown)


tests/bootstrap_project/test/module_quickstart_test.dart
```diff
@@ -335,6 +335,21 @@ void main() {
             isTrue,
           );
         });
+
+        test('has agent skills installed', () {
+          expect(
+            Directory(
+              path.join(tempPath, projectName, '.agent', 'skills'),
+            ).existsSync(),
```
marcelomendoncasoares, line 343: Suggestion: There is a bug on the `skills` package: [the standard is `.agents`](https://github.com/agentskills/agentskills/issues/15) (with the trailing "s"), not `.agent`. [While the fix doesn't land](https://github.com/serverpod/skills/pull/12), either we move the files after calling `getSkills` or we create a symlink.

The easiest would be to move, but we must test whether it will duplicate the skills on subsequent calls to `skills get` by the user - or if this duplication is fine, since the `.agent` won't be discovered anyway.

tools/serverpod_cli/pubspec.yaml
```diff
@@ -52,6 +52,7 @@ dependencies:
   serverpod_serialization: 3.5.0-beta.5
   serverpod_service_client: 3.5.0-beta.5
   serverpod_shared: 3.5.0-beta.5
+  skills: ^0.2.2
```
marcelomendoncasoares: Question: Since we're using implementation imports and the `skills` package is receiving external contribution, should we pin the exact version instead?

tests/bootstrap_project/test/module_test.dart
```diff
@@ -346,6 +346,21 @@ void main() {
...
+              path.join(tempPath, projectName, '.agent', 'skills'),
+            ).existsSync(),
+            isTrue,
+          );
+          expect(
+            Directory(
+              path.join(tempPath, projectName, '.claude', 'skills'),
+            ).existsSync(),
+            isTrue,
+          );
```
marcelomendoncasoares, line 362: Question: Are these folders on the `.gitignore` as well? Should they be?

marcelomendoncasoares (commented):
Only three considerations for discussion.

tests/bootstrap_project/test/module_quickstart_test.dart
Crazelu: Interesting issue.
We can rename the directory after calling getSkills. I expect subsequent calls to skills get to recreate the .agent directory and ignore the renamed one so landing a proper fix should be prioritized on the skills library.

tools/serverpod_cli/pubspec.yaml
Crazelu: We should!

tests/bootstrap_project/test/module_test.dart
Crazelu: They are not gitignored. My thinking is that checking into version control allows for collaborative use of skills in projects. Of course, developers can gitignore if they don’t want this.

tests/bootstrap_project/test/module_quickstart_test.dart
Crazelu: On a second thought, a symlink would be better. Subsequent calls to skills get will not be a problem and that should do until the fix lands.

Crazelu: A problem with symbolic links on windows is that elevated privileges are required. I have opted for moving the files into `.agents` directory.

tools/serverpod_cli/pubspec.yaml
Crazelu: Done in [a1f0c0b](https://github.com/serverpod/serverpod/pull/5052/commits/a1f0c0b19777652ef4673828945cd4c93ae6b9b6)

tests/bootstrap_project/test/module_quickstart_test.dart
Crazelu: Moving files done in [b25393b](https://github.com/serverpod/serverpod/pull/5052/commits/b25393beb80d8001bfa80d09748b8ae15a48a23c)

tests/bootstrap_project/test/module_test.dart
marcelomendoncasoares: I agree that it depends on each team. My only thought on it is that there are a lot of skill files that will pollute the first experience looking at the working tree of the created project. Making them ignored allow the user to focus on the actual project files instead.

tools/serverpod_cli/pubspec.yaml
marcelomendoncasoares: Worth adding a note on the pubspec to inform why we're pinning the exact version.

marcelomendoncasoares (approved):
LGTM! ⛷️

PR #5089, opened 2026-05-08
feat(cli): Add hot_restart and create_repair_migration MCP tools and TUI buttons

### LINKED ISSUE #5088: Support hot-restart in TUI and MCP
### Problem to Solve

Now that apply migrations happens without restart of the pod there is no longer a way to do that without stopping and starting the orchestrating `serverpod start` command.

### Proposal

Add support for hot-restart to TUI and MCP. Here hot-restart means tooling (TUI/MCP) stays up, only the pod restarts. From the pods perspective it is a true process restart - unlike Flutter where the process stays up, but the main isolate is killed and the entry point (main) is re-run.

### Use Case

Like with Flutter it is occasionally useful to clear memory and re-run main. This proposal will restart the process not just the roor isolate, so it will also clear out any memory associated with FFI bounds native libs.

### Alternatives

Just restarting the root isolate (actually more work for us, and inherits the FFI binding problems that plague Flutter).

### Additional context

_No response_

- [x] I want to work on this

### LINKED ISSUE #5117: Make it possible to `create-migration --force` from the TUI
It's currently not possible to create a destructive migration from the TUI; we should make that possible. Also, the output is very "ordinary-CLIy", we could provide a better error message here when running the TUI. (It may be extra important if an agent reads the logs, and want to use the CLI over the MCP.)

Image

nielsenko:
Adds two new tools to the serverpod MCP server, each with a matching TUI button, and establishes a Flutter-style Shift convention for the TUI: lowercase keys are the common case, Shift+key is the more disruptive sibling.

- `hot_restart` (MCP) / `Shift+R` (TUI) - Full server-process restart with a fresh recompile.
- `create_repair_migration` (MCP) / `Shift+P` (TUI) 

## Migration refactor

- Now warnings throw `MigrationAbortedException` (matching the create-migration sibling); null means only "no changes detected".
- Exceptions (`MigrationVersionLoadException`, `MigrationRepairTargetNotFoundException`, etc.) now own their user-facing prose via `toString`. 

## TUI niceties

- Help overlay now shrink-wraps and scrolls when the terminal is short.
- New `ShrinkWrapScrollView` component (`tui/components/shrink_wrap_scroll_view.dart`) for that pattern.

Closes: #5088 
Closes: #5117

## Pre-launch Checklist

## Breaking changes

None. (If you don't count Human interface - which we don't).

### DIFF
--- a/packages/serverpod_shared/lib/src/migration_exceptions.dart
+++ b/packages/serverpod_shared/lib/src/migration_exceptions.dart
@@ -15,6 +15,13 @@ class MigrationVersionLoadException implements Exception {
     required this.moduleName,
     required this.exception,
   });
+
+  @override
+  String toString() =>
+      'Unable to determine latest database definition due to a corrupted '
+      'migration. Please re-create or remove the migration version and try '
+      'again. Migration version: "$versionName" for module "$moduleName".\n'
+      '$exception';
 }
 
 /// Exception thrown when trying to load the live database definition.
@@ -26,6 +33,11 @@ class MigrationLiveDatabaseDefinitionException implements Exception {
   MigrationLiveDatabaseDefinitionException({
     required this.exception,
   });
+
+  @override
+  String toString() =>
+      'Unable to fetch live database schema from server. Make sure the '
+      'server is running and is connected to the database.\n$exception';
 }
 
 /// Exception thrown when writing a migration fails.
@@ -37,6 +49,9 @@ class MigrationRepairWriteException implements Exception {
   MigrationRepairWriteException({
     required this.exception,
   });
+
+  @override
+  String toString() => 'Unable to write repair migration.\n$exception';
 }
 
 /// Exception thrown when a migration target is not found.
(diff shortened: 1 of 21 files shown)


tools/serverpod_cli/lib/src/commands/tui/help_overlay.dart
```diff
@@ -37,7 +37,7 @@ class _HelpOverlayState extends State<HelpOverlay> {
     (
       'Tabs',
       [
-        ('Tab / →', 'Next tab'),
+        ('Tab / ->', 'Next tab'),
         ('←', 'Previous tab'),
```
marcelomendoncasoares, line 43: Question: Isn't this asymmetric now?

marcelomendoncasoares (approved):
Only one minor question on the help text, but LGTM! ♻️

nielsenko:
@marcelomendoncasoares This grew a bit, so calling in for a second round

PR #5109, opened 2026-05-11
fix(auth): Disable sign-in UI while authentication is in progress (issue #4572)

### LINKED ISSUE #4572: The interface is not disabled while signing in
When using sign in with Apple or Google, the interface is still active while we are waiting for a response from the server. It should probably not be possible to start two sign in methods at once.

abdulawalarif:
This PR fixes an issue where the sign-in screen stays interactive while an authentication flow is in progress. Users could start another provider sign-in or navigate away before Serverpod finished login, which made overlapping sign-in attempts possible.

The change adds a shared `SignInFlowCoordinator` that blocks interaction and back navigation for the full attempt. `SignInWidget` installs it automatically, and the built-in provider widgets use it without extra app wiring. `onAuthenticated` and `onError` behavior is unchanged.

`GoogleAuthController.signIn()` now stays active until the Google SDK step and the Serverpod login both finish, so the UI lock also covers the server round-trip on Google.

## Fixed issues

Fixes [The interface is not disabled while signing in #4572](https://github.com/serverpod/serverpod/issues/4572)

## Testing

- Ran `dart analyze --fatal-infos` in `modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter`
- Ran `flutter test` in the same package (55 tests, including `sign_in_flow_coordinator_test.dart`)
- Manual: Google sign-in on Android — other controls stay non-interactive until the flow completes

## Pre-launch Checklist

- [ ] This update contains only one single feature or bug fix and nothing else.
- [ ] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation.

## Breaking changes

None


abdulawalarif:

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/common/sign_in_flow_coordinator.dart
```diff
@@ -0,0 +1,94 @@
...
+  /// Creates a coordinator that exposes the current authentication lock state.
+  const SignInFlowCoordinator({
+    super.key,
+    required ValueNotifier<bool> isAuthenticating,
+    required super.child,
+  }) : super(notifier: isAuthenticating);
+
+  /// Returns the active coordinator state from [context], if present.
+  static SignInFlowCoordinatorState? of(BuildContext context) {
+    return context.findAncestorStateOfType<SignInFlowCoordinatorState>();
```
FXschwartz: The coordinator itself is solid, this is just about the `SignInFlowCoordinator` InheritedNotifier. Since `of()` looks the state up with `findAncestorStateOfType` and the PopScope reactivity comes from the `ValueListenableBuilder` below, nothing actually reads the inherited widget, so the notifier: plumbing isn't really doing anything.

I'd just drop the `InheritedNotifier` and have `build()` return the `ValueListenableBuilder/PopScope` directly. The behavior's identical and it's one less layer.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/common/sign_in_flow_coordinator.dart
```diff
@@ -0,0 +1,94 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return SignInFlowCoordinator(
+      isAuthenticating: _isAuthenticating,
+      child: ValueListenableBuilder<bool>(
+        valueListenable: _isAuthenticating,
+        builder: (context, isAuthenticating, child) {
+          return PopScope(
+            canPop: !isAuthenticating,
```
FXschwartz: Since the lock only releases once the SDK + server round-trip finishes, and there's no timeout or cancel, a stalled `endpoint.login` would freeze the user on a blocked screen with back disabled. 

Could we add a timeout (or some escape hatch) so we can't trap them if the server hangs?

FXschwartz (changes requested):
Really nice work on this @abdulawalarif! The SignInFlowCoordinator approach is clean.

Just a couple of small things I commented on
- The lock has no timeout/cancel, so a stalled server call would keep the screen blocked.
- The SignInFlowCoordinator InheritedNotifier isn't actually being read, so I think it can just be dropped.

Thanks for tackling this one!

abdulawalarif:
Hi, @FXschwartz 
I removed the unused InheritedNotifier and added a 30s UI unlock timeout.
Please let me know if it's okay now

FXschwartz (approved):
@abdulawalarif Approved! Thanks again for taking care of this one. Great work!

PR #5114, opened 2026-05-11
docs: Improves serverpod-overview skill.

vlidholt:
Improves the `serverpod-overview` skill. As this skill is likely to almost always be included, it's been slimmed down to only contain information that is crucial to the agent (e.g., installation instructions and flutter checks are removed). The skill makes the agent much more likely to correctly use the MCP server. The skill is validated over multiple runs with lower-tier LLMs.


packages/serverpod/skills/serverpod-overview/SKILL.md
```diff
@@ -1,71 +1,38 @@
...
 
-## Prerequisites
+There can also be packages that share code, e.g., `my_project_shared`.
 
-- **Flutter** installed (`flutter doctor`)
-- **Docker** (for local PostgreSQL and Redis via project's `docker-compose.yaml`)
+The server exposes endpoint classes that the client calls via generated RPC client. Add methods to the endpoints, the code generation will recreate them on the client side. Models are defined in YAML and generate Dart classes for both server and client.
 
-## Installation
+Serverpod projects use a Postgres database for persistence and include an ORM, caching, real-time streaming (using Dart streams), file uploads, scheduling (called future calls), logging, and a built-in web server (Relic). Each of these features have specific skills that you can use to get more details.
```
marcelomendoncasoares, line 18: Question: Should we also mention SQLite?

marcelomendoncasoares (approved):
Seems like a good improvement, making the skill more slim and focused. I have only one small question, but LGTM! 🤖

packages/serverpod/skills/serverpod-overview/SKILL.md
vlidholt: I'm thinking, for now, no. We optimize for the happy path.

PR #5123, opened 2026-05-12
docs: Improvements to AI agent skills.

vlidholt:
Did a scan over all skills and made some improvements. Main additions:

- Added examples for how to work with users, authentication, and scopes. I think it's better than what we have, but could probably use a second pass sometime in the future.
- Some extra example code and clarifications.


packages/serverpod/skills/serverpod-auth/SKILL.md
```diff
@@ -0,0 +1,183 @@
...
+              },
+            ),
+          );
+  }
+}
+```
+
+**Check signed-in state**: `_isSignedIn = client.auth.isAuthenticated;`
+**Sign out**: `client.auth.signOutAllDevices()` or `client.auth.signOutDevice()`
+**Get user profile (email, full name, etc)**: `final userProfile = await client.modules.serverpod_auth_core.userProfileInfo.get()`
```
marcelomendoncasoares, line 60: Question: Should we mention somewhere that there is localization support using the `SignInLocalizationProvider` widget? It is a pretty common feature and could be useful to add a reference.

marcelomendoncasoares (approved):
Only one minor question, but LGTM! 🚀

packages/serverpod/skills/serverpod-auth/SKILL.md
vlidholt: It's not mentioned in the docs either. 😅 I'm leaving this for now, but adding an issue.

marcelomendoncasoares: Indeed... I'm adding this to the docs still from the set of features included in 3.5.0.

PR #5131, opened 2026-05-12
feat: Add IDE selection on serverpod create TUI

### LINKED ISSUE #5079: Allow users to check which IDEs they want skills and MCP to be installed to on `serverpod create`
Currently, we bloat the project with a lot of MCP configurations because there is no standard of where the file should be (or even its shape, as seen by Codex). We should explicitly ask users which IDEs/tools they use to only add those.

Crazelu:
This PR adds the ability to select multiple IDEs from the `serverpod create` TUI for which AI skills and MCP configurations be added.

By default, skills and MCP configurations are added for Claude and Cursor if the interactive create is not used.
This default also applies to `serverpod quickstart`.

Closes [5079](https://github.com/serverpod/serverpod/issues/5079)

## Pre-launch Checklist

## Breaking changes

None


tests/bootstrap_project/test_perform_create/ide_test.dart
```diff
@@ -0,0 +1,168 @@
...
+    },
+    "dart": {
+      "command": "dart",
+      "args": ["mcp-server"]
+    }
+  }
+}
+''';
+
+          test('for Antigravity', () {
```
nielsenko: This grouping is a bit unconventional. We usually don't use _then_ on group descriptions, or start tests with `for ...`.

tools/serverpod_cli/lib/src/commands/create/tui/config.dart
```diff
@@ -151,3 +172,18 @@ extension ServerpodTemplateTypeExtension on ServerpodTemplateType {
     ),
   };
 }
+
+extension IdeOptionsExtension on Set<IdeOption> {
+  List<TemplateIde> get toTemplateIdes {
+    return map((option) {
+      return switch (option) {
+        IdeOption.antigravity => TemplateIde.antigravity,
```
nielsenko, line 181: Should we start using dot shorthand notation, now that we require dart 3.10?

tools/serverpod_cli/lib/src/commands/create/tui/main_screen.dart
```diff
@@ -147,6 +155,51 @@ class MainScreen extends StatelessComponent {
...
+              config.label,
+              style: const TextStyle(
+                color: Color.defaultColor,
+                fontWeight: FontWeight.bold,
+              ),
+            ),
+            Wrap(
+              spacing: 2,
+              children: [
+                for (final option in config.options.indexed)
```
nielsenko: Why `.indexed`?! You are not using $1 anywhere..

tools/serverpod_cli/lib/src/commands/create/tui/state.dart
```diff
@@ -105,13 +105,32 @@ class CreateConfigState extends ServerpodState {
     configState._updateFocusedOption(delta);
     final focusedOptionIndex = configState.focusedOptionIndex;
     final newSelection = config.options[focusedOptionIndex];
-    _stateValues[config] = newSelection;
+
+    if (config.multiSelect) {
+      final selections = _stateValues[config];
+      if (selections?.contains(newSelection) == true) {
```
nielsenko: Maybe check for `null` here to avoid the prolific use of `?`and `!` later.

tools/serverpod_cli/lib/src/commands/create/tui/state.dart
```diff
@@ -105,13 +105,32 @@ class CreateConfigState extends ServerpodState {
...
+    if (config.multiSelect) {
+      final selections = _stateValues[config];
+      if (selections?.contains(option) == true) {
+        _stateValues[config] = selections!.difference({option});
+      } else {
+        _stateValues[config] = {...?selections, option};
+      }
+    } else {
+      _stateValues[config] = {option};
+    }
```
nielsenko, line 140: duplicated

tools/serverpod_cli/lib/src/commands/create/tui/state.dart
```diff
@@ -155,22 +174,35 @@ class CreateConfigState extends ServerpodState {
...
   }
 
   /// Returns true if [option] is the selected option for [config].
   bool getStatus<T extends ConfigOption>(
     ServerpodCreateConfig<T> config,
     T option,
   ) {
-    return _stateValues[config] == option;
+    final value = _stateValues[config];
+    return value?.contains(option) == true;
```
nielsenko: Maybe it is just me, but I prefer
```suggestion
    return value?.contains(option) ?? false;
```
it smells like js to me

tools/serverpod_cli/lib/src/commands/tui/components/checkbox.dart
```diff
@@ -0,0 +1,32 @@
...
+  });
+
+  final String label;
+  final bool value;
+  final TextStyle? style;
+
+  @override
+  Component build(BuildContext context) {
+    final theme = ServerpodTheme.of(context);
+    final marker = value ? '[✔]' : '[ ]';
```
nielsenko: I thought we went with the unicode characters  🤔

tools/serverpod_cli/lib/src/commands/tui/render_wrap.dart
```diff
@@ -0,0 +1,308 @@
+import 'dart:math' as math;
+import 'package:nocterm/nocterm.dart';
+// ignore: implementation_imports
+import 'package:nocterm/src/framework/terminal_canvas.dart';
+
+/// Render object for wrap layouts.
+class RenderWrap extends RenderObject
```
nielsenko: Why is this duplicated? _wrap.dart_ already defines this class (and friends). This file is not used anywhere by a quick scan.

tools/serverpod_cli/lib/src/commands/create/tui/state.dart
```diff
@@ -155,22 +174,35 @@ class CreateConfigState extends ServerpodState {
   }
 
   /// Returns the selected [ConfigOption] for [config].
+  /// For multi-select configs, use [getSelectedOptionsFor] instead.
   T? getSelectedOptionFor<T extends ConfigOption>(
     ServerpodCreateConfig<T> config,
   ) {
-    return _stateValues[config] as T?;
+    final value = _stateValues[config];
+    return value?.first as T?;
```
nielsenko: What if `value != null`, but `value.isEmpty`? 
```suggestion
    return value?.firstOrNull as T?;
```

nielsenko (dismissed):
Nice extension of the existing pattern. 

I did find few things..

Also, should we have defaults for the TUI as well (we do for non-interactive)?

tools/serverpod_cli/lib/src/commands/tui/components/checkbox.dart
Crazelu: The changes following the last discussion are not in yet

tools/serverpod_cli/lib/src/commands/tui/render_wrap.dart
Crazelu: Great catch. I missed a deletion

tests/bootstrap_project/test_perform_create/ide_test.dart
Crazelu: There are numerous group descriptions with _then_ which I've seen and influenced the style in this test.
However the tests in the group could have updated descriptions although those will duplicate some texts.

tools/serverpod_cli/lib/src/commands/create/tui/config.dart
Crazelu: Perhaps. I don't have a strong opinion on this.

tools/serverpod_cli/lib/src/commands/create/tui/main_screen.dart
Crazelu: 👍
$1 was used before but this UI has now been refactored. `.indexed` is no longer necessary here and in some other places.

tests/bootstrap_project/test/module_quickstart_test.dart
```diff
@@ -416,17 +382,6 @@ args = ["mcp-server", "--force-roots-fallback"]
...
-          test('for VSCode', () {
-            final vscode = File(
-              path.join(tempPath, projectName, '.vscode/mcp.json'),
-            );
-            expect(vscode.existsSync(), isTrue);
-            expect(
-              vscode.readAsStringSync(),
-              genericConfig.replaceAll('mcpServers', 'servers'),
-            );
-          });
```
marcelomendoncasoares, line 429: Suggestion: Since we already have the `.vscode` folder, it doesn't hurt to have the MCP configured for it by default. It is also the most used IDE - together with Cursor. So we'll stay with Claude, Cursor and VSCode by default.

tools/serverpod_cli/lib/src/commands/create/tui/config.dart
marcelomendoncasoares: Neither do I... Although I like the brevity, I know we'll also lose searchability - specially for agents, that rely so much on `grep`. Personally for now, I'm leaning towards not using for non-third-party properties.

tools/serverpod_cli/test/commands/create/tui/state_test.dart
```diff
@@ -127,32 +134,125 @@ void main() {
...
 
-          expect(
-            state.configValues,
-            containsAllInOrder([
-              ServerpodCreateConfig.template,
-              ServerpodCreateConfig.skills,
-            ]),
+          test(
+            'then getSelectedOptionsFor returns the selected ide',
+            () {
```
marcelomendoncasoares, line 164: Suggestion: This test description is invalid ("given" on a group inside another group with "Given a CreateConfigState with server template'"). Maybe these tests - and the other ones below - should be moved to the top-level?

tools/serverpod_cli/lib/src/create/template_context.dart
```diff
@@ -42,7 +44,6 @@ class TemplateContext {
       'web': web,
       'docker': docker,
       'database': database,
-      'skills': skills,
     };
   }
```
marcelomendoncasoares: Suggestion: Might be worth renaming to something like toMustachMap or so to make it clear that it is not a JSON representation. Otherwise future developers might wonder why the `ides` is not present, nor a `fromJson`.

tools/serverpod_cli/lib/src/create/ide.dart
```diff
@@ -0,0 +1,102 @@
...
+
+extension TemplateIdeExtension on TemplateIde {
+  String get effectiveConfig {
+    String result = config;
+    for (final replacement in replacements) {
+      result = result.replaceAll(replacement.from, replacement.to);
+    }
+    return result;
+  }
+}
```
marcelomendoncasoares, line 49: Question: This structure is pretty similar to the `Copier`/`Replacement` that we already have on the CLI. Should we just reuse it instead?

marcelomendoncasoares (changes requested):
Besides the comments, I also noticed we're not symmetric on the boxes between dark/light:

Image

The box is narrow on the inverted version, which makes it weird to understand what is selected. Is this because you couldn't find a white box of similar size?

---

Also, should we have the default IDE options marked by default?

Image

Crazelu:
@marcelomendoncasoares the issue you described [here](https://github.com/serverpod/serverpod/pull/5131#pullrequestreview-4319539571) seems to be font related. It looks good on macOS. I will change it to the ballot box with a cross like we discussed previously (U+2612).

tools/serverpod_cli/lib/src/create/ide.dart
Crazelu: We can reuse `Replacement` but not `Copier` as that is tied to file content modification; requiring directories.

tools/serverpod_cli/test/commands/create/tui/state_test.dart
Crazelu: Should be a `when` not a `given`

tools/serverpod_cli/lib/src/create/ide.dart
Crazelu: Done in [349842a](https://github.com/serverpod/serverpod/pull/5131/commits/349842ae16557a35044bb909beca78ab184565db)

tools/serverpod_cli/lib/src/create/template_context.dart
Crazelu: Done in [939a790](https://github.com/serverpod/serverpod/pull/5131/commits/939a7905d2bb68e003dca2aa42a92cb67d6a337e)

tests/bootstrap_project/test/module_quickstart_test.dart
Crazelu: Done in [a3d2a1f](https://github.com/serverpod/serverpod/pull/5131/commits/a3d2a1f274c7c12f350b5e70842ab8ed07b54ab6)

tools/serverpod_cli/test/commands/create/tui/state_test.dart
Crazelu: Done in [9465134](https://github.com/serverpod/serverpod/pull/5131/commits/946513460ab23c95c901004d905d0fbe48cadb27)

tests/bootstrap_project/test_perform_create/ide_test.dart
marcelomendoncasoares: You're probably referring to compilation and code generation tests. We do have a ton there that break the convention since the boilerplate to create each step is too high. I think in this particular case it is indeed harmless.

marcelomendoncasoares (approved):
LGTM! 🚀 

I did only one adjustment on the checkbox char to make it more symmetric (same size checked and unchecked). Tested on Mac, Linux and Windows, and they seem fine on all three.

As a final nit, I also fixed the stuck generator spinner during create by reusing the `IsolatedAnalyzers` that we use on `start`.

PR #5132, opened 2026-05-13
feat: Run flutter app on serverpod start

### LINKED ISSUE #5073: Web page reload is not triggered when editing code
The web page reload (on server hot reload) is only triggered when editing the templates. It is not triggered when editing the Dart code, which can also affect the webpage. The web page should be reloaded on any hot reload.

### LINKED ISSUE #5127: Manage `flutter run` from `serverpod start`
Basically:

- Run `flutter run -d chrome` (or `flutter run -d web-server` if `--flutter-headless` is passed) as a sub-process to `serverpod start`.
- Pipe the output of `flutter run` to a new tab in the `serverpod start` TUI.
- Monitor changes to the files in the Flutter app and trigger a hot reload on changes.
- If a hot restart is triggered on `serverpod start`, also do a hot restart in the app (send `r` or `R` to `stdin` of `flutter run`).
- Show a link to the app in the TUI.
- Add some sort of config for if we're running the Flutter app (initially, this can be as simple as passing a `--no-flutter` flag to `serverpod start`. But in the longer term, we should have a settings screen.

Crazelu:
This PR adds `flutter run` to `serverpod start`. 
The running flutter app is reloaded when dart files in the `lib` directory are modified.
In tui mode, a separate tab is added for the raw logs from the flutter process.
In non-tui mode, output is streamed to stdout with support for sending stdin to the flutter process to keep support for the Flutter command keys.

This new behaviour is turned on by default and opt-out with `--no-flutter`

Closes [5127](https://github.com/serverpod/serverpod/issues/5127)
Closes #5073

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -10,6 +10,7 @@ import 'package:serverpod_cli/analyzer.dart';
 import 'package:serverpod_cli/src/commands/generate.dart';
 import 'package:serverpod_cli/src/commands/messages.dart';
 import 'package:serverpod_cli/src/commands/start/file_watcher.dart';
+import 'package:serverpod_cli/src/commands/start/flutter_process.dart';
 import 'package:serverpod_cli/src/commands/start/kernel_compiler.dart';
 import 'package:serverpod_cli/src/commands/start/mcp_server.dart';
 import 'package:serverpod_cli/src/commands/start/mcp_socket.dart';
@@ -74,6 +75,13 @@ enum StartOption<V> implements OptionDefinition<V> {
       helpText: 'Show interactive terminal UI.',
     ),
   ),
+  flutter(
+    FlagOption(
+      argName: 'flutter',
+      defaultsTo: true,
+      helpText: 'Launch the Flutter app.',
+    ),
+  ),
   ;
 
   const StartOption(this.option);
@@ -117,6 +125,7 @@ class StartCommand extends ServerpodCommand<StartOption> {
   ) async {
     final watch = commandConfig.value(StartOption.watch);
     final useTui = commandConfig.value(StartOption.tui) && stdout.hasTerminal;
+    final launchFlutterApp = commandConfig.value(StartOption.flutter);
 
     // In TUI mode, start the UI immediately and do all setup in onReady.
     // This avoids a visible delay from config loading and Docker checks.
@@ -135,6 +144,7 @@ class StartCommand extends ServerpodCommand<StartOption> {
       final exitCode = await _runWithTui(
         commandConfig: commandConfig,
         watch: watch,
+        launchFlutterApp: launchFlutterApp,
         serverArgs: argResults?.rest ?? [],
         config: config,
       );
@@ -184,6 +194,7 @@ class StartCommand extends ServerpodCommand<StartOption> {
         serverArgs: serverArgs,
         watch: watch,
         docker: docker,
+        launchFlutterApp: launchFlutterApp,
         shutdown: shutdown,
       );
       switch (result) {
@@ -333,9 +344,13 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
   required ServerArgsRef serverArgs,
   required bool watch,
   required bool docker,
+  required bool launchFlutterApp,
   required _ShutdownSignal shutdown,
+  bool useTui = false,
   IOSink? serverStdoutSink,
   IOSink? serverStderrSink,
+  IOSink? flutterStdoutSink,
+  IOSink? flutterStderrSink,
   Future<void> Function(ServerProcess server)? onServerStart,
   List<Object> Function()? mcpGetLogHistory,
 }) async {
@@ -489,6 +504,19 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
     return true;
   });
 
+  FlutterProcess? flutterProcess;
+
+  if (launchFlutterApp) {
+    flutterProcess = FlutterProcess(
+      flutterPackageDir: p.joinAll(config.flutterPackagePathParts),
+      spawnFromTui: useTui,
+      stdoutSink: flutterStdoutSink,
+      stderrSink: flutterStderrSink,
+    );
+
+    await flutterProcess.start();
+  }
+
   // Construct the watch session.
   final runMode = runModeFromServerArgs(serverArgs.value);
   session = WatchSession(
@@ -505,6 +533,7 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
     },
     createServer: serverProcessFactory,
     initialServer: initialServerProcess,
+    flutterProcess: flutterProcess,
     generatedDirPaths: config.generatedDirPaths,
     applyMigrationsAction: () => _applyMigrationsForSession(
       serverDir: serverDir,
@@ -547,6 +576,7 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
 
   // File watcher (watch mode only).
   StreamSubscription<void>? fileChangeSub;
+  StreamSubscription<void>? flutterFileChangeSub;
   if (watch) {
     final watcher = FileWatcher(
       watchPaths: {
@@ -558,9 +588,24 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
         ),
       },
     );
+
     fileChangeSub = watcher.onFilesChanged
         .asyncMapBuffer((events) => session.handleFileChange(events.merge()))
         .listen((_) {});
+
+    FileWatcher? flutterWatcher;
+
+    if (launchFlutterApp) {
+      flutterWatcher = FileWatcher(
+        watchPaths: {
+          p.absolute(p.joinAll([...config.flutterPackagePathParts, 'lib'])),
+        },
+      );
+
+      flutterFileChangeSub = flutterWatcher.onFilesChanged
+          .asyncMapBuffer((events) async => flutterProcess?.reload())
+          .listen((_) {});
+    }
   }
 
   return WatchLoopReady(
@@ -571,6 +616,10 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
       fileChangeSub: fileChangeSub,
       closeAnalyzers: closeAnalyzers,
       stopDocker: startedDocker ? () => _stopDockerServices(serverDir) : null,
+      stopFlutterProcess: () async {
+        await flutterFileChangeSub?.cancel();
+        await flutterProcess?.stop();
+      },
       vmServiceInfoFile: vmServiceInfoFile,
     ),
   );
@@ -707,10 +756,12 @@ Future<String?> _checkExistingServer(String infoPath) async {
 Future<int> _runWithTui({
   required Configuration<StartOption> commandConfig,
   required bool watch,
+  required bool launchFlutterApp,
   required List<String> serverArgs,
   required GeneratorConfig config,
 }) async {
-  final holder = StartAppStateHolder(ServerWatchState());
+  final state = ServerWatchState()..showFlutterOutput = launchFlutterApp;
+  final holder = StartAppStateHolder(state);
   var backendStarted = false;
 
   // Shared shutdown signal
@@ -731,6 +782,7 @@ Future<int> _runWithTui({
           holder: h,
           commandConfig: commandConfig,
           watch: watch,
+          launchFlutterApp: launchFlutterApp,
           serverArgs: serverArgs,
           config: config,
           shutdown: shutdown,
@@ -764,6 +816,7 @@ Future<void> _runTuiBackend({
   required StartAppStateHolder holder,
   required Configuration<StartOption> commandConfig,
   required bool watch,
+  required bool launchFlutterApp,
   required List<String> serverArgs,
   required GeneratorConfig config,
   required _ShutdownSignal shutdown,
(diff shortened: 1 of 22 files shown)


tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
```diff
@@ -0,0 +1,159 @@
...
+    if (_process == null) return;
+    _process?.stdin.write('r');
+    log.info(flutterAppReloaded);
+  }
+
+  void restart() {
+    if (_process == null) return;
+    _process?.stdin.write('R');
+    log.info(flutterAppRestarted);
+  }
```
nielsenko: Why not actually connect to the VmService? and use that? This seems a bit hacky - to trigger reload/restart by writing r/R keys to flutter process's stdin.

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -558,9 +588,24 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
...
+      flutterWatcher = FileWatcher(
+        watchPaths: {
+          p.absolute(p.joinAll([...config.flutterPackagePathParts, 'lib'])),
+        },
+      );
+
+      flutterFileChangeSub = flutterWatcher.onFilesChanged
+          .asyncMapBuffer((events) async => flutterProcess?.reload())
+          .listen((_) {});
+    }
```
nielsenko, line 665: Actually the FlutterApp is also dependent on the generated client library. We should reload when that changes. But worse, this reload needs to be coordinated with the server. They should really happen together, when there are protocol impacting changes.

nielsenko (commented):
How will this play together with debugging from vscode?

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
nielsenko: BTW, where is this called (`restart`)?

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -489,6 +504,19 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
     return true;
   });
 
+  FlutterProcess? flutterProcess;
+
+  if (launchFlutterApp) {
+    flutterProcess = FlutterProcess(
+      flutterPackageDir: p.joinAll(config.flutterPackagePathParts),
```
nielsenko: What if this is empty?

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
```diff
@@ -0,0 +1,161 @@
...
+      originalLineMode = stdin.lineMode;
+      stdin.echoMode = false;
+      stdin.lineMode = false;
+    }
+
+    final process = await Process.start(
+      'flutter',
+      ['run', '-d', 'web-server'],
+      workingDirectory: _flutterPackageDir,
+    );
```
marcelomendoncasoares: Discussion: Should we expose a `--flutter-device` argument to allow the `start` to invoke other targets? And maybe a `--flutter-options=""` to receive options to forward to the `flutter run` command - like the `--web-hostname` and `--web-port` that are required for testing Google IDP locally from `-d chrome`.

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
```diff
@@ -0,0 +1,161 @@
...
+    if (!Platform.isWindows) {
+      _sigtermSub = ProcessSignal.sigterm.watch().listen(
+        (_) => process.kill(ProcessSignal.sigint),
+      );
+    }
+
+    // Forward stdin if process is spawn in --no-tui mode.
+    if (!_spawnFromTui) {
+      _stdinSub = stdin.listen((data) {
+        if (utf8.decode(data) == 'q') _restoreTerminal();
```
nielsenko: Isn't this redundant? `_cleanup()` already restores the terminal via the `exitCode` handler.

Also that `decode` is unsafe on multibyte input.

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
```diff
@@ -0,0 +1,161 @@
...
+  StreamSubscription? _stdinSub;
+  StreamSubscription? _stdoutSub;
+  StreamSubscription? _stderrSub;
+
+  bool? originalEchoMode;
+  bool? originalLineMode;
+
+  /// Matches the `flutter run` output that points to the
+  /// URL where the app is served.
+  late final _hostRegex = RegExp(r'(served)\s+(at)\s+(http:\/\/localhost:\d+)');
```
nielsenko: why 3 groups when only one (the last) is used?

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
```diff
@@ -0,0 +1,161 @@
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:serverpod_cli/src/commands/messages.dart';
+import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart';
+
+/// Manages the flutter run subprocess lifecycle.
+/// Handles spawning, signal forwarding, output streaming and graceful shutdown.
+class FlutterProcess {
```
nielsenko, line 8: Should we have a test file with more tests? Currently I think it is only `stopFlutterProcess` that is tested?

tools/serverpod_cli/lib/src/commands/start.dart
marcelomendoncasoares: I understood that this is taken care hooking on the current reload process with the `_runWithFlutterProcessReload`, right @Crazelu?

```dart
    if (_compiler == null) {
      return _chain(() => _runWithFlutterProcessReload(() => _reload(null)));
    }
    return _chain(
      () => _runWithFlutterProcessReload(
        () => _compileAndReload(
          dartFiles: const {},
          packageConfigChanged: false,
          forceFullCompile: true,
        ),
      ),
    );
```

The other watcher is for app only changes.

nielsenko (changes requested):
More findings..

- Restart is not wired up (as far as I see it)

- Where is the url displayed in the TUI? Just leaving it in the logs is perhaps not the best experience.

- If reload/restart went through the VM service instead of `stdin.write('r'/'R')`, you'd get a real future to await on, and know if it succeeded or not.

- And really, `session.handleFileChange` is the place to coordinate this, instead of spawning a new watcher.

That said, we could choose to address that in a follow-up PR. I guess the public interface is okay.

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -558,9 +588,24 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
...
     fileChangeSub = watcher.onFilesChanged
         .asyncMapBuffer((events) => session.handleFileChange(events.merge()))
         .listen((_) {});
+
+    FileWatcher? flutterWatcher;
+
+    if (launchFlutterApp) {
+      flutterWatcher = FileWatcher(
+        watchPaths: {
+          p.absolute(p.joinAll([...config.flutterPackagePathParts, 'lib'])),
```
nielsenko, line 658: `flutterPackagePathParts` be empty?

tools/serverpod_cli/lib/src/commands/start.dart
nielsenko: This hook up the key (missing similar for restart - but I guess that is waiting for another PR to land 😄 .. now that I come to think of it). 

The _flutter_ watcher only monitors the `lib` directory - not the path dependency to the generated client lib.

Crazelu:
> How will this play together with debugging from vscode?

@nielsenko 
Assuming we go with the approach of spawning a subprocess, we can update the vscode task to opt-out since it will launch the app. Hot reload from vscode already works for the server and app (not a new change).

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
Crazelu: I didn't consider connecting to the VmService. I'll check that.

> BTW, where is this called?

Hot restart is not wired yet. My idea was that this will be used when hot restart lands.

tools/serverpod_cli/lib/src/commands/start.dart
Crazelu: `_runWithFlutterProcessReload` handles forced reloads (button presses on the tui). That will reload both server and flutter app.

Changes to the client library originate in the server (which is reloaded). When you add a breaking change to a model's spy file, for instance, the flutter app picks it up. Same goes for endpoints. Perhaps that's a side effect?

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
Crazelu: Yes and yes

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
Crazelu: Should definitely be one!

tools/serverpod_cli/lib/src/commands/start.dart
Crazelu: For a module, yes.
Will account for that

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
Crazelu: It is indeed redundant

Crazelu:
@nielsenko 
> Where is the url displayed in the TUI? Just leaving it in the logs is perhaps not the best experience.

The URL is displayed in the "Log Messages" tab. Also in the flutter process's raw logs.
However, we could also just launch the browser.

Screenshot 2026-05-13 at 18 28 18

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
Crazelu: We can do this and even mirror the flutter run flags

vlidholt:
Wow, great to already have a PR for this! @Crazelu 

I think we may need to rename the tabs to make this feel more consistent. Perhaps:

**Server logs** | **Flutter logs** | **Raw output**

The URL to the Flutter app can be placed underneath the Flutter logs tab:

example

Question @nielsenko: Do we really need the raw output? I have never looked at it once.

nielsenko:
> Question @nielsenko: Do we really need the raw output? I have never looked at it once.

@vlidholt It was mostly for when we didn't have proper structured logging in place. I rarely need to go there either. Can happen if people use `print` or `stdout` in their servers.

We don't have any structured logging in the flutter app so there it is more important.

nielsenko:
**Server logs** | **Flutter logs** | **Raw output**

Is slightly misleading I think .. **Raw output** is only from the server, and **Flutter logs** are not structured like **Server logs** are. 

If we want something short.. what about
**Logs | Server output | App output**

perhaps in a different order.

nielsenko:
> > How will this play together with debugging from vscode?
> 
> @nielsenko Assuming we go with the approach of spawning a subprocess, we can update the vscode task to opt-out since it will launch the app. Hot reload from vscode already works for the server and app (not a new change).

It would be nice if we at least we leave a breadcrumb with the info people need to attach to the app. 

Extra nice if we could auto attach to existing app (like we do on the server), if/when you run "again".

nielsenko:
> @nielsenko
> 
> > Where is the url displayed in the TUI? Just leaving it in the logs is perhaps not the best experience.
> 
> The URL is displayed in the "Log Messages" tab. Also in the flutter process's raw logs. However, we could also just launch the browser.

I mean the logs will scroll by. You may want to refer to this url long after it is scrolled out of view (and possibly out of buffer). I think it would be great if we displayed it persistently on screen for as long as the app is up.

Crazelu:
> > > How will this play together with debugging from vscode?
> > 
> > 
> > @nielsenko Assuming we go with the approach of spawning a subprocess, we can update the vscode task to opt-out since it will launch the app. Hot reload from vscode already works for the server and app (not a new change).
> 
> It would be nice if we at least we leave a breadcrumb with the info people need to attach to the app.
> 
> Extra nice if we could auto attach to existing app (like we do on the server), if/when you run "again".

How I'm thinking about this:

- vscode's launch configuration contains `vmservice-out-file`
- `FlutterProcess` polls and connects to the VM Service
- When start is run again from the IDE, we write a kill sequence to the `vmservice-out-file`
- `FlutterProcess` polls every second to check for changes to the `vmservice-out-file`
- When kill sequence is detected, any running flutter subprocess that was spawn by `FlutterProcess` is shutdown
- The app launched from the IDE will write to `vmservice-out-file`
- `FlutterProcess` polls and connects to the VM Service

tools/serverpod_cli/lib/src/commands/start.dart
marcelomendoncasoares: Hmm I got it wrong, then. In this case, I also think it is better to reload the app on client changes. Because, for breaking changes, I agree that app will break -> user will fix -> save -> app reload. But non-breaking changes would require a no-op save (or active reload) to work. Some examples are adding fields with default values to models, optional parameters to endpoint methods (that change their behavior), or even some non-generated code that the user have inside the client package.

Crazelu:
@nielsenko 
Some changes since my last comment:

Using the VM Service for hot reload doesn't work well for flutter especially on the web where the compilation steps are different. It is also impossible to hot reload the flutter app through the VM Service. 

I went for the debug adapter approach instead. 
- We spawn a subprocess for `flutter debug_adapter`
- Add a proxy server and expose the port for vscode to connect
- Writes to that port are piped into the debug adapter subprocess' stdin
- A control server is added for `serverpod start` to send structured commands for hot reload and hot restart
- Those structured commands are transformed to the DAP-compliant format and piped into the debug adapter subprocess' stdin

The IDE retains all debugging capabilities and we're able to hot reload/restart from the `serverpod start` tui.

The changes I made to the tui and `WatchSession` for hot restart are for testing and illustration purposes (for the flutter app) as I'm aware there's ongoing work to support hot restart.

PR #5133, opened 2026-05-14
feat: Add FlutterWebAuth2RedirectRoute for OAuth2 PKCE web sign-in flow

vfiruz97:
Regarding @vlidholt comment in https://github.com/serverpod/serverpod_docs/pull/494#issuecomment-4441744264. This PR adds `FlutterWebAuth2RedirectRoute` for generic`OAuth2` web sign-in flow. 
Users have to register a callback router like  `pod.webServer.addRoute(FlutterWebAuth2CallbackRoute(), '/auth/callback');` before `pod.start()` instead of putting `web/auth.html` in setup Idp. This route that serves some Identity Providers callback. By default it serves in `/auth/callback` path but user can set his `path` and `host`. 

**NOTE:** This route should be in the same `host` with frontend app. 

This route is **provider-agnostic** — register it once and share it across all OAuth2 PKCE-based identity providers (Google, GitHub, Microsoft, etc.).

## Pre-launch Checklist

## Breaking changes

No breaking changes

## Same-origin requirement

This route **must** be served from the same host and port as your Flutter web application. The redirect page uses `window.postMessage` and browsers enforce that`postMessage` is only delivered when the receiving window has the same origin (scheme + host + port).

### example
Server
```dart
pod.webServer.addRoute(
  FlutterRoute(Directory(Uri(path: 'web/app').toFilePath()), host: 'cloud.serverpod.dev'),
  '/',
);
pod.webServer.addRoute(
  FlutterRoute(Directory(Uri(path: 'web/app').toFilePath()), host: 'accounts.serverpod.dev'),
  '/',
);

pod.webServer.addRoute(
  FlutterWebAuth2CallbackRoute(host: 'cloud.serverpod.dev'), // this route should be in the same-origin
  '/auth/callback',
);
```

Frontent
```dart
client.auth.initializeGoogleSignIn(
  clientId:'.apps.googleusercontent.com',
  redirectUri: 'https://cloud.serverpod.dev/auth/callback',
);

client.auth.initializeGitHubSignIn(
  clientId: '0000cc000cc',
  redirectUri: 'https://cloud.serverpod.dev/auth/callback',
);
```

## When NOT to use

If your Flutter web app is hosted separately.
In that case, place the `auth.html` file provided by (flutter_web_auth_2 README)[https://pub.dev/packages/flutter_web_auth_2#web] in your Flutter app's `web/` directory and use its URL as the `redirectUri`.

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/common/oauth2_pkce/flutter_web_auth2_callback_route.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/common/oauth2_pkce/flutter_web_auth2_callback_route.dart
@@ -2,9 +2,9 @@ import 'dart:async';
 
 import 'package:serverpod/serverpod.dart';
 
-/// {@template flutter_web_auth2_redirect_route}
+/// {@template flutter_web_auth2_callback_route}
 /// Route that serves the [flutter_web_auth_2](https://pub.dev/packages/flutter_web_auth_2)
-/// redirect page for the OAuth2 PKCE web sign-in flow.
+/// callback page for the OAuth2 PKCE web sign-in flow.
 ///
 /// This route is **provider-agnostic** — register it once and share it across
 /// all OAuth2 PKCE-based identity providers (Google, GitHub, Microsoft, etc.).
@@ -12,7 +12,7 @@ import 'package:serverpod/serverpod.dart';
 /// ## Same-origin requirement
 ///
 /// This route **must** be served from the same host and port as your Flutter
-/// web application. The redirect page uses `window.postMessage` to deliver
+/// web application. The callback page uses `window.postMessage` to deliver
 /// the OAuth result back to the Flutter app. Browsers enforce that
 /// `postMessage` with a specific `targetOrigin` is only delivered when the
 /// receiving window has the same origin (scheme + host + port).
@@ -29,7 +29,7 @@ import 'package:serverpod/serverpod.dart';
 ///
 /// ```dart
 /// // server.dart
-/// pod.configureOAuth2WebRedirectRoute(
+/// pod.configureFlutterWebAuth2CallbackRoute(
 ///   host: 'example.com',
 /// );
 ///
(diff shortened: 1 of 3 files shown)


Zfinix (approved):
LGTM 🚀

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/common/oauth2_pkce/flutter_web_auth2_callback_route.dart
```diff
@@ -0,0 +1,110 @@
...
+  void configureFlutterWebAuth2CallbackRoute({
+    final String path = '/auth/callback',
+    final String? host,
+  }) {
+    webServer.addRoute(
+      FlutterWebAuth2CallbackRoute(host: host),
+      path,
+    );
+  }
+}
```
marcelomendoncasoares: Suggestion: Remove this extension method in favor of users configuring just the `FlutterWebAuth2CallbackRoute` directly - as they do with the `FlutterRoute` or `SpaRoute`. The reason for this is to avoid introducing more ways to do the same thing, since the extension method is just a plain wrap of the `webServer.addRoute` call that users already know. The only justification to do an extension here would be if we would wrap more code, as we do with the `configureAppleIdpRoutes` for Apple IDP.

marcelomendoncasoares: We should also revisit the examples, template and all places where we add the `auth.html` file directly to replace by this route configuration.

marcelomendoncasoares (changes requested):
Just one suggestion to narrow down the interface and use it on all our examples.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/lib/src/common/oauth2_pkce/flutter_web_auth2_callback_route.dart
vfiruz97: Makes sense.

vfiruz97: done in 8693080

vfiruz97:
@Zfinix take into accont that we removed `configureFlutterWebAuth2CallbackRoute` extension method.
Now users have to register route on server side like this:

```dart
pod.webServer.addRoute(
  FlutterWebAuth2CallbackRoute(),
  '/auth/callback',
);
```

marcelomendoncasoares (approved):
LGTM!

PR #5140, opened 2026-05-15
feat: Make the `global` cache fallback to local in development/testing

### LINKED ISSUE #5138: Remove the need for docker to run the global cache on development/test
Besides eliminating the need for docker to use Postgres with #5087, users will still require docker for Redis support. Since cache is a first-class support on Serverpod, we don't want to remove it from the `create` templates or `quickstart` command.

A better alternative is to fallback to a local cache if running on development/testing run modes. This way, we can completely skip using docker, but still have the cache and database. The fallback also still allow users to run with Redis if they start an instance locally with docker.

marcelomendoncasoares:
Closes: #5138

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod/lib/src/redis/controller.dart
+++ b/packages/serverpod/lib/src/redis/controller.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:io';
 
 import 'package:redis/redis.dart';
 import 'package:serverpod_shared/log.dart';
@@ -28,6 +29,13 @@ class RedisController {
   /// require ssl
   final bool requireSsl;
 
+  /// Maximum time to wait while opening the TCP/TLS connection to Redis.
+  ///
+  /// After this duration, the attempt fails with [TimeoutException] instead of
+  /// relying on OS-level connect timeouts (which can be very long when traffic
+  /// is dropped).
+  final Duration connectTimeout;
+
   final Map<String, RedisSubscriptionCallback> _subscriptions = {};
 
   Command? _command;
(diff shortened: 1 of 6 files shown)


packages/serverpod/lib/src/redis/controller.dart
```diff
@@ -46,13 +54,23 @@ class RedisController {
...
   /// Starts the controller and connects to Redis. Maintains an open connection
   /// until [stop] is called.
-  Future<void> start() async {
-    await _connect();
-    await _connectPubSub();
+  ///
+  /// If [connectTimeout] is set, it overrides [RedisController.connectTimeout]
+  /// for the TCP/TLS connections opened during this [start] call only.
+  Future<void> start({
+    bool Function(Exception e)? handleError,
```
nielsenko, line 66: Do we really need the flexibility to compose errorhandling at runtime? We only use it for a bit of testing. Otherwise we always use the same callback.

nielsenko (approved):
This is okay, but I'm not super big fan of adding error handling callback to `start` (and friends).

packages/serverpod/lib/src/redis/controller.dart
marcelomendoncasoares: As discussed, this change was mostly to avoid having to rewrite the entire connection lifecycle handling, that currently swallows errors to avoid breaking the server when the connection drops. We can refactor in the future to make only the `start` bubble up the connection issues.

PR #5146, opened 2026-05-19
fix: Fix `copyWith` method on `dynamic` fields

### LINKED ISSUE #5145: Calling `copyWith` on a model with `dynamic` fields lead to values being replaced by `_Undefined`
### Describe the bug

When copying a model that has `dynamic` fields without changing their value, such fields are having their value replaced by the sentinel `_Undefined` type.

### To reproduce

Create a model with a `dynamic` field and call `copyWith` without changing the `dynamic` field value.

### Expected behavior

Only the changed fields should be changed.

### Library version

3.5.0 (unpublished, main)

### Platform information

N/A

### Additional context

The root cause is the comparison with the sentinel being done with `is! _Undefined` instead of `!= _Undefined` on the generated models. This happen only with `dynamic` fields because they are the only ones comparing with the `_Undefined` sentinel directly. Other fields use a `is! <FieldType>` comparison.

- [x] I want to work on this

marcelomendoncasoares:
Fixes: #5145

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tests/serverpod_test_server/lib/src/generated/object_with_dynamic.dart
+++ b/tests/serverpod_test_server/lib/src/generated/object_with_dynamic.dart
@@ -208,8 +208,8 @@ class _ObjectWithDynamicImpl extends ObjectWithDynamic {
   }) {
     return ObjectWithDynamic(
       id: id is int? ? id : this.id,
-      payload: payload is! _Undefined ? payload : this.payload,
-      jsonbPayload: jsonbPayload is! _Undefined
+      payload: payload != _Undefined ? payload : this.payload,
+      jsonbPayload: jsonbPayload != _Undefined
           ? jsonbPayload
           : this.jsonbPayload,
       payloadList: payloadList ?? this.payloadList.map((e0) => e0).toList(),
--- a/tests/serverpod_test_client/lib/src/protocol/object_with_dynamic.dart
+++ b/tests/serverpod_test_client/lib/src/protocol/object_with_dynamic.dart
@@ -153,8 +153,8 @@ class _ObjectWithDynamicImpl extends ObjectWithDynamic {
   }) {
     return ObjectWithDynamic(
       id: id is int? ? id : this.id,
-      payload: payload is! _Undefined ? payload : this.payload,
-      jsonbPayload: jsonbPayload is! _Undefined
+      payload: payload != _Undefined ? payload : this.payload,
+      jsonbPayload: jsonbPayload != _Undefined
           ? jsonbPayload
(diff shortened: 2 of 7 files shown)


tests/serverpod_test_server/lib/src/generated/object_with_dynamic.dart
```diff
@@ -208,8 +208,8 @@ class _ObjectWithDynamicImpl extends ObjectWithDynamic {
   }) {
     return ObjectWithDynamic(
       id: id is int? ? id : this.id,
-      payload: payload is! _Undefined ? payload : this.payload,
-      jsonbPayload: jsonbPayload is! _Undefined
+      payload: payload != _Undefined ? payload : this.payload,
```
nielsenko, line 211: This looks wrong to my eye? `_Undefined` is an instance of `Type`. Doing compare with `!=` you compare the `payload` instance  to an instance of `Type`. This (unexpectedly) works (since `Type` instance are implicitly `const`, and are also just `Object`s, so an _undefined_ payload actually point to the same `Type` instance, but a `const _Undefined()` would be what I would expect here. 

Preferably stored in a `const` global in the file:
```dart
const _undefined = _Undefined();

//
payload: payload != _undefined ? payload : this.payload;
```

nielsenko (commented):
Test tells me this works, and I understand why, but it feels wrong to use a `Type` instance pointing to `_Undefined` instead of a const instance og `_Undefined` as the marker.

Could we change it? If not, why not?

tests/serverpod_test_server/lib/src/generated/object_with_dynamic.dart
nielsenko: Obviously would need to use the same in the generated `copyWith` default.

marcelomendoncasoares: I wouldn't say that it is unexpected to work because this is just using identity comparison on a `dynamic` variable - that can assume a `Type` value as well. But I agree that it would be clearer to have the comparison against an `const _Undefined()`. I just didn't want to couple such change on this fix because it will touch every model file that has a nullable field and explode to several thousand lines in diff.

@coderabbitai Create an issue to replace the `_Undefined` sentinel comparison on `copyWith` methods to use a `const _undefined = _Undefined()` instead.

nielsenko: I'm fine with deferring

nielsenko (approved):
👍

PR #5148, opened 2026-05-19
refactor: Avoid connect tool and friends

### LINKED ISSUE #5069: `serverpod start` fails silently if there is a server already running
If there is a server running on the same port, `serverpod start` will silently fail without any error message.

### LINKED ISSUE #5111: Improve connecting to MCP server
Currently, to connect to the MCP server, the agent needs to:

- Read the description of the MCP server.
- Understand that it needs to get a list of running servers.
- Connect to the running server.
- Get a new list of available tools.

This is hard for agents to understand and also consumes unnecessary tokens, bloats the context. A better solution would be to allow it to access the MCP server directly without the steps described above.

A possible solution would be to pass the project name to the MCP server through the configuration. E.g., start the server with:

```
serverpod mcp-server my_project
```

Then, we can remove the `connect` commands. Instead, the MCP server would contain all available tools directly on startup.  When the agent attempts to use a tool (e.g., `create_migration`), the MCP server would attempt to connect to `serverpod start`. If successful, it carries out the command. If not, it responds with a message similar to:

> The server is currently not running. Please inform the user to run the `serverpod start` command.

Removing the connection commands also has the added benefit of further reducing the context (fewer commands for the agent to process).

For this to work, the `serverpod create` command needs to also include the project name in the MCP configuration.

nielsenko:
Some (most?) agents/harnesses are not very good at listening for tool changes, as pods come and go.

This PR makes every tool discoverable before pod is started with `serverpod start` and avoid the need for `connect`/`disconnect` tool by having `serverpod mcp` connecting to the current project.

Current project is either inferred from CWD, or specified via `--server-dir` argument.

Includes a lot of simplifications regarding monitoring for runners (ie. `serverpod start` processes) that come and goes.

To handle the (arguably rare) situation where an agent need to orchestrate multiple pods at the same time (multi-server mono-repos) you now need to configure the client to call `serverpod mcp` for each project. 

Closes: #5111 
Fixes: #5069


tools/serverpod_cli/lib/src/mcp/runner_surface.dart
```diff
@@ -0,0 +1,77 @@
...
+final Tool applyMigrationsTool = Tool(
+  name: 'apply_migrations',
+  description:
+      'Apply pending database migrations. The server restarts with '
+      '`--apply-migrations`. Call after `create_migration` (or '
+      '`serverpod create-migration`) has written the migration files.',
+  inputSchema: Schema.object(),
+);
+
+final Tool createMigrationTool = Tool(
```
marcelomendoncasoares, line 14: Suggestion(NIT): Remove the redundant type after `final`.

tools/serverpod_cli/lib/src/mcp/bridge_mcp_server.dart
```diff
@@ -5,337 +5,176 @@ import 'dart:io';
...
-            'Bridge to a running `serverpod start --watch` instance. Read '
-            'serverpod://instances to discover running instances, then call '
-            'the `connect` tool to attach to one. Tools and resources from '
-            'the connected instance are forwarded over its MCP socket and '
-            'appear here automatically.',
+            'Bridge to the `serverpod start --watch` runner for this '
+            'project. Tools and resources auto-connect to the runner on '
+            'demand. If the runner is not currently running, calls return '
+            'a message asking the user to start it with '
+            '`serverpod start --watch`.',
```
marcelomendoncasoares: Suggestion: Remove references to the `--watch` flag and simplify the explanation of what happens when the server is not available.

```suggestion
            'Bridge to the server running with `serverpod start` for this '
            'project. Tools and resources auto-connect to the runner on '
            'demand. If the runner is not currently available, ask the '
            'user to start it with `serverpod start`.',
```

tools/serverpod_cli/lib/src/mcp/bridge_mcp_server.dart
```diff
@@ -5,337 +5,176 @@ import 'dart:io';
...
+  CallToolResult _notRunningError() => CallToolResult(
     content: [
       TextContent(
         text:
-            'Not connected to a serverpod instance. Read serverpod://instances '
-            'and call the `connect` tool first, or call `spawn` to start a '
-            'new `serverpod start --watch` process.',
+            'The `serverpod start --watch` runner for this project is not '
+            'running. Please ask the user to run `serverpod start --watch` '
+            'in the project directory.',
```
marcelomendoncasoares:
```suggestion
            'The server is not running with `serverpod start` for this project. '
            'Ask the user to run `serverpod start` in the project directory.',
```

tools/serverpod_cli/lib/src/mcp/runner_surface.dart
```diff
@@ -0,0 +1,77 @@
...
+library;
+
+import 'package:dart_mcp/server.dart';
+
+final Tool applyMigrationsTool = Tool(
+  name: 'apply_migrations',
+  description:
+      'Apply pending database migrations. The server restarts with '
+      '`--apply-migrations`. Call after `create_migration` (or '
+      '`serverpod create-migration`) has written the migration files.',
```
marcelomendoncasoares: Suggestion: Update the description to the last migration apply strategy. Also remove the reference to the previous command, since the instruction to call `apply_migrations` was already present there.

```suggestion
      'Apply pending database migrations without restarting the server. '
      'Call after creating a migration.',
```

tools/serverpod_cli/lib/src/mcp/runner_surface.dart
```diff
@@ -0,0 +1,77 @@
...
+  inputSchema: Schema.object(),
+);
+
+final Tool createMigrationTool = Tool(
+  name: 'create_migration',
+  description:
+      'Create a new database migration from the current model definitions. '
+      'Writes the migration files to disk; does not restart the server or '
+      'apply the migration. Follow up with `apply_migrations` to run it '
+      'against the database.',
```
marcelomendoncasoares: Suggestion: Remove the reference to restart the server, since neither the apply does it anymore.

```suggestion
  description:
      'Create a new database migration from the current model definitions. '
      'Only writes the migration files to disk, without applying to the '
      'database. Follow up with `apply_migrations` to apply the changes.',
```

tools/serverpod_cli/lib/src/mcp/runner_surface.dart
```diff
@@ -0,0 +1,77 @@
...
+  inputSchema: Schema.object(
+    properties: {
+      'tag': Schema.string(
+        description: 'Optional tag appended to the migration version name.',
+      ),
+      'force': Schema.bool(
+        description:
+            'Create the migration even if warnings are present (data may '
+            'be destroyed).',
```
marcelomendoncasoares:
```suggestion
        description:
            'Create the migration even if warnings are present (data may '
            'be destroyed). Required for destructive migrations.',
```

tools/serverpod_cli/lib/src/mcp/runner_surface.dart
```diff
@@ -0,0 +1,77 @@
...
+  ),
+);
+
+final Tool hotReloadTool = Tool(
+  name: 'hot_reload',
+  description:
+      'Recompile the server kernel and hot-reload the running isolate. '
+      'Falls back to a full restart if reload is not possible.',
+  inputSchema: Schema.object(),
+);
```
marcelomendoncasoares, line 51: Suggestion: Rename to `hot_restart` since we'll now have `--watch` as mandatory.

tools/serverpod_cli/lib/src/mcp/runner_surface.dart
```diff
@@ -0,0 +1,77 @@
...
+
+final Resource vmServiceResource = Resource(
+  uri: 'serverpod://vm-service',
+  name: 'VM service',
+  description:
+      'Dart VM service HTTP URI for the running server isolate. Stable '
+      'across hot reloads; changes on restart (apply_migrations, crash '
+      'recovery). Subscribe to be notified when the URI changes.',
+  mimeType: 'application/json',
+);
```
marcelomendoncasoares, line 77: Question: Do we still need this resource on the MCP?

marcelomendoncasoares (changes requested):
Only some comments for improvement on the MCP command descriptions.

marcelomendoncasoares (approved):
CI still finishing, but LGTM!

PR #5153, opened 2026-05-19
fix: Fix `dynamic` data being serialized as an encoded string

### LINKED ISSUE #5152: Data on `dynamic` fields should be serialized as `Map` instead of `String`
### Describe the bug

While writing the docs for the `dynamic` type feature, I realized the implementation had a flaw: the data is being **encoded** instead of **serialized** before the actual encoding. This result `dynamic` data being transferred and stored as strings - despite the database column being a JSON/JSONB.

### To reproduce

Create an object with a `dynamic` field and covert it `toJson()`. Notice the `dynamic` field as a `String` with the shape of the `wrapWithClassName` output:

```dart

var object = ObjectWithDynamic(
  payload: 42,
  ...
);

final serialized = object.toJson();
// serialized['payload'] => '{"className":"int","data":42}'
```

### Expected behavior

The data should be serialized as the `Map` itself, not an encoded version of it.

### Library version

3.5.0 (unreleased, main).

### Platform information

N/A

### Additional context

_No response_

- [x] I want to work on this

marcelomendoncasoares:
fixes: #5152

## Pre-launch Checklist

## Breaking changes

This PR is a breaking change since the introduction of the `dynamic` type in `3.5.0-beta.5`, but it is not an issue since it was not released publicly yet.

### DIFF
--- a/packages/serverpod_serialization/lib/src/serialization.dart
+++ b/packages/serverpod_serialization/lib/src/serialization.dart
@@ -65,7 +65,7 @@ abstract class SerializationManager {
     if (value is String) return decodeWithType(value);
     if (value is Map<String, dynamic>) return deserializeByClassName(value);
     throw FormatException(
-      'Dynamic fields are serialized as a Map with the type, but got '
+      'Dynamic fields are encoded as a Map with the type, but got '
       '${value.runtimeType} instead.',
     );
   }
--- a/packages/serverpod_serialization/lib/src/serialization.dart
+++ b/packages/serverpod_serialization/lib/src/serialization.dart
@@ -57,19 +57,6 @@ abstract class SerializationManager {
     return deserializeByClassName(jsonDecode(data));
   }
 
-  /// Decodes a value for a `dynamic` model field from protocol or database
-  /// shapes: a JSON [String] from [encodeWithType], or a [Map] when JSON was
-  /// parsed before deserialization.
-  dynamic decodeDynamicFieldValue(Object? value) {
-    if (value == null) return null;
-    if (value is String) return decodeWithType(value);
-    if (value is Map<String, dynamic>) return deserializeByClassName(value);
-    throw FormatException(
-      'Dynamic fields are encoded as a Map with the type, but got '
-      '${value.runtimeType} instead.',
-    );
-  }
-
   bool _isType<T1, T2>(Type t) => t == T1 || t == T2;
 
   bool _isNullableType<T>(Type t) => _isType<T, T?>(t);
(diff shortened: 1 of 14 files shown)


packages/serverpod_serialization/lib/src/serialization.dart
```diff
@@ -164,41 +151,68 @@ abstract class SerializationManager {
...
+        return raw.map((k, v) => MapEntry(k, deserializeDynamicFieldValue(v)));
+      case 'Map' when raw is List<Map<String, dynamic>>:
+        return Map<dynamic, dynamic>.fromEntries(
+          raw.map(
+            (e) => MapEntry(
+              deserializeDynamicFieldValue(e['k']),
+              deserializeDynamicFieldValue(e['v']),
+            ),
+          ),
+        );
```
nielsenko, line 200: Should this not be:
```dart
case 'Map' when raw is List: // <- after jsonEcode then jsonDecode this will be List<dynamic> at runtime 
    return Map<dynamic, dynamic>.fromEntries(
      raw.cast<Map<String, dynamic>>().map((e) => MapEntry(
        deserializeDynamicFieldValue(e['k']),
        deserializeDynamicFieldValue(e['v']),
      )),
    );
```

You should have a test that roundtrip over encode/decode as well. The current roundtrip test short-cut.

tests/serverpod_test_server/test/dynamic_roundtrip_test.dart
nielsenko, line 232: Roundtrip tests that shortcut like this can hide subtle bugs.
I think you should `jsonEncode`/`jsonDecode` as well.

packages/serverpod_serialization/lib/src/serialization.dart
```diff
@@ -327,6 +341,57 @@ abstract class SerializationManager {
...
+            _dynamicFieldValueToJson(e, encodeForProtocol),
+        ],
+      },
+      Map() when object.keys.every((k) => k is String) => {
+        'className': 'Map',
+        'data': {
+          for (final e in Map<String, dynamic>.from(object).entries)
+            e.key: _dynamicFieldValueToJson(e.value, encodeForProtocol),
+        },
+      },
```
nielsenko:
```suggestion
      Map<String, dynamic>() => {
        'className': 'Map',
        'data': {
          for (final e in object.entries)
            e.key: _dynamicFieldValueToJson(e.value, encodeForProtocol),
        },
      },
      Map() when object.keys.every((k) => k is String) => { // we know key is String now
        'className': 'Map',
        'data': {
          for (final e in object.entries)
            e.key as String: _dynamicFieldValueToJson(e.value, encodeForProtocol),
        },
      },
```
Fast path when type match (no expensive `when` clause) and avoids an extra `Map` allocation on the expensive path.

nielsenko (changes requested):
I think I spotted a bug on the roundtrip..

packages/serverpod_serialization/lib/src/serialization.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5153/commits/49c4654b1beabd9654b0b4f62ba9b1999e07fbfb.

tests/serverpod_test_server/test/dynamic_roundtrip_test.dart
marcelomendoncasoares: Tests added at https://github.com/serverpod/serverpod/pull/5153/commits/0259da2088e6bbdb0a42c9138d6604a16e480c4d.

packages/serverpod_serialization/lib/src/serialization.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5153/commits/984ff6c2e9ffb45c104988333405997d436f5ba3.

nielsenko (approved):
LGTM now 👍 

Not green yet, but looks like a github runner issue.

PR #5155, opened 2026-05-19
refactor: Apply migrations from CLI through a dedicated endpoint

### LINKED ISSUE #5081: Refactor the apply migrations from CLI to use a dedicated endpoint from the server
The current implementation of applying migrations from the CLI (the hidden `migrate` command) has bundled the database verification to ensure that the `serverpod start` sees inconsistencies during development. But this is not ideal, because the command can no longer be used detached from the server. Instead, we should:

- Create a new endpoint method that allows requesting a migration to the server externally. This will be the endpoint used by `serverpod start` after the server has first started with `--apply-migrations`.
- Remove the database integrity verification from the command to apply migrations from the CLI. The database will be allowed to be inconsistent and the server should verify on the next start - given that this command aims to migrate the database with the server offline.
- Completely decouple the required endpoints for managing the database/migrations from the Insights endpoint and make it an official and persistent endpoint class that can't be disabled. This will prevent issues when trying to run repair migration with Insights disabled.

### LINKED ISSUE #5168: Calling `apply_migrations` on the TUI and MCP should also apply repair migrations
Currently, hitting `A` on the TUI or calling `apply_migrations` through the MCP will only apply normal migrations. But #5089 introduces the ability for the user to create repair migrations as well - specially for rollback - so they must also be taken into consideration.

FXschwartz:
Refactors how the CLI applies database migrations.

- New `InsightsEndpoint.applyMigrations` method mirroring the pod's internal `_applyMigrations`. Returns a new `MigrationApplyResult` model. Does not call `verifyDatabaseIntegrity`

- `serverpod start`'s watch loop now applies migrations by calling this endpoint on the running pod, rather than running them in-process against the database.

- The CLI migration runner is no longer dependent on a running pod for loading target table definitions and no longer does the integrity check. 

Note: After discussing it with @marcelomendoncasoares I simplified the implementation and added the endpoint to the `InsightsEndpoint` instead of decoupling it and making it standalone.

Fixes #5081 
Closes #5168

## Pre-launch Checklist

## Breaking changes

None


tools/serverpod_cli/lib/src/migrations/cli_migration_runner.dart
```diff
@@ -212,13 +163,9 @@ T _withServerDir<T>({
 /// verification runs later using target table definitions loaded over
 /// the insights endpoint.
```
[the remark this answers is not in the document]
FXschwartz: Removed verification lines entirely as it no longer has anything to do with this class.

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -305,26 +307,49 @@ List<String> _withApplyMigrations(List<String> serverArgs) {
...
+      applyMigrations: true,
     );
-    log.info(formatAppliedMigrations(applied));
+    log.info(formatAppliedMigrations(result.migrationsApplied ?? const []));
     return const MigrationsApplied();
   } catch (e) {
-    if (!isMissingNativeAssetError(e)) rethrow;
+    log.warning('Failed to apply migrations through the running pod: $e');
     onDeferToPod();
     return const MigrationsRequirePodRestart();
```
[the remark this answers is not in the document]
FXschwartz: Fixed

tests/serverpod_test_server/test_integration/migration_manager/apply_migrations_endpoint_test.dart
```diff
@@ -0,0 +1,48 @@
...
+import 'package:test/test.dart';
+
+import '../test_tools/serverpod_test_tools.dart';
+
+void main() {
+  withServerpod(
+    rollbackDatabase: RollbackDatabase.disabled,
+    testGroupTagsOverride: [TestTags.concurrencyOneTestTag],
+    'Given a database already at the latest migration version',
+    (sessionBuilder, _) async {
```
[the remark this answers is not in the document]
FXschwartz: This is no longer correct and not a needed change

packages/serverpod/lib/src/endpoints/insights.dart
```diff
@@ -231,6 +231,41 @@ class InsightsEndpoint extends Endpoint {
...
+  /// is allowed to be in an inconsistent state until the pod's next start,
+  /// at which point integrity is re-checked.
+  ///
+  /// Used by `serverpod start`'s watch loop to apply newly generated
+  /// migrations without restarting the pod.
+  Future<MigrationApplyResult> applyMigrations(
+    Session session, {
+    required bool applyRepairMigration,
+    required bool applyMigrations,
+  }) async {
```
marcelomendoncasoares, line 249: Actually, this endpoint is the method that will call the `MigrationManager.verifyDatabaseIntegrity`. The one that must skip it is applying migrations from the CLI, since the CLI should do it connecting directly to the database and not through a running server.

marcelomendoncasoares: We should also abstract the entire migration execution logic away from `Serverpod` to reuse there and here:

https://github.com/serverpod/serverpod/blob/f20a029a509f0aa7ad896497cba80c9a221741c7/packages/serverpod/lib/src/server/serverpod.dart#L944-L978

packages/serverpod/lib/src/models/migration_apply_result.spy.yaml
```diff
@@ -0,0 +1,12 @@
+### Result of applying database migrations via the Insights endpoint.
+class: MigrationApplyResult
```
marcelomendoncasoares: Suggestion(NIT): Use on plural, since we apply multiple migrations.

```suggestion
class: MigrationsApplyResult
```

packages/serverpod/lib/src/endpoints/insights.dart
```diff
@@ -231,6 +231,41 @@ class InsightsEndpoint extends Endpoint {
...
 
+  /// Applies pending database migrations to the running pod, mirroring the
+  /// boot-time path triggered by `--apply-migrations` and
+  /// `--apply-repair-migration`. Unlike that path, this method intentionally
+  /// does NOT call [MigrationManager.verifyDatabaseIntegrity] — the database
+  /// is allowed to be in an inconsistent state until the pod's next start,
+  /// at which point integrity is re-checked.
+  ///
+  /// Used by `serverpod start`'s watch loop to apply newly generated
+  /// migrations without restarting the pod.
```
marcelomendoncasoares, line 244: Suggestion: Also document that calling this endpoint expects pending/repair migrations to be available on the `migrations` folder.

packages/serverpod/lib/src/models/migration_apply_result.spy.yaml
```diff
@@ -0,0 +1,12 @@
...
+fields:
+  ### Version name of the repair migration that was applied, or `null` if
+  ### `applyRepairMigration` was false, no repair migration file exists, or
+  ### the repair migration was already applied to this database.
+  repairMigrationApplied: String?
+
+  ### Versions of regular migrations that were applied, or `null` if
+  ### `applyMigrations` was false or the database was already at the latest
+  ### version.
+  migrationsApplied: List<String>?
```
marcelomendoncasoares: Suggestion(NIT): Put the `migrationsApplied` first, since it is the mostly used.

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -305,26 +307,49 @@ List<String> _withApplyMigrations(List<String> serverArgs) {
...
+  required String runMode,
+}) {
+  final originalCwd = Directory.current;
+  Directory.current = Directory(serverDir);
+  try {
+    return ConfigInfo(runMode).createServiceClient();
+  } finally {
+    Directory.current = originalCwd;
+  }
+}
```
marcelomendoncasoares: Question: Can we adapt and reuse the `_loadServerpodConfig` from the CLI migration runner?

https://github.com/serverpod/serverpod/blob/f20a029a509f0aa7ad896497cba80c9a221741c7/tools/serverpod_cli/lib/src/migrations/cli_migration_runner.dart#L180-L194

Then we avoid reimplementing this logic to create the service client in multiple places.

A note on the style, it is better to put auxiliary private classes like this at the bottom (clean code convention).

tests/serverpod_test_server/test_integration/migration_manager/apply_migrations_endpoint_test.dart
```diff
@@ -0,0 +1,48 @@
...
+      final endpoint = InsightsEndpoint();
+
+      test(
+        'when applyMigrations is called requesting regular migrations, '
+        'then no migrations are reported as applied',
+        () async {
```
marcelomendoncasoares, line 18: Suggestion: We must also add migration e2e tests that validate migrations being applied (with integrity verification) when there are new/repair migrations.

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -305,26 +307,49 @@ List<String> _withApplyMigrations(List<String> serverArgs) {
...
+    final result = await client.insights.applyMigrations(
+      applyRepairMigration: false,
+      applyMigrations: true,
     );
-    log.info(formatAppliedMigrations(applied));
+    log.info(formatAppliedMigrations(result.migrationsApplied ?? const []));
     return const MigrationsApplied();
-  } catch (e) {
-    if (!isMissingNativeAssetError(e)) rethrow;
+  } on SocketException catch (e) {
```
marcelomendoncasoares: Suggestion: Will this really pickup connection issues? When using the `Client`, `SocketException` will be replaced by a `ServerpodClientException`:

https://github.com/serverpod/serverpod/blob/f20a029a509f0aa7ad896497cba80c9a221741c7/packages/serverpod_client/lib/src/serverpod_client_io.dart#L77-L78

But, since we'll no longer fallback to `onDeferToPod`, maybe we can remove the try/catch entirely and let the exception bubble up.

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -305,26 +307,49 @@ List<String> _withApplyMigrations(List<String> serverArgs) {
...
+    log.warning('Could not reach the pod to apply migrations: $e');
     onDeferToPod();
     return const MigrationsRequirePodRestart();
```
marcelomendoncasoares: Suggestion: Now that we apply migrations from inside the `pod`, we no longer require the `onDeferToPod` fallback mechanism. Right, @nielsenko?

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -305,26 +307,49 @@ List<String> _withApplyMigrations(List<String> serverArgs) {
...
+  Client? client;
   try {
-    final applied = await applyPendingMigrations(
-      serverDir: serverDir,
-      runMode: runMode,
-      moduleName: moduleName,
+    client = _buildServiceClient(serverDir: serverDir, runMode: runMode);
+    final result = await client.insights.applyMigrations(
+      applyRepairMigration: false,
+      applyMigrations: true,
```
marcelomendoncasoares: Question: You used `applyRepairMigration: false` to mirror the same behavior as previously, right? But, now that we expose repair migrations on the TUI, we'll also need to account for applying repair migrations. Otherwise, users will hit `P` to create the migration, `A` to apply, but it will do nothing.

marcelomendoncasoares (changes requested):
Nice work! A few comments to improve the solution. The main one is to actually verify the integrity from the endpoint, since it will run with access to the `tableDefinitions`. On `serverpod start`, it will have hot-reloaded before applying migrations, so the definitions will be up-to-date and need to match the applied migrations.

packages/serverpod/lib/src/endpoints/insights.dart
FXschwartz: @marcelomendoncasoares Ahh my bad on this. I got it backwards. Fixing now!

packages/serverpod/lib/src/models/migration_apply_result.spy.yaml
FXschwartz: Done!

packages/serverpod/lib/src/endpoints/insights.dart
FXschwartz: Done!

tests/serverpod_test_server/test_integration/migration_manager/apply_migrations_endpoint_test.dart
FXschwartz: ~Done!~

tools/serverpod_cli/lib/src/commands/start.dart
FXschwartz: Removed, but will add back pending his reply.

packages/serverpod/lib/src/models/migration_apply_result.spy.yaml
FXschwartz: Done!

tools/serverpod_cli/lib/src/commands/start.dart
FXschwartz: Yeah you are right. I removed the `SocketException` and `catch` but kepts the `try/finally` to close the client.

tools/serverpod_cli/lib/src/commands/start.dart
FXschwartz: I went ahead and made `_withServerDir` public `withServerDir` since that is really the only piece of duplicated logic. I deleted `_buildServiceClient` and now in `start.dart` it just calls `withServerDir`.

Let me know if you are not ok with it no longer being private or have a different solution in mind.

tools/serverpod_cli/lib/src/commands/start.dart
FXschwartz: Ah yeah you are right, I missed that. Change it to `true`. If one exists and hasn't been applied, it runs it. If not, returns `repairMigrationApplied: null`

packages/serverpod/lib/src/endpoints/insights.dart
FXschwartz: Ok the new endpoint now verifies, and I created a new file `apply_migrations.dart` that contains a shared function `applyMigrationsAndVerify` that gets used in both `serverpod.dart` `_applyMigrations` and `insights.dart` `applyMigrations`

tests/serverpod_test_server/test_integration/migration_manager/apply_migrations_endpoint_test.dart
FXschwartz: @marcelomendoncasoares Well I spoke too soon on this one. So I think the endpoint test is failing because it's reading from the server container's filesystem but our test is creating files in the test containers filesystem. So `migrateToLatest` is always returning `null`. 

It looks like the CLI test works because the CLI runner runs in the test container itself, not the server.

What if I add an e2e test for the verification path by making the shared db drift via `executeSql`. Then move the coverage of the actual new migration via endpoint over to integration tests where it can test that the endpoint handles applying migrations and verifies.

This effectively splits the test coverage you are asking for, so feel free to push back.

tools/serverpod_cli/lib/src/commands/start.dart
nielsenko: @FXschwartz & @marcelomendoncasoares I agree, we can clean this up.

packages/serverpod/lib/src/endpoints/insights.dart
```diff
@@ -231,6 +232,31 @@ class InsightsEndpoint extends Endpoint {
...
+  }) async {
+    return applyMigrationsAndVerify(
+      session: session,
+      projectDirectory: Directory.current,
+      runMode: session.serverpod.runMode,
+      applyRepairMigration: applyRepairMigration,
+      applyMigrations: applyMigrations,
+    );
+  }
+
```
nielsenko: Insights is already admin-grade today, but applying migrations is a new whole new ballgame. Should we gate this on `runMode == development`? @marcelomendoncasoares?

packages/serverpod_database/lib/src/migrations/apply_migrations.dart
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  final databaseMatchesTargetState =
+      await MigrationManager.verifyDatabaseIntegrity(session);
+
+  return MigrationsApplyResult(
+    migrationsApplied: migrationsApplied,
+    repairMigrationApplied: repairMigrationApplied,
+    databaseMatchesTargetState: databaseMatchesTargetState,
+  );
```
nielsenko, line 63: Should we distinguish between `null` and `[]` on results? So `[]` is _tried, nothing pending_ and `null`  is _didn't try_?

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -305,26 +306,26 @@ List<String> _withApplyMigrations(List<String> serverArgs) {
...
-    final applied = await applyPendingMigrations(
-      serverDir: serverDir,
-      runMode: runMode,
-      moduleName: moduleName,
+    final result = await client.insights.applyMigrations(
+      applyRepairMigration: true,
+      applyMigrations: true,
     );
-    log.info(formatAppliedMigrations(applied));
+    log.info(formatAppliedMigrations(result.migrationsApplied ?? const []));
```
nielsenko: Shouldn't we do something with `result.databaseMatchesTargetState` here? Isn't it the whole point of moving verification to the pod that we can report drift?

nielsenko: .. maybe a `formatMigrationApplyResult`?

nielsenko: And perhaps avoid the "Applied 0 migrations: " blemish in the same go 😄

nielsenko: Also, what about `result.repairMigrationApplied`? We should surface that as well.

packages/serverpod/lib/src/endpoints/insights.dart
nielsenko: Also, should we _pause_ the pods handling of requests while `applyMigrationsAndVerify` is ongoing? And not start until all pending requests are settled?

packages/serverpod_database/lib/src/migrations/apply_migrations.dart
```diff
@@ -0,0 +1,49 @@
...
+/// Verification failure is reported via
+/// [MigrationsApplyResult.databaseMatchesTargetState] rather than thrown —
+/// callers decide policy.
+Future<MigrationsApplyResult> applyMigrationsAndVerify({
+  required DatabaseSession session,
+  required Directory projectDirectory,
+  required String runMode,
+  required bool applyRepairMigration,
+  required bool applyMigrations,
+}) async {
```
nielsenko, line 22: Should we exit early on double `false`? Feels a bit weird that you can _verify_ only by passing double `false` 🤔 .. or perhaps document it.

nielsenko (commented):
I'll let @marcelomendoncasoares decide if his feedback has been addressed, but it seems so to me.

I do have some comments/suggestions/questions of my own

packages/serverpod/lib/src/endpoints/insights.dart
marcelomendoncasoares: > Insights is already admin-grade today, but applying migrations is a new whole new ballgame. Should we gate this on `runMode == development`? @marcelomendoncasoares?

Considering that migrations need to be available on the running machine, calling it on production is at most harmless. I'd also argue that nothing is worst than `executeSql` 😅 

> Also, should we _pause_ the pods handling of requests while `applyMigrationsAndVerify` is ongoing? And not start until all pending requests are settled?

Agreed, this is a good hygiene on the process. It will provide the same safety that we expect when restarting. Ideally, we should only be careful to keep health/healthz endpoints alive for correctness. But it is fine to spin this to a separate issue, since I don't think we have this granular level of control on the serving.

packages/serverpod_database/lib/src/migrations/apply_migrations.dart
marcelomendoncasoares: I think it is fine, if we document that calling this method will always verify the database, regardless of applying migrations. This will simplify the usage of this method on the start of the `Serverpod` class, where we always verify the integrity.

packages/serverpod_database/lib/src/migrations/apply_migrations.dart
marcelomendoncasoares: Good suggestion.

tools/serverpod_cli/lib/src/commands/start.dart
marcelomendoncasoares: > Shouldn't we do something with `result.databaseMatchesTargetState` here? Isn't it the whole point of moving verification to the pod that we can report drift?

> .. maybe a `formatMigrationApplyResult`?

> And perhaps avoid the "Applied 0 migrations: " blemish in the same go 😄

Yes, we should use it to report that migrations were not applied. Combined with your suggestion to return `null` if there was nothing to apply, we should be able to no longer report "Applied 0 migrations".

Given that the pod itself will log the drift message, maybe only a generic "Failed to apply existing migrations" on the `formatAppliedMigrations` (to be replaced by `formatMigrationApplyResult`) is enough, right @nielsenko?

tools/serverpod_cli/lib/src/commands/start.dart
marcelomendoncasoares: I'd suggest creating a `static` method on the `ConfigInfo` to load using the `_withServerDir` approach (keeping it private). Then, we can remove both the `_loadServerpodConfig` and `_buildServiceClient` in favor of:

```dart
ConfigInfo.fromDir(serverDir, runMode).createServiceClient();
ConfigInfo.fromDir(serverDir, runMode).config;
```

packages/serverpod/lib/src/endpoints/insights.dart
FXschwartz: @marcelomendoncasoares Just to be clear should @nielsenko entire suggestion about pausing the pod be punted to a separate issue? Or are you only talking about punting the extra level of control by keeping only health endpoints alive?

marcelomendoncasoares: My suggestion is to push only keeping the health endpoints alive to a separate issue. Stopping the pod for now would be a good improvement to land on this PR.

tools/serverpod_cli/lib/src/commands/start.dart
nielsenko: I hadn't thought about that - but yes. The pod will log, and we will surface those log message in the TUI.

tools/serverpod_cli/lib/src/commands/start.dart
FXschwartz: Done!

packages/serverpod/lib/src/endpoints/insights.dart
FXschwartz: Done! Pod is now stopped during the migration and verification

tools/serverpod_cli/lib/src/commands/start.dart
FXschwartz: Updated!

packages/serverpod_database/lib/src/migrations/apply_migrations.dart
FXschwartz: Done!

packages/serverpod_database/lib/src/migrations/apply_migrations.dart
FXschwartz: Updated the documentation

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -366,25 +367,38 @@ List<String> _withApplyMigrations(List<String> serverArgs) {
...
+  ).createServiceClient();
   try {
-    await applyPendingMigrations(
-      serverDir: serverDir,
-      runMode: runMode,
-      moduleName: moduleName,
+    final result = await client.insights.applyMigrations(
+      applyRepairMigration: true,
+      applyMigrations: true,
     );
```
marcelomendoncasoares: Comment: I was going to make a question, but most of my concerns with this PR are kind of covered with the existing migration e2e tests. The main one was that we'll be always calling both repair and normal migrations, which differs from the normal "start server with --apply-migrations". Coverage is not that much robust, but shows that the scenario is officially supported and covered already.

https://github.com/serverpod/serverpod/blob/48e69bd7e5d941ada6138010cc4c7fe39c076be5/tests/serverpod_test_server/test_e2e_migrations/apply_repair_migration_test.dart#L213-L240

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -366,25 +367,38 @@ List<String> _withApplyMigrations(List<String> serverArgs) {
...
+    final applied = result.migrationsApplied;
+    if (applied != null && applied.isNotEmpty) {
+      log.info(formatAppliedMigrations(applied));
+    }
+    if (!result.databaseMatchesTargetState) {
+      log.warning(
+        'Failed to apply existing migrations — database does not match '
+        'target state.',
+      );
+    }
```
marcelomendoncasoares: Comment: Once these messages are unified on the `applyMigrationsAndVerify` function, they will be available on the running `start` through the endpoint logs, with no need to repeat here.

tests/serverpod_test_server/test_integration/migration_manager/apply_migrations_endpoint_test.dart
marcelomendoncasoares: The case is probably the container isolation indeed. But let's defer the testing to another issue. Important points for coverage are:

- Calling with pending migrations.
- Calling with pending repair migrations.
- Calling with pending both repair (targeting old version) and normal migration (created afterwards).
- Verify that the server stopped to apply migrations (we can use a migration with modified SQL to add a delay and try to call other endpoints during this time).

@coderabbitai Please create an issue for this.

tests/docker/tests_integration/run_tests.sh
```diff
@@ -25,6 +25,20 @@ echo "### Apply migrations"
...
+echo "### Installing CLI tools"
+export PATH="$PATH":"$HOME/.pub-cache/bin"
+export CI=true
+(
+  cd /app/tools/serverpod_cli
+  dart pub global activate -s path .
+  serverpod version
+  dart pub global activate -s path .
+)
+
```
marcelomendoncasoares, line 41: Comment: This is a pain when working with migrations, but not the place to fix. If you want to run tests using this script, just ensure to rebuild the CLI between calls. The script itself shouldn't have this because it is not needed on CI. We definitely need a better way to handle it though.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -1232,6 +1226,40 @@ class Serverpod {
...
+  /// e.g. applying migrations through
+  /// [InsightsEndpoint.applyMigrations] — to provide the same safety
+  /// guarantees as a pod restart.
+  ///
+  /// The operator-facing Insights server is not paused, so the call that
+  /// invoked [action] can complete and return its result.
+  Future<T> pauseRequestHandlingDuring<T>(
+    Future<T> Function() action,
+  ) async {
+    await server.shutdown();
```
marcelomendoncasoares, line 1202: Question: Don't we need an `exitProcess: false`?

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -1232,6 +1226,40 @@ class Serverpod {
...
+    final webServerWasRunning =
+        _webServer != null && Features.enableWebServer(_webServer);
+    if (webServerWasRunning) {
+      await _webServer!.stop();
+    }
```
marcelomendoncasoares: Question: Doesn't shutdown already handle it?

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -1232,6 +1226,40 @@ class Serverpod {
...
+    try {
+      return await action();
+    } finally {
+      await server.start(
+        authenticationHandler:
+            authenticationHandler ?? defaultAuthenticationHandler,
+      );
+      if (webServerWasRunning) {
+        await _webServer!.start();
+      }
```
marcelomendoncasoares: Question: Why pass the authentication handler back again? And why start the webserver separately?

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -958,18 +957,13 @@ class Serverpod {
           }
         }
       }
-
-      _internalLogVerbose('Verifying database integrity.');
-      verified = await MigrationManager.verifyDatabaseIntegrity(
-        internalSession,
-      );
```
marcelomendoncasoares: Suggestion: We should also move the logs to the new `applyMigrationsAndVerify` function so we get the same experience regardless from where it is called. We do it already for the `MigrationManager`:

https://github.com/serverpod/serverpod/blob/48e69bd7e5d941ada6138010cc4c7fe39c076be5/packages/serverpod_database/lib/src/migrations/migration_manager.dart#L157-L165

marcelomendoncasoares (changes requested):
Nice work with the changes! Everything seems in place and mostly ready to merge after a few adjustments - and deferring some work to future PRs.

packages/serverpod/lib/src/server/serverpod.dart
FXschwartz: @marcelomendoncasoares I think we would only need that if we were killing the pod like `Serverpod.shutdown()` does. Where as this just closes down the http listeners/websockets. But let me know if I'm missing something

packages/serverpod/lib/src/server/serverpod.dart
FXschwartz: @marcelomendoncasoares I might be confusing the two but I thought `Server.shutdown()` only closes the API server's Relic listener + websockets. Not the `WebServer`

packages/serverpod/lib/src/server/serverpod.dart
marcelomendoncasoares: Sorry, my bad... I read as if it was calling on `Serverpod`... Ignore this and related comments on the shutdown 🤦‍♂️

tests/docker/tests_integration/run_tests.sh
FXschwartz: Yeah sounds good, I saw that `tests/docker/tests_e2e_migrations/run-tests` does the same thing, which is where I got this idea from. Do you want me to create an issue to figure out a better solution in the future?

packages/serverpod/lib/src/server/serverpod.dart
FXschwartz: Both addressed by the `_startUserFacingServers()` extraction. The auth handler and the web server start are now in the helper, called from both the boot path and `withPausedRequestHandling`.

tests/docker/tests_integration/run_tests.sh
marcelomendoncasoares: Hmm Got it. You added migration tests to the integration tests, where the CLI were not available. Ok, leave it like this for now, or CI won't pass also.

packages/serverpod/lib/src/server/serverpod.dart
FXschwartz: Done! Moved all repair/migration/drift logging into `applyMigrationsAndVerify`, stripped duplicates from `_applyMigrations` and the CLI watch loop.

tools/serverpod_cli/lib/src/commands/start.dart
FXschwartz: Done! The logging is now centralized inside `applyMigrationsAndVerify`, so the CLI watch loop and pod boot no longer repeat the messages.

packages/serverpod_database/lib/src/migrations/apply_migrations.dart
```diff
@@ -26,15 +29,35 @@ Future<MigrationsApplyResult> applyMigrationsAndVerify({
   String? repairMigrationApplied;
   if (applyRepairMigration) {
     repairMigrationApplied = await manager.applyRepairMigration(session);
+    if (repairMigrationApplied == null) {
+      log.error('Failed to apply database repair migration.');
+    } else {
```
marcelomendoncasoares: Question: Isn't this the default path if no repair migration is available? It make sense to log an error if applying the repair is explicit (like passing `--apply-repair-migration` to start the server), but that is not the case with the `start`. Since we'll always call with `applyRepairMigration`, it will always log an error, right? On the normal migrations we handle the three possible states through the list being null, empty or filled.

packages/serverpod_database/lib/src/migrations/apply_migrations.dart
```diff
@@ -26,15 +29,35 @@ Future<MigrationsApplyResult> applyMigrationsAndVerify({
...
     migrationsApplied = await manager.migrateToLatest(session) ?? const [];
+    if (migrationsApplied.isEmpty) {
+      log.info('Latest database migration already applied.');
+    } else {
+      log.info(
+        'Applied database migration${migrationsApplied.length > 1 ? 's' : ''}:',
+      );
+      for (final migration in migrationsApplied) {
+        log.info(' - $migration');
+      }
```
marcelomendoncasoares: Suggestion: Group the messages and log one string to avoid messages being split on the CLI.

marcelomendoncasoares (changes requested):
One question and one minor suggestion.

packages/serverpod_database/lib/src/migrations/apply_migrations.dart
FXschwartz: Grouped!

packages/serverpod_database/lib/src/migrations/apply_migrations.dart
FXschwartz: Ahh yeah you are right. Fixed!

marcelomendoncasoares (approved):
Great work, @FXschwartz! LGTM 🚀

PR #5157, opened 2026-05-19
fix: Fix serialization/deserialization of project models on `dynamic` fields of module models

### LINKED ISSUE #5156: Module models with `dynamic` fields fail to serialize/deserialize when holding project model objects
### Describe the bug

When using `dynamic` fields on module models, serialization/deserialization when project model objects are present fail due to the module `Protocol` not knowing project models.

### To reproduce

1. On a module, create a model with a `dynamic` field:
    ```yaml
    class: DynamicOnModule
    fields:
      name: String
      data: dynamic
    ```

2. On a project that uses this module, create a normal model:
    ```yaml
    class: SimpleData
    fields:
      num: int
    ```

3. Store the project model in the module model an try to serialize/deserialize:
    ```dart
    final simpleData = server.SimpleData(num: 1);
    final model = module.DynamicOnModule(name: 'test', data: simpleData);
    final serialized = model.toJson();
    ```

4. See the exception:
    ```console
    Invalid argument(s): Could not find class name for _SimpleDataImpl in serialization.
    package:serverpod_serialization/src/serialization.dart 229:7                     SerializationManager.wrapWithClassName
    package:serverpod_test_module_server/src/generated/protocol.dart 326:18          Protocol.wrapWithClassName
    package:serverpod_serialization/src/serialization.dart 400:25                    SerializationManager._dynamicFieldValueToJson
    package:serverpod_serialization/src/serialization.dart 355:7                     SerializationManager.dynamicFieldToJson
    package:serverpod_test_module_server/src/generated/dynamic_on_module.dart 54:30  DynamicOnModule.toJson
    test/dynamic_on_module_test.dart 27:36                                           main.<fn>.<fn>
    FormatException: No deserialization found for type named serverpod_test.SimpleData
    ```

### Expected behavior

All `dynamic` fields used inside the namespace of a project should be able to store/serialize/deserialize any known object in the project domain, regardless of which `Protocol` is serializing/deserializing under the hood.

### Library version

3.5.0 (unreleased, main)

### Platform information

N/A

### Additional context

_No response_

- [ ] I want to work on this

marcelomendoncasoares:
Fixes: #5156

The fix goes beyond the issue to also fix a missing delegation to the shared protocol for the `getClassNameForObject` and `deserializeByClassName` methods. This is required for complete `dynamic` support, but also needed for streaming serialization of shared models.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/examples/auth/auth_client/lib/src/protocol/protocol.dart
+++ b/examples/auth/auth_client/lib/src/protocol/protocol.dart
@@ -23,9 +23,9 @@ export 'client.dart';
 class Protocol extends _i1.SerializationManager {
   Protocol._();
 
-  factory Protocol() => _instance.._registerHostProtocols();
+  factory Protocol() => _instance;
 
-  static final Protocol _instance = Protocol._();
+  static final Protocol _instance = Protocol._().._registerHostProtocols();
 
   static String? getClassNameFromObjectJson(dynamic data) {
     if (data is! Map) return null;
--- a/examples/auth/auth_client/lib/src/protocol/protocol.dart
+++ b/examples/auth/auth_client/lib/src/protocol/protocol.dart
@@ -23,7 +23,7 @@ export 'client.dart';
 class Protocol extends _i1.SerializationManager {
   Protocol._();
 
-  factory Protocol() => _instance;
+  factory Protocol() => _instance.._registerHostProtocols();
 
   static final Protocol _instance = Protocol._();
 
@@ -100,11 +100,15 @@ class Protocol extends _i1.SerializationManager {
     }
     className = _i4.Protocol().getClassNameForObject(data);
     if (className != null) {
-      return 'serverpod_auth_idp.$className';
+      return className.contains('.')
+          ? className
+          : 'serverpod_auth_idp.$className';
     }
     className = _i5.Protocol().getClassNameForObject(data);
     if (className != null) {
-      return 'serverpod_auth_core.$className';
+      return className.contains('.')
+          ? className
+          : 'serverpod_auth_core.$className';
     }
     return null;
   }
@@ -127,6 +131,11 @@ class Protocol extends _i1.SerializationManager {
       return _i5.Protocol().deserializeByClassName(data);
     }
     return super.deserializeByClassName(data);
+  }
+
+  void _registerHostProtocols() {
+    _i4.Protocol().registerHostProtocol('auth', this);
+    _i5.Protocol().registerHostProtocol('auth', this);
   }
 
   /// Maps any `Record`s known to this [Protocol] to their JSON representation
--- a/modules/legacy/serverpod_auth/serverpod_auth_server/lib/src/generated/protocol.dart
+++ b/modules/legacy/serverpod_auth/serverpod_auth_server/lib/src/generated/protocol.dart
@@ -517,6 +517,15 @@ class Protocol extends _i1.DatabaseSerializationManager {
     ),
   ];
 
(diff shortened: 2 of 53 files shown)


examples/auth/auth_client/lib/src/protocol/protocol.dart
```diff
@@ -23,7 +23,7 @@ export 'client.dart';
 class Protocol extends _i1.SerializationManager {
   Protocol._();
 
-  factory Protocol() => _instance;
+  factory Protocol() => _instance.._registerHostProtocols();
```
nielsenko: Doesn't `_registerHostProtocols` happen on every `Protocol()` .. used to be cheap (just returning a singleton). 

I don't think it need to, right?

```suggestion
  factory Protocol() => _instance; // move .._registerHostProtocols(); to instance
```

examples/auth/auth_client/lib/src/protocol/protocol.dart
```diff
@@ -23,7 +23,7 @@ export 'client.dart';
...
 
   static final Protocol _instance = Protocol._();
```
nielsenko:
```suggestion
  static final Protocol _instance = Protocol._().._registerHostProtocol();
```
We only need to call this once right?

modules/legacy/serverpod_auth/serverpod_auth_server/lib/src/generated/protocol.dart
```diff
@@ -771,6 +780,61 @@ class Protocol extends _i1.DatabaseSerializationManager {
...
+        }
+        final hostData = Map<String, dynamic>.from(value);
+        hostData['className'] = strippedClassName;
+        return protocol.deserializeByClassName(hostData);
+      }
+    }
+    if (className.contains('.')) {
+      for (final protocol in _hostProtocols.values) {
+        return protocol.deserializeByClassName(value);
+      }
```
nielsenko, line 835: This looks fishy.. iterating over `_hostProtocol.values` but always returning on the first 🤔

examples/auth/auth_client/lib/src/protocol/protocol.dart
marcelomendoncasoares: Placed at the wrong line...🤦‍♂️ 

Should be the `static final` initialization.

```dart
  factory Protocol() => _instance.._registerHostProtocols();

  static final Protocol _instance = Protocol._(); // <-- Here instead.
```

modules/legacy/serverpod_auth/serverpod_auth_server/lib/src/generated/protocol.dart
marcelomendoncasoares: Good catch! Forgot to add the try/catch... I'll add a test with two hosts and a model that only the second can deserialize.

marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5157/commits/37a6e5c1c9e7c487e21f7f45eb6ea29851edbeb1.

examples/auth/auth_client/lib/src/protocol/protocol.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5157/commits/f8253568e1a56b2ab95831d755b6237bf8f1c262.

nielsenko (approved):
LGTM 👍

PR #5159, opened 2026-05-20
feat: Spawn flutter app from serverpod start

### LINKED ISSUE #5073: Web page reload is not triggered when editing code

### LINKED ISSUE #5127: Manage `flutter run` from `serverpod start`

nielsenko:
Auto-launches the project's companion Flutter app alongside the server in `serverpod start`, with coordinated reload and IDE attach for both processes.

- Spawns `flutter run --machine` when a companion package is present; opt out with `--no-flutter`.
- File-watch triggers a single codegen + server-reload + Flutter-reload pass, serialized through `WatchSession._chain`.
- New flags: 
   - `--flutter-device=<device>` (default `chrome`),
   - `--flutter-option=<arg>` (repeatable). 
- `--verbose` propagates to `flutter --verbose`.
- Reload routes through the daemon's `app.restart` over stdin, not `vmService.reloadSources` (the latter skips flutter_tools' recompile and breaks DWDS).
- Daemon protocol client vendored from flutter_tools' DAP adapter.
- `FlutterProcess` spawns `flutter_tools.dart` directly through dart, bypassing puro/fvm/asdf wrappers so signals reach the daemon.
- Project template `launch.json` attaches to (instead of launches) the Flutter app via a new `flutter-vm-service-info.json` sibling of the server's info file; compound config gets `stopAll: true`.
- TUI gains a Flutter output tab and a status line; `app.log` and the VM service's `Logging` stream both feed it.
- `ext.serverpod.log` events from the Flutter VM flow into the structured Log Messages tab via the same handler the server uses.
- Tests use Dart shims (`flutter_shims/`) so CI does not need a Flutter SDK.

Closes #5127 

## Pre-launch Checklist

## Breaking changes

None. Opt-out via `--no-flutter`; silently skipped for projects without a `<name>_flutter` directory containing a `pubspec.yaml`. The template change only affects newly generated projects.

### DIFF
--- a/templates/serverpod_templates/vscode/launch.json
+++ b/templates/serverpod_templates/vscode/launch.json
@@ -10,10 +10,8 @@
       "internalConsoleOptions": "neverOpen"
     },
     {
-      // Attaches to the Flutter app that `serverpod start` spawns via
-      // its `--flutter` flag (default on when a Flutter package is
-      // present). Don't launch a second Flutter instance from VS Code -
-      // that would race the spawn for the same web-server port.
+      // Attaches to the Flutter app `serverpod start` spawns. Don't
+      // launch a second instance here - it would race for the port.
       "name": "projectname_flutter",
       "type": "dart",
       "request": "attach",
@@ -26,8 +24,7 @@
     {
       "name": "projectname (full stack)",
       "configurations": ["projectname_server", "projectname_flutter"],
-      // Stop both debug sessions when the user stops one. Without this
-      // the orphan keeps running and the next launch fails to re-attach.
+      // Stopping one session stops both; the orphan blocks re-attach.
       "stopAll": true,
       "presentation": {
         "order": 1
--- a/templates/serverpod_templates/vscode/launch.json
+++ b/templates/serverpod_templates/vscode/launch.json
@@ -2,20 +2,20 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "name": "projectname_server",
-      "type": "dart",
-      "request": "attach",
-      "vmServiceInfoFile": "${workspaceFolder}/projectname_server/.dart_tool/serverpod/vm-service-info.json",
-      "preLaunchTask": "serverpod_start",
-      "internalConsoleOptions": "neverOpen"
-    },
-    {
       // Attaches to the Flutter app `serverpod start` spawns. Don't
       // launch a second instance here - it would race for the port.
       "name": "projectname_flutter",
       "type": "dart",
       "request": "attach",
       "vmServiceInfoFile": "${workspaceFolder}/projectname_server/.dart_tool/serverpod/flutter-vm-service-info.json",
+      "preLaunchTask": "serverpod_start",
+      "internalConsoleOptions": "neverOpen"
+    },
+    {
+      "name": "projectname_server",
+      "type": "dart",
+      "request": "attach",
+      "vmServiceInfoFile": "${workspaceFolder}/projectname_server/.dart_tool/serverpod/vm-service-info.json",
       "preLaunchTask": "serverpod_start",
       "internalConsoleOptions": "neverOpen"
     }
--- a/templates/serverpod_templates/vscode/launch.json
+++ b/templates/serverpod_templates/vscode/launch.json
@@ -2,17 +2,22 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "name": "projectname_flutter",
-      "cwd": "${workspaceFolder}/projectname_flutter",
-      "request": "launch",
-      "type": "dart",
-      "program": "lib/main.dart"
-    },
-    {
       "name": "projectname_server",
       "type": "dart",
       "request": "attach",
       "vmServiceInfoFile": "${workspaceFolder}/projectname_server/.dart_tool/serverpod/vm-service-info.json",
+      "preLaunchTask": "serverpod_start",
+      "internalConsoleOptions": "neverOpen"
+    },
+    {
+      // Attaches to the Flutter app that `serverpod start` spawns via
+      // its `--flutter` flag (default on when a Flutter package is
+      // present). Don't launch a second Flutter instance from VS Code -
+      // that would race the spawn for the same web-server port.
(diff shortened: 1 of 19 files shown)


templates/serverpod_templates/vscode/launch.json
```diff
@@ -2,11 +2,14 @@
...
   "configurations": [
     {
+      // Attaches to the Flutter app `serverpod start` spawns. Don't
+      // launch a second instance here - it would race for the port.
       "name": "projectname_flutter",
-      "cwd": "${workspaceFolder}/projectname_flutter",
-      "request": "launch",
       "type": "dart",
-      "program": "lib/main.dart"
+      "request": "attach",
```
Crazelu, line 9: What if this 'launch' comes after the pod is started using `serverpod start --no-flutter`?

marcelomendoncasoares (approved):
LGTM! 🚀

templates/serverpod_templates/vscode/launch.json
marcelomendoncasoares: Didn't see your comment before approving. Testing this scenario you mentioned (call start without flutter, then hit F5), it is stuck forever in "Starting debug session..."

image

But I think we can open a bug issue and treat it in sequence, for the RC.

nielsenko:
rebased on main

templates/serverpod_templates/vscode/launch.json
marcelomendoncasoares: Opened the issue #5162.

nielsenko:
fixed conflict .. now it is rebased on main

Crazelu:
One more thing:
Looks to me like this approach will prevent users from adding launch args (e.g dart-defines and flavor) directly to their vscode configurations. This is usually a setup-once-use-always kind of configuration. Having to provide the args to `serverpod start` every time is repetitive.

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -76,6 +77,38 @@ enum StartOption<V> implements OptionDefinition<V> {
...
+    MultiOption(
+      argName: 'flutter-option',
+      multiParser: MultiParser(StringParser()),
+      helpText:
+          'Extra argument forwarded to `flutter run`. Repeatable, e.g. '
+          '--flutter-option=--web-hostname=0.0.0.0 '
+          '--flutter-option=--web-port=8090.',
+      defaultsTo: [],
+    ),
+  ),
```
marcelomendoncasoares, line 111: Suggestion: Instead of having multiple flutter-specific arguments, we could expose a single `--flutter-options=""` parameter to allow full control of the underlying `flutter run`. This follows [the previous discussion on the first implementation PR](https://github.com/serverpod/serverpod/pull/5132#discussion_r3235587845).

> Discussion: Should we expose a `--flutter-device` argument to allow the `start` to invoke other targets? And maybe a `--flutter-options=""` to receive options to forward to the `flutter run` command - like the `--web-hostname` and `--web-port` that are required for testing Google IDP locally from `-d chrome`.

> We can do this and even mirror the flutter run flags

marcelomendoncasoares: Documented on issue #5163.

marcelomendoncasoares:
> One more thing: Looks to me like this approach will prevent users from adding launch args (e.g dart-defines and flavor) directly to their vscode configurations. This is usually a setup-once-use-always kind of configuration. Having to provide the args to `serverpod start` every time is repetitive.

Documented the approach we have discussed on your PR on issue #5163.

Crazelu:
> > One more thing: Looks to me like this approach will prevent users from adding launch args (e.g dart-defines and flavor) directly to their vscode configurations. This is usually a setup-once-use-always kind of configuration. Having to provide the args to `serverpod start` every time is repetitive.
> 
> Documented the approach we have discussed on your PR on issue #5163.

That's a different thing. I'm referring to launch args that can be set on the IDE configurations. Like this for vscode:

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Some App Dev",
            "request": "launch",
            "type": "dart",
            "args": [
                "--no-enable-impeller",
                "--flavor",
                "dev",
                "--dart-define-from-file",
                "some-env.json"
            ]
        }
    ]
}
```

nielsenko:
> > > One more thing: Looks to me like this approach will prevent users from adding launch args (e.g dart-defines and flavor) directly to their vscode configurations. This is usually a setup-once-use-always kind of configuration. Having to provide the args to `serverpod start` every time is repetitive.
> > 
> > 
> > Documented the approach we have discussed on your PR on issue #5163.
> 
> That's a different thing. I'm referring to launch args that can be set on the IDE configurations. Like this for vscode:
> 
> ```json
> {
>     "version": "0.2.0",
>     "configurations": [
>         {
>             "name": "Some App Dev",
>             "request": "launch",
>             "type": "dart",
>             "args": [
>                 "--no-enable-impeller",
>                 "--flavor",
>                 "dev",
>                 "--dart-define-from-file",
>                 "some-env.json"
>             ]
>         }
>     ]
> }
> ```

Nothing prevents people from doing their own `launch.json`. We bring a template that uses `serverpod start` - if they don't like they can handle startup and attach on their own.

PR #5161, opened 2026-05-20
feat: Replace SQLite on `quickstart` by the embedded Postgres

### LINKED ISSUE #5080: Replace the SQLite on `quickstart` by the embedded Postgres local alternative
After implementing the support for running Postgres without docker on development, we should make the `quickstart` command use this option by default instead of SQLite. This will provide the seamless experience of not depending on docker, not having port conflicts, but at the same time have support for all the features that depend on Postgres currently (future calls, auth, persistent logs, concurrent transactions).

### LINKED ISSUE #5174: Deleting a model under `generate --watch` or `start` is not deleting the generated file
### Describe the bug

With `serverpod start --watch`, deleting a `.spy.yaml` model does not remove the generated client protocol file (e.g. `test_entity.dart`). The server reloads, but stale codegen remains.

### To reproduce

1. Create a mini project: `serverpod create <name> --mini`
2. Run `serverpod start --watch` in `<name>_server`
3. Add `lib/src/models/test_entity.spy.yaml`, modify it, then delete it
4. Check `<name>_client/lib/src/protocol/test_entity.dart`

(Also fails in e2e: `tests/serverpod_cli_e2e_test/test/start_test.dart` — “model file is added, modified, and deleted”.)

### Expected behavior

After the model file is deleted, incremental generate should drop the model from analyzer state, remove generated entity files, then reload the server.

### Library version

3.5.0 (unreleased, main)

### Platform information

Happened on Windows on CI.

### Additional context

In `Analyzers.update`, removed models used `Uri.parse(p.absolute(path))` while added models use `Uri.file(absolutePath)` from `ModelHelper.createModelSourceForPath`. On Windows those `.path` keys can differ, so `StatefulAnalyzer.removeYamlModel` misses the entry and codegen keeps the old entity.

**Fix:** use `Uri.file(p.absolute(path))` on remove (same as on add).

marcelomendoncasoares:
Closes: #5080

~On modules, the `docker-compose.yaml` won't be generated.~ The original approach was to not generate `docker-compose.yaml` for modules, but we'll have some challenges to make it work. So it will stay to a second moment with issu #5164.

Either way, to allow a conditional rendering taking the project template into account, the `TemplateContext` now includes the project template. This will also facilitate merging the `modulename` and `projectname` templates in the future.

This PR also fixes:

- ~Project create tests were not running on CI.~ Fixed and reverted due to the long running job.
- The server `Dockerfile` was being conditionally rendered with Postgres/Redis.
- Some project create tests had references to old `tasks.json` and `launch.json` use.

And fixes #5174 with https://github.com/serverpod/serverpod/pull/5161/commits/c964ef41776515183b7da7ea53db981e25c2c0a3.

## Pre-launch Checklist

## Breaking changes

None.


tests/bootstrap_project/test/module_quickstart_test.dart
```diff
@@ -182,33 +182,41 @@ void main() {
             );
           });
 
-          test('does not have docker compose file', () {
-            final dockerComposeFile = File(
-              path.join(tempPath, serverDir, 'docker-compose.yaml'),
-            );
-            expect(dockerComposeFile.existsSync(), isFalse);
-          });
+          test(
```
FXschwartz, line 185: Doesn't this contradict your description in the PR of `On modules, the docker-compose.yaml won't be generated`.

Just flagging in case it's not needed.

tools/serverpod_cli/lib/src/commands/quickstart.dart
```diff
@@ -119,11 +119,11 @@ class QuickstartCommand extends ServerpodCommand<QuickstartOption> {
 
     final projectPath = await performCreate(
       name,
-      template,
       force,
       interactive: interactive,
       context: TemplateContext(
-        sqlite: true,
+        template: template,
+        postgres: true,
```
FXschwartz, line 126: Same thing here, I think this would cause module quickstart to now emit `docker-compose.yaml` and `passwords.yaml`. Which is probably the intent but flagging based on the PR description.

tools/serverpod_cli/lib/src/create/template_context.dart
```diff
@@ -1,8 +1,10 @@
+import 'package:serverpod_cli/src/create/create.dart';
 import 'package:serverpod_cli/src/create/ide.dart';
 
 /// Context containing values for rendering templates.
 class TemplateContext {
```
FXschwartz, line 5: It's a nitpick but `create.dart` imports `TemplateContext` and now `template_context.dart` is importing from `create.dart` as well. No harm but might be worth moving `ServerpodTemplateType` into it's own file.

FXschwartz (changes requested):

tests/bootstrap_project/test/module_quickstart_test.dart
marcelomendoncasoares: You're right, I had to give up on this and threw an issue to solve later. I'll update the PR description.

tools/serverpod_cli/lib/src/create/template_context.dart
marcelomendoncasoares: I wouldn't bother much, since Dart handles circular imports very well. And I'd really like to avoid running this CI again just for a nit 😅

FXschwartz (approved):
LGTM 👍

PR #5177, opened 2026-05-21
perf(cli): Speed up serverpod create template rendering

nielsenko:
The current code in `tools/serverpod_cli/lib/src/create/template_renderer.dart` traverses way too many files, invoking expensive operations like mustache rendering, or per-file subprocesses for dart formatting. This PR fixes the worst offenders.

- format Dart files in-process with DartFormatter instead of spawning `dart format` per file,
- skip heavy directories (`.git`, `.dart_tool`, `build`, ...) when traversing,
- and bail early on names/contents without `{{`.

In one example this makes `serverpod create .` (ie. upgrade) run go down from 10m37s to 2s, so 318x faster.

The PR does not fix the fact that blocking operations are run on the nocterm render thread, but the staggering obviously becomes much less pronounced.

### DIFF
--- a/tools/serverpod_cli/lib/src/create/template_renderer.dart
+++ b/tools/serverpod_cli/lib/src/create/template_renderer.dart
@@ -14,7 +14,6 @@ class TemplateRenderer {
     '.git',
     '.dart_tool',
     '.idea',
-    '.vscode',
     '.github',
     'node_modules',
     'Pods',
--- a/tools/serverpod_cli/lib/src/create/template_renderer.dart
+++ b/tools/serverpod_cli/lib/src/create/template_renderer.dart
@@ -1,19 +1,52 @@
 import 'dart:io';
 
+import 'package:dart_style/dart_style.dart';
 import 'package:path/path.dart' as p;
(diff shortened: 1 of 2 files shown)


Crazelu:
🙌

tools/serverpod_cli/lib/src/create/template_renderer.dart
```diff
@@ -1,20 +1,53 @@
...
 
 /// Responsible for rendering template files in a directory using provided context.
 /// It processes both file contents and directory names,
 /// allowing for dynamic project structures.
 class TemplateRenderer {
+  static const Set<String> _defaultExcludedDirNames = <String>{
+    '.git',
+    '.dart_tool',
+    '.idea',
+    '.vscode',
```
Crazelu: We have template directives in `.vscode/tasks.json`

nielsenko: I should know 😆

nielsenko: Fixed in 3a3a7f9

Crazelu (approved):

PR #5185, opened 2026-05-21
fix: Avoid depending on, or manipulating Directory.current

nielsenko:
- require explicit serverDir in ConfigInfo,
- plumb through ServerpodConfig.load,
- and anchor GeneratorConfig path at resolution time


packages/serverpod/test/server/server_directory_test.dart
```diff
@@ -0,0 +1,79 @@
...
+        }
+      }
+      tempServerDir = null;
+      cwdSentinel = null;
+    });
+
+    test(
+      'when Serverpod is constructed with serverDirectory pointing at the '
+      'temp dir, then it loads config from there even though cwd is unrelated',
+      () async {
```
marcelomendoncasoares, line 62: Question: Do we have a test for starting `Serverpod` from `Directory.current`? Or we're considering that all untouched tests that did not receive `serverDir` (how many?) cover this scenario?

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -508,7 +523,9 @@ class Serverpod {
 
     // Load passwords
     _passwordManager = PasswordManager(runMode: runMode);
-    _passwords = _passwordManager.loadPasswords();
+    _passwords = _passwordManager.loadPasswords(
+      ServerpodConfig.passwordsConfigPath(serverDir: serverDirectory.path),
+    );
```
marcelomendoncasoares, line 528: Suggestion: Instead of passing the complete passwords path, should we just add the `serverDir` to `loadPasswords`?

marcelomendoncasoares (commented):
Nice improvement! Way safer and a nice gain of parallelism on tests. Can we also benefit from this change on the bootstrap tests?

nielsenko:
Rebased on main

packages/serverpod/test/server/server_directory_test.dart
nielsenko: We do now ad8a5ef

nielsenko:
> Nice improvement! Way safer and a nice gain of parallelism on tests. Can we also benefit from this change on the bootstrap tests?

I would like to wait until after another change lands #5271

nielsenko:
Rebased on main

packages/serverpod/lib/src/server/serverpod.dart
nielsenko: Done in 701c18f.. Unless you meant to switch to `Directory` type here as well. I started down that path, but once you start there is a lot of follow-up changes.

marcelomendoncasoares: Is it too asymmetric with the `serverDir` at `loadPasswords` as `String`?

nielsenko: In general I prefer not to traffic data in `String`s and `Directory` is hence better. But in Dart `Directory` is not really a file-system entity (there is no inode/handle behind it - it really just a wrapper of a string).

I don't think we should pile more on this PR.

marcelomendoncasoares (approved):
LGTM then! 🚀

PR #5193, opened 2026-05-22
fix: Web don't send app.log events.

nielsenko:
- app.log events are now intentionally ignored
- instead we forward from vmService named `'Stdout'`, `'Stderr'`, and `'Logging'` streams


marcelomendoncasoares (approved):
👍

PR #5194, opened 2026-05-23
fix: Fixed new project templates for GitHub workflows

christerswahn:
Fix the templates for GitHub workflows that are included when a new Serverpod project is created.

Since a new Serverpod project by default includes a flutter app and the project is a workspace, `dart pub get` must be replaced with `flutter pub get`. Which also means that the workflows must install flutter rather than just dart.

There was also an issue with the `--quiet-build` flag passed to `docker compose up`, it is not supported by the standard docker version in GitHub (v28). So this flag is removed here.

The flutter version installed in the workflows must match the `pubspec.yaml` in the created flutter package. I attempted to use the `FLUTTER_VERSION` label, but that got replaced by a really old flutter version `3.32.8`. _This might be a separate bug/issue maybe worth looking into._

So here we continue to hardcode with `3.38.4`, same as the flutter package template's pubspec.yaml.

(Note: In the workflow flutter setup action, the flutter version must be distinct, and ^ or range is not supported.)

## Pre-launch Checklist

## Breaking changes

n/a

### DIFF
--- a/templates/serverpod_templates/github/workflows/format.yml
+++ b/templates/serverpod_templates/github/workflows/format.yml
@@ -13,13 +13,13 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: dart-lang/setup-dart@v1
+      - uses: subosito/flutter-action@v2
         with:
-          sdk: 3.10.3
+          flutter-version: '3.38.4'
 
       - name: Install dependencies
         working-directory: projectname_server
-        run: dart pub get
+        run: flutter pub get
 
       - name: Check formatting
         working-directory: projectname_server
--- a/templates/serverpod_templates/github/workflows/format.yml
+++ b/templates/serverpod_templates/github/workflows/format.yml
@@ -18,7 +18,6 @@ jobs:
           flutter-version: '3.38.4'
 
       - name: Install dependencies
-        working-directory: projectname_server
         run: flutter pub get
 
       - name: Check formatting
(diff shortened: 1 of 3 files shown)


templates/serverpod_templates/github/workflows/format.yml
```diff
@@ -13,13 +13,13 @@ jobs:
...
-      - uses: dart-lang/setup-dart@v1
+      - uses: subosito/flutter-action@v2
         with:
-          sdk: 3.10.3
+          flutter-version: '3.38.4'
 
       - name: Install dependencies
         working-directory: projectname_server
-        run: dart pub get
+        run: flutter pub get
```
marcelomendoncasoares, line 21: Question: Feels kind of weird to call `flutter pub get` in the server directory, but I get that this is because it will defer to the workspace, and it requires `flutter`, right? Should we remove the `working-directory` parameter then to make the call run through the workspace explicitly?

marcelomendoncasoares (commented):
Thanks for this fix, @christerswahn! I have only one comment on the change to do `flutter pub get`, but still from the server directory.

templates/serverpod_templates/github/workflows/format.yml
christerswahn: Good point, yes let's remove it. 👍

marcelomendoncasoares (approved):
LGTM! Thanks for this fix, @christerswahn! 🚀

PR #5197, opened 2026-05-25
feat: Update tui implementation to use shared tui library

Crazelu:
`serverpod create` and `serverpod start` TUI implementations now use the shared `serverpod_tui` library.

## Pre-launch Checklist

## Breaking changes

None


tools/serverpod_cli/lib/src/commands/create/tui/main_screen.dart
```diff
@@ -43,7 +34,7 @@ class MainScreen extends StatelessComponent {
               children: [
                 _buildHeader(theme),
                 Expanded(
-                  child: creatingProject ? _buildLogView() : _buildForm(theme),
+                  child: creatingProject ? _buildLogView() : _buildForm(),
```
nielsenko, line 37: Feels a bit overkill to have `_buildForm` now it is simply
```dart
    return Form(
      state: state.form,
      scrollController: scrollController,
      rebuild: holder.markDirty,
    );
```

tools/serverpod_cli/lib/src/commands/create/tui/state.dart
```diff
@@ -1,260 +1,68 @@
...
 /// Central state for [ServerpodCreateApp] rendered by nocterm.
-class CreateConfigState extends ServerpodState {
+class CreateConfigState extends TuiState {
   CreateConfigState(this.startingTemplate) {
-    _resetState(startingTemplate);
+    form.updateSelectedOption(
+      ServerpodCreateConfig.template,
+      startingTemplate.toConfigOption,
+    );
   }
```
nielsenko, line 13: This is not only related to this PR, but Ill give feedback here.

I'm not fan of using inheritance for state augmentation on app state side of thing.  On the widget site our hands are tied.

The smell is that TuiState is doing two unrelated jobs through one extends:
1. It's the read contract the framework consumes, ie. 
    ```dart
    SpinnerScope(active: state.activeOperations.isNotEmpty);
    LogViewer(state:)
    ```
    The holder's markDirty loop all reach in for logHistory / activeOperations.

2. It's the base for the app's whole mutable state bag.

Bundling those means the base holds fields that only some subclasses want. `showHelp` used to be a mutable field on the abstract base (`ServerpodState`), so `CreateConfigState`, which has no help overlay, inherited a meaningless field. 

This PR fixes that by pushing it down onto `ServerWatchState`. Good fix, but the architecture makes that anti-pattern the default. Same story with `rawLines` defaulting to `BoundedQueueList(0)` on the base.

The shape I'd prefer: the shared package defines a narrow read interface for what each component actually needs
```dart
  abstract interface class TuiLogView {
    BoundedQueueList<Object> get logHistory;
    Map<String, TrackedOperation> get activeOperations;
  }
```
and either the app state implements it, or the framework log components hold their own buffer and the app just composes them.

Now that we are promoting `serverpod_tui` as a separate package for public consumption, we should avoid tight coupling.

nielsenko (approved):
As discussed on standup, we will land PR as is (now that @FXschwartz changes are included). We can address the architectural concerns later, and document that the `serverpod_tui` package is not intended for external use (similar to `serverpod_database`).

PR #5201, opened 2026-05-26
fix: Fix the serialization of `dynamic` classes when importing both client and server

### LINKED ISSUE #5200: Serialization of `dynamic` classes fail when importing both client and server
### Describe the bug

Since we serialize/deserialize dynamic fields delegating to registered host `Protocol` instances, when both the client and server packages of a given project are imported, one will override the registration of the other on module and shared `Protocol` instances, leading to unexpected and hard-to-debug runtime errors.

### To reproduce

1. Create a module or shared model with a dynamic field.
2. Create a project model.
3. On a test file, import both client and server, instantiate the server protocol first, then the client protocol last.
4. Create an instance of the project object with the server class generated class.
5. Create an instance of the module object holding the server project object in the dynamic field.
6. Try to serialize/deserialize the module object.

The operation will fail because the client protocol overwrote the server protocol as the registered host for that project in the module protocol.

### Expected behavior

Importing the client/server should not break the serialization/deserialization of objects created with the other one.

### Library version

3.5.0 (unreleased, main)

### Platform information

N/A

### Additional context

The root cause is that we store the registered protocols through a `Map<String, SerializationManager>, with the key being the host project name. If we change to a `Set<SerializationManager>`, both client and server `Protocol` instances will be allowed to be registered and everything will work as expected.

- [x] I want to work on this

marcelomendoncasoares:
Fixes: #5200

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod_database/lib/src/interface/serialization_manager.dart
+++ b/packages/serverpod_database/lib/src/interface/serialization_manager.dart
@@ -7,9 +7,6 @@ import '../../serverpod_database.dart';
 /// overridden by generated code. [DatabaseSerializationManager] is an
 /// extension to also handle [Table]s.
 abstract class DatabaseSerializationManager extends SerializationManager {
-  /// The name of the module that defines the serialization.
-  String getModuleName();
-
   /// Maps [Type]s to subclasses of [Table].
   Table? getTableForType(Type t);
 
--- a/packages/serverpod_serialization/lib/src/serialization.dart
+++ b/packages/serverpod_serialization/lib/src/serialization.dart
@@ -47,7 +47,7 @@ Type getType<T>() => T;
 /// extended by generated code.
 abstract class SerializationManager {
   /// The name of the module that defines the serialization.
-  String getModuleName();
+  String get moduleName;
 
   /// Decodes the provided json [String] to an object of type [t] or [T].
   T decode<T>(String data, [Type? t]) {
--- a/packages/serverpod_serialization/lib/src/serialization.dart
+++ b/packages/serverpod_serialization/lib/src/serialization.dart
@@ -47,7 +47,11 @@ Type getType<T>() => T;
 /// extended by generated code.
 abstract class SerializationManager {
   /// The name of the module that defines the serialization.
-  String get moduleName;
+  ///
+  /// This method will be implemented by the generated code.
+  String getModuleName() {
+    throw UnimplementedError('This protocol does not have a module name.');
+  }
 
   /// Decodes the provided json [String] to an object of type [t] or [T].
   T decode<T>(String data, [Type? t]) {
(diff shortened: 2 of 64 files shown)


packages/serverpod_serialization/lib/src/serialization.dart
```diff
@@ -46,6 +46,9 @@ Type getType<T>() => T;
 /// serialization, but also for serializing objects. This class is typically
 /// extended by generated code.
 abstract class SerializationManager {
+  /// The name of the module that defines the serialization.
```
FXschwartz, line 49: This is potentially a breaking change no? For anything that extends `SerializationManager `

packages/serverpod_database/lib/src/interface/serialization_manager.dart
```diff
@@ -7,9 +7,6 @@ import '../../serverpod_database.dart';
 /// overridden by generated code. [DatabaseSerializationManager] is an
 /// extension to also handle [Table]s.
 abstract class DatabaseSerializationManager extends SerializationManager {
-  /// The name of the module that defines the serialization.
-  String getModuleName();
-
```
FXschwartz, line 12: Same thing here, this would break any manually written code that extends `SerializationManager` or calls `getModuleName()` directly.

FXschwartz (changes requested):
Nice PR! Besides the breaking changes what do you think about having a regression test that reproduces the scenario of #5200? 

Right now the closest coverage is `dynamic_host_protocol_delegation_test.dart`, but that registers two different modules (`serverpod_test_sqlite` + `serverpod_test`), under the old `Map<String, ...>`, which used distinct keys and never collided, so that test would have passed before this fix too.

Since the failure is same project. Client overwrites server under the same key: register a project's client and server protocol together into a module/shared protocol, then round-trip a server-built object through a dynamic field.

A third group in the existing delegation test with something like `"Given a module protocol with both the client and server protocol of the same project registered`?

packages/serverpod_database/lib/src/interface/serialization_manager.dart
marcelomendoncasoares: It would, on a pure software perspective, but we reserve ourselves the right to change generated code (or code that exists as base for the generation) freely. Manual extensions are not a concern, because these classes are not meant to be manually extended outside of Serverpod.

On a second thought, though, you're right that this is quite a public interface that might be used outside of the generated code. I was a bit eager to refactor to make the code cleaner, but let's stick to the old method.

packages/serverpod_serialization/lib/src/serialization.dart
marcelomendoncasoares: Reverted both changes at https://github.com/serverpod/serverpod/pull/5201/commits/06defd6554f2ae0d22a24b555b97a78f57ce7461.

FXschwartz (approved):

PR #5203, opened 2026-05-26
feat: Implement a `cloud` Serverpod CLI command to forward to `scloud`

### LINKED ISSUE #5202: Users should be able to access `scloud` from the `serverpod` command
To deploy to cloud right now, users need to manually install a second `scloud` command, which lowers the discoverability of the service. We should expose the `scloud` features through a `serverpod cloud` command, but without creating a constraint on the development/publishing for both ends.

marcelomendoncasoares:
Closes: #5202

~When invoked through `serverpod cloud`, all occurrences of `scloud` are also replaced for consistency:~
> image

> Revoked at https://github.com/serverpod/serverpod/pull/5203/commits/d8c33cd856f14d3c65b82a5752019310e8024d74 due to the issues that it would bring when `scloud` features a TUI.
## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/cloud.dart
+++ b/tools/serverpod_cli/lib/src/commands/cloud.dart
@@ -6,6 +6,7 @@ import 'package:args/args.dart';
 import 'package:async/async.dart';
 import 'package:cli_tools/cli_tools.dart';
 import 'package:config/config.dart';
+import 'package:dart_data_home/dart_data_home.dart';
 import 'package:path/path.dart' as p;
 import 'package:serverpod_cli/src/runner/serverpod_command.dart';
 import 'package:serverpod_cli/src/util/sdk_path.dart';
@@ -96,18 +97,28 @@ Future<Process> _startScloudProcess(
   List<String> args, {
   required bool installIfMissing,
 }) async {
-  try {
-    return await Process.start(
-      'scloud',
-      args,
-      workingDirectory: Directory.current.path,
-    );
-  } on ProcessException catch (exception) {
+  final scloudExecutable = getScloudExecutablePath();
+
+  if (!File(scloudExecutable).existsSync()) {
     if (installIfMissing) {
       await _installScloud();
       return _startScloudProcess(args, installIfMissing: false);
     }
 
+    log.error(
+      'Failed to start Serverpod Cloud CLI: '
+      'executable not found at $scloudExecutable',
+    );
+    throw ExitException(ServerpodCommand.commandInvokedCannotExecute);
+  }
+
+  try {
+    return await Process.start(
+      scloudExecutable,
+      args,
+      workingDirectory: Directory.current.path,
+    );
+  } on ProcessException catch (exception) {
     log.error(
       'Failed to start Serverpod Cloud CLI: ${exception.message}',
     );
@@ -146,4 +157,44 @@ Future<void> _closeProcessStdin(IOSink sink) async {
   } on IOException {
     // Ignore broken pipe errors during shutdown.
   }
+}
+
+/// Resolves the full path to the Serverpod Cloud CLI (`scloud`) executable.
+///
+/// Tries in order:
+/// 1. The `dart install` bin directory (where `_installScloud` places it)
+/// 2. The legacy pub-cache bin directory
+///
+/// Returns the preferred `dart install` path even when the executable is
+/// missing so callers can trigger installation.
+String getScloudExecutablePath() {
+  final name = Platform.isWindows ? 'scloud.bat' : 'scloud';
+
+  final dartInstallPath = p.join(getDartDataHome('install'), 'bin', name);
+  if (File(dartInstallPath).existsSync()) {
+    return dartInstallPath;
+  }
+
+  final pubCachePath = p.join(_resolvePubCacheBinDirectory(), name);
+  if (File(pubCachePath).existsSync()) {
+    return pubCachePath;
+  }
+
+  return dartInstallPath;
+}
+
+String _resolvePubCacheBinDirectory() {
+  final pubCache = Platform.environment['PUB_CACHE'];
+  if (pubCache != null && pubCache.isNotEmpty) {
+    return p.join(pubCache, 'bin');
+  }
+
+  if (Platform.isWindows) {
+    final localAppData = Platform.environment['LOCALAPPDATA'];
+    if (localAppData != null && localAppData.isNotEmpty) {
+      return p.join(localAppData, 'Pub', 'Cache', 'bin');
+    }
+  }
+
+  return p.join(Platform.environment['HOME'] ?? '', '.pub-cache', 'bin');
 }
--- a/tools/serverpod_cli/lib/src/commands/cloud.dart
+++ b/tools/serverpod_cli/lib/src/commands/cloud.dart
@@ -1,0 +1,141 @@
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:args/args.dart';
+import 'package:async/async.dart';
+import 'package:cli_tools/cli_tools.dart';
+import 'package:config/config.dart';
+import 'package:path/path.dart' as p;
+import 'package:serverpod_cli/src/runner/serverpod_command.dart';
+import 'package:serverpod_cli/src/util/sdk_path.dart';
+import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart';
+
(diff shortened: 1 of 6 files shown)


FXschwartz (approved):
LGTM!

tools/serverpod_cli/lib/src/generated/completion_script_carapace.dart
```diff
@@ -43,6 +43,8 @@ commands:
             tool: ["completely", "carapace"]
             write-dir: ["$directories"]
 
+  - name: cloud
+
   - name: create
     flags:
       -f, --force: "Create the project even if there are issues that prevent it from running out of the box."
```
nielsenko, line 50: This is a symptom of the _shell out_ approach. We don't get proper completion support.

tools/serverpod_cli/lib/src/commands/cloud.dart
```diff
@@ -0,0 +1,149 @@
...
+  }
+}
+
+Future<void> _installScloud() async {
+  final dartExecutable = p.join(getSdkPath(), 'bin', 'dart');
+  log.info('Installing Serverpod Cloud CLI...');
+  final installProcess = await Process.start(dartExecutable, [
+    'install',
+    'serverpod_cloud_cli',
+  ]);
```
nielsenko: Use `log.progress` for nicer experience

tools/serverpod_cli/lib/src/commands/cloud.dart
```diff
@@ -0,0 +1,149 @@
...
+    log.error('Failed to install Serverpod Cloud CLI.');
+    throw ExitException(ServerpodCommand.commandInvokedCannotExecute);
+  }
+}
+
+/// Rewrites [scloud] branding in child process output for the parent CLI.
+Stream<List<int>> _replaceScloudInOutput(Stream<List<int>> stream) {
+  return stream
+      .transform(utf8.decoder)
+      .map((text) => text.replaceAll('scloud', 'serverpod cloud'))
```
nielsenko: Not fan, but it should work 99% of the time..

nielsenko: Actually this can be chunked, so you see `sc` followed by `loud`. In that case replace won't happen.

nielsenko: It will also replace `scloud` in emitted paths fx. `~/.scloud/` becomes `~/.serverpod cloud/` which I'm sure won't work. 

Not sure if this is an actual issue today, but easy to hit down the line.

tools/serverpod_cli/lib/src/commands/cloud.dart
```diff
@@ -0,0 +1,149 @@
...
+}) async {
+  try {
+    return await Process.start(
+      'scloud',
+      args,
+      workingDirectory: Directory.current.path,
+    );
+  } on ProcessException catch (exception) {
+    if (installIfMissing) {
+      await _installScloud();
```
nielsenko: There is no guarantee that the installed `scloud` is available on `$PATH`. after this, is there? Should we ensure to use full path when calling `scloud`? Saw feedback on similar issue after hackathon.

nielsenko (changes requested):
This shell out approach allows us to run all `scloud <x>` as `serverpod cloud <x>`.
However the UX is not as good as if we took a dep on the dart package.  Subcommands and options of `scloud` won't be included in usage, nor in completion script. We also introduce some brittleness regarding finding the `scloud` exe on `PATH`.

Let's discuss, but I would prefer doing this differently.

tools/serverpod_cli/lib/src/generated/completion_script_carapace.dart
marcelomendoncasoares: I would also say that we don't want to. Because the goal of this change is to provide an access point, but we don't want to create a development constraint that would require us to frequently upgrade the dependency and carapace completions to match changes in the `scloud` - or them having to worry about adding dependencies that are incompatible with `serverpod_cli`.

tools/serverpod_cli/lib/src/commands/cloud.dart
marcelomendoncasoares: Good suggestion!

tools/serverpod_cli/lib/src/commands/cloud.dart
marcelomendoncasoares: This is mostly to fix help messages, that will suggest calling `scloud ...` and might be confusing if the command was accessed from `serverpod cloud`. But I did not though on the path replacement indeed - and `.scloud` exists on as a transitive path on the repository... Maybe it is better to exhaust the patterns that can happen in help messages instead of trying a global replacement approach.

tools/serverpod_cli/lib/src/commands/cloud.dart
marcelomendoncasoares: Indeed, specially since `dart install` will place the binary in a different place that users still using `dart pub global` might not have mapped in their PATH. We can use the full package path to be safer.

marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5203/commits/81ce389aa1922bcc59a1536dfd737b4648547761.

tools/serverpod_cli/lib/src/commands/cloud.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5203/commits/f95b437faab29276d1fedd869c26b1930d6e812c.

tools/serverpod_cli/lib/src/commands/cloud.dart
marcelomendoncasoares: Made more secure at https://github.com/serverpod/serverpod/pull/5203/commits/7447b52a5b21de64900f3d08aa8c2e10acd1c03a.

tools/serverpod_cli/lib/src/commands/cloud.dart
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    final emit = pending.substring(0, lastSpace + 1);
+    pending = pending.substring(lastSpace + 1);
+    yield _encodeReplacingScloudInText(emit);
+  }
+
+  if (pending.isNotEmpty) {
+    yield _encodeReplacingScloudInText(pending);
+  }
```
nielsenko: this seems overly complex. Would:
```dart
stream
    .transform(const Utf8Decoder(allowMalformed: true)) // for robustness - utf8 not guaranteed
    .transform(const LineSplitter()) // handles chunking
    .map((line) => line.replaceAllMapped(pattern, ...) + Platform.lineTerminator) // where pattern is your regex
    .transform(const Utf8Encoder());
```
do?

tools/serverpod_cli/lib/src/commands/cloud.dart
```diff
@@ -0,0 +1,231 @@
...
+}
+
+List<int> _encodeReplacingScloudInText(String text) {
+  return utf8.encode(
+    text.replaceAllMapped(
+      RegExp(r'''(["' ])scloud(["' ])?'''),
```
nielsenko: This seems wrong.. this will not match "scloud", since it doesn't start with either '"', "'", or ' '?!

nielsenko (dismissed):
Fix not quite there ..

nielsenko:
Actually.. what about interactive `scloud` commands? Won't this mangling/buffering of `stdout`/`stderr` break them?

tools/serverpod_cli/lib/src/commands/cloud.dart
marcelomendoncasoares: But this is actually the idea. We only replace occurrences of the word `scloud` when in isolation, to avoid replacing `.../.scloud/...` for example. Looking at the help messages, we should only expect occurrences of "scloud <comand>" or "... scloud ...".

nielsenko: But "scloud " won't be matched? (" scloud" will)

tools/serverpod_cli/lib/src/commands/cloud.dart
marcelomendoncasoares: Better indeed.

marcelomendoncasoares: Removed everything in favor of the safer inherited IO we discussed, to later support the TUI as well.

Issue #591, opened 2023-01-30
OpenAPI support (as base or as export)


jonasbark:
**Is your feature request related to a problem? Please describe.**
While the flutter first approach is appealing I am afraid that exposing the server for other (web/other) clients is only possible with a lot of extra work that feels like duplicating the same thing. The models are already in yaml format and I'd have to write an additional OpenAPI/ swagger file and keep it up to date manually.

**Describe the solution you'd like**
serverpod could be built up based on an OpenAPI definition as single source of truth: models and routes are generated based on the swagger definition instead of the custom yaml format. OpenAPI definitions are the most widely used definition in backend development and creating another yaml definition seems a lot.
Doing so makes it easy to expose the API definitions to other clients.

Alternatively an option could be provided to easily export compliant OpenAPI definition.

**Describe alternatives you've considered**
-

**Additional context**
-

fischerscode:
Generating an OpenAPI definition should be (easily) possible.

But I don't really know, how using OpenAPI as a single source of truth would work. Especially since Serverpod also manages the database connection.

jonasbark:
Database definitions is a very different case indeed and OpenAPI in itself is unrelated to database entities anyway. Only the public facing models (and routes) are part of it.

In regards to single source of truth: the definition of the returned models via yaml seems veeeery similar to what an OpenAPI definition offers. It also defines e.g. model names, attribute names and types (int, string etc) and supports types probably not covered yet (I may be wrong) by serverpod such as enums. So imo it makes sense to reuse the popular structure from OpenAPI instead of creating a new one.

You'd end up with a OpenAPI server (serverpod) and client generator with the added benefit of building up on a well supported, documented and popular API definition system and be able to reuse all of its tools.

vlidholt:
I got a feel that you only had a superficial look at Serverpod.

Serverpod strips away a couple of steps compared to other API generators. Serverpod will actually analyze your server code and add the corresponding methods in a generated client library. It supports things like null safety, optional and named parameters and passing of native Dart objects between the client and the server.

Having generated bindings to the database is super useful. You can use native Dart types to access records and send them directly to the client without any manual transformation. (If needed, you can specify columns that should be stripped out if they are internal to the server.) It also allows you to add these objects straight the cache, simplifying another common task.

[...]

PR #5209, opened 2026-05-26
fix: Call endpoints passing the method on the route instead of the body

### LINKED ISSUE #5229: Generated code should use the endpoint method name in the path instead of using a parameter
The server already has support for paths over parameters; we should update the generated code to use it over the parameter we're currently using.

One use case is that it makes it easier to inspect the network traffic.

marcelomendoncasoares:
Complimentary work for the changes done at #2676.
One small step towards #591.
Closes #5229.

## Pre-launch Checklist

## Breaking changes

None, we're only changing the default behavior to explicitly passing the method name on the route instead of the body, as it was already supported since #2676.

### DIFF
--- a/packages/serverpod_client/lib/src/serverpod_client_shared.dart
+++ b/packages/serverpod_client/lib/src/serverpod_client_shared.dart
@@ -558,8 +558,10 @@ abstract class ServerpodClientShared extends EndpointCaller {
       var authenticationValue = authenticated
           ? await authKeyProvider?.authHeaderValue
           : null;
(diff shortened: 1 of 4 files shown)


nielsenko (approved):
This was a surprisingly small PR 🤔

marcelomendoncasoares:
> This was a surprisingly small PR 🤔

I only did this because @Isakdl said it was already supported since a long time, just not wired on the client. But I was also surprised to see that not even the generated code had to change!

Issue #5167, opened 2026-05-20
Quitting `serverpod start` from the terminal does not end the Flutter app


marcelomendoncasoares:
When the Flutter app is started from `serverpod start` on the terminal (not on the IDE), quitting `start` does not close the app. Only when running from the IDE that the app is closed when hitting Q on the TUI.

PR #5240, opened 2026-05-27
feat: Replace the default Flutter device by a special `web-server` that launches the browser

### LINKED ISSUE #5226: Using `serverpod start`, opening a new browser (using the Flutter link) does not handle hot reload
Hot reload does not work when opening a second window. We should investigate if this is possible to solve. Sometimes the connection is lost to the original window, and it does not reconnect.

marcelomendoncasoares:
Closes #5226 by replacing the default `chrome` device with a special `web-server` that launches automatically in the default browser when the app is started. Automatic launch is opt-out by passing `--flutter-device=web-server` to `serverpod start`.

Also lower the priority of #5167 - which proves very hard to do properly - because the browser is no longer a responsibility of the `serverpod start` command.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
+++ b/tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
@@ -80,7 +80,7 @@ class FlutterProcess {
   String? _flutterAppUrl;
   String? _dtdUri;
   VmService? _vmService;
-  bool _browserOpened = false;
+  bool _browserOpening = false;
 
   // `null` result means the process exited before publishing a URI.
   final Completer<String?> _vmServiceUriCompleter = Completer<String?>();
@@ -501,15 +501,15 @@ class FlutterProcess {
   }
 
   Future<void> _openBrowser(Uri url) async {
-    if (_browserOpened) return;
-    _browserOpened = true;
+    if (_browserOpening) return;
+    _browserOpening = true;
     final open = _openBrowserForTesting ?? BrowserLauncher.openUrl;
     if (!await open(url)) {
       log.warning('Could not open browser at $url');
     }
 
     // Reset the flag so we can open the browser again if the app is restarted.
-    _browserOpened = false;
+    _browserOpening = false;
   }
 
   bool _appStopHandled = false;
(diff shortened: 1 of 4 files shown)


tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
```diff
@@ -482,6 +500,18 @@ class FlutterProcess {
     }
   }
 
+  Future<void> _openBrowser(Uri url) async {
+    if (_browserOpened) return;
+    _browserOpened = true;
```
nielsenko: If I read correct this flags prevents multiple ongoing open calls. Should we rename it `_browserOpening`?

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
```diff
@@ -482,6 +500,18 @@ class FlutterProcess {
...
+    final open = _openBrowserForTesting ?? BrowserLauncher.openUrl;
+    if (!await open(url)) {
```
nielsenko, line 507: `open` return as soon as the page has opened (at least on macos). If this expected to keep the `_browerOpened` flag high while the page is opened it fails.

But I suspect it is just the wording of the variable that is off.

nielsenko (approved):
Had a nit about clarity of the naming of the `_browserOpened` flag.

Also, perhaps we should not close #5167 with this. The problem still exists, it is just not on the default path anymore.

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
marcelomendoncasoares: Yes, I'll rename to "_browserOpening". The idea is indeed to just avoid launching twice (which happened to me) while a launch is in progress.

marcelomendoncasoares:
> Also, perhaps we should not close #5167 with this. The problem still exists, it is just not on the default path anymore.

Fair enough. I changed to only lower the priority of #5167, since now we won't be facing this out-of-the-box.

PR #5241, opened 2026-05-27
fix: Show global options in command help

### LINKED ISSUE #5228: The `--no-interactive` option is not being shown on the help command for `create`
We do support using `--no-interactive` to skip the TUI and create directly (like the old experience), but this option should be shown in the help command.

FXschwartz:
Global options like `--no-interactive`, `--verbose`, and `--analytics` are defined on the command runner and apply to every command. They do work after a subcommand, but the `args` package only lists a command's *own* options in its `--help` output and merely appends a "Run 'serverpod help' to see global options."

This PR overrides `usage` in the shared `ServerpodCommand` base class to render the runner's global options and substitute them in place of that note.

Instead of just fixing this for `--no-interactive` I wen't ahead and appended the global commands to **ALL** subcommands. If this it not desired it's a very easy change.

Running `serverpod create --help`

Before
image
After
image

Fixes #5228 

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tools/serverpod_cli/lib/src/runner/serverpod_command.dart
+++ b/tools/serverpod_cli/lib/src/runner/serverpod_command.dart
@@ -1,3 +1,4 @@
+import 'package:args/args.dart';
 import 'package:cli_tools/cli_tools.dart';
 import 'package:config/config.dart';
 import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart';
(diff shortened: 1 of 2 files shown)


tools/serverpod_cli/lib/src/runner/serverpod_command.dart
```diff
@@ -18,4 +19,36 @@ abstract class ServerpodCommand<O extends OptionDefinition>
...
+
+    // Replace the `args` note that points to the top-level help with the
+    // global options themselves, falling back to appending them if the note
+    // can no longer be found.
+    final globalOptionsNote = RegExp(
+      r'Run\s+"\S+\s+help"\s+to\s+see\s+global\s+options\.',
+    );
+    if (baseUsage.contains(globalOptionsNote)) {
+      return baseUsage.replaceFirst(globalOptionsNote, globalOptionsSection);
+    }
```
nielsenko: Not fan..

nielsenko (commented):
I think this is fixed at the wrong layer. In my opinion this belongs in `cli_tools`. Every `BetterCommand` attached to a `BetterCommandRunner` with `globalOptions` has the same `--help` gap, fx `scloud deploy --help`.

Also, substitution couples to `package:args`'s private wording.
```dart
    final globalOptionsNote = RegExp(
      r'Run\s+"\S+\s+help"\s+to\s+see\s+global\s+options\.',
    );
```
That string _"Run " help" to see global options."_ is generated inside `CommandRunner.usage`. Any wording change in a future `args` becomes a problem. Could we avoid using regex for this? At least add a test to spot regressions.

@christerswahn what is your take?

tools/serverpod_cli/lib/src/runner/serverpod_command.dart
FXschwartz: Yeah it's a pretty gross solution to be honest. I've removed it and just appended the global options. The downside is we see the redundant "Run serverpod help etc.." and then below it are the global options. Are we are ok with this for now until we have more customizable `cli_tools`?
image

FXschwartz: Per this morning's standup we are good with this "hack" and long term will look at moving this logic into the `cli_tool` package when we have more flexibility.

nielsenko (approved):
Let's go with this for now, and change once the option is available upstream

PR #5242, opened 2026-05-27
fix: Refresh the web page on Dart code changes in serverpod start

### LINKED ISSUE #5222: `serverpod start` does not reload the web page when Dart files are changed
Currently, `serverpod start` will only reload the web page if templates are changed. However, changes in the Dart code will also impact when the web needs to be reloaded. We should reload the web page on any type of server change.

FXschwartz:
In `serverpod start` watch mode, the browser was only auto-refreshed when a static file (HTML, CSS, JS, or template) changed. A change to Dart code hot-reloaded the server but left any open server-rendered web page showing stale content until it was manually reloaded.

This wires the existing browser-refresh mechanism into the Dart-change path, which now refreshes for *any* server change, not just static-file changes.

The refresh only fires when the server actually reloaded, code generation or compilation failures don't trigger a refresh to a stale page.

Fixes #5222 

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start/watch_session.dart
+++ b/tools/serverpod_cli/lib/src/commands/start/watch_session.dart
@@ -185,8 +185,7 @@ class WatchSession {
         event.modelFiles.isNotEmpty ||
         event.packageConfigChanged;
 
-    // Static-only changes (HTML, JS, CSS, templates): no compilation needed,
-    // just refresh the browser.
+    // Static-only changes (HTML, JS, CSS, templates): no compilation needed.
     if (!hasDartChanges) {
       log.debug('Static files changed.');
       await _notifyBrowserRefresh();
--- a/tools/serverpod_cli/lib/src/commands/start/watch_session.dart
+++ b/tools/serverpod_cli/lib/src/commands/start/watch_session.dart
@@ -187,7 +187,6 @@ class WatchSession {
 
     // Static-only changes (HTML, JS, CSS, templates): no compilation needed.
     if (!hasDartChanges) {
-      log.debug('Static files changed.');
       await _notifyBrowserRefresh();
       return;
     }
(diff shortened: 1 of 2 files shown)


tools/serverpod_cli/lib/src/commands/start/watch_session.dart
```diff
@@ -434,6 +454,7 @@ class WatchSession {
     return _chain(() async {
       if (await _fullCompileAndRestart()) {
         await _restartFlutter();
+        await _notifyBrowserRefresh();
```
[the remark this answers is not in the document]
FXschwartz: Test added!

tools/serverpod_cli/lib/src/commands/start/watch_session.dart
```diff
@@ -187,7 +187,8 @@ class WatchSession {
 
     // Static-only changes (HTML, JS, CSS, templates): no compilation needed.
     if (!hasDartChanges) {
-      await _handleStaticChange();
+      log.debug('Static files changed.');
```
nielsenko: This log line is not always correct. Perhaps drop it.

nielsenko (commented):
I think I misunderstood the [issue](https://github.com/serverpod/serverpod/issues/5222) when we discussed it yesterday. This does fix the issue mentioned.

One downside is that state on pages (if any) is lost, on any dart change, even if the page is not really affected. Fixing that will require detailed dependency tracking, and not something I see happening anytime soon. 

So I guess the question becomes what is the lesser evil - having to reload explicitly on dart changes, or loosing state due to page reloads?

tools/serverpod_cli/lib/src/commands/start/watch_session.dart
FXschwartz: Done!

FXschwartz:
Talked to Viktor during standup and it was decided that we are good with reloading after every dart change.

nielsenko (approved):
LGTM 🥇 (concerns regarding frequent reloads causing state loss on forms was discussed at SU)

PR #5243, opened 2026-05-27
fix: Make the interactive project selector work on Windows

### LINKED ISSUE #5172: Interactive project selection on Windows is not rendering
Whenever there is more than one server on the current directory, the CLI should invoke an interactive project selector. On Windows, this is resulting in a blank screen, with the arrows and selection working, but without rendering anything. It is possible that this bug has been introduced by [a last minute fix before 3.5.0-beta.6 release](https://github.com/serverpod/serverpod/pull/5063/commits/de490350688110f47a20f1cbc46426ab8a0518b6).

FXschwartz:
The selector shown when multiple Serverpod servers are detected was broken on Windows: blank screen, arrow keys didn't navigate, and Enter threw StdinException (errno 87).

This keeps the standard terminal prompt and fixes three Windows-specific bugs in cli_tools_select.dart:
- Flush `stdout` after each render so the menu reaches the Windows console before the blocking readByteSync();
- Restore `lineMode` before `echoMode` on exit.
- Enable `ENABLE_VIRTUAL_TERMINAL_INPUT` on Windows so cursor keys arrive as ESC[A/B (Windows Terminal/ConPTY otherwise drops them), restored on exit. Add j/k as navigation aliases, matching `mason_logger`.

Note: There was some discussion about whether or not we should go with the arrow key functionality vs listing each option tied to a number key for selection. Packages like Melos for Windows revert to just using the number approach but several other packages like Mason implement the full support to get arrow keys to render. I went with how Mason implements the support in order to keep it consistent between MacOS and Linux.

Extra Note: I do think long term we should look at using our new `serverpod_tui` for this as there will always be platform/OS edge cases that come up. But this is an easy win for now.

Fixes #5172 

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tools/serverpod_cli/lib/src/util/windows_console_input.dart
+++ b/tools/serverpod_cli/lib/src/util/windows_console_input.dart
@@ -1,0 +1,69 @@
+import 'dart:ffi';
+
+import 'package:ffi/ffi.dart';
+
+/// Enables virtual-terminal input on the Windows console so that special keys
+/// (notably the arrow keys) are delivered to `stdin` as ANSI escape sequences.
+///
+/// Windows Terminal/ConPTY otherwise reports arrow keys as character-less key
+/// events that `stdin.readByteSync()` discards, so an escape-sequence based
+/// prompt never sees them. Enabling this mode makes the cursor keys arrive as
+/// the same `ESC [ A` / `ESC [ B` sequences they do on POSIX terminals.
+///
+/// Returns the previous console mode to hand to
+/// [restoreWindowsConsoleInputMode], or `null` if the mode could not be read
+/// or changed (for example when stdin is not a console).
+///
+/// Only call this on Windows; the `kernel32.dll` bindings are loaded lazily
+/// and are unavailable on other platforms.
+int? enableWindowsVirtualTerminalInput() {
+  final handle = _getStdHandle(_stdInputHandle);
+  final modePointer = calloc<Uint32>();
+  try {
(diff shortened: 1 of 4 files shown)


tools/serverpod_cli/lib/src/util/windows_console_input.dart
```diff
@@ -0,0 +1,69 @@
...
+typedef _SetConsoleModeDart = int Function(int hConsoleHandle, int dwMode);
+
+final DynamicLibrary _kernel32 = DynamicLibrary.open('kernel32.dll');
+
+final _getStdHandle = _kernel32
+    .lookupFunction<_GetStdHandleC, _GetStdHandleDart>('GetStdHandle');
+final _getConsoleMode = _kernel32
+    .lookupFunction<_GetConsoleModeC, _GetConsoleModeDart>('GetConsoleMode');
+final _setConsoleMode = _kernel32
+    .lookupFunction<_SetConsoleModeC, _SetConsoleModeDart>('SetConsoleMode');
```
nielsenko: Should we consider depending on [win32](https://pub.dev/packages/win32) instead? Already exposes these APIs I believe.

nielsenko (approved):
I would suggest using [win32](https://pub.dev/packages/win32) instead of using FFI directly.

tools/serverpod_cli/lib/src/util/windows_console_input.dart
FXschwartz: Done! Good call

PR #5246, opened 2026-05-28
feat: Expose L key to clear logs on serverpod start

### LINKED ISSUE #5199: Expose `L` to clear logs on `serverpod start`
When running for too long, logs grow and the TUI becomes very slow. Even with optimizations that are on the way, it is still worth having a way to clear the TUI logs to get rid of older irrelevant messages.

We can expose this functionality on the `L` key to be close to CTRL/CMD+L - and also to avoid a failed CTRL to clear logs when trying to CTRL+C an error.

FXschwartz:
Adds an `L` button to the `serverpod start` TUI button bar that empties every log buffer in one keystroke, structured server log history, raw server stdout/stderr, and Flutter stdout/stderr.

I went with the behavior to clear logs across **ALL** tabs as the split pain makes it a bit difficult to understand which logs the user would want to clear, but if we would rather it just clear the current active tab it would be a very simple change. It's just then we would need to decide if the TUI has the side-by-side layout with server logs & raw server logs and then the right side is **always** flutter logs, what's the "active" tab?

Fixes #5199 

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start/tui/app.dart
+++ b/tools/serverpod_cli/lib/src/commands/start/tui/app.dart
@@ -189,6 +189,10 @@ class ServerpodWatchAppState extends TuiAppState<ServerpodWatchApp> {
         onCreateMigration: onCreateMigration,
         onCreateRepairMigration: onCreateRepairMigration,
         onApplyMigration: onApplyMigration,
+        onClearLogs: () {
+          state.clearLogs();
+          _rebuild();
+        },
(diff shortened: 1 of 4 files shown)


marcelomendoncasoares (approved):
LGTM! 🚀 

Agree with clearing all logs. In the future we'll sure have a more structured UX to allow selectively clearing, but clearing all is sufficient for now.

PR #5253, opened 2026-05-29
refactor: Replace project-tree walk with explicit paths in TemplateRenderer

nielsenko:
The old `TemplateRenderer`  took a _directory_ and walked the entire project tree, rendering everything it found. 

To avoid clobbering files it shouldn't touch and improve performance a recent change #5177 carried a hardcoded blocklist, 
but  that is still a fragile design. 

The new design flips it. `Copier` now records the exact paths it wrote (`copyFiles()` returns them), and the renderer is handed *only that explicit list* via `renderPaths(paths, context)`. It can only ever touch files the copiers actually produced.

The renderer also gained directory-rename + merge logic so a second upgrade run doesn't leave literal `{{...}}` directories on disk.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/create/template_renderer.dart
+++ b/tools/serverpod_cli/lib/src/create/template_renderer.dart
@@ -6,24 +6,27 @@ import 'package:pub_semver/pub_semver.dart';
 import 'package:serverpod_cli/src/create/template_context.dart';
 import 'package:whiskers/whiskers.dart';
 
-/// Responsible for rendering template files in a directory using provided context.
-/// It processes both file contents and directory names,
-/// allowing for dynamic project structures.
+/// Renders Mustache `{{...}}` directives in a known set of file paths
+/// (typically the files [Copier] just wrote) and their ancestor
+/// directory names.
+///
+/// Operating on an explicit path list - rather than walking the project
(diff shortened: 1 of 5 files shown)


tools/serverpod_cli/lib/src/create/template_renderer.dart
```diff
@@ -6,24 +6,27 @@ import 'package:pub_semver/pub_semver.dart';
...
+  /// anything else is left untouched as a defensive sanity guard.
+  static const Set<String> _renderableExtensions = <String>{
+    '.dart',
+    '.yaml',
+    '.yml',
+    '.json',
+    '.html',
+    '.css',
+    '.md',
+    '.svg',
```
Crazelu: Should we be rendering content in SVGs?

Crazelu (approved):
Really nice!
Only one comment on svg as a renderable extension

tools/serverpod_cli/lib/src/create/template_renderer.dart
nielsenko: Fixed in 84dbe02. No need to white list extensions when we know what file paths are to be updated.

marcelomendoncasoares (approved):

PR #5257, opened 2026-05-29
fix: Disable WASM by default new projects

### LINKED ISSUE #5211: Remove `--wasm` from default `flutter_build` command on new projects
When creating a project, we add a `serverpod run flutter_build` command that uses `--wasm` by default and will serve to build the app to deploy to Serverpod Cloud. But WASM can push unexpected COOP and COEP (cross-origin) issues on users that will only happen when they try to deploy their apps. Since the issue won't happen during development and the option is never asked, it gets very hard to debug, specially for developers that are less familiar with Flutter web.

To avoid this kind of issue, we should remove the `--wasm` by default, which aligns with Flutter still treating WASM as opt-in instead of default.

marcelomendoncasoares:
Closes: #5211

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod/lib/src/web_server/routes/flutter_route.dart
+++ b/packages/serverpod/lib/src/web_server/routes/flutter_route.dart
@@ -4,10 +4,10 @@ import 'dart:io';
 import 'package:path/path.dart' as path;
 import 'package:serverpod/serverpod.dart';
 
-/// Route for serving Flutter web applications with WASM support.
+/// Route for serving Flutter web applications.
 ///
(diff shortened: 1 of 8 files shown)


nielsenko (approved):
LGTM, except we should perhaps also remove the `--wasm` flag in pubspec template's `flutter_build ` script?

marcelomendoncasoares:
> LGTM, except we should perhaps also remove the `--wasm` flag in pubspec template's `flutter_build ` script?

It is removed! For both the project and project upgrade.

PR #5258, opened 2026-05-29
fix: Surface serverpod start crashes to the terminal on exit

### LINKED ISSUE #5217: No error message when `serverpod start` crashes
Currently, there is no error message when `serverpod start` crashes. If something happens, we should exit TUI mode and write the stack trace to the terminal so that we have a chance of figuring out what went wrong.

Consideration when fixing. Make the error message fun and friendly (like the "Yikes" message we currently have). Users get less upset when messages are friendly. 😅

FXschwartz:
When `serverpod start` crashed while running in TUI mode, nothing was printed to the terminal.

This change captures the fatal error and stack trace, tears the TUI down so we leave the alternate screen, and then replays the crash to the restored `stdout` logger via a `ServerpodTerminalBackend` `preExit` hook.

This mirrors the strategy `serverpod create` already uses to flush errors to the console after exiting the alternate screen, so the crash survives in the user's scrollback.

Fixes #5217 

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -31,6 +31,7 @@ import 'package:serverpod_cli/src/migrations/create_migration_action.dart';
 import 'package:serverpod_cli/src/migrations/create_repair_migration_action.dart';
 import 'package:serverpod_cli/src/runner/serverpod_command.dart';
 import 'package:serverpod_cli/src/runner/serverpod_command_runner.dart';
+import 'package:serverpod_cli/src/util/internal_error.dart';
 import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart';
 import 'package:serverpod_cli/src/vm_proxy/proxy.dart';
 import 'package:serverpod_cli/src/vm_proxy/serverpod_hooks.dart';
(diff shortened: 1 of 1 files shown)


tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -946,35 +944,55 @@ Future<int> _runWithTui({
...
+
+  // Runs after the TUI tears down (alternate screen restored) but before the
+  // process exits. Restores the stdout logger and replays any captured crash
+  // so it survives in the user's scrollback - mirrors how `serverpod create`
+  // flushes its errors to the terminal on exit.
+  Future<void> preExit() async {
+    final crash = fatalCrash;
+    if (crash == null) return;
+    await closeLogger();
+    initializeLogger();
```
marcelomendoncasoares, line 992: Suggestion: Document that this close-then-open is to replace the TUI-backed logger by a stdout-backed one. Without this context it can be confusing for future developers to understand.

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -946,35 +944,55 @@ Future<int> _runWithTui({
...
-          commandConfig: commandConfig,
-          watch: watch,
-          launchFlutterApp: launchFlutterApp,
-          flutterDevice: flutterDevice,
-          flutterExtraArgs: flutterExtraArgs,
-          serverArgs: serverArgs,
-          config: config,
-          shutdown: shutdown,
-        ).catchError((Object e, StackTrace st) {
-          // Show the error in the TUI.
```
marcelomendoncasoares, line 971: Discussion: This PR made the TUI break and no longer show the error from within. But this means that users no longer see the error with the history that came before. A better UX would be to show the error inside the TUI (without quitting, as previously), but also push the error to the terminal after quitting for it to be accessible.

Image

marcelomendoncasoares (commented):
Only one minor suggestion and one topic for discussion on the UX.

tools/serverpod_cli/lib/src/commands/start.dart
FXschwartz: Yeah you're right this is better. Done! I kept the in-TUI behavior so the error now shows inside the TUI alongside the preceding logs without the auto quit and the crash is still captured and replayed to the terminal on quit.

tools/serverpod_cli/lib/src/commands/start.dart
FXschwartz: Done!

marcelomendoncasoares (approved):
LGTM!

PR #5261, opened 2026-06-01
fix: Fix error when closing method stream command with error after already closed

### LINKED ISSUE #4948: StateError: Cannot add event after closing in callStreamingServerEndpoint
### Describe the bug

When a client cancels a streaming endpoint connection (e.g. user navigates away, app goes to background), the server may still deliver a final event or error. The `catchError` callback in `callStreamingServerEndpoint` calls `outputController.addError()` on an already-closed `StreamController`, throwing:

```
StateError: Bad state: Cannot add event after closing
```

This is a stream lifecycle race condition — the controller is closed via `onCancel` → `_tryCloseInboundStream` → `_closeControllers` before the asynchronous `catchError` fires.

### To reproduce

1. Connect to a streaming server endpoint via `callStreamingServerEndpoint`
2. Cancel the stream quickly (navigate away, dispose the widget, or background the app)
3. The server sends a final error/event after the client has already closed the `outputController`
4. The `catchError` callback at `serverpod_client_shared.dart:620` calls `addError()` on the closed controller → `StateError`

This is timing-dependent and most easily observed with Sentry/crash reporting in a production or staging build. We hit it on iOS when the app went to background during an active stream.

### Expected behavior

Late messages arriving on an already-closed `StreamController` should be silently dropped rather than throwing a `StateError`. The `_closeControllers` helper already checks `!c.isClosed` before calling `close()`, but the `add`/`addError` calls that precede it do not have equivalent guards.

### Library version

- `serverpod_client`: 3.4.1 (also verified unpatched in 3.4.2)
- Dart: 3.11.3 (stable)
- Flutter: 3.41.5 (stable)

### Platform information

- iOS 26.4 (iPhone 17,1, arm64)
- Observed in release/AOT build via Sentry crash reporting
- Likely affects all platforms since the bug is in pure Dart client code

### Additional context

There are 3 unguarded call sites total:

| File | Line | Call | Context |
|------|------|------|---------|
| `serverpod_client_shared.dart` | 620 | `outputController.addError(error)` | `catchError` callback in `callStreamingServerEndpoint` |
| `client_method_stream_manager.dart` | 451 | `controller.add(message.object)` | `_dispatchMessage` |
| `client_method_stream_manager.dart` | 483 | `controller.addError(message.exception)` | `_dispatchSerializableException` |

Additionally, `_tryCloseInboundStream` (line 389) calls `addError(ConnectionClosedException())` without an `isClosed` guard when `reason == CloseReason.error`.

Note that `_closeControllers` (line 332) already guards with `!c.isClosed` — the pattern exists, it's just not applied before the `add`/`addError` calls.

Stacktrace from Sentry:

```
at ServerpodClientShared.callStreamingServerEndpoint.<fn> (serverpod_client_shared.dart:620)
at _StreamController.addError (stream_controller.dart:623)
```

We have an open PR with a minimal fix: #4939

Intermediate

marcelomendoncasoares:
Fixes: #4948

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod_client/lib/src/client_method_stream_manager.dart
+++ b/packages/serverpod_client/lib/src/client_method_stream_manager.dart
@@ -385,7 +385,8 @@ final class ClientMethodStreamManager {
     // a close message to the server.
     inboundStreamContext.controller.onCancel = null;
 
-    if (reason == CloseReason.error) {
+    if (reason == CloseReason.error &&
+        !inboundStreamContext.controller.isClosed) {
       inboundStreamContext.controller.addError(
         const ConnectionClosedException(),
       );
--- a/packages/serverpod_client/lib/src/client_method_stream_manager.dart
+++ b/packages/serverpod_client/lib/src/client_method_stream_manager.dart
@@ -229,6 +229,7 @@ final class ClientMethodStreamManager {
 
     if (exception != null) {
       for (var c in inputControllers) {
+        if (c.isClosed) continue;
         c.addError(exception);
       }
     }
@@ -445,7 +446,8 @@ final class ClientMethodStreamManager {
     );
 
     var inboundStreamContext = _inboundStreams[inboundStreamKey];
-    if (inboundStreamContext == null) {
+    if (inboundStreamContext == null ||
+        inboundStreamContext.controller.isClosed) {
       return false;
     }
 
@@ -481,7 +483,9 @@ final class ClientMethodStreamManager {
       return;
     }
 
-    inboundStreamContext.controller.addError(message.exception);
+    if (!inboundStreamContext.controller.isClosed) {
+      inboundStreamContext.controller.addError(message.exception);
+    }
     await _closeControllers([inboundStreamContext.controller]);
   }
 
--- a/packages/serverpod_client/lib/src/serverpod_client_shared.dart
+++ b/packages/serverpod_client/lib/src/serverpod_client_shared.dart
@@ -622,7 +622,9 @@ abstract class ServerpodClientShared extends EndpointCaller {
         error = e;
       }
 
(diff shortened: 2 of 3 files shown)


packages/serverpod_client/lib/src/client_method_stream_manager.dart
```diff
@@ -385,7 +385,8 @@ final class ClientMethodStreamManager {
     // a close message to the server.
     inboundStreamContext.controller.onCancel = null;
 
-    if (reason == CloseReason.error) {
+    if (reason == CloseReason.error &&
+        !inboundStreamContext.controller.isClosed) {
```
nielsenko, line 390: This fixes `_tryCloseInboundStream` but there are other places we may be adding events to an already closed stream. Are we sure this is the full story? 

What about:
- serverpod_client_shared.dart:625
- client_method_stream_manager.dart:451
- client_method_stream_manager.dart:483

I have a hunch you analyzed this, since you closed the original PR - but just making sure.

nielsenko (commented):
Some questions

packages/serverpod_client/lib/src/client_method_stream_manager.dart
marcelomendoncasoares: I preferred keeping the fix scoped to the issue, so I did not hunt other places. But, given your suggestion, I have expanded the scope with tests (https://github.com/serverpod/serverpod/pull/5261/commits/f3d15b339864d9aa75890e10af159881292f2766) and fixed the other occurrences (https://github.com/serverpod/serverpod/pull/5261/commits/8ea6a30d9d84816422b14e0f41b5e6e65f0a3a00).

nielsenko (approved):
LGTM

PR #5264, opened 2026-06-01
fix: Add the missing `--empty` flag to `create-migration`

### LINKED ISSUE #4999: Question about reviving `create-migration --empty` from PR #3978
Hi Serverpod team,

I saw PR #3978, which added support for `serverpod create-migration --empty`, and noticed that it was eventually closed after some review discussion.

I wanted to ask whether this is still something the team is interested in supporting.

The `--empty` flag would be really useful for advanced database use cases where a project still uses Serverpod’s generated models/client/ORM helpers, but also needs to add a small amount of custom SQL manually, such as:

- generated columns
- full-text-search support
- GIN indexes
- partial indexes
- other PostgreSQL-specific objects

In those cases, it would be helpful to have Serverpod create a valid migration folder with the correct structure and registry metadata, without requiring a model/schema diff first.

I saw in the PR review that migration E2E tests should probably follow the existing `tests/serverpod_test_server/test_e2e_migrations` pattern and be run through `util/run_tests_migrations_e2e`.

Would you be open to a new PR that revives the `--empty` migration part of #3978 with the requested tests?

Thanks!

marcelomendoncasoares:
Closes: #4999

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tests/serverpod_cli_e2e_test/test/create_migration_exit_code_test.dart
+++ b/tests/serverpod_cli_e2e_test/test/create_migration_exit_code_test.dart
@@ -56,6 +56,35 @@ void main() async {
           );
         },
       );
+
+      test(
+        'when create-migration is called with --empty then an empty migration is created.',
+        () async {
+          var migrationsDirectory = Directory(
+            path.join(serverDir, 'migrations'),
+          );
+          var migrationsBefore = migrationsDirectory
+              .listSync()
+              .whereType<Directory>()
+              .length;
+
+          var result = await runServerpod(
+            ['create-migration', '--empty'],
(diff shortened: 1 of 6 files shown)


nielsenko (approved):
LGTM. One thing that I was thinking was if `allowEmpty` would be a better word that `empty` here, but I'll jut mention it as an option.

marcelomendoncasoares:
> LGTM. One thing that I was thinking was if `allowEmpty` would be a better word that `empty` here, but I'll jut mention it as an option.

I used the `--empty` as we defined on the RFC and to keep short/consistent with `--force`.

PR #5266, opened 2026-06-01
feat: Remove the need for the raw server logs tab in serverpod start

### LINKED ISSUE #5223: Remove the need for raw server logs in `serverpod start`
We should remove the need for a raw server log tab in the TUI.

**Exceptions**
Either make exceptions expandable in the TUI, or just print the full stack trace?

**Possible solution for `stdout` 1**
When using `serverpod start`, configure the logger to not log to `stdout`. Then, we can pass all `stdout` messages to the pretty log.

**Possible solution for `stdout` 2**
Just ignore the stdout in the pretty log and add a rule to the `analysis_options.yaml` to treat `print` as an error.

FXschwartz:
Removes the raw server logs tab from the serverpod start TUI now that stack traces surface in the structured view.

- Attaches stackTrace from ext.serverpod.log events to the log entry and renders it in the Server logs tab collapsed to a ▸ N-line stack trace affordance, toggled inline with e.
- Moves raw server logs from a tab to a backtick (`) overlay, Esc or ` closes it, listed in the help panel.
- Collapses the tabs to Server logs + Flutter logs, which also fixes the empty buffer on no Flutter raw tab and the two swapped scroll controllers.

Fixes #5223 

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start/tui/state.dart
+++ b/tools/serverpod_cli/lib/src/commands/start/tui/state.dart
@@ -23,8 +23,7 @@ class ServerWatchState extends TuiState {
   /// Currently selected tab index.
   ///
   /// - 0 = structured server logs
-  /// - 1 = Flutter logs (narrow layout only)
-  /// - 2 = raw server logs
+  /// - 1 = Flutter logs (only when [showFlutterOutput])
   int selectedTab = 0;
 
   /// Latest measured content width from the main log area.
@@ -67,6 +66,12 @@ class ServerWatchState extends TuiState {
   /// When false, an error entry that carries a trace shows a compact
   /// affordance instead; toggled with `e` on the structured log tab.
   bool expandStackTraces = false;
+
+  /// Whether the raw server logs overlay (the "dev console") is visible.
+  ///
+  /// Raw server stdout/stderr is hidden from the default tabs and reached
+  /// on demand via the backtick (`` ` ``) shortcut.
+  bool showRawServerLogs = false;
 
   /// Maximum number of log entries to keep.
   static const maxLogEntries = 10000;
--- a/tools/serverpod_cli/lib/src/commands/start/tui/state.dart
+++ b/tools/serverpod_cli/lib/src/commands/start/tui/state.dart
@@ -9,11 +9,12 @@ class ServerWatchState extends TuiState {
   @override
   final logHistory = BoundedQueueList<Object>(maxLogEntries);
 
-  /// Raw stdout/stderr lines for the "Raw server output" tab.
+  /// Raw stdout/stderr lines shown in the raw server logs overlay
+  /// (toggled with the backtick shortcut).
   @override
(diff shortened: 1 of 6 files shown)


tools/serverpod_cli/lib/src/commands/start/tui/state.dart
```diff
@@ -23,8 +23,7 @@ class ServerWatchState extends TuiState {
   /// Currently selected tab index.
   ///
   /// - 0 = structured server logs
-  /// - 1 = Flutter logs (narrow layout only)
-  /// - 2 = raw server logs
+  /// - 1 = Flutter logs (only when [showFlutterOutput])
   int selectedTab = 0;
```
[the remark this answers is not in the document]
FXschwartz: Done!

marcelomendoncasoares:
> * Moves raw server logs from a tab to a backtick (\`) overlay, Esc or ` closes it, listed in the help panel.

The backtick might not be a good key because it can be rather inaccessible on keyboard layouts of other languages. In PT-BR, for example, it is shift+(key at right of P)+space. Maybe `.` or `ctrl/cmd`+`.` is a more universal key?

tools/serverpod_cli/lib/src/commands/start/tui/state.dart
```diff
@@ -62,6 +62,18 @@ class ServerWatchState extends TuiState {
...
+  /// Whether stack traces attached to log entries are shown inline.
+  ///
+  /// When false, an error entry that carries a trace shows a compact
+  /// affordance instead; toggled with `e` on the structured log tab.
+  bool expandStackTraces = false;
+
+  /// Whether the raw server logs overlay (the "dev console") is visible.
+  ///
+  /// Raw server stdout/stderr is hidden from the default tabs and reached
+  /// on demand via the backtick (`` ` ``) shortcut.
```
nielsenko: Will this render in docs as you would like? Alternative
```suggestion
  /// on demand via the backtick (```) shortcut.
```
Anyway TIL.

nielsenko (approved):
LGTM 👍

FXschwartz:
> > * Moves raw server logs from a tab to a backtick (`) overlay, Esc or ` closes it, listed in the help panel.
> 
> The backtick might not be a good key because it can be rather inaccessible on keyboard layouts of other languages. In PT-BR, for example, it is shift+(key at right of P)+space. Maybe `.` or `ctrl/cmd`+`.` is a more universal key?

@marcelomendoncasoares How would you feel about supporting both? We could show both in the help screen? 

Backtick feels pretty standard with the applications I use but totally get it's not accessible on other keyboard layouts so it would be nice if we could support both.

tools/serverpod_cli/lib/src/commands/start/tui/state.dart
FXschwartz: Updated to include both backticks and period key bindings.

PR #5267, opened 2026-06-02
refactor: Remove the assumption that the migration list on `RuntimeListMigrationArtifactStore` is sorted

### LINKED ISSUE #5031: Refactor RuntimeListMigrationArtifactStore to use SplayTreeSet with comparator
### Context

In PR #4994 (comment: https://github.com/serverpod/serverpod/pull/4994#discussion_r3172652855), it was suggested to improve `RuntimeListMigrationArtifactStore` in `packages/serverpod_database/lib/src/migrations/migration_artifacts_store/runtime_list.dart`.

### Proposed Improvement

Replace the current `Map<version, MigrationVersionSql>` backing store with a `SplayTreeSet<MigrationVersionSql> _migrations` that uses a proper compare function. This would:

1. Ensure migrations are always stored in sorted order.
2. Remove the need for a separate assert that migrations are sorted.
3. Provide efficient ordered traversal of migrations.

### Requested by

@nielsenko

marcelomendoncasoares:
Closes: #5031

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod_database/lib/src/migrations/migration_artifacts_store/runtime_list.dart
+++ b/packages/serverpod_database/lib/src/migrations/migration_artifacts_store/runtime_list.dart
@@ -6,7 +6,7 @@ class RuntimeListMigrationArtifactStore
     implements MigrationArtifactStoreReader {
   /// Creates a store from the given [migrations] (same [moduleName] for all).
   RuntimeListMigrationArtifactStore(
-    this.migrations, {
+    List<MigrationVersionSql> migrations, {
     required this.moduleName,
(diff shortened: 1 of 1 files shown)


FXschwartz (approved):
Looks good to me! Only thing to flag is that this looks like its newer unreleased code. If not then this looks like it would be a breaking change.

marcelomendoncasoares:
> Looks good to me! Only thing to flag is that this looks like its newer unreleased code. If not then this looks like it would be a breaking change.

Yes, it is unreleased code. And, since the `migrations` list is not public on the filesystem store, we can hide it in the runtime list as well.

PR #5268, opened 2026-06-02
fix: Throw eagerly if invalid chars are used as migration tag name

### LINKED ISSUE #5029: Sanitize migration version tag to avoid path-breaking characters
## Summary

When creating a client-side database migration, the version name is built by `createVersionName` which appends an unsanitized `tag` directly to the formatted timestamp:

```dart
static String createVersionName(String? tag) {
  var now = DateTime.now().toUtc();
  var fmt = DateFormat('yyyyMMddHHmmssSSS');
  var versionName = fmt.format(now);
  if (tag != null) {
    versionName += '-$tag';
  }
  return versionName;
}
```

Characters such as spaces, backslashes, quotes (`'`, `"`), and others could be passed in via `tag`, breaking the generated file paths or the Dart `part` directives that reference the version.

The `_partUri` function in `client_migration_dart_emitter.dart` already contains a partial guard for `..` and `\`, but the root cause should be addressed upstream by sanitizing the `tag` inside `createVersionName`.

## Expected Behaviour

The `tag` parameter should be sanitized before being appended to the version name, replacing or rejecting characters that are invalid in file system paths or Dart URI strings (e.g. spaces, `\`, `/`, `'`, `"`, `:`, `*`, `?`, `<`, `>`, `|`).

A richer `Version` object (rather than a plain `String`) could also be considered to enforce invariants at the type level, as suggested in the original review.

## Context

Raised in PR #4994 (review comment): https://github.com/serverpod/serverpod/pull/4994#discussion_r3172712087

Requested by @marcelomendoncasoares.

marcelomendoncasoares:
Closes: #5029

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/migrations/generator.dart
+++ b/tools/serverpod_cli/lib/src/migrations/generator.dart
@@ -34,7 +34,7 @@ class MigrationGenerator {
     var fmt = DateFormat('yyyyMMddHHmmssSSS');
     var versionName = fmt.format(now);
     if (tag != null && tag.isNotEmpty) {
-      if (RegExp(r'[/\\:*?"<>|\x00-\x1f]').hasMatch(tag)) {
+      if (!RegExp(r'^[a-zA-Z0-9_-\s.]+$').hasMatch(tag)) {
         throw FormatException(
           'Invalid migration tag: "$tag". Only characters that are supported '
           'in file names are allowed.',
--- a/tools/serverpod_cli/lib/src/migrations/generator.dart
+++ b/tools/serverpod_cli/lib/src/migrations/generator.dart
@@ -33,7 +33,13 @@ class MigrationGenerator {
     var now = DateTime.now().toUtc();
     var fmt = DateFormat('yyyyMMddHHmmssSSS');
     var versionName = fmt.format(now);
-    if (tag != null) {
+    if (tag != null && tag.isNotEmpty) {
+      if (RegExp(r'[/\\:*?"<>|\x00-\x1f]').hasMatch(tag)) {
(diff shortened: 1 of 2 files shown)


tools/serverpod_cli/lib/src/migrations/generator.dart
```diff
@@ -33,7 +33,13 @@ class MigrationGenerator {
     var now = DateTime.now().toUtc();
     var fmt = DateFormat('yyyyMMddHHmmssSSS');
     var versionName = fmt.format(now);
-    if (tag != null) {
+    if (tag != null && tag.isNotEmpty) {
+      if (RegExp(r'[/\\:*?"<>|\x00-\x1f]').hasMatch(tag)) {
```
nielsenko: What about `\x7f` (DEL)? Only non-printable ascii char above `\x1f`.
Anyway, what about a whitelist instead?
```suggestion
      if (!RegExp(r'^[a-zA-Z0-9_-]+$').hasMatch(tag)) {
```
Then I don't have to think so hard 😄

nielsenko (approved):
I suggest using a white list of characters instead. Alternatively also block `\x7f`.
Otherwise LGTM.

tools/serverpod_cli/lib/src/migrations/generator.dart
marcelomendoncasoares: I initially went with this subset, but then it would be a behavioral breaking change for users that are using spaces, dots and other valid chars for file names as part of their tags. However, considering that we will now be bumping major, it might be a good time to do it.

PR #5271, opened 2026-06-02
feat: Run integration tests on embedded PostgreSQL

nielsenko:
Run integration tests on embedded PostgreSQL, in parallel (no Docker)

`withServerpod` now boots an embedded PostgreSQL on demand and gives each test group its own database on a shared postmaster. Isolation comes from per-group databases (Postgres) / files (SQLite).

This gets rid of most uses of `concurrency_one` tags.

`serverpod_test_server` project not updated in this PR. This is addressed in the follow-up PR: https://github.com/serverpod/serverpod/pull/5402.


nielsenko:
Converted to draft to address feedback from Friday Cycle Wrap.

nielsenko:
> Converted to draft to address feedback from Friday Cycle Wrap.

The feedback will be addressed in a separate pr sitting on top of #5381 instead. We need pg embedded to support PostGIS and pgvector to complete this work, and doing surgery on history to move that PR below this one proved too much of a hurdle.

PR #5278, opened 2026-06-02
fix: Ignore `runtimeParametersBuilder` on SQLite instead of throwing `UnsupportedError`

marcelomendoncasoares:
Fixes an asymmetry on the behavior of SQLite regarding runtime parameters (no-op on transactions, throw on global level). With this change, both places will be a no-op, with a proper warning on the global level.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod_database/lib/src/adapters/sqlite/database_provider.dart
+++ b/packages/serverpod_database/lib/src/adapters/sqlite/database_provider.dart
@@ -1,3 +1,4 @@
+import 'package:serverpod_shared/log.dart';
 import 'package:serverpod_shared/serverpod_shared.dart';
 
 import '../../../serverpod_database.dart';
@@ -24,7 +25,9 @@ class SqliteDatabaseProvider implements DatabaseProvider {
     SqliteDatabaseConfig config,
   ) {
     if (runtimeParametersBuilder != null) {
-      throw UnsupportedError('SQLite does not support runtime parameters.');
+      log.warning(
+        'Runtime parameters are not supported on SQLite and will be ignored.',
+      );
     }
     return SqlitePoolManager(
       serializationManager,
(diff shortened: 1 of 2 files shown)


nielsenko:
Is this really better? Should it be throw on transactions as well? (Now we are bumping major)

marcelomendoncasoares:
> Is this really better? Should it be throw on transactions as well? (Now we are bumping major)

The idea for being a no-op is to allow users to seamlessly switch the database dialect without breaking the code. Postgres remain the first-class, with SQLite being a best effort. It is opened for discussion, though.

nielsenko:
> > Is this really better? Should it be throw on transactions as well? (Now we are bumping major)
> 
> The idea for being a no-op is to allow users to seamlessly switch the database dialect without breaking the code. Postgres remain the first-class, with SQLite being a best effort. It is opened for discussion, though.

But wouldn't it just be silent breakage instead? Wouldn't fail-fast-hard be preferable?

I agree it should be consistent. Just thinking that perhaps it is better to break that limb on?

marcelomendoncasoares:
> > The idea for being a no-op is to allow users to seamlessly switch the database dialect without breaking the code. Postgres remain the first-class, with SQLite being a best effort. It is opened for discussion, though.
> 
> But wouldn't it just be silent breakage instead? Wouldn't fail-fast-hard be preferable?
> 
> I agree it should be consistent. Just thinking that perhaps it is better to break that limb on?

Not for the type of configs one can set using `RuntimeParameterBuilder`. The ones that exist right now are all for tweaking vector indexes behavior (indexes that we skip creating with a warning on migrations), tweaking indexing performance, setting search paths, etc.

nielsenko (approved):

PR #5279, opened 2026-06-03
perf: Use staleness check to avoid redundant work on generate

### LINKED ISSUE #4746: Continuous generator: Handle directory creation/deletion events
## Problem

In the continuous generator's watch mode, directory creation/deletion events may not trigger file events on all platforms. This means that when a directory is created or deleted, the individual file events within that directory may not be properly detected.

## Context

This issue was identified in the file `tools/serverpod_cli/lib/src/generator/generator_continuous.dart` during review of PR #4737.

The current implementation uses `DirectoryWatcher` which may not reliably capture all file changes when directories themselves are added or removed from the file system.

## Consideration

We should consider traversing the subtree on directory add and pruning the full subtree on directory delete, as we cannot depend on individual file events in these scenarios.

## Related

- PR: https://github.com/serverpod/serverpod/pull/4737
- Comment: https://github.com/serverpod/serverpod/pull/4737#discussion_r2820788935
- Reported by: @nielsenko

### LINKED ISSUE #5231: Run `serverpod generate` through `serverpod start`
If `serverpod start` is running, we could run `serverpod generate` via `serverpod start`, which would make it much lighter and faster (no need to spin up the Dart analyzer).

nielsenko:
Closes: https://github.com/serverpod/serverpod/issues/4746

- skip generate when output is already up to date
- track all generation inputs and outputs in the staleness check
- enumerate and stats sources once in the staleness check

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/generate.dart
+++ b/tools/serverpod_cli/lib/src/commands/generate.dart
@@ -158,12 +158,12 @@ class GenerateCommand extends ServerpodCommand<GenerateOption> {
 /// Generates code if it is stale, returning whether it was already [upToDate]
 /// and, when generation ran, whether it reported [success].
 ///
-/// [createAnalyzers] must return an *unprimed* analyzer (e.g.
-/// [Analyzers.create]); the single full prime happens inside [analyzeAndGenerate].
+/// [createAnalyzers] must return an unprimed analyzer (ie. [Analyzers.create]).
+/// The priming happens inside [analyzeAndGenerate].
 ///
-/// When [keepPrimedWhenFresh] is `false` (one-shot generate) a fresh project
-/// returns before the analyzer is created. When `true` (the watch loops) the
-/// analyzer is always created and primed for the incremental loop.
+/// When [keepPrimedWhenFresh] is `false` a fresh project returns before the
+/// analyzer is created. When `true` the analyzer is always created and primed
+/// for an incremental loop.
 Future<({bool upToDate, bool success})> generateIfStale({
   required GeneratorConfig config,
   required Future<Analyzers> Function() createAnalyzers,
@@ -180,8 +180,9 @@ Future<({bool upToDate, bool success})> generateIfStale({
   final result = await analyzeAndGenerate(
     config: config,
     analyzers: analyzers,
-    affectedPaths: allSources,
+    affectedPaths: allSources.keys.toSet(),
     verifyStaleness: keepPrimedWhenFresh,
+    sourceStats: allSources,
   );
   // A full run only returns an empty file set when it skipped generation
   // because the output was already up to date.
@@ -225,6 +226,7 @@ Future<GenerateResult> analyzeAndGenerate({
   required Set<String> affectedPaths,
   bool incremental = false,
   bool verifyStaleness = true,
+  Map<String, FileStamp>? sourceStats,
   GenerationRequirements requirements = GenerationRequirements.full,
 }) async {
   bool needsGenerate = false;
@@ -239,7 +241,8 @@ Future<GenerateResult> analyzeAndGenerate({
   if (incremental) {
     if (!needsGenerate) return (success: true, generatedFiles: <String>{});
   } else if (verifyStaleness &&
-      await isGenerationUpToDate(config, affectedPaths)) {
+      sourceStats != null &&
+      await isGenerationUpToDate(config, sourceStats)) {
     log.debug('Generated code is up to date, skipping.');
     return (success: true, generatedFiles: <String>{});
   }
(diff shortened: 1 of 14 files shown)


nielsenko:
Rebased on main

tools/serverpod_cli/lib/src/generator/generation_staleness.dart
```diff
@@ -29,14 +44,18 @@ Future<bool> isGenerationUpToDate(
...
   final lines = content.split('\n');
+  if (lines.length < 3) return false;
   if (lines.first.trim() != templateVersion) return false;
 
-  // Remaining lines (after version and timestamp) are generated file paths.
-  // If any are missing, generation must run.
-  if (lines.length > 2) {
-    final generatedFiles = lines.skip(2).where((l) => l.isNotEmpty);
+  // A fingerprint mismatch means a source file was added or removed.
+  if (lines[2].trim() != _sourceSetFingerprint(paths)) return false;
```
FXschwartz: We should probably also add the server `pubspec.yaml/pubspec.lock` and shared-model-package pubspecs to the checks alongside `generator.yaml` right? Since modules are discovered from the `pubspec/package` config and flow into the generated output. Otherwise I think if a user runs `serverpod generate` and then adds a module then the next `generate` wouldn't detect anything that changed.

tools/serverpod_cli/lib/src/commands/generate.dart
```diff
@@ -155,28 +155,62 @@ class GenerateCommand extends ServerpodCommand<GenerateOption> {
...
+  }
+
+  final analyzers = await createAnalyzers();
+  // Re-verify staleness only when we skipped the up-front check; the one-shot
+  // path already verified it, so generation runs unconditionally.
   final result = await analyzeAndGenerate(
     config: config,
     analyzers: analyzers,
     affectedPaths: allSources,
-    // Always run generation to prevent bad output from the watch command to be
```
FXschwartz, line 168: This comment was guarding against trusting the stamp when the output itself is bad and I think that can still happen. 

The new check verifies generated files exist, but not that they're untouched. Reverting `lib/src/generated/` with git, or hand-editing a generated file, leaves a valid stamp, and `serverpod generate` would probably say "already up to date" instead of fixing it. 

Maybe we should add a --force flag so there's still an easy way to regenerate from scratch?

tools/serverpod_cli/lib/src/generator/generation_staleness.dart
```diff
@@ -74,19 +93,26 @@ Future<bool> isGenerationUpToDate(
...
+  Set<String>? sourceFiles,
 }) async {
+  final sources = sourceFiles ?? await enumerateSourceFiles(config);
+  final fingerprint = _sourceSetFingerprint(sources);
   final file = File(_stampFilePath(config));
   await file.create(recursive: true);
   final buf = StringBuffer()
     ..writeln(templateVersion)
-    ..writeln(DateTime.now().toIso8601String());
+    ..writeln(DateTime.now().toIso8601String())
```
FXschwartz: I think there's a small race here. The stamp's mtime is set when generation finishes, but the sources were read when it started.

If a file is saved while generation is running, it wasn't included in the run, but its mtime is still older than the stamp, so the next `serverpod generate` the edit is never picked up. The old one always regenerated. 

Maybe we could backdate the stamp to when source enumeration started (setLastModified)? Then an edit mid generation just triggers one extra regeneration instead of potentially being silently missed.

FXschwartz (changes requested):

tools/serverpod_cli/lib/src/commands/generate.dart
nielsenko: Kind of the purpose of the PR to get rid of this 😏 

Anyway you are right about editing generated files, and in particular the git switch/revert is a real issue.
I'll change the hash to include _mtime_ and _size_ for all _source_ AND _target_ paths as well. This will catch these situations at very little cost.

I'll also relent and add the `--force` flag on `serverpod generate`.

tools/serverpod_cli/lib/src/generator/generation_staleness.dart
nielsenko: Good catch 👍 . I'll change `enumerateSourceFiles` to return `(path, mtime)` pairs, so we have the _mtime_ from when we start the read in the hash.

tools/serverpod_cli/lib/src/generator/generation_staleness.dart
nielsenko: point taken 🤔

nielsenko: Fixed in 8d906df

tools/serverpod_cli/lib/src/commands/generate.dart
nielsenko: Fixed in 8d906df (`--force` in eec2e204)

tools/serverpod_cli/lib/src/generator/generation_staleness.dart
nielsenko: Fixed in 8d906df (also considers _size_ since it can mitigate same-tick edits somewhat)

FXschwartz (approved):
LGTM! Sorry we beat you into submission on the `--force` topic 😅

PR #5281, opened 2026-06-03
perf: Avoid running flutter_build on create

### LINKED ISSUE #5232: Building the Flutter app on `serverpod create` makes users pay the compilation time twice
The introduction of the `flutter build` during `serverpod create` was to allow a create -> launch workflow for quickly getting up on cloud. But, with then new `start` command, our expectation is that users will create -> start instead, and the `start` command will build the Flutter app again.

To avoid pushing the cost of a double-build time (which is up to 90s each), we should skip building on create - at least until we grab ownership of the build that happen before running the Flutter app for a faster startup.

nielsenko:
Fixes #5232 

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/create/create.dart
+++ b/tools/serverpod_cli/lib/src/create/create.dart
@@ -6,7 +6,6 @@ import 'dart:io';
 import 'dart:math' as math;
 
 import 'package:cli_tools/cli_tools.dart';
-import 'package:cli_tools/execute.dart';
 import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
 import 'package:serverpod_cli/src/config/experimental_feature.dart';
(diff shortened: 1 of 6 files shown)


nielsenko:
Rebased on main

tools/serverpod_cli/lib/src/create/create.dart
```diff
@@ -253,53 +249,10 @@ Future<String?> performCreate(
...
-        ]);
-
-        return exitCode == 0;
-      },
-    );
-  }
+  // The Flutter web app is intentionally not built here. The `serverpod start`
+  // command builds the Flutter app before running it, so building during
+  // `create` would make users pay the (up to ~90s) compilation cost twice.
+  // See https://github.com/serverpod/serverpod/issues/5232.
```
marcelomendoncasoares: Suggestion: No need for this comment, since we're completely removing the reference for building the Flutter app during create.

marcelomendoncasoares (approved):
Just one nit, but LGTM! 🚀

tools/serverpod_cli/lib/src/create/create.dart
nielsenko: Done

PR #5283, opened 2026-06-03
feat: Automatically restart the Flutter app when its dependencies change

### LINKED ISSUE #5167: Quitting `serverpod start` from the terminal does not end the Flutter app

### LINKED ISSUE #5225: Make it possible for `serverpod start` to run `pub get` and recompile the app
For some actions, such as adding packages or assets to your project, a hot restart may not be sufficient. Ideally, we want to detect these cases and recompile and restart the Flutter app without restarting `serverpod start`. There needs to be a clear indication of what's happening in the interface.

FXschwartz:
On `pub get`, the watcher classifies how the Flutter app's dependency closure changed and responds with the step that picks the change up:

  | Change | Response |
  | --- | --- |
  | No change to closure (server-only deps, no-op pub get) | nothing |
  | Pure-Dart deps (and removals) | hot restart (~1s) |
  | Native code (pluginClass / ffiPlugin / build hook) | full relaunch |

One minor change I snuck into this PR is a fix for misleading logs for when the flutter app restarts via CTRL+R or via this automated path.

Fixes #5225 #5167 

**NOTE:** Before this is merged in and completed It needs a new version of `serverpod_tui` with this [PR](https://github.com/serverpod/serverpod_tui/pull/13) merged. 

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -584,6 +584,9 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
   // Calling while previous instance is still running is a no-op.
   FlutterProcess? flutterProcess;
   var spawnInFlight = false;
+  // Set by the restart action when the spawn replaces a previously running
+  // app, so the progress message reads as a relaunch rather than a first run.
+  var flutterRelaunchInProgress = false;
   spawnFlutterAppIfNeeded = () async {
     if (runMode != 'development') return;
     if (!config.hasFlutterPackage) {
@@ -594,6 +597,8 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
     if (existing != null && existing.isRunning) return;
     if (spawnInFlight) return;
     spawnInFlight = true;
+    final isRelaunch = flutterRelaunchInProgress;
+    flutterRelaunchInProgress = false;
 
     final fp = FlutterProcess(
       flutterPackageDir: p.joinAll(config.flutterPackagePathParts),
@@ -626,7 +631,9 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
     // On `-d web-server` `launched` pends until a browser attaches.
     unawaited(() async {
       await log.progress(
-        'Launching Flutter app (first run may take 30-60s)',
+        isRelaunch
+            ? 'Relaunching Flutter app'
+            : 'Launching Flutter app (first run may take 30-60s)',
         () async {
           await fp.launched;
           return true;
@@ -704,6 +711,7 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
     // button. `spawnFlutterAppIfNeeded` self-guards on development mode.
     flutterAppRestartAction: config.hasFlutterPackage
         ? () async {
+            flutterRelaunchInProgress = flutterProcess != null;
             await flutterProcess?.stop();
             await spawnFlutterAppIfNeeded();
           }
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -669,29 +669,31 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
   FlutterDependencyTracker? flutterDependencyTracker;
   if (config.hasFlutterPackage) {
     final flutterPackageDir = p.joinAll(config.flutterPackagePathParts);
-    final dartToolDir = FlutterDependencyTracker.resolveDartToolDir(
-      flutterPackageDir,
-    );
-    if (dartToolDir != null) {
-      try {
+    try {
+      final flutterPackageName = parsePubspec(
+        File(p.join(flutterPackageDir, 'pubspec.yaml')),
+      ).name;
+      final dartToolDir = FlutterDependencyTracker.resolveDartToolDir(
+        flutterPackageDir,
+        flutterPackageName: flutterPackageName,
+      );
+      if (dartToolDir != null) {
         flutterDependencyTracker = FlutterDependencyTracker(
           dartToolDir: dartToolDir,
-          flutterPackageName: parsePubspec(
-            File(p.join(flutterPackageDir, 'pubspec.yaml')),
-          ).name,
+          flutterPackageName: flutterPackageName,
         );
-      } catch (e) {
-        // A malformed Flutter pubspec must not prevent the server from
-        // starting; it only disables dependency tracking.
+      } else {
         log.debug(
-          'Flutter dependency tracking disabled: could not read the Flutter '
-          'package name from $flutterPackageDir ($e).',
+          'Flutter dependency tracking disabled: no resolution listing '
+          '$flutterPackageName found above $flutterPackageDir.',
         );
       }
-    } else {
+    } catch (e) {
+      // A malformed Flutter pubspec must not prevent the server from
+      // starting; it only disables dependency tracking.
       log.debug(
-        'Flutter dependency tracking disabled: no '
-        '.dart_tool/package_config.json found for $flutterPackageDir.',
+        'Flutter dependency tracking disabled: could not read the Flutter '
+        'package name from $flutterPackageDir ($e).',
       );
     }
   }
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -673,12 +673,21 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
       flutterPackageDir,
     );
     if (dartToolDir != null) {
-      flutterDependencyTracker = FlutterDependencyTracker(
-        dartToolDir: dartToolDir,
-        flutterPackageName: parsePubspec(
-          File(p.join(flutterPackageDir, 'pubspec.yaml')),
-        ).name,
-      );
+      try {
+        flutterDependencyTracker = FlutterDependencyTracker(
+          dartToolDir: dartToolDir,
+          flutterPackageName: parsePubspec(
+            File(p.join(flutterPackageDir, 'pubspec.yaml')),
+          ).name,
+        );
+      } catch (e) {
+        // A malformed Flutter pubspec must not prevent the server from
+        // starting; it only disables dependency tracking.
+        log.debug(
+          'Flutter dependency tracking disabled: could not read the Flutter '
+          'package name from $flutterPackageDir ($e).',
+        );
+      }
     } else {
       log.debug(
         'Flutter dependency tracking disabled: no '
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -10,6 +10,7 @@ import 'package:serverpod_cli/analyzer.dart';
 import 'package:serverpod_cli/src/commands/generate.dart';
 import 'package:serverpod_cli/src/commands/messages.dart';
 import 'package:serverpod_cli/src/commands/start/file_watcher.dart';
+import 'package:serverpod_cli/src/commands/start/flutter_dependency_tracker.dart';
 import 'package:serverpod_cli/src/commands/start/flutter_process.dart';
 import 'package:serverpod_cli/src/commands/start/kernel_compiler.dart';
 import 'package:serverpod_cli/src/commands/start/mcp_server.dart';
@@ -32,6 +33,7 @@ import 'package:serverpod_cli/src/migrations/create_repair_migration_action.dart
 import 'package:serverpod_cli/src/runner/serverpod_command.dart';
 import 'package:serverpod_cli/src/runner/serverpod_command_runner.dart';
 import 'package:serverpod_cli/src/util/internal_error.dart';
+import 'package:serverpod_cli/src/util/pubspec_helpers.dart';
 import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart';
 import 'package:serverpod_cli/src/vm_proxy/proxy.dart';
 import 'package:serverpod_cli/src/vm_proxy/serverpod_hooks.dart';
@@ -654,6 +656,30 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
     await spawnFlutterAppIfNeeded();
   }
 
+  // Track the Flutter app's resolved dependency closure so a dependency change
+  // (which a hot restart can't pick up) auto-triggers a full relaunch. Disabled
+  // when there is no Flutter package or its dependencies haven't been resolved.
+  FlutterDependencyTracker? flutterDependencyTracker;
+  if (config.hasFlutterPackage) {
+    final flutterPackageDir = p.joinAll(config.flutterPackagePathParts);
+    final dartToolDir = FlutterDependencyTracker.resolveDartToolDir(
+      flutterPackageDir,
+    );
+    if (dartToolDir != null) {
+      flutterDependencyTracker = FlutterDependencyTracker(
+        dartToolDir: dartToolDir,
+        flutterPackageName: parsePubspec(
+          File(p.join(flutterPackageDir, 'pubspec.yaml')),
+        ).name,
+      );
+    } else {
+      log.debug(
+        'Flutter dependency tracking disabled: no '
+        '.dart_tool/package_config.json found for $flutterPackageDir.',
+      );
+    }
+  }
+
   // Construct the watch session.
   session = WatchSession(
     compiler: compiler,
@@ -682,6 +708,8 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
             await spawnFlutterAppIfNeeded();
           }
         : null,
+    flutterDependenciesChangedSinceLastCheck:
+        flutterDependencyTracker?.refreshIfChanged,
     applyMigrationsAction: () => _applyMigrationsForSession(
       serverDir: serverDir,
       runMode: runMode,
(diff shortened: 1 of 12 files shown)


tools/serverpod_cli/lib/src/commands/start/watch_session.dart
```diff
@@ -194,8 +206,20 @@ class WatchSession {
...
+              FlutterDependencyChange.none)
+        : FlutterDependencyChange.none;
+
     // Static-only changes (HTML, JS, CSS, templates): no compilation needed.
     if (!hasDartChanges) {
+      if (flutterDependencyChange != FlutterDependencyChange.none) {
+        await _refreshFlutterApp(flutterDependencyChange);
+      }
       await _notifyBrowserRefresh();
       return;
```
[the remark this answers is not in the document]
FXschwartz: Done!

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -654,6 +663,30 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
...
+          File(p.join(flutterPackageDir, 'pubspec.yaml')),
+        ).name,
+      );
+    } else {
+      log.debug(
+        'Flutter dependency tracking disabled: no '
+        '.dart_tool/package_config.json found for $flutterPackageDir.',
+      );
+    }
+  }
```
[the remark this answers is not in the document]
FXschwartz: Fixed!

tools/serverpod_cli/lib/src/commands/start/watch_session.dart
```diff
@@ -274,12 +301,31 @@ class WatchSession {
...
+    await _refreshFlutterApp(flutterDependencyChange);
     if (reloaded) await _notifyBrowserRefresh();
   }
 
+  /// Refreshes the Flutter app after a file change with the lightest step that
+  /// picks the change up: a hot reload by default, a hot restart when
+  /// pure-Dart dependencies changed, or a full process relaunch when native
+  /// code changed. The relaunch calls the action directly rather than via
+  /// [restartFlutterApp] since we are already running inside [_chain].
+  Future<void> _refreshFlutterApp(FlutterDependencyChange change) async {
```
marcelomendoncasoares: Suggestion(NIT): Use `_reloadOrRestartFlutterApp` for parity with `_reloadOrRestart` for the server.

tools/serverpod_cli/lib/src/commands/start/flutter_dependency_tracker.dart
```diff
@@ -0,0 +1,258 @@
...
+    var dir = p.absolute(flutterPackageDir);
+    while (true) {
+      final dartTool = p.join(dir, '.dart_tool');
+      if (File(p.join(dartTool, 'package_config.json')).existsSync()) {
+        return dartTool;
+      }
+      final parent = p.dirname(dir);
+      if (parent == dir) return null; // Reached the filesystem root.
+      dir = parent;
+    }
```
marcelomendoncasoares, line 92: Suggestion: Might be an over-protective measure, but we could verify that the Flutter package exists in the found `package_config.json` before returning. It would prevent pointing at a wrong `package_config.json` if it was not found (`.dart_tool` deleted, for example) but a parent happened to exist.

tools/serverpod_cli/test/commands/start/watch_session_test.dart
```diff
@@ -1079,6 +1085,226 @@ void main() {
...
+          expect(flutter.calls, ['restart']);
+          expect(restartActionCalls, 0);
+        },
+      );
+    },
+  );
+
+  group(
+    'Given a watch session where the Flutter dependency file changed but the '
+    'closure did not,',
```
marcelomendoncasoares:
```suggestion
    'Given a watch session where the Flutter dependency file changed but the closure did not,',
```

tools/serverpod_cli/test/commands/start/file_watcher_test.dart
```diff
@@ -103,6 +103,71 @@ void main() {
...
+        // server FES restart, and .dart_tool churn is not a static change.
+        expect(event.packageConfigChanged, isFalse);
+        expect(event.staticFilesChanged, isFalse);
+        expect(event.dartFiles, isEmpty);
+      },
+    );
+
+    test(
+      'when package_config.json and other .dart_tool files change alongside '
+      'package_graph.json, '
```
marcelomendoncasoares:
```suggestion
      'when package_config.json and other .dart_tool files change alongside package_graph.json, '
```

tools/serverpod_cli/test/commands/start/file_watcher_test.dart
```diff
@@ -103,6 +103,71 @@ void main() {
...
+        ).writeAsString('3.12.0');
+        await File(
+          p.join(dartToolDir, 'package_graph.json'),
+        ).writeAsString('{"roots":[],"packages":[]}');
+
+        final event = await firstEvent;
+
+        expect(event.flutterDependenciesChanged, isTrue);
+        expect(event.packageConfigChanged, isFalse);
+        expect(event.staticFilesChanged, isFalse);
```
marcelomendoncasoares, line 165: Question: If I understood correctly, are we swallowing `packageConfigChanged` when `flutterDependenciesChanged` is true? If so, won't this prevent expected server reloads?

tools/serverpod_cli/lib/src/commands/start/flutter_dependency_tracker.dart
```diff
@@ -0,0 +1,258 @@
...
+  none,
+
+  /// Only pure-Dart dependencies changed; a hot restart picks them up.
+  dartOnly,
+
+  /// A dependency with native code changed; only a full process relaunch
+  /// rebuilds the native host (a hot restart would leave the new native code
+  /// out of the binary, failing at runtime with a MissingPluginException).
+  native,
+}
```
marcelomendoncasoares, line 21: Comment: @coderabbitai Please create a follow-up issue to also track changes to assets and fonts on the `pubspec.yaml` of the Flutter package.

tools/serverpod_cli/test/commands/start/flutter_dependency_tracker_test.dart
```diff
@@ -0,0 +1,486 @@
...
+      {'name': 'flutter_test', 'version': '0.0.0', 'dependencies': <String>[]},
+    ],
+  };
+
+  /// Pure-Dart source dirs + package_config for the app's closure packages.
+  Map<String, String> closureRootUris() => {
+    'app_flutter': writePackage('app_flutter', 'name: app_flutter\n'),
+    'app_client': writePackage('app_client', 'name: app_client\n'),
+    'http': writePackage('http', 'name: http\n'),
+  };
```
marcelomendoncasoares: Question: Can we chose real packages instead of mocking the `package_graph.json` and `package_config.json`? It would make these tests robust to the real files `pub` will generate and prevent errors due to chasing incorrectly mocked structures.

marcelomendoncasoares (commented):
Great work on this, @FXschwartz! And pretty nice to see that we managed to go with the more robust approach that we discussed. Have some questions and small suggestions. The main one is about testing the `FlutterDependencyTracker`.

tools/serverpod_cli/test/commands/start/flutter_dependency_tracker_test.dart
FXschwartz: Done! Converted everything pub can actually produce to a workspace built in a temp dir with a real `dart pub get --offline`, using only path/workspace dependencies.

tools/serverpod_cli/test/commands/start/file_watcher_test.dart
FXschwartz: So I know @nielsenko said differently during standup but from what I can see `packageConfigChanged` never had a live source on main. `package_config.json` only exists inside `.dart_tool`, and no watch root included `.dart_tool` until this PR.

Alex also messaged me with the same question and since we now have the watcher, debounce, and .dart_tool resolution it will be a minor PR to get the server hooked up to it. I can do that in a follow up PR today or tomorrow so the behavior is consistent between Flutter and the Server.

tools/serverpod_cli/test/commands/start/file_watcher_test.dart
FXschwartz: Done!

tools/serverpod_cli/test/commands/start/watch_session_test.dart
FXschwartz: Done!

tools/serverpod_cli/lib/src/commands/start/flutter_dependency_tracker.dart
FXschwartz: This is a good call and made me realize It should be improved even more so that if it fails to find the pubspec for whatever reason, instead of preventing it from starting we just disabled tracking.

Done!

tools/serverpod_cli/lib/src/commands/start/watch_session.dart
FXschwartz: Done!

tools/serverpod_cli/test/commands/start/file_watcher_test.dart
marcelomendoncasoares: @coderabbitai Please create a follow up issue for this fix.

tools/serverpod_cli/lib/src/commands/start/flutter_dependency_tracker.dart
```diff
@@ -65,20 +65,20 @@ class FlutterDependencyTracker {
...
+    String flutterPackageDir, {
+    required String flutterPackageName,
+  }) {
     var dir = p.absolute(flutterPackageDir);
     while (true) {
       final dartTool = p.join(dir, '.dart_tool');
-      if (File(p.join(dartTool, 'package_config.json')).existsSync()) {
+      if (_packageConfigContains(dartTool, flutterPackageName)) {
         return dartTool;
       }
```
marcelomendoncasoares: Question: Can we expect a scenario where an inner `package_config.json` does not have the package name, but an outer one has it? Or should we return null in case the first found `package_config.json` does not have the Flutter package name?

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
```diff
@@ -426,6 +427,31 @@ class FlutterProcess {
     await _sigtermSub?.cancel();
     _sigtermSub = null;
 
+    // Ask the daemon to stop the app first: in --machine mode a bare SIGINT
+    // can terminate flutter_tools without device cleanup, leaving e.g. the
+    // browser window it spawned open. `app.stop` tears the device session
+    // down properly; signals below remain as the fallback.
```
nielsenko, line 433: nice 👍 .. if true 😊

nielsenko (changes requested):
Now I haven't tested this.. and maybe my understanding is wrong..

But how can this work at all. We need to pass the .`dart_tool/package_config.json` to the FES on startup, and since we don't control it (as we do on the pod) - but only spawn `flutter run` - how do we ensure it is picked up? As far as I can tell you don't respawn `flutter run`.
  
  I truly hope I'm wrong 😊

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
marcelomendoncasoares: Will this fix #5167?

nielsenko:
> Now I haven't tested this.. and maybe my understanding is wrong..
> 
> But how can this work at all. We need to pass the .`dart_tool/package_config.json` to the FES on startup, and since we don't control it (as we do on the pod) - but only spawn `flutter run` - how do we ensure it is picked up? As far as I can tell you don't respawn `flutter run`.
> 
> I truly hope I'm wrong 😊

I was wrong 🎉 Check _incremental_compiler.dart_ and _run_hot.dart_

nielsenko (approved):
LGTM 🚀

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
nielsenko: it should

tools/serverpod_cli/lib/src/commands/start/flutter_dependency_tracker.dart
FXschwartz: Good point! I ended up going with returning null. 

Thinking about it more, the keep walking version only helps in pretty heavily nested layouts, but package names aren't unique, so an unrelated resolution further up that happens to contain a package with the same name would match and we'd silently track the wrong root. Stopping at the first one found is much easier to reason with.

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
FXschwartz: It does quit the app but just to be clear it won't close the browser window that the flutter app was running in.

nielsenko: Ahh.. then it doesn't fix #5167 and the comment is somewhat overreaching. Cannot trust those pesky LLMs.

FXschwartz: I don't actually know how possible this is though. I know when running the Flutter web app through vscode it opens it's own browser instance of the browser you choose and that gets killed when stopping the flutter app. But Serverpod seems to open just a new tab in an already open instance of a browser. 

Maybe that has something to do with the difference?

nielsenko: To be honest, I recall deliberately choosing **_not_** to send `app.stop`.. but I fail to recall why at the moment 🤷

marcelomendoncasoares: > But Serverpod seems to open just a new tab in an already open instance of a browser.

Currently yes, but the issue was for when we were running with device `chrome`. Right now, with `webserver`, we can not close the tab, as it lives on the users browser. To test if we can solve #5167 with this change we should try running `serverpod start --flutter-device chrome` to see if it closes the window when quitting.

tools/serverpod_cli/test/commands/start/flutter_dependency_tracker_test.dart
```diff
@@ -0,0 +1,610 @@
...
+        'name: [unclosed',
+      );
+    }
+    return tracker.refresh();
+  }
+
+  group('Given a changed app dependency declaring a native plugin class,', () {
+    test(
+      'when the closure is refreshed, '
+      'then a native change is reported',
```
marcelomendoncasoares: Suggestion: No need for all these `group` blocks that just split the description with a single test inside. It is only needed to split the "Given" to a group if the setup is also moved up to keep the test focused on the acting + result or if the test condition applies to multiple when + then.

tools/serverpod_cli/test/commands/start/flutter_dependency_tracker_test.dart
```diff
@@ -0,0 +1,610 @@
...
+          // ancestor's resolution is real but doesn't list the app.
+          final ancestorDir = p.join(tempDir.path, 'other_project');
+          write(
+            p.join(ancestorDir, 'pubspec.yaml'),
+            _pubspec('something_unrelated'),
+          );
+          await _pubGet(ancestorDir);
+          final pkg = await Directory(
+            p.join(ancestorDir, 'nested', 'app_flutter'),
+          ).create(recursive: true);
```
marcelomendoncasoares: Comment: This is an example of a test that is worth moving the setup to the "Given" group just for the test itself be clear to read.

tools/serverpod_cli/test/commands/start/flutter_dependency_tracker_test.dart
```diff
@@ -0,0 +1,610 @@
...
+          );
+
+          expect(resolved, isNull);
+        },
+      );
+
+      test(
+        'when an even higher resolution does list the package name, '
+        'then it still resolves to null',
+        () async {
```
marcelomendoncasoares: Suggestion: This is actually another "Given".

tools/serverpod_cli/test/commands/start/flutter_dependency_tracker_test.dart
```diff
@@ -0,0 +1,610 @@
...
+        );
+
+        expect(tracker.refresh(), FlutterDependencyChange.none);
+      },
+    );
+
+    test(
+      'when the graph file is malformed, '
+      'then no change is reported',
+      () {
```
marcelomendoncasoares, line 554: Suggestion: These (and the below) tests are all individual "Given" groups. The "when" is describing part of the test scenario here, not the acting.

marcelomendoncasoares (commented):
Just some reorganization on the tests to better adhere to the test pattern.

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
FXschwartz: @marcelomendoncasoares Confirmed that running with `--flutter-device chrome` the browser window that opens completely closes when quitting and when the flutter app restarts via dep change or manually with `CTRL+R`.

I've updated the PR description to close it

tools/serverpod_cli/test/commands/start/flutter_dependency_tracker_test.dart
FXschwartz: Fixed!

tools/serverpod_cli/test/commands/start/flutter_dependency_tracker_test.dart
FXschwartz: Moved!

tools/serverpod_cli/test/commands/start/flutter_dependency_tracker_test.dart
FXschwartz: Moved!

tools/serverpod_cli/test/commands/start/flutter_dependency_tracker_test.dart
FXschwartz: Restructured to their own groups

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5285, opened 2026-06-03
fix: Hide the start TUI splash screen when it is remounted hidden

### LINKED ISSUE #5284: `serverpod start` splash screen reappears and gets stuck after pressing Ctrl+C on Windows
In the `serverpod start` TUI, the Serverpod splash banner initially dismisses correctly after startup. However, pressing **Ctrl+C** (which shows the "Press Ctrl-C again to exit" / "Copied to clipboard" hint) makes the splash banner reappear in the center of the screen and it never goes away again.

FXschwartz:
In the `serverpod start` TUI, the splash banner dismissed correctly after startup, but pressing Ctrl+C made it reappear and stick on screen forever.

`LoadingScreen` only hid itself after observing a `visible: true → false` transition in `didUpdateComponent`, which arms the fade-out. When an ancestor changes its tree shape, nocterm remounts the whole subtree, and a freshly mounted `LoadingScreen(visible: false)` never sees that transition so its `build` rendered the splash at full opacity indefinitely. One trigger that would cause this is the Ctrl-C hint row appearing or disappearing.

This PR makes `LoadingScreen.build` return nothing whenever it is not visible and no fade-out is in flight, so the splash is robust against subtree remounts regardless of their origin. The normal startup fade-out behavior is unchanged. Includes regression tests that exercise the Ctrl-C remount path (verified to fail without the fix).

Fixed #5284 

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start/tui/loading_screen.dart
+++ b/tools/serverpod_cli/lib/src/commands/start/tui/loading_screen.dart
@@ -59,7 +59,15 @@ class _LoadingScreenState extends State<LoadingScreen>
   Component build(BuildContext context) {
     final theme = ServerpodTheme.of(context);
     final t = _controller.value;
-    if (t <= 0 && _fadingOut) return const SizedBox.shrink();
+    // Hidden when not visible, unless a fade-out is still in flight. The
+    // explicit !visible check matters: ancestors changing their tree shape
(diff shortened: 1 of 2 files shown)


marcelomendoncasoares (approved):
Nice fix!

PR #5289, opened 2026-06-04
fix: Always map joined data to the correct object.

### LINKED ISSUE #5287: `include` column-alias collision bleeds one relation's column into another relation's field
### Describe the bug

When loading an entity with a large / overlapping relational `include` graph via
`Model.db.findById(include: ...)`, Serverpod populates a field of one included
relation with a **column value from a completely different included relation**.

Concretely, in our case an `Address.addressLine3` value (a `String`) is written
into the `Organization.id` field (an `int?`) of a *different* branch of the
include tree. On deserialization this throws:

```text
type 'String' is not a subtype of type 'int?' in type cast
  at OrganizationDTO.fromJson  (id: jsonSerialization['id'] as int?)
```

When the bleeding source value happens to be `null`/empty, no exception is
thrown, but the target field is still silently set to the wrong value (`null`) —
i.e. **silent data corruption** even when it does not crash.

## Relevant model shapes (trimmed to the parts that matter)

Root entity with several FK relations, three of which point at the **same child
table** (`Address`) and, at runtime, at the **same row**:

```yaml
# OfferHeader
class: OfferHeaderDTO
table: edm_transactions_offer_header
fields:
  customer: module:edm_parties:PartyCustomerDTO?, relation(optional)
  billToParty:           module:edm_parties:PartyDTO?,    relation(optional)
  billToAddress:         module:edm_parties:AddressDTO?,  relation(optional)
  shipToParty:           module:edm_parties:PartyDTO?,    relation(optional)
  shipToAddress:         module:edm_parties:AddressDTO?,  relation(optional)
  invoiceRecipientParty:   module:edm_parties:PartyDTO?,   relation(optional)
  invoiceRecipientAddress: module:edm_parties:AddressDTO?, relation(optional)
  # ... currency / paymentTerm / offerLines(list, deep) ...
```

```yaml
# PartyCustomer  ->  Party  ->  Organization   (a 3-level nested one-to-one chain)
class: PartyCustomerDTO
fields:
  party: PartyDTO?, relation(...)

class: PartyDTO
fields:
  organization: OrganizationDTO?, relation(name=party_organization)   # one-to-one
  addresses: List<AddressDTO>?, relation(name=party_addresses)

# Organization <-> Party : one-to-one, unique index, explicit field=
class: OrganizationDTO
table: edm_parties_organization
fields:
  partyId: int
  party: PartyDTO?, relation(name=party_organization, field=partyId)
  organizationName: String?
  # ... (id is the implicit auto PK, int) ...
indexes:
  organization_party_unique_idx:
    fields: partyId
    unique: true

class: AddressDTO
table: edm_parties_address
fields:
  partyId: int
  addressLine1: String?
  addressLine2: String?
  addressLine3: String?    # <-- the value that bleeds into Organization.id
  # ...
```

## The include that triggers it

```dart
ProjectOfferDTO.include(
  offerHeader: OfferHeaderDTO.include(
    customer: PartyCustomerDTO.include(
      party: PartyDTO.include(organization: OrganizationDTO.include()),   // 3-level nested one-to-one
    ),
    billToParty: PartyDTO.include(organization: OrganizationDTO.include()),
    billToAddress: AddressDTO.include(country: CountryDTO.include()),
    shipToParty: PartyDTO.include(),
    shipToAddress: AddressDTO.include(country: CountryDTO.include()),
    invoiceRecipientParty: PartyDTO.include(),
    invoiceRecipientAddress: AddressDTO.include(country: CountryDTO.include()),
    // + a deep offerLines list include (itemVariant -> item -> unit/vatInfo, options)
  ),
)
```

At runtime, `billToParty`, `shipToParty`, `invoiceRecipientParty` and
`customer.party` all reference the **same Party row**, and `billToAddress`,
`shipToAddress`, `invoiceRecipientAddress` all reference the **same Address
row**. So the same physical rows are joined multiple times in one query.

## Observed behaviour

A diagnostic `print(jsonSerialization)` added at the top of the generated
`OrganizationDTO.fromJson` produced:

```text
id=zusa 2 (String)
keys=[id, partyId, organizationName, tradingName, legalFormId, vatId, vatIdValidated, taxNumber, isSmallBusiness]
json={id: zusa 2, partyId: 1, organizationName: App Ahoi GmbH, tradingName: ..., legalFormId: 1, vatId: ...}
```

`"zusa 2"` is exactly the string the user typed into the **`Address.addressLine3`**
field of the offer's address. Every other organization field is correct — only
`id` is wrong, and it holds a value belonging to a *different relation* (Address)
in a *different branch* of the include tree.

So the JSON Serverpod assembled for the included `Organization` has
`Address.addressLine3` under the `id` key.

### To reproduce

All of the following were present; we have not yet minimized to the smallest set:

1. A root entity with **many** included relations.
2. The **same child row referenced by multiple relations** in one query (same
   `Address` row via 3 FK relations; same `Party` row via 4 relations).
3. A **deep (3-level) nested one-to-one include** (`customer → party →
   organization`) where the one-to-one is declared with
   `relation(name=..., field=partyId)` + a unique index.
4. A **non-empty text column** (`Address.addressLine3`) on the duplicated child —
   when it is empty/null the cast does not throw, but the org `id` is still
   silently corrupted to `null`.

Removing the 3-level `customer.party.organization` branch from the include makes
the crash go away — i.e. reducing the include complexity changes the alias layout
enough to avoid the collision.

### Expected behavior

Each included relation's fields must be populated only from its own row's
columns. No column value from one relation should ever appear in another
relation's field, regardless of how many joins / duplicate rows / nested
one-to-one relations the include graph contains.

## Hypothesis

The SQL Serverpod generates for this include graph (column aliasing / JSON
aggregation across the LEFT JOINs) produces a **collision**: a column from one
joined relation is emitted under the same key/position as a field of another
relation, so deserialization reads the wrong value. The trigger appears to be the
combination of (a) the same table joined multiple times, and (b) a deep nested
one-to-one (`field=`) relation, pushing the alias generation past a point where
keys are no longer unique.

## Workaround (app side)

Reduce the include so the colliding (deep, redundant) branch is dropped:

```dart
// before
customer: PartyCustomerDTO.include(
  party: PartyDTO.include(organization: OrganizationDTO.include()),
),
// after
customer: PartyCustomerDTO.include(
  party: PartyDTO.include(),
),
```

### Library version

3.4.8

### Platform information

Every Platform

### Additional context

_No response_

- [ ] I want to work on this

Isakdl:
Address the problem of collisions in the query builder when truncating long column names. There was a risk of collisions previously, especially in larger queries with many joins or tables with long column names that are near identical, where the difference between them would be in the end of the string.

This PR aims to address this without adding too much plumbing around the code.

I have added a few tests that verified the bug exist (test failing), after the production change they now pass.

The solution I ended up with will detect collisions and then append a counter at the end of the string. After that we check for collisions again and retry until a non-colliding match is found.

The solution is fairly easy to understand and has a pretty small surface area in the code, hence I think it ends up with a good tradeoff.  

I have considered other approaches.

1. Longer hash, this will reduce the collision risk but not eliminate it, that solution will also obscure the column name even more. 

2. Code generate an alias for long names at generation time, a pretty good solution but has the drawback of affecting a lot of code and tests and would be a huge change. Secondly, you could still get into the truncation and collision problems with enough deeply nested joins. It would also obscure the column name more than today.

Closes: #5287 

I suggest cherry-picking this commit for a patch release as this is a rather nasty bug that will silently return invalid data from the database. 

## Pre-launch Checklist

## Breaking changes

None


packages/serverpod_database/lib/src/adapters/postgres/sql_query_builder.dart
```diff
@@ -1516,9 +1516,23 @@ LinkedHashMap<String, String> _gatherIncludeJoins(
...
+      if (existingJoin != null && existingJoin != joinStatement) {
+        // Two different table relations resolved to the same join alias, which
+        // would silently drop one join (and corrupt included data). This can
+        // happen when relation names are long enough to be truncated to the
+        // same identifier. See https://github.com/serverpod/serverpod/issues/5287
+        throw StateError(
+          'Table alias collision in include query: two different relations '
+          'resolved to the same join alias "$alias". This is likely caused by '
+          'very long or similar relation names.',
+        );
```
marcelomendoncasoares, line 1538: Suggestion: Include a recommendation on what to do (like remove some joins or open an issue).

packages/serverpod_database/test/database_query/long_relation_field_name/column_alias_collision_test.dart
```diff
@@ -0,0 +1,62 @@
...
+///
+/// When a query selects columns from two different relations that point at the
+/// same table, and the relation field names are long enough to be truncated,
+/// the truncated column aliases can collide. When that happens the database
+/// returns both columns under the same key, so one relation's value bleeds into
+/// the other relation's field on deserialization.
+///
+/// The two relation field names below are chosen so that the relations get
+/// distinct join aliases (they are genuinely separate joins), but the `id`
+/// column of each truncates to the *same* select alias.
```
marcelomendoncasoares, line 18: Suggestion: We usually do not include references for the issues under fix on the tests. Might be worth only explaining the scenario on the table/column name below?

tests/serverpod_test_server/test_integration/database_operations/entity_relations/column_alias_collision_test.dart
```diff
@@ -0,0 +1,92 @@
...
+///
+///   type 'String' is not a subtype of type 'int?' in type cast
+///
+/// This test asserts the correct behaviour (each relation keeps its own column
+/// values), so it is expected to pass once the collision is fixed.
+void main() async {
+  var session = await IntegrationTestServer().session();
+
+  group(
+    'Given two relations to the same table with long relation field names (issue #5287)',
```
marcelomendoncasoares:
```suggestion
    'Given two relations to the same table with long relation field names',
```

packages/serverpod_database/test/database_query/long_relation_field_name/column_alias_collision_test.dart
marcelomendoncasoares: Same for the comments on all other tests and descriptions referencing the issue number.

marcelomendoncasoares (approved):
Nice solution! Very small footprint, and makes the aliases safe. I made only a few suggestions regarding the explanation comments and to instruct users on what to do if the resolver fails.

Once merged I'll cherry-pick the squashed commit and release a patch 3.4.9.

packages/serverpod_database/test/database_query/long_relation_field_name/column_alias_collision_test.dart
Isakdl: I know we haven't done that. But I also know that the Flutter team does this and one benefit is that you will have a clear reference for an actual bug that happened, which is the reason for the test existing. 

With that said I can of course remove the comment / issue reference.

marcelomendoncasoares: Ok, let's experiment with adding those. It can be indeed a good practice and AI makes very cheap to add good short docs like these.

PR #5292, opened 2026-06-04
feat: Add IDE selection screen to quickstart command

### LINKED ISSUE #5216: Make it possible to select IDEs in `quickstart` command
Make it possible to select IDEs so we don't clutter the project too much with duplicated, unused files. Require the user to pick at least one IDE. Take the screen from the updated create command. #5218

FXschwartz:
This PR reuses the `create` command's TUI for `quickstart`, exposing only the IDEs row. The selection starts empty and at least one IDE must be selected before the project can be created.

- `CreateConfigState` now accepts the set of configs to expose, fallback values for configs not shown in the form
- The TUI launch plumbing is extracted from `CreateCommand` into a shared `performCreateWithTui()` used by both commands.
- In non-interactive runs (`--no-interactive` or CI), where no selection screen is possible, `quickstart` keeps its previous behavior including the default IDE set.

Fixes #5216 

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/create/tui/state.dart
+++ b/tools/serverpod_cli/lib/src/commands/create/tui/state.dart
@@ -5,16 +5,35 @@ import 'package:serverpod_tui/serverpod_tui.dart';
 
 /// Central state for [ServerpodCreateApp] rendered by nocterm.
 class CreateConfigState extends TuiState {
-  CreateConfigState(this.startingTemplate) {
-    form.updateSelectedOption(
-      ServerpodCreateConfig.template,
-      startingTemplate.toConfigOption,
-    );
+  CreateConfigState(
+    this.startingTemplate, {
+    this.configs = ServerpodCreateConfig.values,
+    TemplateContext? defaults,
+    this.requireIde = false,
+  }) : defaults = defaults ?? TemplateContext() {
+    if (configs.contains(ServerpodCreateConfig.template)) {
+      form.updateSelectedOption(
+        ServerpodCreateConfig.template,
+        startingTemplate.toConfigOption,
+      );
+    }
   }
 
   final ServerpodTemplateType startingTemplate;
 
-  late final form = FormState(ServerpodCreateConfig.values);
+  /// The configurations exposed in the form.
+  final List<ServerpodCreateConfig> configs;
+
+  /// Fallback values for configurations not exposed in the form.
+  /// Configurations hidden by an unsatisfied [FormRequirement] are
+  /// not affected and resolve from the form as usual.
+  final TemplateContext defaults;
+
+  /// Whether at least one IDE must be selected
+  /// before the project can be created.
+  final bool requireIde;
+
+  late final form = FormState(configs);
 
   bool _creatingProject = false;
   bool get creatingProject => _creatingProject;
@@ -29,6 +48,14 @@ class CreateConfigState extends TuiState {
       form.getSelectedOptionFor(ServerpodCreateConfig.template)?.toTemplate ??
       startingTemplate;
 
+  /// True when all required selections are made
+  /// and the project can be created.
+  bool get canCreate {
+    if (!requireIde) return true;
+    final selectedIdes = form.getSelectedOptionsFor(ServerpodCreateConfig.ide);
+    return selectedIdes != null && selectedIdes.isNotEmpty;
+  }
+
   /// Called when project creation starts.
   /// This transitions the UI to a log viewer.
   void markCreatingProject() {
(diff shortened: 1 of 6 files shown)


tools/serverpod_cli/lib/src/commands/create/tui/state.dart
```diff
@@ -29,6 +48,14 @@ class CreateConfigState extends TuiState {
       form.getSelectedOptionFor(ServerpodCreateConfig.template)?.toTemplate ??
       startingTemplate;
 
+  /// True when all required selections are made
+  /// and the project can be created.
+  bool get canCreate {
+    if (!requireIde) return true;
+    final selectedIdes = form.getSelectedOptionsFor(ServerpodCreateConfig.ide);
+    return selectedIdes != null && selectedIdes.isNotEmpty;
+  }
```
marcelomendoncasoares, line 57: Question: Why forbid creating a project without selecting an IDE?

tools/serverpod_cli/lib/src/commands/quickstart.dart
```diff
@@ -117,16 +121,48 @@ class QuickstartCommand extends ServerpodCommand<QuickstartOption> {
...
+        dryRun: true,
+        interactive: interactive,
+        context: context,
+      );
+
+      if (dryRunProjectPath == null) {
+        throw ExitException.error();
+      }
+
+      await performCreateWithTui(
```
marcelomendoncasoares: Suggestion: Move this dry run exercise to inside the `performCreateWithTui` to avoid the duplication.

tools/serverpod_cli/lib/src/commands/quickstart.dart
```diff
@@ -117,16 +121,48 @@ class QuickstartCommand extends ServerpodCommand<QuickstartOption> {
       }
     }
 
+    final context = TemplateContext(
+      template: template,
+      postgres: true,
+      web: true,
+      ides: [TemplateIde.claude, TemplateIde.cursor, TemplateIde.vscode],
+    );
```
marcelomendoncasoares, line 131: Suggestion: This is a change from today's behavior, but `quickstart` should also have `auth`. We initially did not add to keep the code simple for an experimentation, but @vlidholt clarified that the idea behind this command is just being a fast way to startup, without having to think on options as the `create` command presents.

tools/serverpod_cli/test/commands/create/tui/state_test.dart
```diff
@@ -141,4 +142,154 @@ void main() {
...
+        state = CreateConfigState(
+          ServerpodTemplateType.server,
+          configs: const [ServerpodCreateConfig.ide],
+          defaults: TemplateContext(postgres: true, web: true),
+        );
+      });
+
+      test(
+        'then TemplateContext has the default values '
+        'for configs not exposed in the form',
```
marcelomendoncasoares:
```suggestion
        'then TemplateContext has the default values for configs not exposed in the form',
```

tools/serverpod_cli/test/commands/create/tui/state_test.dart
```diff
@@ -141,4 +142,154 @@ void main() {
       );
     },
   );
+
+  group(
+    'Given a CreateConfigState exposing only the ide config '
+    'with default values for the other configs, '
```
marcelomendoncasoares:
```suggestion
    'Given a CreateConfigState exposing only the ide config with default values for the other configs, '
```

tools/serverpod_cli/test/commands/create/tui/state_test.dart
```diff
@@ -141,4 +142,154 @@ void main() {
...
+        () {
+          state.form.updateSelectedOption(
+            ServerpodCreateConfig.ide,
+            IdeOption.claude,
+          );
+
+          final context = state.toTemplateContext();
+          expect(context.ides, [TemplateIde.claude]);
+        },
+      );
```
marcelomendoncasoares, line 205: Suggestion: This test is out of the pattern.

marcelomendoncasoares (changes requested):
Few comments for improvement and one question around requiring at least one IDE.

tools/serverpod_cli/lib/src/commands/create/tui/state.dart
FXschwartz: The original issue said for it to be required, but it's an easy thing for me to revert if that has changed since it was created!

tools/serverpod_cli/lib/src/commands/quickstart.dart
FXschwartz: Moved!

tools/serverpod_cli/lib/src/commands/quickstart.dart
FXschwartz: Added! Just to confirm we want to keep redis off for quickstart though correct?

tools/serverpod_cli/test/commands/create/tui/state_test.dart
FXschwartz: Fixed!

tools/serverpod_cli/test/commands/create/tui/state_test.dart
FXschwartz: Fixed!

tools/serverpod_cli/test/commands/create/tui/state_test.dart
FXschwartz: Moved to it's own test group!

tools/serverpod_cli/lib/src/commands/quickstart.dart
marcelomendoncasoares: Didn't realize this one! No, let's add it also. Since Redis also have a fallback for development without need for docker, it is good to have so users can experimente with [the full suite of Serverpod features](https://docs.serverpod.dev/overview#key-capabilities).

FXschwartz: Added!

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5295, opened 2026-06-05
feat: Show only Hot restart action in the TUI when watch mode is enabled

### LINKED ISSUE #5166: Hide the `Hot reload` button legend on `serverpod start`
The `serverpod start` command is meant to be used with hot reload and incremental compilation (watch mode). We initially shipped it with a `--watch`/`--no-watch` options, ~but this adds unnecessary user-facing options, branching and complexity. To run the server without watch, there will always be the traditional `dart run bin/main.dart` direct execution.~ 
> The UX of running the server + app with hotkeys to reload + create migrations is not possible to match using `dart run bin/main dart`. Since we already have `--no-watch`, it is best to keep it to support advanced use cases where users might want to be completely in control. However...

We should replace the `Hot reload / restart` button on the TUI by only `Hot restart` when running in `--watch` mode (default), since hot reload is already taken care of by the incremental compiler.

FXschwartz:
When running `serverpod start` in watch mode (the default), hot reload is already handled automatically by the incremental compiler on every file change.

With this change, in watch mode the button bar shows `R Hot restart`, plain `R` triggers a hot restart, and the help overlay lists `R → Hot restart`. `Shift+R` also still triggers a hot restart, so existing muscle memory keeps working. When running with `--no-watch`, the previous behavior is unchanged: `R` hot reloads and `Shift+R` hot restarts.

Fixes #5166 

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -1017,7 +1017,9 @@ Future<int> _runWithTui({
   required List<String> serverArgs,
   required GeneratorConfig config,
 }) async {
-  final holder = StartAppStateHolder(ServerWatchState());
+  final holder = StartAppStateHolder(
+    ServerWatchState()..watchModeEnabled = watch,
+  );
   var backendStarted = false;
 
   // Shared shutdown signal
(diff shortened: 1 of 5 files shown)


marcelomendoncasoares (approved):
Nice surgical change! 🪡

PR #5306, opened 2026-06-10
feat: Project-local MCP config for Antigravity via .agents plugin

### LINKED ISSUE #5220: MCP server not working in Antigravity IDE
The MCP server is not working in Antigravity IDE. Screenshot of error message (not sure if it's possible to get more details).

Image

nielsenko:
Antigravity discovers project-local MCP servers through plugin folders under .agents/plugins/ that contain a plugin.json marker.

- `TemplateIde` gains `additionalFiles`: companion files rendered through the same slot/replacement pipeline as the main config.
- `serverpod create .` (upgrade) now also writes IDE MCP configs, so projects created by earlier CLI versions pick up the new layout.
- `_createFileAndWrite` logs write failures instead of swallowing them
- `_moveDirectoryContents` no longer overwrites destination files, so the .agent -> .agents merge cannot clobber _.agents/plugins/_.
- The plugin dir and manifest name derive from one constant

Fixes: #5220 (at least for antigravity-cli)

## Pre-launch Checklist

## Breaking changes

None (only changes behaviour of pre-release)

### DIFF
--- a/tools/serverpod_cli/lib/src/create/create.dart
+++ b/tools/serverpod_cli/lib/src/create/create.dart
@@ -261,19 +261,9 @@ Future<String?> performCreate(
     });
   }
 
-  if (context.ides.isNotEmpty) {
-    await log.progress('Configuring Serverpod MCP server', () async {
-      await _configureMcpServer(
-        serverpodDirs.projectDir.path,
-        context.ides,
-        serverDirRelative: p.relative(
-          serverpodDirs.serverDir.path,
-          from: serverpodDirs.projectDir.path,
-        ),
-      );
-      return true;
-    });
+  await _configureProjectMcpServers(serverpodDirs, context.ides);
 
+  if (context.ides.isNotEmpty) {
     await log.progress('Installing agent skills', () async {
       try {
         if (context.template != ServerpodTemplateType.module &&
@@ -374,6 +364,10 @@ Future<void> _moveDirectoryContents(
     final newPath = p.join(destination.path, p.basename(entity.path));
 
     if (entity is File) {
+      // Never overwrite files already present at the destination, such as the
+      // Antigravity plugin config written under .agents/plugins/ before this
+      // move runs.
+      if (await File(newPath).exists()) continue;
       await entity.rename(newPath);
     } else if (entity is Directory) {
       final newDir = await Directory(newPath).create(recursive: true);
@@ -381,6 +375,25 @@ Future<void> _moveDirectoryContents(
       await entity.delete();
     }
   }
+}
+
+/// Writes the MCP config files for [ides] under [dirs], showing progress.
+Future<void> _configureProjectMcpServers(
+  ServerpodDirectories dirs,
+  List<TemplateIde> ides,
+) async {
+  if (ides.isEmpty) return;
+  await log.progress('Configuring Serverpod MCP server', () async {
+    await _configureMcpServer(
+      dirs.projectDir.path,
+      ides,
+      serverDirRelative: p.relative(
+        dirs.serverDir.path,
+        from: dirs.projectDir.path,
+      ),
+    );
+    return true;
+  });
 }
 
 Future<void> _configureMcpServer(
@@ -395,6 +408,12 @@ Future<void> _configureMcpServer(
         p.join(projectDirPath, ide.filePath),
         ide.effectiveConfig(serverDirRelative: serverDirRelative),
       );
+      for (final entry in ide.additionalFiles.entries) {
+        await _createFileAndWrite(
+          p.join(projectDirPath, entry.key),
+          ide.render(entry.value, serverDirRelative: serverDirRelative),
+        );
+      }
     },
   );
 }
@@ -404,8 +423,8 @@ Future<void> _createFileAndWrite(String path, String content) async {
   try {
     await file.create(recursive: true);
     await file.writeAsString(content);
-  } on FileSystemException {
-    //
+  } on FileSystemException catch (e) {
+    _logError('Failed to write $path: ${e.osError?.message ?? e.message}');
   }
 }
 
@@ -474,6 +493,8 @@ Future<String?> _performUpgrade({
       interactive: interactive,
     );
   });
+
+  await _configureProjectMcpServers(serverpodDir, context.ides);
 
   if (success) {
     log.info(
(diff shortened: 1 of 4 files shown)


tools/serverpod_cli/lib/src/create/create.dart
```diff
@@ -261,19 +261,9 @@ Future<String?> performCreate(
...
-        serverDirRelative: p.relative(
-          serverpodDirs.serverDir.path,
-          from: serverpodDirs.projectDir.path,
-        ),
-      );
-      return true;
-    });
+  await _configureProjectMcpServers(serverpodDirs, context.ides);
 
+  if (context.ides.isNotEmpty) {
```
marcelomendoncasoares: Suggestion(NIT): Move the `_configureProjectMcpServers` to inside the `context.ides.isNotEmpty` condition. I saw it also skips if IDEs are empty, but it reads weird that the action is not gated together with the skills.

tools/serverpod_cli/lib/src/create/create.dart
```diff
@@ -374,6 +364,10 @@ Future<void> _moveDirectoryContents(
     final newPath = p.join(destination.path, p.basename(entity.path));
 
     if (entity is File) {
+      // Never overwrite files already present at the destination, such as the
+      // Antigravity plugin config written under .agents/plugins/ before this
+      // move runs.
+      if (await File(newPath).exists()) continue;
```
marcelomendoncasoares: Question: Should we warn in this case? If a user creates one project, deletes, then create another with the same name but in a different folder, what will happen with the stored plugins?

tools/serverpod_cli/lib/src/create/ide.dart
```diff
@@ -56,6 +71,19 @@ extension TemplateIdeExtension on TemplateIde {
...
+const _antigravityPluginDir = '.agents/plugins/$_antigravityPluginName';
+const _antigravityPluginName = 'serverpod-local';
+
+/// Marker that registers the generated folder as an Antigravity plugin so its
+/// sibling mcp_config.json is ingested.
+const _antigravityPluginManifest =
+    '''{
+  "name": "$_antigravityPluginName"
+}
+''';
```
marcelomendoncasoares, line 85: Question: Shouldn't it be something tied to the project, so different projects can exist simultaneously?

tools/serverpod_cli/test/commands/create/ide_test.dart
marcelomendoncasoares, line 1: Question: Isn't this test file mostly redundant with `tests/bootstrap_project/test_perform_create/ide_test.dart`?

marcelomendoncasoares (changes requested):
Just one NIT and a few questions.

tools/serverpod_cli/lib/src/create/create.dart
nielsenko: Done

tools/serverpod_cli/lib/src/create/create.dart
nielsenko: The plugin lives inside the project tree (<project>/.agents/plugins/serverpod-local/), so deleting the project deletes it, and a new project gets its own fresh copy. This function only merges _.agent/_ into _.agent**s**/_ within a single create run

tools/serverpod_cli/lib/src/create/ide.dart
nielsenko: Antigravity scopes plugin discovery per project (_.agents/plugins/_). So two projects using the same plugin name never collide.

A multi-root workspace spanning two serverpod projects is an edge-case - but we don't handle that for any other agent/IDE.

tools/serverpod_cli/test/commands/create/ide_test.dart
nielsenko: Yes, but the bootstrap suite only fails after a full create (pub get + flutter create) - this fails in about a second.
Should I remove it?

marcelomendoncasoares: I think it is fine to keep it as a unit test, then.

tools/serverpod_cli/test/commands/create/ide_test.dart
```diff
@@ -0,0 +1,95 @@
...
+      test('then the server dir slot and replacements are applied', () {
+        final rendered = ide.render(
+          '"dart": "{serverDirRelative}"',
+          serverDirRelative: 'my_app_server',
+        );
+
+        expect(rendered, '"dart-mcp-server": "my_app_server"');
+      });
+    });
+  });
```
marcelomendoncasoares: Suggestion: Collapse unnecessary groups like this into a test-only.

marcelomendoncasoares (approved):
Only one small suggestion on the tests, but apply as you feel needed!

tools/serverpod_cli/test/commands/create/ide_test.dart
nielsenko: Fixed

nielsenko:
I have been holding back a bit on merging this PR, since I cannot see the config from inside Antigravity. However if I ask the agent, it does have access as it should:
```
> What mcp servers do you have available?

I have the following MCP servers available:

1. **dart-mcp-server**: Providing Dart and Flutter tools such as code analysis, formatting, package searching, widget inspection, running Flutter Driver commands, hot reload/restart, and connection to the Dart Tooling Daemon (DTD).
2. **serverpod-local** (loaded via plugin): Providing tools for interacting with and managing Serverpod backend servers, including managing database migrations, tailing logs, and hot restarting.
```

So far so good. if you launch `agy` (Antigravity CLI) from a shell that has serverpod and dart in its $PATH then it works fine.

But Antigravity2 is a UI application that is spawned with hardly any environment - in particular it is not spawning mcp commands in the users shell, if any shell at all (I suspect it just exec directly). Even `agy-ide` when spawned from project dir will not set the environment path correctly.

Both see the mcp server config, but fails to actually start them, since the tools are not on the `$PATH`.

I tried setting 
```
{
  "mcpServers": {
    "serverpod": {
      "command": "serverpod",
      "args": ["mcp-server", "--server-dir", "gedefisken_ide_server"],
      "env": {
        "PATH": "<set to the content of $PATH when calling serverpod create>"
      }
    },
    # ...
  }
}
```

but it is not honored. I tried fiddling with getting absolute paths, but then the relative server path doesn't work.. Also we need `dart` and `flutter` in the path as well it gets ugly quickly.

Will merge as is. People can use `agy` and we'll see how many actually use `agy-ide` or _Antigravity 2.0_ UI app. I suspect the situation will improve with time. Compared to the competition Antigravity feels like a train-wreck for now - when it comes to configuration of project local mcp servers.

PR #5142, opened 2026-05-17
feat: Add inheritance and sealed support to exception models

cscoderr:
This PR adds `inheritance` and `sealed` support to exception models, matching the existing capabilities available for regular models.
 
Closes #4249 

Usage:

```yaml
# app_exception.spy.yaml
exception: AppException
sealed: true
fields:
    message: String
```

```yaml
# not_found_exception.spy.yaml
exception: NotFoundException
extends: AppException
fields:
    statusCode: int
```

## Pre-launch Checklist

- [x] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [x] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
- [x] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.
- [x] I added new tests to check the change I am making.
- [x] All existing and new tests are passing.
- [x] Any breaking changes are documented below.

If you need help, consider asking for advice on the [discussion board](https://github.com/serverpod/serverpod/discussions).

## 1. ExceptionClassDefinition constructor now requires isSealed
```dart
// Before
  ExceptionClassDefinition(
    className: 'BaseException',
    fields: [...],
    fileName: 'base_exception',
    sourceFileName: '...',
    serverOnly: false,
    type: ...,
  );

  // After
  ExceptionClassDefinition(
    className: 'BaseException',
    fields: [...],
    fileName: 'base_exception',
    sourceFileName: '...',
    serverOnly: false,
    type: ..,
    isSealed: false, // <-- now required
  );
```
Existing usages should add isSealed: false, unless the exception is intended to be sealed.

## 2. ResolvedInheritanceDefinition is now generic (internal API)
```dart
 // Before
  class ResolvedInheritanceDefinition extends InheritanceDefinition {
    final ModelClassDefinition classDefinition;
  }

  // After
  class ResolvedInheritanceDefinition<T extends ClassDefinition>
      extends InheritanceDefinition {
    final T classDefinition;
  }
```
[...]

PR #5307, opened 2026-06-10
feat: Support `extends` and `sealed` properties on `Exception` models

### LINKED ISSUE #4249: Support `extends` and `sealed` properties on `Exception` models
### Problem to Solve

Currently, Serverpod exception models defined in `.spy.yaml` files do not support inheritance properties (`extends` and `sealed`) that are available for regular model classes. This limitation prevents developers from creating exception hierarchies and sealed exception types, which are common patterns in modern Dart applications.

When building complex applications, developers often need to:
- Create base exception classes with common fields that are shared across multiple exception types
- Use sealed classes to ensure exhaustive pattern matching when handling exceptions
- Organize exceptions into hierarchies for better code organization and type safety

### Proposal

Add support for `extends` and `sealed` properties to exception model definitions in `.spy.yaml` files, mirroring the functionality already available for regular model classes.

### Use Case

Better exception handling for Serverpod apps.

### Alternatives

Create a wrapper exception that contains an `enum` and apply `switch` expressions on this `enum` for exhaustive matching of exception scenarios. This is what is being used on the new auth module, but it is not as idiomatic and straightforward as `sealed` exceptions.

### Additional context

- Regular model classes already support `extends` and `sealed` properties (see `ClassYamlDefinition` and `ModelClassDefinition`).
- The infrastructure for handling inheritance exists in the codebase (e.g., `InheritanceDefinition`, `ResolvedInheritanceDefinition`).
- Exception models currently only support `exception`, `serverOnly`, and `fields` properties (see `ExceptionYamlDefinition`).
- This feature would align exception models with the capabilities of regular model classes, providing a consistent developer experience.

- [ ] I want to work on this

marcelomendoncasoares:
Closes: #4249

This PR continues the work from #5142, adding all required tests, fixing the code generation and adding test models. The generator tests are largely copied between server/client/shared and a nearly 1:1 projection from the extends/sealed class test files - only excluding table-related tests.

While adding the extends/sealed tests for exceptions on shared packages, this PR also closes the gap between the coverage of extends/sealed classes on shared packages.

## Pre-launch Checklist

## Breaking changes

None.


tests/serverpod_test_shared/test/inheritance_exception_test.dart
```diff
@@ -0,0 +1,182 @@
...
+    'when round-tripped through JSON '
+    'then inherited and child fields are preserved.',
+    () {
+      final exception = SharedExtendedAppException(
+        message: 'Base message',
+        detail: 'Extra detail',
+      );
+
+      final json = exception.toJson();
+      final deserialized = SharedExtendedAppException.fromJson(json);
```
nielsenko, line 176: I'd recommend going all the way to `String` in these round-trip tests. There has been bugs lurking in that corner before.

tests/serverpod_test_client/lib/src/protocol/inheritance/exception/extended_app_exception.dart
```diff
@@ -0,0 +1,84 @@
...
+    return {
+      '__className__': 'ExtendedAppException',
+      'message': message,
+      'detail': detail,
+    };
+  }
+
+  @override
+  String toString() {
+    return 'ExtendedAppException(detail: $detail)';
```
nielsenko: Hmm.. what about inherited properties, such as message?

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
```diff
@@ -70,8 +70,100 @@ sealed class ClassDefinition extends SerializableModelDefinition {
   }
 }
 
+/// A shared inheritance and sealed mixin on [ClassDefinition]
+mixin InheritanceClassDefinition<T extends InheritanceClassDefinition<T>>
```
nielsenko: If this wasn't a mixin we would not have to give up on exhaustive switching.
Could do like this:
```dart
  sealed class ClassDefinition extends SerializableModelDefinition { ... }

  sealed class InheritanceClassDefinition<T extends InheritanceClassDefinition<T>>
      extends ClassDefinition {
      // ...
  }

  final class ModelClassDefinition extends InheritanceClassDefinition<ModelClassDefinition> { ... }
  final class ExceptionClassDefinition extends InheritanceClassDefinition<ExceptionClassDefinition> { ... }
```
then `_generateClassLibrary` will not need the unexpected `default: throw..` if I'm not mistaken

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -689,26 +712,26 @@ class Restrictions {
...
+            span,
+          ),
+          ExceptionClassDefinition() => _validateDuplicateInheritedFieldName(
+            currentModel,
+            fieldName,
+            span,
+          ),
+          _ => throw StateError(
+            'Invalid inheritance class type: ${currentModel.runtimeType}',
+          ),
```
nielsenko: If we drop using mixin would this not disappear?

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -2766,3 +2818,13 @@ class Restrictions {
...
   }
 }
+
+extension on InheritanceClassDefinition {
+  String get typeName => switch (this) {
+    ModelClassDefinition() => 'model',
+    ExceptionClassDefinition() => 'exception',
+    _ => throw StateError(
+      'Invalid inheritance class type: $runtimeType',
+    ),
```
nielsenko: If we drop using mixin would this not disappear?

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -467,11 +467,34 @@ class Restrictions {
...
+      var ancestorServerOnlyClass = switch (currentModel) {
+        ModelClassDefinition() => _findServerOnlyClassInParentClasses(
+          currentModel,
+        ),
+        ExceptionClassDefinition() => _findServerOnlyClassInParentClasses(
+          currentModel,
+        ),
+        _ => throw StateError(
+          'Invalid inheritance class type: ${currentModel.runtimeType}',
+        ),
```
nielsenko: If we drop using mixin would this not disappear?

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
```diff
@@ -57,35 +57,59 @@ class SerializableModelLibraryGenerator {
         return _generateExceptionLibrary(classDefinition);
       case ModelClassDefinition():
         return _generateModelClassLibrary(classDefinition);
+      default:
+        throw StateError(
+          'Unhandled ClassDefinition subtype: ${classDefinition.runtimeType}',
+        );
```
nielsenko: If we drop using mixin would this not disappear?

nielsenko (changes requested):
Some suggestion to keep switching exhaustive with using `default:`/`_ =>`

tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5307/commits/62e39481b9900eafd391f279abffb1b74a874212. We could actually get rid completely from the `InheritanceClassDefinition` and move all inheritance/sealed properties to the base `ClassDefinition` directly. Then, most of the code to resolve inheritance just needed a downgrade on the type from `ModelClassDefinition` to `ClassDefinition` and is now working for both models and exceptions.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
marcelomendoncasoares: Removed with https://github.com/serverpod/serverpod/pull/5307/commits/62e39481b9900eafd391f279abffb1b74a874212.

tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart
marcelomendoncasoares: Removed with https://github.com/serverpod/serverpod/pull/5307/commits/62e39481b9900eafd391f279abffb1b74a874212.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
marcelomendoncasoares: Removed with https://github.com/serverpod/serverpod/pull/5307/commits/62e39481b9900eafd391f279abffb1b74a874212.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
marcelomendoncasoares: Removed with https://github.com/serverpod/serverpod/pull/5307/commits/62e39481b9900eafd391f279abffb1b74a874212.

tests/serverpod_test_shared/test/inheritance_exception_test.dart
marcelomendoncasoares: I did not worry to add these here because we don't have anything fancy for encoding - as we had for dynamic. So a basic coverage equal to the one we have for the inheritance on the models is enough.

tests/serverpod_test_client/lib/src/protocol/inheritance/exception/extended_app_exception.dart
marcelomendoncasoares: Good catch! One missing replacement for the fields on the model generator. Fixed at https://github.com/serverpod/serverpod/pull/5307/commits/6092e808a0cc9416e63e90dfe839d076d7f74c22.

nielsenko (approved):
LGTM 🚀 
Nice simplifications on the second round 👍

PR #5316, opened 2026-06-11
fix: Recover orphaned embedded Postgres on Windows after unclean exit

### LINKED ISSUE #5315: Embedded Postgres crash-loops on Windows after an unclean exit
On Windows, if a `Serverpod` server exits uncleanly, the embedded Postgres postmaster is left orphaned. Every subsequent start then fails with `PostmasterLockBusyException` and crash-loops until the user manually kills the `postgres` process. This does not happen on macOS/Linux.

FXschwartz:
On Windows, if the server exits uncleanly (e.g. the TUI/process crashes), the embedded Postgres process gets orphaned and every restart after that crash-loops with `PostmasterLockBusyException` until you manually kill it. Doesn't happen on macOS/Linux.

The supervisor saved the postmaster's executable path without the `.exe` extension, Windows `CreateProcess` tacks it on automatically at spawn time, so launching worked fine. But on the next start, when we check whether an orphan is ours, Windows reports the real path *with* `.exe`, so the two don't match and we end up treating our own orphan as a foreign process, and refuse to clean it up.

The fix is a small `binExecutable` helper that adds `.exe` on Windows, used for the `postgres`, `pg_ctl`, and `initdb` paths.

Fixes #5315 

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/packages/serverpod_embedded_postgres/test/binary/executable_test.dart
+++ b/packages/serverpod_embedded_postgres/test/binary/executable_test.dart
@@ -1,0 +1,27 @@
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+import 'package:serverpod_embedded_postgres/src/binary/executable.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('Given binExecutable', () {
+    var installDir = Directory(p.join('install', '16.13.0'));
+
+    test('when resolving a binary then it lands in the bin/ directory.', () {
+      var path = binExecutable(installDir, 'postgres');
+
+      expect(p.dirname(path), p.join(installDir.path, 'bin'));
+    });
+
+    test(
+      'when on Windows then the name carries the .exe extension, '
+      'otherwise the bare name is used.',
+      () {
+        var name = p.basename(binExecutable(installDir, 'postgres'));
+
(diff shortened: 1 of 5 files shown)


packages/serverpod_embedded_postgres/test/binary/executable_test.dart
```diff
@@ -0,0 +1,27 @@
...
+    test(
+      'when on Windows then the name carries the .exe extension, '
+      'otherwise the bare name is used.',
+      () {
+        var name = p.basename(binExecutable(installDir, 'postgres'));
+
+        expect(name, Platform.isWindows ? 'postgres.exe' : 'postgres');
+      },
+    );
+  });
```
marcelomendoncasoares: Suggestion: Invert the given, since the platform is the test scenario and the when is the call to `binExecutable` ("when getting the binary executable").

marcelomendoncasoares (approved):
LGTM! Only one suggestion for the test, but feel free to merge when addressed.

packages/serverpod_embedded_postgres/test/binary/executable_test.dart
FXschwartz: Done!

PR #5320, opened 2026-06-12
feat: Auth cookie

### LINKED ISSUE #4045: Make web authentication use strict httpOnly cookies
### Problem to Solve

Web authentication regarding token storage currently do not follow best practices. We should never store tokens (session tokens, jwt pairs, etc.) in local storage, indexedDb, etc. We should use cookies and in-memory storage.

### Proposal

We should save tokens in cookies, where we enable httpOnly so that the client (browser/JS) cannot read the token. Cookies are automatically sent by the browser in the `Cookie` header.

Current implementation includes the server sending the token to the client and the client saving it in local storage. My proposal changes this where the server now needs to send a `Set-Cookie` header and the client doesn't do anything since saving of the cookie is handled by the browser as well.

In the case of JWT pairs, only the refresh token should be in cookies. The access token should be in-memory for the lifetime of the token and/or the user's session (session here meaning the time the user uses the app before closing the tab). On loading the app the next time, an API request should be made to get a fresh access token to be used. Access tokens SHOULD NOT be used across tabs or page reloads.

### Use Case

Providing a secure authentication storage mechanism for web applications

### Alternatives

Continue to use the current implementation

### Additional context

_No response_

- [ ] I want to work on this

nielsenko:
Adds opt-in web authentication that carries the session token in a strict HttpOnly cookie instead of the response body, so the token is never reachable by JavaScript. It only ever travels in an HttpOnly, Secure, SameSite Set-Cookie. This closes the XSS token-theft exposure that header/bearer auth has for browser clients.

CSRF is handled in multiple layers:
- SameSite (default Lax), 
- server-side Origin and WS handshake validation against allowedOrigins, and 
- a marker-header requirement (forces a CORS preflight, which the browser blocks if the origin isn't allowlisted).

Fixes: #4045

## Pre-launch Checklist

## Breaking changes

- Auth tokens are no longer accepted via the ?auth= query parameter
- When allowedOrigins is configured, browser WebSocket handshakes with bad Origin are rejected with HTTP 403.

### DIFF
--- a/docs/design/web_auth_httponly_cookies.md
+++ b/docs/design/web_auth_httponly_cookies.md
@@ -1,0 +1,402 @@
+# Design: Web Authentication with httpOnly Cookies
+
+Moves Serverpod's web authentication off JavaScript-readable token storage and
+onto httpOnly cookies. Consolidates the requirement spike and the phased plans
+(issue https://github.com/serverpod/serverpod/issues/4045).
+
+Scope: browser (Flutter/Dart web) clients. Native and desktop clients already use
+OS-backed secure storage (Keychain / encrypted shared preferences) and keep
+header / in-band authentication; they are unchanged.
+
+## Summary
+
+On web today the auth token is held in JS-readable storage and travels in request
+URLs and headers. Target:
+
+- Opaque session token (`AuthStrategy.session`, "SAS"): httpOnly, Secure,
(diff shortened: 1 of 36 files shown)


marcelomendoncasoares:
Closing this, since #5555 already merged the feature.

PR #5321, opened 2026-06-12
refactor: Reload package_config.json in place instead of restarting t…

### LINKED ISSUE #5290: Hook server reload to package_config.json changes via .dart_tool watcher
## Context

In PR #5283, a `.dart_tool` directory watcher was added that correctly detects `package_graph.json` changes and sets `flutterDependenciesChanged`. However, `packageConfigChanged` is currently swallowed when `flutterDependenciesChanged` is true inside the `.dart_tool` scope.

## Problem

`package_config.json` only lives inside `.dart_tool`, and no watch root included `.dart_tool` before PR #5283. Now that the watcher, debounce, and `.dart_tool` resolution are in place, the server is not yet hooked up to react to `package_config.json` changes for server FES (fast edit/save) restarts.

This means package dependency changes are not consistently handled between Flutter and the Server.

## Expected behavior

Server reload behavior should be consistent with Flutter when `package_config.json` changes inside `.dart_tool`.

## Follow-up work

Wire up the server reload trigger to the `.dart_tool` watcher output so that `packageConfigChanged` (or an equivalent signal) from `package_config.json` changes also drives a server restart, not just a Flutter reload.

## References

- PR: https://github.com/serverpod/serverpod/pull/5283
- Comment: https://github.com/serverpod/serverpod/pull/5283#discussion_r3352947331

/cc @marcelomendoncasoares

nielsenko:
KernelCompiler.compile now takes invalidatePackageConfig and appends to the invalidated set, turning the restart + full compile into an in-place incremental recompile, when deps changes.

Fixes: #5290

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start/flutter_dependency_tracker.dart
+++ b/tools/serverpod_cli/lib/src/commands/start/flutter_dependency_tracker.dart
@@ -4,31 +4,12 @@ import 'package:path/path.dart' as p;
 import 'package:serverpod_cli/src/commands/start/package_dependency_tracker.dart';
 import 'package:yaml/yaml.dart';
 
-/// How the Flutter app's dependency closure changed, ordered by the severity
-/// of the response required to pick the change up.
-enum FlutterDependencyChange {
-  /// No change to the closure; source edits are covered by a hot reload.
-  none,
-
-  /// Only pure-Dart dependencies changed; a hot restart picks them up.
-  dartOnly,
-
-  /// A dependency with native code changed; only a full process relaunch
-  /// rebuilds the native host (a hot restart would leave the new native code
-  /// out of the binary, failing at runtime with a MissingPluginException).
-  native,
-
-  /// Assets or fonts declared in `pubspec.yaml` changed. These are bundled at
-  /// build time so only a full process relaunch picks them up.
-  assets,
-}
-
 /// A [PackageDependencyTracker] for a companion Flutter app, adding the
 /// asset/font fingerprint on top of the shared dependency-closure tracking.
 ///
 /// Beyond the closure (native vs. pure-Dart) classification inherited from the
 /// base, a change to the `flutter.assets` or `flutter.fonts` sections of the
-/// app's `pubspec.yaml` maps to [FlutterDependencyChange.assets]: those are
+/// app's `pubspec.yaml` maps to [PackageDependencyChange.assets]: those are
 /// bundled at build time, so only a full process relaunch picks them up.
 class FlutterDependencyTracker extends PackageDependencyTracker {
   /// The package directory of the Flutter app, whose `pubspec.yaml` is read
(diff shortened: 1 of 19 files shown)


marcelomendoncasoares (approved):
Didn't test locally, but the changes LGTM! 🚀

FXschwartz (changes requested):
This looks right, but heads up: the path it touches doesn't actually fire yet. `package_config.json` changes never set `packageConfigChanged`, the watcher drops them in the `_isWithinDartTool` branch.

I confirmed it on a real project: add a dep, pub get, import it, analyzer's happy but the server still can't resolve the package until a full restart.

To make it do something the watcher needs to start emitting `packageConfigChanged` for `package_config.json`.

FXschwartz (approved):
LGTM! 🚀

tools/serverpod_cli/test/commands/start/watch_paths_test.dart
```diff
@@ -0,0 +1,72 @@
...
+        );
+
+        expect(paths, containsAll([serverDartTool, flutterDartTool]));
+      },
+    );
+
+    test(
+      'when the server and Flutter share a resolution (workspace layout), '
+      'then the shared .dart_tool is watched exactly once',
+      () {
```
marcelomendoncasoares, line 54: Comment: Nice optimization!

tools/serverpod_cli/test/commands/start/kernel_compiler_test.dart
```diff
@@ -129,6 +129,99 @@ void main() {
...
+
+    tearDown(() async {
+      await compiler.dispose();
+      await tempDir.delete(recursive: true);
+    });
+
+    test(
+      'when compile is called with invalidatePackageConfig, '
+      'then the new package resolves without a restart',
+      () async {
```
marcelomendoncasoares, line 176: Question: Should we also add tests for added dependencies that contains build hooks? Are we handling them?

marcelomendoncasoares (commented):
Just one question.

tools/serverpod_cli/test/commands/start/kernel_compiler_test.dart
nielsenko: Point taken. Will return

tools/serverpod_cli/lib/src/commands/start/flutter_dependency_tracker.dart
```diff
@@ -24,201 +23,53 @@ enum FlutterDependencyChange {
...
-        }
-      }
-    }
-
-    return {
-      for (final name in closure) name: '${byName[name]?['version'] ?? ''}',
+    return switch (refreshClosure()) {
+      PackageDependencyChange.none => FlutterDependencyChange.none,
+      PackageDependencyChange.dartOnly => FlutterDependencyChange.dartOnly,
+      PackageDependencyChange.native => FlutterDependencyChange.native,
```
marcelomendoncasoares: Question: Do we need the specific `FlutterDependencyChange`? Why not use only the `PackageDependencyChange` to avoid this mapping?

marcelomendoncasoares (commented):
Only one question before approving.

tools/serverpod_cli/lib/src/commands/start/flutter_dependency_tracker.dart
nielsenko: I kept it separate as `FlutterDependecyChange.assets` is unique, but I can collapse and just never have pod emit that.

nielsenko: Fixed in 0461676

PR #5343, opened 2026-06-16
feat: Allow using `unique(per=...)` shorthand for composite unique indexes

### LINKED ISSUE #5336: Allow using `unique` shorthand for multi-column indexes
### Problem to Solve

After #1141, users can create unique indexes just adding the `unique` keyword after a column for single-column indexes. However, for multi-column unique indexes, it is still required to create an explicit index, with an arbitrary name and a much more verbose structure. Ideally, users should also have a way to create multi-column unique indexes using the `unique` shorthand.

### Proposal

Allow using a `per` keyword on after the `unique` property to specify the list of columns that compose the multi-column index (in order).

For single two-column indexes, we could do:

```yaml
class: Example
table: example
fields:
  category: String
  title: String, unique(per=category)
```

Or allow a list to support indexes with more than two columns:

```yaml
class: Example
table: example
fields:
  tenant: String
  category: String
  title: String, unique(per=[tenant, category])
```

The column with the `unique(per=...)` would be the leaf column, with the index listing the columns in order: all previous + unique leaf. The above examples would result in the following, respectively:

```yaml
indexes:
  example__category__title__unique_idx
    fields: category, title
    unique: true
  example__tenant__category__title__unique_idx
    fields: tenant, category, title
    unique: true
```

### Use Case

Creating unique enforcement with the same nice UX as the `unique` shorthand and no verbose index structure.

### Alternatives

Use the verbose indexes only for multi-column indexes.

### Additional context

_No response_

- [x] I want to work on this

marcelomendoncasoares:
Closes: #5336

## Pre-launch Checklist

## Breaking changes

None.


tools/serverpod_cli/test/integration/analyzer/unique_field_e2e_test.dart
```diff
@@ -1,152 +0,0 @@
-import 'package:serverpod_cli/src/analyzer/models/definitions.dart';
```
FXschwartz, line 1: What was the reason for this deletion? The `field_unique_per_test.dart` looks like it covers the model level but I think we are loosing some coverage that a unique field round-trips into a unique `IndexDefinition`

FXschwartz (changes requested):
Just a minor clarification and possible missed test coverage!

Do we need to update the docs for this change as well?

tools/serverpod_cli/test/integration/analyzer/unique_field_e2e_test.dart
marcelomendoncasoares: Forgot to mention this in the description... This file was out of pattern and redundant. It was testing a bridge that is not the responsibility of this feature. We already have coverage for `IndexDefinition` -> `DatabaseDefinition`, so a file that tests both parsing and database definition is redundant on both ends.

FXschwartz (approved):
LGTM!

PR #5344, opened 2026-06-16
feat: Add a `databaseInterceptor` parameter to `Serverpod`

### LINKED ISSUE #5337: Allow intercepting database requests on the server
### Problem to Solve

Despite the `Database` class being defined in the internal `serverpod_database` package, there are some advanced use cases in which having access to intercept database calls with richer info is necessary - such logging, tracing, metrics, lifecycle hooks, policy enforcement and others. Currently, the `Serverpod` class exposes no such option.

### Proposal

Expose a `Database Function(Session session, Database inner)? databaseInterceptor` parameter on the `Serverpod` class that will be invoked when creating each session to use the intercepted database instead.

### Use Case

Allow creating lifecycle hooks (like requested at #4887) or database metrics (a follow-up of #4422).

### Alternatives

Rely on wrapping the `Session` object inside each endpoint to provide a custom `logQuery` method, which is both very limited and error prone.

### Additional context

Since the `Database` class is treated as internal, using this parameter must be documented as accepting possible breaking changes on the interface between minor versions. Given that the feature is targeted at power users for advanced use cases, if properly documented, it should not be a real issue.

- [x] I want to work on this

marcelomendoncasoares:
Closes: #5337

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod/lib/src/server/server.dart
+++ b/packages/serverpod/lib/src/server/server.dart
@@ -53,10 +53,12 @@ class Server implements RouterInjectable {
       throw ArgumentError('Database config not set');
     }
 
-    return DatabaseConstructor.create(
+    final inner = DatabaseConstructor.create(
       session: session,
       poolManager: databasePoolManager,
     );
+
+    return serverpod.databaseInterceptor?.call(session, inner) ?? inner;
   }
 
   /// The [SerializationManager] used by the server.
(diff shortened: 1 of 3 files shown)


Crazelu (approved):
Simple approach. I like!

PR #5357, opened 2026-06-18
feat: Allow running multiple Flutter apps from `serverpod start` and other misc improvements

### LINKED ISSUE #5236: Users should be able to launch more than one Flutter app with `serverpod start`
It is not rare that users need more than one running app instance to test features - like chat, sync, etc. Currently, `serverpod start` will start only one, but we can expose a key like CTRL+N to allow launching another(s).

Each new app instance should get a separate tab on the TUI - which first require a small refactor to how we're dealing with dynamically available tabs.

### LINKED ISSUE #5348: Allow configuring the companion Flutter app path for serverpod start
### Problem to Solve

serverpod start --flutter auto-launches a companion Flutter app, but it only looks for it at a hardcoded location — the <name>_flutter sibling of the server directory. In GeneratorConfig.load (tools/serverpod_cli/lib/src/config/config.dart):

final relativeFlutterPackagePathParts = ['..', '${name}_flutter'];

In projects that don't follow that sibling layout, this path doesn't exist, so hasFlutterPackage is false and the app is silently skipped — the server starts but the app never launches. The client package already has an override (client_package_path in config/generator.yaml), but there is no equivalent for the Flutter package, nor a flag on serverpod start to point at a custom app directory.

### Proposal

Add two overrides for the companion Flutter app path, mirroring client_package_path:

1. A flutter_package_path key in config/generator.yaml (path relative to the server directory).
2. A --flutter-path flag on serverpod start that takes precedence over the config key.

Resolution precedence: --flutter-path > flutter_package_path > <name>_flutter (current default, unchanged — fully backwards compatible).

# config/generator.yaml
type: server

client_package_path: ../../app/my_app_client
flutter_package_path: ../../app/my_app

serverpod start --flutter-path ../../app/admin_app

### Use Case

We run a Melos workspace monorepo with two Flutter apps backed by a single Serverpod server — a user-facing app and an admin app, both under app/:

repo/
├── app/my_app             # user app
├── app/my_admin           # admin app (same server)
├── backend/my_app_server
└── backend/my_app_client

The apps aren't <name>_flutter siblings, so --flutter silently skips them today. And because one server backs multiple apps, a single fixed path can't express the layout — we need to pick, per run, which app to launch alongside the backend:

serverpod start --flutter-path ../../app/my_app     # user app
serverpod start --flutter-path ../../app/my_admin   # admin app

### Alternatives

- Symlink the expected <name>_flutter location to the real app dir — fragile, only one app at a time, and confuses workspace tooling.
- --no-flutter + a separate flutter run — works, but loses the integrated full-stack hot-reload loop serverpod start provides.
- Restructuring an established monorepo to the sibling layout — not viable.

### Additional context

The --flutter-path flag only needs to override relativeFlutterPackagePathParts in GeneratorConfig.load, parallel to how client_package_path already works, so the change is small and isolated. I have a working implementation (config key + flag + integration tests following the Given-When-Then convention) ready.

How experienced are you with this library?

Are you interested in working on a PR for this?

- [x] I want to work on this

- [x] I want to work on this

marcelomendoncasoares:
Closes: #5236
Closes: #5348

This PR introduces a much finer control over the Flutter apps that can be launched with the `serverpod start` command. Now users can:

- Configure multiple Flutter apps to use with the `serverpod start` command.
- Configure which apps are automatically launched or only on demand with `Ctrl+R`.
- Customize launch configs persistently for each app.
- Centralize Flutter app configuration when using the TUI or VS Code debugger.
- Stop running apps from inside the `start` command.
- Ask agents to `spawn_flutter_app` from the MCP.

During the implementation, a bug as also fixed when trying to launch a Flutter app that failed to build. Previously, the `start` command would try to connect anyway and wait forever instead of timing out and releasing the app for relaunch.

## Configuration

To configure multiple apps, modify the `flutter_apps` section of the `pubspec.yaml` file (under `serverpod`):

```yaml
  flutter_apps:
    projectname:
      path: ../projectname_flutter
      auto_launch: true
      target: lib/driver.dart
    Admin:
      path: ../apps/admin_flutter
      auto_launch: true
      device: chrome
      target: lib/main.dart
      web-hostname: 0.0.0.0
      web-port: 8090
      release: true
      pub: false
      dart-define:
        API_URL=https://admin.example.com
        FLAVOR=production
```

## How it works

Each extra launched app will receive a tab with its own logs and breadcrumb. If running in side-by-side mode, the server logs will be shown in a fixed pane on the left, and the apps will be shown in a tab strip on the right. Numbers/tab/arrows navigation will only cycle through the tabs on the right, since the left pane has only one tab.

### Hot reload/restart

Hot reload/restart is per-app, so if a dependency changes in one app, only that app will be reloaded/restarted. When doing a full restart, all apps will be restarted.

### Get app DTD MCP tool

The MCP commands are also updated to support the new app configuration. The `get_flutter_app_dtd` tool now returns a map of app ids to DTD URIs.

### Tail Flutter logs MCP tool

The `tail_flutter_logs` tool now supports an optional `appId` parameter to filter the logs for a specific app. If no appId is provided and there is only one app configured, the logs for that app will be returned. If no appId is provided and there are multiple apps configured, an error will be returned with the list of available app ids.

## Spawn Flutter app MCP tool

The new `spawn_flutter_app` behaves exactly like the `tail_flutter_logs`. If it is called with the `appId` of an already running app, it will ignore and return an exception saying that the app is already running.

## Extra features

While this PR is focused on the multi-app support, it also introduces some extra features that improve the overall user experience.

### Areas and panes

The TUI layout is now based on areas and panes instead of tabs. This abstraction is the base for the dynamic navigation that changes when new apps are launched. It was built in a way that allows us to add more areas in the future without breaking the existing code - following a TMUX/CMUX pattern. These should be pushed down to the `serverpod_tui` in a follow-up PR.

### Flutter driver target

To facilitate the navigation through the app by agents, this PR also adds a `lib/driver.dart` file to the Flutter app template that will be the default target for starting the app on the configuration. The purpose is to launch the app with the `flutter_driver` already enabled for agents to use with the Dart MCP.

## Pre-launch Checklist

## Breaking changes

This PR removes a lot of Flutter-related options from the `serverpod start` command in favor of the configuration through the `serverpod: flutter_apps:` key in the `pubspec.yaml` file.


tools/serverpod_cli/lib/src/commands/start/mcp_server.dart
```diff
@@ -244,37 +254,147 @@ base class ServerpodMcpServer extends MCPServer
...
+      return CallToolResult(
+        content: [TextContent(text: 'No Flutter apps are configured.')],
+        isError: true,
+      );
+    }
+
+    // Resolve which app to launch using the same single-app/multi-app rules as
+    // `tail_flutter_logs`: one configured app is implicit, a missing `appId`
+    // with more than one app is an error naming the options.
+    final (appId, error) = _resolveFlutterAppId(request, ids);
```
nielsenko, line 310: How does the agent learn what ids exists?

tools/serverpod_cli/lib/src/commands/start/mcp_server.dart
```diff
@@ -244,37 +254,147 @@ base class ServerpodMcpServer extends MCPServer
...
+  if (appId == null) {
+    if (ids.length > 1) {
+      return (
+        null,
+        CallToolResult(
+          content: [
+            TextContent(
+              text:
+                  'Multiple Flutter apps are available. Pass `appId` to '
+                  'choose one of: ${ids.join(', ')}.',
```
nielsenko, line 370: If it tries without, it gets a list of valid ids I see.

tools/serverpod_cli/lib/src/commands/start/mcp_server.dart
```diff
@@ -244,37 +254,147 @@ base class ServerpodMcpServer extends MCPServer
...
+  }
+
+  if (!ids.contains(appId)) {
+    return (
+      null,
+      CallToolResult(
+        content: [
+          TextContent(
+            text:
+                'Unknown Flutter app id "$appId". Available: ${ids.join(', ')}.',
```
nielsenko, line 387: Or if it uses a wrong one..

But it cannot ask to have them listed. Nor can it easily relate them to config.. or am I mistaken?

tools/serverpod_cli/lib/src/commands/start/watch_session.dart
```diff
@@ -612,14 +617,22 @@ class WatchSession {
...
+  Future<void> _reloadAllFlutterApps() async {
+    await _flutterManager?.runningAppIds.map(_reloadFlutter).wait;
+  }
+
+  /// Reloads a Flutter app and logs the outcome. Never throws.
+  Future<void> _reloadFlutter(String appId) async {
+    final flutter = _flutterManager?.processFor(appId);
     if (flutter == null) return;
     if (!flutter.isVmServiceConnected) {
       log.debug('Flutter not ready; skipping reload.');
```
nielsenko: Perhaps trace the appId?

tools/serverpod_cli/lib/src/commands/start/watch_session.dart
```diff
@@ -633,9 +646,9 @@ class WatchSession {
...
 
-  /// Hot-restarts the Flutter app and logs the outcome. Never throws.
-  Future<void> _restartFlutter() async {
-    final flutter = _flutterProcess;
+  /// Hot-restarts a Flutter app and logs the outcome. Never throws.
+  Future<void> _restartFlutter(String appId) async {
+    final flutter = _flutterManager?.processFor(appId);
     if (flutter == null) return;
     if (!flutter.isVmServiceConnected) {
       log.debug('Flutter not ready; skipping restart.');
```
nielsenko: Here as well..

tools/serverpod_cli/lib/src/config/flutter_app_config.dart
```diff
@@ -0,0 +1,292 @@
...
+  return serverpodSection.nodes['flutter_apps'];
+}
+
+List<FlutterAppConfig> _synthesizeDefaultFlutterApps({
+  required List<String> serverPackageDirectoryPathParts,
+  required String projectName,
+}) {
+  final relativePathParts = ['..', '${projectName}_flutter'];
+  final app = FlutterAppConfig(
+    id: _slugFromName(projectName, relativePathParts),
```
nielsenko: Should this move to ctor? How does agents learn these ids except through error messages?

tools/serverpod_cli/lib/src/config/flutter_app_config.dart
```diff
@@ -0,0 +1,292 @@
...
+}
+
+/// Properties the parser interprets itself; every other key under an app's
+/// entry is forwarded to `flutter run`.
+const _reservedFlutterAppKeys = {'path', 'auto_launch', 'device'};
+
+/// Builds `flutter run` arguments from the non-reserved properties of
+/// [propsNode], preserving the order they appear in the pubspec.
+List<String> _flutterRunArgsFromProps(YamlMap propsNode, String alias) {
+  final args = <String>[];
```
nielsenko, line 204: Nice 👍

tools/serverpod_cli/test/commands/start/tui/main_screen_layout_test.dart
```diff
@@ -0,0 +1,267 @@
...
+  );
+  await tester.pump();
+  return tester;
+}
+
+void main() {
+  group(
+    'Given a rendered terminal at least as wide as the side-by-side cutoff',
+    () {
+      late NoctermTester tester;
```
nielsenko, line 86: Nice 👍

tools/serverpod_cli/test/commands/start/mcp_server_test.dart
```diff
@@ -478,68 +495,313 @@ void main() {
...
+          server.getFlutterAppIds = () => ['admin', 'customer'];
+          server.onSpawnFlutterApp = (appId) async {
+            spawnCalls.add(appId);
+            return false;
+          };
+        });
+
+        test(
+          'when calling spawn_flutter_app without an appId, '
+          'then it errors listing the available app ids and launches nothing',
```
nielsenko, line 679: I keep looking for an actual way for agents to learn these ids.. I guess it is just in the error messages.

But perhaps that is fine 🤔

nielsenko (commented):
Comments and questions?

tools/serverpod_cli/lib/src/commands/start/mcp_server.dart
marcelomendoncasoares: From the configurations directly. We don't need to expose a tool to list because it can read the `pubspec.yaml` and see the configured apps per id. Maybe we can make it more explicit on the MCP tool description if needed.

nielsenko: But it is not actually the id right. It is the alias, then name, or the path.. which are used to derive the slug.. but the slug itself

marcelomendoncasoares: I'll simplify this and make it the ID.

tools/serverpod_cli/lib/src/config/flutter_app_config.dart
marcelomendoncasoares: Removed at https://github.com/serverpod/serverpod/pull/5357/commits/d04834f80a774871c716a110abb401d59b65e259.

tools/serverpod_cli/lib/src/commands/start/mcp_server.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5357/commits/d04834f80a774871c716a110abb401d59b65e259.

tools/serverpod_cli/lib/src/commands/start/mcp_server.dart
marcelomendoncasoares: Now with https://github.com/serverpod/serverpod/pull/5357/commits/d04834f80a774871c716a110abb401d59b65e259 it is clarified on the MCP tools how to get it.

tools/serverpod_cli/lib/src/commands/start/watch_session.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5357/commits/5ef3b718faec025c6a85bb1ed21e41e469d38832.

tools/serverpod_cli/lib/src/commands/start/watch_session.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5357/commits/5ef3b718faec025c6a85bb1ed21e41e469d38832.

FXschwartz (approved):
Approved! Only one minor nit is it looks like App id is used unsanitized as a filename. The risk is really low but it might cause issues if someone puts a space or invalid character in it.

But I'm happy with or without that change

PR #5364, opened 2026-06-19
feat: Add withSession method on Serverpod for manual session usage

### LINKED ISSUE #3881: Expose `withSession` method on Serverpod class for manual usage
### Problem to Solve

When the user needs a session in places where a `session` object does not exist, it must create the session manually and handle its closing carefully to avoid leakages.

### Proposal

As discussed in [this comment](https://github.com/serverpod/serverpod/issues/3161#issuecomment-3177641013), a better experience would be to use a method that would take care of the session lifecycle automatically, such as:

```dart
Future<T> withSession<T>(Future<T> Function(Session session) callback) async {
  final session = await createSession();
  try {
    final result = await callback(session);
    return result;
  } finally {
    await session.close();
  }
}
```

This way, the user can have the same experience when interacting with the `session` as they have on endpoints.

### Use Case

Below is an example of one such case where the session is handled manually:

https://github.com/serverpod/serverpod/blob/e4dbee3256b6a166353f15b2e9941283cb01851c/examples/chat/chat_server/lib/server.dart#L55-L80

### Alternatives

Manually handing the session, as already happens in the example.

### Additional context

N/A

- [ ] I want to work on this

nicowalter256:
Expose a withSession helper that creates a session, runs a callback with it, and closes it afterwards (including on error), so callers outside of endpoint contexts don't need to manage session lifecycle manually. Updates the chat example to use it.

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/packages/serverpod/CHANGELOG.md
+++ b/packages/serverpod/CHANGELOG.md
@@ -1,6 +1,5 @@
 ## 3.5.0-beta.10
 
-- feat: Adds a `withSession` method to `Serverpod` that creates a session, runs a callback with it, and closes it afterwards, including on error.
 - feat: Hides the "Hot restart" action in the `start` TUI when on watch mode (default).
 - feat: Restarts the Flutter app automatically when its dependencies change.
 - feat: Adds IDE selection screen to the `quickstart` command for a complete experience.
--- a/packages/serverpod/CHANGELOG.md
+++ b/packages/serverpod/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 3.5.0-beta.10
 
+- feat: Adds a `withSession` method to `Serverpod` that creates a session, runs a callback with it, and closes it afterwards, including on error.
 - feat: Hides the "Hot restart" action in the `start` TUI when on watch mode (default).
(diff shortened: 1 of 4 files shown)


packages/serverpod/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 3.5.0-beta.10
 
+- feat: Adds a `withSession` method to `Serverpod` that creates a session, runs a callback with it, and closes it afterwards, including on error.
```
marcelomendoncasoares: Suggestion: Remove the entry, as we do it only when preparing the release. Otherwise we can't assert that all CHANGELOG files are in sync.

marcelomendoncasoares (changes requested):
LGTM! Just revert the change on the CHANGELOG and I'll approve :)

packages/serverpod/CHANGELOG.md
nicowalter256: I have made this fix @marcelomendoncasoares

marcelomendoncasoares (approved):
Thanks for this contribution, @nicowalter256!

Issue #4735, opened 2026-02-16
Allow defining a `relation` on a base classe for inherited table models


marcelomendoncasoares:
### Problem to Solve

Currently, relations can only be defined on table models, making impossible to declare a base class with a relation that should be present on all subclasses. It also mean that such relations would be unidirectional only instead of allowing a bidirectional polymorphic relation on the related table.

### Proposal

Allow a base class of table models to declare the `relation` keyword on fields, such as in the below example:

1. Define a base class `Person` class with a relation to an `Address` class.
    ```yaml
    # person.spy.yaml
    class: Person
    sealed: true
    fields:
        name: String
        address: Address?, relation
    ``` 
    ```yaml
    # address.spy.yaml
    class: Address
    table: addresses
    fields:
        street: String
    ```

2. Then extend the base class to `Employee` and `Guest`.
    ```yaml
    # employee.spy.yaml
    class: Employee
    extends: Person
    table: employees
    fields:
      userPCs: String?
    ```
    ```yaml
    # guest.spy.yaml
    class: Guest
    extends: Person
    table: guests
    fields:
        company: String
    ```

The classes `Employee` and `Guest` would inherit the relation to `Address`. Additionally, the `Address` class could declare a `List<Person>, relation` (if one-to-many) or `Person, relation` (if one-to-one) as a polymorphic bidirectional relation.

### Use Case

Defining relations that should exist for all child models of a hierarchy and extract the children through the related table in a single query as a polymorphic object.

### Alternatives

Copy the relation for each inherited model, but with no support for the polymorphic extraction through the related table class.

### Additional context

As commented on #4647, supporting relations defined on base classes must be implemented carefully due to how relations work on the database:

- **Only relations where the foreign key lives in the subclass table would be supported**. Otherwise, this would mean expecting that a same column on the related table could serve as FK to multiple tables at the same time, which is not possible.
- To support polymorphic [bidirectional relations](https://docs.serverpod.dev/concepts/database/relations/one-to-one#bidirectional-relations), the `name` parameter would have to accept a pattern to be checked against subclasses. The type of the model on the related class would have to be the base type.

This issue is a more detailed specification of the request created at #4648.

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

PR #5368, opened 2026-06-19
feat: Allow defining a relation on a base class for inherited table models (#4735)

rgiese:
This change presents a scoped/partial solution to #4735.

## Problem to solve

The part of #4735 this is addressing: _Currently, relations can only be defined on table models, making impossible to declare a base class with a relation that should be present on all subclasses._

As requested, this change allows a (non-table) base class to declare a `relation` that is then automatically implemented as a foreign key on derived (`extends:`-ing) `table` classes. The `relation` can be implicit (i.e. just `relation`) or specify the field establishing the relationship (e.g. `relation(field=parentId)`).

The part of #4735 this is intentionally leaving unsolved: _It also mean that such relations would be unidirectional only instead of allowing a bidirectional polymorphic relation on the related table._

I'm proposing that we do not (yet?) enable a back-reference from the destination of the relationship. 

If we _did_ enable that, it means the destination class would need to either enumerate the derived `table` classes (which feels like it would break layers of abstraction), _or_ it would need to reference the parent/base class which one could then either treat universally or try to downcast from.

Neither of these is wrong; however, I think this current change already provides self-contained value without back-references, has the benefit of fewer lines of code changed, and does not preclude back-references from being supported in the future. (For full disclosure, this is perhaps just an elaborate rationalization of "this solves _my_ problem of the day and I did not want to scope-creep this PR".)

## Additional considerations implemented

I am proposing that we make this capability opt-in because its success path is otherwise difficult to distinguish from the failure path of "I declared a `relation`, forgot to make it a `table` class, now I am confused why things aren't working".

To that end, I am proposing a new table-level tag, `tableBase: true`, to make it explicit that a given `table`-less base class is exactly that (a base class for `table` classes), which then activates this feature.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
@@ -187,6 +187,10 @@ final class ModelClassDefinition extends ClassDefinition {
   /// If set to true the class is immutable.
   final bool isImmutable;
 
+  /// If set to true, this class is a base for table-backed subclasses and may
+  /// declare relations without having a table itself.
+  final bool isTableBase;
+
   /// If set, the default data type used for serialization of the JSON columns in this class.
   /// It can be overridden for each field.
   final SerializationDataType? serializationDataType;
@@ -202,6 +206,7 @@ final class ModelClassDefinition extends ClassDefinition {
     required super.type,
     required super.isSealed,
     required this.isImmutable,
+    this.isTableBase = false,
     super.childClasses,
     super.extendsClass,
     this.database = ModelDatabaseDefinition.server,
--- a/tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/definitions.dart
@@ -189,6 +189,10 @@ final class ModelClassDefinition extends ClassDefinition {
   /// If set to true the class is immutable.
   final bool isImmutable;
 
+  /// If set to true, this class is a base for table-backed subclasses and may
+  /// declare relations without having a table itself.
+  final bool isTableBase;
+
   /// If set, the default data type used for serialization of the JSON columns in this class.
   /// It can be overridden for each field.
   final SerializationDataType? serializationDataType;
@@ -204,6 +208,7 @@ final class ModelClassDefinition extends ClassDefinition {
     required super.type,
     required super.isSealed,
     required this.isImmutable,
+    this.isTableBase = false,
     super.childClasses,
     super.extendsClass,
     this.database = ModelDatabaseDefinition.server,
--- a/tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
@@ -194,6 +194,10 @@ class ModelDependencyResolver {
     var relation = fieldDefinition.relation;
     if (relation is! UnresolvedObjectRelationDefinition) return;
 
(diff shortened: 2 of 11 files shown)


rgiese:

rgiese:
Tagging @marcelomendoncasoares as the "most likely to be interested in this" party. :)

rgiese:
recheck

marcelomendoncasoares (changes requested):
Hi @rgiese! Sorry for the delay in getting back to you. I agree with your approach to postpone the backlink on the related table, as allowing the `relation` on base tables already provides a lot of value. However, the PR still requires a bit more work to inherited relations completely.

This feature might look simple on the surface, but lifting the restrictions and implementing a `tableBase` keyword is only the starting point. The current tests are not seeing it because they are only validating the first inherited table, but the second onward will get into trouble.

Firstly, the `ModelDependencyResolver` will mark the relation as resolved on the first and will skip resolving for the other models. Then, when creating the model, we must ensure that the relation is not stored on the same Dart object for all inherited classes (as we do with the `id` by cloning the field instead of reusing).

Tests for this feature also need a broader coverage, including at least:

- Base tables must receive the `id` field and apply all same restrictions to it as `table` models.
- One parent class with relation (object and FK field) and two children, validating:
  - that the relation is resolved as expected on each child.
  - that the columns are independent between child models (not the shared Dart object from the parent).
- One parent with object relation and a child class that declares a field that clashes on the automatically generated FK id.
- One child that declared a relation on an inherited field that already contains a relation (for object and FK field relations).
- Integration test models on the `serverpod_test_server` that use the feature for real.

rgiese:
@marcelomendoncasoares Just to make sure I'm understanding your feedback correctly, you're pointing out that the first layer (abstract-base -> derived-class) will work correctly, but subsequent layers (abstract-base -> derived-class -> derived-derived-class) would not (if so, I 100% did not think about that), _or_ that the first derived class will work (abstract-base -> derived-class) but a second peer class (abstract-base -> derived-class-2) would not?

marcelomendoncasoares:
> _or_ that the first derived class will work (abstract-base -> derived-class) but a second peer class (abstract-base -> derived-class-2) would not?

This. ☝️ 

The first layer will fail for the second child of the same parent (parent -> [child1, child2]). Since the relation is stored on the `SerializableModelFieldDefinition` Dart object of the parent class, the first child will resolve it and mark as resolved. The second child will skip on the following guard:

https://github.com/serverpod/serverpod/blob/105c001bec8c4040d8710735000fc3d3262c0894/tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart#L194-L195

This is why we need to ensure that fields with relations are copied into each child instead of passing through from the parent by `fieldsIncludingInherited`. It is what we do for the `id` field when resolving:

https://github.com/serverpod/serverpod/blob/105c001bec8c4040d8710735000fc3d3262c0894/tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart#L78-L134

Then we skip passing it onwards from the parent on inherited fields:

https://github.com/serverpod/serverpod/blob/105c001bec8c4040d8710735000fc3d3262c0894/tools/serverpod_cli/lib/src/analyzer/models/definitions.dart#L250-L253

https://github.com/serverpod/serverpod/blob/105c001bec8c4040d8710735000fc3d3262c0894/tools/serverpod_cli/lib/src/analyzer/models/definitions.dart#L265-L275

rgiese:
Awesome, thank you for the detailed analysis. It might take me a week or so to get this ironed out, but I'm on it.

Abdurrahman98XX:
news on this?

PR #5371, opened 2026-06-20
feat: Add a `noReturn` to all ORM methods that return a list

### LINKED ISSUE #5118: A database updateWhere method that does not return all the updated rows
### Problem to Solve

The `updateWhere` method is a good fit for when lots of rows need to be updated, for example in a large data migration. But in some cases the updated objects are not necessary to retrieve, and if a large data set was changes this can be slow, occupy resources, and even cause the operation to abort due to taking too long.

### Proposal

To enable using `updateWhere` for large data sets, there should be a version which only returns the number of affected rows (this is PostgreSQL standard behavior for UPDATE).

### Use Case

Updating large sets of data, for example in data migrations with large tables.

### Alternatives

Breaking up the update work in smaller chunks, or use the pagination functionality. But this still has the unnecessary overhead, and requires more complex user code.

### Additional context

_No response_

- [ ] I want to work on this

marcelomendoncasoares:
Closes: #5118

Adds a `noReturn` parameter to all ORM methods that return a list of rows. When set to `true`, the rows are not read back from the database and an empty list is returned.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/examples/legacy/chat/chat_server/lib/src/generated/channel.dart
+++ b/examples/legacy/chat/chat_server/lib/src/generated/channel.dart
@@ -335,16 +335,22 @@ class ChannelRepository {
   /// If [ignoreConflicts] is set to `true`, rows that conflict with existing
   /// rows are silently skipped, and only the successfully inserted rows are
   /// returned.
+  ///
+  /// If [noReturn] is set to `true`, the inserted rows are not read back from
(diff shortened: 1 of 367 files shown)


nielsenko (approved):
LGTM 👍

PR #5387, opened 2026-06-24
feat: Improve serverpod start TUI experience

### LINKED ISSUE #5218: Improvements to `serverpod start` TUI interface
## Startup sequence

- [ ] We should always have a bit more padding around the "Serverpod" start text. If the window is narrow, it does not look great. (It's even worse than the picture if it's a bit narrower the borders will overlap)

Image

## Interface

- [ ] There is an extra white character at the start of the top tab bar. It should be removed.

- [ ] Rename `Flutter app |` to just `Flutter:`. This will avoid the bad visual tangent to the `Serverpod logs` tab.

- [ ] If easy, apply a shimmer effect to the Flutter status messages while things are loading.

- [ ] The different Flutter status messages currently have different capitalization and punctuation (may have been fixed). Keep it consistent. Generally, avoid the `...`. Use the following sequence:
  - Launching Flutter app
  - Connecting to Chrome

- [ ] There are T-shaped characters for drawing boxes, we should use them instead of having small gaps in the UI.
Image

- [ ] The divider (in split screen) should go all the way up to the top bar.
Image

### Help screen

The help screen has too much information. No need to repeat the keys already displayed in the bottom bar or showing multiple options for the same thing. It can also use some other refinements.

- [ ] For text and outline, use the same purple as all other highlights.
- [ ] Remove `Help` from the top bar, instead add a blank line followed by `Help` (bold face) and a blank line.
- [ ] We only need to show the navigation options, keep the `Shift-Up/Down` and `Home / g` and `End G`, and `Tab` `Next tab`. Remove the other options (5 lines total).
- [ ] `Esc to close` -remove the `H` option. Make `Esc` bold and purple.
- [ ] Same indentation for `Help` and the options.
- [ ] Clicking anywhere on the screen should dismiss the Help dialog.

Crazelu:
This PR harmonises the look and feel of the serverpod start TUI. The help options have been trimmed down for simplicity. Most of the existing shortcuts are still supported even though they are no longer shown in help. The only shortcut that has been replaced is ``period (.)`` to show raw server logs. It is replaced with `S`.

The behaviour of CTRL+R for launching apps is now consistent regardless of how many apps are configured.
If at least one app is configured, then CTRL+R will show the apps panel instead of automatically launching the app.
The Flutter status line gains a shimmer when the app is still launching and stops when the app is launched. Stopping a running app from the TUI now produces a Flutter status message for the top bar.

The key events handling has been tweaked to allow for raw server logs and apps panel events to be treated correctly.

Closes #5218 

https://github.com/user-attachments/assets/48e7a9dc-f6f7-4d98-b4b7-f1a1b34ace07

https://github.com/user-attachments/assets/99f9c14d-b4fc-4020-bed4-587ad229bac2

## Pre-launch Checklist

## Breaking changes

None


tools/serverpod_cli/test/commands/start/watch_loop_test.dart
```diff
@@ -187,6 +187,7 @@ void main() {
         onProgress: (_, _) {},
         onReady: (_, _) {},
         onStart: (_, _) async {},
+        onStop: (_) {},
```
nielsenko, line 190: Not related to this PR, but perhaps we should take an interface instead of all these individual callbacks.
It is getting a bit messy

nielsenko (approved):
LGTM 🚀 

I did think perhaps our keyboard mappings could benefit from being more intuitive.
Perhaps we should bike-shed it 😬 on SU. This is not related to this PR.

Crazelu:
@serverpod/framework 
As discussed, app tabs for stopped apps can now be closed by pressing X.

https://github.com/user-attachments/assets/3e7d975b-7085-4b97-ac72-1461c8f38513

PR #5137, opened 2026-05-15
feat: Add PostGIS geography type support

charlesarchibong:
## PostGIS Geography Type Support

This PR adds native PostGIS geography type support to Serverpod, enabling server-side spatial data modeling with `GeographyPoint`, `GeographyLineString`, `GeographyPolygon`, and `GeographyGeometryCollection` column types.

### What's changing

**New column types** — four geography types can now be declared in `.spy.yaml` model files:
- `GeographyPoint` → maps to `geography(Point,4326)` in PostgreSQL / `TEXT` in SQLite
- `GeographyLineString` → `geography(LineString,4326)` / `TEXT`
- `GeographyPolygon` → `geography(Polygon,4326)` / `TEXT`
- `GeographyGeometryCollection` → `geography(GeometryCollection,4326)` / `TEXT`

**Dart serialisation** — new types in `serverpod_serialization` with full WKT/GeoJSON/WKB parse & encode support, `copyWith`, and null-safe equality.

**Database integration** — `ColumnType` enum extended; Postgres dialect emits `CREATE EXTENSION IF NOT EXISTS postgis` and correct `geography(...)` DDL; SQLite stores as `TEXT`.

**Code generation** — `serverpod_cli` generator maps each geography type to its Dart class and SQL type string.

**Docker** — the local test Postgres container (`docker-local/`) now uses a custom image based on `pgvector/pgvector:pg16` with `postgresql-16-postgis-3` installed, so both pgvector and PostGIS are available in development.

**Tests** — serialisation unit tests, column query tests, spatial-operations tests (`ST_DWithin`, `ST_Intersects`, `ST_Distance`, `ST_Covers`), code-generation tests for both Postgres and SQLite dialects.

### Issues fixed

Fixes #2439

## Pre-launch Checklist

- [x] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [x] This update contains only one single feature or bug fix and nothing else.
- [x] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with `dart format`.
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.
- [x] I added new tests to check the change I am making.
- [x] All existing and new tests are passing.
- [x] Any breaking changes are documented below.

## Breaking changes

[...]

PR #5392, opened 2026-06-25
feat: Support GiST and SP-GiST indexes on geography columns

### LINKED ISSUE #5389: Support GiST and SP-GiST indexes on geography columns
### Problem to Solve

PostGIS spatial queries benefit from GiST and SP-GiST indexes, but Serverpod does not yet support declaring these index types on geography model fields. Without them, proximity and spatial-filter queries may not perform well at scale.

### Proposal

Extend Serverpod's index support so geography model fields can declare GiST and SP-GiST indexes. Generated migrations and table definitions should create the appropriate PostGIS index types when specified in the model.

### Use Case

A store-locator API filters locations with `ST_DWithin` and orders results by distance. Developers add a spatial index on the geography column in their model so common location queries stay fast as the dataset grows.

### Alternatives

- Rely on sequential scans and accept slower queries at scale.
- Manage spatial indexes manually outside Serverpod's migration and code generation workflow.

### Additional context

Follow-up from https://github.com/serverpod/serverpod/pull/5137. GiST/SP-GiST support was intentionally scoped out of the initial PostGIS geography PR.

charlesarchibong:
Follow-up to #5137. This adds the index piece we left out of the initial PostGIS PR, so geography fields can declare GiST and SP-GiST indexes for faster spatial queries.

You can now set `type: gist` or `type: spgist` on an index over a geography field. If no type is given it defaults to `gist`, which is the usual choice for spatial data. The other index types (`btree`, `gin`, `brin`, `hash`) are rejected on geography fields with an error since they don't work on geography columns. I also disallowed mixing geography and non-geography fields in the same index, matching what we already do for vector fields. Multiple geography columns in one index is allowed though, since a multi-column GiST index is valid in PostGIS.

Most of the plumbing was already there (the generic `USING <type>` path handles gist/spgist fine), so the changes are small and stick close to the existing pgvector handling: a default in `_parseIndexType`, the restriction in `validateIndexType`, the field-mixing guard in `_validateIndexFieldNames`, and an `ObjectWithGeographyIndex` model for the integration test.

## Testing

- Validation: default falls back to `gist`, `gist`/`spgist` are accepted, unsupported types and geography/non-geography mixing are rejected, and a multi-geography index is allowed.
- SQL generation: `gist`/`spgist` produce `CREATE INDEX ... USING gist/spgist (...)` with no operator class.
- Integration (`index_test.dart`): the generated migration creates the indexes, checked by analysing the database schema (default, explicit gist, explicit spgist).

I also confirmed against PostGIS 3.6 that both gist and spgist work on all four geography types (Point, LineString, Polygon, GeometryCollection), so there's no per-type restriction needed.

One thing worth a note for the docs follow-up: SP-GiST on the geography type needs a recent enough PostGIS. It works on the version we run in CI, and a declared index on an older install would fail at migration time with a clear Postgres error, same as the existing PostGIS extension requirement.

Closes #5389.


charlesarchibong:
@marcelomendoncasoares, this is the GiST/SP-GiST follow-up from #5137 that you assigned me on #5389, ready for review whenever you get a chance. Thanks again for all the guidance on the first one! 🙏

charlesarchibong:
@marcelomendoncasoares the failing "Unit tests IO (3.41.5)" check isn't a real failure, the tests passed, and the runner was killed with exit 143 ("received a shutdown signal") right at the end. Mind re-running the failed job? Should be green on a fresh runner. Thanks!

tests/serverpod_test_server/lib/src/models/object_with_geography_index.spy.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+indexes:
+  geography_index_default:
+    fields: point
+  geography_index_gist:
+    fields: pointIndexedGist
+    type: gist
+  geography_index_spgist:
+    fields: pointIndexedSpgist
+    type: spgist
```
marcelomendoncasoares: Suggestion: Instead of adding one file with indexes, add the indexes to the existing `ObjectWith*` for the Geography classes as we do with vector. This will allow testing the index generation for all types - each object should receive the whole set of indexes.

tests/serverpod_test_server/test_integration/index_test.dart
```diff
@@ -410,4 +410,60 @@ void main() async {
...
     },
   );
+
+  group(
+    'Given declared ObjectWithGeographyIndex class when analyzing database schema',
+    () {
+      late List<IndexDefinition> indexes;
+
+      setUpAll(() async {
+        var databaseDefinition = await session.db.analyzer.analyze();
```
marcelomendoncasoares, line 420: Comment: After moving indexes to the `ObjectWith*` classes, we should expand this test to validate all combinations - as we do with Vector, HalfVector, SparseVector and Bit.

marcelomendoncasoares (changes requested):
Great work again, @charlesarchibong! I have only one suggestion to expand the indexes coverage on the tests to all Geography test models, so we validate the full set. Apart from this, changes looks good!

tests/serverpod_test_server/lib/src/models/object_with_geography_index.spy.yaml
charlesarchibong: Done, replaced the single model with one `ObjectWith*` per geography type (Point, LineString, Polygon, GeometryCollection), each carrying the full index set (default → gist; explicit gist; explicit spgist), mirroring the vector models.

tests/serverpod_test_server/test_integration/index_test.dart
charlesarchibong: Done, `index_test.dart` now has a group per geography type validating every default/gist/spgist combination, like the Vector/HalfVector/SparseVector/Bit groups. All 12 combinations are verified against PostGIS.

charlesarchibong:
Thanks @marcelomendoncasoares! Addressed both, there's now an `ObjectWith*` model per geography type (Point, LineString, Polygon, GeometryCollection), each with the full default/gist/spgist index set, and `index_test.dart` validates every combination across all four types. Ready for another look 🙏

charlesarchibong:
@marcelomendoncasoares the red "Unit tests IO (3.41.5, windows-latest)" is a flaky Flutter app-launch test (`flutter_app_manager_test.dart` → "Failed to launch" on the Windows runner), unrelated to this change. Could you re-run the failed job when you get a chance? 🙏

Also put up the docs for this feature: serverpod/serverpod_docs#661

marcelomendoncasoares (approved):
LGTM! 🚀

Thank you one more time for this great contribution, @charlesarchibong!!

PR #5394, opened 2026-06-25
feat: Add the tail keyword to configure inherited fields order on models

### LINKED ISSUE #5355: Add posility to change position of database columns which are inherited from parent model
### Problem to Solve

Let's say I have a base `Entity` model
```yaml
class: Entity
fields:
  - id
  - createdDT
  - updatedDT
```
which is inherited
```yaml
class: Child
table: childs
extends: Entity
fields:
  - name
```
When will creates all inherited fields will be first columns of table, i.e.
```SQL
CREATE TABLE "childs" (
  "id" ...,
  "createdDT" ...,
  "createdDT" ...,
  "name" ...
  ...
)
```
But basically audit columns are not so important to be first.

### Proposal

So I suggest to create a field property `tail` (for example) which will notify SQL generator to place such columns after `Child`'s last column.
```yaml
class: Entity
fields:
  - id
  - createdDT, tail
  - updatedDT, tail
```

### Use Case

 It would be good to move them to the end on the create table SQL query
```SQL
CREATE TABLE "childs" (
  "id" ...,
  "name" ...
  ...
  "createdDT" ...,
  "createdDT" ...,
)
```

### Alternatives

Or create a field property similar to `position(seq=N)` (where N is the index number), which will allow the column position to be changed upon initial creation. The column with the higher index will be moved to the right.

### Additional context

_No response_

- [ ] I want to work on this

marcelomendoncasoares:
Closes: #5355

## Pre-launch Checklist

## Breaking changes

None.


tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -2088,6 +2088,23 @@ class Restrictions {
     return [];
   }
 
+  List<SourceSpanSeverityException> validateTailKey(
+    String parentNodeName,
+    String key,
+    SourceSpan? span,
+  ) {
+    if (parentNodeName == defaultPrimaryKeyName) {
```
FXschwartz, line 2096: This is a nit and an edge case. But I believe this would reject usage on an ID field for a non table class.

FXschwartz (approved):
One super minor nit. Otherwise looks great!

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
marcelomendoncasoares: It is intentional, because non-table classes can be base classes for table classes, and then the `tail` would be inconsistently ignored.

PR #5395, opened 2026-06-26
fix: AGENTS.md proposes using the Dart CLI instead of the MCP server.

vlidholt:
Minor change to ask AI agents to use the CLI for `dart format` and `dart analyze` over using the MCP server. The reason is that there seems to be a bug in the MCP server that often causes it to hang.


templates/serverpod_templates/projectname/AGENTS.md
```diff
@@ -14,11 +14,11 @@ ALWAYS use the MCP server instead of the command line. Use the MCP server to:
...
 
-1. `dart analyze` (`dart` MCP)
-2. `dart format` (`dart` MCP)
-3. `create_migration` and `apply_migrations` (only if necessary)
+1. `dart analyze` (CLI)
+2. `dart format` (CLI)
+3. `create_migration` and `apply_migrations` (MCP - only if necessary)
 4. Do `serverpod` MCP `hot_restart` if required (hot reload is done automatically). Will also hot restart Flutter app
-5. Run tests, if applicable (`dart` MCP)
+5. Run tests, if applicable (`dart` CLI)
```
marcelomendoncasoares, line 21: Discussion: Before we give up and instruct the agent to use the CLI, could you try to add the `--force-roots-fallback` flag to the `dart` MCP command, @vlidholt? This is a possible fix [recommended on the MCP page](https://docs.flutter.dev/ai/mcp-server#overview) that I have read around to be required for Claude, Cursor and Codex (the latter is explicitly mentioned on the page).

Both `analyze` and `format` tools (among others) have a `roots` parameter that actually have no description, so the agent might no be passing the required argument there.

Image

I found some people recommending to [tell the agent to pass the correct root argument](https://www.linkedin.com/posts/peter-trost-builds_getting-errors-when-claude-uses-the-dart-activity-7371243334868037632-VGUr), but seems like a weaker approach for me than the `--force-roots-fallback`.

My preference for trying to fix the MCP before giving up is because there are other useful tools on the `dart` MCP that won't also work without correct roots, and it would be exhaustive to try to list them all for the agent to ignore on the `dart` MCP.

vlidholt: I have tried it now, and it still does not work. I suggest we apply this until we can figure out if it's fixable with the MCP.

marcelomendoncasoares (approved):

PR #5420, opened 2026-07-01
feat: Serve flutter web app under root if website is not enabled

Crazelu:
The Flutter web app will now be served under `/` if Flutter app only option is enabled for the web server.

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tests/bootstrap_project/test_perform_create/web_test.dart
+++ b/tests/bootstrap_project/test_perform_create/web_test.dart
@@ -639,10 +639,13 @@ void main() {
 
       group('Given the Flutter web app is built and the pod is restarted', () {
         setUp(() async {
-          final flutterBuildProcess = await startProcess(
-            'serverpod',
+          final flutterBuildProcess = await startServerpodCli(
             ['run', 'flutter_build'],
+            rootPath: rootPath,
             workingDirectory: project.serverDir,
+            environment: {
+              'SERVERPOD_HOME': rootPath,
+            },
           );
           expect(await flutterBuildProcess.exitCode, 0);
 
--- a/tests/bootstrap_project/test_perform_create/web_test.dart
+++ b/tests/bootstrap_project/test_perform_create/web_test.dart
@@ -357,7 +357,7 @@ void main() {
       );
 
       test(
-        'then theserver.dart has a fallback middleware for the default route',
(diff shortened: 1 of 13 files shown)


tests/bootstrap_project/test_perform_create/web_test.dart
```diff
@@ -327,6 +331,127 @@ void main() {
...
+          );
+          final content = await serverFile.readAsString();
+          expect(content, contains('// Serve the flutter web app under /.'));
+          expect(content, contains("'/'"));
+          expect(content, contains("'/**'"));
+        },
+      );
+
+      test(
+        'then theserver.dart has a fallback middleware for the default route',
```
marcelomendoncasoares:
```suggestion
        'then the server.dart has a fallback middleware for the default route',
```

marcelomendoncasoares (changes requested):
Nice work! Both website are showing on the root now according to the chosen option. Two suggestions only:

- It would be good to add two tests on the bootstrap (one for each scenario) that validates the correct resource being accessed on the `localhost:8082` after starting the server.
- When using the `serverOnly` -> `Website` context, the resulting page is the same that we use for the app, with a button that will point to a 404. We should have a different placeholder in this case.
    Image

Crazelu:
> When using the serverOnly -> Website context, the resulting page is the same that we use for the app, with a button that will point to a 404. We should have a different placeholder in this case.

@marcelomendoncasoares  this is now fixed in [6b9d044](https://github.com/serverpod/serverpod/pull/5420/commits/6b9d04437c6b2073af7e47c156f1095f209b1d91)

Crazelu:
> It would be good to add two tests on the bootstrap (one for each scenario) that validates the correct resource being accessed on the localhost:8082 after starting the server.

Done in [94eae73](https://github.com/serverpod/serverpod/pull/5420/commits/94eae739b3c0cb226425612b3129e3db942b81fa). There was an important fix that needed to be made for server only projects to only setup website by default as well -> [035302c](https://github.com/serverpod/serverpod/pull/5420/commits/035302c084f7dc10f513c6e97e39a38a9d2a1142)

marcelomendoncasoares (approved):
Great work with the tests! 🚀 🚀 

Once CI is green, feel free to merge.

PR #5428, opened 2026-07-03
fix: Fix `upsert` with `updateWhere` that matches rows on SQLite

marcelomendoncasoares:
The SQLite adapter implements batch `upsert` by decomposing it into single-row upserts and then applying a duplicate-conflict heuristic that compares the number of distinct returned ids against the number of *input* rows. This conflates two situations:

- A genuine duplicate when two input rows resolve to the same conflict target. In such cases, Postgres raises "ON CONFLICT DO UPDATE command cannot affect row a second time".
- A row whose `DO UPDATE` is legitimately skipped by `updateWhere`, which returns no row and, therefore, makes the counts differ.

There was a bug in the implementation that would falsely throw an integrity-constraint-violation error in the latter case and roll back the transaction.

The fix is to detect duplicates among the *returned* ids, which still catches the true "same row affected twice" case while no longer counting rows skipped by `updateWhere`.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
+++ b/packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
@@ -337,10 +337,12 @@ class SqliteDatabaseConnection extends DatabaseConnection<SqlitePoolManager> {
             ),
         ];
 
-        // NOTE: Since we transform batch inserts into multiple single-row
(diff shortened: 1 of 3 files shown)


FXschwartz (approved):
LGTM!

PR #5434, opened 2026-07-06
feat: Remove experimental flag from `serverpod start` command

developerjamiu:
The `serverpod start` command was marked as experimental for the 3.5.0-beta release, but the flag was not removed for 4.0.0 as intended. This PR removes the `EXPERIMENTAL!` description prefix and unhides the command, so it appears in `serverpod --help` and in the generated CLI reference docs.

## Pre-launch Checklist

- [ ] I added new tests to check the change I am making. (No behavior to test — existing tests pass.)

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -112,8 +112,9 @@ class StartCommand extends ServerpodCommand<StartOption> {
 
   @override
   final description =
-      'Generate code and start the server. '
-      'Use --watch to watch for changes and hot reload.';
+      'Start the full development stack with hot reload: generates code, '
+      'runs the server, and launches the companion Flutter apps in an '
+      'interactive terminal UI.';
 
   @override
   String get invocation => 'serverpod start [-- <server-args>]';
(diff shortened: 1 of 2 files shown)


tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -110,12 +110,9 @@ class StartCommand extends ServerpodCommand<StartOption> {
...
   final name = 'start';
 
-  @override
-  bool get hidden => true;
-
   @override
   final description =
-      'EXPERIMENTAL! Generate code and start the server. '
+      'Generate code and start the server. '
       'Use --watch to watch for changes and hot reload.';
```
marcelomendoncasoares: Suggestion: The description of the command seems rather simplistic for what it does now. We should mention the "fullstack" experience with hot reload, not only the generate + run server.

marcelomendoncasoares (changes requested):
Only one suggestion around how we present the `start` command to users.

tools/serverpod_cli/lib/src/commands/start.dart
developerjamiu: Updated

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5436, opened 2026-07-07
feat: Make the WidgetRoute's build method return a nullable WebWidget

### LINKED ISSUE #4855: Make the `WidgetRoute`'s `build` method return a nullable `WebWidget`
The web server's `WidgetRoute` build method currently returns `WebWidget`. It would be helpful if it could also return null. I.e., `WebWidget?`. If `null` is returned, the route would create a 404 (not found) response.

Current implementation:

```dart
/// Override this method to build your web widget from the current [session]
/// and [request].
Future<WebWidget> build(Session session, Request request);
```

Proposed implementation:

```dart
/// Override this method to build your web widget from the current [session]
/// and [request].
Future<WebWidget?> build(Session session, Request request);
```

With this update, the `handleCall` method would check if the value returned by the `build` method is `null` and in that case return a `Response.notFound()`.

```dart
final widget = await build(session, context);
if (widget == null) {
  return Response.notFound(body: Body.fromString('Not found'));
}
```

The use case is that we sometimes catch multiple paths in the router (e.g., `/blog/**`), but we don't know if there is a matching entry until we build the route (e.g., `/blog/non-existing-entry`). By handling `null` as a 404 Not Found, it would be an easy case to handle.

Crazelu:
Closes #4855 

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/packages/serverpod/lib/src/web_server/web_server.dart
+++ b/packages/serverpod/lib/src/web_server/web_server.dart
@@ -457,7 +457,7 @@ abstract class WidgetRoute extends Route {
 
   /// Override this method to build your web widget from the current [session]
   /// and [request].
-  Future<WebWidget> build(Session session, Request request);
+  Future<WebWidget?> build(Session session, Request request);
 
   @override
   FutureOr<Result> handleCall(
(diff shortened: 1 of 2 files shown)


marcelomendoncasoares (approved):
LGTM! 🚀

PR #5438, opened 2026-07-07
fix: Allow unverified emails in Firebase IDP default account validation

### LINKED ISSUE #5154: Firebase IDP default email validation blocks normal signup flows
## Problem

`FirebaseIdpConfig`'s default `firebaseAccountDetailsValidation` rejects any signup where `email` is present but `email_verified` is `false`. `FirebaseIdpConfigFromPasswords()` (the constructor recommended in the docs) inherits this strict default with no obvious way to discover it.

This breaks a very common UX pattern: a user signs up via Firebase Email/Password and is auto-signed-in immediately, before they have clicked the verification link Firebase emails them. The very first `controller.login(user)` call from the Flutter client fails.

## Reproduce

1. Configure the server with `FirebaseIdpConfigFromPasswords()` (no overrides).
2. From a Flutter client, sign up a brand-new user via Firebase Email/Password. Firebase creates the account with `email_verified: false`.
3. Call `controller.login(firebaseUser)` immediately after signup.
4. Server rejects. Client `onError` fires with a token verification error.

## Root cause

`serverpod_auth_idp_server/lib/src/providers/firebase/business/firebase_idp_config.dart`:

```dart
static void validateFirebaseAccountDetails(
  final FirebaseAccountDetails accountDetails,
) {
  if (accountDetails.email != null && accountDetails.verifiedEmail != true) {
    throw FirebaseUserInfoMissingDataException();
  }
}
```

## Compounding issues

1. **Misleading exception name.** `FirebaseUserInfoMissingDataException` is thrown when the email is *present but unverified*, not when data is missing.
2. **Exception swallowing.** `_parseAccountDetails` rewraps any exception from the validation callback into `FirebaseUserInfoMissingDataException`, then `fetchAccountDetails` rewraps *that* into `FirebaseIdTokenVerificationException`. The client cannot distinguish "bad token" from "unverified email" from "invalid data".
3. **Discoverability.** Default behavior is not surfaced in the dartdoc on `FirebaseIdpConfigFromPasswords`. New users hitting this on their first signup have no signal in the error message about why it failed.

## Proposed direction

One of:

**a)** Flip the default to allow unverified email, since this matches the typical Firebase UX. `firebaseAccountDetailsValidation` becomes opt-in for stricter teams.

**b)** Keep the strict default but:
- Rename to `FirebaseEmailNotVerifiedException` (or similar) and pass it through without rewrapping so the client can react meaningfully.
- Add an explicit `requireVerifiedEmail` flag on `FirebaseIdpConfig` (default `true`) so the behavior is discoverable from the API surface.
- Document that the default blocks signup until verification on the Customizations docs page.

## Notes

- Phone-only auth is already handled correctly by the `email != null` guard.
- Current workaround: override `firebaseAccountDetailsValidation` with a no-op, but you have to read the source to discover this is the issue.

FXschwartz:
The Firebase identity provider's default `firebaseAccountDetailsValidation` rejected any sign-in where an email was present but not verified. Firebase Email/Password accounts start out with `email_verified: false` and users are typically signed in immediately after signup, before they have clicked the verification link, so the default configuration blocked the very first `login()` call of every new user. 

Additionally, the rejection surfaced as a generic `FirebaseIdTokenVerificationException`, giving no hint that email verification was the cause.

- **The default validation now accepts all account details**, including unverified emails. Account linking is keyed on the Firebase UID, so accepting unverified emails does not allow linking into existing accounts.
- **Strict validation is now opt-in** via the new static `FirebaseIdpConfig.requireVerifiedEmail`, which can be passed as `firebaseAccountDetailsValidation`.
- **A new serializable `FirebaseEmailNotVerifiedException`** is thrown by the strict validation.
- **Validation exceptions implementing `SerializableException` are no longer swallowed.** Previously any exception from the validation callback was rewrapped twice, ending up as a generic `FirebaseIdTokenVerificationException`.

Fixes #5154 

## Pre-launch Checklist

## Breaking changes
- The default `firebaseAccountDetailsValidation` on `FirebaseIdpConfig` no longer rejects unverified emails. Projects relying on the previous strict behavior must opt back in with `firebaseAccountDetailsValidation: FirebaseIdpConfig.requireVerifiedEmail`, which now throws the new `FirebaseEmailNotVerifiedException` (passed through to the client) instead of an exception rewrapped into `FirebaseIdTokenVerificationException`.

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/firebase/firebase_idp_config_test.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/firebase/firebase_idp_config_test.dart
@@ -6,9 +6,13 @@ void main() {
     late FirebaseAccountDetails accountDetails;
 
     setUp(() {
-      accountDetails = _createAccountDetails(
+      accountDetails = (
+        userIdentifier: 'firebase-uid',
         email: 'test@example.com',
+        fullName: null,
+        image: null,
         verifiedEmail: false,
+        phone: null,
       );
     });
 
@@ -25,8 +29,8 @@ void main() {
     );
 
     test(
-      'when applying the requireVerifiedEmail validation then it throws a '
-      'FirebaseEmailNotVerifiedException',
+      'when applying the requireVerifiedEmail validation '
+      'then it throws a FirebaseEmailNotVerifiedException',
       () {
         expect(
           () => FirebaseIdpConfig.requireVerifiedEmail(accountDetails),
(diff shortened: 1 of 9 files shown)


modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/firebase/firebase_idp_config_test.dart
```diff
@@ -0,0 +1,153 @@
...
+            accountDetails,
+          ),
+          returnsNormally,
+        );
+      },
+    );
+
+    test(
+      'when applying the requireVerifiedEmail validation then it throws a '
+      'FirebaseEmailNotVerifiedException',
```
marcelomendoncasoares:
```suggestion
      'when applying the requireVerifiedEmail validation '
      'then it throws a FirebaseEmailNotVerifiedException',
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/firebase/firebase_idp_config_test.dart
```diff
@@ -0,0 +1,153 @@
...
+        expect(
+          () => FirebaseIdpConfig.requireVerifiedEmail(accountDetails),
+          throwsA(isA<FirebaseEmailNotVerifiedException>()),
+        );
+      },
+    );
+  });
+
+  group('Given account details with an email whose verification status is '
+      'unknown', () {
```
marcelomendoncasoares:
```suggestion
  group(
    'Given account details with an email whose verification status is unknown', () {
```

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/firebase/firebase_idp_config_test.dart
```diff
@@ -0,0 +1,153 @@
...
+}) {
+  return (
+    userIdentifier: 'firebase-uid',
+    email: email,
+    fullName: null,
+    image: null,
+    verifiedEmail: verifiedEmail,
+    phone: phone,
+  );
+}
```
marcelomendoncasoares: Suggestion: Remove this abstraction in favor of creating the `FirebaseAccountDetails` inline on the "Given" groups. This is preferrable because the abstraction hides the central aspect of the test condition: the account details itself. Users then need to jump to the bottom to understand how the abstraction relates to the "Given" instead of reading directly.

marcelomendoncasoares (changes requested):
Just some small comments on the tests.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/firebase/firebase_idp_config_test.dart
FXschwartz: Done!

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/firebase/firebase_idp_config_test.dart
FXschwartz: Done!

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_server/test/firebase/firebase_idp_config_test.dart
FXschwartz: Removed!

marcelomendoncasoares (approved):
LGTM!

PR #5439, opened 2026-07-07
feat: Allow table models on shared packages

### LINKED ISSUE #5400: Support `table` on models of shared packages
### Problem to Solve

Currently, models on shared packages are not allowed to have the `table` keyword. The reason is because shared models were introduced before client-side databases and tables were only possible on the server. Now that tables can exist on both server and client with `databases: all`, it make sense to allow a shared table model.

### Proposal

Allow users to declare shared models with the `table` keyword and `databases: all`. If the `databases` keyword is omitted, an error should be collected to ensure that the shared model is intentional, because its existence will generate repository classes and migrations on the client.

### Use Case

Create tables that are common to the server and the client without having to duplicate the models. This is specially useful for synced tables with shared methods.

### Alternatives

Have two separate classes for the server and client.

### Additional context

_No response_

- [x] I want to work on this

marcelomendoncasoares:
Closes: #5400

To implement support for the `table` keyword on models of shared packages, this PR makes all client and shared models also implement the `ProtocolSerialization` class - previously only used on the server. This interface exposes the `toJsonForProtocol` and is responsible by filtering `serverOnly` fields or hidden FK fields. By doing this, we:

1. Enable adding `clientOnly` in the future for symmetry.
2. Fix hidden FK fields on client-side relations leaking into the wire (existing bug and requirement for the change).
3. Simplify the code generation, reducing branching between server and client.

This PR also broaden coverage on the tests for the interaction between shared packages and modules with the client-side. Migrations are only generated on the client has at least one host table, which prevents a shared package with a table to generate client-side migrations and force clients to import the `serverpod_database`.

### New SQLite shared test package

To build the integration tests, this PR introduces a shared package for the SQLite test project. Such project contains the same tables as the `serverpod_test_shared`, but is tied only to the SQLite test server. The reason to have a new instead of reusing the `serverpod_test_shared` is that [we can't have a shared package be owned by two server packages](https://github.com/serverpod/serverpod/issues/5446).

### On the legacy auth changes

Due to adding the `serverpod_database` package as a dependency on the `serverpod_test_shared`, we generated a conflict between the `win32` dependency and legacy `file_picker` from `serverpod_auth_test_flutter`. The conflict required bumping `file_picker` and solving the API changes.

## Pre-launch Checklist

## Breaking changes

None.


tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -106,9 +106,15 @@ class LibraryGenerator {
     var nonModelStreamTypes = protocolDefinition
         .getNonModelOrPrimitiveStreamTypes(modules: config.modules);
 
+    final protocolDatabaseSerializationManager =
```
Crazelu: I was surprised to see that this variable is for a boolean.
I suggest renaming

tools/serverpod_cli/test/generator/dart/shared_code_generator/class_field_test.dart
```diff
@@ -186,55 +190,163 @@ void main() {
...
+            initializerMethod: '${testClassName}Table',
+          ),
+          isTrue,
+          reason: 'Missing declaration for ${testClassName}Table singleton.',
+        );
+      });
+
+      test('has a table getter.', () {
+        expect(
+          CompilationUnitHelpers.hasMethodDeclaration(
```
Crazelu, line 274: Suggestion:
Add an optional `isGetter` parameter to `CompilationUnitHelpers.hasMethodDeclaration` to tighten this test.
`table()` would pass, for instance.

Crazelu (approved):
Apart from two minor suggestions, this looks good.

tools/serverpod_cli/test/generator/dart/shared_code_generator/class_field_test.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5439/commits/006731f1a1a31c663c113458b16cef77fd729454.

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
marcelomendoncasoares: Good call. Done at https://github.com/serverpod/serverpod/pull/5439/commits/4ba0455dfc435f5e114c127a2651eba09bdb09f4.

PR #5444, opened 2026-07-08
feat: Add support for cache busting with dedicated syntax in templating system

### LINKED ISSUE #5378: Make the cache busting pre-configured and with nicer syntax in templating system
It's currently pretty hard to use cache busting, and it requires a custom setup.

Let's move from the `package:mustache_template` to `package:whiskers`. Whiskers provides an `onMissingVariable` method that we can use to cache bust parts using the following syntax:

```html
<img src="{{!/foo/bar.png}}" />
```

When doing this, we should also add a `StaticRoute.withCacheBusting` constructor, which sets up a `StaticRoute.directory` with default settings taken from a `CacheBustingConfig`. 

E.g., we can replace:

```dart
final buster = CacheBustingConfig(
  mountPrefix: '/',
  fileSystemRoot: Directory(
    'web/static',
  ),
);


final root = Directory(Uri(path: 'web/static').toFilePath());
pod.webServer.addRoute(
  StaticRoute.directory(
    root,
    cacheBustingConfig: buster,
    cacheControlFactory: (ctx, fileInfo) {
      return CacheControlHeader(
        maxAge: 31536000, // 1 year - safe with cache busting
        publicCache: true,
        immutable: true,
      );
    },
  ),
);
```

with

```dart
pod.webServer.addRoute(
  StaticRoute.withCacheBusting(
    mountPrefix: '/',
    fileSystemRoot: 'web/static',
  ),
)
```

Cache busting should be configured by default in the template. Preferably, no more code than what's listed above.

Crazelu:
This PR adds a dedicated syntax for cache busting asset paths: `{{@/path/to/asset}}`. The proposed `{{!/path/to/asset}}` does not work because `!` is used to mark a comment in mustache templates which are rendered as empty strings. 

The serverpod package now depends on `whiskers` instead of `mustache_template` to use the `onMissingVariable` callback support when rendering.

`StaticRoute.withCacheBusting` factory constructor has been introduced to make it simpler to handle cache busting setup.
The template for new projects now comes with cache busting pre-configured.

Other changes: 

- Templates are created with `lenient` set to true to allow for rendering with missing variables.
- Templates are created with `htmlEscapeValues` set to to false to prevent escaping special characters such as `/` in file paths.

Closes #5378 

## Pre-launch Checklist

## Breaking changes

- `WebWidget` gains a `render` method with `onMissingVariable` callback to support cache busting using introduced mustache syntax.

### DIFF
--- a/packages/serverpod/lib/src/web_server/templates.dart
+++ b/packages/serverpod/lib/src/web_server/templates.dart
@@ -1,7 +1,7 @@
 import 'dart:io';
 
-import 'package:mustache_template/mustache.dart';
 import 'package:path/path.dart';
+import 'package:whiskers/whiskers.dart';
 
 /// Global access to all templates loaded when starting the webserver.
 final Templates templates = Templates();
@@ -37,6 +37,8 @@ class Templates {
         _templates[templateKey] = Template(
           data,
           name: templateKey,
+          lenient: true,
+          htmlEscapeValues: false,
         );
       } else if (entity is Directory) {
         var subDirName = basename(entity.path);
--- a/packages/serverpod/lib/src/web_server/templates.dart
+++ b/packages/serverpod/lib/src/web_server/templates.dart
@@ -38,7 +38,6 @@ class Templates {
           data,
           name: templateKey,
           lenient: true,
-          htmlEscapeValues: false,
         );
       } else if (entity is Directory) {
         var subDirName = basename(entity.path);
--- a/packages/serverpod/lib/src/web_server/web_server.dart
+++ b/packages/serverpod/lib/src/web_server/web_server.dart
@@ -455,8 +455,11 @@ abstract class WidgetRoute extends Route {
   /// to a specific virtual host (defaults to `null`, matching any host).
   WidgetRoute({super.methods, super.path, super.host, this.cacheBustingConfig});
 
-  /// Used to cache-bust paths for {{@/path/to/asset}} patterns in the template.
+  /// Used to cache-bust paths for `{{{@/path/to/asset}}}` patterns in the template.
   final CacheBustingConfig? cacheBustingConfig;
+
+  /// Matches `@/path/to/asset` patterns in the template.
+  final _cacheBustingPathRegExp = RegExp(r'@/[^}]+');
 
   /// Override this method to build your web widget from the current [session]
   /// and [request].
@@ -487,13 +490,22 @@ abstract class WidgetRoute extends Route {
       ),
     );
 
-    // Cache-bust paths for {{@/path/to/asset}} patterns in the template.
+    // Cache-bust paths for {{{@/path/to/asset}}} patterns in the template.
     // This warms up CacheBustingConfig's cache so that the
     // cache-busted paths are available synchronously when rendering.
     if (cacheBustingConfig case final buster?) {
-      final regex = RegExp(r'\{\{@(/[^}]+)\}\}');
+      final regex = RegExp(r'\{\{\{@(/[^}]+)\}\}\}');
       final matches = regex
-          .allMatches(widget.render(onMissingVariable: (name) => '{{$name}}'))
+          .allMatches(
+            widget.render(
+              onMissingVariable: (name) {
+                if (_cacheBustingPathRegExp.hasMatch(name)) {
+                  return '{{{$name}}}';
+                }
+                return '{{$name}}';
+              },
+            ),
+          )
           .toList();
       for (final match in matches) {
         await buster.tryAssetPath(match.group(1)!);
@@ -503,10 +515,11 @@ abstract class WidgetRoute extends Route {
     final renderedContent = widget.render(
       onMissingVariable: (name) {
         if (cacheBustingConfig case final buster?) {
-          // Strip leading '@' from the name
-          // which is used to indicate that the asset should be cache-busted.
-          if (name.startsWith('@')) name = name.substring(1);
-          return buster.tryAssetPathSync(name);
+          if (_cacheBustingPathRegExp.hasMatch(name)) {
+            // Strip leading '@' from the path
+            // which is used to indicate that the asset should be cache-busted.
+            return buster.tryAssetPathSync(name.substring(1));
+          }
         }
         return null;
       },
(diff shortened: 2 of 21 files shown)


templates/serverpod_templates/projectname_server_upgrade/lib/server.dart
```diff
@@ -56,7 +56,10 @@ void run(List<String> args) async {
   // Serve all files in the web/static relative directory under /web.
   // These are used by the default web page.
   final root = Directory(Uri(path: 'web/static').toFilePath());
-  pod.webServer.addRoute(StaticRoute.directory(root), '/web');
+  pod.webServer.addRoute(
+    StaticRoute.withCacheBusting(root, mountPrefix: '/web'),
```
vlidholt: What I do not like here is that we are duplicating the information in the `cacheBustingConfig`. If we need to keep it in a separate file, this method should be something like:

`StaticRoute.withCacheBusting(cacheBustingConfig)`

We may need to add some property to the config, possibly.

vlidholt: Also, make sure we are consistent in the interface if we are using `Directory` or a string.

Crazelu: Done in [dd382fb](https://github.com/serverpod/serverpod/pull/5444/commits/dd382fbede36c74fb656f0dbdcda4cc9fdfb1ebd)

templates/serverpod_templates/projectname_server_upgrade/lib/server.dart
```diff
@@ -55,8 +58,10 @@ void run(List<String> args) async {
   // {{#webserver}}
   // Serve all files in the web/static relative directory under /web.
   // These are used by the default web page.
-  final root = Directory(Uri(path: 'web/static').toFilePath());
-  pod.webServer.addRoute(StaticRoute.directory(root), '/web');
+  pod.webServer.addRoute(
+    StaticRoute.withCacheBusting(cacheBustingConfig),
+    '/web',
```
vlidholt: Should this be taken from `cacheBustingConfig.mountPrefix`?

Crazelu: I don't think so. 
That would require changing the web server interface to branch on static routes that have cache busting configured for `addRoute`.

Crazelu: I had another look at this and I realize you mean replacing `/web` with `cacheBustingConfig.mountPrefix` which is accurate.

Done in [14ad37d](https://github.com/serverpod/serverpod/pull/5444/commits/14ad37df05822f7c04e69bd75f34535977bc6d55)

packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -484,8 +487,33 @@ abstract class WidgetRoute extends Route {
...
+      final regex = RegExp(r'\{\{@(/[^}]+)\}\}');
+      final matches = regex
+          .allMatches(widget.render(onMissingVariable: (name) => '{{$name}}'))
+          .toList();
+      for (final match in matches) {
+        await buster.tryAssetPath(match.group(1)!);
+      }
+    }
+
+    final renderedContent = widget.render(
```
FXschwartz, line 515: I think there's a subtle bug in this callback that goes against your new empty string test checks. `onMissingVariable` is called for every unresolved variable, and `tryAssetPathSync` returns its input unchanged when the path isn't in the cache. 

So on a route with cache busting configured, a plain missing variable like {{userName}} renders as the literal text "userName" instead of the empty string you'd get on a route without the config.

I think we should do something like 
```dart
if (name.startsWith('@')) {
  return buster.tryAssetPathSync(name.substring(1));
}
```

packages/serverpod/lib/src/web_server/templates.dart
```diff
@@ -37,6 +37,8 @@ class Templates {
         _templates[templateKey] = Template(
           data,
           name: templateKey,
+          lenient: true,
+          htmlEscapeValues: false,
```
FXschwartz: I'm a bit worried about `htmlEscapeValues: false` here. I'm guessing this was intentional and probably already discussed but it disabled for both cache busting tags and every `{{var}}` in every existing user template. So apps rendering user provided data through {{var}} lose their XSS protection silently on upgrade.

Could the templates use triple stache tags instead (`{{{@/path/to/asset}}}`)? Those skip escaping per tag, so escaping could stay defaulted to on here and the warmup regex just needs to match the extra brace.

packages/serverpod/lib/src/web_server/web_server.dart
```diff
@@ -484,8 +487,33 @@ abstract class WidgetRoute extends Route {
       ),
     );
 
+    // Cache-bust paths for {{@/path/to/asset}} patterns in the template.
+    // This warms up CacheBustingConfig's cache so that the
+    // cache-busted paths are available synchronously when rendering.
+    if (cacheBustingConfig case final buster?) {
```
FXschwartz, line 496: Does this render the widget twice then? Once to find the `{{@...}}` tags and warm the cache, then again for the response?

If so could we use `CacheBustingConfig.indexAssets()` instead? Calling it once lazily would pre populate the cache for everything under the root, so this whole warm up block could goe away and the single render resolves every @ tag through `tryAssetPathSync`.

FXschwartz (changes requested):

packages/serverpod/lib/src/web_server/web_server.dart
Crazelu: > Does this render the widget twice then? Once to find the {{@...}} tags and warm the cache, then again for the response?

Yes.

> If so could we use CacheBustingConfig.indexAssets() instead? 

We could however that is more expensive because it indexes every asset in the file system. The current approach makes sure that only the assets referenced in the template are warmed in cache. Calling `CacheBustingConfig.indexAssets()` lazily won't work because there's no guarantee that all of the assets will be indexed before the synchronous rendering.

packages/serverpod/lib/src/web_server/templates.dart
Crazelu: Resolved in [79505dd](https://github.com/serverpod/serverpod/pull/5444/commits/79505dd4f50d935c485113c134c8514ac57441ae)

packages/serverpod/lib/src/web_server/web_server.dart
Crazelu: Resolved in [79505dd](https://github.com/serverpod/serverpod/pull/5444/commits/79505dd4f50d935c485113c134c8514ac57441ae)

FXschwartz (approved):
LGTM! 🚀

PR #5450, opened 2026-07-09
fix: Fix the quick close-open streaming method bug

### LINKED ISSUE #5449: Opening a second stream in quick succession races with websocket teardown
### Describe the bug

Underneath streaming endpoints, `ClientMethodStreamManager` shares one websocket across all method streams and closes it when the last stream ends. That teardown is fired without synchronizing against new `openMethodStream` calls, so a client that closes one method stream and promptly opens another can crash with an undocumented `StateError`, or have the new stream silently killed.

### To reproduce

Any usage pattern with short-lived, back-to-back method streams (e.g. an RPC-style call that opens a stream, completes, and is immediately called again) can hit it. Long-lived streams rarely exercise the window, which is probably why it has never happened until now.

### Race 1: teardown fires while a new stream is being opened

`_tryCloseConnection` calls `closeAllConnections()` unawaited when both stream maps are empty:

```dart
void _tryCloseConnection() {
  if (_inboundStreams.isEmpty && _outboundStreams.isEmpty) {
    closeAllConnections();
  }
}
```

Sequence:

1. The server closes the last method stream; `_dispatchCloseMethodStreamCommand` empties the stream maps and `_tryCloseConnection` schedules `closeAllConnections()` (unawaited).
2. The application immediately calls `openMethodStream`. `_openMethodStream` checks `if (_webSocket == null)` — still non-null because the teardown has not run its lock body yet — so it skips `_connectSynchronized`.
3. `_openMethodStream` suspends at `await connectionDetails.authKeyProvider?.authHeaderValue`, yielding to the event loop; the scheduled teardown runs and sets `_webSocket = null`.
4. `_addMessageToWebSocket(openCommand)` throws:

```console
Bad state: Message posted when web socket connection is closed: {"type":"omsc","data":{...,"args":"{...}","is":["..."],"auth":"Bearer token"}}
  package:serverpod_client/src/client_method_stream_manager.dart 606:7   ClientMethodStreamManager._addMessageToWebSocket
  package:serverpod_client/src/client_method_stream_manager.dart 159:5   ClientMethodStreamManager._openMethodStream
  ===== asynchronous gap ===========================
  package:serverpod_client/src/client_method_stream_manager.dart 110:14  ClientMethodStreamManager.openMethodStream
```

Note that the payload in the error is the *open* command — the new stream's open raced the previous connection's teardown. `openMethodStream`'s documented failure modes are `OpenMethodStreamException`, `WebSocketConnectException`, and `ConnectionAttemptTimedOutException`; a raw `StateError` from an internal helper is not part of the contract and callers cannot handle it meaningfully. Depending on the interleaving, the open command can instead be registered into `_inboundStreams` just before teardown clears the map, in which case `openMethodStream` never completes (the open completer is never resolved).

### Race 2: a dead connection's listener cleanup destroys its replacement

Method `_listenToWebSocketStream`'s `finally` block calls `closeAllConnections(closeException)` unconditionally:

```dart
} finally {
  _cancelConnectionTimer();
  ...
  await closeAllConnections(closeException);
}
```

The call to `closeAllConnections` operates on the manager's *current* state, not on the connection whose listener just died. If a new websocket was established while the old listener was still winding down (its `closeAllConnections` is queued on the lock behind `_connectSynchronized`), the old connection's cleanup then nulls `_webSocket` (which now points at the **new** socket), cancels the new connection's ping/idle timers, closes the new socket, and errors every freshly registered stream with `WebSocketClosedException`. From the caller's perspective, a just-opened stream dies immediately, or the open command hits the nulled socket and throws the same `StateError` as above.

### Expected behavior

A new streaming method connection should not be closed due to previous tearing-down/errored connections.

### Library version

4.0.0-beta.0

### Platform information

Any

### Additional context

The fix consists in two combined changes in `ClientMethodStreamManager` that won't introduce behavior changes to the documented contract ("the connection will be closed when all streams are done"):

1. **Make the close-when-unused decision atomic**: Replace the unawaited `closeAllConnections()` in `_tryCloseConnection` with a teardown that runs under the existing connection `Lock` and re-checks that the stream maps are still empty after acquiring it. A stream opened while the teardown waited for the lock keeps the connection open (consistent with how concurrent streams already share the socket).

2. **Make listener cleanup generation-aware**: The `finally` cleanup receives the socket it belongs to and, under the lock, returns without touching manager state when `_webSocket` already points at a different (new) connection. The dead connection's own streams were already removed when it closed; the new connection's timers, socket, and streams are left intact.

The public API `closeAllConnections()` (and its use for idle timeouts) is unchanged and will still unconditionally close everything.

- [x] I want to work on this

marcelomendoncasoares:
Fixes: #5449

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod_client/lib/src/client_method_stream_manager.dart
+++ b/packages/serverpod_client/lib/src/client_method_stream_manager.dart
@@ -194,8 +194,24 @@ final class ClientMethodStreamManager {
 
   void _tryCloseConnection() {
     if (_inboundStreams.isEmpty && _outboundStreams.isEmpty) {
-      closeAllConnections();
+      unawaited(_closeConnectionIfUnused());
     }
+  }
+
+  /// Closes the websocket connection if no streams are using it.
+  ///
+  /// The emptiness check is repeated once the lock is acquired, since a new
+  /// method stream may have been opened while this cleanup was waiting for
+  /// the lock. In that case the connection must be kept open.
+  Future<void> _closeConnectionIfUnused() async {
+    await _lock.synchronized(() async {
+      if (_inboundStreams.isNotEmpty || _outboundStreams.isNotEmpty) return;
+      _cancelConnectionTimer();
+      var webSocket = _webSocket;
+      _webSocket = null;
+      await webSocket?.sink.close();
+      await _webSocketListenerCompleter.future;
+    });
   }
 
   /// Builds a unique key for a stream.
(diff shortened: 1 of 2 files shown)


Crazelu (approved):
🚀

PR #5459, opened 2026-07-10
fix: Mark Flutter app tabs ready on non-web devices

### LINKED ISSUE #5454: Flutter app tab never leaves loading state on non-web devices when running serverpod start
When `serverpod start` launches a Flutter app on a non-web device the app's tab in the TUI never transitions to the running state, even after the app is fully built and running on the device.

- The status-line text keeps shimmering forever and never updates once the app is up.
- The loading spinner next to the app name (tab strip / launch panel) freezes mid-frame.
- The tab indicator never becomes the green "running" dot, and the status line never shows "App running".

FXschwartz:
When `serverpod start` launches a Flutter app on a non-web device (desktop or mobile), the app's tab in the TUI never left the loading state: the spinner froze, the status text kept shimmering forever.

The ready signal (`onReady`, which sets `tab.ready`) only fired when the daemon published a web URL. `flutterAppUrl` comes from the `app.webLaunchUrl` machine event, which only exists for web devices, so on any other device the tab stayed "launching" forever.

This PR signals ready from the daemon's `app.started` event, which fires on every device type:
- `FlutterProcess` gains an `onStarted` callback, invoked on `app.started`.
- `FlutterAppManager` funnels both ready sources through a `_signalReady` helper that fires `onReady` at most once per launch. 
- `isLaunching` is now keyed on the ready signal instead of URL publication.
- New `emits_desktop_machine_events.dart` shim with a manager test asserting `onReady` fires once with a null URL, plus a `FlutterProcess` unit test for `onStarted`.

Fixes #5454 

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -376,7 +376,7 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
   IOSink Function(FlutterAppConfig app)? flutterStderrSinkFor,
   void Function(FlutterAppConfig app)? onEnsureFlutterAppTab,
   void Function(FlutterAppConfig app, String stage)? onFlutterProgress,
-  void Function(FlutterAppConfig app, String url)? onFlutterReady,
+  void Function(FlutterAppConfig app, String? url)? onFlutterReady,
   void Function(FlutterAppConfig app)? onFlutterLaunchFailed,
   void Function(FlutterAppConfig app)? onFlutterStop,
   Future<void> Function(ServerProcess server)? onServerStart,
@@ -1142,7 +1142,9 @@ Future<void> _runTuiBackend({
       onFlutterReady: (app, url) {
         final tab = holder.state.appLogTabFor(app.id);
         if (tab != null) {
-          tab.url = url;
+          // Non-web devices publish no URL; the status line then falls back
+          // to its generic running label.
+          tab.url = url ?? tab.url;
           tab.ready = true;
           tab.stopped = false;
           holder.markDirty();
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -1142,9 +1142,9 @@ Future<void> _runTuiBackend({
       onFlutterReady: (app, url) {
         final tab = holder.state.appLogTabFor(app.id);
         if (tab != null) {
-          // Non-web devices publish no URL; the status line then falls back
-          // to its generic running label.
-          tab.url = url ?? tab.url;
+          // Null on non-web devices, which publish no URL; the status line
+          // then falls back to its generic running label.
+          tab.url = url;
           tab.ready = true;
           tab.stopped = false;
           holder.markDirty();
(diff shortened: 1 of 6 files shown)


tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
```diff
@@ -550,6 +557,7 @@ class FlutterProcess {
           }
         case 'app.started':
           _onProgress?.call('ready');
```
Crazelu: Should we remove this now that we have a dedicated `onStarted`?

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -1142,7 +1142,9 @@ Future<void> _runTuiBackend({
       onFlutterReady: (app, url) {
         final tab = holder.state.appLogTabFor(app.id);
         if (tab != null) {
-          tab.url = url;
+          // Non-web devices publish no URL; the status line then falls back
+          // to its generic running label.
+          tab.url = url ?? tab.url;
           tab.ready = true;
           tab.stopped = false;
           holder.markDirty();
```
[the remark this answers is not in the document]
FXschwartz: Done

tools/serverpod_cli/lib/src/commands/start/flutter_app_manager.dart
```diff
@@ -242,6 +248,9 @@ class FlutterAppManager {
         onProgress(runtime.app, stage);
         log.info('  ${runtime.app.name}: $stage');
       },
+      // Non-web devices never publish an `app.webLaunchUrl`, so `app.started`
+      // is their only launch-complete signal.
+      onStarted: () => _signalReady(runtime, process.flutterAppUrl),
```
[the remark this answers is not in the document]
FXschwartz: Done

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
FXschwartz: Yeah you are right, good simplification. Removed and `onStarted` is the only signal for `app.started` now.

tools/serverpod_cli/lib/src/commands/start/flutter_app_manager.dart
```diff
@@ -242,6 +248,13 @@ class FlutterAppManager {
         onProgress(runtime.app, stage);
         log.info('  ${runtime.app.name}: $stage');
       },
+      // Only non-web devices ready on `app.started`; web-server devices ready
+      // on URL publication, which `app.started` must never preempt.
+      onStarted:
```
Crazelu: I didn't test this but my hunch is that the previous `ready` signal on `app.started` always happened after URL publication. If that's the case, should there be a distinction here about the device type? What if it is a different web device like chrome?

FXschwartz: Yeah I think you're right. I tested this against a live `flutter run -d web-server --machine` and `app.webLaunchUrl` comes in right before `app.started`, about 1ms apart in the same event.

Just to note that the old code never actually signaled ready on `app.started`, that event only updated the progress text. Ready only ever came from the URL publication, which is why non-web devices got stuck forever.

I've removed the device check and `onStarted` is wired unconditionally now! Since it passes the currently known `flutterAppUrl` and `_signalReady` dedupes, web devices still get their URL either way.

Crazelu (approved):
🚀

PR #5462, opened 2026-07-10
fix(cli): validate relation database actions

### LINKED ISSUE #5401: Using `onDelete=SetNull` on relations should require a nullable FK column
### Describe the bug

The Serverpod model analyzer does not validate that `onDelete=SetNull` (or `onUpdate=SetNull`) is compatible with the nullability of the foreign key column.

When a relation uses a non-optional object relation or an explicit non-nullable FK field with `SetNull`, `serverpod generate` succeeds and migrations are generated with `ON DELETE SET NULL` on a `NOT NULL` column. PostgreSQL rejects that constraint at migration time. The failure is deferred until migration/DDL instead of being caught during model validation.

### To reproduce

1. Create a model with a non-nullable explicit FK field and `onDelete=SetNull`:

```yaml
class: Parent
table: parent
fields:
  name: String
```
```yaml
class: Child
table: child
fields:
  parentId: int
  parent: Parent?, relation(field=parentId, onDelete=SetNull)
```

2. Run `serverpod generate`.
3. Observe that generation completes with no analyzer errors.
4. Run `serverpod create-migration` (or apply the migration to PostgreSQL).
5. Observe that PostgreSQL rejects the foreign key constraint because `ON DELETE SET NULL` is not allowed on a `NOT NULL` column.

For comparison, existing test models use nullable FK columns with `SetNull`, e.g. `arenaId: int?` in `tests/serverpod_test_server/lib/src/models/models_with_relations/nested_one_to_many/team.spy.yaml`.

### Expected behavior

The analyzer should report a validation error when either `onDelete=SetNull` or `onUpdate=SetNull` is used on a relation whose FK column is non-nullable.

Example error message:

> The field "parentId" must be nullable when using `onDelete=SetNull` (or `onUpdate=SetNull`).

This would match PostgreSQL semantics and fail fast at model definition time instead of at migration time.

### Library version

3.5.0-beta.10, but exists since a long time.

### Platform information

N/A

### Additional context

The same validation gap also applies to `onDelete=SetDefault` and `onUpdate=SetDefault`, but with different database rules than `SetNull`. PostgreSQL requires the FK column to have a database-level default; it does not require the column to be nullable. Serverpod still does not validate this — it accepts `SetDefault` whenever the enum value is valid and `onDelete` is on the FK-holding side, then emits `ON DELETE SET DEFAULT` regardless of whether the column actually has a SQL default.

That leads to invalid combinations passing analysis and failing at migration time, such as a non-nullable FK with no persist default (`parentId: int`) or with only a model default (`parentId: int, defaultModel=0`), since `_parseColumnDefault()` only uses `defaultPersistValue` and `defaultModel` alone does not produce a database default. A complete fix should treat `SetNull` and `SetDefault` as separate rules: `SetNull` requires a nullable FK column; `SetDefault` requires a database default on the FK column (`default` or `defaultPersist`, not `defaultModel` alone).

- [ ] I want to work on this

eukalpia:
Validates relation database actions against the resolved foreign-key column before migrations are generated.

For `SetNull`, the foreign-key database column must be nullable. For `SetDefault`, it must have a database-level default from `default` or `defaultPersist`; `defaultModel` alone is not sufficient because it only initializes Dart objects.

The validation resolves both explicit and implicitly generated foreign-key fields and covers `onDelete` and `onUpdate`. Database defaults are supported on direct foreign-key id relations, while remaining invalid on non-persisted object and list relation fields.

Fixes #5401

## Validation

- `dart format --output=none --set-exit-if-changed` on the changed Dart files
- `dart analyze` in `tools/serverpod_cli`
- `dart test test/analyzer/models/stateful_analyzer/model_validation/relation/relation_database_action_test.dart`
- `dart test test/analyzer/models/stateful_analyzer/model_validation/relation`
- `git diff --check`

## Pre-launch Checklist

- [x] I read the contribution guidelines and kept the change focused on one bug.
- [x] The code follows Dart formatting and style conventions.
- [x] The PR references the issue it fixes.
- [x] Relevant regression tests were added.
- [x] Existing and new relation analyzer tests pass.
- [x] No breaking API changes are introduced.

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/CHANGELOG.md
+++ b/tools/serverpod_cli/CHANGELOG.md
@@ -6,6 +6,7 @@
 - feat: Adds support for creating server only projects.
 - fix: Fixes `upsert` with `updateWhere` throwing on SQLite when conflicts are filtered out.
 - fix: Adds missing export of `DeepCollectionEquality` for shared models. Backported to 3.4.11.
+- fix: Validates `SetNull` and `SetDefault` relation actions against foreign key column constraints.
 - fix: Prevents the creation of orphaned images on subsequent IDP logins. Backported to 3.4.11.
 - refactor: Refines the serverpod start TUI app status, stop/close, and launcher.
 - chore: Changes the template to serve the Flutter web app under root if website is not enabled.
--- a/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
@@ -537,9 +537,10 @@ class Restrictions {
     var definition = documentDefinition;
     if (definition is! ClassDefinition) return [];
 
-    var field = definition.findField(parentNodeName);
+    var relationField = definition.findField(parentNodeName);
+    var relation = relationField?.relation;
 
-    if (field?.relation?.isForeignKeyOrigin == false) {
+    if (relation?.isForeignKeyOrigin == false) {
       return [
         SourceSpanSeverityException(
           'The "$key" property can only be set on the side holding the foreign key.',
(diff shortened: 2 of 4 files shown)


eukalpia:

PR #5465, opened 2026-07-13
feat: Show the device platform in the start TUI Flutter app status line

FXschwartz:
For non-web Flutter apps, the `serverpod start` TUI status line previously showed a generic `App running` once the app was up. It now shows which platform the app is running on:

The daemon's `app.start` event carries the resolved `deviceId`:
- Desktop ids name their platform directly (`windows`, `macos`, `linux`)
- Mobile ids are opaque serials (`emulator-5554`, iOS UDIDs), so they are matched against a `device.getDevices` daemon request and mapped via the device's `platformType`.

Resolution is best-effort: on any failure the status line just keeps the generic `Running on device` label. Web devices are not mapped since their status line always shows the published URL.

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -1149,6 +1149,8 @@ Future<void> _runTuiBackend({
         tab.ready = false;
         tab.stopped = false;
         tab.url = null;
+        // Refreshed per launch; the configured device may have changed.
+        tab.device = app.device;
         // Focus the tab only when the launch was initiated from the launch
         // panel (which is open at that point). Apps auto-started by
         // `serverpod start` launch with the panel closed, so the Server logs
--- a/tools/serverpod_cli/test/commands/start/tui/main_screen_layout_test.dart
+++ b/tools/serverpod_cli/test/commands/start/tui/main_screen_layout_test.dart
@@ -313,51 +313,49 @@ void main() {
     );
   });
 
-  group('Given a ready Flutter app tab', () {
-    test(
-      'when it has a configured device '
-      'then the status line names it',
-      () async {
-        final tester = await _pumpReadyAppTab(device: 'emulator-5554');
+  group('Given a ready Flutter app tab with a configured device', () {
+    late NoctermTester tester;
 
-        expect(
-          tester.terminalState.containsText('Running on device emulator-5554'),
-          isTrue,
-        );
-      },
-    );
+    setUp(() async {
+      tester = await _pumpReadyAppTab(device: 'emulator-5554');
+    });
 
-    test(
-      'when it has no configured device '
-      'then the status line shows the generic running label',
-      () async {
-        final tester = await _pumpReadyAppTab();
+    test('then the status line names the device', () {
+      expect(
+        tester.terminalState.containsText('Running on device emulator-5554'),
+        isTrue,
+      );
+    });
+  });
 
-        expect(
-          tester.terminalState.containsText('Running on device'),
-          isTrue,
-        );
-      },
-    );
+  group('Given a ready Flutter app tab without a configured device', () {
+    late NoctermTester tester;
 
-    test(
-      'when a URL is published '
-      'then the status line shows the URL, not the device',
-      () async {
-        final tester = await _pumpReadyAppTab(
-          device: 'chrome',
-          url: 'http://localhost:8080',
-        );
+    setUp(() async {
+      tester = await _pumpReadyAppTab();
+    });
 
-        expect(
-          tester.terminalState.containsText('http://localhost:8080'),
-          isTrue,
-        );
-        expect(
-          tester.terminalState.containsText('Running on device'),
-          isFalse,
-        );
-      },
-    );
+    test('then the status line shows the generic running label', () {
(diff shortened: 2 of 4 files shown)


tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
```diff
@@ -566,6 +582,52 @@ class FlutterProcess {
...
+  };
+
+  /// Labels for the daemon's `platformType` device property.
+  static const _platformLabelsByPlatformType = <String, String>{
+    'android': 'Android',
+    'ios': 'iOS',
+    'macos': 'macOS',
+    'windows': 'Windows',
+    'linux': 'Linux',
+  };
```
marcelomendoncasoares: Question: Why not just capitalize? Seems an unnecessary mapping.

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
```diff
@@ -566,6 +582,52 @@ class FlutterProcess {
...
+          platform = _platformLabelsByPlatformType[platformType];
+        }
+      } catch (e) {
+        log.debug('Could not resolve the platform of device $deviceId: $e');
+      }
+    }
+    if (platform == null) return;
+    _devicePlatform = platform;
+    _onDevicePlatform?.call(platform);
+  }
```
marcelomendoncasoares: Question: Why resolve instead of just getting from the config?

tools/serverpod_cli/lib/src/commands/start.dart
```diff
@@ -590,6 +591,8 @@ Future<WatchLoopSetupResult> _setupWatchLoop({
     serverPackageDirectoryPathParts: config.serverPackageDirectoryPathParts,
     onProgress: (app, stage) => onFlutterProgress?.call(app, stage),
     onReady: (app, url) => onFlutterReady?.call(app, url),
+    onDevicePlatform: (app, platform) =>
+        onFlutterDevicePlatform?.call(app, platform),
```
marcelomendoncasoares: Question: This is also not necessary if we get the name directly from the config, right?

marcelomendoncasoares (commented):
A few questions on the source of the device info.

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
FXschwartz: It's definitely some extra fluff. I did it because iOS and macOS unique brand casings. I'm also using it as a sort of allow list for when `platformType` is a value that we don't want to render raw like an emulator with it's ID.

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
FXschwartz: Because the config devices only handles some cases. If its an emulator or an iOS it could just be a generic "emulator-555" or and iOS udid. `flutter run -d` also accepts device prefixes like "pixel" or "iphone".

The resolve only get's called when we cannot get a valid device name from the config devices.

tools/serverpod_cli/lib/src/commands/start.dart
FXschwartz: Similar to the "why the resolve" question above the label is asynchronous so the call back makes sure the status line is updated whenever resolution is complete.

FXschwartz:
@marcelomendoncasoares You're questions about the resolve does have me thinking. I went the OS name route for the flutter status line. But if we'd rather just display the device name "Running on device emulator-555" we can just go off of the config devices and not have the resolve method.

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
marcelomendoncasoares: > @marcelomendoncasoares You're questions about the resolve does have me thinking. I went the OS name route for the flutter status line. But if we'd rather just display the device name "Running on device emulator-555" we can just go off of the config devices and not have the resolve method.

Let's go this path and simplify. A plain `Running on device "emulator-555"` is better, because it was the user that wrote the device ID and they will know what it means. Trying to beautify it can only mask the info.

tools/serverpod_cli/test/commands/start/tui/main_screen_layout_test.dart
```diff
@@ -291,4 +312,52 @@ void main() {
...
+          tester.terminalState.containsText('http://localhost:8080'),
+          isTrue,
+        );
+        expect(
+          tester.terminalState.containsText('Running on device'),
+          isFalse,
+        );
+      },
+    );
+  });
```
marcelomendoncasoares: Suggestion: This "Given" is being used as generic group instead of targeting the test scenario (the configured device), leading to part of the condition leaking into the "when". Each of these is a separate "Given".

marcelomendoncasoares (changes requested):
Only one comment on the tests.

tools/serverpod_cli/test/commands/start/tui/main_screen_layout_test.dart
FXschwartz: Done!

marcelomendoncasoares (approved):
LGTM!

PR #5466, opened 2026-07-14
fix: Fix watch-mode feedback loops and Flutter build disk I/O

marcelomendoncasoares:
## Summary

Fixes watch-mode feedback loops and excessive disk I/O during Flutter builds, while preserving dependency-change detection across missing, deleted, and recreated pub artifacts.

## Changes

- Excludes generated output paths from generate --watch events before invoking generation, preventing generated files from feeding back into the next analysis pass.

- Avoids touching protocol.dart during unchanged generations. The temporary analysis stub is now applied through an analyzer overlay instead of being written to disk. The existing disk-based approach remains as a fallback, and overlay teardown refreshes the analysis context to prevent stale stub content.

- Regenerates only when the set of files containing analyzer errors changes. Persistent errors no longer cause unrelated file events to trigger generation indefinitely, while introducing or resolving an error still regenerates as expected.

- Watches package_config.json and Flutter package_graph.json as exact files instead of recursively watching .dart_tool. This avoids scanning churning build output such as flutter_build, which caused the disk-I/O spike during cold Flutter desktop builds.

- Makes exact-file watches persistent:
  - Missing artifacts are checked individually every 500 ms.
  - Native file watching takes over once an artifact appears.
  - Deleting and recreating an artifact no longer ends its watch.
  - Expected Flutter graph paths are registered before the first Flutter run.
  - Flutter dependency tracking initializes lazily when the first valid resolution appears.

## Testing

Added regression coverage for:

- Generated events being ignored by generate --watch, with source-change liveness checks.
- Unchanged generation preserving both protocol.dart content and modification time.
- Persistent endpoint and future-call errors not triggering generation for unrelated changes.
- Error appearance and resolution still triggering generation.
- Workspace and non-workspace pub-artifact watch paths.
- Pub artifacts initially being absent.
- Artifact deletion, recreation, and subsequent modification.
- Flutter dependency tracking initialized after the first resolution is created.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/analyzer/dart/endpoints_analyzer.dart
+++ b/tools/serverpod_cli/lib/src/analyzer/dart/endpoints_analyzer.dart
@@ -78,7 +78,7 @@ class EndpointsAnalyzer {
         .where((f) => p.isWithin(absoluteIncludedPaths, p.absolute(f)))
         .toSet();
 
-    final erroredBefore = _erroredFiles;
+    final erroredFilesBefore = _erroredFiles;
     final keysBefore = _fileCache.keys.toSet();
 
     await analyze(
@@ -86,7 +86,7 @@ class EndpointsAnalyzer {
       changedFiles: relevantPaths,
     );
 
-    final erroredAfter = _erroredFiles;
+    final erroredFilesAfter = _erroredFiles;
     final keysAfter = _fileCache.keys.toSet();
 
     // Regenerate when the set of endpoint files changed, or when any file's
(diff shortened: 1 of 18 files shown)


tools/serverpod_cli/lib/src/analyzer/dart/endpoints_analyzer.dart
```diff
@@ -72,21 +78,27 @@ class EndpointsAnalyzer {
         .where((f) => p.isWithin(absoluteIncludedPaths, p.absolute(f)))
         .toSet();
 
-    final errorsBefore = _fileCache.values.any((r) => r.hadErrors);
```
Crazelu, line 75: I think we should keep the `errorsBefore` and `errorsAfter` name as they read more correctly for the collection.

Crazelu (approved):
Looks good
I have a minor comment on some variable names.

tools/serverpod_cli/lib/src/analyzer/dart/endpoints_analyzer.dart
marcelomendoncasoares: Actually, the previous `errorsBefore` and `errorsAfter` were booleans, not a collection. The current variable holds a list of errored files. I improved a bit to make it clearer on https://github.com/serverpod/serverpod/pull/5466/commits/7783ad77ab82c702601750ac6745c6209dcf5c43.

PR #5467, opened 2026-07-14
feat: Add structured Flutter logs to `serverpod start`

### LINKED ISSUE #5169: Logs on the `Flutter logs` tab should also be structured
On the `Flutter logs` tab, we don't offer the nice experience that users have on the `Server logs`.

Image

Ideally, we should also group similar logs (as Flutter run does) and allow filtering - which is particularly useful on Flutter due to the large amount of log messages.

marcelomendoncasoares:
Closes: #5169

This PR started as a fix for missing important Flutter log events, such as rendering errors or non-initialized providers. But, since such messages were structured, the PR evolved into making all Flutter logs structured. Now, all app tabs show the logs with the same format as the Server logs.

Important to mention that, since stdout/stderr are not structured, they are heuristically grouped into structured entries, with an inferred log level based on the channel (stdout -> info, stderr -> error). Grouping is done by coalescing adjacent raw lines that arrive in a very short time interval (~25ms) with no other messages in between.

We acknowledge that this is not perfect, but it provides a greatly improved UX over the raw logs, and the worst side-effect that can happen is a message being bundled on the wrong group.

image

image

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
+++ b/tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
@@ -5,12 +5,14 @@ import 'dart:io';
 import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
 import 'package:serverpod_cli/src/commands/messages.dart';
+import 'package:serverpod_cli/src/commands/start/flutter_log_event.dart';
 import 'package:serverpod_cli/src/commands/start/server_process.dart'
     show vmServiceWsUri;
 import 'package:serverpod_cli/src/util/browser_launcher.dart';
 import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart';
 import 'package:serverpod_cli/src/vendored/flutter_daemon_protocol.dart';
 import 'package:serverpod_cli/src/vm_proxy/proxy.dart';
+import 'package:serverpod_shared/log.dart' show LogLevel;
 import 'package:vm_service/vm_service.dart';
 import 'package:vm_service/vm_service_io.dart';
 
@@ -45,6 +47,7 @@ class FlutterProcess {
   final List<String> _extraArgs;
   final IOSink _stdout;
   final IOSink _stderr;
+  final void Function(FlutterLogEvent event)? _onLog;
 
   /// IDE-facing proxy. When non-null, the process binds upstream on
   /// VM-service connect and unbinds on shutdown so the IDE sees a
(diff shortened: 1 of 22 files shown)


tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
```diff
@@ -559,13 +650,252 @@ class FlutterProcess {
...
+      (line) =>
+          now - line.receivedAtMilliseconds >
+          _rawLogDeduplicationWindow.inMilliseconds,
+    );
+
+    final lines = stripAnsi(event.message)
+        .split('\n')
+        .map(
+          (line) =>
+              line.endsWith('\r') ? line.substring(0, line.length - 1) : line,
```
Crazelu, line 776: Should this be `line.length - 2`?
Looks to me like it will get rid of `r` but not `\`

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
```diff
@@ -77,13 +83,18 @@ class FlutterProcess {
...
   StreamSubscription<Event>? _loggingSub;
   StreamSubscription<Event>? _stdoutEventSub;
   StreamSubscription<Event>? _stderrEventSub;
   Timer? _vmServiceHeartbeat;
+  Timer? _logCoalesceTimer;
+  _PendingFlutterLog? _pendingLog;
+  _Utf8LineDecoder? _vmStdoutDecoder;
+  _Utf8LineDecoder? _vmStderrDecoder;
+  final Stopwatch _logClock = Stopwatch()..start();
+  final List<_RecentRawLogLine> _recentRawLogLines = [];
```
Crazelu: Suggestion: This could be a `BoundedQueueList` to remove the explicit operation later in this file and also define the max lines at the top level.

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
marcelomendoncasoares: Actually, `\r` is a single carriage-return character (U+000D), not the two characters `\` and `r`.

tools/serverpod_cli/lib/src/commands/start/flutter_process.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5467/commits/c68b212cf80cb37d2be6c24e62974bf1ec34ffce.

Crazelu (approved):
A pretty dense one. Looks good!

PR #5470, opened 2026-07-14
feat: Allow booting the embedded PostgreSQL from the CLI

### LINKED ISSUE #5469: Missing an easy way to connect to a local embedded database
The local embedded database starts with Unix transport by default, making it only accessible from clients that can connect using Unix sockets. Moreover, if the path where the database is located is too long, `psql` will reject the path. A fallback to TCP would be possible, but the database is not being created with a password for later TCP access.

We should then:

1. Default to creating the database with the generated password from `passwords.yaml`. It is harmless for Unix sockets, but enables the usage of TCP on a later manual boot over the same database.
2. Expose a command on `serverpod` to start the embedded database and print the connection string. If no port is passed, it will boot the database on a random port and print the complete connection string for local access.

This will allow users to have a clear UX when inspecting what is going on inside the database.

marcelomendoncasoares:
Closes: #5469

After this PR, users will be able to do: 

```console
12:21 $ serverpod database start
Embedded PostgreSQL is ready.
Connection URI: postgres://postgres:d1rxfknXbFWEWjru3H13vB7j7UJu7Q7Q@127.0.0.1:8090/demo_offline
Press Ctrl+C to stop.
```

In the future, we can also introduce more commands under `serverpod database` to manage the embedded PostgreSQL, such as a `--detach` option on `start`, a `status` and `stop` commands.

## Pre-launch Checklist

## Breaking changes

None, but this feature will only work with databases created after this PR.

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/database.dart
+++ b/tools/serverpod_cli/lib/src/commands/database.dart
@@ -1,0 +1,175 @@
+import 'dart:io';
+
+import 'package:ci/ci.dart' as ci;
+import 'package:cli_tools/cli_tools.dart';
+import 'package:config/config.dart';
+import 'package:path/path.dart' as p;
+import 'package:serverpod_cli/src/runner/serverpod_command.dart';
+import 'package:serverpod_cli/src/runner/serverpod_command_runner.dart';
+import 'package:serverpod_cli/src/util/server_directory_finder.dart';
+import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart';
+import 'package:serverpod_embedded_postgres/serverpod_embedded_postgres.dart';
+import 'package:serverpod_shared/serverpod_shared.dart'
+    show PasswordManager, PostgresDatabaseConfig, ServerpodConfig;
+
+/// Database-related development commands.
+class DatabaseCommand extends ServerpodCommand<OptionDefinition> {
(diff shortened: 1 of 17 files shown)


tools/serverpod_cli/lib/src/commands/database.dart
```diff
@@ -0,0 +1,175 @@
...
+      );
+
+      log
+        ..info('Embedded PostgreSQL is ready.')
+        ..info('Connection URI: ${postgres.connectionString}')
+        ..info(
+          Platform.isMacOS ? 'Press ⌃C to stop.' : 'Press Ctrl+C to stop.',
+        );
+
+      while (postgres.isRunning) {
```
Crazelu, line 116: Would it be cleaner to use a Completer here instead of a loop?

Crazelu: On a second look, the Completer suggestion doesn't seem possible here.

Crazelu (approved):

tools/serverpod_cli/lib/src/commands/database.dart
marcelomendoncasoares: Tried it, but it would span a lot more lines of code than these three to loop awaiting. And we would still need a watch loop down the line.

PR #5472, opened 2026-07-16
fix: Fixed VS Code debugger not working

### LINKED ISSUE #5452: Launching the Flutter app from the VS Code debugger no longer works
At some point (possible after the introduction of #5357) the debugger is no longer attaching to the Flutter app started from `serverpod start`. This makes the debugging experience poorer, as wiring the Flutter app to the debugger manually is not an easy task.

We need to restore the old behavior that made calls to launch the Flutter app from the debugger be intercepted by `serverpod start`, launch the app internally and expose for attaching by VS Code.

Crazelu:
There were two issues here:
1. VM service client requests that are sent before the upstream is available are permanently lost. This caused VS Code to never receive responses for some initial requests it made that were necessary to subsequently setting up breakpoints.
2. The `vmServiceInfoFile` path used in VS Code launch configuration became invalid after the introduction of multiple Flutter apps support.

VM service client requests are now buffered when no upstream exists and replayed when an upstream becomes available.
The `vmServiceInfoFile` path has also been updated in the launch template.

Closes #5452 

## Pre-launch Checklist

## Breaking changes

None


templates/serverpod_templates/projectname_server/pubspec.yaml
```diff
@@ -45,7 +45,8 @@ serverpod:
   # Companion Flutter apps that `serverpod start` can launch (Ctrl+R), keyed
   # by a display alias. The default device is the web server, opening the
   # machine's default browser when ready. All non-reserved properties are
-  # forwarded to `flutter run`. See the example below:
+  # forwarded to `flutter run`. For the best IDE debugging experience, use
+  # `chrome` or a native target as the device. See the example below:
```
marcelomendoncasoares, line 49: Suggestion: Since users are going to add the app here, add a reference that they also need to configure the `launch.json` to add multiple apps with debugger support.

marcelomendoncasoares (changes requested):
Small suggestion on the discovery of the configuration. Can we also add an integration test that represents what the debugger does, so we don't break it again?

Crazelu:
> Can we also add an integration test that represents what the debugger does, so we don't break it again?

The test on the VM Service proxy already covers that path.

templates/serverpod_templates/projectname_server/pubspec.yaml
Crazelu: This reference is already in `launch.json` where developers will check if they are looking to do this sort of thing.
It is also VS Code specific. I don't think we should be adding IDE instructions to the server's pubspec

marcelomendoncasoares: I was thinking only of a brief reference that would point to the documentation already present on the `launch.json` file, but you're right that users that want the debugger know that they have to configure this file.

marcelomendoncasoares:
> > Can we also add an integration test that represents what the debugger does, so we don't break it again?
> 
> The test on the VM Service proxy already covers that path.

The VM proxy test validates the attach behavior, but not against a running `serverpod start --no-flutter`, right? What I mean is an integration test that will validate that launching the app through the `start` correctly attaches to the VM proxy and receives the expected events - the test that should have broken and prevented this issue.

tests/serverpod_cli_e2e_test/test/test_assets/fake_flutter_vm_service.dart
```diff
@@ -0,0 +1,125 @@
...
+
+  _emitMachineEvent('app.start', {'appId': 'fake-flutter-app'});
+  await _waitForDebugPortGate();
+
+  final wsUri = 'ws://${server.address.address}:${server.port}/ws';
+  _emitMachineEvent('app.debugPort', {
+    'wsUri': wsUri,
+    'port': server.port,
+  });
+  _emitMachineEvent('app.started', {});
```
marcelomendoncasoares, line 28: Question: Why not use a real Flutter run process?

tests/serverpod_cli_e2e_test/test/start_test.dart
```diff
@@ -317,6 +317,256 @@ fields:
...
+          'templates',
+          'serverpod_templates',
+          'vscode',
+          'launch.json',
+        ),
+      ).readAsStringSync().replaceAll('projectname', projectName);
+      final configuredLaunch = originalLaunch.replaceFirst(
+        'flutter-vm-service-info-$projectName.json',
+        'flutter-vm-service-info-$appId.json',
+      );
```
marcelomendoncasoares: Suggestion: Use the originally created app name to avoid too much replacements. It will reduce complexity on the test setup that is unrelated to its goal.

tests/serverpod_cli_e2e_test/test/start_test.dart
```diff
@@ -317,6 +317,256 @@ fields:
...
+      serverProcess = null;
+      stdoutSubscription = null;
+      stderrSubscription = null;
+      processOutput.clear();
+      if (debugPortGate.existsSync()) debugPortGate.deleteSync();
+    });
+
+    test(
+      'serverpod start exposes the app-id VM-service info file and forwards '
+      'a VS Code-style handshake to Flutter',
```
marcelomendoncasoares: Suggestion: This description is out of the pattern.

marcelomendoncasoares (changes requested):
Some improvement suggestions on the new integration test.

tests/serverpod_cli_e2e_test/test/test_assets/fake_flutter_vm_service.dart
Crazelu: This part is simulating the IDE. The real flutter run process is still used inside of `serverpod start`

tests/serverpod_cli_e2e_test/test/start_test.dart
Crazelu: Done in [4ad0cc5](https://github.com/serverpod/serverpod/pull/5472/commits/4ad0cc506e4f28318d73f3daa36a01ff02355e2f)

tests/serverpod_cli_e2e_test/test/start_test.dart
Crazelu: Done in [4ad0cc5](https://github.com/serverpod/serverpod/pull/5472/commits/4ad0cc506e4f28318d73f3daa36a01ff02355e2f)

marcelomendoncasoares (changes requested):
Before looking at the code, on manual tests, I managed to start with success and see the debugger working on the server, but Flutter app breakpoints are not working - even though the app connects and the reload button works as expected.

From the image below, you can see that the breakpoint is not even evaluated.

Image

---

This test was done launching the compound with F5 directly, not trying the mixed setup of manual `serverpod start --no-flutter` + F5 to attach Flutter app.

marcelomendoncasoares (approved):
I have proved it working on `chrome`, which means that the issues with `webserver` might be related to the relation between the WSL and the browser on Windows. So we can merge this and fix the WSL issue later (or change the default device in this case).

PR #5479, opened 2026-07-17
fix: Keep target schema fresh after hot reload

### LINKED ISSUE #5433: After manually changing a model and creating and applying a migration I still get a warning in `serverpod start`
Steps to reproduce:

1. Start the server with `serverpod start`.
2. Make a change in a database model.
3. Create a database migration `M`.
4. Apply the database migration `A`.

The start command still reports that the table does not match the target database. The expected outcome would be to have no warning.

Image

abdulawalarif:
## Summary
- Fixes #5433.
- Changes generated `Protocol.targetTableDefinitions` from a `static final` field to a static getter.
- This ensures `serverpod start` re-evaluates the target schema after hot reload, preventing false database mismatch warnings after creating and applying migrations.

## Problem
After editing a model, creating a migration, and applying it from `serverpod start`, the integrity check could still warn that the database did not match the target state.

This happened because Dart hot reload does not re-run `static final` initializers, so `Protocol.targetTableDefinitions` could remain stale while the live database had already been updated.

## Test Plan
- Ran protocol generator tests:
  `cd tools/serverpod_cli && dart test test/generator/dart/client_code_generator/protocol/protocol_test.dart test/generator/dart/server_code_generator/protocol/protocol_test.dart`
- Regenerated protocols with `util/generate_all_parallel`.


marcelomendoncasoares (approved):
Thanks for the contribution investigating and fixing this issue, @abdulawalarif! I have just merged main and reduced a bit on the diff, but otherwise LGTM! 🚀

PR #5481, opened 2026-07-17
feat: Unify create and apply migrations

### LINKED ISSUE #5458: Unify `M Create migration` and `A Apply migrations` on the `serverpod start`

Crazelu:
Apply migrations button has been removed from the TUI. Migrations are now applied once they are created from the TUI button/keyboard shortcut. This also applies to repair migrations.

Closes #5458 

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start/tui/main_screen.dart
+++ b/tools/serverpod_cli/lib/src/commands/start/tui/main_screen.dart
@@ -23,7 +23,6 @@ class MainScreen extends StatelessComponent {
     this.onHotReload,
     this.onHotRestart,
     this.onCreateMigration,
-    this.onApplyMigration,
     this.onClearLogs,
     this.onLaunchApp,
     this.onTabSelected,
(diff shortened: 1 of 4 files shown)


marcelomendoncasoares:
As discussed, we should give a bit more thought on this to ensure that the workflow is clear when create succeeds, but apply fails - since the operation will leave the migration behind. One possibility is to then show a message informing the apply failure and suggesting `A` to apply manually again.

We'll discuss this in the team before proceeding with the change.

tools/serverpod_cli/lib/src/commands/start/tui/main_screen.dart
marcelomendoncasoares, line 785: Suggestion: Change the label to `Migrate`, so it also implies the apply action.

marcelomendoncasoares (changes requested):
Just one small suggestion to change the `Create migration` label.

tools/serverpod_cli/lib/src/commands/start/tui/main_screen.dart
Crazelu: Done in [26ca4b8](https://github.com/serverpod/serverpod/pull/5481/commits/26ca4b8647fa6e171f7fb0d230d78ba763435f4f)

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5402, opened 2026-06-29
feat: Run integration tests on embedded PostgreSQL (last mile)

nielsenko:
Follow up to #5271 , #5280 , and #5381 

Closes: #5119

Fixes: #5037
Fixes: #5144
Fixes: #5249

Run integration tests on embedded PostgreSQL, in parallel (no Docker).

`withServerpod` already boots an embedded PostgreSQL on demand and gives each test
group its own database on a shared postmaster. Isolation comes from per-group
databases (Postgres) / files (SQLite).

`serverpod_test_server` is hereby migrated the same way. Each test suite provisions
its own database on the shared postmaster, and servers that bind sockets
(`IntegrationTestServer.create`/`start`) use ephemeral ports, so these suites
also run concurrently, with no fixed-port or shared-state collisions. Tests
reach servers by their bound port via `IntegrationTestServer.apiUrl`/`webUrl`/…
helpers.

`concurrency_one` is removed and the run scripts drop `--concurrency=1`.

PR #5482, opened 2026-07-17
feat: Autostart docker if using PostgreSQL from `docker-compose.yaml`

marcelomendoncasoares:
This PR improves the automatic behavior of the `serverpod start` command regarding the use of Docker with multiple database configurations. If a project uses PostegreSQL from `docker-compose.yaml`, the command will now automatically start the PostgreSQL container before starting the server. Previously, users would have to explicitly specify `--docker` to start the Docker services. The automatic behavior can still be disabled by passing `--no-docker` to the command.

This behavior is important because it will be the fallback in case embedded database services are not available - for example, due to no compatible binary available.

This work is part of #5402.

## Pre-launch Checklist

## Breaking changes

None.


FXschwartz (approved):
LGTM!

Issue #5346, opened 2026-06-16
☂️ Remove deprecated elements


marcelomendoncasoares:
Since we're going to do a major bump, we should cleanup deprecated elements from the existing code. This issue will collect all of them at once, and we can decided whether to split up the actual work on sub-issues for simplicity:

- [x] ~~[Old static directory cache control classes from `serverpod/lib/src/web_server/routes/static_directory.dart`.](https://github.com/serverpod/serverpod/pull/5445)~~
- [x] ~~[All legacy widgets from `serverpod/lib/src/web_server/widgets/legacy_widget.dart`.](https://github.com/serverpod/serverpod/pull/5445)~~
- [x] [The `mini` option on `create` and all references to "Serverpod mini" (that no longer exists in the docs).](https://github.com/serverpod/serverpod/issues/5441)
- [x] [All `orderDescending` boolean parameters and the public `Order` constructor on the ORM.](https://github.com/serverpod/serverpod/issues/5440)
- [x] ~~[The `SerializationManagerServer` class (that is now a typedef to `DatabaseSerializationManager`).](https://github.com/serverpod/serverpod/pull/5445)~~
- [x] ~~[The `ignoreEndpoint` annotation (superseded by `doNotGenerate`).](https://github.com/serverpod/serverpod/pull/5445)~~
- [x] ~~[The `ScriptsParseException` on the CLI.](https://github.com/serverpod/serverpod/pull/5445)~~

## What should not be removed

- **Legacy auth interfaces**: since the e-mail user migration guide from legacy auth to the new auth has been recently finished, we should give more time for users to migrate before removing the legacy auth completely.

## Up for discussion

- **Old streaming methods**: It would be great to remove the legacy streaming interface completely, but this requires us to decouple the current legacy chat modules from the monorepo overrides.
- **Google Sign-In Web implementation**: Now that we have a [better Sign-In implementation for web using OAuth2 #5093](https://github.com/serverpod/serverpod/pull/5093), we can completely remove the native web implementation and solve all issues below:
  - #5358
  - #5188
  - #5042
  - #4569
  - #4164

marcelomendoncasoares:
As discussed, both the legacy streaming and the native Google Sign-In web will be removed.

PR #5486, opened 2026-07-20
refactor: Remove deprecated future call methods

### LINKED ISSUE #5448: Remove the legacy future call methods from Serverpod
Part of #5346

Remove the deprecated `Serverpod.futureCallWithDelay`, `Serverpod.futureCallAtTime`, and `Serverpod.cancelFutureCall` methods, replaced  by the generated `serverpod.futureCalls` API (`callWithDelay` / `callAtTime` / `callRecurring` / `cancel`).

This was originally part of the combined removal PR (#5445) but was split out because it is not a simple removal.

We need to:
- Remove the three deprecated methods and migrate the test server to the generated API.
- Keep `registerFutureCall` but hide it. It is still used by the generated code, but with name based scheduling gone it no longer makes sense to be public.
- Move future-calls generation before endpoint analysis in the CLI, so endpoints can import the generated `future_calls.dart`. Today that import fails `serverpod generate` on a clean tree because the file is generated after endpoints are analyzed.

FXschwartz:
Removes the deprecated string-based future call methods for the 4.0 release (part of #5346). Since the generated future calls API is now the only way to schedule calls, this PR also makes two changes:
- **Code generation order**: `future_calls.dart` is now generated before endpoint analysis (previously it was after endpoints were analyzed). This means an endpoint that imports the generated `future_calls.dart` to schedule a call now generates successfully from a clean tree on the first run, instead of failing with "Target of URI doesn't exist".
- **`Serverpod.registerFutureCall` is marked `@internal`**: the generated `FutureCalls.initialize()` handles registration, and with name based scheduling removed there is no public way to schedule a manually registered call, so it no longer makes sense as user-facing API.

Fixes #5448 

## Pre-launch Checklist

## Breaking changes
- `Serverpod.futureCallWithDelay`, `Serverpod.futureCallAtTime`, and `Serverpod.cancelFutureCall` have been removed.
- `Serverpod.registerFutureCall` ~~is now annotated `@internal`.~~ has been removed


packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -1,6 +1,7 @@
 import 'dart:async';
 import 'dart:io';
 
+import 'package:meta/meta.dart' as meta;
```
marcelomendoncasoares: Question: Is this alias really needed to avoid a conflict or just a projection of the below imports using aliases? If not needed, we should avoid the unnecessary alias.

packages/serverpod/lib/src/server/serverpod.dart
```diff
@@ -1134,70 +1140,6 @@ class Serverpod {
     _futureCallManager?.registerFutureCall(call, name);
   }
```
marcelomendoncasoares: Suggestion: Instead of using an `internal` marker, we could expose this only on the `ServerpodInternalMethods` and use this class on the generated code. This will make the interface cleaner for users, that won't see the `registerFutureCall` on completion.

marcelomendoncasoares (changes requested):
Only two comments for now, but I have one broader discussion point: can we also remove the `invoke` method from the base class?

This method received a null body on the implementation of generated future calls to enable extending the `FutureCall` class to declare random methods (like endpoints). But, now that we no longer support the legacy format - directly overriding `invoke` -, we could move it to an interface class that the generated code will implement together with extending `FutureCall`.

This would remove a confusion on the generation layer around `FutureCall` classes assuming different roles depending on the generics (either a generation spec or an actual executor). What do you think @FXschwartz and @Crazelu?

packages/serverpod/lib/src/server/serverpod.dart
FXschwartz: Done! Moved it to `ServerpodInternalMethods` and dropped the @internal annotation.

One thing I noticed is the generated code never calls this method, wrappers register through `FutureCallManager.registerFutureCall` via `FutureCalls.initialize()`. The only callers left are our own integration tests. 

We could delete it entirely and expose a `futureCallManager` getter on the extension instead, so the manager is the single registration API. @marcelomendoncasoares thoughts?

packages/serverpod/lib/src/server/serverpod.dart
FXschwartz: It was needed because `serverpod.dart` imports the generated endpoints/protocol as internal, so a `@internal` resolves to that import prefix and fails to compile ("annotation must be a const variable reference"). 

But per your other suggestion I moved `registerFutureCall` into `ServerpodInternalMethods`, so the `@internal` marker and the meta import are gone now!

packages/serverpod/lib/src/server/serverpod.dart
marcelomendoncasoares: Let's remove, then! I guess we don't even need to expose the `futureCallManager` getter, right? Because the `initialize` is called by `Serverpod` itself, passing the private `_futureCallManager`.

https://github.com/serverpod/serverpod/blob/b17aa3d3898c414e45c3a03a907aef621d420871/packages/serverpod/lib/src/server/serverpod.dart#L895-L901

FXschwartz: Removed! We do still need access for `broken_future_calls_test` though. It registers calls on the pod's own manager before start() to set up the broken/missing registration scenarios.

tests/serverpod_test_server/lib/src/endpoints/future_calls.dart
```diff
@@ -1,25 +1,12 @@
...
-      data,
-      const Duration(seconds: 1),
-    );
+  Future<void> makeFutureCall(Session session, SimpleData data) async {
+    await session.serverpod.futureCalls
+        .callWithDelay(const Duration(seconds: 1))
+        .testCall
+        .run(data);
   }
 }
```
marcelomendoncasoares, line 22: Question: Why remove the other endpoint that would call the future call that throws? Won't this reduce coverage?

tests/serverpod_test_server/lib/src/futureCalls/test_call.dart
```diff
@@ -1,14 +1,8 @@
...
-    if (object != null) {
-      var data = object;
-      session.log('${data.num}');
-    } else {
-      session.log('null');
-    }
+class TestCall extends FutureCall {
+  Future<void> run(Session session, SimpleData data) async {
+    session.log('${data.num}');
   }
```
marcelomendoncasoares, line 11: Question: Can't we still keep the object type nullable on the `TestCall` to achieve the same coverage?

tests/serverpod_test_server/lib/src/endpoints/future_calls.dart
marcelomendoncasoares, line 1: Suggestion: We should also add a future call that declares the `invoke` method to exercise the conflict - which should not be a problem as it was implemented previously.

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
```diff
@@ -142,12 +139,13 @@ void main() {
...
+        'when a scheduled future call throws '
         'then the diagnostic event handler gets called',
         () async {
-          await client.testFutureCalls.makeFutureCallThatThrows(
-            SimpleData(num: 42),
-          );
+          await pod.futureCalls
+              .callWithDelay(const Duration(seconds: 1))
+              .testExceptionCall
+              .run(p.SimpleData(num: 42));
```
marcelomendoncasoares: Suggestion: My suggestion is to keep this test through an endpoint to avoid changing the exact coverage surface - since we don't have the history on why this test goes through an endpoint.

tools/serverpod_cli/lib/src/analyzer/dart/future_call_analyzers/future_call_class_analyzer.dart
```diff
@@ -214,18 +214,14 @@ extension on ClassElement {
...
+    return allSupertypes.any(
+      (type) => type.element.name == 'InvokableFutureCall',
     );
-
-    final onlyDeclaresInvokeMethod =
-        methods.length == 1 && methods.first.name == 'invoke';
-
-    return extendsFutureCallDirectly && onlyDeclaresInvokeMethod;
   }
 }
```
marcelomendoncasoares, line 230: Question: What would be the case for a direct `name == 'InvokableFutureCall'`? Since we're checking if the class extends/implements and the `InvokableFutureCall` is an interface, it should not be possible for the `name` to match it exactly, right?

tools/serverpod_cli/lib/src/generator/analyzers.dart
```diff
@@ -262,6 +262,31 @@ class Analyzers {
...
+      var generatedFutureCallFiles =
+          await ServerpodCodeGenerator.generateFutureCalls(
+            protocolDefinition: ProtocolDefinition(
+              endpoints: const [],
+              models: allModels,
+              futureCalls: futureCalls,
+            ),
+            config: config,
+          );
+      changedFiles.addAll(generatedFutureCallFiles);
```
marcelomendoncasoares: Question: By bringing this here, are we now generating the future call files twice? Seems like we're breaking the layer abstractions by calling the generator in here. We should update the context (which is part of the logic inside the `generateFutureCalls`, but is it really needed that we write the files in this step?

tools/serverpod_cli/lib/src/generator/serverpod_code_generator.dart
```diff
@@ -58,6 +58,27 @@ abstract class ServerpodCodeGenerator {
...
+  }) async {
+    var allFiles = const DartServerCodeGenerator().generateFutureCallsCode(
+      protocolDefinition: protocolDefinition,
+      config: config,
+    );
+    await _writeFiles(allFiles);
+
+    return allFiles.keys.toList();
+  }
+
```
marcelomendoncasoares: Comment: If we manage to maintain the layers isolated, this method should not need to be introduced.

marcelomendoncasoares (changes requested):
Few comments to improve on the implementation.

tests/serverpod_test_server/lib/src/endpoints/future_calls.dart
FXschwartz: Good point and no good reason. Its only caller was the diagnostics test which I rewrote to schedule directly. I'll bring makeFutureCallThatThrows back and keep the diagnostics test going through it.

tools/serverpod_cli/lib/src/analyzer/dart/future_call_analyzers/future_call_class_analyzer.dart
FXschwartz: No user class should hit it through supertypes.

But for generating `packages/serverpod` itself, the analyzer scans serverpod's own lib and finds the `InvokableFutureCall` declaration, which extends `FutureCall`, so without this it gets classified as an abstract spec and its invoke(`Session, T?`) fails param validation with "The type T? is not a supported future call parameter type". 

I ran into this on the first `generate_all` run.

I can add a comment in the code explaining the reasoning better if you are good with it

tools/serverpod_cli/lib/src/generator/analyzers.dart
FXschwartz: The disk write isn't duplicated (second pass skips identical content) but yeah, the content is built twice and you are right it doesn't need to write files at this step. 

I'll build the future calls content and setOverlay it for endpoint analysis, then let protocol generation do the single real write. Same lifecycle as the protocol.dart stub. That also removes `ServerpodCodeGenerator.generateFutureCalls` entirely.

tests/serverpod_test_server/lib/src/endpoints/future_calls.dart
FXschwartz: Sounds good! I don't think anything special cases the name so it looks like it should just generate a `<Class>InvokeFutureCall` wrapper like any other method. I'll add one to the test server with an assertion that it schedules.

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
FXschwartz: I'll revert it to go through the endpoint. Only difference will be adapting it to the new harness from the integrated Postgres PR `startWithDatabase() + Client(pod.apiUrl)` instead of the fixed port.

tests/serverpod_test_server/lib/src/futureCalls/test_call.dart
FXschwartz: Yes we can, the generator handles a nullable model param fine, the dispatcher just passes it through instead of null-checking. I'll revert `TestCall` to `SimpleData?` so we keep the null-branch coverage from the old invoke implementation.

tools/serverpod_cli/lib/src/generator/serverpod_code_generator.dart
FXschwartz: Yep, this goes away with the overlay approach from your other comment.

tests/serverpod_test_server/lib/src/endpoints/future_calls.dart
FXschwartz: Done

tests/serverpod_test_server/lib/src/futureCalls/test_call.dart
FXschwartz: Done

tests/serverpod_test_server/lib/src/endpoints/future_calls.dart
FXschwartz: Done

tests/serverpod_test_server/test_integration/diagnostics/diagnostic_events_test.dart
FXschwartz: Done

tools/serverpod_cli/lib/src/generator/analyzers.dart
FXschwartz: Done

tools/serverpod_cli/lib/src/generator/dart/server_code_generator.dart
```diff
@@ -60,12 +60,10 @@ class DartServerCodeGenerator extends CodeGenerator {
...
-      if (protocolDefinition.shouldGenerateFutureCalls)
-        p.joinAll([
-          ...config.generatedServerFutureCallFilePathParts,
-        ]): serverClassGenerator
-            .generateServerFutureCalls()
-            .generateCode(),
+      ...generateFutureCallsCode(
+        protocolDefinition: protocolDefinition,
+        config: config,
+      ),
```
marcelomendoncasoares, line 66: Question: Do we need this code change? Why the need for the `generateFutureCallsCode` with a second `serverClassGenerator` instead of reusing the existing one as before?

tools/serverpod_cli/test/integration/generate/generate_from_scratch_test.dart
```diff
@@ -114,6 +114,117 @@ class MyFutureCall extends FutureCall {
     },
   );
 
+  group(
+    'Given a future call and an endpoint importing the generated future '
+    'calls file with no generated files',
```
marcelomendoncasoares:
```suggestion
    'Given a future call and an endpoint importing the generated future calls file with no generated files',
```

tools/serverpod_cli/lib/src/generator/analyzers.dart
```diff
@@ -262,6 +265,52 @@ class Analyzers {
...
+          // No overlay provider backing the analysis context; fall back to
+          // writing the file to disk. It is overwritten (with identical
+          // content) by generateProtocolDefinition below.
+          final file = File(entry.key);
+          await file.create(recursive: true);
+          await file.writeAsString(entry.value, flush: true);
+        }
+        changedFiles.add(entry.key);
+      }
+
```
marcelomendoncasoares: Question: I still wonder if we need to write here. Doesn't the endpoints analyzer uses provided models instead of re-analyzing models? Because, if models are not analyzed again, we should not need to write them here, even as an overlay - as nothing seems to be consuming these changes to overlay.

examples/legacy/chat/chat_server/lib/src/future_calls/example_future_call.dart
marcelomendoncasoares, line 1: Suggestion: Now that we decoupled the legacy chat examples from the monorepo, we can't change them together or we'll break CI on the next commit after merging to main.

marcelomendoncasoares (changes requested):
Just a few more comments on the analyzer changes.

tools/serverpod_cli/lib/src/generator/analyzers.dart
FXschwartz: Ok so you're right that the endpoints analyzer uses the provided models and doesn't re-analyze them, but that's separate from what consumes the overlay. I went and proved it by disabling this block and ran the from scratch check (rm -rf lib/src/generated + one generate):

```bash
ERROR: Endpoint analysis skipped due to invalid Dart syntax.
File: .../lib/src/endpoints/future_calls.dart
Error: Target of URI doesn't exist: 'package:serverpod_test_server/src/generated/future_calls.dart'.
Error: The getter 'futureCalls' isn't defined for the type 'Serverpod'.
```
With the block back, same run exits 0 with byte-identical output.

The provided models only feed signature validation. The analyzer still resolves each endpoint's Dart source through the real Dart analyzer and fails generation on severe errors, that resolution is the consumer of the overlay. The second error above is that `futureCalls` doesn't exist in the framework since it's the generated extension inside future_calls.dart, so an endpoint scheduling a future call can't resolve without that file having content somewhere. And since analysis gates generation, the run aborts before the file would ever get written.

I think this never came up before because nothing scheduled from an endpoint. The legacy `futureCallWithDelay` was a real method, so no generated import was needed, and our own generated API usage lives in test files the CLI never scans. With the legacy methods gone, the endpoint import is the only option, which is exactly what the from scratch CI gate hits. Same chicken-and-egg that protocol.dart solves with the temp stub. 

The "endpoint importing future calls" group in `generate_from_scratch_test.dart` covers it and fails if this block is removed.

@Crazelu Feel free to check me on this when you have time. The code here is slightly difficult to follow

Crazelu: You're right @FXschwartz 
This is only necessary now because we have a test that invokes a future call from an endpoint using the generated interface.

tools/serverpod_cli/lib/src/generator/dart/server_code_generator.dart
FXschwartz: The method needs to be callable from analyzers.dart with an endpoints free definition for the overlay, so it takes a ProtocolDefinition and builds its own LibraryGenerator

tools/serverpod_cli/lib/src/generator/analyzers.dart
marcelomendoncasoares: Makes sense. Then, my suggestion is to abstract this part to a separate private method with a doc comment explaining why it is needed. The use case of future calls being called from endpoints is indeed valid and should be supported.

tools/serverpod_cli/test/integration/generate/generate_from_scratch_test.dart
FXschwartz: Fixed!

examples/legacy/chat/chat_server/lib/src/future_calls/example_future_call.dart
FXschwartz: Reverted everything under `examples/legacy/chat`!

tools/serverpod_cli/lib/src/generator/analyzers.dart
FXschwartz: Done! Pulled it out into `_shadowFutureCallsFile` with a doc comment explaining the why and the lifecycle.

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5490, opened 2026-07-21
fix: Correct Windows embedded PostgreSQL builds and tests

marcelomendoncasoares:
## Summary

This PR fixes the Windows embedded PostgreSQL bundle and hardens the runtime and CI paths that exercise embedded PostgreSQL across supported platforms.

- build a portable MinGW pgvector binary that does not crash while constructing `halfvec` indexes
- run Windows PostgreSQL tests as a real standard user, with deterministic logging and the precise filesystem/tooling access they need
- fall back to dynamic loopback TCP when Unix domain sockets are unavailable
- shut PostgreSQL down through `pg_ctl` on Windows and wait for resources to be released before cleanup
- isolate mutable test artifacts and bootstrap fixtures so concurrent jobs and test discovery cannot interfere with them
- preserve the complete integration-test surface while making slow and destructive tests explicit in the CI matrix

Bundle revision remains `r1` because no `r1` artifact was published.

### Windows bundle correctness

- Undefine `__FLT16_MAX__` for the MinGW pgvector build so pgvector selects its portable `uint16` half-precision implementation.
- Keep runtime SIMD dispatch disabled for redistributable bundles without using the broader `-fno-isolate-erroneous-paths-dereference` and `-fno-delete-null-pointer-checks` workarounds.
- Retain the Windows smoke battery that initializes PostgreSQL, loads pgvector and PostGIS, and exercises the supported vector, half-vector, index, geography, and geometry contract before an artifact becomes publishable.

### Windows process and test execution

- Centralize PsExec command preparation in `util/prepare_windows_psexec_process` for writable temporary storage, deterministic log capture, command argument preservation, and exit-code propagation.
- Layer `util/prepare_windows_standard_user_test` on top for Dart and Flutter tests that must run outside the hosted runner's Administrators group.
- Extract the repeated PsExec invocation and log handling into `util/run_windows_standard_user_test.cmd`.
- Create an ephemeral `serverpod_ci` standard account instead of relying on `PsExec -l`; Low Integrity prevented Flutter plugin symlink creation.
- Grant that account Modify access to the checkout, pub cache, Flutter SDK, and isolated test workspace. An investigated Full Control grant was reverted because it did not address the remaining failure.
- Enable plugin symlinks, isolate `APPDATA` and `LOCALAPPDATA`, trust the checkout and Flutter SDK as Git repositories, clear disposable Flutter startup locks, and run a bounded Flutter preflight under the actual test account.
- Forward only the required CI, tool, shard, CLI bundle, and embedded PostgreSQL variables rather than copying the runner's complete environment.
- Reuse the CI-built Serverpod CLI bundle in bootstrap tests instead of rebuilding or resolving the CLI separately in every shard.
- Allow embedded integration scripts to honor an explicitly supplied `SERVERPOD_DATABASE_DATA_PATH` while preserving their existing local default.
- Keep bootstrap project database configuration intact; the Windows wrapper no longer overrides it.

### Embedded PostgreSQL runtime

- Select Unix domain sockets by default where Dart supports them and otherwise fall back automatically to `TcpTransport` on `127.0.0.1` with a dynamic port and configured or generated password.
- Update the package API documentation, README, platform guide, shared database configuration documentation, package descriptions, and transport tests for that behavior.
- Add shared `pg_ctl` shutdown support for Windows supervisors, attached supervisors, and stale-lock repair, with signal escalation retained as a fallback.
- Represent `pg_ctl stop` modes with the typed `PgCtlStopMode` enum instead of an arbitrary string.
- Wait for the postmaster process, output streams, and log sink to close before considering shutdown complete.
- Stop the shared test postmaster before deleting its data directory and retry deletion for a bounded period while PostgreSQL child processes release their handles.
- Route the auth, module, and non-vector embedded integration runners through the same cleanup implementation.

### Test isolation and determinism

- Keep the fake postmaster alive with an active timer on every supported Dart VM and await the spawned process's actual exit during stale-lock repair tests.
- Create server-template bootstrap temporary directories and randomized fixtures inside `setUpAll`, then remove them in `tearDownAll`, avoiding filesystem allocation during test discovery.
- Build expected bootstrap MCP configuration only after its randomized project name has been initialized.
- Accept a future-call timestamp equal to, but never before, the expected time to account for Windows clock/database precision.
- Tag the migration-artifact test and run it in a dedicated matrix job because it mutates checked-out protocol and migration directories.
- Generate that test's migration through the production `createMigrationAction` in-process. It still writes real protocols and migration artifacts, applies the migration and repair through the endpoint, and verifies migration integrity; dedicated CLI and bootstrap suites continue to cover the executable boundary.
- Give the large tagged `serverpod_test_server` integration suite a 30-minute timeout for slower ARM and macOS runners.

## Root causes

The Windows bundle crash was caused by MinGW defining `__FLT16_MAX__`, which made pgvector select native `_Float16`. MinGW's `isinf` fallback does not support that type and emits `__builtin_trap()`, terminating the PostgreSQL backend with `0xC000001D` during `halfvec` index construction.

The hosted Windows runner executes with an administrator token, which PostgreSQL rejects. The initial Low Integrity workaround allowed PostgreSQL to start but prevented normal Dart and Flutter behavior, including writing tool state and creating plugin symlinks. Running as an explicit standard account matches the user configuration being tested.

Additional failures came from Windows retaining PostgreSQL child-process file handles during cleanup, bootstrap fixtures being created before test setup, and one migration test mutating checkout paths while the broader suite ran concurrently.

## Coverage

- No integration suite or supported database behavior was removed.
- The migration-artifact test moved out of the concurrent tagged suite into its own job on every embedded PostgreSQL platform.
- The endpoint migration test still produces and applies real migration artifacts; only its unrelated child-CLI wrapper was removed.
- The Windows bundle smoke test remains a release gate and exercises the operation that originally crashed.
- Bootstrap tests still cover project generation through the packaged CLI on both supported Flutter versions.


tests/bootstrap_project/test/project_server_test.dart
```diff
@@ -7,15 +7,34 @@ import 'package:test/test.dart';
...
+    tempPath = tempDirectory!.path;
+
+    final project = createRandomProjectName(tempPath);
+    projectName = project.projectName;
+    final projectFolders = createProjectFolderPaths(projectName);
+    serverDir = projectFolders.serverDir;
+    flutterDir = projectFolders.flutterDir;
+    clientDir = projectFolders.clientDir;
+
+    final runningProject = createRandomProjectName(tempPath);
```
Crazelu: Why is the running project name different from the project name created above?

marcelomendoncasoares: This change was to isolate the test that runs the database from the globally created test that is used for inspection. But it is indeed not clean... I'll refactor this test to completely separate the group that needs to start the database from the global project.

marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5490/commits/b17aa3d3898c414e45c3a03a907aef621d420871.

Crazelu (approved):
🙌

PR #5384, opened 2026-06-24
feat: Unify and customize social sign-in button styling

FXschwartz:
This PR gives them one uniform, theme-aware appearance by default and a shared `SignInButtonStyle` to customize them globally or per button.

- A centralized `SignInButtonStyle` (size, shape, logo alignment, text, minimum width, text style, and background/foreground/border colors), exposed to every button.
- All buttons share one neutral appearance that follows the theme.
- Set a provider widget's own `style`/`theme` to put that button into its brand colors.
- Every button inherits the app's theme font instead of a hardcoded one.
- A `center` and a `left` alignment mode, both computed from the button width so the logos and labels line up with the native Apple button at any size.

Apple is drawn by the `sign_in_with_apple` native button, which hardcodes the system font and only offers black/white/outlined
presets, so it approximates the shared colors with its nearest preset and ignores `textStyle`. We overlay the shared border on it so it matches the others.

Note:
- Because we are having to dynamically fix other buttons to line up with the native app button this means at smaller button sizes it can be obvious the button and text is not centered. For instance "Continue with Facebook" button.

Fixes #5329.
Fixes #4717.
Fixes #4571.
Fixes #5143.

## Pre-launch Checklist

- [x] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [x] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
- [x] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.
- [x] I added new tests to check the change I am making.
- [x] All existing and new tests are passing.
- [x] Any breaking changes are documented below.

If you need help, consider asking for advice on the [discussion board](https://github.com/serverpod/serverpod/discussions).

### Shared enums replace all per-provider layout enums:
```dart
SignInButtonSize, SignInButtonTextVariant, SignInButtonShape, SignInButtonLogoAlignment.
```
Replaces:
- `GitHubButtonSize/Text/Shape/LogoAlignment`
- `MicrosoftButton{Size,Text,Shape,LogoAlignment}`
- `FacebookButton{Size,Text,Shape,LogoAlignment}`
[...]

PR #5093, opened 2026-05-08
refactor(auth): Implement Google Sign-In for web with OAuth2 PKCE flow

vfiruz97:
Implement Google Sign-In for web with OAuth2 PKCE flow. It opens url in new tab and redirect back with auth code.

It solves several issues of google_sign_is has on web.
 - Auto translating text on button
 - Forces use of build-in rendered button
 - Two times asks choose accounts popup
 - Makes use inly popups whiich Browsers blocking 

## Pre-launch Checklist

- [x] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [x] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
- [x] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.
- [x] I added new tests to check the change I am making.
- [x] All existing and new tests are passing.
- [x] Any breaking changes are documented below.

If you need help, consider asking for advice on the [discussion board](https://github.com/serverpod/serverpod/discussions).

## Breaking changes

No breaking changes. 

### Before (web)
```dart
// main.dart or startup
await client.auth.initializeGoogleSignIn(clientId: 'web-client-id.apps.googleusercontent.com');

// widget — showed Google iFrame button
GoogleSignInWidget(client: client, locale: 'en', onAuthenticated: ...)
```

### After (web)

**Step 1** — Create `web/auth.html` in your Flutter project with this exact content:
```html
<!DOCTYPE html>
<title>Authentication complete</title>
<p>Authentication is complete. If this does not happen automatically, please close the window.</p>
<script>
  function postAuthenticationMessage() {
    const message = {
      'flutter-web-auth-2': window.location.href
    };
    if (window.opener) {
      window.opener.postMessage(message, window.location.origin);
      window.close();
    } else if (window.parent && window.parent !== window) {
      window.parent.postMessage(message, window.location.origin);
    } else {
      localStorage.setItem('flutter-web-auth-2', window.location.href);
      window.close();
    }
  }
  postAuthenticationMessage();
</script>
```

**Step 2** — Google Cloud Console:
[...]

PR #5498, opened 2026-07-22
refactor: Remove the native Google Sign-In web implementation

### LINKED ISSUE #5442: Remove the native Google Sign-In web implementation
Part of #5346. Blocked by #5329 - the sign-in UI refactor already removes the Google web button, so this should land on top of it to avoid conflicts.

With the OAuth2 (PKCE) web sign-in flow introduced in #5093, the native GSI-based web implementation in `serverpod_auth_idp_flutter` can be removed completely.

FXschwartz:
With the OAuth2 (PKCE) web sign-in flow introduced in #5093, the native GSI-based web implementation in `serverpod_auth_idp_flutter` is no longer needed. The Google-hosted web button was already removed in #5384. 

This PR removes the remaining runtime fallback so the PKCE redirect flow is the only Google sign-in path on web.

- `initializeGoogleSignIn` now throws an ArgumentError on web when redirectUri is not provided, instead of silently falling back to the google_sign_in package.
- `GoogleAuthController` no longer initializes or subscribes to `google_sign_in` on web.
- Removed the `google_sign_in_web` dependency.
- Removed web-only workarounds that existed for the old button.

Note: this PR is stacked on #5384 and should be merged after it.

Fixes #5442 

## Pre-launch Checklist

## Breaking changes
- On web, initializeGoogleSignIn requires redirectUri. Apps relying on the GSI fallback must configure the PKCE redirect flow.
- Lightweight sign-in (FedCM / One Tap) is no longer available on web; attemptLightweightSignIn now only has an effect on native platforms.

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/google/google_sign_in_service.dart
+++ b/modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/google/google_sign_in_service.dart
@@ -6,8 +6,10 @@ import 'package:serverpod_auth_core_flutter/serverpod_auth_core_flutter.dart';
 
 import 'google_web_sign_in_service.dart';
 
-/// Service to manage Google Sign-In and ensure it is only initialized once
-/// throughout the app lifetime.
+/// Service to manage Google Sign-In on native platforms (iOS, Android, macOS)
+/// and ensure it is only initialized once throughout the app lifetime.
+///
+/// On web, [GoogleWebSignInService] is used instead.
 class GoogleSignInService {
   /// Singleton instance of the [GoogleSignInService].
   static final GoogleSignInService instance = GoogleSignInService._internal();
@@ -58,7 +60,7 @@ class GoogleSignInService {
 
       await googleSignIn.initialize(
         clientId: clientId,
-        serverClientId: !kIsWeb ? serverClientId : null,
+        serverClientId: serverClientId,
         nonce: nonce,
         hostedDomain: hostedDomain,
       );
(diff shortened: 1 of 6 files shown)


modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/google/google_sign_in_service.dart
```diff
@@ -140,7 +141,15 @@ extension DisconnectGoogleSignIn on FlutterAuthSessionManager {
     String? redirectUri,
```
marcelomendoncasoares, line 141: Suggestion: Now that only the OAuth2 workflow will be supported, we should make the `redirectUri` explicitly required. Then, we won't need the exception below. A migration guide for this will be importat.

marcelomendoncasoares (changes requested):
Only one change to make the configuration safer - now that we have only one path.

modules/serverpod_auth/serverpod_auth_idp/serverpod_auth_idp_flutter/lib/src/google/google_sign_in_service.dart
FXschwartz: @marcelomendoncasoares The main reason I did not do this is because then it becomes required on the native non web apps and wouldn't be used.

FXschwartz: We could split the api so that native uses `initializeGoogleSignIn` and web would use something like `initializeGoogleWebSignIn`

marcelomendoncasoares: You're right! I missed that the method is used for all platforms. Let's leave it as is.

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5509, opened 2026-07-27
fix: Release sessions from MessageCentral when streams are cancelled before session close

### LINKED ISSUE #5502: MessageCentral keeps closed sessions alive when a method stream is cancelled before its session closes
## Summary

Every websocket teardown strands one `Session` per open method stream that used `session.messages.createStream`. The session stays reachable from `MessageCentral._sessionToCleanupCallbacksLookup` for the lifetime of the process, and with it its `Request`, its `SessionLogManager` buffer and — for a `StreamingSession` — the socket it arrived on.

Reproduced on **3.4.10** and **3.4.11** (the relevant files are identical in both).

## Where it happens

`MessageCentral.createStream` registers a cleanup callback keyed by the session:

```dart
addListener(session, channelName, addToStream);
_addCleanupCallback(session, controller.close);

controller.onCancel = () {
  _removeCleanupCallback(session, controller.close);
  removeListener(session, channelName, addToStream);
};
```

Two things then combine:

**1. `_removeCleanupCallback` empties the set but leaves the entry behind**, so the session stays as a map key:

```dart
void _removeCleanupCallback(Session session, Function() callback) {
  var callbacks = _sessionToCleanupCallbacksLookup[session];
  if (callbacks == null) return;
  callbacks.remove(callback);   // the entry for `session` is never removed
}
```

**2. `removeListenersForSession` — the only place that drops that entry — returns before reaching it** once the other per-session lookups are gone (which `removeListener`, called just above, has already done):

```dart
void removeListenersForSession(Session session) {
  var channelNames = _sessionToChannelNamesLookup[session];
  if (channelNames == null) return;      // <-- returns before the cleanup map
  var listeners = _sessionToCallbacksLookup[session];
  if (listeners == null) return;
  ...
```

So whenever `controller.onCancel` runs *before* `Session.close()`, the entry is orphaned.

## Why that is the normal order

`MethodStreamManager.closeAllStreams()` cancels the output subscription first:

```dart
var closeSubscriptionFutures = outboundStreamContexts.map(
  (c) => c.subscription.cancel().timeout(...)
);
```

Cancelling it tears down `addStream` → the endpoint's `async*` → the inner `createStream` controller, so that controller's `onCancel` (`_removeCleanupCallback` + `removeListener`) runs first, and only then does the chain reach the output controller's `onCancel`, which calls `session.close()`.

And `closeAllStreams()` sits in the `finally` around the socket's event loop in `MethodWebsocketRequestHandler.handleWebsocket`, so **every** websocket teardown takes this path — graceful or not.

The other path, `closeStream()` (a client closing one stream while the socket stays up), awaits the output controller's `onCancel` first, so `session.close()` runs while the lookups are still populated and the entry is cleaned up. That asymmetry is why this only becomes visible when connections actually drop.

## Impact

Seen on a production server behind an unstable network: RSS 579 MB → 3.09 GB over ~10 hours with only ~120 live sockets. A forced-GC heap snapshot showed thousands of retained `_WebSocketImpl` / `_HttpParser` / `_HttpResponse` / `_HttpOutgoing` (identical counts), `MethodStreamSession` and `SessionLogManager` instances — roughly one retained connection graph every 7 seconds, matching how often clients reconnected.

## Reproduction

Prints `100` for the socket teardown order and `0` for the single-stream close.

<details>
<summary>repro.dart (needs only <code>serverpod: 3.4.10</code>)</summary>

```dart
import 'dart:async';
import 'dart:mirrors';

class StubServerpod implements Serverpod {
  @override
  dynamic noSuchMethod(Invocation invocation) => null; // redisController: null
}

class StubSession implements Session {
  StubSession(this._messageCentral);

  final MessageCentral _messageCentral;
  final _serverpod = StubServerpod();
  final _willCloseListeners = <WillCloseListener>[];

  @override
  Serverpod get serverpod => _serverpod;

  @override
  void addWillCloseListener(WillCloseListener listener) =>
      _willCloseListeners.add(listener);

  @override
  void removeWillCloseListener(WillCloseListener listener) =>
      _willCloseListeners.remove(listener);

  /// Session.close(), minus logging and the database.
  Future<int?> close({dynamic error, StackTrace? stackTrace}) async {
    final listeners = _willCloseListeners.toList();
    _willCloseListeners.clear();
    for (final listener in listeners) {
      await listener(this);
    }
    _messageCentral.removeListenersForSession(this);
    return null;
  }

  @override
  dynamic noSuchMethod(Invocation invocation) => null;
}

/// An endpoint method returning a channel stream.
Stream<SerializableModel> subscribeToChannel(
  MessageCentral messageCentral,
  Session session,
) async* {
  yield* messageCentral.createStream<SerializableModel>(session, 'a-channel');
}

/// The plumbing MethodStreamManager builds around such a method.
({StreamController<dynamic> output, StreamSubscription<dynamic> subscription})
openMethodStream(MessageCentral messageCentral, StubSession session) {
  var cancelled = false;
  late final StreamController<dynamic> output;
  output = StreamController<dynamic>(
    onCancel: () async {
      if (cancelled) return;
      cancelled = true;
      await session.close(); // method_stream_manager.dart:365
    },
  );

  final subscription = output.stream.listen((_) {});
  output.addStream(subscribeToChannel(messageCentral, session));

  return (output: output, subscription: subscription);
}

Future<void> settle() async {
  for (var tick = 0; tick < 10; tick++) {
    await Future<void>.delayed(Duration.zero);
  }
}

int retainedSessions(MessageCentral messageCentral) {
  final library = reflectClass(MessageCentral).owner as LibraryMirror;
  final map = reflect(messageCentral)
      .getField(
        MirrorSystem.getSymbol('_sessionToCleanupCallbacksLookup', library),
      )
      .reflectee;
  return (map as Map).length;
}

Future<void> main() async {
  // 1. The socket goes away: closeAllStreams() cancels the subscription first,
  //    and session.close() runs from inside that cancel chain.
  final socketTeardown = MessageCentral();
  for (var connection = 0; connection < 100; connection++) {
    final session = StubSession(socketTeardown);
    final stream = openMethodStream(socketTeardown, session);
    await settle();

    await stream.subscription.cancel(); // method_stream_manager.dart:181
    await settle();
  }
  print('after 100 socket teardowns     '
      '_sessionToCleanupCallbacksLookup: ${retainedSessions(socketTeardown)}');

  // 2. The client closes one stream while the socket stays up:
  //    closeStream() awaits the output controller's onCancel first.
  final singleStreamTeardown = MessageCentral();
  for (var connection = 0; connection < 100; connection++) {
    final session = StubSession(singleStreamTeardown);
    final stream = openMethodStream(singleStreamTeardown, session);
    await settle();

    await stream.output.onCancel?.call(); // method_stream_manager.dart:222
    unawaited(stream.subscription.cancel()); // method_stream_manager.dart:223
    await settle();
  }
  print('after 100 single-stream closes '
      '_sessionToCleanupCallbacksLookup: '
      '${retainedSessions(singleStreamTeardown)}');
}
```

</details>

```
after 100 socket teardowns     _sessionToCleanupCallbacksLookup: 100
after 100 single-stream closes _sessionToCleanupCallbacksLookup: 0
```

The stub session is only there to keep the repro dependency-free — `MessageCentral` reaches for nothing on a session besides its identity and `serverpod.redisController`, and `close()` is a copy of `Session.close()`. The same numbers show up in a real server; that is where this was found.

## Suggested fix

Either would close it, and both look worth doing:

- drop the map entry in `_removeCleanupCallback` when the set becomes empty;
- clean `_sessionToCleanupCallbacksLookup` in `removeListenersForSession` before the early returns, so a session is always fully released no matter which side got there first.

Related: #5503 — `Endpoint.setUserObject` also keeps sessions alive for the life of the process.

FXschwartz:
`MessageCentral` keeps cleanup callbacks in a map keyed by `Session`. Cancelling a stream created with `createStream` emptied the session's callback set but left the entry in the map, and `removeListenersForSession` early returned before reaching it once the listener lookups were empty. So whenever a stream was cancelled before its session closed, the session stayed in the map for the lifetime of the server.

- `_removeCleanupCallback` removes the map entry when the callback set becomes empty
- `removeListenersForSession` now purges all session lookups and runs the cleanup callbacks unconditionally, so release no longer depends on teardown order
- Added `hasSessionReferences` (`@visibleForTesting`) and a regression test covering both orderings, the cancel-first tests fail without the fix

Fixes #5502 

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/packages/serverpod/lib/src/server/message_central.dart
+++ b/packages/serverpod/lib/src/server/message_central.dart
@@ -246,6 +246,10 @@ class MessageCentral {
     }
   }
 
+}
+
+/// Internal methods for [MessageCentral].
+extension MessageCentralInternalMethods on MessageCentral {
   /// Whether [session] is still referenced by any internal lookup.
   @visibleForTesting
   bool hasSessionReferences(Session session) =>
--- a/packages/serverpod/lib/src/server/message_central.dart
+++ b/packages/serverpod/lib/src/server/message_central.dart
@@ -245,7 +245,6 @@ class MessageCentral {
       callback();
     }
   }
-
 }
 
 /// Internal methods for [MessageCentral].
--- a/packages/serverpod/lib/src/server/message_central.dart
+++ b/packages/serverpod/lib/src/server/message_central.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 
+import 'package:meta/meta.dart';
 import 'package:serverpod/serverpod.dart';
 
 /// Channels that are listened to by the Serverpod Framework.
@@ -155,20 +156,20 @@ class MessageCentral {
   /// Removes all listeners from the specified [Session]. This method is
   /// automatically called when [StreamingSession] is closed.
   void removeListenersForSession(Session session) {
-    // Get subscribed channels
-    var channelNames = _sessionToChannelNamesLookup[session];
-    if (channelNames == null) return;
-    var listeners = _sessionToCallbacksLookup[session];
-    if (listeners == null) return;
+    var channelNames = _sessionToChannelNamesLookup.remove(session);
+    var listeners = _sessionToCallbacksLookup.remove(session);
 
-    for (var channelName in channelNames) {
-      for (var listener in listeners) {
-        _removeListener(session, channelName, listener);
+    if (channelNames != null && listeners != null) {
+      for (var channelName in channelNames) {
+        for (var listener in listeners) {
+          _removeListener(session, channelName, listener);
+        }
       }
     }
 
-    _sessionToChannelNamesLookup.remove(session);
-    _sessionToCallbacksLookup.remove(session);
+    // Executed unconditionally, so the session is fully released even if the
+    // listener lookups were already emptied by removeListener.
+    _executeCleanupCallbacks(session);
   }
 
   void _removeListener(
@@ -184,8 +185,6 @@ class MessageCentral {
         session.serverpod.redisController!.unsubscribe(channelName);
       }
     }
-
-    _executeCleanupCallbacks(session);
   }
 
   /// Creates a stream that listens to a specified channel.
@@ -233,16 +232,24 @@ class MessageCentral {
     if (callbacks == null) return;
 
     callbacks.remove(callback);
+    if (callbacks.isEmpty) {
+      _sessionToCleanupCallbacksLookup.remove(session);
+    }
   }
 
   void _executeCleanupCallbacks(Session session) {
-    var callbacks = _sessionToCleanupCallbacksLookup[session];
+    var callbacks = _sessionToCleanupCallbacksLookup.remove(session);
     if (callbacks == null) return;
 
     for (var callback in callbacks) {
       callback();
     }
-
-    _sessionToCleanupCallbacksLookup.remove(session);
   }
+
+  /// Whether [session] is still referenced by any internal lookup.
+  @visibleForTesting
+  bool hasSessionReferences(Session session) =>
+      _sessionToChannelNamesLookup.containsKey(session) ||
+      _sessionToCallbacksLookup.containsKey(session) ||
+      _sessionToCleanupCallbacksLookup.containsKey(session);
 }
--- a/packages/serverpod/lib/server.dart
+++ b/packages/serverpod/lib/server.dart
@@ -8,7 +8,7 @@ export 'src/server/future_call_dispatch.dart';
 export 'src/server/future_call_manager/cron.dart';
 export 'src/server/future_call_manager/future_call.dart';
 export 'src/server/future_call_manager/future_call_manager.dart';
-export 'src/server/message_central.dart';
+export 'src/server/message_central.dart' hide MessageCentralInternalMethods;
 export 'src/generated/protocol.dart'
     show
         CronFutureCallScheduling,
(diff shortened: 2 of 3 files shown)


packages/serverpod/lib/src/server/message_central.dart
```diff
@@ -233,16 +232,24 @@ class MessageCentral {
...
-
-    _sessionToCleanupCallbacksLookup.remove(session);
   }
+
+  /// Whether [session] is still referenced by any internal lookup.
+  @visibleForTesting
+  bool hasSessionReferences(Session session) =>
+      _sessionToChannelNamesLookup.containsKey(session) ||
+      _sessionToCallbacksLookup.containsKey(session) ||
+      _sessionToCleanupCallbacksLookup.containsKey(session);
```
marcelomendoncasoares, line 257: Suggestion: Place this on a separate extension class to avoid leaking test specifics on the public class. Then, hide this extension from the barrel export.

marcelomendoncasoares (changes requested):
The solution is clean! Only one minor suggestion to avoid leaking `visibleForTesting` members into the public interface.

packages/serverpod/lib/src/server/message_central.dart
FXschwartz: Done!

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5511, opened 2026-07-27
chore: Detach the legacy chat module from the monorepo overrides

### LINKED ISSUE #5443: Detach the legacy chat module from the monorepo overrides
Part of #5346. Prerequisite for removing the legacy streaming API.

`modules/legacy/serverpod_chat` is the last consumer of the legacy streaming interface, `streamOpened` / `handleStreamMessage` / `sendStreamMessage` on the server (`chat_endpoint.dart`) and `StreamingConnectionHandler` / `openStreamingConnection` on the client (`chat_controller.dart`, `chat_dispatch.dart`). It also depends on the legacy auth module.

FXschwartz:
`modules/legacy/serverpod_chat` is the last consumer of the legacy streaming interface.  As discussed in the issue, the module stays in the repo but no longer builds against the local packages: it is frozen at hosted `4.0.0-beta.0`.

## Changes
- The chat module and chat example templates in `templates/pubspecs/` are now slot free and hard pinned to hosted `4.0.0-beta.0`, with all path overrides to the main/auth packages removed.
- `analyze-pubspecs` ignores the two frozen directories, since the frozen serverpod pin would otherwise be reported as a version mismatch once the monorepo version moves on.
- Chat is removed from the `serverpod_generate` CI matrix and from `SERVERPOD_SERVER_PATHS`, so `generate_all` / `create_migrations` and CI no longer regenerate the frozen module with the HEAD CLI. It stays in `run_tests_analyze` and the melos workspace as a regression guard.
- `util/update_pubspecs` no longer copies the monorepo CHANGELOG/README into the chat packages, so future releases don't stamp their changelogs onto the frozen module.
- **Partially reverts the chat part of #5439**: `chat_input.dart` and the `file_picker` constraint go back to their pre-#5439 state. The `file_picker: ^12` migration only resolved against the local (unpublished) `serverpod_auth_shared_flutter`; the hosted `4.0.0-beta.0` release caps `file_picker <11.0.0`, so the frozen module cannot use it. The regenerated protocol files from #5439 are kept.

@marcelomendoncasoares Tagging you as the code I partially reverted belonged to you.

Fixes #5443. Part of #5346

## Pre-launch Checklist

## Breaking changes
None


marcelomendoncasoares (approved):
LGTM! Only the dart format is not passing, but we can merge right after fixing.

PR #5518, opened 2026-07-29
fix: Update language server state when model files change on disk

### LINKED ISSUE #5493: language-server doesn't update as intended
### Describe the bug

whenever a change like renaming or moving content to another file the language server doen't catch up these update

### To reproduce

rename a directory holding model .spy.yaml files

, or rename a model .spy.yaml file

, or copy a model content to another file

### Expected behavior

catch up files, folders, and content changes

### Library version

Serverpod version: 4.0.0-beta.0

Flutter 3.44.6 • channel stable • https://github.com/flutter/flutter.git
Framework • revision ee80f08bbf (13 days ago) • 2026-07-08 15:02:06 -0700
Engine • hash d3a3293399556a85388faf8c6f0723a7a5597aa8 (revision 83675ed276) (21 days ago) • 2026-06-30 16:59:03.000Z
Tools • Dart 3.12.2 • DevTools 2.57.0

### Platform information

Edition	Windows 11 Pro
Version	25H2
Installed on	‎5/‎12/‎2026
OS build	26200.8875
Experience	Windows Feature Experience Pack 1000.26100.334.0

### Additional context

_No response_

- [ ] I want to work on this

FXschwartz:
The language server read model files from disk exactly once, at `initialize`. After that it only reacted to `didOpen`/`didChange`/`didClose` for documents opened in the editor.

- Renaming a model file left stale diagnostics on the old path and never validated the new file.
- Renaming a directory holding models left everything stale.
- Copying model content to a new file (or any file created outside the editor) was invisible until the file was opened.

Additionally, removing a model from the analyzer never published empty diagnostics, so stale entries persisted in the Problems panel indefinitely.

## This PR
- The server now asks the client to watch the file system via `client/registerCapability` for `workspace/didChangeWatchedFiles`, gated on the client's `dynamicRegistration` capability. Two watchers are registered: `**/*.{spy,yaml,yml}` for model files, and `**/*` restricted to create/delete events which is needed because directory renames and deletions only surface as events on the directory itself.
- On relevant events, the server performs a debounced, serialized rescan of the model files on disk (reusing the same loader as `initialize`) and diffs against the analyzer state: vanished models are removed and their diagnostics cleared, new and changed files are (re)validated.
- Open documents are tracked and never replaced or removed by disk state.
- Closing a document whose file was deleted now also clears its diagnostics.

## Also
- `didOpen` now registers models via `ModelHelper.createModelSourceForPath`, resolving the correct module alias for shared package models instead of always using the default alias.
- Fixed a malformed activation event in the VS Code extension (`"activationEvents:serverpod"` to `"onLanguage:serverpod"`). The `.spy.yaml`/`.spy.yml`/`.spy` extensions map to the `serverpod` language ID, so with only `onLanguage:yaml` valid, the extension never activated in sessions where only `.spy` model files were opened.

Fixed #5493 

## Pre-launch Checklist

## Breaking changes
None


tools/serverpod_cli/test/integration/language_server/language_server_test.dart
```diff
@@ -0,0 +1,250 @@
...
+
+void main() {
+  setUpAll(() {
+    CommandLineExperimentalFeatures.initialize([]);
+  });
+
+  group(
+    'Given an initialized language server for a project with an invalid model '
+    'file in a models subdirectory',
+    () {
```
marcelomendoncasoares, line 92: Question: Should we also have a test for the success case? (only valid models, no errors)

tools/serverpod_cli/test/integration/language_server/language_server_test.dart
```diff
@@ -0,0 +1,250 @@
...
+          var reported = session.nextDiagnosticsFor(
+            newPath,
+            where: (params) => params.diagnostics.isNotEmpty,
+          );
+          session.sendWatchedFileEvents([(newPath, FileChangeType.Created)]);
+
+          var diagnostics = (await reported).diagnostics;
+          expect(diagnostics.single.message, contains('invalid datatype'));
+        },
+      );
```
marcelomendoncasoares, line 171: Suggestion: This is another Given that should ideally comes before the current group - since it asserts the premise that writing an invalid file publishes diagnostics for it.

tools/serverpod_cli/test/integration/language_server/language_server_test.dart
```diff
@@ -0,0 +1,250 @@
...
+          await expectLater(oldUriCleared, completes);
+          var diagnostics = (await newUriReported).diagnostics;
+          expect(diagnostics.single.message, contains('invalid datatype'));
+        },
+      );
+
+      test(
+        'when the model file is deleted on disk '
+        'then empty diagnostics are published for its URI.',
+        () async {
```
marcelomendoncasoares, line 134: Suggestion: Move this test above the previous to keep the scenarios in a "growing" pattern. Combined with the next comment, it would make reading this file flow as:

1. Creating an invalid file publishes issues.
2. Deleting an invalid file removes issues.
3. Renaming a file executes both operations as expected.

tools/serverpod_cli/test/integration/language_server/language_server_test.dart
```diff
@@ -0,0 +1,250 @@
...
+          session.sendWatchedFileEvents([(newPath, FileChangeType.Created)]);
+
+          var diagnostics = (await reported).diagnostics;
+          expect(diagnostics.single.message, contains('invalid datatype'));
+        },
+      );
+
+      test(
+        'when the model directory is renamed on disk and only directory-level '
+        'events are delivered then diagnostics move to the new URIs.',
```
marcelomendoncasoares: Suggestion: Break on the "then" to make the boundary of the clauses read clearer.

```suggestion
        'when the model directory is renamed on disk and only directory-level events are delivered '
        'then diagnostics move to the new URIs.',
```

tools/serverpod_cli/test/integration/language_server/language_server_test.dart
```diff
@@ -0,0 +1,250 @@
...
+          await expectLater(oldUriCleared, completes);
+          var diagnostics = (await newUriReported).diagnostics;
+          expect(diagnostics.single.message, contains('invalid datatype'));
+        },
+      );
+
+      test(
+        'when a watched change event arrives for an open document '
+        'then the editor buffer content stays authoritative.',
+        () async {
```
marcelomendoncasoares, line 246: Suggestion: This seems like a separate "Given", unrelated to the invalid model file (the open document is the target of the scenario).

tools/serverpod_cli/test/integration/language_server/language_server_test.dart
```diff
@@ -0,0 +1,250 @@
...
+            reason:
+                'The open document still holds the invalid model, so its '
+                'errors must not be cleared by the valid content on disk.',
+          );
+        },
+      );
+
+      test(
+        'when the model file is deleted on disk and the document is then '
+        'closed then empty diagnostics are published for its URI.',
```
marcelomendoncasoares: Suggestion: This also seem like a separate "Given" targeting an opened that was deleted. Otherwise it reads pretty similar to the first file delete scenario.

tools/serverpod_cli/lib/src/language_server/language_server.dart
```diff
@@ -124,7 +173,149 @@ Future<void> runLanguageServer() async {
...
+      // Model files anywhere in the workspace.
+      FileSystemWatcher(globPattern: const Either2.t1('**/*.{spy,yaml,yml}')),
+      // Directory renames and deletions only surface as events on the
+      // directory itself, which the model file glob never matches.
+      FileSystemWatcher(
+        globPattern: const Either2.t1('**/*'),
+        kind: const WatchKind(5), // WatchKind.Create | WatchKind.Delete
+      ),
+    ],
+  );
```
marcelomendoncasoares, line 203: Discussion: Should we restrict to `.spy.yaml`/`.spy.yml`? We don't need to support raw `yaml`, and it will prevent us from watching too many unrelated files (pubspecs, CI workflows, analysis options, pubspec overrides, and many other config files).

Also, as it is now, we're watching all files from the repository, regardless of them not being under the loaded project paths (loaded from `_loadServerProject`). Should we restrict it?

marcelomendoncasoares (changes requested):
Two points for discussion and a few suggestions for improvement on the tests.

tools/serverpod_cli/lib/src/language_server/language_server.dart
FXschwartz: We currently support plain `yaml` files and our docs even mention that while we recommend `.spy.yaml` you can use plain `yaml`. Removing support would mean the LSP won't recognize any disc changes to them.

I could scope it: `**/*.{spy,spy.yaml,spy.yml}` anywhere, and `.yaml/.yml` only under `**/lib/src/{models,protocol}/**`. That gets rid of pubspecs, workflows and the rest of the noise. Also narrowing the directory watcher from `**/*` to `**/lib/**`.

On restricting to the loaded project paths, the globs are only relative to the workspace already, and doing it properly means watchers per root, which is basically the multi-project work that we discussed. The relevance filter already drops anything that can't affect the loaded project. And I'd rather tackle the multi project support in a separate PR.

As a side note the start command's file watcher looks like it's already `.spy` only, so plain `.yaml` models don't hot regenerate there. Might be a good separate issue if we care to discuss if we want to push more toward `.spy.yaml` usage or fully support both.

marcelomendoncasoares: If we are to keep .yaml`, we can't restrict to `**/lib/src/{models,protocol}/**` because models can be anywhere below `lib` - still safe to get rid to random `yaml` files with the exception of `analysis_options.yaml` that can be randomly placed to change rules for specific folders. But removing support for bare `.yaml`/`.yml` is indeed a breaking change that we can do separately.

Given we're bumping a major, I think it is a good timing to push the behavioral breaking change and requiring `.spy.yaml` for the Serverpod features to work (`start` watcher, syntax highlight, LSP, etc.). It makes our life a lot easier because we get rid of heuristics to make sure we're not watching/parsing the wrong files.

tools/serverpod_cli/test/integration/language_server/language_server_test.dart
FXschwartz: Good call! Added, creating a valid model on disk now asserts empty diagnostics get published for it.

tools/serverpod_cli/test/integration/language_server/language_server_test.dart
FXschwartz: Restructured

tools/serverpod_cli/test/integration/language_server/language_server_test.dart
FXschwartz: Restructured

tools/serverpod_cli/test/integration/language_server/language_server_test.dart
FXschwartz: Done!

tools/serverpod_cli/test/integration/language_server/language_server_test.dart
FXschwartz: Restructured

tools/serverpod_cli/test/integration/language_server/language_server_test.dart
FXschwartz: Restructured

tools/serverpod_cli/lib/src/language_server/language_server.dart
FXschwartz: Created #5537  to track the `.spy.yaml` requirement.

For this PR I kept `.yaml` support but scoped the watchers. The watchers are now: `**/*.{spy,spy.yaml,spy.yml}` anywhere, `.yaml/.yml` only under `**/lib/src/{models,protocol}/**`, and the directory watcher narrowed from `**/*` to `**/lib/**`.

That cuts out `pubspecs`, workflows, `analysis_options` and the rest of the noise without dropping anything we currently support. Once #5537  lands the yaml watcher can get deleted.

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5520, opened 2026-07-29
feat: Forward shared package models through the owning module's server/client packages

### LINKED ISSUE #5447: Forward shared package models through the owning module's server/client packages
### Problem to Solve

A shared package (`shared_packages` in `config/generator.yaml`) lets a server project or module host models (soon including table models with #5400) that can be consumed without a dependency on the server runtime.

By design, a shared package is an **implementation detail of its owner**: the generated identity (protocol `getModuleName()`, dynamic `__className__` prefixes, `TableDefinition.module`) comes from the owning project or module, so models can move between the owner and the shared package without any wire-visible change.

The problem is that there is no supported way for a third project to consume models or tables from a shared package it does not own. Today the only wiring available is listing the package in the consumer's own `shared_packages`, which silently makes the consumer a **co-owner**: it generates into the package, stamps its own identity, and produces migrations for tables it does not own (see the companion issue on enforcing single ownership). It is also not feasible for modules that are not co-located on the same repository.

The natural fit for cross-project consumption already exists: **modules**. When a module owns a shared package, projects depending on the module should get the shared models and tables transitively, in the same way they get everything else the module provides, without ever referencing the shared package directly.

### Proposal

Make a module's shared package fully forwarded through the module's own server and client packages:

- The module's server and client packages re-export the shared models, so hosts import them from the module as usual. If desired, a shared package of the host project should also be able to import the shared package of the module.
- The module's protocol serves the shared models under the **module's identity**, together with all other normal server/client models. The shared package name never appears in host code or in wire payloads.
- The module's migrations ship the shared tables, so hosts install and version them through the existing `installedModules` mechanism instead of generating their own migrations for foreign tables.

### Use Case

Create a module that has shared logic between server and client and such logic must be available to server/client packages of users' projects. This will be the case for the Offline Sync package, that is implemented as a shared package.

### Alternatives

No real alternative to share logic between server and client in modules.

### Additional context

_No response_

- [x] I want to work on this

marcelomendoncasoares:
Closes: #5447

While implementing this feature, I closed the symmetric hole for a project's own shared package: models used in endpoint signatures were being imported as `package:x/src/generated/...` into `endpoints.dart`, `client.dart` and `serverpod_test_tools.dart`. Now every shared-package reference in generated code goes through the package's public library. Now there are zero private shared-package imports in generated files.

The testing strategy is the heaviest part of this PR, because we needed a few new packages to fully test the feature. The existing `serverpod_test_module` could not be reused, because it is important to keep coverage of regular modules with no shared packages. So a new `serverpod_test_shared_module` was created containing a model that allows us to exercise the most complex interaction of such modules: `dynamic` fields roundtrip.

```yaml
class: SharedModuleTable
table: shared_module_table
database: all
fields:
  name: String
  data: dynamic
```

For a complex coverage, both `serverpod_test_sqlite_server` and `serverpod_test_server` now consume this new module with shared packages. On the `serverpod_test_sqlite_server` side, we validate that both client and server tables are forwarded correctly. On the `serverpod_test_server`, we validate that client tables do not force the creation of a client-side database (since the project lacks own client-side tables). On both, we exercise the `dynamic` field roundtrip.

## Pre-launch Checklist

## Breaking changes

None.


tools/serverpod_cli/lib/src/util/locate_modules.dart
```diff
@@ -143,6 +147,37 @@ Map<dynamic, dynamic> loadConfigFile(File file) {
...
+    if (sharedPath is! String || path.isAbsolute(sharedPath)) {
+      continue;
+    }
+    try {
+      var pubspecFile = File.fromUri(
+        packageSrcRoot.resolve('$sharedPath/pubspec.yaml'),
+      );
+      var pubspec = parsePubspec(pubspecFile);
+      result[pubspec.name] = path.split(sharedPath);
+    } catch (_) {
```
FXschwartz: Does this work for modules installed from pub.dev? The hosted cache layout is flat, so a sibling path like `../serverpod_test_shared_module_shared`, and since this catch just skips the entry, the feature would silently no op. All the tests use path deps, where the path resolves fine.

If I'm reading it right, the consumer would still generate without errors, just with the shared models missing, deep `src/` imports come back and the shared tables drop out of `definition.json`. 

A warning here would help, but maybe the better route is resolving the shared package through `package_config` instead of the relative path, since it's a transitive dep anyway.

Thoughts?

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
```diff
@@ -700,6 +728,14 @@ class LibraryGenerator {
     ]);
   }
 
+  Code _buildDeserializeByClassNameForwarding(String protocolImportPath) {
+    return Code.scope(
+      (a) =>
+          'try{return ${a(refer('Protocol', protocolImportPath))}().deserializeByClassName(data);}'
+          'on FormatException catch(_){}',
```
FXschwartz, line 735: Is `FormatException` too broad a signal here? 

If I'm following the base class right, it's thrown both for "unknown className" and for real data errors inside a class the shared package does handle like a malformed dynamic field via `deserializeDynamicFieldValue`. If that second case happens, I think the error would get swallowed here and we'd fall through to the module's own switch, ending in a "No deserialization found" message for a class that's actually handled. Could be a confusing one to debug, since dynamic fields on shared models are the main use case.

Not sure there's an easy fix given the shared names are unprefixed so the usual prefix check doesn't apply. Maybe a dedicated exception for the unknown className case, like `DeserializationTypeNotFoundException` on the deserialize path?

FXschwartz (changes requested):

tools/serverpod_cli/lib/src/generator/dart/library_generators/library_generator.dart
marcelomendoncasoares: You're right on this one, we do have the possibility of shadowing real `FormatException` occurrences here that should not be swallowed. However, it seems to me that this issue predates this PR, given that we use `FormatException` as fallback on the code generation broadly. To avoid growing this PR even more, I scoped this issue on #5529 for us to tackle separately.

tools/serverpod_cli/lib/src/util/locate_modules.dart
marcelomendoncasoares: This is indeed a real issue that I completely overlooked during implementation... Fixed at https://github.com/serverpod/serverpod/pull/5520/commits/c0eef0f00f72f9bf8d8c08b0f2785b4d493fcd61 by storing the shared package names on a new `manifest.yaml` that will sit inside the `generated` folder (like the `protocol.yaml`). Then, when loading a module, we search for this file and also load models from the installed dart package (the same mechanism we already use for the very modules). 

I intentionally scoped this file as a generic `manifest.yaml` to make it extensible in case we need more metadata in the future - or to collapse the `protocol.yaml` into it.

FXschwartz (approved):
LGTM! 🚀

PR #5521, opened 2026-07-30
feat: Add analytics tracking for used Serverpod features

marcelomendoncasoares:
This PR adds analytics tracking on the CLI to allow us to better understand guide the roadmap, next features and enhancements based on usage data.

## Pre-launch Checklist

## Breaking changes

None.


tools/serverpod_cli/lib/src/analytics/project_metadata_store.dart
```diff
@@ -0,0 +1,147 @@
...
+      p.join(metadataDirectory(serverDir), metadataFileName);
+
+  static Future<ProjectMetadata> loadOrCreate(String serverDir) =>
+      _synchronized(() => _loadOrCreate(serverDir));
+
+  static Future<ProjectMetadata> _loadOrCreate(String serverDir) async {
+    final file = File(metadataFilePath(serverDir));
+    if (await file.exists()) {
+      final json =
+          jsonDecode(await file.readAsString()) as Map<String, dynamic>;
```
FXschwartz: It's a low probability but if this file is ever corrupt like because of a crash, this will throw on every load, and since writes only happen after a successful load, nothing ever rewrites the file. Every capture path swallows the throw, so analytics for the checkout go silent permanently.

Could we catch the parse failure and fall through to the recreate branch? Losing the counters beats losing all future events. 

Also noticed the corrupt-metadata test in `session_analytics_test.dart` only covers the disabled path, so this case isn't exercised anywhere.

tools/serverpod_cli/lib/src/commands/generate.dart
```diff
@@ -289,6 +302,15 @@ Future<GenerateResult> analyzeAndGenerate({
     );
     log.debug(incrementalCodeGenerationComplete);
   }
+
+  await reportGenerateAnalytics(
```
FXschwartz, line 306: Looks like this is skipped whenever `performGenerate` throws.  `log.progress` rethrows runner errors, so a codegen crash or I/O error never produces an event. `generation_succeeded: false` ends up covering only validation failures that completed the pipeline, while crashes seem like something we'd also like to track and falls in line with the "where does generation fail" question in analytics.md.

Could we catch here and still report the attempt?

tools/serverpod_cli/lib/src/analytics/project_identity.dart
```diff
@@ -0,0 +1,144 @@
...
+        if (!p.isAbsolute(commonDir)) {
+          commonDir = p.normalize(p.join(gitDir, commonDir));
+        }
+        return commonDir;
+      default:
+        return null;
+    }
+  }
+
+  static String? _findDotGit(String startDir) {
```
FXschwartz, line 93: I haven't tested this but I don't believe `serverpod create` initializes a git repo. And I don't see any kind of ceiling check here which would cause a fresh project to resolve through whatever ancestor repo happens to exist until the user inits their own.

Then the metadata would land in that repo's .git

I can't think of a perfect ceiling that keeps monorepos working, but refusing a .git at or above $HOME would kill the worst case. Or if that doesn't work this might be worth a note in analytics.md as a known identity tradeoff.

FXschwartz (changes requested):

tools/serverpod_cli/test/analytics/session_analytics_test.dart
```diff
@@ -0,0 +1,200 @@
...
+
+      final properties = recording.eventNamed('cli.flutter_launch');
+      expect(properties['device_category'], 'desktop');
+      expect(properties['device_platform'], 'macos');
+      expect(properties['is_relaunch'], isFalse);
+    },
+  );
+
+  test(
+    'Given analytics are disabled and the project metadata is corrupt, '
```
Crazelu, line 214: We have another test for when analytics is enabled and project metadata is corrupt.

tools/serverpod_cli/test/analytics/project_metadata_store_test.dart
```diff
@@ -0,0 +1,157 @@
...
+          d.file('pubspec.yaml', 'name: myapp_server\n'),
+        ]),
+      ]).create();
+
+      serverDir = p.join(
+        d.sandbox,
+        'metadata_project_a',
+        'myapp_server',
+      );
+      metadata = await ProjectMetadataStore.loadOrCreate(serverDir);
```
Crazelu: Should this not only be in the `when` clause?

tools/serverpod_cli/test/analytics/project_identity_test.dart
```diff
@@ -0,0 +1,183 @@
...
+        p.join(d.sandbox, 'clone_https', 'app_server'),
+      );
+
+      expect(sshId, isNotNull);
+      expect(sshId, httpsId);
+    },
+  );
+
+  test(
+    'Given a git checkout without a remote, '
```
Crazelu, line 360: Should we not have a persistent anonymously generated project id fallback if git remote (or git entirely) is not available?

tools/serverpod_cli/lib/src/analytics/project_metadata_store.dart
marcelomendoncasoares: Good catch, we should indeed make it safer. I’ll add a test that pins this scenario and a recover from decode/deserialization failures by recreating the metadata.

tools/serverpod_cli/test/analytics/project_identity_test.dart
marcelomendoncasoares: This already falls back to the persisted random checkout ID when Git or its remote is unavailable. What we don't have when remote is not available is a "durable ID": an ID that we expect to repeat on all machines/checkouts that target the same repository. I'll add an end-to-end test to make that behavior explicit.

tools/serverpod_cli/lib/src/commands/generate.dart
marcelomendoncasoares: This was actually intentional. Given our purpose with the analytics is feature tracking, I preferred skipping a partial payload than catching and reporting exceptions. The `generation_succeeded` was not intended for actual crashes - that we expect to be reported through GitHub issues by the users. Going this path also prevents us from worrying about bleeding private information through exceptions and stacktraces. If we later need more insight into common handled failure paths, then we can add explicit allowlisted categories rather than treating every exception as an analytics event.

tools/serverpod_cli/lib/src/analytics/project_identity.dart
marcelomendoncasoares: It doesn't initialize git indeed - although, as we have discussed, it harmlessly could. And, even then, we could never expect this premise to hold for any repository: users might copy from other projects, create the folders manually, remove the `.git` intentionally, etc. So the process must be safe regardless of `.git` being eventually initialized by `serverpod create`.

I'll add the other boundary detection we also discussed: checking for a workspace `pubspec.yaml` together with the `.git` search, limited to the user `$HOME`.

tools/serverpod_cli/test/analytics/project_metadata_store_test.dart
marcelomendoncasoares: You're right, it should! I'll move it down to the test.

tools/serverpod_cli/test/analytics/project_identity_test.dart
marcelomendoncasoares: Added the test at https://github.com/serverpod/serverpod/pull/5521/commits/ac2218df248241045d6ff95b6fd4626e1a9c9e78.

tools/serverpod_cli/test/analytics/project_metadata_store_test.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5521/commits/9f88ce3771d265f2bc93ee4fae122c55553072c8.

tools/serverpod_cli/lib/src/analytics/project_metadata_store.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5521/commits/d807082dd1b3ae936db9f47f2ea52a559d447133.

tools/serverpod_cli/lib/src/analytics/project_identity.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5521/commits/548163045d02c15c3257614b2d7af07ae6cbe740. While doing it, I realized that we were reusing the same ID for all servers on a given repo, which would make monorepos with multiple servers generate alternative analytics payloads with different feature sets. I solved this by adding the server path before hashing the ID at https://github.com/serverpod/serverpod/pull/5521/commits/aa5d875e6b779cba096d39bb8e84ea3a42dcc98e.

FXschwartz (approved):
Approved!

PR #5535, opened 2026-08-03
fix: Serve flutter app config under root when project is created with webapp enabled only

Crazelu:
Previously, projects created with Flutter web app only enabled served `assets/assets/config.json` under `/app` which causes dynamically injected config to fail. This config is now served under root.

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tests/bootstrap_project/test_perform_create/web_test.dart
+++ b/tests/bootstrap_project/test_perform_create/web_test.dart
@@ -625,7 +625,13 @@ void main() {
         final testConfig = File(
           p.join(project.serverDir, 'config', 'test.yaml'),
         );
-        final config = await testConfig.readAsString();
+        // Templates checked out on Windows use CRLF. Normalize before matching
+        // the multiline YAML blocks below so the replacements work on every
+        // platform.
+        final config = (await testConfig.readAsString()).replaceAll(
+          '\r\n',
+          '\n',
+        );
         await testConfig.writeAsString(
           config
               .replaceFirst(
--- a/tests/bootstrap_project/test_perform_create/web_test.dart
+++ b/tests/bootstrap_project/test_perform_create/web_test.dart
@@ -716,9 +716,6 @@ void main() {
         },
       );
     },
-    skip: Platform.isWindows
-        ? 'Windows does not support postgres in github actions'
-        : null,
   );
 
   group(
@@ -738,7 +735,7 @@ void main() {
       setUpAll(() async {
         startProjectProcess = await startProcessAndWaitForKeywords(
           'dart',
-          ['bin/main.dart', '--apply-migrations'],
+          ['bin/main.dart'],
           workingDirectory: project.serverDir,
           keywords: ['Webserver listening on'],
         );
@@ -799,7 +796,7 @@ void main() {
           startProjectProcess.kill();
           startProjectProcess = await startProcessAndWaitForKeywords(
             'dart',
-            ['bin/main.dart', '--apply-migrations'],
+            ['bin/main.dart'],
             workingDirectory: project.serverDir,
             keywords: ['Webserver listening on'],
           );
@@ -827,8 +824,5 @@ void main() {
         );
       });
     },
-    skip: Platform.isWindows
-        ? 'Windows does not support postgres in github actions'
-        : null,
   );
 }
--- a/tests/bootstrap_project/test_perform_create/web_test.dart
+++ b/tests/bootstrap_project/test_perform_create/web_test.dart
@@ -716,6 +716,9 @@ void main() {
         },
       );
     },
+    skip: Platform.isWindows
+        ? 'Windows does not support postgres in github actions'
+        : null,
   );
 
   group(
--- a/tests/bootstrap_project/test_perform_create/web_test.dart
+++ b/tests/bootstrap_project/test_perform_create/web_test.dart
@@ -688,9 +688,10 @@ void main() {
           };
       });
 
-      tearDownAll(() {
+      tearDownAll(() async {
         client?.close(force: true);
-        startProjectProcess?.kill();
+        final process = startProjectProcess;
+        if (process != null) await stopProcess(process);
       });
 
       test(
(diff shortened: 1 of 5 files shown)


tests/bootstrap_project/test_perform_create/web_test.dart
```diff
@@ -576,6 +595,132 @@ void main() {
...
+          expect(apiUrl.scheme, 'https');
+          expect(apiUrl.host, 'api.managed.host');
+          expect(apiUrl.port, 443);
+          expect(responseBody, isNot(contains('invalid.url')));
+        },
+      );
+    },
+    skip: Platform.isWindows
+        ? 'Windows does not support postgres in github actions'
+        : null,
```
marcelomendoncasoares: Question: By overriding the config without the `database` keyword, running on Windows should not be a problem, right?

tests/bootstrap_project/test_perform_create/web_test.dart
```diff
@@ -576,6 +595,132 @@ void main() {
...
+          final response = await request.close();
+          final responseBody = await utf8.decoder.bind(response).join();
+          final appConfig = jsonDecode(responseBody) as Map<String, dynamic>;
+          final apiUrl = Uri.parse(appConfig['apiUrl'] as String);
+
+          expect(response.statusCode, HttpStatus.ok);
+          expect(apiUrl.scheme, 'https');
+          expect(apiUrl.host, 'api.managed.host');
+          expect(apiUrl.port, 443);
+          expect(responseBody, isNot(contains('invalid.url')));
```
marcelomendoncasoares, line 703: Question: Shouldn't it be the opposite? Why is the configuration shadowing the config asset value, since it is meant to override the config? My understanding is that this file should be served as it is, and fetched on the Flutter app by the `getServerUrl()` to build the `Client` with the injected config (when deploying to Cloud, for example). Or am I missing something on this fix?

marcelomendoncasoares (commented):
Only two questions on the fix.

tests/bootstrap_project/test_perform_create/web_test.dart
Crazelu: It shouldn't be a problem

tests/bootstrap_project/test_perform_create/web_test.dart
Crazelu: `assets/config.json` file is fetched from the server when the web app is hosted. The documentation on `getServerUrl` mentions that it allows the server to change the API URL at runtime.

So on cloud, the API URL is swapped by `AppConfigRoute` based on API server configurations.
The issue is that the file is served under the root when flutter only app is deployed but the path wrongly included `/app` which pointed to the actual asset that hardcodes localhost.

This test uses a loopback http client to mirror what happens when a production API server is hit for that file.
To reproduce the issue in test, change

```dart
final request = await client!.getUrl(
            Uri.http(
              'app.managed.host',
              '/assets/assets/config.json',
            ),
          );
```

to 

```dart
final request = await client!.getUrl(
            Uri.http(
              'app.managed.host',
              '/app/assets/assets/config.json',
            ),
          );
```

and it will serve the original config file containing `invalid.url`.

tests/bootstrap_project/test_perform_create/web_test.dart
Crazelu: The issue here was a combination of windows MAX_PATH and windows path normalization, not really database since database is disabled for those tests.

Those are fixed now.

tests/bootstrap_project/test_perform_create/web_test.dart
```diff
@@ -590,23 +740,49 @@ void main() {
...
+      Future<void> startPod() async {
+        final serverOutput = StringBuffer();
+        final webServerStartedPattern = RegExp(
+          r'Webserver listening on http://localhost:(\d+)',
+        );
+        int? nextWebPort;
+        final process = await startProcessAndWaitForKeywords(
           'dart',
-          ['bin/main.dart', '--apply-migrations'],
+          ['bin/main.dart', '--mode=test'],
```
marcelomendoncasoares: Question: It might not affect these tests, but removing the `--apply-migrations` can lead to the server not starting properly, right? Seems an optimization that is not worthy and can introduce flakiness later.

tests/bootstrap_project/test_perform_create/web_test.dart
```diff
@@ -641,6 +817,12 @@ void main() {
 
       group('Given the Flutter web app is built and the pod is restarted', () {
         setUp(() async {
+          final runningProcess = startProjectProcess;
+          if (runningProcess != null) {
+            await stopProcess(runningProcess);
+            startProjectProcess = null;
+          }
+
```
marcelomendoncasoares: Suggestion: This test is actually out-of-pattern. This "Given" is inside the parent group that is also a "Given". Might be worth splitting the fix on this test to a separate PR to avoid bundling unrelated broad test changes. There is also an interaction between this `setUp` and the start process created on the `setUpAll` that can lead to issues if we later introduce more groups on the same level as this one.

marcelomendoncasoares (changes requested):
Just two suggestions on the later changes to the other tests to work on Windows. I think it is worth splitting them to a separate PR to avoid mixing up the context.

tests/bootstrap_project/test_perform_create/web_test.dart
Crazelu: Removing `--apply-migrations` is correct here since the created project has database disabled. It is harmless to add it; a warning will just be logged. But this doesn't affect the server starting properly.

Crazelu: I took another look. It is not harmless like I said.
It'll actually cause the pod to exit after startup. So removing it here is correct and the right way to avoid flakiness

tests/bootstrap_project/test_perform_create/web_test.dart
Crazelu: Fixed in [9678c82](https://github.com/serverpod/serverpod/pull/5535/commits/9678c82afad24c48d5e722e6f4c16bb3b5cb8262)

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5538, opened 2026-08-04
feat: Allow overriding the default cache header for `FlutterRoute` from env vars

### LINKED ISSUE #5527: Expose an environment variable to inject default cache control for Serverpod Cloud CDN
Now that Serverpod Cloud [supports CDN caching](https://docs.serverpod.dev/cloud/concepts/cdn#enable-caching-for-your-app) with active cache invalidation on deploy, we need to expose a safe way for users to to have this enabled on their deployed projects without affecting projects deployed elsewhere.

To achieve this, we should expose an environment variable named `SERVERPOD_WEB_SERVER_FLUTTER_CACHE_CONTROL` that can receive raw `Cache-Control` header strings like:

| Use case | Value |
|----------|-------|
| CDN recommended (default opt-in) | `public, max-age=60, s-maxage=86400` |
| Current FlutterRoute default | `private, no-cache` |
| Revalidate every request, CDN can store | `no-cache, public, s-maxage=86400` |

Since `Relic` already ships a parser on `CacheControlHeader`, on Dart we can do:

```dart
final header = CacheControlHeader.parse([
  'public, max-age=60, s-maxage=86400',  // or extracted from the environment variable
]);
// header.publicCache == true
// header.maxAge == 60
// header.sMaxAge == 86400
```

And wire this value as a default inside the `FlutterRoute` when no `cacheControlFactory` was supplied by the user. A missing environment variable would still default to `StaticRoute.privateNoCache()`.

Note that **the feature is only the environment variable**, without altering default values. The variable will be used on Cloud deployments to inject the default that is compatible with CDN, which is overridable by the user when setting from Dart directly.

marcelomendoncasoares:
Closes: #5527

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod/lib/src/web_server/routes/flutter_route.dart
+++ b/packages/serverpod/lib/src/web_server/routes/flutter_route.dart
@@ -4,6 +4,9 @@ import 'dart:io';
 import 'package:path/path.dart' as path;
 import 'package:serverpod/serverpod.dart';
 
+const _flutterCacheControlEnvironmentVariable =
+    'SERVERPOD_WEB_SERVER_FLUTTER_CACHE_CONTROL';
+
 /// Route for serving Flutter web applications.
 ///
 /// Combines static file serving with automatic fallback to index.html for
@@ -21,9 +24,9 @@ import 'package:serverpod/serverpod.dart';
 ///
 /// ## About Caching
 ///
-/// By default, all files are served with `private, no-cache` headers so the
-/// browser always revalidates with the server via ETag. This is the only safe
-/// default because:
+/// Unless `SERVERPOD_WEB_SERVER_FLUTTER_CACHE_CONTROL` is set, all files are
+/// served with `private, no-cache` headers so the browser always revalidates
+/// with the server via ETag. This is the only safe built-in default because:
 ///
 /// - Flutter's service worker is deprecated and will be removed
 ///   (see [flutter#156910](https://github.com/flutter/flutter/issues/156910)).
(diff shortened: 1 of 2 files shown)


packages/serverpod/lib/src/web_server/routes/flutter_route.dart
```diff
@@ -91,9 +98,18 @@ class FlutterRoute extends Route {
...
-           cacheControlFactory ?? StaticRoute.privateNoCache(),
+           cacheControlFactory ?? _cacheControlFactoryFromEnvironment(),
        super(methods: {Method.get, Method.head});
 
+  static CacheControlFactory _cacheControlFactoryFromEnvironment() {
+    final cacheControl =
+        Platform.environment[_flutterCacheControlEnvironmentVariable];
+    if (cacheControl == null) return StaticRoute.privateNoCache();
+
+    final parsedCacheControl = CacheControlHeader.parse([cacheControl]);
```
Crazelu, line 109: If there are parsing errors, that would break the UX. Do we want that to be explicit? Should there be a fallback here?

marcelomendoncasoares: This was intentional indeed. Because we're already talking about caching, which can malfunction quietly and only impose performance penalties. If the override mechanism can also fail silently, the feature becomes unreliable. So it is better to fail loudly upfront - which will be a deployment error fixed before the app is live - than requiring the user to inspect the container and served app to be sure that the cache is working.

Crazelu (approved):
🚀

PR #5539, opened 2026-08-04
fix: Generate detach/detachRow for named list relations without order dependence

### LINKED ISSUE #4868: Missing `detach`/`detachRow` for nullable named list relations when object side has unnamed relation
### Describe the bug

For named list relations where the foreign row uses the implicit object resolution path, `ListRelationDefinition` could be resolved with `nullableRelation: false` even when the object side is optional. Generated table repositories then omitted `detach`/ `detachRow` for list owners (examples from the `serverpod_test_server` are `Student` and `Order`).

The object field’s `ObjectRelationDefinition` does not set `RelationDefinition.name` (only the FK field carries the name). List-side code matched fields only by `relation.name`, so it matched the non-nullable id field but not the optional object field, producing wrong nullability and missing detach methods. Reordering of models (e.g. moving YAML), watch mode, or incremental generation could expose the bug.

This was perceived during tests on the new `serverpod start --watch` command, when the reordering of models due to cached runs not analyzing all models would result in the missing detach methods for the `CourseUuid` model - which luckily had them, despite being in the same condition as `Student` and `Order`.

### To reproduce

1. Define models with `List<Child>?, relation(name=…)` on the owner and `Parent?, relation(name=…)` on the child (optional object), implicit FK + object field on the child.
2. Run `serverpod generate` (optionally `--watch` or after moving model files).
3. Inspect generated `*Repository` for the list owner.

**Observed:** `detach` / `detachRow` missing.

### Expected behavior

Nullable list relations with optional foreign object sides should set `nullableRelation` correctly and generate `detach` / `detachRow` on the owning repository.

### Library version

3.4.4

### Platform information

WSL

### Additional context

Fix direction: pair object fields (null relation name) with the named FK on the same class when resolving list relations; optionally keep topological class order so list elements are resolved before list owners. Staleness/`skipStalenessCheck` for generate vs watch is a separate concern from nullability but affects whether codegen runs when expected.

- [x] I want to work on this

Crazelu:
`detach/detachRow` methods were previously conditionally generated depending on the order by which models are analyzed.

- **Child class analyzed first**: its object field's relation has already been resolved into an ObjectRelationDefinition which carries no name so the name-based match finds the injected FK id field instead. Its type nullability correctly reflects the optional keyword for a required FK.

- **Owner analyzed first**: the FK field doesn't exist yet, so the match finds the child's UnresolvedObjectRelationDefinition object field, and nullability is taken from the object field's Dart type, which is essentially always nullable.

The models that had generated `detach/detachRow` methods were silently broken as they called `copyWith` and passed `null` for a non-nullable field which doesn't drop the relation.

The fix is to derive nullability from the foreign key regardless of the order by which models are analyzed.

Fixes #4868 

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tools/serverpod_cli/test/generator/dart/server_code_generator/database_repository_class_list_relation_additions_test.dart
+++ b/tools/serverpod_cli/test/generator/dart/server_code_generator/database_repository_class_list_relation_additions_test.dart
@@ -560,4 +560,130 @@ void main() {
       });
     },
   );
+
+  group(
+    'Given a class with table name and explicit non-nullable list relation field, '
+    'when generating code',
+    () {
+      var models = [
+        ModelClassDefinitionBuilder()
+            .withClassName(testClassName)
+            .withFileName(testClassFileName)
+            .withTableName('example_table')
+            .withListRelationField(
+              'people',
+              'Person',
+              'organizationId',
+              nullableRelation: false,
+            )
+            .build(),
+      ];
+
+      var codeMap = generator.generateSerializableModelsCode(
+        models: models,
+        config: config,
+      );
+
+      var compilationUnit = parseString(
+        content: codeMap[expectedFilePath]!,
+      ).unit;
+
+      var repositoryClass = CompilationUnitHelpers.tryFindClassDeclaration(
+        compilationUnit,
+        name: '${testClassName}Repository',
+      );
+
+      group(
+        'then the class name ${testClassName}Repository',
+        () {
+          test('has a final attach field', () {
+            var field = CompilationUnitHelpers.tryFindFieldDeclaration(
+              repositoryClass!,
+              name: 'attach',
+            );
+
+            expect(
+              field?.toSource(),
+              'final attach = const ${testClassName}AttachRepository._();',
+              reason: 'Missing final attach field.',
+            );
+          });
+
+          test('has no detachRow field', () {
+            var field = CompilationUnitHelpers.tryFindFieldDeclaration(
+              repositoryClass!,
+              name: 'detachRow',
+            );
+
+            expect(
+              field,
+              isNull,
+              reason:
+                  'The field detachRow was found but was expected to not exist.',
+            );
+          });
+
+          test('has no detach field', () {
+            var field = CompilationUnitHelpers.tryFindFieldDeclaration(
+              repositoryClass!,
+              name: 'detach',
+            );
+
+            expect(
+              field,
+              isNull,
+              reason:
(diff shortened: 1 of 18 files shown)


FXschwartz (approved):
Only note here is whether it's worth adding a breaking change since now any usages that produced the broken detach will now get a compile error instead of it just silently failing? Maybe worth a note in the changelog?

Otherwise LGTM!

tools/serverpod_cli/test/generator/dart/server_code_generator/database_repository_class_list_relation_additions_test.dart
marcelomendoncasoares, line 1: Suggestion: We should add a similar test to the client and shared generator tests as well.

tools/serverpod_cli/test/generator/dart/server_code_generator/database_repository_detach_resolution_order_test.dart
marcelomendoncasoares, line 1: Suggestion: Same as above, we should add a test for client and shared.

marcelomendoncasoares (changes requested):
Just two requests on test coverage for client and shared generators.

marcelomendoncasoares:
> Only note here is whether it's worth adding a breaking change since now any usages that produced the broken detach will now get a compile error instead of it just silently failing? Maybe worth a note in the changelog?
> 
> Otherwise LGTM!

Not sure if we need a warning, since the existing methods would produce a runtime error. If we haven't received reports on this until now, it is probably due to users not using or ignoring after facing errors.

tools/serverpod_cli/test/generator/dart/server_code_generator/database_repository_detach_resolution_order_test.dart
Crazelu: Done in [2ba68e4](https://github.com/serverpod/serverpod/pull/5539/commits/2ba68e4d33bc14da78fc9d6a3f87ca1fadb1409f)

tools/serverpod_cli/test/generator/dart/server_code_generator/database_repository_class_list_relation_additions_test.dart
Crazelu: Done in [2ba68e4](https://github.com/serverpod/serverpod/pull/5539/commits/2ba68e4d33bc14da78fc9d6a3f87ca1fadb1409f)

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5542, opened 2026-08-04
fix: Make generated Dart code `dart format` clean

### LINKED ISSUE #5473: serverpod generate: enum terminator `;` emitted on its own line — generated enums are never `dart format`-clean
### Describe the bug

`serverpod generate` emits the terminating `;` of a generated enum on its own line:

```dart
enum ChatRole implements _i1.SerializableModel {
  user,
  assistant,
  system
  ;
```

No `dart format` configuration reproduces this. It is dirty under **both** `--trailing-commas=preserve` and `--trailing-commas=automate`, so a project cannot reconcile it by configuring its formatter — every generated enum is permanently `dart format`-dirty.

In our project this affects **81 of 295** generated server files (every file declaring an enum; verified — 81/81 contain an `enum` declaration, 0 do not).

This appears distinct from the trailing-commas difference in the rest of the generated output (filed separately), which *is* explainable by formatter settings. The `;` placement suggests it is concatenated onto the source after formatting rather than being part of the formatted output.

### To reproduce

1. Create `lib/src/models/chat_role.spy.yaml`:

```yaml
enum: ChatRole
serialized: byName
values:
  - user
  - assistant
  - system
```

2. Run `dart run serverpod_cli generate`

3. Inspect `lib/src/generated/chat_role.dart`:

```dart
enum ChatRole implements _i1.SerializableModel {
  user,
  assistant,
  system
  ;
```

4. Run `dart format --output=none --set-exit-if-changed lib/src/generated/chat_role.dart` → exits 1.

Both settings collapse it identically:

```
$ dart format --output=show --trailing-commas=preserve lib/src/generated/chat_role.dart
...
  system;

$ dart format --output=show --trailing-commas=automate lib/src/generated/chat_role.dart
...
  system;
```

The diff `dart format` wants:

```diff
-  system
-  ;
+  system;
```

### Expected behavior

Generated enums are `dart format`-clean, i.e. the terminator is emitted as:

```dart
enum ChatRole implements _i1.SerializableModel {
  user,
  assistant,
  system;
```

### Library version

- `serverpod_cli` 4.0.0-beta.0
- `serverpod` / `serverpod_test` / `serverpod_lints` 4.0.0-beta.0
- `dart_style` 3.1.5 (pub-resolved) — note the SDK's bundled 3.1.6 produces identical output, so this is not a dart_style version skew

### Platform information

- Dart SDK 3.12.2 (stable) on `windows_x64`
- Flutter 3.44.6 (stable)
- Windows 11

### Additional context

Impact is on projects that verify formatting in CI. Because generated enums can never be `dart format`-clean at any setting, such projects must either run a normalization pass after every `serverpod generate` (an invariant held only by discipline — ours lapsed once and turned 563 files red in CI while nothing was actually broken), or exclude generated paths from formatting checks entirely.

`dart format` has no `--exclude` flag, so the exclusion is awkward to express and makes a whole-package `dart format` actively harmful in such a project.

Happy to provide the full generated file or additional samples if useful.

### LINKED ISSUE #5474: serverpod generate: honor the project’s analysis_options.yaml `formatter:` settings for generated code
### Problem to Solve

`serverpod generate` formats its output with trailing commas **preserved**, and does not read the project's `analysis_options.yaml` `formatter:` block.

Measured across all **295** generated server files in our project (`serverpod_cli` 4.0.0-beta.0, Dart 3.12.2):

| Check | Dirty files |
|---|---|
| `dart format` (default, `trailing_commas: automate`) | **288 / 295** |
| `dart format --trailing-commas=preserve` | **81 / 295** |

The **214** files that are clean under `preserve` are *byte-identical* to `dart format --trailing-commas=preserve` output — confirming generated code is dart_style with `preserve`, regardless of what the project configures.

(The residual 81 are all enums, caused by a separate defect — #5473. Those are dirty under *every* setting.)

So a project that pins:

```yaml
# analysis_options.yaml
formatter:
  page_width: 80
  trailing_commas: automate
```

can never have `dart format`-clean generated code. Note this is not a version skew: the SDK's `dart_style` 3.1.6 and the pub-resolved 3.1.5 produce identical output — only the *setting* differs.

The consequence is that every single `serverpod generate` must be followed by a manual `dart format` normalization pass, forever, held together purely by discipline. When one run in our project skipped it, CI went red across 563 files while nothing was actually broken — pure noise, and it cost real time to diagnose.

### Proposal

When formatting generated output, read the target package's `analysis_options.yaml` `formatter:` block (`page_width`, `trailing_commas`) and pass those settings to `DartFormatter`, so generated code matches the project's canonical style.

Failing that: **document** that generated code is formatted with fixed settings and is not expected to be `dart format`-clean under a differing project configuration. Right now this behavior is undocumented — I checked the Serverpod docs and found no mention of formatting of generated code anywhere — so projects discover it via a red CI run rather than a deliberate decision.

### Use Case

Any project that (a) commits generated code, which Serverpod's layout encourages, and (b) verifies formatting in CI — a common Dart setup. Today those two facts are in permanent conflict unless the project happens to use dart_style's `preserve` behavior.

Concretely: we run `dart format --output=none --set-exit-if-changed` as a CI gate. With `formatter: trailing_commas: automate` pinned, that gate can never pass over `lib/src/generated/**`.

### Alternatives

1. **Normalize after every generate** (what we did until now) — an invariant held only by discipline. It lapsed exactly once and produced a 563-file red CI run. Rejected: unreliable by construction.
2. **Adopt `trailing_commas: preserve` project-wide to match serverpod** — does not work. The 81 enum files (#5473) are dirty under `preserve` too, so generated code stays dirty, and it means degrading hand-written style to match a generator's default.
3. **Exclude generated paths from formatting checks** (what we now do). Works, but `dart format` has no `--exclude` flag, so it requires a git-pathspec filter, and it makes a plain `dart format <package>` actively harmful — it silently re-dirties the generated tree.
4. **Stop committing generated code** — viable but expensive: every CI job grows a generate step and every clone needs codegen before the IDE resolves the generated client.

### Additional context

Related: #5473 (generated enum `;` placement) — that one is dirty under *all* settings, so fixing this issue alone would not make generated output fully `dart format`-clean. Fixing both would.

Happy to test a patch against a reasonably large real-world schema (295 generated files) if that's useful.

marcelomendoncasoares:
Fixes: #5473
Fixes: #5474

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/generator/dart/client_code_generator.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/client_code_generator.dart
@@ -84,8 +84,16 @@ class DartClientCodeGenerator extends CodeGenerator {
         'lib',
         'migrations',
         'migration_registry.dart',
-      ])] = _migrationDartEmitter
-          .emitPlaceholderRegistry();
+      ])] = _migrationDartEmitter.emitPlaceholderRegistry(
+        formatter: GeneratedDartFormatters.of(
+          p.joinAll([
+            ...config.clientPackagePathParts,
+            'lib',
+            'migrations',
+            'migration_registry.dart',
+          ]),
+        ),
+      );
     }
     return files;
   }
--- a/tools/serverpod_cli/lib/src/generator/dart/client_code_generator.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart/client_code_generator.dart
@@ -5,6 +5,7 @@ import 'package:serverpod_cli/src/generator/code_generator.dart';
 import 'package:serverpod_cli/src/generator/dart/library_generators/library_generator.dart';
 import 'package:serverpod_cli/src/generator/dart/library_generators/model_library_generator.dart';
 import 'package:serverpod_cli/src/generator/dart/library_generators/util/model_generators_util.dart';
+import 'package:serverpod_cli/src/generator/dart_formatters.dart';
 import 'package:serverpod_cli/src/migrations/client_side/client_migration_dart_emitter.dart';
 
 /// A [CodeGenerator] that generates the client side dart code of a
@@ -33,15 +34,18 @@ class DartClientCodeGenerator extends CodeGenerator {
       config,
     );
 
-    return {
-      for (var entry in modelAllocatorContext.entries)
-        entry.model.getFullFilePath(
-          config,
-          serverCode: false,
-        ): clientSideGenerator
-            .generateModelLibrary(entry.model)
-            .generateCode(allocator: entry.allocator),
-    };
+    var codeMap = <String, String>{};
+    for (var entry in modelAllocatorContext.entries) {
+      var path = entry.model.getFullFilePath(config, serverCode: false);
+      codeMap[path] = clientSideGenerator
+          .generateModelLibrary(entry.model)
+          .generateCode(
+            allocator: entry.allocator,
+            formatter: GeneratedDartFormatters.of(path),
+          );
+    }
+
+    return codeMap;
   }
 
   @override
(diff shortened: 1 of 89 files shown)


tools/serverpod_cli/lib/src/generator/dart/client_code_generator.dart
```diff
@@ -55,11 +59,23 @@ class DartClientCodeGenerator extends CodeGenerator {
...
-      p.joinAll([...config.generatedDartClientModelPathParts, 'client.dart']):
-          clientClassGenerator.generateClientEndpointCalls().generateCode(),
+      protocolPath: clientClassGenerator.generateProtocol().generateCode(
+        formatter: GeneratedDartFormatters.of(protocolPath),
+      ),
+      clientPath: clientClassGenerator
+          .generateClientEndpointCalls()
+          .generateCode(formatter: GeneratedDartFormatters.of(clientPath)),
     };
     if (protocolDefinition.models.hasHostClientDatabaseTables &&
```
FXschwartz, line 80: Looks like the client migration files under `lib/migrations/` don't get this treatment. `ClientMigrationDartEmitter._formatSource` is still hard-coded to language version 3.8 / preserve / width 80, and the same CI gate covers those files.

FXschwartz (changes requested):

tools/serverpod_cli/lib/src/generator/dart/client_code_generator.dart
marcelomendoncasoares: Good catch! I completely ignored that client-side migrations are also Dart code and only worried about the generate path. Fixed at https://github.com/serverpod/serverpod/pull/5542/commits/53577d69f95ec76c44e75b0795ceeb4109b2870f.

FXschwartz (approved):
LGTM!

PR #5543, opened 2026-08-04
fix: Squash the most frequent flaky tests

marcelomendoncasoares:
Fixes 4 of the most common flaky tests that we're having after the CI improvements.

### Future-call claim synchronization

The test assumed that a claim existed 100 ms after starting the future-call manager. Since starting the manager only schedules asynchronous scanning, slower CI workers could reach the assertion first. The fixed test waits for the future call to begin invocation, which can only happen after its claim has been inserted.

### PostgreSQL transaction error normalization

Depending on transaction timing, PostgreSQL reported the serialization conflict either during an update or during commit. Update errors passed through the existing query adapter and became `DatabaseQueryException`, while commit errors bypassed that adapter and escaped raw because the transaction boundary performed no error translation. The fix added that translation around `runTx()` (while still preserving SQLSTATE `40001`).

### Windows SQLite cleanup retry

Windows may briefly retain SQLite worker handles after the database session has closed, preventing immediate deletion of the temporary database directory. Cleanup now retries only transient `PathAccessException` failures with bounded backoff and still reports persistent failures.

### Generate-watch completion sequencing

Generate-watch emits an incremental-generation completion followed by a watcher-ready initial-generation completion during startup, and the test treated both as equivalent. Because the stream helper cleared its state before its settling delay, the second marker could remain queued and satisfy the next file-change wait, leaving assertions one generation behind. Clearing the state after the delay coalesces those markers and makes each subsequent wait require a new event.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
+++ b/packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
@@ -866,17 +866,27 @@ class PostgresDatabaseConnection
     );
 
     final connection = await _postgresConnection;
-    return connection.runTx<R>(
-      (ctx) {
-        var transaction = _PostgresTransaction(
-          ctx,
-          session,
-          this,
-        );
-        return transactionFunction(transaction);
-      },
-      settings: pgTransactionSettings,
-    );
+    try {
+      return await connection.runTx<R>(
+        (ctx) {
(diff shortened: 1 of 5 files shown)


FXschwartz (approved):
LGTM! Down with the flaky tests!

PR #5545, opened 2026-08-05
feat: Implement device selection for VS Code launch

Crazelu:
After the changes to the VS Code launch configurations to attach instead of launch and with the introduction of multiple flutter app support to `serverpod start`, there's been no way for users to pick a device when running from VS Code with F5. The device selection is driven by configuration in the server pubspec and defaults to web-server when not set.

One downside of this default device is that it doesn't support debugging out of the box.

This PR brings back device selection when running from VS Code to retain the IDE's UX and also run on a debuggable target. Device selection prompt is only shown when there is one app configured. This also covers when no app is configured in the server's pubspec but one is synthesized. If a device has already been selected in the IDE, the prompt is not shown again for subsequent launches.

The prompt will not appear when running projects with multiple flutter apps configured. Those still require manual configuration in pubspec for the target devices. 

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start/flutter_app_manager.dart
+++ b/tools/serverpod_cli/lib/src/commands/start/flutter_app_manager.dart
@@ -53,9 +53,14 @@ class FlutterAppManager {
     required this.serverPackageDirectoryPathParts,
     required this.projectName,
     required this.launchFlutterApp,
+    this.environmentOverrideForTesting,
     this.flutterExecutableForTesting,
     this.argsOverrideForTesting,
   });
+
+  /// Test-only override for process environment.
+  /// Decides whether this run was launched by the IDE - see [ideLaunchEnvVar].
+  final Map<String, String>? environmentOverrideForTesting;
 
   /// Test-only override for the `flutter` executable path.
   final String? flutterExecutableForTesting;
@@ -408,12 +413,17 @@ class FlutterAppManager {
   }
 
   /// Closes every proxy and deletes info files.
+  ///
+  /// Only an IDE launch removes the device info; a plain terminal run leaves
+  /// it alone, since it may belong to a pending IDE launch.
   Future<void> dispose() async {
     await stopAll();
     await _runtimes.values.map((runtime) => runtime.proxy.close()).wait;
-    await File(
-      p.join(serverpodToolDir, ideDeviceInfoFile),
-    ).deleteIfExists();
+    if (_isIdeLaunch) {
+      await File(
+        p.join(serverpodToolDir, ideDeviceInfoFile),
+      ).deleteIfExists();
+    }
   }
 
   _AppRuntime? _runtimeFor(String appId) => _runtimes[appId];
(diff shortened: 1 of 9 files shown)


tools/serverpod_vscode_extension/package.json
```diff
@@ -18,7 +18,9 @@
   ],
   "activationEvents": [
     "onLanguage:yaml",
-    "onLanguage:serverpod"
+    "onLanguage:serverpod",
+    "onDebugResolve:dart",
+    "activationEvents:serverpod"
```
FXschwartz: I've run into this before but `activationEvents:` isn't a prefix VS Code recognizes, so this entry never fires. `onLanguage:serverpod` + the new `onDebugResolve:dart` should cover everything, so this line can just be dropped.

tools/serverpod_cli/lib/src/commands/start/flutter_app_manager.dart
```diff
@@ -410,13 +411,49 @@ class FlutterAppManager {
...
 
   String _infoFileFor(String appId) =>
       p.join(serverpodToolDir, 'flutter-vm-service-info-$appId.json');
 
+  /// The file where the serverpod VS Code extension writes the
+  /// IDE-selected Flutter device to, as `{"deviceId": "<id>"}`, inside
+  /// [serverpodToolDir].
+  ///
+  /// Scoped to one session: the extension writes it as it launches, and
+  /// [dispose] removes it when the session ends, so a later `serverpod start`
```
FXschwartz: Looks like closing the VS Code window kills the `preLaunchTask` terminal without SIGINT/SIGTERM, so `dispose()` never runs and the file survives. The next `serverpod start` from a plain terminal then silently runs the app on whatever device was last picked in the IDE. The next F5 does overwrite the file, so it only happens going from the IDE session to a terminal-session, but I think that's a pretty normal order, at least for me.

We could do something similar as the stale-file check `_checkExistingServer` does for `vm-service-info.json`. Have `ideDevice()` consume the file and fall back to the cache when it's gone. The extension rewrites it on every F5 so re-selection still lands, restarts reuse the cache, and a killed session can't leak into a later terminal one.

What do you think?

FXschwartz (changes requested):

tools/serverpod_cli/lib/src/commands/start/flutter_app_manager.dart
Crazelu: I thought about this but didn't land on a great way to prevent it from happening.
If `ideDevice()` consumes, the file will be left behind if a crash happens before the app is launched.
Because of this, the file needs to be consumed early on and a timestamp can be introduced for staleness check.

Another approach is to introduce an environment variable that will only be provided by the IDE's `task.json` to enable `ideDevice()`. That env variable should never be present when running from the terminal so it simplifies where, how and when to consume the file.

Crazelu: I opted for the environment variable approach [4ac8e06](https://github.com/serverpod/serverpod/pull/5545/commits/4ac8e06b4996b9c59ab156fe6201798652478ad9)

FXschwartz (approved):
LGTM!

PR #5547, opened 2026-08-05
feat: Support `nulls_distinct` key on unique indexes

### LINKED ISSUE #5494: Support unique DB indexes where NULLs are distinct
### Problem to Solve

The default behavior of PostgreSQL unique indexes are to create NULL values as distinct. This is not always desirable.

### Proposal

See PostgreSQL documentation:
https://www.postgresql.org/docs/current/indexes-unique.html

The Serverpod ORM could expand the model definition along these lines:

```yaml
class: Product
table: product
fields:
  tenantId: int
  category: String
  archivedAt: String?
indexes:
  product_unique_idx:
    fields: tenantId, category, archivedAt
    unique: true
    nulls_distinct: false
```

This would append `NULLS NOT DISTINCT` to the `CREATE UNIQUE INDEX` statement.

### Use Case

For example in cloud we typically have a nullable `archivedAt` column in our tables. We need to include this in the unique index to enable marking old records as archived, and creating a new record with the same key values (but with archivedAt = NULL).

Since the default unique index treats null values as distinct from one another, this unique index does not actually guard against duplicate entries in our tables.

### Alternatives

A possible work-around is to stop using NULL in the column and instead use a static dummy value, which would then prevent collisions. But this required more manual logic, is brittle, and is not always workable depending on the data type of the column.

### Additional context

_No response_

- [ ] I want to work on this

marcelomendoncasoares:
Closes: #5494

## Pre-launch Checklist

## Breaking changes

None.


tools/serverpod_cli/test/integration/migrations/create_migration_dialect_rejection_test.dart
```diff
@@ -0,0 +1,129 @@
...
+      test(
+        'when creating a migration, '
+        'then the client migration fails naming the offending index.',
+        () async {
+          var result = await createMigrationAction(config: buildConfig());
+
+          var clientResult =
+              (result as CreateMigrationServerClientCreated).clientResult;
+
+          expect(clientResult, isA<CreateMigrationFailed>());
```
FXschwartz: Are we ok with this behavior? By the time the client migration is rejected, the server migration is already written and registered. So a failed run leaves a stray server migration behind. I have not tested what rerunning the migration would do but I think it would give a "no changes" on the server plus the same client error.

Since the offending index is knowable before anything touches disk, could we generate both sides with `write: false` and only persist when both succeed? Catching `nulls_distinct: false` on `database: all` models at generate time would be even nicer if possible.

FXschwartz (changes requested):

tools/serverpod_cli/test/integration/migrations/create_migration_dialect_rejection_test.dart
marcelomendoncasoares: No, this is not a desired behavior. Although rare with this feature, it is better to close the precedent now than allowing the creation of a migration that is impossible to complete and might require a reversible migration afterwards (with force). I've changed the `create_migration_action` behavior at https://github.com/serverpod/serverpod/pull/5547/commits/46c51bd53ac2b955501b4bf78f749cb36506feac to first generate the migration artifacts and only persist if both were correctly generated.

FXschwartz (approved):
LGTM!

PR #5551, opened 2026-08-05
test: Reduce flakiness on `pgvector` index test

marcelomendoncasoares:
Fixes/reduces the flakiness of the vector index query tests.

The "conservative" query wasn't conservative. It used `efSearch: 2` with `IterativeScan.relaxed` — and iterative scan exists precisely to defeat a small `efSearch`: pgvector keeps re-scanning the index with a growing candidate list until the `LIMIT` is satisfied (up to `maxScanTuples`). So the conservative query happily converged on the true top-30, which is exactly the same set the aggressive query returns → intersection ratio 1.0.

Whether it converged that far depended on the shape of the HNSW graph, which pgvector builds with its own PRNG on every run. Nothing in the test controlled that, so the assertion was a coin flip. The third test (`aggressiveAvg < conservativeAvg`, `strict <`) was flaky for the identical reason — when both queries return the same set the averages are equal.

The fix replaces the statistical assertions with structural ones that pgvector guarantees:

- Conservative query now uses `IterativeScan.off`. With iterative scan off, an HNSW scan cannot return more rows than its candidate list, so `efSearch: 10` returns exactly 10 rows even though the query asks for 30. Confirmed empirically against the embedded PG 16.13 / pgvector 0.8.3 bundle: `ef=1->1 row, 2->2, 5->5, 10->10, 30->30, 40->30, 500->30`.
- Overlap test is now bounded by construction. 10 rows vs 30 rows means the intersection ratio can never exceed `smallEfSearch / resultLimit` (1/3), so that's the tight bound rather than the arbitrary 0.5.
- Replaced the two distance tests with one rank-wise comparison. A larger candidate list visits a superset of the nodes a smaller one visits, so the large-`efSearch` result at rank `i` can never be farther from the query than the small-`efSearch` result at rank `i`. This subsumes both the old "best is at least as close" and "average is closer" checks and is deterministic.
- Added coverage for what caused the flake. Two new tests assert that relaxed and strict iterative scans lift the `efSearch` cap back to the full limit - the behavior that was silently sabotaging the old test is now pinned down explicitly (`strict` had lost its real-query coverage otherwise, since the aggressive query moved to `off`).
- Swapped print for `printOnFailure`, so the diagnostic IDs and distances still show up on CI failures without spamming passing runs.

The dataset generation (5000 near-equidistant vectors, seeded `Random(42)`) is unchanged.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/vector_query_test.dart
+++ b/tests/serverpod_test_server/test_integration/database_operations/runtime_parameters/vector_query_test.dart
@@ -8,6 +8,19 @@ import '../../test_tools/serverpod_test_tools.dart';
 const totalVectors = 5000;
 const vectorDimension = 512;
 
+/// The number of vectors each query asks for.
+const resultLimit = 30;
+
+/// An efSearch below [resultLimit]. Without an iterative scan, the HNSW scan
+/// cannot return more rows than the size of its dynamic candidate list, so this
+/// caps the query below the requested [resultLimit].
+const smallEfSearch = 10;
+
+/// An efSearch well above [resultLimit], leaving the scan free to return the
+/// full [resultLimit] and to explore a superset of the candidates visited with
+/// [smallEfSearch].
+const largeEfSearch = 500;
+
 void main() {
   withServerpod(
     'Given a large number of vectors with very similar distances',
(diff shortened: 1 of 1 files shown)


FXschwartz (approved):
LGTM!

PR #5552, opened 2026-08-05
chore: Review and update dependencies

### LINKED ISSUE #4745: Revisit all dependencies to tighten the ranges and replace stale packages
Several dependencies on the Serverpod packages are not updated since a long time, presenting ranges that are not even supported anymore due to the lower bound of the SDK version. An example is the `meta: ^1.15.0` dependency, [as pointed in PR #4737](https://github.com/serverpod/serverpod/pull/4737#discussion_r2820884538).

We should revisit our constraints and ensure that all lower bound versions are consistent with the repository constraints.

In this revision, we should also map which dependencies are no longer being maintained and can present issues for the future - as the [redis](https://pub.dev/packages/redis) package, that does not receive updates since 2 years. If replacements exist, they should be done as part of this same "dependency sanity" effort.

### LINKED ISSUE #5536: Update sign_in_with_apple to latest version
### Problem to Solve

The following plugins do not support Swift Package Manager for macos:
  - sign_in_with_apple
This will become an error in a future version of Flutter. Please contact the plugin maintainers to request Swift
Package Manager adoption.

### Proposal

Version 8.0.0 [#](https://pub.dev/packages/sign_in_with_apple/changelog#800)
Support Swift Package Manager (SPM)

### Use Case

When I use auth in client.

### Alternatives

Without

### Additional context

_No response_

- [ ] I want to work on this

marcelomendoncasoares:
Audits the repository's direct dependencies and records the constraint policy in a new `docs/process/dependency_maintenance.md` (as part of the new effort to build context incrementally into the repository).

Closes: #4745
Closes: #5536

- Removes unused direct declarations: `retry`, `built_collection`, and `super_string`. The handful of `super_string` helpers are replaced by local extensions in `serverpod_cli` and `serverpod_shared`, with unit tests.
- Adds three narrow compatibility floors that the downgrade check exposed: `ffi_leak_tracker`, `quiver`, and `win32 6.0.1`. Each is commented with the invalid solver selection it prevents.
- Bumps `melos` to the latest compatible 6.x. A follow-up PR will bump `melos` again introducing proper workspace usage on the monorepo.
- Extends `sign_in_with_apple` to `>=7.0.1 <9.0.0` in `serverpod_auth_idp_flutter`. 8.x's only Dart API break renames the `SignInWithAppleButton` icon-alignment enum, a widget Serverpod does not use; every API it does use is unchanged. 8.0.0 raises its own floors to Dart 3.11 / Flutter 3.41, so the range self-gates: 7.0.1 resolves on the minimum supported SDKs, 8.1.0 on current ones.
- Makes `melos` and `serverpod analyze-pubspecs` ignore `tests/temp`. Leftover CLI scratch projects there previously aborted melos workspace discovery while still exiting 0, so the validation commands silently did nothing.

## Dependency maintenance audit

The July 2026 audit screened 115 direct hosted dependencies. Thirty had not published a stable release in at least one year, and seven in at least two years.

Unused direct declarations for `built_collection`, `retry`, and `super_string` were removed. The original example of `redis` as an inactive dependency published version 4.1.0 in March 2026 and no longer requires replacement.

### CLI follow-ups

- Remove the remaining transitive `super_string` dependency from `cli_tools` by replacing its uses with local, tested helpers.
- Replace the archived `ci` package with explicit CI detection in `cli_tools`, covering the standard CI variable and supported provider-specific variables.
- Evaluate replacing `recase` with either a compatibility wrapper around `change_case` or a small locally maintained implementation. Any migration must include golden tests for generated identifiers, especially acronyms, digits, and punctuation.

## Pre-launch Checklist

## Breaking changes

None.


packages/serverpod_shared/lib/src/util.dart
```diff
@@ -44,7 +43,9 @@ String databaseTypeToLowerCamelCase(String databaseType) {
     if (previousValue.isEmpty || element.isEmpty) {
       return '$previousValue${element.toLowerCase()}';
     } else {
-      return '$previousValue${element.capitalize()}';
+      var first = String.fromCharCode(element.runes.first).toUpperCase();
+      var rest = element.substring(1).toLowerCase();
+      return '$previousValue$first$rest';
```
FXschwartz: nit - This ended up as both an inline port and a duplicate. It's the same logic as `StringManipulationExtension.capitalize` over in `serverpod_cli`. Since cli already depends on shared, could shared host the extension and cli import it?

Also worth noting the description says the `super_string` helpers were replaced "with unit tests", but `databaseTypeToLowerCamelCase` has no coverage anywhere so this is the only ported code without a test.

tools/serverpod_cli/pubspec.yaml
```diff
@@ -70,12 +71,11 @@ dependencies:
   source_span: ^1.10.0
   stream_channel: ^2.1.1
   stream_transform: ^2.1.0
-  super_string: ^1.0.3
   uuid: ^4.5.3
   vm_service: '>=13.0.0 <16.0.0'
   watcher: ^1.0.2
   whiskers: ^0.1.1
-  win32: ^6.0.0
+  win32: ^6.0.1
```
FXschwartz, line 79: nit - the description says each compatibility floor gets a comment with the solver selection it prevents, but win32 didn't get one here or in `serverpod_shared`.

FXschwartz (dismissed):
PR looks great, but #4745 also asked for mapping which dependencies are no longer maintained, with redis as the example to replace. I don't see that mapping anywhere in the PR or the new process doc.

Also very minor nit - the description says docs/process/dependency_maintenance.md but the file landed at docs/design/process/.

tools/serverpod_cli/pubspec.yaml
marcelomendoncasoares: Added a comment on each at https://github.com/serverpod/serverpod/pull/5552/commits/ef5219c500d6e973f5e2cb3817eeb10387c1cd64.

packages/serverpod_shared/lib/src/util.dart
marcelomendoncasoares: Unified and test-covered at https://github.com/serverpod/serverpod/pull/5552/commits/e26d72f00f176924f662d9ce144bc9a0b3c151b8.

marcelomendoncasoares:
> PR looks great, but #4745 also asked for mapping which dependencies are no longer maintained, with redis as the example to replace. I don't see that mapping anywhere in the PR or the new process doc.
> 
> Also very minor nit - the description says docs/process/dependency_maintenance.md but the file landed at docs/design/process/.

I also updated the PR description to rescue a summary of the changes - including why `redis` was no longer a problem. The file was on the wrong place, so I moved it at https://github.com/serverpod/serverpod/pull/5552/commits/39e97146dc655be1d7c76117f9e6ab1734d8fa01 (the PR description was right).

docs/process/dependency_maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+reflected in the declared lower bound.
+
+The downgrade check can surface a transitive package whose oldest permitted
+release cannot compile on a supported SDK. Pin such a package with a direct
+constraint that excludes the invalid releases, and comment the constraint with
+the selection it prevents. These pins should stay as narrow as the defect
+requires, and be removed once the upstream graph no longer permits the broken
+selection.
+
+Repository-wide tooling migrations, such as moving the workspace to a new major
```
nielsenko, line 31: This is very admirable, but less trivial to effectuate. We will need try remove the pinned floor, then re-resolve and see if we still end on a never version than the pinned floor, and if so remove the redundant dep.

Doable, but a small project in itself.

docs/process/dependency_maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+  break for all consumers, not for this repository.
+- Determine which of the changed APIs Serverpod uses. A rename or removal in a
+  widget or helper the repository never references does not affect it.
+- Check the transitive majors the bump carries. A major caused only by additive
+  enum cases is breaking solely for exhaustive switches, so it matters only if
+  the repository switches over that enum.
+- Compare the candidate's own SDK constraints against the floors the depending
+  packages declare. When the candidate requires newer SDKs, a widened range
+  self-gates: the floor keeps resolving to the older release while current
+  toolchains select the newer one. Both ends must then be verified.
```
nielsenko, line 61: The both ends must be verified is not actually done today. Do we intend to add jobs for this on CI? We do the upgrade/downgrade dance, but tests are not run (we only lint).

docs/process/dependency_maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+- For Flutter plugins, review the native and packaging changes as well. Platform
+  build baselines, deployment targets, and plugin layouts usually follow from
+  the required Flutter version rather than from Serverpod.
+- Consider whether the changed API is re-exported. An API the repository does
+  not use itself can still reach downstream users through a barrel file or a
+  public signature.
```
nielsenko, line 67: We should simply stop re-exporting other packages. Maybe 4.0 is time?

docs/process/dependency_maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Validation
+
+Validate changes with the minimum SDK versions and run:
+
+```console
+melos bootstrap
+melos downgrade
+melos lint_loose
+serverpod analyze-pubspecs
```
nielsenko, line 97: Should this be part of the cli, or a separate tool? What is the benefit for customers that they can run this?

packages/serverpod_shared/lib/src/util.dart
```diff
@@ -3,10 +3,44 @@ import 'dart:math';
...
+      value.allMatches(substring(start, end)).length;
+
+  /// Converts space- or underscore-separated words to camel case.
+  String toCamelCase({bool isLowerCamelCase = false}) {
+    var separator = RegExp(r'[ _]');
+    var parts = split(separator);
+
+    if (!contains(separator)) {
+      return isLowerCamelCase ? toLowerCase() : capitalize();
+    }
```
nielsenko: Why is this needed? `parts` will just be a one item list in this case. We already payed for the `split`, why pay for the `contains`?

packages/serverpod_shared/lib/src/util.dart
```diff
@@ -3,10 +3,44 @@ import 'dart:math';
...
+
+    return [
+      isLowerCamelCase ? parts.first.toLowerCase() : parts.first.capitalize(),
+      ...parts.skip(1).map((part) => part.capitalize()),
```
nielsenko, line 30: `skip(1)` works fine on a one-item list - it is just an empty iterator.

packages/serverpod_shared/test/string_manipulation_test.dart
```diff
@@ -0,0 +1,73 @@
...
+import 'package:test/test.dart';
+
+void main() {
+  test(
+    'Given a mixed-case string, '
+    'when it is capitalized, '
+    'then only its first character is uppercase.',
+    () {
+      expect('hELLO'.capitalize(), 'Hello');
+    },
```
nielsenko, line 11: I know this is common behavior for `capitalize` across languages, and why I always disliked it 😄 _rant over_

templates/pubspecs/tools/serverpod_cli/pubspec.yaml
```diff
@@ -57,6 +56,8 @@ dependencies:
   pub_api_client: '>=2.4.0 <4.0.0'
   pub_semver: ^2.1.4
   pubspec_parse: ^1.5.0
+  # Nocterm permits quiver 3.0.0, which cannot compile on Dart 3.
+  quiver: ^3.2.2
```
nielsenko, line 60: We should bump constraint in `nocterm` instead. We are about to roll a new release anyway, and nocterm requires dart 3.5 anyway.

pubspec.yaml
```diff
@@ -4,7 +4,7 @@ environment:
   sdk: '^3.10.3'
 
 dev_dependencies:
-  melos: ^6.1.0
+  melos: ^6.3.3
```
nielsenko, line 7: This is getting old. Not for this PR perhaps, but we should consider upgrading

nielsenko (approved):
LGTM. 

I have some comments. Mostly follow-up work if we want to full-fill the full ambitions here.

And a small efficiency thing, but `toCamelCase` is unlikely on any critical path.

pubspec.yaml
marcelomendoncasoares: Will be fixed on the follow up PR to use Dart workspaces on the monorepo.

templates/pubspecs/tools/serverpod_cli/pubspec.yaml
marcelomendoncasoares: I'll leave this for us to fix when bumping the TUI to get the nocterm fixes.

docs/process/dependency_maintenance.md
marcelomendoncasoares: Well, agents do it quite thoroughly without much trouble. I had Codex running for a while on it, and it didn't even spent much tokens because most of the work was waiting the resolution. It first fetched the ranges from all packages then checked floors and went on the loop until the end. A project for a human, but pretty boring and repetitive work that agents handle nicely.

docs/process/dependency_maintenance.md
marcelomendoncasoares: I think this is for the deps review process in itself, that we should do from time to time, but not something we need to be actively on top to justify a CI for now.

docs/process/dependency_maintenance.md
marcelomendoncasoares: Not so sure... This is a broader design change that I don't personally like. When using third party libraries, I find pretty annoying to have to explicitly declare transitive dependencies just to be able to use a type that is part of the signature of such package. Because, from the user perspective, it doesn't matter whether a type is owned by the lib or used from a transitive dep, since they are using that specific library. So exporting provides a nice UX with little compromise - since the package is already depending on it anyway. And the maintenance cost is no different from maintaining owned exports on barrel files.

docs/process/dependency_maintenance.md
marcelomendoncasoares: We sure should remove it from the CLI sometime. It is not meant for our users anyway, so no reason to have it there - instead of as a `melos` script, as all our other tooling. Just need to be on a separate PR.

packages/serverpod_shared/lib/src/util.dart
marcelomendoncasoares: Removed at https://github.com/serverpod/serverpod/pull/5552/commits/af286fb69f6a0fe63512775529104b6b3d078a10.

PR #5553, opened 2026-08-05
feat: Post message central events globally by default with fallback to local delivery

### LINKED ISSUE #2733: Post message central events globally by default and downgrade if "Redis" is not available.
Currently `MessageCentral` takes a flag named `global` when messages are posted. If true, the message is posted through Redis and if `false` the message is posted locally within the server.

When `global` is true and Redis is not available, an exception is thrown.

```dart
  Future<bool> postMessage(
    String channelName,
    SerializableModel message, {
    bool global = false,
  }) async {
```

This behavior forces users of the API to make decisions around the environement configuration that the server is run on.

The MessageCentral interface should by default try to post messages globally, if no "global" service is available, such as Redis, then the message should automatically fall back to be posted locally.

To allow the same level of configuration as before, an enum that specifies where a message should be posted can be used as a parameter to the method.

```dart
enum MessageMode {
  /// The message is only transmitted locally in the server instance.
  local,

  /// The message is only transmitted through a global message transmitter (Redis).
  /// Throws if none is available.
  global,

  /// The message will first try to transmit through a global message transmitter,
  /// If no global message transmitter is available, the message is transmitted locally.
  all,
}
```

And given the new enum, `all` could then be used as the default value:
```dart
  Future<bool> postMessage(
    String channelName,
    SerializableModel message, {
     MessageMode mode = MessageMode.all,
  }) async {
```

> [!NOTE]
> Not the final naming.

**Additional context**
This is a breaking change and can only be introduced with a major release of Serverpod.

FXschwartz:
`postMessage` took a `global` flag that throws if Redis is not enabled.

This PR replaces the flag with a `MessageScope` enum (`local`, `global`, `auto`) and defaults to `auto`, which posts through Redis when it is enabled and otherwise falls back to local delivery. `authenticationRevoked` already implemented this exact fallback manually with a try/catch and now just relies on the default.

The fallback only applies when Redis is not enabled. If Redis is enabled but unreachable, `auto` behaves like `global` and the post fails best effort, same as before. `MessageScope.global` still throws a `StateError` without Redis.

I went with `auto` instead of the `all` name suggested in the issue since it makes more sense to me personally, but happy to change it.

Fixes #2733 

## Pre-launch Checklist

## Breaking changes
**serverpod**
- `MessageCentral.postMessage` and `session.messages.postMessage` now take `MessageScope scope = MessageScope.auto` instead of `bool global = false`.
- With Redis enabled, messages posted with the default scope are now delivered asynchronously through Redis, and listeners receive a deserialized copy of the message instead of the posted instance.

### DIFF
--- a/packages/serverpod/lib/src/server/message_central.dart
+++ b/packages/serverpod/lib/src/server/message_central.dart
@@ -18,6 +18,20 @@ abstract class MessageCentralServerpodChannels {
 typedef MessageCentralListenerCallback =
     void Function(SerializableModel message);
 
+/// Determines how a message posted to the [MessageCentral] is delivered.
+enum MessageScope {
+  /// The message is only delivered within this server instance.
+  local,
+
+  /// The message is delivered across the cluster via Redis. Requires Redis
+  /// to be enabled.
+  global,
+
+  /// The message is delivered across the cluster if Redis is enabled,
+  /// otherwise it is delivered locally within this server instance.
+  auto,
+}
+
 /// The [MessageCentral] handles communication within the server, and between
 /// servers in a cluster. It is especially useful when working with streaming
 /// endpoints. The message central can pass on any serializable to a channel.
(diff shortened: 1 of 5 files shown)


Crazelu (approved):
Looks good

PR #5555, opened 2026-08-06
feat: Add opt-in httpOnly cookie authentication for web clients

### LINKED ISSUE #4045: Make web authentication use strict httpOnly cookies

FXschwartz:
Adds opt-in web authentication that keeps tokens out of JavaScript reach. With `authCookie` configured on the server and `client.cookieAuth = true` on a browser client:

- **SAS (session tokens):** issued as an `HttpOnly`, `Secure`, `SameSite` cookie via `Set-Cookie`; the response body carries a blank token, so the credential never reaches JS.
- **JWT:** the refresh token lives in an `HttpOnly` cookie `Path`-scoped to the refresh endpoint; the access token stays in memory, per tab, and is re-minted through the refresh cookie on page load.
- **Per-call auth intent:** `@unauthenticatedClientCall` methods send a `cookie-transport` marker, cookies are carried but the main auth cookie never authenticates the call. Method streams carry the same intent in `OpenMethodStreamCommand.authMode`.
- **Stream lifecycle:** the method stream WebSocket reconnects when the signed-in identity changes, so streams never run under a stale handshake cookie identity.
- **Cross-tab refresh coordination:** cookie-mode JWT refreshes serialize across tabs via the Web Locks API, and server-side rotation runs under a row lock, concurrent tabs can no longer trip reuse revocation and kill the session. Cookie-mode JWT fails with a clear error on browsers without Web Locks.

Native and desktop clients are unchanged.

## CSRF
1. `SameSite` on the cookie (default `Lax`; `None` requires `Secure`).
2. Server-side `Origin` validation against `allowedOrigins` for every cookie-transport request and WS handshake (`authCookie` requires a non-empty allow-list).
3. A marker header (`x-serverpod-auth-mode`) that forces a CORS preflight, which fails for non-allow-listed origins.

Fixes #4045 

## Pre-launch Checklist

## Breaking changes


modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/test/session_manager_test.dart
```diff
@@ -115,6 +117,465 @@ void main() {
...
+                'b0eebc99-9c0b-4ef8-bb6d-6bb9bd380a22',
+              ),
+              authStrategy: AuthStrategy.session.name,
+              token: '',
+            ),
+          );
+        });
+
+        test(
+          'then modern method streaming connections are closed.',
```
marcelomendoncasoares: Suggestion: No need to call "modern method streaming", since there is only one possibility right now.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/test/session_manager_test.dart
```diff
@@ -115,6 +117,465 @@ void main() {
...
+        );
+
+        await expectLater(
+          sessionManager.updateSignedInUser(authSuccess),
+          throwsA(isA<StateError>()),
+        );
+        expect(await storage.get(), isNull);
+        expect(sessionManager.authInfo, isNull);
+      },
+    );
```
marcelomendoncasoares: Suggestion: These tests are mixing the test case with the action and hiding the actual action of the test. For example, the "JWT sign-in succeeds with an access token and no refresh token" describes part of the test scenario, not an action. The action itself (updating the signed in user) is hidden on the description.

I get that part of this might be to reuse the generic cookie-auth `ClientAuthSessionManager`, but then we can move these tests to a separate file focused on the cookie-auth and put this `setUp` on the `main`.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/test/session_manager_test.dart
```diff
@@ -115,6 +117,465 @@ void main() {
...
+        _authSuccess(authStrategy: AuthStrategy.session.name, token: 'token'),
+      );
+
+      expect(client.closeStreamingMethodConnectionsCallCount, 0);
+    },
+  );
+
+  group('Given a JWT auth key provider in cookie mode', () {
+    late AuthSuccess? authInfo;
+    late _TestRefreshJwtTokensEndpoint refreshEndpoint;
```
marcelomendoncasoares: Suggestion: Same observation on the `'Given a cookie-auth ClientAuthSessionManager'` group is valid here.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/test/session_manager_test.dart
```diff
@@ -259,3 +720,165 @@ class _InMemoryKeyValueStorage implements KeyValueStorage {
...
+    bool authenticated = true,
+  }) {
+    events.add('call $endpoint.$method');
+    throw UnimplementedError();
+  }
+}
+
+class _CookieRequestDelegate extends ServerpodClientRequestDelegate {
+  @override
+  bool get supportsCookieAuth => true;
```
marcelomendoncasoares: Discussion: Mocking the entire generated client goes a bit against the principle of testing the code as closer to production as possible. Can we write these tests as integration tests instead? This follows the current pattern we have for testing JWT, SSS and all IDPs.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -152,6 +231,7 @@ abstract class ServerpodClientShared extends EndpointCaller {
     this.onSucceededCall,
     bool? disconnectStreamsOnLostInternetConnection,
     http.Client? httpClientOverride,
+    @visibleForTesting ServerpodClientRequestDelegate? requestDelegate,
```
marcelomendoncasoares: Suggestion: We should not leak test only injections into the public interface of the class that the generated client implements. Can't the injection use the lower level `httpClientOverride`? Otherwise we can just skip unit testing these classes and assert the behavior through integration tests - always preferred.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/integrations/adapters/jwt_token_manager.dart
```diff
@@ -19,14 +20,20 @@ class JwtTokenManager implements TokenManager {
...
   /// Creates a new [JwtTokenManager] instance.
   JwtTokenManager({
     required final JwtConfig config,
     final AuthUsers authUsers = const AuthUsers(),
-  }) : jwt = Jwt(
-         config: config,
-         authUsers: authUsers,
-       );
+    final Jwt? jwt,
+  }) : _config = config,
```
marcelomendoncasoares: Suggestion: This also seem a parameter introduced to allow building unit tests that won't be necessary if we move to integration tests (also removing an entire Request -> Session -> Server -> Serverpod mock chain on `jwt_cookie_auth_test.dart`).

marcelomendoncasoares (changes requested):
I haven't gone through the entire PR, but I'll suggest we address the current comments first before I dive on the server-side changes and integration tests. The central point around the tests is avoiding elaborated mocks and injections in favor of integration tests. This has important advantages:

- Speeds up review, given that the mocks are hard to review.
- Reduces maintenance cost, because the mocks are brittle and generate more future diff.
- Increases confidence, since the code is tested under real usage scenario.
- Prevents the need to cover the feature twice (integration is needed anyway).

Unit tests are great when testing isolated components, but value flips when they require plumbing too much to isolate the component. We either end up exposing parameters that should not be part of the interface or mocking implementation details that decouple the test from the rest of the code.

My suggestion then is that we heavily rely on our integration tests being so cheap and easy to run now, and only use unit tests when they come near free - with no need to mock everything that is above.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/test/session_manager_test.dart
FXschwartz: Gone! Those tests moved to the browser E2E as part of the mock cleanup and the wording went with them.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/test/session_manager_test.dart
FXschwartz: Moved! The sanitization and `StateError` guards are in the browser E2E now with the action in the when, and `session_manager_test` is back to what's on main.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/test/session_manager_test.dart
FXschwartz: Same as above. Group's gone!

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/test/session_manager_test.dart
FXschwartz: Dropped the mock. Everything it covered runs through the E2E rig now.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
FXschwartz: Removed it. Browser tests go through `httpClientOverride` like you suggested, the lock tests assert against the real Web Locks delegate on chrome, and the VM wire-format tests are replaced by header-echo assertions in the browser E2E. 

One thing I kept as a unit test is `webAuthHeaders` composition on the delegate base, the `/api` base-path case has no E2E home since the rig runs at the root.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/integrations/adapters/jwt_token_manager.dart
FXschwartz: Removed, and `jwt_cookie_auth_test` with the whole fake chain. One catch: `withServerpod` sessions have request: null so the cookie code isn't really reachable there.

tests/serverpod_auth_test/serverpod_auth_test_server/lib/src/endpoints/auth_test_endpoint.dart
```diff
@@ -40,11 +48,15 @@ class AuthTestEndpoint extends Endpoint {
...
+  ///
+  /// Unauthenticated like real sign-in endpoints, so that cookie issuance
+  /// treats the call as the caller signing in.
+  @unauthenticatedClientCall
   Future<AuthSuccess> createJwtToken(
     final Session session,
     final UuidValue authUserId,
   ) async {
-    return jwt.createTokens(
+    return AuthServices.getTokenManager<JwtTokenManager>().issueToken(
```
marcelomendoncasoares: Question: After reverting the `jwt` parameter injection, is this change still needed?

tests/serverpod_auth_test/serverpod_auth_test_server/lib/src/endpoints/auth_test_endpoint.dart
```diff
@@ -17,6 +21,10 @@ class AuthTestEndpoint extends Endpoint {
   }
 
   /// Creates a new session authentication for the test user.
+  ///
+  /// Unauthenticated like real sign-in endpoints, so that cookie issuance
+  /// treats the call as the caller signing in.
+  @unauthenticatedClientCall
```
marcelomendoncasoares: Discussion: We should not make this assumption, because there is one valid workflow for logging in as a logged in user: **account linking** (not yet implemented). This will be trigger to invoke the account merging process. Also, other `login` endpoints are not really `unauthenticatedClientCall`.

We do have to guard against replacing A's `httpOnly` cookie with B's session (even when merging), but we should do it following this matrix:

| Already signed in as A | Credentials for | Merge request | Result |
|---|---|---|---|
| yes | A | n/a | Re-issue A's session (cookie replace / new header tokens) |
| yes | B | present | Merge path; do not install B as the session |
| yes | B | absent | Throw (must sign out to switch) |
| no | B | n/a | Normal login; may set cookie |

Then, we should unify native with web to always forbid logging in as a different user from an authenticated account without a prior logout (when no account merge request exist). The cookie-only `StateError` ("secrets came back in the body") is replaced by an error on both platforms.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/integrations/adapters/jwt_token_manager.dart
```diff
@@ -19,11 +20,14 @@ class JwtTokenManager implements TokenManager {
...
 
+  final JwtConfig _config;
+
   /// Creates a new [JwtTokenManager] instance.
   JwtTokenManager({
     required final JwtConfig config,
     final AuthUsers authUsers = const AuthUsers(),
-  }) : jwt = Jwt(
+  }) : _config = config,
+       jwt = Jwt(
```
marcelomendoncasoares: Suggestion: Make the `jwt.config` public instead to avoid duplicating the parameter. The `JwtTokenManager` is already a class to be used internally on the server, so no reason to hide the config.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/integrations/adapters/jwt_token_manager.dart
```diff
@@ -36,13 +40,33 @@ class JwtTokenManager implements TokenManager {
...
+    final refreshToken = authSuccess.refreshToken;
+    if (refreshToken == null) return authSuccess;
+
+    final maxAgeSeconds = _config.refreshTokenLifetime.inSeconds;
+    session.writeWebAuthRefreshCookie(
+      refreshToken,
+      maxAgeSeconds: maxAgeSeconds > 0 ? maxAgeSeconds : null,
+      path: jwtRefreshCookiePath(session),
+    );
+    return authSuccess.copyWith(refreshToken: null);
```
marcelomendoncasoares: Suggestion: Issuing the cookie here seems like breaking the token manager abstraction. Providers should not need to be aware of this feature, that is a transport/delivery concern only. We should move from here and the `ServerSideSessionsTokenManager` to the upper `issueToken` level.

If the request is in cookie mode and the minted session belongs to the caller (no authenticated caller yet (sign-in), or `session.authenticated` is the same user as `AuthSuccess.authUserId`), then:

- If `refreshToken` present: write the refresh cookie (path-scoped), send `refreshToken: null` in the body.
- Else If `token` present: write the auth cookie, send `token: ''` in the body.
- Else: leave secrets in the body (covers the admin minting a session for a different user).

The last case should not set the cookie, because the body is for the caller (admin) to pass ahead to a user. Reading the refresh cookie when the refresh argument is omitted stays on the refresh endpoint (JWT-specific).

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/util/jwt_auth_success_extension.dart
```diff
@@ -7,8 +7,17 @@ import '../jwt.dart';
...
+      throw const FormatException(
+        'Cannot parse JWT refresh token ID from an empty '
+        '`AuthSuccess.refreshToken`. In cookie-mode sign-in the JWT refresh '
+        'token is carried in an HttpOnly cookie instead of the response body.',
+      );
+    }
+
     try {
-      return RefreshTokenString.parseRefreshTokenString(refreshToken!).id;
+      return RefreshTokenString.parseRefreshTokenString(refreshToken).id;
```
marcelomendoncasoares, line 20: Suggestion: This throw is a consequence of us removing the tokens right when we issue them. But it breaks the intended workflow of using `jwtRefreshTokenId`/`serverSideSessionId` to attach metadata after login (see the `token_metadata_integration_test.dart` file). Instead, a possible approach can be to mask the tokens on the transport layer.

If we create a `CookieAuthSuccess` that extends (or implements) the `AuthSuccess`, we can override the `toJson`/`toJsonForProtocol` methods and apply the masking there. This way, server code that expects the recently emitted tokens to exist will continue to work normally.

The throw here can still be maintained, in case the `AuthSuccess` is returned from the `Client` through an endpoint, but at least the base case for these getters will still be supported.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/lib/src/session_manager.dart
```diff
@@ -160,11 +164,49 @@ class ClientAuthSessionManager implements RefresherClientAuthKeyProvider {
...
+    if (_usesCookieAuth && identityChanged) {
+      // The method-stream WebSocket authenticates from the cookie captured at
+      // handshake time, so a socket from before the identity change would
+      // keep the old identity. Closing it makes the next stream reconnect
+      // with a fresh handshake; a same-identity update (e.g. a JWT token
+      // rotation) keeps streams running.
+      await _caller?.client.closeStreamingMethodConnections(exception: null);
+    }
     onAuthInfoChanged?.call(_authInfo);
   }
```
marcelomendoncasoares, line 187: Discussion: Do we have a clear use case to support keeping the stream opened for an authenticated user on non-cookie clients? It seems to me that we can make this behavior more symmetric between the modes, which will also be safer.

Streams that start unauthenticated and are promoted to authenticated can still be supported on native (although not possible with cookies), but we can default to always close the stream if the existing authenticated user changes (logout).

Also, should we use a typed exception that clearly informs the connection being closed due to logout?

marcelomendoncasoares (changes requested):
A few more topics for discussion on the design before jumping on the implementation. The central concern behind all them is to reduce the branching on users' code when working with multi-platform apps.

tests/serverpod_auth_test/serverpod_auth_test_server/lib/src/endpoints/auth_test_endpoint.dart
FXschwartz: Makes sense! This also kills the silent body-token fallback which is nice. Two questions:
1. is `MultiTokenManager.issueToken` the right place to enforce the throw? Every IDP login funnels through it via `_tokenIssuer.issueToken`, so combined with your transport suggestion it'd be the single spot for both the policy check and the cookie writes.
2. That would make calling the concrete manager / business layer directly the sanctioned admin path (no cookies, secrets in body). Is that the intent? If so I'll drop the `unauthenticatedClientCall` from the test endpoints so they behave like real login endpoints and test the matrix instead.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/integrations/adapters/jwt_token_manager.dart
FXschwartz: Agreed, will move it. What about the cookie `Max-Age`? For SAS I can derive it from `AuthSuccess.tokenExpiresAt`, but the JWT refresh lifetime isn't on `AuthSuccess` anywhere.

I can add a nullable getter on `TokenManager`, something like `refreshTokenCookieLifetime, null = session cookie` so the generic level can ask the concrete manager. Or would you rather put the lifetime on `AuthSuccess/CookieAuthSuccess` since we're touching that anyway for the masking?

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/lib/src/session_manager.dart
FXschwartz: No real use case for keeping them open on native that I can think of, happy to make it symmetric.

On the typed exception, the current clean close was deliberate so signing out doesn't error blast every open stream listener in the UI. A typed exception is more informative but means every listener needs to handle an error on logout instead of just `onDone`. Fine either way, just want to confirm you want the error behavior before I flip it and add something like `SignedOutException`

tests/serverpod_auth_test/serverpod_auth_test_server/lib/src/endpoints/auth_test_endpoint.dart
marcelomendoncasoares: Can't be on the `MultiTokenManager`, because there is a valid workflow for users to get a specific `TokenManager` to issue a desired type of token. Maybe the best alternative is to:

1. Make the `TokenManager` class no longer an interface, so that all implementations extend it.
2. Make the `issueToken` method concrete with the cookie-aware path and annotated with `@nonVirtual`.
3. Add another abstract `createToken` method that will replace the current implementation.

This will be a breaking change for all custom `TokenManager` classes, but we don't expect to be so many out there. On the other hand, it won't break code that uses the token managers (certainly a lot more). The `@nonVirtual` annotation will prevent that users mistakenly override the `issueToken` method. Instead, subclasses will only implement the `createToken` that will be internally called by the `issueToken` before deciding on the cookie path.

Then you can drop the `unauthenticatedClientCall`, because any call to any `TokenManager.issueToken` will correctly evaluate the matrix above and throw or return a `CookieAuthSuccess` - even for custom implementations.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/integrations/adapters/jwt_token_manager.dart
marcelomendoncasoares: Maybe get directly `jwt.config.refreshTokenLifetime` and deduce a few seconds. Since we're issuing the token at that moment, the `refreshTokenLifetime` is global, and the lifetime is on the order of a few minutes, taking "now - 10s" can be a simple and safe path.

```dart
final refreshExpiresAt = DateTime.now().toUtc().add(
  jwt.config.refreshTokenLifetime - const Duration(seconds: 10),
);
```

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/lib/src/session_manager.dart
marcelomendoncasoares: I don't have a strong opinion on either. Maybe closing gracefully is nicer indeed, if we document this behavior as part of the streaming contract.

tests/serverpod_auth_test/serverpod_auth_test_server/lib/src/endpoints/auth_test_endpoint.dart
FXschwartz: Yes with the `TokenManager` rework, `issueToken` is now the policy + delivery entry point, so the endpoint keeps calling it.

tests/serverpod_auth_test/serverpod_auth_test_server/lib/src/endpoints/auth_test_endpoint.dart
FXschwartz: `issueToken` now is concrete and `nonVirtual` on `TokenIssuer`: it throws a new serializable exception `SignInWhileAuthenticatedException` when an authenticated caller targets a different user, on both platforms, then delegates to the abstract `createToken` and applies the cookie delivery.

Funny note is that the browser E2E caught a stale cookie from a previous test making a sign in look like a user switch so we know the guard is doing its job!

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/integrations/adapters/jwt_token_manager.dart
FXschwartz: Turns out jwt.config was already public. The _config field was just redundant, my bad. I removed it and with the cookie write moved to the base class the adapter doesn't need the lifetime at all anymore.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/integrations/adapters/jwt_token_manager.dart
FXschwartz: I moved to the `TokenIssuer` base per your other comment, so it's out of the adapters and `ServerSideSessions.createSession` is back to pure minting. The refresh cookie Max-Age uses your snippet via a small virtual hook the JWT manager overrides, since the base can't see `jwt.config`

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/jwt/util/jwt_auth_success_extension.dart
FXschwartz: Implemented, with one deviation: it implements `AuthSuccess` rather than extends it, since the generated class only has a private constructor. Real secrets stay on the instance, masking happens in `toJson/toJsonForProtocol`. The `jwtRefreshTokenId` throw stays for genuinely-empty values.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/lib/src/session_manager.dart
FXschwartz: Made it symmetric! Identity change closes streams in both modes now, and kept the graceful close. The contract is documented on `updateSignedInUser`.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/business/multi_token_manager.dart
```diff
@@ -33,15 +33,17 @@ class MultiTokenManager implements TokenManager {
     ];
   }
 
+  // Delegates to the primary manager's createToken: the policy check and
+  // cookie delivery run once, in this manager's inherited issueToken.
```
marcelomendoncasoares: Suggestion: No need for the comment here, since it is redundant with the explanation of the base method.

tests/serverpod_auth_test/serverpod_auth_test_server/lib/src/endpoints/auth_test_endpoint.dart
```diff
@@ -21,12 +25,13 @@ class AuthTestEndpoint extends Endpoint {
...
     final UuidValue authUserId,
   ) async {
-    return _serverSideSessions.createSession(
-      session,
-      authUserId: authUserId,
-      method: 'test',
-      scopes: {},
-    );
+    return AuthServices.getTokenManager<ServerSideSessionsTokenManager>()
+        .issueToken(
```
marcelomendoncasoares: Suggestion: Instead of repeating the call to `AuthServices.getTokenManager`, make the top `_serverSideSessions` and `jwt` be the actual token managers, then call their specific `.serverSideSessions` and `.jwt` on the call sites.

packages/serverpod_shared/test/config_test.dart
```diff
@@ -2013,4 +2013,500 @@ websocketPingInterval: 0
...
+          passwords,
+          loadYaml(apiServerConfig),
+        );
+
+        expect(config.allowedOrigins, isNull);
+      },
+    );
+
+    test(
+      'when set as a YAML list then allowedOrigins is parsed.',
```
marcelomendoncasoares: Suggestion: These tests are presenting part of the condition on the "when" clause. Instead of a generic "Given" grouping all `allowedOrigins` tests, we should split them into each configuration setup (i.e. "Given no allowedOrigins configuration set", "Given allowedOrigins set to a YAML list", ...).

packages/serverpod_shared/test/config_test.dart
```diff
@@ -2013,4 +2013,500 @@ websocketPingInterval: 0
...
+apiServer:
+  port: 8080
+  publicHost: localhost
+  publicPort: 8080
+  publicScheme: http
+allowedOrigins:
+  - https://app.example.com
+''';
+
+    test('when not set then authCookie is null.', () {
```
marcelomendoncasoares: Suggestion: Same split commented above is valid for these tests.

templates/serverpod_templates/projectname_server_upgrade/config/development.yaml
```diff
@@ -71,3 +71,17 @@ sessionLogs:
 #futureCall:
 #  concurrencyLimit: 1 # Defaults to 1, a negative or null value removes the limit
 #  scanInterval: 5000 # Unit in milliseconds, defaults to 5000
+
+# Uncomment to keep web client auth tokens in httpOnly cookies instead of
+# JavaScript-readable storage. Web clients opt in with `client.cookieAuth = true`.
+# Requires `allowedOrigins`; browsers from origins not in the list lose
+# cross-origin access, including to public endpoints.
+#authCookie:
```
marcelomendoncasoares: Suggestion: Give a bit more emphasis that this is recommended, since otherwise web leaves tokens exposed.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -42,12 +42,44 @@ class MethodCallContext {
...
+  /// The cookie-auth request headers for a call with the given [authenticated]
+  /// intent: the [webAuthModeHeaderName] marker and the declared base path,
+  /// or empty when [cookieAuth] is disabled.
+  Map<String, String> webAuthHeaders({required bool authenticated}) => {
+    if (cookieAuth)
+      webAuthModeHeaderName: authenticated
+          ? webAuthModeCookie
+          : webAuthModeCookieTransport,
+    if (cookieAuth) webBasePathHeaderName: cookieAuthBasePath,
+  };
```
marcelomendoncasoares: Question: Since this will only be used on web clients, do we really need to surface it on the generic interface? It makes native clients cookie-aware, but they will never be able to implement it.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -141,6 +173,52 @@ abstract class ServerpodClientShared extends EndpointCaller {
...
+  /// no secrets.
+  CrossTabLock? get authRefreshCrossTabLock {
+    if (!_authRefreshCrossTabLockCreated) {
+      _authRefreshCrossTabLockCreated = true;
+      _authRefreshCrossTabLock = _requestDelegate.createCrossTabLock(
+        'serverpod-auth-refresh:${Uri.parse(host).origin}$_hostBasePath',
+      );
+    }
+    return _authRefreshCrossTabLock;
+  }
```
marcelomendoncasoares: Question: Same question as above regarding this change: can we make it specific to a web-only interface, so that native code don't get polluted with a feature that they can never implement? I get that this is exposed to be used on the `JwtAuthKeyProvider`, but there we already split the interface. Maybe we can remove all this by replacing the injection on the `JwtAuthKeyProvider` by passing the actual `ServerpodClientShared`.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/lib/src/auth_key_providers/jwt_auth_key_provider.dart
```diff
@@ -1,93 +1,218 @@
...
+    /// cookie would trip the server's reuse revocation.
+    CrossTabLock? Function()? getCrossTabRefreshLock,
+
+    /// Optional function to invalidate the cached authentication info before
+    /// refreshing the token.
+    Future<void> Function()? invalidateCachedAuthInfo,
+
+    /// Tolerance to add to the token expiration time before refreshing.
+    Duration refreshJwtTokenBefore = const Duration(seconds: 30),
+  }) : super(
```
marcelomendoncasoares: Comment: If instead of requiring `usesCookieAuth`, `getCrossTabRefreshLock` and `refreshEndpoint` we ask the very `caller.client` (which is a `ServerpodClientShared`), the `JwtAuthKeyProvider` can have access to cookie-aware properties with no need to leak such parameters on the general interface.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/lib/src/auth_key_providers/jwt_auth_key_provider.dart
```diff
@@ -1,93 +1,218 @@
...
+    // single-use, so an uncoordinated concurrent refresh from another tab
+    // would be treated as credential reuse and revoke the session.
+    final crossTabLock = getCrossTabRefreshLock();
+    if (crossTabLock == null) {
+      throw StateError(
+        'Cookie-based JWT auth requires cross-tab refresh coordination, '
+        'but the platform does not support it (the browser is missing the '
+        'Web Locks API).',
+      );
+    }
```
marcelomendoncasoares: Discussion: I suggest we remove this throw and accept the race. The only downside of refreshing with no cross-tab coordination is the exposure to a race condition when using the app in multiple tabs opened nearly at the same instant: when they refresh at the same time, the first succeeds and the second presents a stale key that leads to the deletion of the first one (and the user must login again).

On the other hand, seems that `navigator.locks` is secure-context only. Which means that any `http://` origin (like a LAN dev address other than `localhost`) will receive `WebLocksCrossTabLock.isSupported == false`. The consequence is that any developer who enables cookie auth against a LAN IP would get a `StateError` thrown from `refreshAuthKey` at the moment that the access token expires.

To avoid a silent error, we could log that the client is running with no support for `navigator.locks` and that the race condition can happen. Then, any developer in the above scenario that happen to hit the race will have a quick way to be informed that it is just an artifact of the test environment.

packages/serverpod_client/test/integration/method_stream_connection_url_test.dart
```diff
@@ -0,0 +1,108 @@
...
+
+import 'websocket_extensions.dart';
+import '../test_utils/test_auth_key_providers.dart';
+import '../test_utils/test_serverpod_client.dart';
+import '../test_utils/test_web_socket_server.dart';
+
+class TestSerializationManager extends SerializationManager {}
+
+void main() {
+  group('Given a client with an auth key provider', () {
```
marcelomendoncasoares: Suggestion: 'Given a client with an authenticated auth key provider'

packages/serverpod_client/test/integration/method_stream_connection_url_test.dart
```diff
@@ -0,0 +1,108 @@
...
+      'when opening a method stream '
+      'then the authentication is carried in-band in the open command.',
+      () async {
+        client.callStreamingServerEndpoint<Stream<String>, String>(
+          'test',
+          'method',
+          {},
+          {},
+        );
+        await openCommandReceived.future;
```
marcelomendoncasoares: Suggestion: Group the two tests into a shared `group('when opening a method stream')`.

packages/serverpod_client/test/integration/method_stream_connection_url_test.dart
```diff
@@ -0,0 +1,108 @@
...
+
+    tearDown(() async {
+      client.close();
+      await closeServer();
+    });
+
+    test(
+      'when opening a method stream '
+      'then the connection URL carries no credentials.',
+      () async {
```
marcelomendoncasoares: Question: Maybe I missed these, but are we exercising that signing-out disconnects opened streams?

tests/serverpod_auth_test/serverpod_auth_test_flutter/test/jwt_cookie_auth_test.dart
```diff
@@ -0,0 +1,324 @@
...
+    setUp(() async {
+      userId = await _createTestUser();
+    });
+
+    test(
+      'when creating a JWT token '
+      'then the refresh token is moved from the body to a path-scoped '
+      'HttpOnly cookie.',
+      () async {
+        final response = await _call(
```
marcelomendoncasoares: Suggestion: Maybe worth splitting this group and move the "Given" to the tests, describing the body of the request (which is the subject being tested on the "when creating a JWT token" action).

tests/serverpod_auth_test/serverpod_auth_test_flutter/test/sas_cookie_auth_test.dart
```diff
@@ -0,0 +1,213 @@
...
+  group('Given an SAS sign-in request in cookie mode', () {
+    late String userId;
+
+    setUp(() async {
+      userId = await _createTestUser();
+    });
+
+    test(
+      'when creating a session '
+      'then the token is moved from the body to an HttpOnly cookie.',
```
marcelomendoncasoares: Suggestion: Same comment as the JWT auth cookie test.

packages/serverpod/test/server/cookie_auth_cors_test.dart
```diff
@@ -0,0 +1,111 @@
...
+          headers.accessControlAllowHeaders?.headers,
+          containsAll(['content-type', ...cookieAuthRequestHeaders]),
+        );
+      },
+    );
+
+    test(
+      'when the cookie-auth headers are already present '
+      'then the list is unchanged.',
+      () {
```
marcelomendoncasoares: Suggestion: Same comment here about the generic group with no shared setup and the subject on the when.

packages/serverpod/test/server/auth_cookie_test.dart
```diff
@@ -0,0 +1,164 @@
...
+    });
+
+    test('when it has no leading dot then it is used as-is.', () {
+      var cookie = const WebAuthCookieConfig(
+        domain: 'example.com',
+      ).buildSetCookieHeader('v');
+      expect(cookie.domain?.toString(), 'example.com');
+    });
+
+    test('when it is malformed then a clear ArgumentError is thrown.', () {
```
marcelomendoncasoares: Suggestion: Same comment here as well (each test has a different `WebAuthCookieConfig` and domain configuration that is part of the subject).

marcelomendoncasoares (changes requested):
I have only one discussion on the implementation around the `StateError` throw that can hurt development workflows, but other comments are mostly around tests.

I had some other comments that I decided to fold into one refactor commit that I landed directly to isolate the auth cookie details from the `ServerpodClientShared` main class (and the general delegate interface).

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/common/business/multi_token_manager.dart
FXschwartz: Removed!

tests/serverpod_auth_test/serverpod_auth_test_server/lib/src/endpoints/auth_test_endpoint.dart
FXschwartz: Done! The fields are the managers now and the call sites go through `.issueToken` / `.serverSideSessions` / `.jwt` off them.

packages/serverpod_shared/test/config_test.dart
FXschwartz: Split! Each configuration variant is its own Given now

packages/serverpod_shared/test/config_test.dart
FXschwartz: Split! Same as above

templates/serverpod_templates/projectname_server_upgrade/config/development.yaml
FXschwartz: Beefed it up in all three run-mode configs!

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_client/lib/src/auth_key_providers/jwt_auth_key_provider.dart
FXschwartz: Agreed. The secure context issue on LAN IPs is much worse than the race. 

I removed the throw. Without a lock the refresh now runs uncoordinated with a one-time warning saying Web Locks is unavailable and concurrent tabs may race and force a re login. The VM test flips to asserting the uncoordinated refresh reaches the server.

packages/serverpod_client/test/integration/method_stream_connection_url_test.dart
FXschwartz: Renamed

packages/serverpod_client/test/integration/method_stream_connection_url_test.dart
FXschwartz: Done!

tests/serverpod_auth_test/serverpod_auth_test_flutter/test/sas_cookie_auth_test.dart
FXschwartz: Same split applied here.

tests/serverpod_auth_test/serverpod_auth_test_flutter/test/jwt_cookie_auth_test.dart
FXschwartz: Split!

packages/serverpod/test/server/cookie_auth_cors_test.dart
FXschwartz: Fair, that group had no shared setup either. Split into two tests with the list state in the Given.

packages/serverpod/test/server/auth_cookie_test.dart
FXschwartz: Split!

marcelomendoncasoares (approved):
LGTM! 🍪

PR #5556, opened 2026-08-06
feat: Allow using `serial` on regular `int` columns

### LINKED ISSUE #5484: Auto-increment fields support (SERIAL and BIGSERIAL)
### Problem to Solve

It is currently impossible to create a model that will have an auto-incremented field in the database. The only field that will utilise Postgres SERIAL data type is `id`.

### Proposal

It would be really helpful to be able to use `default=serial` not only on the `id` field, but on any `int` field of the model.

### Use Case

For example, a data synchronisation algorithm could use a `syncVersion` field for an entity to complement client pulls to retrieve changes from the server that were added after the last sync.

### Alternatives

It is possible to use a timestamp as a field that is always incrementing, but there could be cases where several updates on the server were made at the same second, resulting in field conflicts.

### Additional context

_No response_

- [ ] I want to work on this

marcelomendoncasoares:
Closes: #5484

While addressing this, I removed the default `nextval(...)` that we used to generate for Postgres in the past only to identify that the column should be written with `bigserial` (we have never used the `nextval(...)` on real SQL code). It is safe to remove because this check would only happen when generating SQL, which comes after the migration that already evaluates the field as `serial` (in case a project migrates directly from 3.2 -> 4.0.0, for example).

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/database/dialects/postgres.dart
+++ b/tools/serverpod_cli/lib/src/database/dialects/postgres.dart
@@ -264,6 +264,9 @@ extension PostgresColumnDefinitionPgSqlGeneration on ColumnDefinition {
 
     // The id column is special.
     if (isPrimary) {
+      if (isNullable) {
+        throw const FormatException('The id column must be non-nullable');
+      }
       type = '$type PRIMARY KEY';
       nullable = '';
     }
@@ -608,6 +611,14 @@ extension PostgresColumnMigrationPgSqlGenerator on ColumnMigration {
             'ALTER TABLE "$tableName" ALTER COLUMN "$physicalName"'
             ' DROP DEFAULT;\n';
         return out;
+      } else if (newDefault == defaultIntSerial) {
+        // Adding a serial default requires creating a sequence via the serial
+        // pseudo-type on column (re)create. SET DEFAULT cannot express this.
+        throw StateError(
+          'Cannot SET DEFAULT "$defaultIntSerial" on column "$physicalName" '
+          'of table "$tableName". Auto-increment defaults must be applied by '
+          'recreating the column with the serial type.',
+        );
       } else {
         var newDefaultSql = columnDefinition.columnType.getPgColumnDefault(
           newDefault,
--- a/tools/serverpod_cli/lib/src/database/dialects/postgres.dart
+++ b/tools/serverpod_cli/lib/src/database/dialects/postgres.dart
@@ -123,7 +123,7 @@ extension PostgresTableDefinitionPgSqlGeneration on TableDefinition {
 
     var columnsPgSql = <String>[];
     for (var column in columns) {
-      columnsPgSql.add('    ${column.toPgSqlFragment(tableName: name)}');
+      columnsPgSql.add('    ${column.toPgSqlFragment()}');
     }
     out += columnsPgSql.join(',\n');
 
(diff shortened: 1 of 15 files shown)


tools/serverpod_cli/lib/src/database/dialects/postgres.dart
```diff
@@ -252,24 +250,20 @@ extension PostgresColumnDefinitionPgSqlGeneration on ColumnDefinition {
...
+    if (isIntSerialColumn) {
+      type = columnType == ColumnType.bigint || isPrimary
+          ? 'bigserial'
+          : 'serial';
+      defaultValue = '';
+      nullable = ' NOT NULL';
+    }
+
     // The id column is special.
     if (isPrimary) {
```
FXschwartz, line 266: Any reason to drop the `FormatException('The id column must be non-nullable')` guard here? Looks like sqlite.dart still has it, so the two dialects enforce different invariants now. 

Probably harmless since `PRIMARY KEY` implies `NOT NULL` anyway, but if it's intentional I'd remove it from `sqlite` too for parity.

tools/serverpod_cli/lib/src/database/dialects/postgres.dart
```diff
@@ -618,7 +611,6 @@ extension PostgresColumnMigrationPgSqlGenerator on ColumnMigration {
       } else {
         var newDefaultSql = columnDefinition.columnType.getPgColumnDefault(
```
FXschwartz, line 623: With the `nextval` branch removed, `getPgColumnDefault` now returns the literal string serial for int columns, so if `newDefault='serial'` ever reaches this path we'd emit `SET DEFAULT serial` which is invalid SQL.

I know `canMigrateTo` prevents that migration from being generated today, but nothing here enforces it, could we throw in `getPgColumnDefault` (or here) if a serial default slips through?

FXschwartz (changes requested):
Just a couple of comments and questions:

1. All the new tests stop at SQL/definition generation and nothing actually inserts a row and checks the value increments. Might be worth adding a `syncVersion: int?`, `defaultPersist=serial` model to the test server with a runtime test?
2. Since the issue asked for default=serial and we went with `defaultPersist=serial` only, it might be good for the docs PR to call that out explicitly

tools/serverpod_cli/lib/src/database/dialects/postgres.dart
marcelomendoncasoares: No reason, actually. I refactored the checks to split them and did not pay attention to the removal (this exact part was a manual edit of mine). It is indeed harmless, but the exception prevents from removing validation on the upper layer - if we ever remove, this will fail. I'll return it.

tools/serverpod_cli/lib/src/database/dialects/postgres.dart
marcelomendoncasoares: Good suggestion! We can add this throw as a defense in case some future change allow `serial` to pass through to the SQL generator. The previous `nextval` wasn't safe anyway, because we never emitted a `CREATE SEQUENCE` for it - which is required when using an explicit sequence function.

marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5556/commits/cfa5d62a7048d00841ba28cf8ea9203134a29db2.

tools/serverpod_cli/lib/src/database/dialects/postgres.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5556/commits/cfa5d62a7048d00841ba28cf8ea9203134a29db2.

FXschwartz (approved):
Looks great!

PR #5558, opened 2026-08-07
feat: Add Module.onStartup hooks for module server startup

### LINKED ISSUE #2503: Make it possible for modules to contain code that is executed on server startup
It should be possible for a module to have code that is automatically invoked when the server starts. This could facilitate initial database setup, start periodic tasks, or other features.

To do this we would need to find a hook which can be found by the code analyzer/generator and include it into the generated Protocol class. One way would be to create a `Module` class with a method that can be overridden:

```dart
class MyModule extends Module {
  @override
  Future<void> onStartup(Session session) {
    // Perform startup tasks
  }
```

abdulawalarif:
Adds a package-scoped `Module` base class with an `onStartup` hook so modules (and server packages) can run post-migration initialization automatically. The CLI discovers at most one default-constructable concrete `Module` per package and generates `Endpoints.onStartup`. Hosts invoke the local module (if any), then nested modules once each in flat `module.name` order, after migrations/Redis and before Insights/API servers. The hook is skipped on the maintenance + apply-migrations early-exit path; thrown errors fail `start()`.

Auth/route configuration and `onConfigure` / `onShutdown` are intentionally out of scope for this PR.

Fixes #2503.

## Pre-launch Checklist

## Breaking changes

None. This is additive: packages without a `Module` subclass keep the existing no-op behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@ To allow easily building offline-first Flutter apps, the experimental `database:
 
 #### Server, web server and client:
 
+- feat: Adds `Module.onStartup` hooks invoked after migrations and before servers start.
 - feat: Ensures at-least-once semantics for future calls execution.
 - feat: Adds dedicated support for recurring future calls.
 - feat: Exposes configuration options for finding and deleting broken future calls on server startup.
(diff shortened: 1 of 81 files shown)


marcelomendoncasoares:
See https://github.com/serverpod/serverpod/pull/5763#issuecomment-5801382992.

PR #5560, opened 2026-08-07
refactor: Rework exception hierarchy for ServerpodClientException

Crazelu:
Reworks the exception hierarchy for `ServerpodClientException` and removes hacky way of dealing with network errors using -1 status code.

Part of #3283 

## Pre-launch Checklist

## Breaking changes

- The base `ServerpodClientException` no longer has a `statusCode` field.
- `ServerpodClientNetworkException` is now thrown in place of `ServerpodClientException` with -1 statusCode.


packages/serverpod_client/lib/src/serverpod_client_exception.dart
```diff
@@ -5,21 +5,38 @@ class ServerpodClientException implements Exception {
...
 
-  /// Creates a new [ServerpodClientException].
-  const ServerpodClientException(this.message, this.statusCode);
+  /// Creates a new [ServerpodClientHttpException].
+  const ServerpodClientHttpException(super.message, this.statusCode);
 
   @override
   String toString() {
-    return ('ServerpodClientException: ${message.trim()}, statusCode = $statusCode');
+    return ('ServerpodClientHttpException: ${message.trim()}, statusCode: $statusCode');
```
marcelomendoncasoares: Suggestion: Unnecessary parentheses (on all of them).

```suggestion
    return 'ServerpodClientHttpException: ${message.trim()}, statusCode: $statusCode';
```

packages/serverpod_client/lib/src/serverpod_client_shared.dart
```diff
@@ -296,10 +296,10 @@ abstract class ServerpodClientShared extends EndpointCaller {
             ServerpodClientForbidden(),
           OpenMethodStreamResponseType.invalidArguments =>
             ServerpodClientBadRequest(),
-          OpenMethodStreamResponseType.success => ServerpodClientException(
-            'Unknown error, data: $e',
-            -1,
-          ),
+          OpenMethodStreamResponseType.success =>
+            ServerpodClientNetworkException(
```
FXschwartz: Up for discussion but I feel like a plain `ServerpodClientException('Unknown error, data: $e')` would be a better.  Since this isn't really a network error, just an `OpenMethodStreamException` that came back with a success response type. 

The old -1 was only there because the base class required a `statusCode`, but now that it doesn't, I think `ServerpodClientException` seems like a better fit.

packages/serverpod_client/lib/src/serverpod_client_io.dart
```diff
@@ -75,10 +75,10 @@ class ServerpodClientRequestDelegateImpl
 
       return data;
     } on SocketException catch (e) {
-      throw ServerpodClientException(e.toString(), -1);
+      throw ServerpodClientNetworkException(e.toString());
```
FXschwartz, line 79: Since this is the main change of the PR it'd be good to have a test that actually hits this path. Something like a client pointed at a closed port asserting it throws `ServerpodClientNetworkException`. Right now I don't see anything on the branch that exercises the new type.

packages/serverpod_client/test/exception_from_test.dart
```diff
@@ -63,15 +63,18 @@ void main() {
         statusCode: 499,
       );
 
-      test('then a base ServerpodClientException is created', () {
-        expect(exception, isA<ServerpodClientException>());
-        expect((exception as ServerpodClientException).statusCode, 499);
+      test('then a base ServerpodClientHttpException is created', () {
+        expect(exception, isA<ServerpodClientHttpException>());
```
FXschwartz: Was the removal of the statusCode (499) coverage on purpose? It would be nice to keep that since it's the only coverage for that.

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
```diff
@@ -131,17 +131,22 @@ void main() async {
         var key = 'username-4711:password-4711';
         await incorrectAuthKeyManager.put(key);
 
-        ServerpodClientException? clientException;
+        var clientException;
```
FXschwartz: It would be nice to not have this be dynamic. Since the other tests in this PR moved to `expectLater` + `throwsA`, can we do the same here and drop the try/catch?

```dart
await expectLater(
  () async => await client.echoRequest.echoAuthenticationKey(),
  throwsA(
    isA<ServerpodClientBadRequest>()
      .having((e) => e.statusCode, 'statusCode', equals(400))
      .having((e) => e.message, 'message', startsWith('Bad request: ')),
  ),
);
```
Same at line 232.

packages/serverpod_client/lib/src/serverpod_client_exception.dart
```diff
@@ -5,21 +5,38 @@ class ServerpodClientException implements Exception {
   /// Error message sent from the server.
   final String message;
 
+  /// Creates a new [ServerpodClientException].
+  const ServerpodClientException(this.message);
+
+  @override
+  String toString() {
+    return ('ServerpodClientException: ${message.trim()}');
```
FXschwartz: nit: with the names hardcoded, subclasses print the wrong one. `ServerpodClientBadRequest().toString()` gives `ServerpodClientHttpException: ....` 

Using `$runtimeType: ${message.trim()}` here and in the override below should fix that and remove the duplication.

tests/serverpod_test_server/test_e2e/endpoint_login_hierarchy_test.dart
```diff
@@ -33,7 +33,7 @@ void main() {
           await expectLater(
             () async => await client.myLoggedIn.echo('hello'),
             throwsA(
-              isA<ServerpodClientException>()
+              isA<ServerpodClientHttpException>()
                   .having(
                     (e) => e.message,
                     'statusCode',
```
FXschwartz: Very minor nit and pre-existing: but the matcher description says 'statusCode' but it's reading e.message, same in the four other spots in this file.

FXschwartz (changes requested):
#3283 mentions **all** client exceptions deriving from a common base. Do we also need to handle `MethodStreamException` family and `ServerpodClientGetEndpointException`?

packages/serverpod_client/lib/src/serverpod_client_exception.dart
```diff
@@ -5,21 +5,38 @@ class ServerpodClientException implements Exception {
   /// Error message sent from the server.
```
marcelomendoncasoares, line 5: Discussion: Should we seal it? It provides a better UX when handling exceptions, at the cost of making every new exception a breaking change. However, these exceptions have been stable for a very long time already.

packages/serverpod_client/lib/src/serverpod_client_exception.dart
```diff
@@ -5,21 +5,38 @@ class ServerpodClientException implements Exception {
...
+}
+
+/// [Exception] thrown when a network error occurs.
+class ServerpodClientNetworkException extends ServerpodClientException {
+  /// Creates a new [ServerpodClientNetworkException].
+  const ServerpodClientNetworkException(super.message);
+}
+
+/// [Exception] thrown when HTTP errors occur in communication with the server.
+class ServerpodClientHttpException extends ServerpodClientException {
```
marcelomendoncasoares: Comment: If we decide to seal the base, we should also seal this one.

packages/serverpod_client/lib/src/serverpod_client_shared.dart
marcelomendoncasoares: Maybe we need a `ServerpodClientUnknownException` to handle these cases - part of the double purpose the `-1` was being used to.

marcelomendoncasoares (changes requested):
Few additional comments.

packages/serverpod_client/test/exception_from_test.dart
Crazelu: That was an oversight. I'll bring it back

packages/serverpod_client/lib/src/serverpod_client_shared.dart
Crazelu: @marcelomendoncasoares I agree `ServerpodClientUnknownException` will be clearer.

packages/serverpod_client/lib/src/serverpod_client_exception.dart
Crazelu: Another cost of using sealed here is that the exceptions can not be extended or implemented outside our library. We risk having similar issues with `DatabaseExceptions` where third party libraries can't create specified exceptions for their own use.

tests/serverpod_test_server/test_integration/auth_core/auth_header_test.dart
Crazelu: Done in [b6ba7ba](https://github.com/serverpod/serverpod/pull/5560/commits/b6ba7ba36496bf34c124a3905dc9bfb1c2204014)

packages/serverpod_client/test/exception_from_test.dart
Crazelu: Done in [74a01cb](https://github.com/serverpod/serverpod/pull/5560/commits/74a01cbd4c64320457f23754a4962888e359b981)

packages/serverpod_client/lib/src/serverpod_client_shared.dart
Crazelu: Added `ServerpodClientUnknownException` in [a5fd764](https://github.com/serverpod/serverpod/pull/5560/commits/a5fd76457f5ee6eefdff11d7959a24bc4730cb8f)

packages/serverpod_client/lib/src/serverpod_client_io.dart
Crazelu: Done in [e23471e](https://github.com/serverpod/serverpod/pull/5560/commits/e23471e26fde90ce9d8c9e55a838c767aafd40f4)

packages/serverpod_client/lib/src/serverpod_client_exception.dart
Crazelu: Done in [e7db2ae](https://github.com/serverpod/serverpod/pull/5560/commits/e7db2ae2cc060d72b8efcade966678d7fac59f2a)

tests/serverpod_test_server/test_e2e/endpoint_login_hierarchy_test.dart
Crazelu: Done in [614813f](https://github.com/serverpod/serverpod/pull/5560/commits/614813f835ecac667622edb9d1e4f9c3b405b8f4)

Crazelu:
> #3283 mentions **all** client exceptions deriving from a common base. Do we also need to handle `MethodStreamException` family and `ServerpodClientGetEndpointException`?

It does seem valuable to do that.

Crazelu:
@FXschwartz 
MethodStreamException and ServerpodClientGetEndpointException now derive from ServerpodClientException.
[5b7e87f](https://github.com/serverpod/serverpod/pull/5560/commits/5b7e87f16fd0d737458ca5b8a3cffec2ce1c4537)

packages/serverpod_client/lib/src/serverpod_client_exception.dart
marcelomendoncasoares: The question is whether this is a problem for the client. On the database it became a problem because the `databaseInterceptor` opens the possibility of the `Database` to be overridden with a custom implementation, but do we have a use cases for users to be wrapping the `Client` and throwing exceptions manually?

Crazelu: Right now it is not a problem especially because we do not allow for implementations of `ServerpodClientRequestDelegate` to be provided anywhere. That is one critical part where, if enabled, custom implementations could use extended exceptions defined outside our library.

packages/serverpod_client/lib/src/method_stream/method_stream_manager_exceptions.dart
```diff
@@ -1,14 +1,12 @@
-import 'package:serverpod_client/serverpod_client.dart';
+part of '../serverpod_client_exception.dart';
 
 /// Exceptions thrown by the [ClientMethodStreamManager].
-abstract class MethodStreamException implements Exception {
+abstract class MethodStreamException extends ServerpodClientException {
```
marcelomendoncasoares: Discussion: I'm having a second thought on this change. The `ServerpodClientException` is generic and suitable for happening on any interaction with the client. But the `MethodStreamException` is specific to streaming methods. Specially now that the `ServerpodClientException` chain is sealed, this means that users will be expected to treat `MethodStreamException` on all places they catch a `ServerpodClientException`, which is incorrect. Maybe they should not be mixed as part of the same chain.

marcelomendoncasoares (changes requested):

packages/serverpod_client/lib/src/serverpod_client_get_endpoint_exception.dart
```diff
@@ -0,0 +1,39 @@
+part of 'serverpod_client_exception.dart';
+
+/// Thrown if not able to get an endpoint on the client by type.
+class ServerpodClientGetEndpointException extends ServerpodClientException {
+  /// Creates an Endpoint Missing Exception.
+  const ServerpodClientGetEndpointException(super.message);
+
+  @override
+  String toString() => message;
+}
```
marcelomendoncasoares: Discussion: On the same idea as before, this is an exception that can happen only when working with abstract endpoints. Should we remove from the chain to avoid pushing on every catch of a `ServerpodClientException`? Also worth sealing this one.

marcelomendoncasoares (changes requested):

FXschwartz (approved):
All my comments have been addressed and pending @marcelomendoncasoares comments & CI failures being resolved it looks great to me.

packages/serverpod_client/lib/src/serverpod_client_get_endpoint_exception.dart
Crazelu: Having looked at the usage of `ServerpodClientGetEndpointException`, I agree that we should remove it from the chain. It is unlikely to hit using the generated client. The most likely way to hit it is using `getEndpointOfType` typically in a module.

packages/serverpod_client/lib/src/method_stream/method_stream_manager_exceptions.dart
Crazelu: I've had another look at this and I agree about the specificity of `MethodStreamException`.
I'll remove from the chain

packages/serverpod_client/lib/src/serverpod_client_get_endpoint_exception.dart
Crazelu: Done in [652806f](https://github.com/serverpod/serverpod/pull/5560/commits/652806f97d002bf3b8d3295cfd9f7d40393e8921)

packages/serverpod_client/lib/src/method_stream/method_stream_manager_exceptions.dart
Crazelu: Done in [652806f](https://github.com/serverpod/serverpod/pull/5560/commits/652806f97d002bf3b8d3295cfd9f7d40393e8921)

packages/serverpod_client/lib/src/serverpod_client_browser.dart
```diff
@@ -56,7 +58,9 @@ class ServerpodClientRequestDelegateImpl
       return data;
     } on http.ClientException catch (e) {
       var message = 'Unknown server response code. ($e)';
-      throw ServerpodClientException(message, -1);
+      throw ServerpodClientNetworkException(message);
+    } on TimeoutException {
+      throw const ServerpodClientNetworkException('Request timed out');
```
marcelomendoncasoares: Suggestion(NIT): Maybe have the `TimeoutException` mentioned in this exception so that users can differentiate that it is a Dart-side timeout (both here and on the other similar throw).

marcelomendoncasoares (approved):
LGTM! Only one minor nit on the text of the throw on `TimeoutException`, but already leaving it approved!

PR #5563, opened 2026-08-07
feat: Build, publish and use the `ghcr.io/serverpod/postgres:16` Postgres image

### LINKED ISSUE #5304: Publish official Serverpod PostgreSQL Docker image with PostGIS and pgvector to `ghcr.io`
## Summary

As a follow-up to PR #5137 (PostGIS geography type support), the Serverpod test infrastructure now uses a custom `Dockerfile.postgres` that builds a PostgreSQL image with both **pgvector** and **PostGIS** extensions installed.

To avoid requiring users and CI pipelines to build this image locally every time, an official pre-built image should be published to Docker Hub under an official Serverpod account.

## Motivation

- PR #5137 introduced a custom `Dockerfile.postgres` (based on `pgvector/pgvector:pg16`) to add PostGIS support.
- All test docker-compose files (`tests_integration`, `tests_e2e`, `tests_e2e_migrations`, `tests_flutter`, `docker-local`) were updated to build this image locally.
- Building the image on every developer machine and CI run adds unnecessary overhead.
- An official Serverpod-maintained image on Docker Hub would improve developer experience and reduce setup friction.

## Proposed Solution

- Use the GitHub Container Registry.
- Publish a pre-built image based on `pgvector/pgvector:pg16` with `postgresql-16-postgis-3` installed.
- Update all relevant docker-compose files to reference the published image instead of building locally.
- Set up automation (e.g., GitHub Actions) to rebuild and push the image when the Dockerfile changes.

## References

- PR #5137: https://github.com/serverpod/serverpod/pull/5137
- Related comment: https://github.com/serverpod/serverpod/pull/5137#discussion_r3364480580

## Requested by

@marcelomendoncasoares

marcelomendoncasoares:
Closes: #5304

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/.github/workflows/publish-postgres-image.yaml
+++ b/.github/workflows/publish-postgres-image.yaml
@@ -1,0 +1,247 @@
+name: Publish PostgreSQL image
+
+# Builds and publishes ghcr.io/serverpod/postgres, the official Serverpod
+# PostgreSQL image (PostgreSQL + pgvector + PostGIS) used by the docker-compose
+# file of newly created projects and by this repo's own docker test stacks.
+# Recipe: docker/postgres/Dockerfile.
+#
+# Publishing is triggered by a push that touches the recipe, on ANY branch, and
+# by workflow_dispatch. Branch pushes publish on purpose: the compose files that
+# consume this image - here and in newly created projects - are useless until it
+# exists, so the PR that introduces or changes the recipe has to be able to
+# publish before it lands. Dispatch alone cannot cover that case, since GitHub
+# only offers workflow_dispatch for workflows already present on main.
+#
+# What protects the rolling tags is the smoke gate below, not the branch name;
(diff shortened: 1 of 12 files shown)


nielsenko (approved):
The linux image we build for serverpod-embedded-postgres is way better tested for our use than this is (`smoke.sh`). I still think it would be better to simply download and bundle that.

Anyway, I don't see anything wrong with this, but then I'm no docker expert.

marcelomendoncasoares:
> The linux image we build for serverpod-embedded-postgres is way better tested for our use than this is (`smoke.sh`). I still think it would be better to simply download and bundle that.
> 
> Anyway, I don't see anything wrong with this, but then I'm no docker expert.

I agree, but I did not increase the coverage intentionally. On the docker image we're using the stablished pattern to provide the extensions, which is different from the binary path, where can bump into a lot more trouble because we're compiling.

PR #5565, opened 2026-08-07
fix: Make shared-package logs visible in the CLI

### LINKED ISSUE #5561: Logs emitted from `serverpod_database` through `serverpod_cli` calls are discarded
### Describe the bug

Logs emitted from the `serverpod_database` reports diagnostics through the global `log` from `package:serverpod_shared/log.dart` (a re-export of `serverpod_logging`). That global `Log` forwards to a global `MultiLogWriter` which **starts empty** - entry points are expected to register a writer on it, as the library documents: *"entry points configure logging by mutating `logWriter`, not by replacing"* (`serverpod_logging/src/global_log.dart:14`, `:20`).

- `packages/serverpod` does this. `serverpod.dart:55` initializes `ServerpodLogSetup.installDefaults()`, which calls `addLogWriter(...)` → `logWriter.add(...)` (`log_manager/serverpod_logging.dart:60`).
- `serverpod_cli` never does. `initializeLogger()` (`tools/serverpod_cli/lib/src/util/serverpod_cli_logger.dart:27`) builds a `ServerpodCliLogger` owning a *private* `Log` and writer. The global `logWriter` holds zero writers for the entire CLI process.

**Net effect: every `log.*` call in `serverpod_database` is discarded when that code runs under the CLI.** 12 call sites today:

| File | Lines |
|---|---|
| `src/migrations/migration_manager.dart` | 104, 117, 161, 289, 323, 364 |
| `src/migrations/apply_migrations.dart` | 32, 42, 44, 56 |
| `src/interface/analyzer.dart` | 72 |
| `src/adapters/sqlite/database_provider.dart` | 28 |

These are live CLI paths: `CliMigrationRunner` reaches `MigrationManager` from both `serverpod migrate` (`commands/migrate.dart`) and `serverpod start` (`commands/start.dart`).

### To reproduce

Verified end to end on a freshly created project.

```bash
serverpod create logrepro --template server --no-interactive
cd logrepro/logrepro_server

# Corrupt the SQL the first migration applies. On a fresh database
# MigrationManager takes the `fromVersion == null` branch
# (migration_manager.dart:240) and applies definition.sql, NOT migration.sql.
sed -i '1i SELECT * FROM "table_that_does_not_exist_xyz";' \
  migrations/<version>/definition.sql

serverpod migrate
```

Actual output (exit code **1**):

```
stdout: Applying migrations for "logrepro" (mode: development)...
stderr: ERROR: Migration failed: DatabaseQueryException: { message: Table not found,
        have you applied the database migration? (relation
        "table_that_does_not_exist_xyz" does not exist), code: 42P01, position: 56 }
        <~40 lines of raw Dart stack trace>
```

`grep -c "Failed to apply migration"` → **0** in both stdout and stderr.

The stack trace lands on `migration_manager.dart:283`, the `unsafeSimpleExecute` inside the `try` whose `catch` at `:288` runs `log.error('Failed to apply migration ${code.version}.', error: e)` and rethrows. To confirm the catch really executes, add `print('CATCH-REACHED for ${code.version}');` on the line directly below that `log.error` and re-run:

```
Applying migrations for "logrepro" (mode: development)...
CATCH-REACHED for 20260806161450006
ERROR: Migration failed: DatabaseQueryException: { message: Table not found,
```

Same catch block, same iteration, adjacent lines — `print` is visible, `log.error` is not.

Minimal confirmation of the mechanism alone, no database needed — a script in `tools/serverpod_cli`:

```dart
import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart' as cli;
import 'package:serverpod_shared/log.dart' as shared;

Future<void> main() async {
  cli.initializeLogger();
  cli.log.warning('A: via the CLI logger');       // prints
  shared.log.warning('B: via the global log');    // silent
  shared.log.error('C: via the global log');      // silent
  print('global log level = ${shared.log.logLevel}'); // LogLevel.info
  await cli.closeLogger();
}
```

Only `A` reaches the terminal. The global level is `info`, so `B` and `C` are not level-filtered — they reach an empty writer chain.

### Expected behavior

Log records emitted by shared packages should reach the CLI's terminal output when that code runs under `serverpod_cli`, honoring the CLI's verbosity flags (`-v`, `--quiet`), just as they reach the pod's log when the same code runs in a server.

Concretely for the case above: the failure *is* reported (the CLI's own handler prints the `DatabaseQueryException`), but **which migration version failed is lost** — `20260806161450006` appears nowhere in the output, and that is precisely what the discarded message carries. Other diagnostics vanish entirely, including `migration_manager.dart:364`'s "database does not match the target database" report with its *"Did you forget to run `serverpod generate`…"* hint, and `:161`'s wrong-`Protocol`-class module-name mismatch warning (#4754).

### Library version

Serverpod `4.0.0-beta.1` on `main` @ `fd3e61ae2`.

### Platform information

N/A

### Additional context

Proposed fix: register the CLI's writer on the global `logWriter` in `initializeLogger()`, and remove it in `closeLogger()`, mirroring the add/remove `ServerpodLogSetup` already does at `serverpod_logging.dart:60` / `:109`.

Three things a PR needs to handle:

1. **Log level syncing.** The global `Log` carries its own `logLevel` (defaults to `LogLevel.info`), independent of `ServerpodCliLogger.logLevel`. Without syncing it in the `logLevel` setter, `-v` / `--quiet` won't affect forwarded records.
2. **Double output.** Anything logging through both the CLI logger and the global `log` would print twice.
3. **Output review.** Messages invisible for a long time will start appearing, especially during `serverpod start`.

Supporting evidence that the global writer is the intended channel: tests needing these records wire it manually — `tests/serverpod_test_server/test_integration/migration_manager/migration_module_name_mismatch_test.dart:93` and `tests/serverpod_test_sqlite/.../runtime_parameters_test.dart:12` both call `logWriter.add(testWriter)`.

Related: #5044 adds a warning when empty trailing migration directories are pruned. Its client-side store passes the CLI logger explicitly so that warning is visible, but the server-side store logs through the global `log` — so **that warning is currently invisible during `serverpod create-migration`**, the main scenario it was written for. A `logWarning` constructor parameter was prototyped there and dropped in favor of fixing this centrally.

- [x] I want to work on this

marcelomendoncasoares:
Fixes: #5561

Log records emitted by packages shared with the server (`serverpod_database` above all) were silently discarded whenever that code ran under `serverpod_cli`. Nothing was misreported but the messages simply never appeared.

**What changes for users**

- `serverpod migrate` now names the migration that failed. Previously the exception was reported but the version was lost.
- `serverpod create-migration` now reports pruned empty migration directories — the warning added in #5044, which had never been visible in the scenario it was written for.
- The module-name mismatch warning (#4754) and the "database does not match the target database" report, with its `serverpod generate` hint, now reach the terminal.
- Failures report the exception once, not twice.
- These records honor `-v` and `--quiet` like the CLI's own output, and land in whichever surface is active — plain terminal, the `serverpod start` TUI, or the language server's silent mode.

**No duplicated logging.** The pod runs in its own process with its own log chain, so a record is produced in exactly one place and rendered exactly once. Nothing the CLI relays from a running pod passes back through the new path, and no existing log destination gained a second copy. Verified across the CLI process, the pod process, and the test processes that capture log output.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/test/util/serverpod_cli_logger_test.dart
+++ b/tools/serverpod_cli/test/util/serverpod_cli_logger_test.dart
@@ -20,14 +20,13 @@ Future<void> _logGlobally(
 void main() {
   late TestLogWriter writer;
 
+  setUp(() {
+    writer = TestLogWriter();
+    initializeLoggerWith(ServerpodCliLogger(writer));
+  });
   tearDown(closeLogger);
 
   group('Given a CLI logger is installed,', () {
-    setUp(() {
-      writer = TestLogWriter();
-      initializeLoggerWith(ServerpodCliLogger(writer));
-    });
-
     test(
       'when a shared package writes a warning to the global log, '
       'then the record reaches the CLI logger.',
@@ -97,9 +96,6 @@ void main() {
     'when a shared package writes a debug record to the global log, '
     'then it is filtered out.',
     () async {
-      writer = TestLogWriter();
-      initializeLoggerWith(ServerpodCliLogger(writer));
-
       await _logGlobally((log) => log.debug('Noisy detail.'));
 
       expect(writer.entries, isEmpty);
@@ -111,8 +107,6 @@ void main() {
     'when a shared package writes a debug record to the global log, '
     'then the record reaches the CLI logger.',
     () async {
-      writer = TestLogWriter();
-      initializeLoggerWith(ServerpodCliLogger(writer));
       log.logLevel = cli.LogLevel.debug;
 
       await _logGlobally((log) => log.debug('Noisy detail.'));
@@ -127,8 +121,6 @@ void main() {
     'when a shared package writes an error to the global log, '
     'then it is suppressed.',
     () async {
-      writer = TestLogWriter();
-      initializeLoggerWith(ServerpodCliLogger(writer));
       log.logLevel = cli.LogLevel.nothing;
 
       await _logGlobally((log) => log.error('Suppress me.'));
@@ -139,8 +131,6 @@ void main() {
 
   group('Given a CLI logger has been closed,', () {
     setUp(() async {
-      writer = TestLogWriter();
-      initializeLoggerWith(ServerpodCliLogger(writer));
       await closeLogger();
     });
 
--- a/tools/serverpod_cli/test/util/serverpod_cli_logger_test.dart
+++ b/tools/serverpod_cli/test/util/serverpod_cli_logger_test.dart
@@ -1,0 +1,166 @@
+import 'package:cli_tools/cli_tools.dart' as cli;
+import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart';
+import 'package:serverpod_logging_cli/serverpod_logging_cli.dart';
+import 'package:serverpod_shared/log.dart' as shared;
+// Imported with `show` because the barrel also exposes the global `log`, which
+// would be ambiguous against the CLI logger singleton of the same name.
+import 'package:serverpod_shared/log_io.dart' show LogLevel, TestLogWriter;
+import 'package:test/test.dart';
+
+/// Writes [message] the way a package shared with the server does, and waits
+/// for it to travel through the global chain into the CLI logger.
+Future<void> _logGlobally(
+  void Function(shared.Log log) write,
+) async {
+  write(shared.log);
+  await shared.log.flush();
+  await log.flush();
+}
+
+void main() {
+  late TestLogWriter writer;
+
+  tearDown(closeLogger);
+
+  group('Given a CLI logger is installed,', () {
+    setUp(() {
+      writer = TestLogWriter();
+      initializeLoggerWith(ServerpodCliLogger(writer));
+    });
+
+    test(
+      'when a shared package writes a warning to the global log, '
+      'then the record reaches the CLI logger.',
+      () async {
+        await _logGlobally((log) => log.warning('Removed empty directory.'));
+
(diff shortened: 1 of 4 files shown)


tools/serverpod_cli/test/util/serverpod_cli_logger_test.dart
```diff
@@ -0,0 +1,166 @@
...
+    );
+  });
+
+  test(
+    'Given a CLI logger with the default log level, '
+    'when a shared package writes a debug record to the global log, '
+    'then it is filtered out.',
+    () async {
+      writer = TestLogWriter();
+      initializeLoggerWith(ServerpodCliLogger(writer));
```
FXschwartz: nit: `this writer = TestLogWriter(); initializeLoggerWith()` pair shows up 5 times in the file. Since `tearDown(closeLogger)` is already top-level, could hoist the pair into a top-level setUp, these three tests keep just their log level line and the closed-logger group's `setUp` becomes just `closeLogger()`.

FXschwartz (approved):
Just one minor nit on the test. Approving either way though!

tools/serverpod_cli/test/util/serverpod_cli_logger_test.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5565/commits/5960ae34ae257d6a08be18fab805bf317b863cdc.

PR #5568, opened 2026-08-09
fix: Preserve recognized-class deserialization errors

### LINKED ISSUE #5529: Deserialization errors are swallowed while probing generated protocols
### Describe the bug

Generated protocols use `FormatException` to signal both an unknown `className` and malformed data for a recognized class. Protocol forwarding catches this exception to try another protocol, which can swallow the real data error and replace it with a misleading `No deserialization found` message.

```dart
// tests/serverpod_test_shared_module/serverpod_test_shared_module_server/lib/src/generated/protocol.dart:111-125 (PR #5520)
  @override
  dynamic deserializeByClassName(Map<String, dynamic> data) {
    var dataClassName = data['className'];
    if (dataClassName is! String) {
      return super.deserializeByClassName(data);
    }
    try {
      return _i3.Protocol().deserializeByClassName(data);
    } on FormatException catch (_) {}
    if (dataClassName.startsWith('serverpod.')) {
      data['className'] = dataClassName.substring(10);
      return _i2.Protocol().deserializeByClassName(data);
    }
    return super.deserializeByClassName(data);
  }
```

### To reproduce

1. Define a shared-package model containing a `dynamic` field.
2. Deserialize it through a forwarding module protocol using its recognized `className` but malformed dynamic-field data.
3. The shared protocol throws the relevant `FormatException`, but the forwarding protocol catches it and eventually throws `No deserialization found for type named ...` instead.

### Expected behavior

Only an unknown `className` should trigger protocol fallback. Errors raised while deserializing a recognized class should propagate unchanged.

### Library version

Latest main on 4.0.0-beta.1, unreleased.

### Platform information

Any

### Additional context

A solution can be to introduce a dedicated unknown-class-name exception (like `DeserializationTypeNotFoundException`, preferably extending `FormatException` for compatibility), and throw it only when `deserializeByClassName` cannot find a handler. Generated fallback paths should catch that specific exception instead of `FormatException`.

- [ ] I want to work on this

Emrecil69:
## Summary

* Preserve deserialization errors for recognized classes instead of silently falling back to raw JSON maps.
* Keep the generic fallback behavior for genuinely unrecognized dynamic types.
* Update generated protocol code and add regression coverage for direct and shared-module dynamic fields.

## Problem

Dynamic field deserialization caught errors raised while deserializing recognized classes and returned the original JSON value. This hid malformed payloads and allowed incorrectly typed map values to propagate.

## Solution

The generated protocol now distinguishes an unrecognized class from a recognized class whose value cannot be deserialized. Only unrecognized classes use the generic JSON fallback; actual deserialization failures are preserved.

## Tests

* `dart test test/dynamic_field_serialization_test.dart`
* `dart test test/dynamic_shared_module_test.dart`
* `dart analyze --fatal-infos` for the affected packages and files
* `util/run_tests_serverpod_generate`

## Pre-launch checklist

* [x] This PR contains one focused bug fix.
* [x] The affected code was formatted with `dart format`.
* [x] The related issue is listed below.
* [x] Regression tests were added and the affected tests pass.
* [x] Generated protocol files were verified.

## Breaking changes

None.

Fixes #5529


Emrecil69:

Crazelu:
Thanks for contributing @Emrecil69.
I pushed some commits to bring this up to speed with the current state of the repo.

Crazelu (approved):
🚀

PR #5570, opened 2026-08-10
fix: Clears orphaned server-scope operations when serverpod start res…

### LINKED ISSUE #5540: 4.0.0.beta.1: TUI stream/operation counter never decrements for streams killed by hot restart, eventually crowding out the log
### Describe the bug

The TUI's stream indicator/line for this method keeps growing (e.g. "3 streams" → "4 streams" → ...) with every hot restart, and the accumulating STREAM ... lines end up pinned above the log, eventually crowding out the actual scrolling log until it's unreadable.

### To reproduce

1. Add a long-lived stream endpoint (e.g. Stream<Model> listenToUpdates(Session session, ...) async* { ... await for (...) { yield ...; } } using session.messages.createStream).
2. Run serverpod start, connect a client, and have it subscribe to the stream.
3. Trigger a hot restart via the TUI (R).
4. Repeat step 3 a few times.

### Expected behavior

A hot restart hard-kills the pod process, which implicitly terminates every stream that process was serving. The TUI should not keep counting/displaying those as still active afterward.

### Library version

serverpod_cli 4.0.0-beta.1

### Platform information

macOS
serverpod start (default --watch enabled)
Reproduced with a custom Stream<T> listenToUpdates(...) endpoint using session.messages.createStream(...)

### Additional context

Root cause (best guess, based on investigation):

A hot restart genuinely kills and respawns the pod process (confirmed: a static in-memory counter added to the stream method for debugging reset back to 1 after each restart, rather than continuing to increment).
serverpod_session_log's duration column stays permanently NULL for any stream that was still open at the moment of a restart (confirmed directly in the DB) - the pod dies before it can log completion, so the row is orphaned forever.
The TUI's own stream/operation counter does not come from querying serverpod_session_log (confirmed: manually deleting the stale NULL-duration rows from the DB and then hot-restarting again still showed the counter increasing) - it appears to be tracked in-memory in the (longer-lived) TUI process itself, likely via VM service extension events (scope_start/scope_end) forwarded from the pod. For stream endpoints specifically, the completion-side event apparently never fires when the pod is hard-killed mid-stream - so the TUI's counter never decrements for these, and grows by one with every restart for the rest of the dev session.
Confirmed this is a display-only issue, not an actual leak:

lsof -i :<port> shows exactly one established TCP connection between the running client and server, regardless of how high the TUI's displayed count climbs.
serverpod_session_log entries for streams that did close normally (client reconnects unrelated to a restart) get a real duration recorded correctly - only ones open at restart time stay NULL forever.
Impact: Purely cosmetic during normal use, but becomes a real developer-experience problem over a longer dev session with several hot restarts: the growing block of stacked stream lines eventually pushes the actual scrolling log output out of view.

- [ ] I want to work on this

nicowalter256:
Fixes the TUI's stream/operation counter permanently growing after a hot restart.

`WatchSession.forceRestart()` hard-kills the pod process to restart it, so any
scope that was still open at that moment (e.g. a live stream endpoint) never
gets its matching `scope_end` event - the process that would have sent it is
gone. Those entries stayed pinned in the TUI's `activeOperations` forever,
so the displayed stream/operation count only ever grew across restarts,
eventually crowding out the log.

This adds `clearOrphanedServerScopes`, called from `onServerStart` - which
already fires on every server boot, including a restart - to finalize any
operation still tracked from the previous process (as a failed completion,
matching the existing convention for `scope_end`) before the new process's
event listener attaches. CLI-driven actions (e.g. "Restarting server"
itself) are tracked separately and are untouched by this.

Fixes #5540

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/start.dart
+++ b/tools/serverpod_cli/lib/src/commands/start.dart
@@ -1369,6 +1369,9 @@ Future<void> _runTuiBackend({
         // so a degraded->running transition lights up the action buttons.
         holder.state.serverReady = true;
         holder.state.serverStartable = false;
+        // The previous process (if any) is gone, so none of its open scopes
+        // (e.g. a stream endpoint) can still emit their `scope_end`.
+        clearOrphanedServerScopes(holder);
         holder.markDirty();
         final vmService = server.vmService;
         if (vmService == null) return;
--- a/tools/serverpod_cli/lib/src/commands/start/tui/event_handler.dart
+++ b/tools/serverpod_cli/lib/src/commands/start/tui/event_handler.dart
@@ -10,6 +10,14 @@ import 'tab_model.dart';
 
 int _actionCounter = 0;
 
+/// IDs of [TrackedOperation]s currently open because of a `scope_start`
+/// server log event, as opposed to a CLI-driven action started by
+/// [runTrackedAction]. Tracked separately because [TrackedOperation] (from
+/// the external `serverpod_tui` package) carries no origin of its own, and
+/// [clearOrphanedServerScopes] must only finalize the former: CLI actions
+/// already self-complete through their own future.
+final Set<String> _serverScopeIds = {};
+
 /// Dispatches a structured server log event to the TUI state.
 void handleServerLogEvent(TuiAppStateHolder holder, Event event) {
   if (event.extensionKind != 'ext.serverpod.log') return;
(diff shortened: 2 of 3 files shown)


Crazelu:
Thanks for your contribution @nicowalter256 
This fell into our critical path so we took it over. Please feel free to contribute to other open issues.
Closing this now with #5637 merged.

PR #5571, opened 2026-08-10
refactor: Rework database exception hierarchy

### LINKED ISSUE #3283: Rework exception hierarchy in Serverpod.

Crazelu:
This PR allows `DatabaseException` (and `DatabaseQueryException`) to be extendable from outside the Serverpod library. Specialized exceptions for insert, update and delete are replaced by the generic `DatabaseException`.

Stacked on #5560 
Closes #3283 

## Pre-launch Checklist

## Breaking changes

- The specialized `DatabaseInsertRowException`, `DatabaseUpdateRowException` and `DatabaseDeleteRowException` implementations have been replaced with `DatabaseException`.


tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
```diff
@@ -504,15 +504,15 @@ void main() {
...
+        'when updating by non-existent id then an exception is thrown',
         () async {
           var updated = Types.db.updateById(
             session,
             999999,
             columnValues: (t) => [t.anInt(123)],
           );
 
-          expect(updated, throwsA(isA<DatabaseUpdateRowException>()));
+          expect(updated, throwsA(isA<DatabaseException>()));
```
FXschwartz: Since every query error is also a `DatabaseException` now, this passes even if the update fails for a completely unrelated reason. I think we should change this to match on the message to keep the old coverage:

```dart
expect(
  updated,
  throwsA(
    isA<DatabaseException>()
        .having((e) => e.message, 'message', contains('no rows updated')),
  ),
);
```
Same applies to the other tests switched over in this PR.

packages/serverpod_database/lib/src/concepts/exceptions.dart
```diff
@@ -1,34 +1,49 @@
 /// Exception thrown when an error occurs in the database.
-abstract base class DatabaseException implements Exception {
-  /// Returns the message of the exception.
-  String get message;
+class DatabaseException implements Exception {
+  /// Creates a new [DatabaseException] with the given [message].
+  DatabaseException(this.message);
+
+  /// Error message from the database.
```
FXschwartz: nit: These messages don't necessarily come from the database since the row guards in the adapters build them in Dart, and now that the constructor is public they can say anything. Maybe just something like:
```dart
/// A message describing the error.
```

packages/serverpod_database/lib/src/adapters/postgres/postgres_exceptions.dart
```diff
@@ -1,32 +1,15 @@
 part of 'database_connection.dart';
 
 final class _PgDatabaseQueryException extends DatabaseQueryException {
```
FXschwartz: nit: now that `DatabaseQueryException` is concrete, I don't think this subclass is adding anything since it's private and final so nothing can catch it by type. `fromServerException` could just be a function returning a `DatabaseQueryException`. Same for `_SqliteDatabaseQueryException`.

FXschwartz (changes requested):
Just a couple of comments and one general question:

The breaking changes section lists insert/update/delete, but `DatabaseUpsertRowException` is removed too, would it be worth adding since it was also part of the public API?

packages/serverpod_database/lib/src/adapters/postgres/postgres_exceptions.dart
marcelomendoncasoares: I agree. There is absolutely no need for us to extend the base exceptions for each dialect if we're not adding something specific to that dialect. Having the abstracted exceptions also favors the multi-dialect abstraction.

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
```diff
@@ -209,7 +209,7 @@ class PostgresDatabaseConnection
     );
 
     if (result.length != 1) {
-      throw _PgDatabaseInsertRowException(
+      throw DatabaseException(
         'Failed to insert row, updated number of rows is ${result.length} != 1',
       );
```
marcelomendoncasoares, line 212: Discussion: A plain `DatabaseException` everywhere makes seem like a pretty generic usage (like using `Exception`). Maybe we could have an extension of it that is a `DatabaseExecutionException` to differentiate from the `DatabaseQueryException` or a `DatabaseConnectionException` that we could need in the future - since they all bubble up from the same upper calls.

On the paths that we throw after asserting the result (!= 1, empty, etc.), we could also have a more specific `DatabaseUnexpectedResultException` that extends the `DatabaseExecutionException`.

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
```diff
@@ -371,9 +371,7 @@ class PostgresDatabaseConnection
     );
 
     if (updated.isEmpty) {
-      throw _PgDatabaseUpdateRowException(
-        'Failed to update row, no rows updated',
-      );
+      throw DatabaseException('Failed to update row, no rows updated');
```
marcelomendoncasoares: Comment: All of these would be `DatabaseUnexpectedResultException`, which is different from us rethrowing an exception from Postgres or SQLite that happened inside the engine.

packages/serverpod_database/lib/src/concepts/exceptions.dart
```diff
@@ -1,34 +1,49 @@
...
+  final String message;
 
   @override
   String toString() => 'DatabaseException: $message';
```
marcelomendoncasoares: Suggestion: Valid to use the runtime type here as well as @FXschwartz suggested on the other PR to avoid inconsistencies on the printed names of subclasses.

marcelomendoncasoares (changes requested):
Few more suggestions for discussion to make the exception more specific on some paths.

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
Crazelu: I agree with the general direction of your suggestion. It is in line with making the exceptions explain what went wrong rather than the domain where things went wrong. Plain `DatabaseException` makes that explanation very broad with its current usage. 

I'll adopt this!

Crazelu: `DatabaseUnexpectedResultException` seems to be redundant as it becomes the only subclass of `DatabaseExecutionException`.

We could use `DatabaseExecutionException` exclusively and add specific types that extend it when that need emerges

packages/serverpod_database/lib/src/adapters/postgres/postgres_exceptions.dart
Crazelu: Done in [0bea4d0](https://github.com/serverpod/serverpod/pull/5571/commits/0bea4d00be96b8912b64377e2fa873368af0450d)

packages/serverpod_database/lib/src/concepts/exceptions.dart
Crazelu: Done in [4b5b142](https://github.com/serverpod/serverpod/pull/5571/commits/4b5b142a53b38cd5086177fa8a4b85c431827e70)

packages/serverpod_database/lib/src/concepts/exceptions.dart
Crazelu: Done in [f287624](https://github.com/serverpod/serverpod/pull/5571/commits/f287624d339ca8c39c96811dd63f94c47843af4c)

tests/serverpod_test_server/test_integration/database_operations/crud/update_by_id_test.dart
Crazelu: Done in [9c1c64d](https://github.com/serverpod/serverpod/pull/5571/commits/9c1c64d5e61274af8aa5cc0b60d01ba554dc486f)

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
Crazelu: Added `DatabaseExecutionException` in [aa413ce](https://github.com/serverpod/serverpod/pull/5571/commits/aa413ce13475501bfb4228281bd0a089f8228484)

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
Crazelu: See https://github.com/serverpod/serverpod/pull/5571#discussion_r3812275033

FXschwartz (approved):
All my feedback has been addressed! Approved pending Marcelo's comments.

Crazelu:
@marcelomendoncasoares I've introduced `DatabaseUniqueViolationException` and `DatabaseForeignKey ViolationException ` in [48b3942](https://github.com/serverpod/serverpod/pull/5571/commits/48b394258c91bad0c3779d664338a0500493c13a) for the most relevant error codes.

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
```diff
@@ -1652,3 +1646,31 @@ class _TransactionCancelledException<R> implements Exception {
...
+  if (e is! SqliteException) {
+    int? code;
+    if ([
+      'recursive lock',
+      'LockError',
+    ].any((s) => e.toString().contains(s))) {
+      code = 6;
+    }
+    return DatabaseQueryException(e.toString(), code: code.toString());
+  }
```
marcelomendoncasoares, line 1669: Suggestion: We should have one specific exception here as well. It won't be used on Postgres, it is one very common exception on SQLite (caused by concurrent transactions, or not passing a transaction parameter with a transaction active).

packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart
marcelomendoncasoares: I actually made a confusion on my suggestion. The rename is still valid, but I have changed on https://github.com/serverpod/serverpod/pull/5571/commits/f978bf593a43f36cc6a7b569699663489b1bde0d to `DatabaseUnexpectedResultException` instead of `DatabaseExecutionException` to prevent a confusion with the also execution-related `DatabaseQueryException` (and to make clear that the result exceptions are post-execution, not from the underlying engine).

packages/serverpod_database/lib/src/concepts/exceptions.dart
marcelomendoncasoares: Suggestion: Now that we have a specific `DatabaseForeignKeyViolationException`, we should rename the `SqliteForeignKeyViolationException` to make clear that they point to different parts of the workflow. The latter is thrown when a migration fails to satisfy foreign key constraints (and can contain multiple violations at once).

tests/serverpod_test_server/test_integration/database_operations/crud/insert_ignore_conflicts_test.dart
```diff
@@ -90,20 +90,14 @@ void main() async {
...
             [UniqueData(number: 2, email: 'existing@serverpod.dev')],
           ),
-          throwsA(
-            isA<DatabaseQueryException>().having(
-              (e) => e.code,
-              'code',
-              PgErrorCode.uniqueViolation,
-            ),
-          ),
+          throwsA(isA<DatabaseUniqueViolationException>()),
```
marcelomendoncasoares: Suggestion: It is worth to make both checks (exception is `DatabaseUniqueViolationException` and `e.code` is correct). This is important in case users catch the plain `DatabaseQueryException` and act only on the code. same is valid for the `DatabaseForeignKeyViolationException`.

marcelomendoncasoares (changes requested):
Only a few more suggestions and we're good to go.

tests/serverpod_test_server/test_integration/database_operations/crud/insert_ignore_conflicts_test.dart
Crazelu: Done in [4b6c056](https://github.com/serverpod/serverpod/pull/5571/commits/4b6c056c351dfce573080001e21a3a3fd4d511f7)

packages/serverpod_database/lib/src/concepts/exceptions.dart
Crazelu: Done in [37161c6](https://github.com/serverpod/serverpod/pull/5571/commits/37161c64f6e23d9e71d027fca88c1a3eebbb3be9)

packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart
Crazelu: Done in [41b2c62](https://github.com/serverpod/serverpod/pull/5571/commits/41b2c6294cca18d87754eef267a4dcb039cf2124)

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5572, opened 2026-08-10
fix: Fix MCP server not being able to fetch logs when using `--no-tui`

### LINKED ISSUE #5531: Disabling the TUI in `serverpod start` breaks the MCP logs.
By disabling the TUI by running `serverpod start --no-tui` also disables/breaks the MCP server's `tail_server_logs` and `tail_flutter_logs` commands.

Error description by AI:

### Summary

When the server is started with `--no-tui`, the MCP tools `tailLogs` and `tailFlutterLogs` always return an error, because the callbacks that back them are only supplied from inside `_runTuiBackend`. The log history lives on the TUI's own state object, so with no TUI there is nothing to serve.

The MCP server itself starts fine and the other tools work, which makes this look like a configuration problem rather than a missing wire-up.

### Environment

- `serverpod_cli` **4.0.0-beta.1**
- macOS (arm64)
- Started via `serverpod start --no-tui`, with a Flutter app launched through the `flutter_apps` config

### Steps to reproduce

1. Start a Serverpod project with `serverpod start --no-tui`
2. Connect an MCP client to the CLI's MCP socket
3. Call `tailFlutterLogs` (or `tailLogs`)

### Expected

Recent Flutter app / server log lines.

### Actual

```
Flutter log history not available.
```

and for `tailLogs`:

```
Log history not available.
```

Both are returned with `isError: true`. The same calls work when the TUI is enabled.

### Root cause

`mcpGetLogHistory` and `mcpGetFlutterLogHistory` are parameters of the watch loop, and they are supplied in exactly one place in the file — inside `_runTuiBackend` (`lib/src/commands/start.dart:1362-1363`):

```dart
mcpGetLogHistory: () => holder.state.logHistory.toList(),
mcpGetFlutterLogHistory: (appId) =>
    holder.state.appLogTabFor(appId)?.lines.toList() ?? <String>[],
```

Both read from `holder.state`, i.e. the TUI's state holder. `useTui` is decided at `start.dart:140`:

```dart
final useTui = commandConfig.value(StartOption.tui) && stdout.hasTerminal;
```

and the non-TUI branch never passes either callback. They therefore arrive as `null` at the MCP socket (`start.dart:841-843`):

```dart
getLogHistory: mcpGetLogHistory,
getFlutterAppIds: () => [for (final app in flutterManager.apps) app.id],
getFlutterLogHistory: mcpGetFlutterLogHistory,
```

which trips the guards in `lib/src/commands/start/mcp_server.dart`:

```dart
// _tailLogs, line 235
final get = getLogHistory;
if (get == null) {
  return CallToolResult(
    content: [TextContent(text: 'Log history not available.')],
    isError: true,
  );
}

// _tailFlutterLogs, line 256
final getIds = getFlutterAppIds;
final getLines = getFlutterLogHistory;
if (getIds == null || getLines == null) { ... }
```

Note that for `tailFlutterLogs`, `getFlutterAppIds` *is* populated in both modes — it comes from `flutterManager`. Only `getFlutterLogHistory` is null, so the app is known but its output is not retained.

In short: the TUI is not just a renderer of the log buffer, it **is** the buffer.

### Why this is easy to misdiagnose

The neighbouring capabilities are wired from `flutterManager` rather than the TUI holder (`start.dart:845-846`):

```dart
getVmServiceUri: () => proxy?.httpUri.toString(),
getFlutterDtdUris: () => flutterManager.dtdUris,
```

So in `--no-tui` mode `getFlutterAppDtd` and the VM service resource work normally, and an agent can connect to DTD and use the Dart MCP server. Only the two log tools fail, which reads like a Flutter/DTD problem rather than a CLI wiring gap.

### Impact

Any non-interactive host is affected — CI, a wrapping IDE, or any tool that embeds `serverpod start`. `--no-tui` is required in those contexts because the TUI puts the terminal into raw mode, so there is no workaround from the caller's side: the log data is never collected.

This came up in an IDE that embeds `serverpod start --no-tui` and exposes the agent to MCP. The agent can reach DTD and the Dart MCP server, but can never see server or Flutter logs.

### Suggested fix

Move log retention out of the TUI and into the watch loop, so the buffer exists in both modes and the TUI renders it rather than owning it:

- keep a bounded ring buffer for server output and one per Flutter app, populated where the process output is already being read
- wire `getLogHistory` / `getFlutterLogHistory` next to `getFlutterDtdUris` in the `mcpSocket` setup, so they are mode-independent
- have `_runTuiBackend` read from that buffer instead of being its owner

That would also make `tailLogs` / `tailFlutterLogs` behave consistently regardless of how the server was started.

marcelomendoncasoares:
Fixes #5531

Moves log retention out of the TUI and into the watch loop to allow the MCP tools to work without the TUI.

The buffers behind the MCP `tail_server_logs` and `tail_flutter_logs` tools lived on the TUI's state object, and their callbacks were supplied only from `_runTuiBackend`, so with `--no-tui` they arrived as `null` and both tools always failed. A new `StartLogHistory` now owns a bounded server-entry history and a bounded raw-line buffer per Flutter app; the watch loop fills it — subscribing to each process's VM service `Extension` stream and supplying the Flutter processes' stdout/stderr sinks — and the TUI renders those same buffers instead of owning them. The log-history callbacks are now required when connecting a watch session, so the missing wiring can no longer compile.

## Pre-launch Checklist

## Breaking changes

None.


tools/serverpod_cli/lib/src/commands/start/log_history.dart
```diff
@@ -0,0 +1,345 @@
...
+        ? StackTrace.fromString(stackTrace)
+        : null,
+    metadata: data['metadata'] is Map
+        ? Map<String, Object?>.from(data['metadata'] as Map)
+        : null,
+  );
+}
+
+/// The [LogLevel] named by [level]; unknown names are treated as info.
+LogLevel parseLogLevel(String level) {
```
FXschwartz, line 251: Looks like the old copy of this in `tui/event_handler.dart` survived the move and nothing over there calls it anymore, and since that file imports `log_history.dart` the local copy just shadows this one. I think it can be deleted.

FXschwartz (changes requested):

tools/serverpod_cli/lib/src/commands/start/log_history.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5572/commits/187ab182251130d7f504e0bf70301f9f4f758144.

FXschwartz (approved):
LGTM!

PR #5580, opened 2026-08-12
fix: Resolve generated dart formatters with captured stderr

### LINKED ISSUE #5576: `serverpod create` displays an intermittent error message.
Right after "Running Serverpod generator" the TUI displays an intermittent error message. Screenshot below:

Image

Crazelu:
`dart_style` writes warnings to stderr when it [encounters some issues ](https://github.com/dart-lang/dart_style/blob/ce379a15fc8d30b4d955e24589897b183c476722/lib/src/config_cache.dart#L155) with `analysis_options.yaml`. This became more likely to happen with the introduction of `GeneratedDartFormatters`. Specifically the parts that try to resolve `pageWidth` and `trailingCommas` configurations from `analysis_options.yaml`. 

The resolution of formatters is now down in a fresh zone with overriden stderr and stdout (for future-proof safety) to capture the library's warnings and log them as debug messages.

Closes #5576 

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tools/serverpod_cli/lib/src/generator/dart_formatters.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart_formatters.dart
@@ -39,26 +39,25 @@ final class GeneratedDartFormatters {
   /// `dart_style` reports problems with a target package's
   /// `analysis_options.yaml` by writing to `stderr` itself, which corrupts the
   /// TUI. Those writes are captured and re-emitted through the logger instead.
+  ///
+  /// Anything written to `stdout` is captured the same way
+  /// but logged as debug output.
   static Future<void> resolve(GeneratorConfig config) async {
     final formatterStderr = _BufferedStdout();
+    final formatterStdout = _BufferedStdout();
 
     await IOOverrides.runZoned(
       () => _resolve(config),
       stderr: () => formatterStderr,
-      stdout: () => formatterStderr,
+      stdout: () => formatterStdout,
     );
 
-    // Output directories in the same package share an `analysis_options.yaml`,
-    // so a problem with it is reported once per directory. Report each distinct
-    // line once, as a single warning, keeping multi-line messages intact.
-    final warnings = <String>{};
-    for (final line in const LineSplitter().convert(formatterStderr.text)) {
-      if (line.trim().isEmpty) continue;
-      warnings.add(line);
+    if (formatterStdout.text.trim().isNotEmpty) {
+      log.debug(formatterStdout.text);
     }
 
-    if (warnings.isNotEmpty) {
-      log.warning(warnings.join('\n'));
+    if (formatterStderr.text.trim().isNotEmpty) {
+      log.warning(formatterStderr.text);
     }
   }
 
@@ -167,11 +166,12 @@ Future<bool> _directoryExists(Directory directory) async {
 
 /// A [Stdout] that buffers everything written to it in memory.
 ///
-/// Installed over `stderr` with [IOOverrides.runZoned] so that direct writes
-/// from within `dart_style` can be routed through the logger. The top-level
-/// `stderr` in `dart:io` resolves [IOOverrides.current] on every access, and
-/// the overrides are zone scoped, so the buffer stays in effect across the
-/// `await`s inside the library.
+/// Installed over `stdout` and `stderr` with [IOOverrides.runZoned] so that
+/// direct writes from within `dart_style` can be routed through the logger
+/// rather than corrupting the TUI. The top-level `stdout` and `stderr` in
+/// `dart:io` resolve [IOOverrides.current] on every access, and the overrides
+/// are zone scoped, so the buffers stay in effect across the `await`s inside
+/// the library.
 class _BufferedStdout implements Stdout {
   final StringBuffer _buffer = StringBuffer();
 
--- a/tools/serverpod_cli/lib/src/generator/dart_formatters.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart_formatters.dart
@@ -58,7 +58,7 @@ final class GeneratedDartFormatters {
     }
 
     if (warnings.isNotEmpty) {
-      log.debug(warnings.join('\n'));
+      log.warning(warnings.join('\n'));
     }
   }
 
--- a/tools/serverpod_cli/lib/src/generator/dart_formatters.dart
+++ b/tools/serverpod_cli/lib/src/generator/dart_formatters.dart
@@ -120,12 +120,17 @@ Future<DartFormatter> _formatterFor(
 ) async {
   final outputFile = File(p.join(directory, 'protocol.dart')).absolute;
 
-  // Package config discovery needs a path inside an existing directory. The
-  // generated output folder (e.g. lib/src/protocol) may not exist yet on the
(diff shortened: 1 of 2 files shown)


marcelomendoncasoares (changes requested):
Before we go on with this PR, it would be important to know what is the cause for these messages [being shown on the bootstrap tests](https://github.com/serverpod/serverpod/actions/runs/31435124817/job/93611204462) and [some users machines](https://github.com/serverpod/serverpod/issues/5576) (the original issue).

Image

Capture the message and show properly is good, but maybe we should log them as warning instead of debug, so they get shown during create and we get the chance to fix them properly (this one, and others that could happen in the future).

Crazelu:
> Before we go on with this PR, it would be important to know what is the cause for these messages [being shown on the bootstrap tests](https://github.com/serverpod/serverpod/actions/runs/31435124817/job/93611204462) and [some users machines](https://github.com/serverpod/serverpod/issues/5576) (the original issue).

This issue was caused by using a non-existing file path to attempt lookup for analysis_options.yaml. We already do some directory climbing to resolve the language version file but failed to supply that file for analysis_options lookup. This is fixed in [7094888](https://github.com/serverpod/serverpod/pull/5580/commits/70948886a16f9523d3b98095accf0f9a3c925e58).

> Capture the message and show properly is good, but maybe we should log them as warning instead of debug, so they get shown during create and we get the chance to fix them properly (this one, and others that could happen in the future).

Done in [87fdd0a](https://github.com/serverpod/serverpod/pull/5580/commits/87fdd0a5e20c63a59a770e34bf368f1f51fe864d)

tools/serverpod_cli/lib/src/generator/dart_formatters.dart
```diff
@@ -33,7 +35,34 @@ final class GeneratedDartFormatters {
...
+    // line once, as a single warning, keeping multi-line messages intact.
+    final warnings = <String>{};
+    for (final line in const LineSplitter().convert(formatterStderr.text)) {
+      if (line.trim().isEmpty) continue;
+      warnings.add(line);
+    }
+
+    if (warnings.isNotEmpty) {
+      log.warning(warnings.join('\n'));
+    }
```
marcelomendoncasoares: Question: Can't this lead to a loss on the order of the lines? Do we need to split-then-group or can we just pass it forward, only trimming to ensure we don't log empty content?

tools/serverpod_cli/lib/src/generator/dart_formatters.dart
```diff
@@ -33,7 +35,34 @@ final class GeneratedDartFormatters {
...
+  /// `dart_style` reports problems with a target package's
+  /// `analysis_options.yaml` by writing to `stderr` itself, which corrupts the
+  /// TUI. Those writes are captured and re-emitted through the logger instead.
   static Future<void> resolve(GeneratorConfig config) async {
+    final formatterStderr = _BufferedStdout();
+
+    await IOOverrides.runZoned(
+      () => _resolve(config),
+      stderr: () => formatterStderr,
+      stdout: () => formatterStderr,
```
marcelomendoncasoares: Question: Why capture the `stdout` as `formatterStderr`? Maybe we should split them to log one as warning (or even error) and the other as info.

Crazelu: Capturing `stdout` currently is a no-op as the `dart_style` library does not write anything to stdout. I added it as some kind of future-proofing in case that changes or a dependency of `dart_style` actually does that. So we don't end up here again.

I'd argue that we should ignore anything written to stdout instead of surfacing that to our TUI. We could potentially pollute our TUI with a library's messages.

tools/serverpod_cli/lib/src/generator/dart_formatters.dart
Crazelu: The order is preserved due to the use of a set for the warnings. Iteration is in insertion-order for the default [LinkedHashSet](https://api.dart.dev/dart-collection/LinkedHashSet-class.html).

We can pass the lines forward but that can result in multiple lines of the same message since they may be produced per output directory which is passed to `_formatterFor`.

The split-then-group mechanism enables deduplication.

tools/serverpod_cli/lib/src/generator/dart_formatters.dart
marcelomendoncasoares: Ok, let's remove the `stdout` capture then.

tools/serverpod_cli/lib/src/generator/dart_formatters.dart
```diff
@@ -130,3 +164,75 @@ Future<bool> _directoryExists(Directory directory) async {
...
+
+/// A [Stdout] that buffers everything written to it in memory.
+///
+/// Installed over `stderr` with [IOOverrides.runZoned] so that direct writes
+/// from within `dart_style` can be routed through the logger. The top-level
+/// `stderr` in `dart:io` resolves [IOOverrides.current] on every access, and
+/// the overrides are zone scoped, so the buffer stays in effect across the
+/// `await`s inside the library.
+class _BufferedStdout implements Stdout {
+  final StringBuffer _buffer = StringBuffer();
```
marcelomendoncasoares, line 176: Suggestion: If this is the same class that we use on the test, we can make it `internal` and reuse instead of duplicating.

tools/serverpod_cli/lib/src/generator/dart_formatters.dart
Crazelu: I made a second pass on this in [9feff5e](https://github.com/serverpod/serverpod/pull/5580/commits/9feff5e140854994a0358a4ca8ac1d759a9f2a8c).

The split-then-group step is now completely removed. All warnings are logged as-is. The deduplication effort is not really worth it.

tools/serverpod_cli/lib/src/generator/dart_formatters.dart
Crazelu: I made a second pass on this in [9feff5e](https://github.com/serverpod/serverpod/pull/5580/commits/9feff5e140854994a0358a4ca8ac1d759a9f2a8c).

`stdout` is now captured separately and logged as debug. `dart_style` doesn't currently write to `stdout` but if future versions do, we get that coverage with this change. And those messages won't show in the TUI unless when the command is run with the verbose flag. I maintain that `stdout` messages should not pollute the TUI. Just like we have for `pub get`.

`stderr` remains captured and logged as a warning. This is sufficient because even when warnings are logged, project creation still completes with sane defaults. Logging as error here seems counterintuitive especially as the project creation won't fail.

tools/serverpod_cli/lib/src/generator/dart_formatters.dart
Crazelu: This will result in a lot of changes in unrelated tests. Are you comfortable with that?

marcelomendoncasoares: No more need. On https://github.com/serverpod/serverpod/pull/5580/commits/9feff5e140854994a0358a4ca8ac1d759a9f2a8c you already removed the class from the test.

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5582, opened 2026-08-13
fix: Include the session key salt in the session key hash

nielsenko:
The per-session salt was generated, stored, and handed to validation, but never entered the digest, which covered only `secret + pepper`. The salt is now part of the digest.

Session key secrets are 32-byte CSPRNG output. The salt is defense-in-depth, but not security critical by itself.

The change is non-breaking. Hashes now carry a leading version byte (`[1] + sha512(secret + salt + pepper)`). Pre-fix rows (bare 64-byte saltless digest) still validate and are re-hashed in place on next use with their already stored salt. No user is signed out.

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/util/session_key_hash.dart
+++ b/modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/util/session_key_hash.dart
@@ -7,8 +7,29 @@ import 'package:serverpod_shared/serverpod_shared.dart';
 
 import '../business/server_side_sessions_config.dart';
 
+/// Outcome of validating a presented secret against a stored session key hash.
+enum SessionKeyHashValidation {
+  /// The secret did not match under any supported scheme or pepper.
+  invalid,
+
+  /// The secret matched a hash already stored in the current scheme.
+  valid,
+
+  /// The secret matched, but the stored hash uses a superseded scheme (the
+  /// pre-versioning saltless digest). The caller should re-hash via
+  /// [ServerSideSessionKeyHash.rehashSessionKeyHash] and persist it, upgrading
+  /// the row in place without dropping the session.
+  validButOutdated,
+}
+
 @internal
 final class ServerSideSessionKeyHash {
+  /// Length in bytes of a raw SHA-512 digest.
+  static const int _digestLength = 64;
+
+  /// The scheme written today: `[_saltedVersion] + sha512(secret + salt + pepper)`.
+  static const int _saltedVersion = 1;
+
   final int _sessionKeyHashSaltLength;
   final String _sessionKeyHashPepper;
   final List<String> _fallbackSessionKeyHashPeppers;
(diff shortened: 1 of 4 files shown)


modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/session/integration/server_side_sessions_authentication_handler_test.dart
```diff
@@ -488,4 +490,110 @@ void main() {
...
+      sessionId = row.id!;
+      sessionToken = buildServerSideSessionToken(
+        serverSideSessionId: sessionId,
+        secret: secret,
+      );
+    });
+
+    test(
+      'when it authenticates, then it succeeds and the stored hash is upgraded '
+      'in place to the versioned salted scheme.',
```
marcelomendoncasoares: Suggestion: Split this into two tests to avoid conflating multiple assertions that are independent - the fix could authenticate without upgrading or vice-versa.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/session/util/session_key_hash_test.dart
```diff
@@ -0,0 +1,229 @@
...
+      final secret = Uint8List.fromList(List.generate(32, (final i) => i));
+
+      // Any stored salt: the pre-fix digest never mixed it in.
+      final salt = Uint8List.fromList(List.generate(16, (final i) => 255 - i));
+
+      final legacyHash = _legacySaltlessHash(secret, 'test-pepper');
+
+      test(
+        'when it is validated with the right secret, then it is accepted as '
+        'outdated so the caller upgrades it.',
```
marcelomendoncasoares:
```suggestion
        'when it is validated with the right secret, '
        'then it is accepted as outdated so the caller upgrades it.',
```

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/session/util/session_key_hash_test.dart
```diff
@@ -0,0 +1,229 @@
...
+              salt: salt,
+            ),
+            SessionKeyHashValidation.invalid,
+          );
+        },
+      );
+
+      test(
+        'when it is re-hashed, then the result is versioned and validates as '
+        'current under the same salt.',
```
marcelomendoncasoares:
```suggestion
        'when it is re-hashed, '
        'then the result is versioned and validates as current under the same salt.',
```

marcelomendoncasoares: Suggestion: Split this as a group with two tests.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/session/util/session_key_hash_test.dart
```diff
@@ -0,0 +1,229 @@
...
+            ),
+            SessionKeyHashValidation.valid,
+            reason: 'The upgraded row must validate under the current scheme.',
+          );
+        },
+      );
+
+      test(
+        'when its pepper has since been rotated to a fallback, then it still '
+        'validates as outdated.',
```
marcelomendoncasoares:
```suggestion
        'when its pepper has since been rotated to a fallback, '
        'then it still validates as outdated.',
```

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/util/session_key_hash.dart
```diff
@@ -20,68 +41,136 @@ final class ServerSideSessionKeyHash {
...
+            hash,
+            (final pepper) =>
+                _legacySaltlessDigest(secret: secret, pepper: pepper),
+          )
+          ? SessionKeyHashValidation.validButOutdated
+          : SessionKeyHashValidation.invalid;
+    }
+
+    // Versioned rows carry their scheme in the first byte.
+    switch (hash[0]) {
```
marcelomendoncasoares, line 115: Suggestion: Return `invalid` if the `hash.length` is not valid to prevent a throw on `hash[0]`.

```suggestion
    if (hash.length != 1 + _digestLength) {
      return SessionKeyHashValidation.invalid;
    }

    // Versioned rows carry their scheme in the first byte.
    switch (hash[0]) {
```

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/session/util/session_key_hash_test.dart
```diff
@@ -0,0 +1,229 @@
...
+        expect(
+          first.salt,
+          isNot(second.salt),
+          reason: 'Precondition: each hash must get its own random salt.',
+        );
+
+        expect(
+          first.hash,
+          isNot(second.hash),
+          reason:
```
marcelomendoncasoares, line 49: Suggestion: Use byte-content matcher instead of identity-based to ensure they have different contents.


```suggestion
        expect(
          first.salt,
          isNot(orderedEquals(second.salt)),
          reason: 'Precondition: each hash must get its own random salt.',
        );

        expect(
          first.hash,
          isNot(orderedEquals(second.hash)),
          reason:
```

marcelomendoncasoares (changes requested):
One small suggestion to prevent unexpected server exceptions and a few suggestions on the tests.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/lib/src/session/util/session_key_hash.dart
nielsenko: Good catch

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/session/util/session_key_hash_test.dart
nielsenko: `isNot` wraps a non-matcher in `equals()`, and `equals` on a `List` already compares element-wise, so `isNot(second.salt)` was comparing bytes. Anyway, I have done the change to be more explicit.

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/session/util/session_key_hash_test.dart
nielsenko: Fixed

modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_server/test/session/util/session_key_hash_test.dart
nielsenko: Fixed

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5584, opened 2026-08-13
fix: Refuse to rotate a refresh token for a blocked auth user

nielsenko:
`blocked` was only checked at login (`createTokens`). `rotateRefreshToken` never looked up the user, so blocking a user closed the login path but left every session they already held running. 

Each refresh minted a new access token and, since expiry is measured from `lastUpdatedAt`, extended the refresh token's own expiry. A blocked user kept full scoped access for as long as they kept refreshing.

Now `rotateRefreshToken` now refuses rotation for a blocked user.
- Checked *after* the fixed secret, expiry, and rotating secret are validated, so it can't probe account state without holding the token.
- The token row is left in place. Lifting the block restores the session, and the sliding window keeps running so the token still expires on its own.


marcelomendoncasoares (approved):
LGTM! Already approving, despite requiring the Windows CI fix to get green.

PR #5585, opened 2026-08-13
ci: Pin pstools checksum (while we wait for chocolatey to update)

nielsenko:
Fix CI. There was a release of pstools today, and chocolatey has not updated their hash yet.


marcelomendoncasoares (approved):
LGTM! 🪟

PR #5586, opened 2026-08-13
chore: Ensure project build runs on a clean folder to avoid compilation issues

### LINKED ISSUE #5581: The default template's `build_flutter` command should do a `flutter clean`
When new packages are added, the app may appear to build cleanly, but it doesn’t run correctly. When deploying to Serverpod Cloud, the app silently fails to run with only cryptic exceptions in the logs. A fix is to run a clean before building the app to ensure we always have a fresh build. This update increases the build time by a few seconds, but it’s worth it, especially compared to having users chase hard-to-understand issues in production.

Here are proposed changes to the template (AI generated, the comment may not be required):

Image

marcelomendoncasoares:
Fixes: #5581

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/templates/pubspecs/templates/serverpod_templates/projectname_server/pubspec.yaml
+++ b/templates/pubspecs/templates/serverpod_templates/projectname_server/pubspec.yaml
@@ -22,28 +22,29 @@ serverpod:
     # Build the Flutter web app and copy it to the server's web directory.
     # Windows uses xcopy since flutter's --output flag doesn't work there.
     # See: https://github.com/flutter/flutter/issues/157886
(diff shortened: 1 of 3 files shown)


nielsenko (approved):
Does what it says, but I could be a bit worried about the performance hit for the 99% who don't have any problems.

marcelomendoncasoares:
> Does what it says, but I could be a bit worried about the performance hit for the 99% who don't have any problems.

I agree... Let's try a more conservative approach only running clean if build fails once.

nielsenko (approved):
LGTM 👍 

Eventually we should probably take more control of this instead of using the `run` (anything) command.

PR #5587, opened 2026-08-13
feat: Add dark theme and polish to the default Flutter app

### LINKED ISSUE #5577: Proposed changes to default app created by `serverpod create`
The default app currently doesn’t look very good, and it doesn't support a dark theme (which messes things up for some agents and App Studio).

Proposed changes will improve the style of the app without bloating the code too much:

https://github.com/serverpod/example/commit/3fbd77b07befa950b7318adff451afdc5607b923

marcelomendoncasoares:
The default app from `serverpod create` now follows the system light/dark theme and uses ColorScheme-aware colors, so it looks better and works in App Studio and agent tools. Greeting input uses an outlined field with a send action instead of a separate button.

Based on the proposed changes in https://github.com/serverpod/example/commit/3fbd77b07befa950b7318adff451afdc5607b923.

Fixes #5577.

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/templates/serverpod_templates/projectname_flutter/lib/main.dart
+++ b/templates/serverpod_templates/projectname_flutter/lib/main.dart
@@ -33,6 +33,16 @@ void main() async {
   runApp(const MyApp());
 }
 
+/// Builds a theme for the given [brightness].
+ThemeData _buildTheme(Brightness brightness) {
+  return ThemeData(
+    colorScheme: ColorScheme.fromSeed(
+      seedColor: Colors.blue,
+      brightness: brightness,
+    ),
+  );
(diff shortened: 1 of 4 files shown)


nielsenko (approved):
LGTM 👍

FXschwartz (commented):
@marcelomendoncasoares Looks like the auth variant still has hardcoded `snackbar` colors. Colors.green/Colors.red won't adapt in dark mode, so a project created with auth enabled still hits the problem. Is that something we would want to include in this PR or separately? Might be worth switching these to `ColorScheme` roles too.

PR #5590, opened 2026-08-13
test: Use proper Given-When-Then descriptions

nielsenko:
For the pedantic..


marcelomendoncasoares (approved):
Nothing like enforcing the pattern! Sorry for so many comments to break the line on the clause 😆

PR #5594, opened 2026-08-14
ci: Bump max Flutter version to 3.47.0

nielsenko:
Bump to latest stable flutter on ci


marcelomendoncasoares:
I'm also doing this on the workspaces PR, so we might not need the explicit bump before. Leave it here as draft for now, since I'm still solving a dependency hell issue because of the `storybook` app on the Flutter IDP package.

Issue #5595, opened 2026-08-14
Generate the `field` for relations that require the explicit FK


marcelomendoncasoares:
### Problem to Solve

When working with relations that require a declaration of the side that owns the FK, the user is required to explicitly declare the field instead of letting the generator handle it. This means that the user is responsible for keeping the type of the field in sync with the relation (nullability) and other side primary key type. It is also an asymmetry in the way we handle the foreign key by the user in object unidirectional relations.

### Proposal

Two ways of simplify this UX:

- Allow the user to declare only `field=` on the relation and generate this field if it does not exist. The field will have the type of the PK of the other table and be nullable if the relation is `optional`.
- Allow the user to declare a `fk` flag on the relation (e.g. `relation(name=abc, fk)`) to signal that this is the side that owns the FK. Similar to the `field=`, but for cases in which the desired FK field name follows the same convention as when the field is implicit on unidirectional relations.

### Use Case

Simplify the declaration of bidirectional relations reusing the existing mechanism for keeping the FK field name in sync with the relation and other table PK type.

### Alternatives

Continue declaring explicitly.

### Additional context

N/A

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [x] I want to work on this

PR #5596, opened 2026-08-14
feat: Generate the `field` for relations that require the explicit FK

marcelomendoncasoares:
Partially implements #5595. This PR is the first part of the issue, to support generating the FK field when the user writes `relation(field=`.

## Pre-launch Checklist

## Breaking changes

None.


tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -2928,4 +2965,21 @@ class Restrictions {
         .where((field) => field.jsonKey == jsonKey)
         .toList();
   }
+
+  String? _relationForeignKeyFieldName(RelationDefinition? relation) {
+    return switch (relation) {
+      ObjectRelationDefinition(:final fieldName) => fieldName,
```
FXschwartz, line 2987: This case misfires on the side that doesn't hold the FK: there `fieldName` is the local primary key name (id), not a foreign key. So `validateOptionalValue` errors on this valid model:

```yaml
class: Company
table: company
fields:
  id: UuidValue, defaultModel=random
  member: Member?, relation(name=company_member, optional)
```
An optional relation requires the foreign key field "id" to be nullable. And since the default int id is nullable, whether optional passes on the non-FK side ends up depending on the id type.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_relation_scope_test.dart
```diff
@@ -89,6 +89,68 @@ void main() {
     );
   });
 
+  test(
```
FXschwartz, line 92: Question on this one: The generated `postId` inherits the `serverOnly` scope, so from the client's perspective this is the same situation as the implicit relation we reject with "requires the relation to be optional".

Skipping the guard made sense when the user declared the field themselves, but for a generated field it's implicit. Is this the behavior we want?

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
```diff
@@ -357,7 +357,29 @@ class ModelDependencyResolver {
     String relationFieldName,
   ) {
     var field = classDefinition.findField(relationFieldName);
-    if (field == null) return;
+    if (field == null) {
+      var relationFieldType = relation.nullableRelation
+          ? referenceDefinition.idField.type.asNullable
+          : referenceDefinition.idField.type.asNonNullable;
+
+      field = SerializableModelFieldDefinition(
```
FXschwartz: nit: This mirrors the FK field construction in `_resolveImplicitDefinedRelation`, we could create a shared helper.

Also this one gets the `/// The foreign key of...` doc comment while the implicit one doesn't. Fine either way, just noting the asymmetry.

tools/serverpod_cli/lib/src/analyzer/models/yaml_definitions/class_yaml_definition.dart
```diff
@@ -132,10 +132,7 @@ class ClassYamlDefinition {
                   ValidateNode(
                     Keyword.optional,
                     keyRestriction: restrictions.validateOptionalKey,
-                    valueRestriction: BooleanValueRestriction().validate,
-                    mutuallyExclusiveKeys: {
-                      Keyword.field,
-                    },
+                    valueRestriction: restrictions.validateOptionalValue,
```
FXschwartz, line 148: Probably worth a line in the docs that `optional` + `field=` is now legal for a pre-existing field too, with the new rule that the field must then be nullable.

FXschwartz (changes requested):

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
marcelomendoncasoares: Good catch! Fixed at https://github.com/serverpod/serverpod/pull/5596/commits/8fc0e5829807b7f247d73787e7b514375907f510.

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5596/commits/b9a976351d7dbaf4d168631350d5002661b62537.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/field/field_relation_scope_test.dart
marcelomendoncasoares: This is a real bug, not desired behavior. Fixed at https://github.com/serverpod/serverpod/pull/5596/commits/79ba44bba84ec6934f2808797593d6eb7298ebf1.

tools/serverpod_cli/lib/src/analyzer/models/yaml_definitions/class_yaml_definition.dart
marcelomendoncasoares: Yes, I'll do this as follow-up!

marcelomendoncasoares: Already on the way (still working on it): https://github.com/serverpod/serverpod_docs/pull/743

FXschwartz (approved):
LGTM!

Issue #1147, opened 2023-07-12
Allow to configure the referenced field in a relation


Isakdl:
**Is your feature request related to a problem? Please describe.**
Allow a relation to specify the referencing field of another table/object. If this value is not set we simply use the default id field. 

The reference field can be configured by specifying reference=<field> which always defaults to reference=id. This will set the column we use as a foreign key in the relationship between objects.

If the referenced field is not of the same datatype as the local field with the relation an error should be reported.

Example valid usage:
```yaml
# user.yaml
class: User
table: user
fields:
  addressReference: int, relation(parent=address, reference=zip)

# address.yaml
class: Address
table: address
fields:
  zip: int
```

Example invalid usage:
```yaml
# user.yaml
class: User
table: user
fields:
  addressReference: String, relation(parent=address, reference=zip)

# address.yaml
class: Address
table: address
fields:
  zip: int
```

christerswahn:
For models with 1:1 relation, it would also be nice to be able to share the same ID. This will avoid having to do joins merely to find the id of the referenced record.

PR #5603, opened 2026-08-17
refactor(cli): Remove the hardcoded id reference in relations

Isakdl:
# Changes

This is a pure refactor that makes the `reference=` keyword from #1147 implementable. It ships no user-visible feature, and generated output is unchanged.

Relations used to assume the foreign key always points at the `id` column. The relation resolver wrote the `id` constant into every resolved relation at eight call sites, and `create_definition.dart` wrote a literal `['id']` into every foreign key without ever reading `foreignFieldName`. Adding a configurable reference field on top of that would have meant touching all of it in the same PR as the feature, so I split it out to keep the feature diff small and to isolate the risk of silently changing existing users' generated code.

The unresolved relation definitions now carry the reference field, defaulting to the primary key, and `_parseRelation` threads it into every relation it builds. From there it flows to the resolved relations and to the foreign key definitions. The parser has no way to read a non-default value yet, so every call site still resolves to `id`.

One thing I deliberately did not fix: `referenceColumns` wants column names, while the value now threaded into it is a field name. Today that is harmless and unreachable rather than merely unlikely, because `_resolveIdField` rebuilds the id field without a column name override, so the only field a relation can reference always has `columnName == name`. It only becomes a real bug once `reference=` can name an arbitrary field, which can carry `column=`.

Fixing it properly is also not a one-liner: `_createForeignKeys` only sees the child class, so it would need the full model list plumbed in, a table-to-class lookup, and a fallback for when that lookup misses. It does miss in practice, since foreign keys can target module tables that are not part of the generating package's own models. That is worth doing deliberately in the follow-up, so it is marked with a `TODO` against this issue rather than rushed in here.

I also added a primary key flag to `SerializableModelFieldDefinition`. There was no representation of a primary key in the analyzer's model layer beyond the literal field name `id`, and the follow-up needs to ask that question during validation. It is set where the id field is already synthesised, which runs before validation.

Towards #1147.

### DIFF
--- a/tools/serverpod_cli/lib/src/database/create_definition.dart
+++ b/tools/serverpod_cli/lib/src/database/create_definition.dart
@@ -122,7 +122,9 @@ List<ForeignKeyDefinition> _createForeignKeys(
         columns: [field.columnName],
         referenceTable: relation.parentTable,
         referenceTableSchema: 'public',
-        referenceColumns: ['id'],
+        // TODO(#1147): Map the reference field to its column name. The two only
+        // coincide today because the primary key has no column name override.
+        referenceColumns: [relation.foreignFieldName],
         onDelete: relation.onDelete,
         onUpdate: relation.onUpdate,
       ),
--- a/modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_client/lib/src/protocol/client.dart
+++ b/modules/serverpod_auth/serverpod_auth_bridge/serverpod_auth_bridge_client/lib/src/protocol/client.dart
@@ -76,7 +76,7 @@ class EndpointLegacyEmail extends _i1.EndpointRef {
     },
   );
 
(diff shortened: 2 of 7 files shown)


tools/serverpod_cli/lib/src/database/create_definition.dart
```diff
@@ -122,7 +122,9 @@ List<ForeignKeyDefinition> _createForeignKeys(
         columns: [field.columnName],
         referenceTable: relation.parentTable,
         referenceTableSchema: 'public',
-        referenceColumns: ['id'],
+        // TODO(#1147): Map the reference field to its column name. The two only
+        // coincide today because the primary key has no column name override.
+        referenceColumns: [relation.foreignFieldName],
```
Isakdl, line 127: I'll do a follow up resolving this

PR #5605, opened 2026-08-17
feat: Allow using deferrable foreign key constraints on the models

### LINKED ISSUE #5338: Allow using deferrable foreign key constraints on the models
### Problem to Solve

For the majority of the use cases, constraints can be evaluated when issuing each operation. However, there are some cases - like offline data merge - in which data can be out-of-order and foreign key constraints evaluation must be deferred to the transaction commit time instead. Postgres and SQLite support this though deferrable constraints, but we expose no way to opt-in for such behavior on the models.

### Proposal

Expose a `deferrableConstraints` parameter on the `generator.yaml` that will modify how the database generates foreign key constraints from then on. The accepted enum values should be:

- `off` -> Generates constraints normally, same way as today (`NOT DEFERRABLE`).
- `initiallyImmediate` -> Generates constraints as `DEFERRABLE INITIALLY IMMEDIATE`.
- `initiallyDeferred` -> Generates constraints as `DEFERRABLE INITIALLY DEFERRED`.

```yaml
# config/generator.yaml
type: server

client_package_path: ../serverpod_test_client
server_test_tools_path: test_integration/test_tools

deferrableConstraints: initiallyImmediate
```

Then, on the `TransactionSettings`, expose a `deferConstraints` boolean that will execute `SET CONSTRAINTS ALL DEFERRED` on Postgres and `PRAGMA defer_foreign_keys = ON` on SQLite after starting the transaction to defer the transaction if they are created as `initiallyImmediate`. To be used like:

```dart
await session.db.transaction(
  settings: const TransactionSettings(deferConstraints: true),
  (transaction) async {
    // Operations that transiently violate FK constraints...
  },
);
```

### Use Case

Allow merging out-of-order data inside a transaction without violating foreign key constraints. Working around is not always possible because the data might reflect a state in which original FKs were overridden before being sent and it is no longer possible to restore the causal order of the operations to merge without transiently violating FK constraints.

### Alternatives

Drop and recreate all constraints as deferrable after each migration.

### Additional context

For proper migration support, `ForeignKeyDefinition` objects will have to store a `deferrable` parameter to track how they were created.

- [x] I want to work on this

marcelomendoncasoares:
Closes: #5338

## Pre-launch Checklist

## Breaking changes

None.


tools/serverpod_cli/test/database/migration/deferrable_constraints_migration_test.dart
```diff
@@ -0,0 +1,142 @@
...
+import 'package:serverpod_database/serverpod_database.dart';
+import 'package:test/test.dart';
+
+import '../../test_util/builders/database/column_definition_builder.dart';
+import '../../test_util/builders/database/database_definition_builder.dart';
+import '../../test_util/builders/database/index_definition_builder.dart';
+import '../../test_util/builders/database/table_definition_builder.dart';
+
+void main() {
+  group(
```
FXschwartz: These three groups each hold a single test with no shared `setUp`. Could they be changed to be flat test()s with the full given/when/then description instead, like `relation_deferrable_test.dart` does in this PR?

tools/serverpod_cli/test/database/migration/deferrable_constraints_migration_test.dart
```diff
@@ -0,0 +1,142 @@
...
+          databaseTarget: target,
+        );
+
+        expect(migration.actions, isEmpty);
+      });
+    },
+  );
+}
+
+ForeignKeyDefinition _employeeForeignKey({DeferrableConstraint? deferrable}) =>
```
FXschwartz, line 100: These four helpers are identical to the ones in `deferrable_constraints_sql_generation_test.dart`. Probably worth moving them into `test_util/builders/database/` so both files share them?

tools/serverpod_cli/test/database/extentions/deferrable_constraints_sql_generation_test.dart
```diff
@@ -0,0 +1,221 @@
...
+      'then it includes DEFERRABLE INITIALLY DEFERRED.',
+      () {
+        var sql = SqliteSqlGenerator().generateSql(database);
+
+        expect(sql, contains('DEFERRABLE INITIALLY DEFERRED'));
+      },
+    );
+  });
+
+  test(
```
FXschwartz: nit - The deferrable cases above split Postgres and SQLite into separate tests, but this one asserts both generators in a single test. Split it to match the rest of the file?

packages/serverpod_database/lib/src/adapters/sqlite/sqlite_analyzer.dart
```diff
@@ -187,10 +188,44 @@ class SqliteDatabaseAnalyzer extends DatabaseAnalyzer {
...
+    var start = createTableSql.indexOf('CONSTRAINT "$constraintName"');
+    if (start < 0) return null;
+
+    var end = createTableSql.indexOf(',\n', start);
+    var clause = (end < 0
+            ? createTableSql.substring(start)
+            : createTableSql.substring(start, end))
+        .toUpperCase();
+
+    if (!clause.contains('DEFERRABLE')) return null;
```
FXschwartz: nit - I know we never generate the clause but this would also match `NOT DEFERRABLE`, so a constraint with an explicit NOT DEFERRABLE clause would get gets read back as `initiallyImmediate`.  But since this method exists to parse arbitrary `sqlite_master` text, maybe add a `NOT DEFERRABLE` check that returns null first?

FXschwartz (changes requested):
Looks great! Just two questions and some minor comments
1. For the offline-merge use case, someone now has to annotate every relation individually? Losing the "flip everything at once" option?
2. Why two mutually exclusive booleans instead of a single enum valued key, `deferrable: initiallyDeferred`? That would match how `onDelete`/`onUpdate` take enum values and avoid the need for the mutual-exclusion validation.

tools/serverpod_cli/test/database/migration/deferrable_constraints_migration_test.dart
marcelomendoncasoares: Shame on me 😆

tools/serverpod_cli/test/database/migration/deferrable_constraints_migration_test.dart
marcelomendoncasoares: Since they would only repeat once and they are very specific - a shape that works for testing deferrable constraints, but might not serve for a generic relation test -, I preferred to duplicate.  We can merge them if we see the need to repeat this exact same shape again.

tools/serverpod_cli/test/database/extentions/deferrable_constraints_sql_generation_test.dart
marcelomendoncasoares: I thought in doing it, but decided it was not worth it. Since you also thought the same, I'll do it.

packages/serverpod_database/lib/src/adapters/sqlite/sqlite_analyzer.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5605/commits/5231e372d350dc4f3b76bd0570c88ee8a2c11dda.

tools/serverpod_cli/test/database/extentions/deferrable_constraints_sql_generation_test.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5605/commits/e9e4834f83b2b68b6c843b147a6da3122bb35010.

tools/serverpod_cli/test/database/migration/deferrable_constraints_migration_test.dart
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5605/commits/e9e4834f83b2b68b6c843b147a6da3122bb35010.

FXschwartz (approved):
LGTM!

PR #5606, opened 2026-08-17
chore: Remove unused Insights endpoint methods

FXschwartz:
- Moved to an opt-in abstract `InsightsDatabaseEndpoint`: `executeSql`, `runQueries`, `fetchDatabaseBulkData`, and `getDatabaseRowCount`. Not served by default. A project enables them by extending the endpoint, like with the auth module endpoints.
- Removed: `hotReload`, `getOpenSessionLog` and `shutdown`.
- `fetchFile` stays: it is whitelist guarded and powers the live protocol listing in the Insights app.

Also fixes the generator to support extending endpoints from the `serverpod` package itself: the client generator resolves a parent endpoint's client class through the module mappings, and the internal `serverpod` module's names were derived from the `<name>_server/<name>_client` convention the core package does not follow. `ModuleConfig` now maps it to `serverpod/serverpod_service_client`, with a regression test.

## Pre-launch Checklist

## Breaking changes
- `executeSql`, shutdown, `hotReload`, `getOpenSessionLog`, `runQueries`, `fetchDatabaseBulkData`, and `getDatabaseRowCount` are no longer served by the built-in insights protocol.
- ServerpodSqlException is removed.


packages/serverpod/lib/src/endpoints/insights.dart
marcelomendoncasoares, line 396: Question: Where is this method used? On the monorepo it is only called on a test file (that is inclusively very old and out-of-pattern).

marcelomendoncasoares (commented):
I commented on one, but can we also remove the others that call arbitrary operations on the server? If they are only being used on tests, we can replace them by a normal endpoint on the test server instead of using Insights. Ideally, `fetchDatabaseBulkData`, `runQueries`, `getDatabaseRowCount` and `fetchFile` should also be removed, if they are not used by the Insights app or Cloud.

packages/serverpod/lib/src/endpoints/insights.dart
FXschwartz: I didn't remove them because all four are used by the insights app.
`fetchFile` - [lib/business/config_handler.dart:54](https://github.com/serverpod/serverpod_gui/blob/fd7b54359bf1973f40d1749c994164118224b28b/serverpod_gui/lib/business/config_handler.dart#L54-L56)
`fetchDatabaseBulkData` - [lib/widgets/database/providers.dart:62](https://github.com/serverpod/serverpod_gui/blob/fd7b54359bf1973f40d1749c994164118224b28b/serverpod_gui/lib/widgets/database/providers.dart#L62)
`getDatabaseRowCount` - [lib/widgets/database/providers.dart:101](https://github.com/serverpod/serverpod_gui/blob/fd7b54359bf1973f40d1749c994164118224b28b/serverpod_gui/lib/widgets/database/providers.dart#L101)
`runQueries` - [lib/widgets/database/providers.dart:141](https://github.com/serverpod/serverpod_gui/blob/fd7b54359bf1973f40d1749c994164118224b28b/serverpod_gui/lib/widgets/database/providers.dart#L141)

There's also a commented out usage of `getDatabaseDefinitions` in a Riverpod provider [here](https://github.com/serverpod/serverpod_gui/blob/fd7b54359bf1973f40d1749c994164118224b28b/serverpod_gui/lib/widgets/database/providers.dart#L14)

marcelomendoncasoares: Everything under `lib/widgets/database` is part of a feature that never went live - and, if we ever decide to publish it, must be as an opt-in module, not as built in. We can remove them from `InsightsEndpoint` and later from Insights itself when upgrading the app.

marcelomendoncasoares: To avoid breaking this code completely, one alternative is moving all these methods under a new abstract `InsightsDatabaseEndpoint`. This will make them still available as opt-in if we want to continue the feature (explicitly extending the endpoint, as we do on auth), and the upgrade path on Insights will be as easy as switching the endpoint.

FXschwartz: Done! Went with the abstract endpoint approach. `runQueries`, `fetchDatabaseBulkData`, and `getDatabaseRowCount` now live on an abstract `InsightsDatabaseEndpoint`, removed from `InsightsEndpoint`. Nothing is served unless a project extends it, same as the auth endpoints.

I kept `fetchFile` since the protocol/endpoint listing in the app is a live feature but let me know if you want that one gone too.

FXschwartz:
@marcelomendoncasoares Ok finally ready for another review and the module issue is fixed. `ModuleConfig` now maps it to `serverpod/serverpod_service_client`, with a regression test in the abstract-endpoint suite. It does now mean a client package whose server extends an endpoint from the `serverpod` package needs a `serverpod_service_client` dependency.

`executeSql`, `runQueries`, `fetchDatabaseBulkData`, and `getDatabaseRowCount` now live on the abstract `InsightsDatabaseEndpoint`, and the migration test servers opt in by extending it.

`shutdown` is currently removed rather than moved and `executeSql` now throws `BulkDataException` instead of `ServerpodSqlException`

tests/serverpod_test_nonvector/serverpod_test_nonvector_server/lib/src/endpoints/insights_database.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:serverpod/serverpod.dart';
+
+/// Exposes the opt-in [InsightsDatabaseEndpoint] so the migration tests can
+/// reset the live database.
+class InsightsDatabaseTestEndpoint extends InsightsDatabaseEndpoint {
+  // The migration tests call this endpoint without authentication.
+  @override
+  bool get requireLogin => false;
```
marcelomendoncasoares: Question: Is this override necessary? Doesn't the endpoint already declare it on the base class?

marcelomendoncasoares (approved):
LGTM! 🚀 

We need to follow this PR with an update to the docs on the migration guide and Insights page.

FXschwartz:
@marcelomendoncasoares Docs PR created! - https://github.com/serverpod/serverpod_docs/pull/764

PR #5610, opened 2026-08-18
feat: Allow `fk` flag for declaring the owner of the foreign key on a `relation`

### LINKED ISSUE #5595: Generate the `field` for relations that require the explicit FK

marcelomendoncasoares:
Closes: #5595

## Pre-launch Checklist

## Breaking changes


tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -824,6 +824,45 @@ class Restrictions {
     return [];
   }
 
+  List<SourceSpanSeverityException> validateRelationFkKey(
```
FXschwartz, line 794: This is basically `validateRelationFieldKey` with different strings. Might be nice to pull the shared part into a private helper that takes the property name, so the two can't drift. Not a blocker though.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_fk_keyword_test.dart
```diff
@@ -0,0 +1,550 @@
...
+
+        collector = CodeGenerationCollector();
+        StatefulAnalyzer(
+          config,
+          models,
+          onErrorsCollector(collector),
+        ).validateAll();
+      });
+
+      test('then no errors are collected.', () {
```
FXschwartz, line 305: Could this one also check that the generated `companyId` is `UuidValue`? I ran it locally and it is, but right now the test would still pass if the type fell back to int.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_fk_keyword_test.dart
```diff
@@ -0,0 +1,550 @@
...
+            collector.errors.single.message,
+            'The "fk" property is mutually exclusive with the "field" property.',
+          );
+        },
+      );
+    },
+  );
+
+  group(
+    'Given a list relation with the fk property, '
```
FXschwartz, line 370: Two more error cases that might be worth a test since they're one-liners: `relation(parent=..., fk)` on an id field, and fk on one side with field= on the other. Both already produce the right errors when I tried them, just nothing pins that down.

tests/serverpod_test_server/lib/src/models/models_with_relations/fk_relation/fk_relation_employee.spy.yaml
```diff
@@ -0,0 +1,8 @@
+# Many side of a named one-to-many relation. `companyId` is generated from `fk`.
+# `previousCompanyId` is generated as nullable because of `optional`.
+class: FkRelationEmployee
+table: fk_relation_employee
+fields:
+  name: String
+  company: FkRelationCompany?, relation(name=fk_relation_company_employees, fk)
+  previousCompany: FkRelationCompany?, relation(optional, fk)
```
FXschwartz, line 10: Small question: on an unnamed relation fk is a no-op since that side already owns the key. Is the intent here to show it's allowed, or should we warn that it's redundant?

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
```diff
@@ -333,6 +334,7 @@ class ModelDependencyResolver {
       fieldDefinition,
       foreignRelationField,
     );
+    _resolveFieldIndexes(foreignRelationField, classDefinition);
```
FXschwartz, line 343: This is probably just me but it took me a longer than I'd like to admit and with help from fable to understand why this is needed. Might warrant a code comment.

FXschwartz (changes requested):
Looking great! Just some minor test comments and one code comment addition to help slow people like me in the future 😅

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
marcelomendoncasoares: No problem in documenting. It is indeed unclear from looking at the code. Done at https://github.com/serverpod/serverpod/pull/5610/commits/9bafe1ead1ef5dbe60fdb036e753ea8b07fc7a98.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
marcelomendoncasoares: Good idea! Refactored at https://github.com/serverpod/serverpod/pull/5610/commits/f10badb5cf1d7cf68d89864b9c37a504a221fe59.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_fk_keyword_test.dart
marcelomendoncasoares: Added at https://github.com/serverpod/serverpod/pull/5610/commits/7f7090b0f66403ae186e2818a94e1cb8c5ad9998.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/relation/relation_fk_keyword_test.dart
marcelomendoncasoares: Added at https://github.com/serverpod/serverpod/pull/5610/commits/28b4939c1be2eeeb29c06d1d6f01af5aca0c920b.

tests/serverpod_test_server/lib/src/models/models_with_relations/fk_relation/fk_relation_employee.spy.yaml
marcelomendoncasoares: It is allowed, since it causes no harm. Made explicit at https://github.com/serverpod/serverpod/pull/5610/commits/62835da53fde040c0a08f522c64a0afe2a930c7c.

FXschwartz (approved):
LGTM! Getting closer and closer to having offline sync!

PR #5611, opened 2026-08-18
feat: Simplified Serverpod initialization

### LINKED ISSUE #4544: Simplified `Serverpod` initialization
With a bit of code generation, we can simplify the initialization of the `Serverpod` object. Initialization currently looks like this:

```dart
  final pod = Serverpod(args, Protocol(), Endpoints());
```

Instead, we could hide the `Protocol` and `Endpoints` classes (make them optional for backwards compatibility):

```dart
final pod = Serverpod(args);
```

FXschwartz:
Adds a generated `lib/src/generated/serverpod.dart` to every server package: a `Serverpod` subclass pre-configured with the project's generated `Protocol` and `Endpoints`, so a server can now be created with just:
```dart
import 'src/generated/serverpod.dart';
final pod = Serverpod(args);
```

- The generated class extends the framework's Serverpod, re-declares all of its named constructor parameters, and forwards them while injecting `Protocol()` and `Endpoints()` in the super call.
- The file re-exports `package:serverpod/serverpod.dart`, so `server.dart` needs only the single generated import.
- Generation is gated to PackageType.server

Closes #4544

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tests/serverpod_test_server/lib/src/generated/serverpod.dart
+++ b/tests/serverpod_test_server/lib/src/generated/serverpod.dart
@@ -15,6 +15,7 @@ import 'dart:io' as _i2;
 import 'protocol.dart' as _i3;
 import 'endpoints.dart' as _i4;
 export 'package:serverpod/serverpod.dart' hide Serverpod;
+export 'future_calls.dart' show ServerpodFutureCallsGetter;
 
 /// The Serverpod server for this project.
 ///
--- a/tests/serverpod_test_server/lib/src/generated/serverpod.dart
+++ b/tests/serverpod_test_server/lib/src/generated/serverpod.dart
@@ -1,0 +1,60 @@
+/* AUTOMATICALLY GENERATED CODE DO NOT MODIFY */
+/*   To generate run: "serverpod generate"    */
+
+// ignore_for_file: implementation_imports
+// ignore_for_file: library_private_types_in_public_api
+// ignore_for_file: non_constant_identifier_names
+// ignore_for_file: public_member_api_docs
+// ignore_for_file: type_literal_in_constant_pattern
+// ignore_for_file: use_super_parameters
+// ignore_for_file: invalid_use_of_internal_member
+
+// ignore_for_file: no_leading_underscores_for_library_prefixes
+import 'package:serverpod/serverpod.dart' as _i1;
+import 'dart:io' as _i2;
+import 'protocol.dart' as _i3;
+import 'endpoints.dart' as _i4;
+export 'package:serverpod/serverpod.dart' hide Serverpod;
+
+/// The Serverpod server for this project.
+///
+/// Pre-configured with the generated Protocol serialization manager and
+/// Endpoints, so a server can be created with just the command line
+/// arguments:
+///
+/// ```dart
+/// final pod = Serverpod(args);
+/// ```
+class Serverpod extends _i1.Serverpod {
+  Serverpod(
+    List<String> args, {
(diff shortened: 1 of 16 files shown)


tests/serverpod_test_server/lib/src/generated/serverpod.dart
```diff
@@ -0,0 +1,60 @@
...
+// ignore_for_file: type_literal_in_constant_pattern
+// ignore_for_file: use_super_parameters
+// ignore_for_file: invalid_use_of_internal_member
+
+// ignore_for_file: no_leading_underscores_for_library_prefixes
+import 'package:serverpod/serverpod.dart' as _i1;
+import 'dart:io' as _i2;
+import 'protocol.dart' as _i3;
+import 'endpoints.dart' as _i4;
+export 'package:serverpod/serverpod.dart' hide Serverpod;
```
marcelomendoncasoares, line 17: Question: Should we also export the future calls generated getter?

tools/serverpod_cli/test/generator/dart/server_code_generator/serverpod_class_test.dart
```diff
@@ -0,0 +1,90 @@
...
+            expect(serverpodFile, matches(r'_i\d+\.Endpoints\(\)'));
+          },
+        );
+      });
+    },
+  );
+
+  test(
+    'Given a module package when generating protocol code then no serverpod '
+    'file is created.',
```
marcelomendoncasoares:
```suggestion
    'Given a module package '
    'when generating protocol code '
    'then no serverpod file is created.',
```

tools/serverpod_cli/test/generator/dart/server_code_generator/serverpod_class_constructor_sync_test.dart
```diff
@@ -0,0 +1,107 @@
...
+    'lib',
+    'src',
+    'server',
+    'serverpod.dart',
+  );
+
+  test(
+    'Given the framework Serverpod constructor when comparing with the '
+    'generated Serverpod subclass then all named parameters are declared and '
+    'forwarded.',
```
marcelomendoncasoares:
```suggestion
    'Given the framework Serverpod constructor '
    'when comparing with the generated Serverpod subclass '
    'then all named parameters are declared and forwarded.',
```

marcelomendoncasoares (commented):
Only a few comments around the `futureCalls` getter and two nits on the tests. Nice to see the change was indeed so small!

tests/serverpod_test_server/lib/src/generated/serverpod.dart
FXschwartz: Good catch! We do need it. Added the same conditional export that `endpoints.dart` already does in aa3b93373, with generator tests for both cases.

tools/serverpod_cli/test/generator/dart/server_code_generator/serverpod_class_test.dart
FXschwartz: Done! 8385bd676

tools/serverpod_cli/test/generator/dart/server_code_generator/serverpod_class_constructor_sync_test.dart
FXschwartz: Done! 8385bd676

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5616, opened 2026-08-19
chore: Review all skills for correctness and better guidance

### LINKED ISSUE #5533: Evaluate if agents are receiving good information of whether `serverpod start` is running
The AI agents are often confused about whether the server is running. Adding a tool to check the server status would probably mitigate this.

I suggest a command `is_serverpod_start_running`, which returns `Yes` or `No`.

Description:
```
Checks if the server is currently running with `serverpod start`.
```

marcelomendoncasoares:
This PR closes a few gaps in the skills and related documentation for newly created projects.

Closes: #5533

## Agent skills

- **Correctness.** Every factual claim was checked against the source, and the wrong ones fixed — a tool name that doesn't exist, wrong documented defaults, an API signature, a method Relic never provided, and a directive that forbade a supported workflow.
- **Coverage.** Filled gaps where a real workflow had no documentation at all: common modeling features, the current major-version upgrade, setup that made copy-pasted examples throw at runtime, the default local database story, and editing generated migration SQL.
- **Token budget.** Lookup material moved out of the always-read `SKILL.md` files into `references/` that load on demand (9 files, ~13 KB), and content duplicated between skills was consolidated to one home plus links.
- **Activation.** Descriptions rewritten as the routing layer: sharper boundaries between neighboring skills, negative triggers where a skill was being pulled in wrongly, and one over-eager skill restricted to the case it is actually good for.
- **New `serverpod-cloud` skill.** Introduced to present managed hosting as an option and points at the docs, deliberately without a deploy runbook.
- **Maintenance.** `skills/README.md` now documents the conventions — naming, what belongs in a description, when to use `references/`, and updating skills alongside the API they describe.

## Project creation

- `AGENTS.md` is now rendered per project shape instead of assuming a Flutter app, a database and auth: Flutter-only sections, migration steps and auth advice are gated on `flutterApp`, `database`/`postgres`/`sqlite` and `auth`.
- New `AGENTS.md` for module projects, covering what differs from an app project: no runnable server, generate and migrate from the module server package, prefixed table names.
- `CLAUDE.md` is written for module projects too, so Claude Code loads their `AGENTS.md` (`create.dart`).
- New `skills` template flag (`TemplateContext.skills`, true when an IDE is configured) so `AGENTS.md` mentions the installed skills only when they are actually installed, with tests.
- `AGENTS.md` gained the rules it was missing — never edit generated code (while allowing hand-edits to `migration.sql`), what to do when the MCP is unreachable, that tests need no Docker, and how to enable text entry for driver-based app testing.

## Framework and CLI

- Corrected a stale doc comment in the test-tools generator and regenerated the committed test tools it produces (11 files).
- The MCP bridge's own instructions and not-running error now match the guidance shipped in the skills and `AGENTS.md`.

## Pre-launch Checklist

## Breaking changes

None.


packages/serverpod/skills/serverpod-auth/SKILL.md
```diff
@@ -1,64 +1,71 @@
...
-description: Serverpod Authentication — Signing in users, verify if they are authenticated, assinging scopes (e.g., admin). Use when adding features that require the user to be signed in.
+description: Serverpod authentication — signing in users, checking whether they are authenticated, assigning scopes (e.g. admin). Use when adding features that require the user to be signed in.
 ---
 
 # Serverpod Authentication
 
-Serverpod has authentication built in. Pre-configured with email. Most social sign-ins supported (Apple, Google, GitHub, Facebook, Microsoft, etc) but need configuration.
+Serverpod has authentication built in. Projects created with `serverpod create` have it enabled by default (unless `--no-auth`, or the project has no database), pre-configured with email. Most social sign-ins are supported (Apple, Google, GitHub, Facebook, Microsoft, etc) but need configuration.
+
+## Packages
```
vlidholt: Perhaps we should put this into a reference, like the user management. It's rare that the agent will need to configure the authentication (typically only happens once), but here we include that information every time.

It's probably a good idea to only include things that the agent often needs to know at the top level. (e.g. checking if a user is authenticated, getting the user id, etc).

packages/serverpod/skills/serverpod-cloud/SKILL.md
```diff
@@ -0,0 +1,12 @@
+---
+name: serverpod-cloud
```
vlidholt: It could be smart to make this into a `serverpod-deployment` skill, so it's more likely to be picked up by the agent as a default deployment strategy. Also relevant for the description.

templates/serverpod_templates/projectname/AGENTS.md
```diff
@@ -1,30 +1,96 @@
+<!-- {{#flutterApp}} -->
 # Flutter & Serverpod project
 
-This is a newly created Flutter (frontend) & Serverpod (backend) project. Always build a backend for the app using Serverpod. If it make sense, add support for multiple users (using Serverpod's built in authentication).
+This project is a Flutter app (frontend) backed by a Serverpod server (backend). Always build the app's backend with Serverpod.
+<!-- {{#auth}} -->
+When the project needs users, use Serverpod's built in authentication, which is already set up in `lib/server.dart`.
```
vlidholt:
```suggestion
Build for multiple users, use Serverpod's built-in authentication, which is already set up in `lib/server.dart`.
```

vlidholt: If the user has selected auth, they likely want multiple users.

templates/serverpod_templates/projectname/AGENTS.md
```diff
@@ -1,30 +1,96 @@
...
+<!-- {{/auth}} -->
 
-The user starts the server and Flutter app with `serverpod start`. NEVER start the server yourself, instead STOP and ask the user to start it. When the server is running, interact with it through the `serverpod` MCP. `serverpod start` automatically handles hot reload for both the server and the app (as soon as files change).
+The user starts the server and Flutter app with `serverpod start`. There is no need to check if the server is running: make the changes and call the `serverpod` MCP tools as needed. If the server is not running, an informative error message will be received from the MCP server. Then STOP and ask the user to start the server manually, or ask for permission to start it yourself with `serverpod start --no-tui` — that command runs until it is stopped, so start it in the background and keep working. NEVER start the server without asking first. The Flutter app is started along with it, or can be launched from the MCP tool `spawn_flutter_app`.
```
vlidholt: Suggestion: I think we should only describe the happy path here. Adding many options often confuses the agents. Let's stick with just having the user run the server.

vlidholt: In my experiments, having multiple options, often made the agent choose poorly.

templates/serverpod_templates/projectname/AGENTS.md
```diff
@@ -1,30 +1,96 @@
...
+- `hot_reload` / `hot_restart` to reload or restart the server and the Flutter app. ALWAYS call `hot_restart` after doing changes in the Flutter app that may not work with normal hot reload (which is automatically applied).
+- `spawn_flutter_app` to start a Flutter app declared under `serverpod: flutter_apps:` in the server `pubspec.yaml`.
 - `get_flutter_app_dtd` (Dart tooling daemon) for connecting to the app through the `dart` MCP.
+<!-- {{/flutterApp}} -->
+<!-- {{^flutterApp}} -->
+- `hot_reload` / `hot_restart` to reload or restart the server. Use `hot_restart` for changes that hot reload cannot apply, such as changes to `main()`.
+<!-- {{/flutterApp}} -->
+<!-- {{#skills}} -->
+
+Serverpod agent skills are installed in this project. They cover models, migrations, the database ORM, endpoints, streams, authentication, testing, configuration and the web server — read the skill that matches the task before writing Serverpod code, starting with `serverpod-overview`.
```
vlidholt: I don't think this is needed. The skills are typically announced to the agent automatically.

templates/serverpod_templates/projectname/AGENTS.md
```diff
@@ -1,30 +1,96 @@
...
+<!-- {{/skills}} -->
+
+NEVER edit generated code. The server's `lib/src/generated/` directory and the whole `projectname_client` package are rewritten by the code generator. Change the `.spy.yaml` models, the endpoints, or `lib/server.dart` instead.
+<!-- {{#database}} -->
 
-Checklist after doing changes:
+Migrations are the exception: the `migration.sql` of a generated migration MAY be edited by hand, for example to add a data transformation, or to reach a destructive change through non-destructive steps. Leave the other files in the migration directory alone, and keep the schema that the SQL ends up with identical to `definition.sql` — new databases are created from that file and never run `migration.sql`.
```
vlidholt: Do we really want the agent to modify these?

vlidholt: Open question. Not sure what's best. :)

vlidholt:
I added some comments. I think the biggest one is that it may be smart to break up some of the skills using references, and make sure that only information that the agent _often_ need is present at the top level. For instance, in the authentication, installing the auth package is a rare occurrence, but working with the auth (setting scopes, restricting access, setting up tables with user ids, etc) is common.

vlidholt:
Also, in my testing (at least with lower-tier models), it's a bad idea to give the model more than one option for solving a task. It often chooses the wrong way or gets confused. Let's optimize for the happy path in the skills.

packages/serverpod/skills/serverpod-auth/SKILL.md
marcelomendoncasoares: Good idea. I did this with other parts of the skills, like the env vars, flutter apps, etc., but did not thought about this one. I'll move it to a reference.

packages/serverpod/skills/serverpod-cloud/SKILL.md
marcelomendoncasoares: Indeed, it will facilitate the discovery.

templates/serverpod_templates/projectname/AGENTS.md
marcelomendoncasoares: My thought around this was only to provide a path in which the agent could obey a direct order to start the server without contradicting the `AGENTS.md` or the overview skill - specially given that users won't be able to easily edit the skill to change the behavior in case they want to work on detached workflows, such as remote agents or worktrees.

But I'll revert the change with alternative to start in the background and keep only the direct order for the agent to always stop and ask the user to start the server, then (both here, on the skill and MCP responses).

templates/serverpod_templates/projectname/AGENTS.md
marcelomendoncasoares: I consider this important. Our migration system is very robust, but some destructive actions can be made non-destructive by using a custom SQL that our generator cannot emit (i.e. Changing a type from int to string while persisting the data through conversion). If we don't special case the `migration.sql` file as editable, any request by users to make the migration non-destructive would hit a wall.

templates/serverpod_templates/projectname/AGENTS.md
marcelomendoncasoares: Let's save some tokens, then. I'll remove it.

packages/serverpod/skills/serverpod-logging/SKILL.md
```diff
@@ -26,20 +26,22 @@ Uncaught endpoint exceptions are logged as session failures in `serverpod_sessio
...
+| Setting | Default |
+| ------- | ------- |
+| persistentEnabled | true (with DB) |
+| consoleEnabled | true (no DB, or development) |
+| consoleLogFormat | text (development), json (other modes) |
+| cleanupInterval | 24h |
+| retentionPeriod | 90d |
+| retentionCount | 100000 |
+
+Each setting also has a `SERVERPOD_SESSION_*` environment variable, listed in [Serverpod Configuration](../serverpod-configuration/SKILL.md).
```
FXschwartz: This link points at the configuration SKILL.md, but the env var table moved into `references/environment-variables.md` in this same PR, so the agent has to hop twice to find it. Might as well link the reference file directly.

templates/serverpod_templates/modulename/AGENTS.md
```diff
@@ -0,0 +1,31 @@
+# Serverpod module
+
+This project is a Serverpod module: a reusable package with server code (`modulename_server`) and client code (`modulename_client`) that other Serverpod projects add as a dependency. A module is not a runnable server, so there is no `serverpod start` and no `serverpod` MCP here.
```
FXschwartz, line 3: Looks like `serverpod create` still writes the serverpod MCP config for module projects. `_configureProjectMcpServers` runs for every template when an IDE is selected. So a module project gets an MCP whose tools always error with "ask the user to run serverpod start", which isn't possible here, and that contradicts this line. 

Should we skip the MCP config for module templates?

FXschwartz (changes requested):

packages/serverpod/skills/serverpod-auth/SKILL.md
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5616/commits/90fb43dc1327a2cb6b17336dee824832a7ebd4e9.

packages/serverpod/skills/serverpod-cloud/SKILL.md
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5616/commits/1968519119f7443c655064579481dd806be4882a.

templates/serverpod_templates/projectname/AGENTS.md
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5616/commits/b7531aad4f017322d43637e2b9b28d30db56af6a.

templates/serverpod_templates/projectname/AGENTS.md
marcelomendoncasoares: Reverted the multiple options at https://github.com/serverpod/serverpod/pull/5616/commits/98cd4ceb1728f809ce2449551060e74a0879e64c.

templates/serverpod_templates/projectname/AGENTS.md
marcelomendoncasoares: Removed at https://github.com/serverpod/serverpod/pull/5616/commits/b41756e5a498ca95a2832333d8f793ea08a55b03.

templates/serverpod_templates/projectname/AGENTS.md
marcelomendoncasoares: Made even a bit narrower at https://github.com/serverpod/serverpod/pull/5616/commits/cbd5a8b86ae0cf96732954da369c17451165637c.

templates/serverpod_templates/modulename/AGENTS.md
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5616/commits/6772c726b3dbbddd26d7fefa7314deb143790976.

packages/serverpod/skills/serverpod-logging/SKILL.md
marcelomendoncasoares: Fixed at https://github.com/serverpod/serverpod/pull/5616/commits/a6ef46661defccc718f7d7ba4fff565467750ad1.

FXschwartz (approved):
LGTM!

PR #5597, opened 2026-08-14
fix: Join start TUI borders and dividers with T-junction characters

FXschwartz:
The `serverpod` start TUI drew its borders, pane dividers, and rules independently, leaving visible gaps where lines met. With `nocterm` 0.9.0's box-line blending and `serverpod_tui` 0.12.0's border-matched tab bar underline, the TUI now joins them with junction characters: the pane divider tees into the outer border (┬/┴), the tab bar underline runs border-to-border and crosses the divider (┝/┿/┥), and the status, alert, and raw-log rules tee into the edges (├/┤).

Bumps `nocterm` to ^0.9.0 and `serverpod_tui` to ^0.12.0.

Fixes #5424 
### Before
start-before

### After
start-after-rebased

## Pre-launch Checklist

- [x] I read the [Contribute](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
- [x] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
- [x] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [dart format](https://dart.dev/tools/dart-format).
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.
- [ ] I added new tests to check the change I am making.
- [x] All existing and new tests are passing.
- [x] Any breaking changes are documented below.

If you need help, consider asking for advice on the [discussion board](https://github.com/serverpod/serverpod/discussions).

## Breaking changes
None

PR #5618, opened 2026-08-19
fix: Update start TUI app status on spontaneous exit, launcher divider color, and clickable app URL

### LINKED ISSUE #5613: App shows indicated as running in `serverpod start`, when it is in fact stopped
In `serverpod start`, the top bar can show an app as running (green dot), even though it has stopped. However, the correct state is displayed in the App Launcher.

Also, the "__X__ Stop App" does nothing.

Image

To reproduce:
- Run `serverpod start` with a default project.
- Close the web app.
- Wait a minute or two, and the bug appears.

### LINKED ISSUE #5614: The App Launcher in `serverpod start` has incorrect color for the divider
See screenshot:

Image

### LINKED ISSUE #5615: Clicking the URL for a running web app should open the web browser
In the `serverpod start` TUI, it would be nice if clicking the URL of a running web app actually opened the web browser with the app.

FXschwartz:
Fixes three start-TUI issues:
1. 37f9a8d - The manager never learned about a Flutter process that exited on its own, so the tab strip kept its green running dot, the App Launcher only corrected itself on reopen, and `X Stop App` acted on stale state.
2. fa2ea41 - The App Launcher's title divider was the last rule left on the default divider color after #5597; it now matches the panel border.
3. 717dc10 - A running web app's URL in the status line is now underlined and opens the app in the default browser when clicked.

**Note:** Currently the apps's URL is just rendered as text. In order to render it as a native terminal hyperlink (OSC 8) would require an update to the nocterm repo to support this.

Fixes #5613
Fixes #5614
Fixes #5615

## Pre-launch Checklist

## Breaking changes
None

### DIFF
--- a/tools/serverpod_cli/test/commands/start/flutter_shims/emits_machine_events.dart
+++ b/tools/serverpod_cli/test/commands/start/flutter_shims/emits_machine_events.dart
@@ -1,5 +1,7 @@
 // Test shim: emits a fixed `--machine` event sequence, then waits for
 // SIGINT/SIGTERM. `--ws=<uri>` overrides the wsUri in app.debugPort.
+// `--exit-file=<path>` makes the shim exit on its own once that file
+// appears, simulating a spontaneous app exit (e.g. browser closed).
 import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
@@ -44,6 +46,19 @@ Future<void> main(List<String> args) async {
   });
 
   final stopper = Completer<void>();
+
+  final exitFile = args
+      .where((a) => a.startsWith('--exit-file='))
+      .map((a) => a.substring('--exit-file='.length))
+      .firstOrNull;
+  if (exitFile != null) {
+    Timer.periodic(const Duration(milliseconds: 25), (timer) {
+      if (File(exitFile).existsSync()) {
+        // The signal watchers below keep the isolate alive; exit outright.
+        exit(0);
+      }
+    });
+  }
 
   ProcessSignal.sigint.watch().listen((_) {
     if (!stopper.isCompleted) stopper.complete();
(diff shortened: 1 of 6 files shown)


tools/serverpod_cli/test/commands/start/flutter_shims/emits_machine_events.dart
```diff
@@ -45,6 +47,19 @@ Future<void> main(List<String> args) async {
...
+      .map((a) => a.substring('--exit-file='.length))
+      .firstOrNull;
+  if (exitFile != null) {
+    Timer.periodic(const Duration(milliseconds: 25), (timer) {
+      if (File(exitFile).existsSync()) {
+        // The signal watchers below keep the isolate alive; exit outright.
+        exit(0);
+      }
+    });
+  }
```
nielsenko: Why not use another signal instead of this hack?

nielsenko (commented):
Why the weird signal-by-file-creation?

tools/serverpod_cli/test/commands/start/flutter_shims/emits_machine_events.dart
FXschwartz: The test has no process handle. The scenario needs the shim to die on its own and not through anything `FlutterProcess` initiates. But the test only holds the FlutterAppManager; the underlying Process is private to `FlutterProcess`, so there's no `pid` to send a signal.

This also keeps the behavior the same across all OS's where Windows acts a bit different with Signals.

If you think it's worth some effort I could try exposing the `pid` as `@visibleForTesting` on `FlutterProcess`, pass it through the manager, then have the test call `killPid`

nielsenko: Okay, but the shim could just write the pid somewhere for the test to pickup. Or we could expose it on `FlutterProcess`.

But okay, I was jut curious. Your call.

nielsenko (approved):
LGTM 👍 .. once CI is green 😄

tools/serverpod_cli/test/commands/start/flutter_shims/emits_machine_events.dart
FXschwartz: No you are right that's easier and kills the hack. I've updated it so the shim writes its `pid` on startup and the test kills it with `killPid(sigkill)`. 20179b805

Thanks for pushing back, it's definitely cleaner 🙂.

PR #5621, opened 2026-08-20
refactor: Update CloudStorage interface

### LINKED ISSUE #4848: Discussion: extending CloudStorage vs provider-specific APIs for advanced storage capabilities
Related context:
- [#4846](https://github.com/serverpod/serverpod/pull/4846) adds the first official Tencent COS provider within the current `CloudStorage` contract.
- [#2229](https://github.com/serverpod/serverpod/issues/2229) highlights broader storage support needs for China-region and non-default provider setups.

While working on [#4846](https://github.com/serverpod/serverpod/pull/4846), I noticed that the current shared `CloudStorage` abstraction covers the basic cross-provider workflow well:

- server-side CRUD
- public URLs
- direct upload descriptions
- upload verification
- basic upload options such as `contentLength` and `preventOverwrite`

However, there are still several production-oriented capabilities that are not part of the shared abstraction today, especially for native / non-S3-compatible providers:

- temporary private download URLs / presigned GET
- custom `Content-Type`, metadata, and cache-related headers
- large-file multipart / streaming-oriented upload support
- object-management capabilities such as list / copy / move / rename

I am not proposing to expand [#4846](https://github.com/serverpod/serverpod/pull/4846) itself with all of this. That PR intentionally stays within the current `CloudStorage` contract so the Tencent COS provider can remain reviewable and aligned with the existing official providers.

What I would like to discuss is what the preferred next step should be.

## Option A: extend the shared Serverpod cloud storage abstractions / protocol

This would mean evolving the current shared `CloudStorage` API and, where needed, the direct-upload description protocol so these higher-level capabilities can become first-class optional features for official providers.

Potential direction:
- add a shared way to create temporary private download URLs
- introduce richer upload / object options for content type, metadata, and cache behavior
- define optional capabilities for larger-file multipart / streaming flows
- possibly introduce optional object-management APIs later

Pros:
- gives application developers a more consistent cross-provider API
- makes advanced providers feel more complete inside Serverpod itself
- could benefit not only COS, but also native GCP and future provider integrations

Cons:
- increases the surface area of the shared abstractions
- requires API design and backward-compatibility care
- may be too much to add all at once

## Option B: keep the shared abstraction small and support these features separately

This would mean leaving the current `CloudStorage` contract unchanged, and letting providers expose additional capabilities through provider-specific helpers / extension APIs.

This is already close to what I have explored in my community packages:
- `tencent_cos_dart` provides the low-level COS signer for presigned URLs, including custom headers and query params
- `serverpod_cos` provides `session.cosSigner()` and an endpoint / helper-based way to generate presigned URLs, including `GET`
- my CloudStorage-compatible community package was previously named `serverpod_cloud_storage_cos`, and has since been renamed to `tencent_cos_cloud_storage_serverpod` to avoid confusion with the new official package from [#4846](https://github.com/serverpod/serverpod/pull/4846)

Pros:
- keeps the shared abstraction focused and minimal
- lets provider-specific features evolve faster
- lower risk for the core framework

Cons:
- application code becomes less portable across providers
- advanced features are discovered through provider-specific APIs rather than a shared Serverpod interface
- may feel fragmented over time if multiple official providers need similar capabilities

## Questions

1. Which direction would you prefer for future work here:
   - extending the shared `CloudStorage` abstractions / protocol
   - or keeping the shared contract minimal and adding provider-specific extensions separately?

2. If the shared direction is preferred, which capability would be the best first candidate?
   - temporary private download URLs / presigned GET seems like the most generally useful one to me

3. If helpful, I can also share / adapt the approach used in my community packages and see whether that design is reusable here.

I would be happy to continue in whichever direction fits Serverpod best.

Crazelu:
The CloudStorage interface now supports metadata and temporary URLs. Upload description is now typed. Methods that previously had nullable return types now opt for throwing exceptions instead of returning null.

This refactor comes with hardened validation of provided options with exceptions thrown instead of silently ignoring.

Closes #4848 

## Pre-launch Checklist

## Breaking changes
- Several `CloudStorage` methods have been replaced or changed shape.
- `CloudStorage` methods now throw exceptions instead of doing nothing or returning null.
- Options like `preventOverwrite` are no longer silently ignored by implementations that don't support them; an exception is thrown instead.
- Upload description is now typed.


packages/serverpod/lib/src/cloud_storage/cloud_storage.dart
```diff
@@ -32,23 +33,38 @@ abstract class CloudStorage {
...
+  /// Returns a temporary link to a file in the storage with an [expiration] time.
+  ///
+  /// Throws a [CloudStorageFileNotFoundException] if no file exists at [path].
+  /// Throws a [CloudStorageUnsupportedOperationException] if this storage does not
+  /// provide temporary URLs.
+  Future<Uri> getTemporaryUrl({
+    required Session session,
+    required String path,
+    Duration expiration = const Duration(minutes: 10),
+  });
```
Isakdl: Question: Couldn't this just be an option on the public URL interface? 

This would only matter in the cases where the bucket is private and we need a signed url.

packages/serverpod/lib/src/cloud_storage/cloud_storage.dart
Isakdl, line 151: What's the difference between this and maxFileSize?

packages/serverpod/lib/src/cloud_storage/upload_description.dart
```diff
@@ -0,0 +1,96 @@
+import 'dart:convert';
+
+/// Type of a cloud storage upload.
+enum UploadType {
+  /// Multipart HTTP upload.
+  multipart,
+
+  /// Binary upload.
+  binary,
+}
```
Isakdl, line 10: Question: Does the server need to specify this? Couldn't the client decide how to upload? You could also do streamed uploads too right?

packages/serverpod/lib/src/cloud_storage/cloud_storage.dart
Crazelu: In the current implementations, `contentLength` is passed in the header for some providers and is optional.
`maxFileSize` is required for S3's presigned POST upload policy.

I now realize that this bleeds some provider-specific quirks into the interface.
I'll think a little more on this.

nielsenko: `contentLength` is the _declared exact size_ of the specific file the client is about to upload. It's a hint the caller supplies when they already know the size.

`maxFileSize` is ceiling the server allows for an upload.

nielsenko: So `contentLength <= maxFileSize` is the invariant. The ceiling is what the server _allows_, the content length is what the client _promises_

packages/serverpod/lib/src/cloud_storage/upload_description.dart
Crazelu: The client can choose how it provides the file data (ByteData or Stream) but not the upload format.
`UploadType` here maps to generic provider formats. For instance, Cloudflare R2 requires a binary PUT and an S3 presigned POST requires multipart.

Streamed uploads are already supported in FileUploader for client use.

packages/serverpod/lib/src/cloud_storage/cloud_storage.dart
```diff
@@ -94,52 +114,13 @@ final class CloudStorageOptions {
...
 
+  /// Metadata attached to the upload.
+  final Map<String, String> metadata;
+
   /// Creates a new [CloudStorageOptions].
   const CloudStorageOptions({
     this.contentLength,
     this.preventOverwrite = false,
+    this.metadata = const {},
   });
```
nielsenko, line 101: I suggest splitting this class by usage, so it doesn't carry redundant information on some calls. It still allow for non-breaking additions per call type later (as more implementations require it)

```suggestion
//// Options for [CloudStorage.storeFile].
final class StoreFileOptions {
  /// When set, the file is removed from the storage after this time.
  final DateTime? expiration;

  /// When true, the store fails if an object already exists at the given path.
  ///
  /// Implementations that cannot enforce this must throw a
  /// [CloudStorageException] rather than silently ignoring it.
  final bool preventOverwrite;

  /// Creates a new [StoreFileOptions].
  const StoreFileOptions({
    this.expiration,
    this.preventOverwrite = false,
  });
}

/// Options for [CloudStorage.createDirectFileUploadDescription].
final class UploadOptions {
  /// How long the generated upload URL stays valid.
  final Duration expirationDuration;

  /// The largest upload the generated URL will accept.
  final int maxFileSize;

  /// The exact size of the file about to be uploaded, when known. Lets the
  /// provider pin the accepted size instead of allowing anything up to
  /// [maxFileSize]. Must not exceed [maxFileSize].
  final int? contentLength;

  /// When true, the upload fails if an object already exists at the given path.
  ///
  /// Implementations that cannot enforce this must throw a
  /// [CloudStorageException] rather than silently ignoring it.
  final bool preventOverwrite;

  /// Creates a new [DirectUploadOptions].
  const UploadOptions({
    this.expirationDuration = const Duration(minutes: 10),
    this.maxFileSize = 10 * 1024 * 1024,
    this.contentLength,
    this.preventOverwrite = false,
  }) : assert(
         contentLength == null || contentLength <= maxFileSize,
         'contentLength must not exceed maxFileSize',
       );
}

```

nielsenko: Note that `maxFileSize` (and validation) moves into `DirectUploadOptions`.

packages/serverpod/lib/src/cloud_storage/cloud_storage.dart
Crazelu: It could be but I'd argue it is not quite clean to mix both and then require implementers to branch on `public`.

nielsenko (changes requested):
I think we should take the opportunity to change more. Here are my suggestions (with the help of claude for the table). So `File` on all file related methods, direction on `Uri` related ones (`Download`/`Upload`), drop `Direct`, etc.

## CloudStorage

| Before                                                                          | After                                         | Change                                      |
|---------------------------------------------------------------------------------|-----------------------------------------------|---------------------------------------------|
| `storeFile(byteData, expiration, verified) -> void`                             | `storeFile(byteData, options) -> void`        | `expiration` -> options; `verified` removed |
| `storeFileWithOptions(byteData, expiration, verified, options)`                 | (deleted)                                     | merged into `storeFile`                     |
| `retrieveFile() -> ByteData?`                                                   | `retrieveFile() -> ByteData?`                 | unchanged                                   |
| (none)                                                                          | `statFile() -> FileStat?`                     | new                                         |
| `fileExists() -> bool`                                                          | `fileExists() -> bool`                        | now a default method over `statFile`        |
| `deleteFile() -> void`                                                          | `deleteFile() -> void`                        | unchanged                                   |
| `getPublicUrl() -> Uri?`                                                        | `publicDownloadUrl() -> Uri?`                 | renamed                                     |
| (none)                                                                          | `temporaryDownloadUrl(options) -> Uri?`       | new                                         |
| `createDirectFileUploadDescription(expirationDuration, maxFileSize) -> String?` | `createUploadDescription(options) -> String?` | renamed; both params -> options             |
| `createDirectFileUploadDescriptionWithOptions(..., options)`                    | (deleted)                                     | merged into `createUploadDescription`       |
| `verifyDirectFileUpload() -> bool`                                              | `verifyUpload() -> bool`                      | renamed                                     |

## StorageAccess

| Before                                                                                                | After                                         | Change                                 |
|-------------------------------------------------------------------------------------------------------|-----------------------------------------------|----------------------------------------|
| `storeFile(byteData, expiration, preventOverwrite)`                                                   | `storeFile(byteData, options)`                | flattened params -> `StoreFileOptions` |
| `createDirectFileUploadDescription(expirationDuration, maxFileSize, contentLength, preventOverwrite)` | `createUploadDescription(options)`            | 4 params -> `UploadOptions`            |
| `getPublicUrls(paths) -> List<Uri?>`                                                                  | `publicDownloadUrls(paths)`                   | renamed                                |
| (none)                                                                                                | `temporaryDownloadUrl(options)`, `statFile()` | new pass-throughs                      |
| null-check repeated 8x                                                                                | `_storage(storageId)` helper                  | dedup                                  |
| (none)                                                                                                | `contentLength <= maxFileSize` check          | hoisted from 3 impls                   |

## Types

| Type                          | Fields                                                                               | Used by                                  |
|-------------------------------|--------------------------------------------------------------------------------------|------------------------------------------|
| `StoreFileOptions`            | `expiration`, `preventOverwrite`, `metadata`                                         | `storeFile`                              |
| `UploadOptions`               | `expirationDuration`, `maxFileSize`, `contentLength`, `preventOverwrite`, `metadata` | `createUploadDescription`                |
| `TemporaryDownloadUrlOptions` | `expirationDuration`, `downloadFileName`, `contentType`                              | `temporaryDownloadUrl`                   |
| `FileMetadata`                | `contentType`, `cacheControl`, `contentDisposition`, `contentEncoding`, `custom`     | write-time; in both upload options types |
| `FileStat`                    | `size`, `lastModified`, `contentType`, `etag`, `custom`                              | read-time; `statFile()`                  |
| `CloudStorageOptions`         | (removed)                                                                            | split into the 3 above                   |
| `CloudStorageWithOptions`     | (removed)                                                                            | deleted                                  |

## Behaviour

| Area                                   | Before                                     | After                                            |
|----------------------------------------|--------------------------------------------|--------------------------------------------------|
| Unsupported option                     | silently dropped                           | must throw `CloudStorageException`               |
| `preventOverwrite` on DB direct upload | accepted, ignored                          | throws (or `ON CONFLICT DO NOTHING` + migration) |
| `verified: false`                      | on public interface, ignored by S3 and GCP | `DatabaseCloudStorage.storeUnverifiedFile`       |
| Content type, DB reads                 | guessed from 12-entry map                  | stored value; extension guess as fallback        |
| `fileExists` on S3                     | HEAD, discards headers                     | HEAD via `statFile`, headers kept                |
| `expiration` on DB entries             | written, never enforced                    | enforced (prerequisite for download tokens)      |

packages/serverpod/lib/src/cloud_storage/public_endpoint.dart
```diff
@@ -61,42 +77,111 @@ class CloudStoragePublicEndpoint extends Endpoint {
...
+    required String storageId,
+    required String path,
+    String? contentTypeOverride,
+    String? downloadFileName,
+  }) async {
+    try {
+      final results = await Future.wait([
+        session.storage.retrieveFile(storageId: storageId, path: path),
+        session.storage.statFile(storageId: storageId, path: path),
+      ]);
```
nielsenko: This is a performance observation, not a correctness one. For `DatabaseCloudStorage` both `retrieveFile` and `statFile` go through `_findAvailableEntry`, which selects `byteData`. 

Hence each download reads the whole blob out of Postgres twice and holds two copies in memory at once. Should we special case this?

packages/serverpod/lib/src/cloud_storage/public_endpoint.dart
```diff
@@ -61,42 +77,115 @@ class CloudStoragePublicEndpoint extends Endpoint {
...
+        preventOverwrite: uploadInfo.preventOverwrite,
+        metadata: FileMetadata(
+          contentType: uploadInfo.contentType,
+          cacheControl: uploadInfo.cacheControl,
+          contentDisposition: uploadInfo.contentDisposition,
+          contentEncoding: uploadInfo.contentEncoding,
+          custom: _decodeCustomMetadata(uploadInfo.customMetadata),
+        ),
+      ),
     );
```
nielsenko: This may throw. Should it return `false` in that case, or should upload be `Future<void>`?
```suggestion
try {
  await storage.storeUnverifiedFile(/* ... */);
} on CloudStorageFileAlreadyExistsException {
  return false;
}
```

packages/serverpod/lib/src/cloud_storage/public_endpoint.dart
```diff
@@ -61,42 +77,115 @@ class CloudStoragePublicEndpoint extends Endpoint {
...
 
+    await CloudStorageDirectUploadEntry.db.deleteRow(session, uploadInfo);
```
nielsenko, line 126: What is the intend behind moving delete here? I assume it is to allow _retry-after-failure_, but it does allow the same url to be used repeatedly for upload (until expiration) - I think. The old code was strictly single use.  Maybe it is intentional, but just checking.

packages/serverpod/lib/src/cloud_storage/database_cloud_storage.dart
```diff
@@ -1,236 +1,388 @@
...
+  Future<Uri> temporaryDownloadUrl({
     required Session session,
     required String path,
-    Duration expirationDuration = const Duration(minutes: 10),
-    int maxFileSize = 10 * 1024 * 1024,
+    TemporaryDownloadUrlOptions options = const TemporaryDownloadUrlOptions(),
   }) async {
-    var config = session.server.serverpod.config;
+    options.validate();
+    await statFile(session: session, path: path);
```
nielsenko: Do we need `statFile` here? It still retrieves all bytes. You just added `fileExists` - could we use that?

packages/serverpod/lib/src/cloud_storage/database_cloud_storage.dart
```diff
@@ -1,236 +1,388 @@
...
+
+  void _scheduleTemporaryDownloadDeletion({
+    required Session session,
+    required Duration delay,
+    required Future<void> Function() delete,
+  }) {
+    Timer(delay, () {
+      _temporaryDownloadDeletionQueue = _temporaryDownloadDeletionQueue.then((
+        _,
+      ) async {
```
nielsenko: Thanks for adding the cleanup. Not a fan of the timer here though. What happens across restarts? Should we use a `deleteWhere(expiration < now)` scheduled future call instead?

packages/serverpod/lib/src/server/session.dart
```diff
@@ -384,163 +384,133 @@ final class StorageAccess {
...
-  }
+  }) => _storage(
+    storageId,
+  ).publicDownloadUrl(session: _session, path: path);
 
   /// Bulk lookup of a list of public links to files given a list of paths in
   /// the storage. If any given file isn't public or if no such file exists,
   /// null is stored at the corresponding position in the output list. Saves
   /// on server roundtrips if a large number of public URLs must be fetched,
   /// relative to calling [getPublicUrl] via an endpoint for each one.
```
nielsenko:
```suggestion
  /// relative to calling [publicDownloadUrl] via an endpoint for each one.
```

nielsenko (commented):
Mostly questions

packages/serverpod/lib/src/cloud_storage/database_cloud_storage.dart
Crazelu: `fileExists` should be used here instead

packages/serverpod/lib/src/cloud_storage/database_cloud_storage.dart
Crazelu: I did consider using a future call but that feature could be disabled.
The deletion here is best-effort. Accounting for restarts will require some periodic cleanup mechanism that doesn't fit well in this interface.

I also considered deletion on file retrieval but didn't like that.

packages/serverpod/lib/src/cloud_storage/public_endpoint.dart
Crazelu: Thanks for the callout.
We should definitely special case this. `upload` already does a similar thing with `storeUnverifiedFile`

packages/serverpod/lib/src/cloud_storage/public_endpoint.dart
Crazelu: Your assumption is correct. The intent is to align the implementation more closely with the S3 and GCP ones. Strictly single use (as in, produce at most one successful upload) is now enforced with `preventOverwrite`.

Crazelu: Also in the old code, an invalid key consumed the authorization with the early deletion, preventing a later upload with the valid key from executing.

packages/serverpod/lib/src/cloud_storage/public_endpoint.dart
Crazelu: 👍 It should return `false`

packages/serverpod/lib/src/cloud_storage/database_cloud_storage.dart
Crazelu: Done in [3882975](https://github.com/serverpod/serverpod/pull/5621/commits/3882975fe38c9c4baa0d811e64dd0569d7823d8d)

packages/serverpod/lib/src/cloud_storage/public_endpoint.dart
Crazelu: Done in [4e5bc85](https://github.com/serverpod/serverpod/pull/5621/commits/4e5bc85db8c65bf38d953e78e4da127123777412)

packages/serverpod/lib/src/cloud_storage/public_endpoint.dart
Crazelu: Done in [a45278c](https://github.com/serverpod/serverpod/pull/5621/commits/a45278c906672dfe1f1b60ca38c2f05337d955b5)

packages/serverpod/lib/src/cloud_storage/database_cloud_storage.dart
nielsenko: What about `publicDownloadUrl`?

packages/serverpod/lib/src/cloud_storage/database_cloud_storage.dart
nielsenko: This is not a transient problem, this is a leak. It doesn't sit well with me.

Also, the closure captures `session` for and uses it after `delay`. By then the `MethodCallSession` is long closed, and you are not supposed to call any methods on a closed session. The error will be swallowed here and only be visible if verbose logging is enabled.

nielsenko: Calling `CloudStorageDirectDownloadEntry.db.deleteWhere(expiration < now) `on each `temporaryDownloadUrl` insert (and/or on the failed-lookup path in `temporaryFile`) could be away to avoid using future calls.

Crazelu: You're right. I was avoiding the overhead of deletion on insert/find but that is cleaner and perhaps acceptable since the database cloud storage implementation is not encouraged for production use.

packages/serverpod/lib/src/cloud_storage/database_cloud_storage.dart
Crazelu: Covered for `publicDownloadUrl` in [a6a3eb9](https://github.com/serverpod/serverpod/pull/5621/commits/a6a3eb9f6c3cb6fafc881c4a9ea13d190bc841a8)

packages/serverpod/lib/src/cloud_storage/database_cloud_storage.dart
Crazelu: Applied your suggestion in [bd463f7](https://github.com/serverpod/serverpod/pull/5621/commits/bd463f7700a70f2360f550944a99e54257d666ee)

nielsenko (approved):
LGTM

PR #5625, opened 2026-08-21
feat: Allow overriding the default cache header for `StaticRoute` from env vars

marcelomendoncasoares:
Missing piece for CDN on Cloud, since serving is from the `StaticRoute`.

Note that the env var for `StaticRoute` will also apply to `SpaRoute`, but not to `FlutterRoute`. Hence, the previous `SERVERPOD_WEB_SERVER_FLUTTER_CACHE_CONTROL` env var is still valid.

## Pre-launch Checklist

## Breaking changes

None.


packages/serverpod/lib/src/web_server/routes/spa_route.dart
```diff
@@ -55,15 +55,18 @@ class SpaRoute extends Route {
     subRouter.use(
       '/',
       FallbackMiddleware(
-        fallback: StaticRoute.file(fallback),
+        fallback: StaticRoute.file(
+          fallback,
+          cacheControlFactory: cacheControlFactory,
```
FXschwartz: Question on this one: Since this now inherits the user's `cacheControlFactory` / the env var. It looks like a `SpaRoute` with `publicImmutable` ends up long-caching index.html, and on Cloud with the env var set the app shell gets CDN-cached which could cause a stale app after a deploy.

I think that's why `FlutterRoute` pins its index fallback to `privateNoCache()` for this reason, even when its env var is set. 

Is this the behavior we want here? If so I think it would be a breaking change for `SpaRoute` users.

packages/serverpod/lib/src/web_server/routes/spa_route.dart
```diff
@@ -55,15 +55,18 @@ class SpaRoute extends Route {
...
+        ),
         on: (response) => response.statusCode == 404,
       ).call,
     );
 
     StaticRoute.directory(
       directory,
       cacheBustingConfig: cacheBustingConfig,
-      cacheControlFactory: cacheControlFactory ?? (_, _) => null,
+      cacheControlFactory: cacheControlFactory,
```
FXschwartz, line 105: The `StaticRoute` doc comments all mention the env var now, but `SpaRoute`'s default changed too and its `cacheControlFactory` field/constructor docs don't mention `SERVERPOD_WEB_SERVER_STATIC_CACHE_CONTROL`. 

`FlutterRoute` documents its var on the field, the constructor, and the class doc.

packages/serverpod/test/web_server/static_route_cache_control_test.dart
```diff
@@ -0,0 +1,158 @@
...
+import 'package:test/test.dart';
+
+import '../server/test_helpers/empty_endpoints.dart';
+
+void main() {
+  const environmentCacheControlTestName =
+      'Given a static cache control environment variable and no Dart factory, '
+      'when a static file is requested, '
+      'then the environment cache control is applied.';
+  test(environmentCacheControlTestName, () async {
```
FXschwartz: These cover `StaticRoute.directory`, but the two claims in the description, applies to `SpaRoute`, doesn't apply to `FlutterRoute`, aren't tested.

I traced both and they hold, but a test for each would pin it. The existing `SpaRoute` cache test only requests a real file with an explicit factory, so it doesn't catch either. Same goes for `.file` and `.withCacheBusting` keeping its 1 year default when the var is unset.

packages/serverpod/lib/src/web_server/routes/static_route.dart
```diff
@@ -91,26 +108,42 @@ class StaticRoute extends Route {
...
-        cacheControl: cacheControlFactory,
+        cacheControl:
+            cacheControlFactory ?? _cacheControlFactoryFromEnvironment(),
       ).asHandler,
       tailMatch: false,
       host: host,
     );
   }
 
+  static CacheControlFactory _cacheControlFactoryFromEnvironment({
```
FXschwartz: nit: This is almost the same as `FlutterRoute._cacheControlFactoryFromEnvironment`, and this version already takes the fallback param that covers the difference.

Could be one shared helper taking the env var name and fallback?

packages/serverpod/test/web_server/static_route_cache_control_test.dart
```diff
@@ -0,0 +1,158 @@
...
+    if (await _rerunWithCacheControlEnvironment(
+      testName: environmentCacheControlTestName,
+      value: 'public, max-age=60, s-maxage=86400',
+    )) {
+      return;
+    }
+
+    final (:pod, :tempDir) = await _createStaticFileServer();
```
FXschwartz: nit: This teardown block is duplicated in both server tests, could move the `addTearDown` into `_createStaticFileServer`?

FXschwartz (changes requested):

packages/serverpod/lib/src/web_server/routes/spa_route.dart
marcelomendoncasoares: This is indeed a drift in the behavior that I didn't think completely and I don't think is worth on this PR. Changing to some value was needed because now the `StaticRoute` can have a default cache header from an env var, and passing `null` would silently adopt it on the fallback. Since previously it was impossible to configure a cache header for the fallback, I changed to `noCacheControl`. We can discuss separately if the best would be to let the fallback adopt the `StaticRoute` value from the env var.

To avoid coupling the two, now each one has its own default cache control env var.

packages/serverpod/lib/src/web_server/routes/spa_route.dart
marcelomendoncasoares: Documented at https://github.com/serverpod/serverpod/pull/5625/commits/7fe242ccb0e1fe172974523988c746778b6b15e6 (and enhanced on https://github.com/serverpod/serverpod/pull/5625/commits/cae089e1394ed2ea6d6b80212a573db6fc8c2138 after the behavior change).

packages/serverpod/test/web_server/static_route_cache_control_test.dart
marcelomendoncasoares: Refactored the tests at https://github.com/serverpod/serverpod/pull/5625/commits/7fe242ccb0e1fe172974523988c746778b6b15e6.

packages/serverpod/lib/src/web_server/routes/static_route.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5625/commits/7fe242ccb0e1fe172974523988c746778b6b15e6. Now they all share a common helper.

packages/serverpod/test/web_server/static_route_cache_control_test.dart
marcelomendoncasoares: Tests isolation is much clearer now after https://github.com/serverpod/serverpod/pull/5625/commits/7fe242ccb0e1fe172974523988c746778b6b15e6 (I mirrored the same pattern I applied on the `FlutterRoute` tests).

packages/serverpod/lib/src/web_server/routes/spa_route.dart
FXschwartz: Nice! I like the separate env var. It protects the fallback without coupling the two routes. 

Testing the new behavior looks good. The fallback stays `uncached` even with a Dart factory or either env var set, the STATIC var no longer touches `SpaRoute`, and the tests pin all of it down including the FlutterRoute isolation.

FXschwartz (approved):

PR #5627, opened 2026-08-21
fix: Restore inbound foreign keys when a table is recreated

### LINKED ISSUE #5626: Table recreation drops inbound foreign keys from other tables and never restores them
### Describe the bug

When a model change cannot be applied in place, the migration is resolved as `deleteTable` + `createTable`. On PostgreSQL this emits `DROP TABLE "<table>" CASCADE`, which also drops the foreign key constraints that **other** tables declare against the dropped table. The recreated table brings back only its own outbound keys, and a table that merely *references* it gets no action in the migration — so its constraint is never rebuilt.

The result is a database that silently diverges from the `definition.json` of that same migration: the constraint is gone, and with it its `ON DELETE` action, so deleting a parent now leaves orphaned children instead of removing them. No error, no warning.

Only existing databases are affected. A fresh one is built from `definition.sql`, which is correct — so a test suite that creates its database per run stays green while every already-migrated environment drifts.

### To reproduce

1. Define two models where one holds a relation into the other:

   ```yaml
   class: IngestionStage
   table: ingestion_stage
   fields:
     anInt: int
   ```

   ```yaml
   class: IngestionItemOutcome
   table: ingestion_item_outcome
   fields:
     stage: IngestionStage?, relation(onDelete=Cascade)
   ```

2. Create and apply the migration, so an existing database is in place.
3. Add a non-nullable field with no default to `IngestionStage` — a change that cannot be applied in place, forcing the table to be deleted and recreated.
4. Run `serverpod create-migration --force`.
5. Inspect the generated `migration.sql`:

   ```console
   $ grep -c 'ingestion_item_outcome_fk_0' migration.sql
   0
   ```

   The only `ADD CONSTRAINT` is the recreated table's own outbound key. The same migration's `definition.sql` still declares `ingestion_item_outcome_fk_0` with `ON DELETE CASCADE`.

6. Apply the migration to the existing database and inspect the live schema — the constraint is missing.

### Expected behavior

After applying `migration.sql` to an existing database, the schema matches `definition.json`. Foreign keys that other tables declare against a recreated table are re-added after the `CREATE TABLE`, with their `ON DELETE` / `ON UPDATE` actions intact.

### Library version

Serverpod 4.0.0-beta.2

### Platform information

PostgreSQL. SQLite is unaffected — it keeps the child's inline `REFERENCES` clause across a parent drop and recreate.

### Additional context

Present since migrations shipped — `e384a199b` "feat: Initial work on database migrations. (#936)" (May 2023) introduced both the `CASCADE` drop and the recreate branch with no inbound-key restoration. `6b470e447` only moved the code between files.

`cb4831b41` "fix: Cascade recreation of table to related tables. (#3195)" (Feb 2025, 2.4.0) fixed the sibling case: a table *deleted for good* now drags its referencing tables into the drop and recreate, so their keys return via the create path. It never covered the recreate branch, so the bug survived the one change that looked directly at this area.

Existing deployments are not repaired retroactively by a fix — `serverpod create-repair-migration` diffs the live schema against the target and re-adds the missing constraints.

Suggested fix: when resolving a table recreation, re-emit the foreign keys the target definition declares against the dropped table, skipping constraints that a `createTable`/`alterTable` action already emits (otherwise: duplicate `ADD CONSTRAINT`) and unmanaged tables, which the generated schema never writes to.

- [x] I want to work on this

marcelomendoncasoares:
Fixes: #5626

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/tests/serverpod_test_server/test_e2e_migrations/migrations_roundtrip_indexes_relations_test.dart
+++ b/tests/serverpod_test_server/test_e2e_migrations/migrations_roundtrip_indexes_relations_test.dart
@@ -1,4 +1,5 @@
 @Timeout(Duration(minutes: 5))
+import 'package:serverpod_database/serverpod_database.dart';
 import 'package:serverpod_test_server/test_util/migration_test_utils.dart';
 import 'package:serverpod_test_server/test_util/service_client.dart';
 import 'package:test/test.dart';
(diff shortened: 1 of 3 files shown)


FXschwartz (approved):
LGTM!

PR #5628, opened 2026-08-21
chore: Make generated import prefixes stable

marcelomendoncasoares:
Generated code aliases every import with a prefix that was numbered in the order references happened to be emitted (`_i1`, `_i2`, ...). As an example: adding one model to `serverpod_test_sqlite` for experimentation produced **1,380 insertions and 1,352 deletions** across its two `protocol.dart` files, for a real change of about fifteen lines. Even worst, bare changes on the code that uses the models (like the endpoints) could provide a change in the usage order and churn the prefixes for the entire library.

There are a few main consequences of this:

- It can be very hard to review changes to the generator through the output diff, specially when it touches the `protocol.dart` files.
- Introducing a simple model generates frequent merge conflicts on the `protocol.dart` files.
- Even small PRs can present an enormous diff size, which pollutes review and even degrades GitHub's review UI.

This PR solves this problem by assigning prefixes derived from the import URL alone, so the same URL always gets the same prefix. The same experiment mentioned above now produces **30 insertions and 0 deletions**.

### Strategy

A 7-character hash of the URL, plus a fixed table of short prefixes reserved for the 21 packages generated code imports most (i.e. `_is` for `package:serverpod/serverpod.dart`, `_isc` for the client, `_isd` for the database package, ...). Reserved prefixes are all shorter than a hashed one, so the two sets can never overlap.

Prefixes are assigned once per library from its complete set of imports, in sorted order. Nothing depends on the order references are emitted in.

### Benchmarking

To be sure that this approach wouldn't introduce performance regressions, it was measured:

- **Hash length.** At 7 characters, a 1% collision chance needs ~40,000 imports in a single library; the largest today has a few hundred. Distribution was verified against the birthday expectation over a 200k-URL corpus.
- **Size.** Generated code grows **0.42%** (146,991 → 147,613 lines) on the `serverpod_test_sqlite` library. Reserved prefixes contribute to keep the LOC increase minimal, but even without them it would increase by only 0.99%.
- **Assignment cost.** Emission is **142ms of a 10.8s** generation run, so collecting a library's imports before assigning its prefixes costs ~1%. That measurement is why full order-independence was affordable.

### Collisions

Two URLs can hash to the same prefix. Since the imports involved may belong to packages a user does not control, this never fails the build. Instead, one of them is aliased differently and generated code stays correct. Which one moves is decided deterministic based on the URLs themselves, so the outcome is reproducible run to run.

## Review suggestion

The bulk of the diff on this PR is regenerated code, since it touches absolutely every file in the project. Commits were organized to make review easy: the first one is the implementation, then a small simplification on pre-existing code, then the tests update, and finally the generated code.

## Pre-launch Checklist

## Breaking changes

None.


tools/serverpod_cli/test/generator/dart/import_prefix_test.dart
```diff
@@ -0,0 +1,457 @@
...
+      importPrefixFor('package:example/example.dart'),
+      isNot(importPrefixFor('package:other/other.dart')),
+    );
+  });
+
+  group('Given a reference to a fixnum implementation library,', () {
+    // package:fixnum 1.2.0 moved its implementation into platform specific
+    // libraries under src/ that must not be imported directly.
+    late var allocator = StableImportAllocator();
+    late var allocated = allocator.allocate(
```
FXschwartz: I think there's a hidden ordering dependency here. Allocated is late, so it only gets evaluated when the first test reads it. The second test only reads `allocator.imports`, so running it on its own fails with an empty imports list:
```
dart test test/generator/dart/import_prefix_test.dart --plain-name "then the public library is what gets imported"
```

Moving the allocator setup into a setUp should fix it.

tools/serverpod_cli/test/generator/dart/import_prefix_test.dart
```diff
@@ -0,0 +1,457 @@
...
+    setUp(() {
+      allocator = StableImportAllocator();
+      allocator.allocate(
+        const Reference('Example', 'package:example/example.dart'),
+      );
+      allocator.assignPrefixes();
+    });
+
+    test(
+      'when allocating a reference to an import it never collected, '
```
FXschwartz, line 398: Tiny nit: might be worth swapping these two tests so the happy path comes first.

tools/serverpod_cli/lib/src/generator/dart/library_generators/util/custom_allocators.dart
```diff
@@ -1,12 +1,342 @@
...
+/// URL, see [importPrefixFor].
+///
+/// [resolveUrl] maps a referenced URL to the one to import, for generators
+/// that emit references against one package and import another. Resolving
+/// here rather than rewriting the emitted source keeps the prefix derived from
+/// the URL that actually ends up in the import directive.
+class StableImportAllocator implements AssigningAllocator {
+  final _PrefixRegistry _registry;
+
+  StableImportAllocator({String Function(String url)? resolveUrl})
```
FXschwartz: nit: `String Function(String url)`? shows up in five signatures now, a `typedef UrlResolver = String Function(String url);` could name the concept once. Totally optional.

FXschwartz (approved):
One minor finding and then just some nits. I'm going to approve and it's up to you if the nits are worth fixing. Otherwise let's get the release out! 🚀

tools/serverpod_cli/lib/src/generator/dart/library_generators/util/custom_allocators.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5628/commits/b630821346e219cd53525636c9863ce2a809003b.

tools/serverpod_cli/test/generator/dart/import_prefix_test.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5628/commits/b630821346e219cd53525636c9863ce2a809003b.

tools/serverpod_cli/test/generator/dart/import_prefix_test.dart
marcelomendoncasoares: Done at https://github.com/serverpod/serverpod/pull/5628/commits/b630821346e219cd53525636c9863ce2a809003b.

PR #5630, opened 2026-08-22
feat(database, generator): Type-Safe Model Projections, Relational Joins, and JSON Field Selection

### LINKED ISSUE #5492: Type-Safe Model Projections for Partial Object Fetching and Relational Joins
### Problem to Solve

Currently, Serverpod's ORM always fetches complete model objects from the database. When dealing with large models, especially those containing heavy JSON payloads or large text columns fetching every column for every query can lead to unnecessary network overhead and memory consumption.

There is currently no built in, type safe way to query specific columns (partial object fetching) using the generated repository classes.

### Proposal

Introduce a projections block in the .spy.yaml or .spy model files to define lightweight, partial models. Update the Database ORM to allow fetching these partial models using a type safe select parameter.

1. **YAML Definition (projections keyword)**
Developers can define projections within their existing model definitions. These projections inherit their field definitions from the base model but do not generate their own tables or migrations.

```yaml
class: Post
table: posts
fields:
  title: String
  content: String
  summary: String
  
projections:
  PostFeed:
    fields:
      - title
      - summary
```

**ORM Integration (Type-safe select parameter)**
The code generator creates a lightweight PostFeed class. The base model's repository class is updated to expose projection methods (e.g., findByIdWithSelect, findWithSelect, findFirstRowWithSelect,) that accept a non nullable select parameter.

```dart
final feedItem = await Post.db.findByIdWithSelect<PostFeed>(
  session,
  postId,
  select: (t) => [t.title, t.summary],
);
```

2. Object Relations (Nested Projections)
To support querying partial objects across database relations while maintaining strict type safety, developers can specify a nested projection type in their YAML files.

```.yaml
# author.spy.yaml
class: Author
table: author
fields:
  name: String
  bio: String?

projections:
  AuthorPreview:
    fields:
      - name

# comment.spy.yaml
class: Comment
table: comment
fields:
  text: String
  post: Post?, relation

projections:
  CommentSummary:
    fields:
      - text

# post.spy.yaml
class: Post
table: post
fields:
  title: String
  author: Author?, relation 
  comments: List<Comment>?, relation

projections:
  PostFeedView:
    fields:
      - title
      # Explicitly maps the object relation to the AuthorPreview projection class
      - author: AuthorPreview 
      # Maps the list relation to the CommentSummary projection class
      - comments: CommentSummary
```

This generates a PostFeedView class where the author field is strictly typed as AuthorPreview? and comments as List<CommentSummary>?.

When querying, developers use includeWithSelect<T> (mirroring the standard include pattern) to instruct the ORM to join the table and map the result to the specified projection:

```dart
final result = await Post.db.findFirstRowWithSelect<PostFeedView>(
  session,
  select: (t) => [t.id, t.title],
  include: Post.include(
    // Tell the ORM to join the Author table, but map it to AuthorPreview
    author: Author.includeWithSelect<AuthorPreview>((a) => [a.id, a.name]),
    
    // Also works for list (hasMany) relations!
    comments: Comment.includeListWithSelect<CommentSummary>((c) => [c.id, c.text])
  ),
);
```

**### Why this design avoids user confusion**
Currently, users performing joins with the ORM use Model.include(...), where the resulting object is a full instance of the related model (e.g., Address.include(...) gives back a full Address object with all of its fields).

Introducing Model.includeWithSelect<ProjectionClass>(...) provides an explicit, opt-in behavior for projections:

1. It unambiguously specifies the expected type of the projection via generics (e.g., <AddressStreet>), so the user knows exactly what they are getting back.
2. It requires a select callback ((a) => [a.id, a.street]) which clearly indicates to the developer that only a subset of fields is being fetched.
3. It clearly separates the standard .include(...) (which maps strictly to the parent table's model and automatically infers fields) from projected joining operations.

### **How the SQL queries are built**

The `sql_query_builder.dart` module creates `LEFT JOIN` operations exactly as it does for standard `.include` calls, but with one critical change: **instead of collecting all columns from the joined table, it dynamically checks the `selectedColumns` property introduced in the `Include` base class**.

For **Object Relations (e.g. `1:1` or `n:1`)**:
- The `_gatherIncludeColumns` method only adds the explicitly selected columns to the SQL `SELECT` statement using table aliases (e.g., `SELECT "post"."title", "author"."name" FROM "post" LEFT JOIN "author" ...`).
- When the row returns from Postgres, the `serializationManager.deserialize<PostFeedView>` automatically instantiates the projected container class instead of the standard `Post` class.

For **List Relations (e.g. `1:n` `comments`)**:
- Serverpod inherently handles List Relations by triggering a secondary database query using `IncludeList` (`SELECT "text" FROM "comment" WHERE ...`).
- For projections, the generated `includeListWithSelect<ProjectedClass>` populates `selectedColumns`. The ORM implicitly injects the foreign key field into the `select` clause behind the scenes (even if the user doesn't request it in the projection).
- This hidden foreign key enables the Dart memory parser (`query_result_parser.dart`) to successfully fold and group the returned rows (`CommentSummary`) into nested lists on the parent objects (`PostFeedView`) without exposing the foreign key to the end-user's projected object.
This architecture provides a clean separation of concerns (scalar columns in `select`, relational joins in `include`) that existing Serverpod developers are already comfortable with.

### Use Case

This feature significantly improves database efficiency, memory consumption, and API response payload size for high-traffic applications. For example, when rendering social media feeds containing thousands of posts, fetching only the title, summary, and author's name prevents the transfer of heavy markdown content, media URLs, and extensive metadata. Similarly, in e-commerce applications, displaying product catalogs only requires a subset of data like price, thumbnail, and name. By reducing over-fetching, this approach minimizes mobile data usage and allows clients to deserialize payloads significantly faster. Overall, it provides a first-class, strongly-typed developer experience without the need to fall back to raw SQL queries or `unsafeQuery`.

### Alternatives

One common alternative is creating separate tables or models specifically for summaries (e.g., a post_summary table) to avoid querying the full payload, but this leads to duplicate data storage and synchronization challenges. Another approach is writing raw SQL using unsafeQuery to select only the necessary columns. However, doing so completely sacrifices Serverpod's core value: end-to-end type safety, code completion, IDE linting, and automatic Dart/Flutter class generation. Lastly, developers can fetch the entire object tree from the database to the server and map it into Data Transfer Objects (DTOs) before sending it to the client. While this solves the client-side network overhead, it still strains the internal database-to-server network bandwidth and significantly increases memory consumption on the server.

### Additional context

One common alternative is creating separate tables or models specifically for summaries (e.g., a post_summary table) to avoid querying the full payload, but this leads to duplicate data storage and synchronization challenges. Another approach is writing raw SQL using unsafeQuery to select only the necessary columns. However, doing so completely sacrifices Serverpod's core value: end-to-end type safety, code completion, IDE linting, and automatic Dart/Flutter class generation. Lastly, developers can fetch the entire object tree from the database to the server and map it into Data Transfer Objects (DTOs) before sending it to the client. While this solves the client-side network overhead, it still strains the internal database-to-server network bandwidth and significantly increases memory consumption on the server.

- [x] I want to work on this

binemmanuel:
## Summary

This PR introduces **Model Projections** to Serverpod's ORM and code generator. Projections enable developers to declare and query lightweight, type-safe subsets of database models, relational joins, and nested JSON/JSONB column fields without fetching full table rows or unneeded heavy data.

Fixes [#5492](https://github.com/serverpod/serverpod/issues/5492)

---

### 1. Declarative Projections (`projections` block in `.spy.yaml`)
Define projection models directly within existing `.spy.yaml` files. Projections automatically inherit field types from the parent model and do not generate redundant database tables or migrations (`manageMigration: false`).

### 2. Relational Projections & Field Forwarding
- **Relation Sub-Projections (`relation -> TargetProjection`)**: Fetch related entities (1:1, 1:n, n:n) mapped to their respective projection types using Serverpod's `include` system under the hood.
- **Relational Field Flattening (`relation.field`)**: Flatten a nested relation's column into the parent projection model (e.g. `address.street` -> `addressStreet`).
- **Nullability Preservation**: Accurately computes field nullability based on database foreign key constraints (`ObjectRelationDefinition.nullableRelation`) so non-optional relations remain non-nullable and required.

### 3. JSON Field Selection & JSON Sub-Projections
- **SQL JSON Key Selection (`jsonField.key`)**: Extract specific keys from a JSON/JSONB column directly into flattened fields on the projection model (e.g. `jsonField.text` -> `jsonFieldText`, `jsonField.value` -> `jsonFieldValue`). Generated queries use database column expressions (`jsonKey`) to extract only the specified keys via SQL (`"table"."col"->'key'`).
- **JSON Sub-Model Projections (`jsonField -> ProjectedJsonModel`)**: Extract a typed subset of a JSON model column into a dedicated projected sub-model.
- **Null Safety**: Gracefully handles `null` JSON columns and omitted optional properties.

### 4. Dedicated Type-Safe Repositories
Every projection gets a generated `.db` repository attached (e.g. `ProjectedUserJsonField.db`), providing standard, strongly-typed query methods:
- `find(session, ...)`
- `findFirstRow(session, ...)`
- `count(session, ...)`
- `findById(session, id, ...)`

---

### Model Definition ([`projected_user.spy.yaml`](tests/serverpod_test_server/lib/src/models/projected_user.spy.yaml))

```yaml
class: ProjectedUser
table: projected_user
fields:
  name: String
  address: ProjectedAddress?, relation
  orders: List<ProjectedOrder>?, relation
  jsonField: ProjectedJsonField?

projections:
  # 1. Flattened JSON Key Projection
  ProjectedUserJsonField:
    select:
      - name
      - jsonField.text
      - jsonField.value

  # 2. Typed JSON Sub-Model Projection
  ProjectedUserSimpleJson:
    select:
      - name
      - jsonField -> ProjectedJsonFieldSimple

  # 3. Relational Join Projection
  ProjectedUserStreetAddress:
    select:
      - name
      - address -> ProjectedAddressStreet

  # 4. Relational Field Flattening
  ProjectedUserAddressStreetOnly:
    select:
      - name
      - address.street
```

### JSON Sub-Model Definition ([`projected_json_field.spy`](tests/serverpod_test_server/lib/src/models/projected_json_field.spy))

```yaml
class: ProjectedJsonField
fields:
  text: String
  value: int
  valueA: bool
  valueB: double
  valueC: String?
  dateValue: DateTime
  listA: List<int>
  mapA: Map<String, int>

projections:
  ProjectedJsonFieldSimple:
    select:
      - text
      - value
```

---

## Dart Query Usage

```dart
// 1. Fetch with flattened JSON fields (SQL extracts only jsonField->'text' and jsonField->'value')
var userJson = await ProjectedUserJsonField.db.findFirstRow(session);
print(userJson?.name);          // 'John'
print(userJson?.jsonFieldText);  // 'sample text'
print(userJson?.jsonFieldValue); // 42

// 2. Fetch with typed JSON sub-projection
var userSimpleJson = await ProjectedUserSimpleJson.db.findFirstRow(session);
print(userSimpleJson?.jsonField?.text);  // 'sub-projected'
print(userSimpleJson?.jsonField?.value); // 99

// 3. Fetch with 1:1 relational sub-projection
var userAddress = await ProjectedUserStreetAddress.db.findFirstRow(session);
print(userAddress?.name);
print(userAddress?.address?.street); // Typed ProjectedAddressStreet

// 4. Fetch with flattened relation field
var flatAddress = await ProjectedUserAddressStreetOnly.db.findFirstRow(session);
print(flatAddress?.name);
print(flatAddress?.addressStreet); // 'Main Street'
```

---

### `tools/serverpod_cli`
- **[`analyzers.dart`](tools/serverpod_cli/lib/src/generator/analyzers.dart)**:
  - Implemented parsing and validation for the `projections` YAML block.
  - Added support for dot-notation field forwarding (`relation.field`, `jsonField.key`).
  - Added support for arrow-notation sub-projections (`relation -> TargetProjection`, `jsonField -> JsonProjection`).
  - Resolved field nullability accurately using `ObjectRelationDefinition.nullableRelation` so non-optional relations preserve non-nullable/required types.
  - Generated appropriate field types, nullability, and migration flags (`manageMigration: false`).
- **[`model_library_generator.dart`](tools/serverpod_cli/lib/src/generator/dart/library_generators/model_library_generator.dart)**:
  - Generated projection classes and their `fromJson` / `toJson` / `copyWith` methods.
  - Implemented `include` builders supporting `jsonKey` column expressions and `internal_` constructors.
- **[`repository_classes.dart`](tools/serverpod_cli/lib/src/generator/dart/library_generators/class_generators/repository_classes.dart)**:
  - Generated projection-specific repository classes delegating queries to base tables while preserving type safety.
- **[`class_generators_util.dart`](tools/serverpod_cli/lib/src/generator/dart/library_generators/util/class_generators_util.dart)**:
  - Handled operator precedence for nested JSON fallback deserialization expressions.

### `packages/serverpod_database`
- **[`columns.dart`](packages/serverpod_database/lib/src/concepts/columns.dart)**:
  - Introduced `@internal ColumnJsonField<T>` and `.jsonKey<R>()` on `ColumnSerializable` and `ColumnStructured` for SQL-level JSON key extraction (`"table"."col"->'key'`).
- **[`column_alias_resolver.dart`](packages/serverpod_database/lib/src/util/column_alias_resolver.dart)** & **[`query_result_parser.dart`](packages/serverpod_database/lib/src/util/query_result_parser.dart)**:
  - Added support for resolving and parsing prefixed JSON key column selections.
- **[`database_connection.dart`](packages/serverpod_database/lib/src/interface/database_connection.dart)** (Interface), **[`database.dart`](packages/serverpod_database/lib/src/database.dart)**, **[`database_connection.dart`](packages/serverpod_database/lib/src/adapters/postgres/database_connection.dart)** (Postgres), & **[`database_connection.dart`](packages/serverpod_database/lib/src/adapters/sqlite/database_connection.dart)** (SQLite):
  - Implemented `findAsJson` and `findFirstRowAsJson` across connection adapters and database proxies.

### `packages/serverpod_test`
- **[`test_database_proxy.dart`](packages/serverpod_test/lib/src/test_database_proxy.dart)**:
  - Added proxy forwarding for `findAsJson` and `findFirstRowAsJson`.

### Tests (`tests/serverpod_test_server`)
- Added comprehensive integration tests in **[`projection_relation_test.dart`](tests/serverpod_test_server/test_integration/database_operations/model_projection/projection_relation_test.dart)** testing:
  - 1:1, 1:n, and n:n relation projections.
  - Flattened relational field forwarding.
  - Flattened JSON key selection (`jsonField.text`, `jsonField.value`, `jsonField.mapA`, `jsonField.listA`, `jsonField.dateValue`).
  - JSON sub-model projections (`jsonField -> ProjectedJsonFieldSimple`).
  - Null JSON column handling.

---

## Pre-launch Checklist

- [x] I read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code with `dart format`.
- [x] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation.
- [x] All existing and new tests are passing (`dart test`).

---

## Breaking Changes

_None._


binemmanuel:

binemmanuel:
cc @marcelomendoncasoares ready for review whenever you have some time! This implements the type safe model projections, relational joins, and JSON field extraction discussed in [#5492](https://github.com/serverpod/serverpod/issues/5492)

Silfalion:
Hey, very interested in this PR, thanks for putting it together — the projection support looks really useful. I was reading through the lower-level implementation, and it seems the PR already contains most of what would be needed to also support flexible, one-off selects.

Since `findAsJson` and `Include.selectedColumns` now exist, I think it would be valuable to expose that capability publicly without requiring users to define a named YAML projection for every selection shape.

Currently, the underlying functionality appears to be reachable through:

```dart
final rows = await session.db.findAsJson<Post>(
  include: PostInclude.internal_(
    selectedColumns: [Post.t.id, Post.t.title],
  ),
);
```

A public API could expose this directly:

```dart
final rows = await session.db.findAsJson<Post>(
  selectedColumns: [Post.t.id, Post.t.title],
);
```

Named projections would remain the better option for reusable response types and nested structures, while this would provide a practical escape hatch for one-off queries.

It would also be great to have a typed result instead of exposing only `Map<String, dynamic>`. For example, a `SelectedRow` could allow values to be read using the same column objects used for the selection:

```dart
final rows = await Post.db.select(
  session,
  columns: (t) => [t.id, t.title],
);

final String title = rows.first.read(Post.t.title);
```

`SelectedRow` is just a proposed wrapper here, not something currently in the PR. Its `read(Column<T>)` method would return `T`, avoiding string keys and manual casts. `Map<String, dynamic>` could still be the initial API if adding a typed result is too large for this PR.

While looking through the implementation, I also noticed two possible issues:

- `findByIdAsJson` currently creates a `ColumnInt` and casts the ID to `int`, which breaks models using UUID IDs. It could use the generated ID column, as the existing `findById` does:

  ```dart
  where: table.id.equals(id),
  ```

- Matching rows disappear when every selected value is null. The parser skips null values, treats the resulting empty map as “no row,” and `findAsJson` filters it out. If five rows match and the selected nullable field is null in all five, the result should still contain five rows with null values—not an empty list.

For root rows, preserving the selected null values would retain the rows correctly. For optional joined rows, the ID could be selected internally to distinguish “the related row exists but its selected fields are null” from “there is no related row.”

Overall, this looks very close to supporting both generated projections and flexible selects without needing two separate query implementations.

binemmanuel:
@Silfalion, thanks a lot for taking a close look at the implementation and sharing these insights!

### 1. On `SelectedRow` and Ad-Hoc Column Selections
When building this out, I actually considered exposing ad-hoc column selections and exploring a `SelectedRow` wrapper, but deliberately decided to keep `findAsJson` and `Include.selectedColumns` marked `@internal` due to a few key trade-offs:

- **Developer Ergonomics:** Reading fields directly via `post.title` is much cleaner and more natural than repetitive `.read(Post.t.title)` method calls.
- **True Compile-time Safety:** With YAML projections, unselected columns simply do not exist on the generated model, catching errors at compile time. With a `SelectedRow` pattern, a developer can still call `row.read(Post.t.authorId)` on a column that was omitted from the select list, leading to runtime failures.
- **Client & Endpoint Serialisation:** Named YAML projections seamlessly generate Flutter client models and endpoint serialisers out of the box, whereas `SelectedRow` would be confined strictly to server-side database logic.
- **Nested Relations:** Traversal of complex relations (like `author.name` or `comments`) is much cleaner and intuitive with generated models than traversing dynamic row wrappers.

Keeping `findAsJson` as internal engine plumbing felt like the right architectural boundary to ensure Serverpod's public ORM remains strictly model-driven and type-safe. We could always revisit a `SelectedRow` concept in a dedicated issue if there is a strong need for server-only ad-hoc queries.

---

### 2. Bug Fixes
- **UUID fix in `findByIdAsJson`:** Good catch on the `ColumnInt` / `int` casting. I have updated it to use `table.id.equals(id)` dynamically so it works seamlessly for UUIDs and custom primary key types.
- **Null rows in `query_result_parser.dart`:** In practice, generated YAML model projections never run into this because the code generator always injects `table.id` into `selectedColumns`. However, you are totally right that at the lower-level database engine (if someone runs a query with only nullable columns and no `id`), `query_result_parser.dart` would drop rows where all values are `null`. I am putting together a fix for `query_result_parser.dart` so the engine handles that gracefully as well.

Thanks again for the detailed review!

Silfalion:
Thanks so much for the detailed response and for explaining the reasoning behind the current design!

I agree that generated projections provide the best experience for reusable query shapes, client serialization, compile-time field availability, and complex nested models. They should absolutely be the recommended and documented approach for those cases.

Alongside projections, there is also a useful server-only case: a one-off query that only needs a few columns and does not need a reusable generated model. I think it would be valuable for the public ORM API to cover that case directly:

```dart
final rows = await session.db.findAsJson<Post>(
  selectedColumns: (post) => [
    post.id,
    post.title,
  ],
);
```

Returning `Map<String, dynamic>` would be reasonable for this explicit escape hatch. It would naturally have fewer guarantees than a generated projection, while still using generated, model-aware columns and avoiding raw SQL.

Nested relations could continue following Serverpod’s existing `include` structure, with each relation owning its selection:

```dart
final rows = await session.db.findAsJson<Post>(
  selectedColumns: (post) => [
    post.id,
    post.title,
  ],
  include: Post.include(
    author: Author.include(
      selectedColumns: (author) => [
        author.id,
        author.name,
      ],
    ),
    comments: Comment.includeList(
      selectedColumns: (comment) => [
        comment.id,
        comment.content,
      ],
    ),
  ),
);
```

This keeps the relationship path explicit. If both `author` and `editor` point to the same user table, their selections remain attached to the correct relation. One-to-one relations can be returned as nested maps, while one-to-many relations can remain nested lists.

Since the PR already carries selected columns through the include machinery internally, exposing `selectedColumns` on `findAsJson`, `include`, and `includeList` appears to fit naturally on top of the current implementation.

This could also be introduced in stages: root-table selection through `findAsJson` would cover the main ad-hoc use case, while public nested selection could follow through the include API.

A typed `SelectedRow` can remain a separate follow-up:

```dart
final String title = row.read(Post.t.title);
```

It would not prove at compile time that the requested column was included, so reading an omitted column would need to produce a clear runtime error. It would still preserve the selected column’s Dart type and avoid string-based map access, making it a useful server-side complement to generated projections.

That would leave Serverpod with two complementary paths: projections for reusable, fully generated query models, and `findAsJson` for explicit ad-hoc server-side selections.

And thank you as well for addressing the ID comparison with `table.id.equals(id)` and the case where every selected nullable column is `null`. I really appreciate how quickly you picked those up!

binemmanuel:
@Silfalion, thanks again for the thoughtful feedback!

Regarding the API parameter name: what are your thoughts on using **`select`** vs **`selectedColumns`**?
```dart
// Option A: `select` (concise and familiar, mirroring SQL)
final rows = await session.db.findAsJson<Post>(
  select: (post) => [post.id, post.title],
  include: Post.include(
    author: Author.include(
      select: (author) => [author.id, author.name],
    ),
  ),
);

// Option B: `selectedColumns` (more descriptive)
final rows = await session.db.findAsJson<Post>(
  selectedColumns: (post) => [post.id, post.title],
  include: Post.include(
    author: Author.include(
      selectedColumns: (author) => [author.id, author.name],
    ),
  ),
);

**Comment by Silfalion** [comment]
Thanks for considering this, @binemmanuel! I’d lean toward **`select`**: it’s more inline with Servedpod API's.

**Comment by binemmanuel** [comment]
Hi @Silfalion,

I've pushed an update implementing the ad-hoc column selection approach we discussed alongside model projections.

### Summary of Changes
1. **Public Ad-Hoc Selection (`select`)**:
   - `session.db.findAsJson<T>`, `session.db.findFirstRowAsJson<T>`, and `session.db.findByIdAsJson<T>` are now public and accept a `select` parameter (`SelectColumnsBuilder<dynamic>? select`):
     ```dart
     final rows = await session.db.findAsJson<Post>(
       select: (t) => [t.id, t.title],
     );
     ```
2. **Nested Relations via `include(select:)`**:
   - Generated model `Include` and `IncludeList` methods also accept `select:` for ad-hoc relation filtering:
     ```dart
     final row = await session.db.findFirstRowAsJson<Citizen>(
       select: (t) => [t.name],
       include: Citizen.include(
         company: Company.include(
           select: (c) => [c.name],
         ),
       ),
     );
     ```
3. **Database Engine & Parser**:
   - Handled column scoping and relation presence resolution so that nested relations with null/partial columns correctly resolve to `null` when the relation row is absent.

4. **Testing**:
   - Added comprehensive integration tests in `tests/serverpod_test_server/test_integration/database_operations/model_projection/projection_null_columns_test.dart` covering root and nested `select:` queries, verifying that unselected columns are omitted and return shapes are as expected.

**Comment by Silfalion** [comment]
Hi @binemmanuel,

Thanks so much for implementing this — this is exactly the public API direction I was hoping for, and I really appreciate the work you put into carrying `select` through the root queries and nested includes!

Looking through the updated implementation, I noticed two important cases that still seem unresolved:

1. The root `select` callback is currently declared as `SelectColumnsBuilder<dynamic>?`, which means `t` is dynamic:

```dart
select: (t) => [t.thisColumnDoesNotExist],
```

That passes static analysis and only fails at runtime. The generated nested callbacks are correctly typed—for example, `SelectColumnsBuilder<CompanyTable>`—so it would be great to preserve that same typed column access at the root.

If `session.db.findAsJson<Post>` cannot infer `PostTable`, two possible API shapes would be a generated `Post.db.findAsJson(...)` wrapper or a directly typed list such as:

```dart
select: [Post.t.id, Post.t.title],
```

2. I couldn’t find a test covering `includeList(select:)`, and the current list-relation path appears to require some columns internally even when they are not requested:

```dart
final rows = await session.db.findAsJson<Customer>(
  select: (customer) => [customer.name],
  include: Customer.include(
    orders: Order.includeList(
      select: (order) => [order.description],
    ),
  ),
);
```

The framework still needs the customer ID to fetch the orders and the order’s customer ID to attach each order to the correct customer. A clean way to preserve the public result shape seems to be using those columns only for the internal relation mapping, while still omitting them from the returned maps unless the caller explicitly selected them.

A test where both the root ID and child foreign key are omitted from the public selection would cover this case nicely.

Other than those two cases, this is looking really good. Thanks again for taking the suggestion on board and implementing it so quickly!

binemmanuel:
Hi @Silfalion,

Thank you for the review and suggestions! I've addressed the feedback in the latest update:

### 1. Root Column Selection & Type Safety
* **`session.db.findAsJson<Post>`**: Updated `select` on `session.db` and `DatabaseConnection` to accept `List<Column>? select` (e.g. `select: [Post.t.id, Post.t.title]`), aligning directly with `orderByList: [Post.t.id]`.
* **`Post.db.findAsJson`**: Generated strongly typed `findAsJson`, `findFirstRowAsJson`, and `findByIdAsJson` methods on the model repositories (`SelectColumnsBuilder<PostTable>`), allowing idiomatic typed access like `select: (post) => [post.id, post.title]`.

### 2. `includeList(select:)` & Key Management
* The query builder now automatically injects required relation keys (`_table.id` on the parent query and the foreign key on child list queries) behind the scenes when assembling list relations.
* When constructing the final JSON maps, unrequested internal keys are stripped so only the columns explicitly requested by the user are returned.
* Added integration test coverage in [`projection_null_columns_test.dart`](file:///Users/binemmanuel/Projects/serverpod/tests/serverpod_test_server/test_integration/database_operations/model_projection/projection_null_columns_test.dart) to verify that `includeList(select:)` accurately resolves relation lists while omitting unselected primary/foreign keys from the resulting maps.

### 3. Private Constructors
* Generated `Include` and `IncludeList` constructors have been made private (`_`) so they cannot be instantiated directly outside the builder helpers.

Silfalion:
Hi @binemmanuel,

Thank you so much for taking the time to address all of this so thoroughly! Having both the strongly typed repository API and the lower-level `List<Column>` API is a really thoughtful solution, and the hidden-key handling for `includeList(select:)` is exactly the behavior I was hoping for. I genuinely appreciate how carefully you’ve carried the suggestions through into both the implementation and the tests.

While reading through the updated hidden-key handling, I noticed one mixed-relation case where the current check may still mistake one table’s ID for another. The primary-key check currently compares only `columnName`:

```dart
var hasId = selectColumns.any(
  (column) => column.columnName == _table.id.columnName,
);
```

Since IDs from different tables are commonly all named `id`, an included object’s ID could satisfy that check even when the root ID itself is absent. For example:

```dart
final rows = await Customer.db.findAsJson(
  session,
  select: (customer) => [customer.name],
  include: Customer.include(
    address: Address.include(
      select: (address) => [address.street],
    ),
    orders: Order.includeList(
      select: (order) => [order.description],
    ),
  ),
);
```

`Address.id` is added internally for the object relation. Because its `columnName` is also `id`, the check could interpret it as the required `Customer.id` and skip adding the root ID needed to load `orders`.

Comparing the complete column identity—such as its table or query alias together with its column name—rather than the column name alone seems like it would preserve the intended behavior here. A test combining an object relation and an `includeList`, while omitting the root ID, would cover this case directly.

Thank you again for all the care you’ve put into this. The resulting API is looking genuinely useful!

P.S. One quiet heads-up: the test link in your previous comment currently uses a local `file://` URL rather than a GitHub URL. You may want to replace it with the repository link and remove that revision from the edit history, since the URL contains part of a local path.

binemmanuel:
Hi @Silfalion,

Thank you again for the eagle-eyed review and for the heads up on the link!

### 1. Mixed-Relation Key Comparison Fix
* Updated the primary-key and foreign-key inclusion checks in `sql_query_builder.dart` to compare `c.queryAlias == _table.id.queryAlias` (and `c.queryAlias == foreignColumn.queryAlias`) rather than `columnName` alone. This ensures that columns from included object relations sharing the same column name (e.g. `Address.id`) are not mistaken for root table keys.
* Added an integration test in `projection_null_columns_test.dart` querying `Organization` with both an included `City` (object relation) and `Person` (list relation) while omitting the root `Organization.id`, verifying that relations resolve correctly and unselected keys remain omitted from the resulting maps.

---

### 2. Design Question: `select:` on `Include` with typed `Model.db.find`
While implementing the tests, I noticed a design nuance regarding `select:` on `Include` and wanted to get your thoughts:

Because `Model.include(...)` and `Model.includeList(...)` are shared between `Model.db.findAsJson` and `Model.db.find`:
* In **`findAsJson`**, `select:` works with arbitrary column subsets because it returns raw JSON maps.
* In **`find`**, if a developer passes `select:` on an included relation (e.g. `people: Person.includeList(select: (p) => [p.id])`) and omits a **non-nullable** field required by `Person.fromJson` (e.g. `name: String`), Dart will throw a `TypeError` during deserialization (`type 'Null' is not a subtype of type 'String'`).

Our current approach keeps `Include` unified (with `select:` available for `findAsJson` and Model Projections) and relies on developers using `findAsJson` or a `.spy.yaml` Projection when they need partial fetching of entities with non-nullable fields.

Do you think keeping `Include` unified is the right balance here, or would you prefer a different mechanism (e.g. separate `includeAsJson` helpers, or specific error handling/guards)?

Looking forward to your thoughts!

Silfalion:
Thanks, this is a really good catch. You’re right: I had focused on the `findAsJson` path and missed that the same generated include types are also accepted by typed `find`. Keeping `select` on that shared public type would allow an invalid typed query to look valid until `fromJson` fails.

It seems there are two reasonable ways to handle this:

1. Add a guard to typed `find` that checks the include tree and rejects any include containing `selectedColumns`. This is the smaller change and would provide a clear error instead of the current `TypeError`. The downside is that IntelliSense would still allow the invalid call, and the developer would only discover the problem when running it.

2. Separate the include types used by typed and JSON queries. This requires a little more generator work, but seems like the stronger API because the analyzer and IntelliSense can reject the invalid combination before the query runs.

I think the second approach is preferable, with the guard optionally remaining as a fallback for manually implemented or forcibly cast include objects.

The generated factories could be divided into:

```dart
Person.include()          // Complete Person object
Person.includeList()      // Complete Person objects in a list

Person.includeJson()      // JSON object, optionally selectable
Person.includeJsonList()  // JSON list, optionally selectable
```

Here, “complete” does not mean that every relationship is fetched automatically. Relationships would still only be fetched when explicitly included. It means that every row which is fetched contains all the columns needed to construct its corresponding typed model.

The one-way compatibility could be represented through interfaces:

```dart
abstract interface class JsonCompatibleInclude implements Include {}

abstract interface class FullModelInclude
    implements JsonCompatibleInclude {}
```

The generated object and list types could then conceptually look like:

```dart
abstract interface class PersonJsonInclude
    implements JsonCompatibleInclude {}

abstract interface class PersonJsonIncludeList
    implements JsonCompatibleInclude {}

final class PersonInclude extends IncludeObject
    implements PersonJsonInclude, FullModelInclude {
  // No select argument.
}

final class PersonIncludeList extends IncludeList
    implements PersonJsonIncludeList, FullModelInclude {
  // No select argument.
}

final class _PersonJsonInclude extends IncludeObject
    implements PersonJsonInclude {
  @override
  final List<Column>? selectedColumns;
}

final class _PersonJsonIncludeList extends IncludeList
    implements PersonJsonIncludeList {
  @override
  final List<Column>? selectedColumns;
}
```

`Person.include()` and `Person.includeList()` would therefore produce full-model includes, while `Person.includeJson()` and `Person.includeJsonList()` would produce JSON-compatible includes that may select only certain columns.

The same relationship would apply to nested relations:

```dart
Person.include(
  orders: Order.includeList(),
)
```

This is safe for typed `find` because every included `Order` row is complete.

For partial nested results:

```dart
Person.includeJson(
  // select: null means all Person columns.
  orders: Order.includeJsonList(
    select: (t) => [t.id, t.total],
  ),
)
```

This returns all Person columns while selecting only some Order columns. This is also why `includeJson(select: null)` remains useful: the current table can be complete while one or more explicitly included relationships are partial.

Because `OrderIncludeList` would also implement `OrderJsonIncludeList`, a full nested list remains valid inside a JSON include:

```dart
Person.includeJson(
  orders: Order.includeList(),
)
```

The opposite would not be accepted:

```dart
Person.include(
  orders: Order.includeJsonList(
    select: (t) => [t.id],
  ),
)
```

That combination is unsafe for typed `find`, because Serverpod would still try to construct complete `Order` models from the partial rows.

At the lower level, the distinction could be carried through as:

```dart
find(
  ...,
  FullModelInclude? include,
)

findAsJson(
  ...,
  JsonCompatibleInclude? include,
)
```

So the guard would be a valid and smaller fix, but distinguishing the object and list variants through interfaces seems preferable because it prevents unsafe combinations in normal generated code rather than only detecting them at runtime.

binemmanuel:
Hi @Silfalion,

I’ve implemented the separated include type hierarchy as discussed, giving us compile-time safety across both typed and JSON queries, along with the defensive runtime guard as a fallback.

Here is a summary of what has been implemented:

### 1. Include Interface Hierarchy
Defined in `serverpod_database`:
```dart
/// Marker interface for includes compatible with JSON queries (`findAsJson`).
abstract interface class JsonCompatibleInclude implements Include {}

/// Marker interface for includes producing complete models compatible with typed queries (`find`).
abstract interface class FullModelInclude implements JsonCompatibleInclude {}
```

Because `FullModelInclude implements JsonCompatibleInclude`, we get **one-way compatibility**:
- A complete model include (`Person.include()`) can be passed directly to `findAsJson` or nested inside `Organization.includeJson(...)`.
- A partial JSON include (`Person.includeJson(select: ...)`) cannot be passed to typed `find`, as the compiler rejects it.

### 2. Code Generation Updates
For each model (e.g. `Person`):
- **Interfaces**:
  - `abstract interface class PersonJsonInclude implements JsonCompatibleInclude {}`
  - `abstract interface class PersonJsonIncludeList implements JsonCompatibleInclude {}`
- **Typed Includes (for `find`)**:
  - `final class PersonInclude extends IncludeObject implements PersonJsonInclude, FullModelInclude`
  - `final class PersonIncludeList extends IncludeList implements PersonJsonIncludeList, FullModelInclude`
  - Generated methods `Person.include(...)` and `Person.includeList(...)` do not expose a `select:` parameter.
- **JSON Includes (for `findAsJson`)**:
  - `final class _PersonJsonInclude extends IncludeObject implements PersonJsonInclude`
  - `final class _PersonJsonIncludeList extends IncludeList implements PersonJsonIncludeList`
  - Generated methods `Person.includeJson({..., select})` and `Person.includeJsonList({..., select})` expose `SelectColumnsBuilder<PersonTable>? select`.
- **Model Projections**:
  - Projection models generate static `include()` and `includeList()` methods that return `${baseClassName}JsonInclude` and `${baseClassName}JsonIncludeList` pre-populated with the projection's selected columns and relations.
- **Repositories**:
  - Model repositories accept `${className}Include? include` for typed `find*` methods and `${className}JsonInclude? include` for `find*AsJson` methods.

### 3. Defensive Runtime Guard
`Database.find`, `findFirstRow`, and `findById` maintain an `_assertNoSelectedColumns(include)` check to catch any custom or manually constructed includes that attempt partial selection during a typed query, throwing a descriptive `ArgumentError`.

### 4. Query Alias Disambiguation
Updated `SelectQueryBuilder` (`sql_query_builder.dart`) to check column aliases (`c.queryAlias == _table.id.queryAlias` and `c.queryAlias == foreignColumn.queryAlias`) rather than only column names, ensuring primary key checks across joins are disambiguated even when relation tables share column names like `id`.

### 5. Tests & Verification
- Updated generator test suite (`include_class_test.dart` — 21/21 passing).
- Added comprehensive integration tests covering typed queries, `findAsJson` partial selection, nested `includeJson`, one-way `FullModelInclude` compatibility, and relation key stripping (`projection_null_columns_test.dart` and `projection_relation_test.dart` — 20/20 passing).
- `dart analyze` across `serverpod_database`, `serverpod_cli`, `serverpod_test`, and `serverpod_test_server` reports 0 issues.

Silfalion:
Hi @binemmanuel,

Thank you so much again for everything you’ve put into this. The current separation between typed and JSON includes feels very clean, and I really appreciate how thoughtfully you worked through the discussion.

There are just two edge cases I wanted to mention for completeness:

- Because "IncludeList.include" remains publicly mutable, someone could construct a full typed include list and then manually assign a partial JSON include to it afterward. Normal factory usage is fully checked by IntelliSense, and the runtime guard already handles this unusual mutation safely.
- Generic code that deliberately stores an include using the broad "Include" type would need to retain or recover its more specific type before passing it to a typed query. That seems like a reasonable consequence of providing the stronger compile-time guarantee, particularly while this API is still unreleased.

You have the fullest view of the implementation and the direction of the PR, so I’ll leave the significance of those edge cases entirely to your judgment. I mainly wanted to make sure they were visible alongside the rest of the discussion.

Thank you again for all the time and care you’ve given this!

binemmanuel:
Hi @Silfalion,

Thank you for bringing up these edge cases! Here is how I've addressed and evaluated them:

### 1. `IncludeList.include` Mutability
To eliminate the possibility of mutating `.include` (or any other field) on a typed include list after instantiation:
- Made all properties on `IncludeList` (`include`, `where`, `limit`, `offset`, `orderBy`, `orderByList`) `final`.
- Updated the generator so that generated `[Model]IncludeList` and `_[Model]JsonIncludeList` constructors pass these values directly to `super(...)`.
- `IncludeList` instances are now completely immutable after construction, preventing partial JSON includes from being reassigned to full model include lists.

### 2. Generic Code Storing Broad `Include` Types
- `FullModelInclude` is deliberately required on typed query methods (`find`, `findFirstRow`, `findById`) to provide compile-time safety against passing partial column selections into full model queries.
- Generic helpers and abstractions designed for full model queries can type their parameters as `FullModelInclude?`. Code dealing with dynamic or broad `Include` types can narrow them with `if (include is FullModelInclude?)` before dispatching to typed methods.

Silfalion:
Thanks for addressing both points! Making "IncludeList" immutable cleanly closes the remaining gap, and the handling of broad "Include" types makes sense. I really appreciate the work you’ve put into this!

PR #5631, opened 2026-08-24
fix: support flutter_secure_storage 11.0.0 in serverpod_auth_core_flu…

### LINKED ISSUE #5620: Support flutter_secure_storage 11.0.0 in serverpod_auth_core_flutter
`serverpod_auth_core_flutter` pins `flutter_secure_storage: '>=9.2.4 <11.0.0'`. v11 removed the deprecated `encryptedSharedPreferences` option from `AndroidOptions`, which we still pass in `SecureClientAuthSuccessStorage`, so we don't compile against it. That's the only API we use that changed.

Dropping the flag isn't compatible with keeping 9.x in range: on 9.x the flag picks the Android storage backend, so removing it would log out existing users on 9.x resolutions. On 10.x it's a no-op. So the constraint needs to become `'>=10.0.0 <12.0.0'`, not just a ceiling bump.

Session migration on Android: 9 ->10 and 10 ->11 both preserve the stored session. Going 9.x straight to 11.x does not. v11 dropped the pre-v10 migration code, so those users get logged out. Unlikely in practice (the current constraint already resolves to 10.3.1) but needs a release-notes mention.

Adopting v11 also raises app requirements: minSdk 24  and **compileSdk 37** which is *not* the Flutter default yet, even on current stable. Anyone resolving v11 has to set `compileSdk = 37` themselves or the Android build fails. Because of that, a fresh `serverpod create` project resolving v11 wouldn't build out of the box, so we should probably keep the `^10.0.0` override the CLI writes into new projects until Flutter's default catches up

ibrahimMohamed124:
# fix: Support flutter_secure_storage 11.0.0

## Description

Bumps `flutter_secure_storage` constraint from `>=9.2.4 <11.0.0` to
`>=10.0.0 <12.0.0` in `serverpod_auth_core_flutter`, and removes the
deprecated `encryptedSharedPreferences` flag from
`SecureClientAuthSuccessStorage` so the package compiles against v11.

v11 removed `encryptedSharedPreferences` from `AndroidOptions`, which
this package still passed, breaking compilation against v11. Simply
raising the ceiling isn't safe: on 9.x the flag selects the Android
storage backend, so dropping it while still allowing 9.x resolutions
would log out existing users. Raising the floor to `>=10.0.0` avoids
that, since the flag is a no-op from v10 onward.

The `^10.0.0` override the CLI writes into new projects (via
`serverpod_templates`) was kept as-is, since v11 requires
`compileSdk = 37`, which isn't Flutter's default yet — a fresh
`serverpod create` project resolving v11 would fail to build out of
the box.

Fixes #5620

## Pre-launch Checklist

* [x] I read the [[Contribute](https://docs.serverpod.dev/contribute)](https://docs.serverpod.dev/contribute) page and followed the process outlined there for submitting PRs.
* [x] This update contains only one single feature or bug fix and nothing else. (If you are submitting multiple fixes, please make multiple PRs.)
* [x] I read and followed the [[Dart Style Guide](https://dart.dev/guides/language/effective-dart/style)](https://dart.dev/guides/language/effective-dart/style) and formatted the code with [[dart format](https://dart.dev/tools/dart-format)](https://dart.dev/tools/dart-format).
* [x] I listed at least one issue that this PR fixes in the description above.
* [ ] I updated/added relevant documentation (doc comments with `///`), and made sure that the documentation follows the same style as other Serverpod documentation. I checked spelling and grammar.
* [ ] I added new tests to check the change I am making.
* [x] All existing and new tests are passing.
* [x] Any breaking changes are documented below.

No doc comments (`///`) reference `encryptedSharedPreferences` or the
removed option, so nothing needed updating there. No new test was
added because the change removes a deprecated, no-op configuration
flag rather than altering observable behavior — the existing 11
tests in `secure_client_auth_success_storage_test.dart` continue to
exercise `SecureClientAuthSuccessStorage`'s read/write behavior
unchanged, and all pass.

## Breaking changes

Not breaking for typical usage, since `flutter_secure_storage`
resolves automatically within its constraint range. One edge case
worth noting for release notes: a project that upgrades directly
from a 9.x resolution straight to 11.x (skipping 10.x) will lose
existing user sessions on Android, since v11 dropped the pre-v10
session-migration code. Projects that pass through 10.x first are
unaffected.

### DIFF
--- a/modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/analysis_options.yaml
+++ b/modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/analysis_options.yaml
@@ -7,3 +7,12 @@ linter:
 
 formatter:
   trailing_commas: preserve
+analyzer:
+  exclude:
+    - build/**
+    - android/**
+    - ios/**
+    - web/**
+    - windows/**
+    - macos/**
+    - linux/**
--- a/modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/secure_client_auth_success_storage.dart
+++ b/modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/lib/src/storage/secure_client_auth_success_storage.dart
@@ -31,10 +31,6 @@ class FlutterSecureKeyValueStorage implements KeyValueStorage {
 
   static FlutterSecureStorage _defaultSecureStorage() =>
       const FlutterSecureStorage(
-        aOptions: AndroidOptions(
-          // ignore: deprecated_member_use
-          encryptedSharedPreferences: true,
-        ),
         iOptions: IOSOptions(
           accessibility: KeychainAccessibility.first_unlock,
         ),
--- a/modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/pubspec.yaml
+++ b/modules/serverpod_auth/serverpod_auth_core/serverpod_auth_core_flutter/pubspec.yaml
@@ -16,7 +16,7 @@ dependencies:
   clock: ^1.1.1
   flutter:
     sdk: flutter
-  flutter_secure_storage: '>=9.2.4 <11.0.0'
+  flutter_secure_storage: '>=10.0.0 <12.0.0'
   meta: ^1.15.0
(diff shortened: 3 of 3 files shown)


ibrahimMohamed124:

PR #5635, opened 2026-08-24
fix: Fix SQLite missing exception mapping at transaction COMMIT

### LINKED ISSUE #5634: SQLite adapter does not map driver exceptions raised by `COMMIT`
### Describe the bug

`_SqliteDatabaseConnection.transaction()` runs the write transaction without wrapping it in the exception mapping that every other SQLite code path uses. Any driver error raised while committing therefore propagates to the caller as a raw `sqlite3` `SqliteException` instead of a `DatabaseQueryException` subtype.

The Postgres adapter already does this correctly, so the two dialects disagree on the exception type for the same logical failure.

### To reproduce

1. Declare a relation with the new `deferred` keyword:

   ```yaml
   class: Child
   table: child
   database: all
   fields:
     id: UuidValue?, defaultPersist=random_v7
     parent: Parent?, relation(optional, deferred)
   ```

2. On a SQLite transaction, insert a child pointing at a parent id that does not exist:

   ```dart
   try {
     await session.db.transaction((tx) async {
       await Child.db.insertRow(
         session,
         Child(parentId: someIdThatDoesNotExist),
         transaction: tx,
       );
     });
   } catch (e) {
     print(e.runtimeType); // SqliteException
   }
   ```

   The `INSERT` succeeds, because the constraint is deferred. The violation is raised by `COMMIT`.

**Actual**

```console
SqliteException(787): while selecting from statement, FOREIGN KEY constraint failed,
constraint failed (code 787)
  Causing statement: COMMIT
```

**Expected**

```console
DatabaseForeignKeyViolationException: { message: FOREIGN KEY constraint failed, code: 787, ... }
```

The same scenario on Postgres correctly yields a `DatabaseQueryException` subtype, so code that catches `DatabaseException` works on Postgres and breaks on SQLite.

### Expected behavior

All exceptions raised from the database should be wrapped by a `DatabaseException` for centralized (and abstracted) exception treatment.

### Library version

4.0.0-beta.4

### Platform information

N/A

### Additional context

This was perceived while updating the `serverpod_offline_sync` to use the `deferred` relations feature.

The cause comes from `lib/src/adapters/sqlite/database_connection.dart:886` - the only `catch` is for transaction cancellation, so nothing maps an exception thrown by `COMMIT`:

```dart
try {
  _currentTransactionParentZone ??= Zone.current;
  final connection = await _sqliteConnection;
  return await connection.writeTransaction<R>((tx) async {
    // ...
  });
} on _TransactionCancelledException catch (e) {
  return e.result;
} finally {
  _currentTransactionParentZone = null;
}
```

Compare `lib/src/adapters/postgres/database_connection.dart:851`, which does map:

```dart
try {
  return await connection.runTx<R>(/* ... */);
} on pg.ServerException catch (exception, trace) {
  var serverpodException = _queryExceptionFromServerException(exception);
  Error.throwWithStackTrace(serverpodException, trace);
} on pg.PgException catch (exception, trace) {
  var serverpodException = DatabaseQueryException(exception.message);
  Error.throwWithStackTrace(serverpodException, trace);
}
```

- [x] I want to work on this

marcelomendoncasoares:
Fixes: #5634

## Pre-launch Checklist

## Breaking changes

None.

### DIFF
--- a/packages/serverpod/skills/serverpod-models/SKILL.md
+++ b/packages/serverpod/skills/serverpod-models/SKILL.md
@@ -178,7 +178,7 @@ fields:
 
 **Many-to-many:** Use a join table model with two relation fields.
 
-**Referential actions:** `relation(onDelete=Cascade)` and `relation(onUpdate=...)` map to the SQL foreign key actions (`Cascade`, `SetNull`, `SetDefault`, `Restrict`, `NoAction`). `SetNull` requires a nullable foreign key (a nullable object relation field, or `relation(optional)`). Add `deferrable` (or `deferred` for initially deferred) to check the constraint at the end of the transaction instead of per statement.
+**Referential actions:** `relation(onDelete=Cascade)` and `relation(onUpdate=...)` map to the SQL foreign key actions (`Cascade`, `SetNull`, `SetDefault`, `Restrict`, `NoAction`). `SetNull` requires a nullable foreign key (a nullable object relation field, or `relation(optional)`). Add `deferrable` to allow constraint checks to be deferred explicitly, or `deferred` to defer them by default until commit. `Restrict` actions are always checked immediately, even on a deferred constraint; use `NoAction` when the check must be deferrable.
 
 Querying: `include` for eager loading, `includeList` with `where`/`orderBy`/`limit`/`offset` for list relations. `attach`/`detach` for managing relations.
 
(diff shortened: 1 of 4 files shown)


Crazelu (approved):
🎯

Issue #5633, opened 2026-08-24
Automatically generate the list of sync tables when using `database: sync`


marcelomendoncasoares:
As part of the `serverpod_offline_sync` setup, developers need to configure the [server](https://pub.dev/packages/serverpod_offline_sync#3-configure-the-sync-engine-on-the-server) an [client](https://pub.dev/packages/serverpod_offline_sync#4-configure-the-sync-engine-on-the-client) with a `DatabaseSession` wrapper and the list of tables to be synced. This list must be manually kept synchronized between server and client with all the tables that will sync, or a sync operation can fail.

Since both the client and [now also the server](https://github.com/serverpod/serverpod/issues/4544) have a generated class, we can generate the wrapper and tables list by the user. This will reduce the usage of the `serverpod_offline_sync` package to annotating models as `database: sync` and adding the respective dependencies.

PR #5638, opened 2026-08-24
feat: Add the `database: sync` model option as an experimental feature

### LINKED ISSUE #5632: Add the `database: sync` option as experimental feature
Using the `serverpod_offline_sync` package [imposes a few restrictions to the models](https://pub.dev/packages/serverpod_offline_sync#data-modeling-limitations) that are currently evaluated during the initialization of the database, but only because the package was built as an external module. As it becomes part of the official suite of packages, the restrictions must be eagerly evaluated as parte of the `serverpod generate` phase.

Below are the data-modeling limitations that the package enforces to correctly support the merge conflicts:

- Every table with `database: sync` table must:
  - Exist on both the client and the server (same as `database: all`).
  - Have a UUID primary key.
  - Declare the field `scopeId: int?` holding a relation to the `crdt_scopes` table with `onDelete=Cascade`.
- Unique indexes must include `scopeId` together with the target columns.
- Global unique indexes are unsupported, except for FK-only indexes.
- Unique indexes are only supported with at least one `String`/`UuidValue`/nullable column.
- All 1:1 relations must have the foreign-key column nullable (`optional` relation).
- The only allowed non-synced-to-synced relation is `scopeId -> crdt_scopes.id`.
- All foreign-key relations must be `deferred` (`DeferrableConstraint.initiallyDeferred`).

All such restrictions will have to be evaluated for models marked as `database: sync`. Since users can still declare models with `database: all` and satisfy all conditions (as already possible today), the same constraints will continue to be applied on runtime as a runtime safety check.

FXschwartz:
Adds `database: sync` as a new value for the `database` keyword on models, gated behind the new `databaseSync` experimental feature. A sync table generates table code on both the client and the server, like `database: all`, and is additionally validated against the data-modeling restrictions of the `serverpod_offline_sync` package, so violations are reported by `serverpod generate` instead of failing when the sync engine initializes.

Tables with `database: sync` must:
- Have a UUID primary key.
- Declare `scopeId: int?, relation(parent=crdt_scopes, onDelete=Cascade)`.
- Mark every other foreign key as `deferred`.
- Only relate to other `database: sync` tables, and never be referenced from non-synced tables.
- Scope unique indexes with `scopeId` and include at least one releasable column. The only global unique indexes allowed are those composed exclusively of nullable relations to other sync tables.

Additionally, `onDelete=Restrict` is replaced by `NoAction` on the generated table definition of sync tables.

Generating the list of sync tables and the session wrappers from the models is left to #5633.

Closes #5632

## Pre-launch Checklist

## Breaking changes
None


tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -241,11 +243,12 @@ class Restrictions {
...
     if (definition is ModelClassDefinition &&
         definition.isSharedModel &&
-        definition.database != ModelDatabaseDefinition.all) {
+        definition.database != ModelDatabaseDefinition.all &&
+        definition.database != ModelDatabaseDefinition.sync) {
       return [
         SourceSpanSeverityException(
           'The "table" property in shared packages requires the "database" '
-          'property to be set to "all".',
+          'property to be set to "all" or "sync".',
```
marcelomendoncasoares: Suggestion: We need to gate `sync` behind the experimental flag on this error message as well.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -328,9 +347,36 @@ class Restrictions {
...
+      ?validateSyncIdField(definition),
+      ...validateSyncScopeIdField(definition),
+      for (var index in definition.inheritedIndexes)
+        ?validateSyncUniqueIndex(definition, index, parsedModels),
+    ];
+
+    return [
+      for (var message in messages) SourceSpanSeverityException(message, span),
+    ];
+  }
```
marcelomendoncasoares: Suggestion: Since `database: sync` is a deliberate user choice, the errors should happen on the keys that are not compliant with the `sync` requirements instead of on the `database` keyword. All restrictions must follow the current pattern of returning a `SourceSpanSeverityException` that is tied to the analyzed `span`.

marcelomendoncasoares (changes requested):
Only one comment for now, since it will reshape a lot of the introduced restrictions. The main concern is that the errors should be localized at the exact spot that is off on the spec, not on the opt-in keyword that created the restriction. It is the same pattern that we apply for `serverOnly` models.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
FXschwartz: Done! Every sync restriction reports on the offending key, and the tests assert the text for each. The database keyword only carries the experimental-feature gate now.

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
```diff
@@ -549,6 +588,50 @@ class Restrictions {
...
+        ];
+      }
+      return [];
+    }
+
+    if (key == Keyword.deferrable &&
+        relation.deferrable != DeferrableConstraint.initiallyDeferred) {
+      return [SourceSpanSeverityException(syncRelationDeferredError, span)];
+    }
+
```
marcelomendoncasoares: Suggestion: This validation should be split to match the architecture of the YAML validation/parsing. The way it was implemented we're evaluating the value of different keywords together and on the `keyRestriction`.

The mental model comes from the `class_yaml_definition.dart`. There we declare each key with its restrictions and value restrictions.

A key restriction is a rule that prevents the key to be present at all. For example, the `database` keyword requiring the presence of the `table` keyword:

https://github.com/serverpod/serverpod/blob/84229f82da0e911476a3df6b3521bdbf7bd3c4dc/tools/serverpod_cli/lib/src/analyzer/models/yaml_definitions/class_yaml_definition.dart#L57-L61

A value restriction is a rule that prevents a given value on a key. On the flags and options of the `relation` configuration, this is currently just validating that the option values are correct:

https://github.com/serverpod/serverpod/blob/84229f82da0e911476a3df6b3521bdbf7bd3c4dc/tools/serverpod_cli/lib/src/analyzer/models/yaml_definitions/class_yaml_definition.dart#L125-L144

Lastly, if a given key is required and absent, we validate its requirement inside the value restriction of the key that creates the requirement. This is what I done at https://github.com/serverpod/serverpod/pull/5638/commits/e9828b47a4e0607e754de48a609766e6c0aee7db by moving the missing ID check to the `database` value restriction.

So, we need to split these checks somewhat like:

- On the `relation`, if `sync`:
  - For `scopeId`, we validate the cases in which the `onDelete` key is missing.
  - For others, we validate the cases in which `deferred` and `deferrable` are missing (they are mutex).
- On the `onDelete`, if `sync` and `scopeId`, we reject values other than `Cascade`.
- On the `deferrable`, we reject if the model is `sync` and field is not `scopeId` (it is ok for `scopeId`).

All these validations must happen in the `valueRestriction` of the respective keys, replacing and encapsulating the bool/enum validations that already exist today - not on the `keyRestriction` as they are now.

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
```diff
@@ -45,6 +47,58 @@ class ModelDependencyResolver {
...
+    var scopeIdField = SerializableModelFieldDefinition(
+      name: syncScopeIdFieldName,
+      type: TypeDefinition.int.asNullable,
+      scope: ModelFieldScopeDefinition.all,
+      shouldPersist: true,
+      isRequired: false,
+      documentation: [
+        '/// The scope owning this row. Maintained by the sync engine, which',
+        '/// sets it on writes and clears it on reads, so it is null unless',
+        '/// explicitly assigned.',
```
marcelomendoncasoares: Suggestion: Make it a little less verbose.

```suggestion
        '/// The scope owning this row. Maintained by the sync engine.',
```

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
```diff
@@ -45,6 +47,58 @@ class ModelDependencyResolver {
...
+      ],
+      relation: ForeignRelationDefinition(
+        parentTable: syncScopesTableName,
+        foreignFieldName: defaultPrimaryKeyName,
+        onDelete: ForeignKeyAction.cascade,
+      ),
+    );
+
+    classDefinition.fields.add(scopeIdField);
+    _resolveFieldIndexes(scopeIdField, classDefinition);
```
marcelomendoncasoares: Suggestion: Inject as the second field, right below the `id` field, since it is part of the row identity.

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/database_sync_test.dart
```diff
@@ -0,0 +1,796 @@
...
+      'on the id type.',
+      () {
+        var errors = validate([
+          crdtScopeModel,
+          personModel(
+            fields:
+                '  id: int?, defaultPersist=serial\n'
+                '  scopeId: int?, relation(parent=crdt_scopes, onDelete=Cascade)',
+          ),
+        ]);
```
marcelomendoncasoares: Suggestion: Tests on this file are out of the pattern with the other validation tests. There is too much abstraction hiding core parts of the tests, which requires going up and down to understand what each test is validating. I suggest we mirror the same pattern as the surrounding tests, with the models, analyzer and collector explicitly created inside each test.

tools/serverpod_cli/test/database/create_definition/sync_foreign_key_action_definition_test.dart
```diff
@@ -0,0 +1,71 @@
...
+    return ModelClassDefinitionBuilder()
+        .withTableName('example')
+        .withDatabase(database)
+        .withField(field)
+        .build();
+  }
+
+  group(
+    'Given a class definition with "database: sync" and a foreign relation '
+    'with onDelete set to "Restrict" when generating a database definition',
```
marcelomendoncasoares: Suggestion(NIT): Align the clauses to the beginning of the sentences for an easier reading.

```suggestion
    'Given a class definition with "database: sync" and a foreign relation with onDelete set to "Restrict", '
    'when generating a database definition,',
```

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/database_test.dart
```diff
@@ -68,4 +69,93 @@ void main() {
...
+        var config = GeneratorConfigBuilder().withEnabledExperimentalFeatures([
+          ExperimentalFeature.databaseSync,
+        ]).build();
+
+        var collector = CodeGenerationCollector();
+        StatefulAnalyzer(
+          config,
+          syncModels,
+          onErrorsCollector(collector),
+        ).validateAll();
```
marcelomendoncasoares: Suggestion: Experimental flags configured are part of the test scenario, not of the action. Should be a different Given instead.

marcelomendoncasoares (changes requested):
I made a few initial changes to better align the architecture of the YAML validation/parsing, but stopped on the relation actions validation and left an explanation on the direction we need to change. Also a few comments on the tests to reduce the abstraction and better separate the scenarios.

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
```diff
@@ -45,6 +47,58 @@ class ModelDependencyResolver {
...
+  }
+
+  /// Injects the `scopeId` field on tables with `database: sync` that do not
+  /// declare it, following the same convention as implicit foreign keys: a
+  /// declared field is validated as is, an omitted one is generated.
+  ///
+  /// The field is a system field maintained by the sync engine, which stamps
+  /// it on writes and strips it on reads, so applications never need to
+  /// declare it. It is skipped when the scopes table is unknown, which the
+  /// validation layer reports as a missing module.
```
FXschwartz: Trimmed both comments down and moved the injection to right under id. Added a quick assertion on the field order too.

6d6542cf4

tools/serverpod_cli/lib/src/analyzer/models/validation/restrictions.dart
FXschwartz: Makes sense and I moved everything over. Relation now only checks that the required keys are there, and onDelete / deferred / deferrable each validate their own value. While I was in there I also lined it up with 0.0.3, so only non-nullable FKs need deferred now and scopeId can be deferred if someone really wants to.

Done in - 4b670aa31

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/database_sync_test.dart
FXschwartz: Flattened them so each test builds its own models, analyzer and collector like the rest of the folder. I kept the span assertions since that's what shows where each error lands. 90992534c

tools/serverpod_cli/test/analyzer/models/stateful_analyzer/model_validation/database_test.dart
FXschwartz: Good call! I split it into three tests with the flag config in the Given (disabled / enabled / all features) instead of inside the when. 90992534c

tools/serverpod_cli/test/database/create_definition/sync_foreign_key_action_definition_test.dart
FXschwartz: Done, one clause per line now. 90992534c

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
FXschwartz: Done - 6d6542cf4

tools/serverpod_cli/lib/src/analyzer/models/entity_dependency_resolver.dart
FXschwartz: Done - 6d6542cf4

marcelomendoncasoares (approved):
LGTM! 🚀🚀

PR #5643, opened 2026-08-25
fix: Allow cross-origin WebSocket upgrades under relic 2.0.0-rc.1

FXschwartz:
relic 2.0.0-rc.1 (published 2026-08-25) rejects cross-origin WebSocket upgrades with 403 by default. We depend on `^2.0.0-beta.1` with no lockfile, so every CI run since then resolves rc.1 and all E2E web shards fail with
`WebSocketException: Failed to connect WebSocket`: the test page is served from `localhost` while the socket goes to `serverpod_test_server`. The `vm` shards pass because dart:io clients send no `Origin` header.

This is not a test-only problem. Any web client served from a different host than the API server (e.g. `app.example.com` → `api.example.com`) would lose method streaming as soon as it picks up rc.1.

Serverpod already gates the handshake with `allowedOrigins`, `Server._isOriginDisallowed`, and that allow-list is required whenever cookie auth is enabled. This PR:
- Passes `allowAnyOrigin: true` to `WebSocketUpgrade` so Serverpod's own origin policy stays the one that decides. Behaviour is unchanged from before rc.1: with `allowedOrigins` configured only listed origins may connect; with it unset any origin may.
- Raises the relic floor to `^2.0.0-rc.1` (the `downgrade` analyzer jobs resolved beta.1, which has no `allowAnyOrigin` parameter).

Not in this PR, worth deciding separately: whether an unset `allowedOrigins` should default to same-host only (relic's new default) rather than any origin.

## Pre-launch Checklist

## Breaking changes

### DIFF
--- a/.github/workflows/dart-tests.yaml
+++ b/.github/workflows/dart-tests.yaml
@@ -842,7 +842,12 @@ jobs:
       - name: Alias serverpod_test_server to localhost
         run: echo "127.0.0.1 serverpod_test_server" | sudo tee -a /etc/hosts
       # Firefox for the browser run; xvfb provides the virtual display (xvfb-run).
+      # Pinned: Firefox 154.0.1 extended Local Network Access permission prompts to
+      # WebSocket connections from HTTP pages, which blocks ws://serverpod_test_server
+      # under xvfb (nobody can click "Allow"). See the tracking issue linked in the PR.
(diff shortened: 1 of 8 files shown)


marcelomendoncasoares (approved):
LGTM! 🚀

marcelomendoncasoares (approved):

PR #5644, opened 2026-08-25
feat: Require the .spy.yaml extension for model files

### LINKED ISSUE #5537: Require .spy.yaml for model files and drop bare .yaml/.yml model support
Follow-up from the discussion on #5518 

Model files with bare `.yaml/.yml` extensions are still supported when placed under `lib/src/models` or `lib/src/protocol` (hardcoded path parts in `ModelLoadConfig`, checked via substring in `ModelHelper.isModelFile`). The recommended format is `.spy.yaml`, all templates generate `.spy.yaml`, and no package in this repo uses bare `yaml` models anymore.

Supporting both formats forces heuristics to tell model files apart from other `yaml`, with inconsistent results:

- An `analysis_options.yaml` placed inside `lib/src/models` passes `isModelFile` and gets parsed as a model.
- The `serverpod start` file watcher only watches `.spy` extensions, so bare `yaml` models don't trigger regeneration there, while `generate --watch` handles them.
- The language server has to watch `yaml` files in the model directories to keep diagnostics fresh for a format we no longer recommend.

Since 4.0 is a major bump, it's a good time to make the behavioral breaking change and require `.spy.yaml/.spy.yml/.spy`.

Scope
- [ ] Remove `yamlModelFileExtensions` from `ModelHelper.isModelFile` (affects generate, generate --watch, and the language server)
- [ ] Remove the `**/models/**/*.yaml / **/protocol/**/*.yaml` patterns from the VS Code extension's `documentSelector` (and the `onLanguage:yaml` activation event)
- [ ] Simplify the language server file watchers to spy extensions only
- [ ] Update docs (concepts/models) and add an upgrade note for 3.x → 4.0 (rename .yaml models to .spy.yaml)

FXschwartz:
Serverpod 4.0 requires the `.spy.yaml` extension for model files. Bare `.yaml` / `.yml` files in `lib/src/models` and `lib/src/protocol`, are no longer treated as models.

This removes the directory heuristics from model discovery, the extra bare-yaml watcher in the language server, and the `onLanguage:yaml` activation of the VS Code extension. It also makes `serverpod start` consistent with `generate`: its file watcher was already `.spy`-only, so bare yaml models never hot-regenerated.

To keep the upgrade from silently dropping models, `generate`, `create-migration` and `start` scan the legacy directories of the project, its shared packages and its dependent modules for bare yaml model files and exit with an error listing them:

```console
ERROR: Model files must use the .spy.yaml extension. The following files are
ignored:
  lib/src/models/user.yaml
Rename the files to use the .spy.yaml extension and run the command again.
```

Only files whose top-level yaml has a `class`, `enum` or `exception` key are reported, so unrelated yaml in those directories is ignored.

Closes #5537 

## After Merge
- serverpod_docs: update the models page wording and add an upgrade-guide step (drafted).
- Publish VS Code extension 1.5.0.

## Pre-launch Checklist

## Breaking changes
- Model files must use the `.spy.yaml`, `.spy.yml` or `.spy` extension. Bare `.yaml` / `.yml` files in `lib/src/models` and `lib/src/protocol` the CLI reports them as an error and asks for them to be renamed.
- The VS Code extension only attaches the language server to `.spy.yaml`, `.spy.yml` and `.spy` files and no longer activates on other yaml files.

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/generate.dart
+++ b/tools/serverpod_cli/lib/src/commands/generate.dart
@@ -16,6 +16,7 @@ import 'package:serverpod_cli/src/generator/generation_staleness.dart';
 import 'package:serverpod_cli/src/runner/serverpod_command.dart';
 import 'package:serverpod_cli/src/runner/serverpod_command_runner.dart';
 import 'package:serverpod_cli/src/serverpod_packages_version_check/serverpod_packages_version_check.dart';
+import 'package:serverpod_cli/src/util/legacy_model_files.dart';
 import 'package:serverpod_cli/src/util/pubspec_lock_parser.dart';
 import 'package:serverpod_cli/src/util/pubspec_plus.dart';
 import 'package:serverpod_cli/src/util/serverpod_cli_logger.dart';
@@ -89,6 +90,10 @@ class GenerateCommand extends ServerpodCommand<GenerateOption> {
     } catch (e) {
       log.error('$e');
       throw ExitException(ServerpodCommand.commandInvokedCannotExecute);
+    }
+
+    if (await LegacyModelFiles.report(config)) {
+      throw ExitException.error();
     }
 
     var serverPubspecFile = File(
--- a/tools/serverpod_cli/lib/src/util/legacy_model_files.dart
+++ b/tools/serverpod_cli/lib/src/util/legacy_model_files.dart
@@ -128,7 +128,8 @@ abstract final class LegacyModelFiles {
   }
 
   /// Whether [file] has a bare yaml extension and a top-level model key.
-  /// Files that do not parse as yaml maps are not models, and are ignored.
+  /// Files that cannot be read or parsed as yaml maps are not models, and
+  /// are ignored.
   static bool _isLegacyModelFile(File file) {
     var path = file.path;
     if (ModelHelper.isModelFile(path)) return false;
@@ -138,6 +139,8 @@ abstract final class LegacyModelFiles {
       var document = loadYaml(file.readAsStringSync());
       return document is YamlMap &&
           _modelKeys.any((key) => document.containsKey(key));
+    } on FileSystemException {
+      return false;
     } on YamlException {
       return false;
     }
--- a/tools/serverpod_cli/lib/src/util/legacy_model_files.dart
+++ b/tools/serverpod_cli/lib/src/util/legacy_model_files.dart
@@ -1,0 +1,145 @@
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+import 'package:serverpod_cli/src/config/config.dart';
+import 'package:serverpod_cli/src/util/model_helper.dart';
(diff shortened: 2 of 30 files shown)


tools/serverpod_cli/lib/src/util/legacy_model_files.dart
```diff
@@ -0,0 +1,145 @@
...
+  static bool _isLegacyModelFile(File file) {
+    var path = file.path;
+    if (ModelHelper.isModelFile(path)) return false;
+    if (!_legacyExtensions.any(path.endsWith)) return false;
+
+    try {
+      var document = loadYaml(file.readAsStringSync());
+      return document is YamlMap &&
+          _modelKeys.any((key) => document.containsKey(key));
+    } on YamlException {
```
nielsenko, line 144: What about files that are not utf8, or unreadable? Should we not return `false` on such as well?

tools/serverpod_cli/lib/src/commands/generate.dart
```diff
@@ -91,6 +92,10 @@ class GenerateCommand extends ServerpodCommand<GenerateOption> {
       throw ExitException(ServerpodCommand.commandInvokedCannotExecute);
     }
 
+    if (await LegacyModelFiles.report(config)) {
```
nielsenko, line 95: So a single legacy file will abort generate completely? 
What if it belong to an imported module that you cannot modify?

nielsenko: Also this runs before `isGenerationUpToDate` staleness stamp check.  Seems a bit expensive and redundant.

nielsenko (commented):
Some questions..

tools/serverpod_cli/lib/src/commands/generate.dart
nielsenko: The point about imported modules is perhaps less of a concern given that we pin versions explicitly. Module authors will have to bump as well.

The run too often/early is still an issue I think

Crazelu:
It seems to me that large server projects without legacy models will always pay the cost of the scan on `create-migration`, `generate` and `start`. Worse if they have many shared models without legacy models.

Maybe this can be benchmarked? Perhaps that could inform the decision to move this legacy model reporting into the generator after it has collected analysis results (since it has to do the scan anyway).

tools/serverpod_cli/lib/src/commands/generate.dart
FXschwartz: Discussed during standup the increased time for this is in the sub 10 ms for small and large projects. Having it run before the staleness check also prevents missed .yaml file additions in the `generate/create` commands.

tools/serverpod_cli/lib/src/util/legacy_model_files.dart
FXschwartz: Yes this is a bug. Fixed in d31680397

nielsenko (approved):
LGTM 🥇

PR #5645, opened 2026-08-26
docs: Add runner.md design to pin down attach

nielsenko:
For discussing design of `serverpod attach`. This PR will not be merged in isolation, but combined with implementation if accepted.

### DIFF
--- a/docs/design/runner.md
+++ b/docs/design/runner.md
@@ -1,0 +1,501 @@
+# Design: The Serverpod runner
+
+`serverpod start` runs the terminal UI, the MCP server, the file watcher, the
+Frontend Server, and the pod in a single process. The terminal UI binds a pty,
+an agent working in a worktree has none. It can use the `--no-tui` flag, but 
+then a human can no longer participate.
+
+This document separates the long-lived part of `serverpod start` from the
+terminal UI, so a UI can be attached later, attached several times, or not at
+all. 
+
+>[!NOTE]
+>[`mcp_server.md`](mcp_server.md) already calls the `serverpod start`
+process the *runner*, and this document narrows the term to exclude the UI.
+
+## Current state
+
+`StartCommand` (`tools/serverpod_cli/lib/src/commands/start.dart`) owns Docker
+Compose, code generation, the `KernelCompiler` (and its resident Frontend
+Server), the `ServerProcess`, the `FileWatcher`, the `WatchSession` reload loop,
+Flutter app spawning, the MCP socket, the vm-service proxy, and the nocterm
+terminal UI. 
+
+Two consumers already attach from outside without a terminal.
+
+- An IDE, over the vm-service proxy. The CLI mounts a proxy in front of the
+  pod's VM service, publishes the proxy URI to `vm-service-info.json`, and the
+  IDE connects over TCP for debugging.
+- An agent, over the Unix socket at
+  `<serverDir>/.dart_tool/serverpod/mcp.sock`, where `serverDir` is the server
+  package, the one that depends on `serverpod`. It connects through the
+  `serverpod mcp-server` bridge. [`mcp_server.md`](mcp_server.md) describes its reconnect
+  and lifetime semantics.
+
+The terminal UI is the only consumer that must run inside the runner process.
+
+>[!NOTE]
+>[`hot_reload.md`](hot_reload.md) describes the reload pipeline and
+[`tui.md`](tui.md) the UI.
+
+
+## Goals
+
+- Start the development stack without a terminal, and attach a UI later.
+- Make `serverpod start` idempotent, so a caller need not know whether the stack
+  is already running in this worktree.
+- Run several runners concurrently in different worktrees of the same codebase.
+
+Out of scope is sharing a database or Docker Compose stack between worktrees.
+Each worktree provides its own, through embedded Postgres or its own compose
+project. The reload pipeline, the logging protocol, and the visual design of the
+UI are unchanged.
+
+Fixed ports in `config/development.yaml` collide when several worktrees run at
+once, so the concurrency goal depends on [Address
+publication](#address-publication). That change is largely independent of the
+split, but it lands after it, since the manifest it publishes into has to exist
+first.
+
+
+## Overview
+
+Three commands now augment `start`, which keeps its name and its meaning, but
+become idempotent.
+
+```shell
+serverpod start   # Ensure the runner is up, then attach a UI
+serverpod attach  # Attach to an already-running runner
+serverpod stop    # Shut the runner down
+serverpod status  # Print the runner's state and addresses
+```
+
+`serverpod start` now spawns a runner if none is running (for the server package),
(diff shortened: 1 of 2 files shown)


docs/design/runner.md
```diff
@@ -0,0 +1,490 @@
...
+
+Three commands now augment `start`, which keeps its name and its meaning, but
+become idempotent.
+
+```shell
+serverpod start   # Ensure the runner is up, then attach a UI
+serverpod attach  # Attach to an already-running runner
+serverpod stop    # Shut the runner down
+serverpod status  # Print the runner's state and addresses
+```
```
marcelomendoncasoares, line 71: Suggestion: Instead of placing all these commands at the root, we could group them as `serverpod runtime` to be clear on what each command is acting - since we already have `serverpod database` that has `start` and later can receive `status` / `stop`. The `serverpod start` would remain as the idempotent command that can either start + attach (if not running) or just attach directly.

docs/design/runner.md
```diff
@@ -0,0 +1,490 @@
...
+its address, and returns, which is the path an agent takes.
+
+`--tui` / `--no-tui` keeps its current meaning and selects the renderer. The
+terminal UI is used when `--tui` holds and `stdout.hasTerminal`, and a
+plain-text log stream in the foreground otherwise. `serverpod attach` takes the
+same pair.
+
+| Invocation | Result |
+|------------|--------|
+| `serverpod start` | Runner up, terminal UI if a pty is available |
```
marcelomendoncasoares, line 96: Comment: Agree on gating the `--tui` only if PTY is available. It allows a CI or automation shells to invoke plain `serverpod start` with no issues.

docs/design/runner.md
```diff
@@ -0,0 +1,490 @@
...
+### Lifetime
+
+Detaching a UI does not stop the runner, whoever started it. The stack is
+stopped with `serverpod stop`, or with `⇧+Q` in the UI, next to the `Q` that
+only detaches.
+
+>[!NOTE]
+>The rejected alternative is to shut down on detach when the detaching invocation
+spawned the runner. That keeps Ctrl+C closer to its current behaviour, but the
+same key then stops the server or does not, depending on how the session began.
```
marcelomendoncasoares, line 121: Comment: Good UX on the Shift+Q to stop, Q to detach only. I agree that the consistent behavior of the commands is better than stopping if it was the one starting it.

docs/design/runner.md
```diff
@@ -0,0 +1,490 @@
...
+
+The lock is an advisory lock on an open file descriptor, through
+`RandomAccessFile.lock`. The kernel releases it when the process dies, so a
+crashed runner leaves nothing to clean up, unlike the manifest and the socket
+file.
+
+A detached runner survives `dart pub global activate serverpod_cli`, so a new
+client can meet an old runner. A client refuses to attach when `protocolVersion`
+differs and prints the stop-and-restart instruction. A differing `cliVersion` at
+equal `protocolVersion` is a warning.
```
marcelomendoncasoares, line 202: Question: Should trying to attach to an old runner expose the possibility to kill and replace it?

docs/design/runner.md
```diff
@@ -0,0 +1,490 @@
...
+Two consequences elsewhere.
+
+- `ServerConfig` in `serverpod_shared/lib/src/config.dart` distinguishes the
+  bind `port` from the advertised `publicHost`, `publicPort`, and
+  `publicScheme`, and the server builds client-facing URLs from the latter.
+  `publicPort` has to be derived from the resolved bind port, or the server
+  advertises 8080 while listening elsewhere.
+- Flutter apps the runner launches receive the resolved URL through
+  `--dart-define`. Apps started by hand, browser tabs, and `curl` read it from
+  the manifest, which `serverpod status` prints.
```
marcelomendoncasoares, line 325: Question: This is the separate part that we need to address to allow concurrent worktrees, right? If so, I would suggest moving this to a separate file to avoid mixing the contexts (although we can reference it in this one).

docs/design/runner.md
```diff
@@ -0,0 +1,490 @@
...
+  path does today.
+- In an attached client, SIGINT detaches and cannot reach the pod.
+- Ordinary shutdown goes over the protocol, from `serverpod stop`.
+
+### Flutter apps
+
+`--flutter` currently launches the apps marked `auto_launch: true` when the
+stack starts. Auto-launch moves to UI attachment, so the apps launch when a UI
+first attaches in a runner lifetime. A later attach does not relaunch them, so
+an app the developer stopped stays stopped across a detach.
```
marcelomendoncasoares, line 401: Question: Why default to launching the Flutter app only when attaching? Since the server will be running, why can't the app be running as well?

docs/design/runner.md
```diff
@@ -0,0 +1,490 @@
...
+### `serverpod`
+
+- Resolve the bind ports dynamically when the configured ports are unavailable,
+  and derive `publicPort` from the resolved value.
+
+### Repository
+
+- `DART_VERSION` raised to `^3.11.0`, with the checked-in pubspecs regenerated
+  from `templates/pubspecs`. `FLUTTER_VERSION` and the floor entries of the CI
+  version matrices move to the lowest Flutter release bundling Dart 3.11.
```
marcelomendoncasoares, line 468: Comment: Already solved after merging the workspaces PR.

docs/design/runner.md
nielsenko: It can. It won't launch the app if already running - exactly so you can detach and attach again without changing the state. To match current behavior somewhat I think the first attach should try to launch the app (if `auto_launch: true`), but if it is stopped, you detach, and re-attach then it shouldn't start it again.

docs/design/runner.md
nielsenko: Yes, but it seems a bit too small to have its own file.

docs/design/runner.md
nielsenko: If attach succeeds,  you can stop the runner like "normal" (actually Shift-Q, as Q will just detach)

docs/design/runner.md
nielsenko: I agree that `serverpod runner` with subcommand `start`, `stop`, `attach`, `status` makes sense, but this would make `serverpod start` become `serverpod runner start` - isn't it a bit late in the game for that (with docs, and marketing material under way)?

nielsenko: Perhaps we could still have `serverpod start` that basically does `serverpod runner start && serverpod runner attach` for you.

marcelomendoncasoares (approved):
Giving a thumbs up in advance, since we have already agreed upon the discussed points. If there are no other comments, feel free to merge after updating the docs.

Crazelu (approved):

docs/design/runner.md
nielsenko: This will stay for now, as it is part of the goals. If it proves big enough to warrant its own design it can move out later.

PR #5646, opened 2026-08-26
fix: Run serverpod create without TUI in default mode when stdin/stdout is not a TTY

### LINKED ISSUE #5639: serverpod create exits 1 with no output when stdin is not a TTY
`serverpod create <name>` fails when stdin is not a TTY. It prints nothing and creates nothing.

### Symptom

```
$ serverpod create bareproj >out.txt 2>err.txt
EXIT=1
stdout bytes: 0  stderr bytes: 0
```

Zero bytes on **both** stdout and stderr. No project directory appears. Exit code is 1.

With `-v` the command prints only the Dart and Flutter version banner, then stops. Still no project.

### Steps to reproduce

1. Change to an empty directory.
2. Run `serverpod create bareproj` from a shell with no TTY (CI, a script, a non-interactive agent shell).

### Expected

Either the project is scaffolded, or an error names the cause. For example:

```
Interactive prompts are disabled (no TTY). Pass --template, --database and --auth explicitly.
```

### Workaround

Passing every choice explicitly works:

```
serverpod create myproject --no-interactive --template server --database --no-redis --no-auth --ide none
```

That prints `✅ Serverpod project created.` and scaffolds normally.

### Impact

Scaffolding cannot be automated without knowing in advance that every prompt must be pre-answered. A silent exit 1 gives no signal about what went wrong, so the failure is hard to diagnose from CI logs.

### Environment

- Serverpod CLI 4.0.0-beta.3
- Dart SDK 3.13.1 (stable), macOS arm64
- Flutter 3.47.1

Crazelu:
`serverpod create` and `serverpod quickstart` gated TUI use with CI check and the optional `interactive` flag (treated as enabled by default). When executed with stdin and/or stdout that is not a tty, the TUI is silently broken and project creation doesn't happen. This PR extends the TUI readiness check by probing stdin and stdout for tty.

Closes #5639 

## Pre-launch Checklist

## Breaking changes

None

### DIFF
--- a/tools/serverpod_cli/lib/src/commands/create/tui/runner.dart
+++ b/tools/serverpod_cli/lib/src/commands/create/tui/runner.dart
@@ -1,5 +1,6 @@
 import 'dart:io';
 
+import 'package:ci/ci.dart' as ci;
 import 'package:cli_tools/cli_tools.dart';
 import 'package:path/path.dart' as p;
 import 'package:serverpod_cli/src/commands/create/tui/app.dart';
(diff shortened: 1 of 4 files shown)


tools/serverpod_cli/lib/src/commands/create/tui/runner.dart
```diff
@@ -105,6 +106,46 @@ Future<CreateConfigStateResult> getCreateConfigState({
...
+/// Starting the TUI captures stdin's echo and line modes so it can restore them
+/// on exit, and paints frames to stdout. When stdin cannot report those modes
+/// the capture throws before anything is rendered, and because the TUI logger
+/// is already installed by then the failure never reaches the user.
+///
+/// `stdin.hasTerminal` does not answer this: Dart reports every character
+/// device as a terminal, so `< /dev/null` passes that check and then throws
+/// anyway. Reading the mode is the only reliable probe.
+bool shouldUseCreateTui(bool? interactive) {
+  if (interactive == false) return false;
```
nielsenko, line 121: I prefer using `??` to resolve `null`. 
```suggestion
  if (interactive ?? true) return false;
```
or even better make `interactive` parameter mandatory.

tools/serverpod_cli/lib/src/commands/create/tui/runner.dart
```diff
@@ -105,6 +106,46 @@ Future<CreateConfigStateResult> getCreateConfigState({
...
+/// treats every error as "unsupported".
+bool get _stdinSupportsTerminalModes {
+  try {
+    stdin.echoMode;
+    return true;
+  } catch (_) {
+    return false;
+  }
+}
+
```
nielsenko, line 148: Nice defense in depth 👍 .. While it would be exceedingly odd for a user to _pipe_ anything into `serverpod create` we might as well guard against it.

nielsenko (approved):
A nit, but you choose.

LGTM 👍

tools/serverpod_cli/lib/src/commands/create/tui/runner.dart
Crazelu: This will change the behaviour when `interactive` is null (which is its default). By default, we want to show the TUI but only if it is possible.

nielsenko: You could resolve interactive to a non-nullable from the outset when you read it from cli args. 

The suggestion does not change behaviour - semantics is identical.

PR #5647, opened 2026-08-27
fix: Print correct version on serverpod upgrade, and prevent accidental downgrades

### LINKED ISSUE #5500: `serverpod upgrade` outputs pre-upgrade version upon completion
### Describe the bug

When running `serverpod upgrade`, the CLI successfully executes `dart install serverpod_cli` to download and install the latest package. However, the final completion message displays:

Serverpod is up to date: <INITIAL_VERSION> version.

### Root Cause
In `tools/serverpod_cli/lib/src/commands/upgrade.dart`, the final log message references `$templateVersion`:
```dart
log.info('Serverpod is up to date: $templateVersion version.');


### To reproduce

1. Install a previous version of Serverpod CLI (or run `serverpod upgrade` when a newer version is published).
2. Execute `serverpod upgrade` in the terminal.
3. Wait for the `Updating Serverpod Cli...` progress indicator and `dart install serverpod_cli` to complete.
4. Observe the final output message: "Serverpod is up to date: <PRE_UPGRADE_VERSION> version."


### Expected behavior

The CLI should report the actual updated version, e.g.:
Serverpod CLI updated to version <LATEST_VERSION>.

### Library version

serverpod_cli 4.0.0-beta.0 and Dart 3.12.2 (stable)

### Platform information

All supported platforms (Linux, macOS, Windows)


### Additional context

_No response_

None

- [ ] I want to work on this

### PR DESCRIPTION
**Pull request by nielsenko**
Fixes #5500.

`serverpod upgrade` printed the version compiled into the binary already running, so by definition the pre-upgrade one. The target is now resolved from pub.dev.

Some related issues was fixed in this PR as well:

- A pre-release install was silently downgraded to the latest stable, since a bare `dart install` never selects a pre-release. Moving backwards now needs an explicit `--version`.
- The output was identical whether or not anything was installed.
- The update prompt now says `serverpod upgrade`, not `dart install serverpod_cli` (that would downgrade a beta).
- Pub cache fallback no longer resolves to `/.pub-cache/bin` when `HOME` is unset.

### CODE REVIEW

File: tools/serverpod_cli/lib/src/commands/upgrade/cli_installation.dart [hunk 1]
```diff
@@ -0,0 +1,137 @@
...
+    }
+  }
+
+  return null;
+}
+
+/// Whether both paths point at the same executable, ignoring the platform
+/// extension that distinguishes a binstub from the binary it wraps.
+bool isSameExecutable(final String a, final String b) =>
+    p.equals(p.dirname(a), p.dirname(b)) &&
```
FXschwartz, line 173: I can validate this on my Windows machine tomorrow but..

I think this could come out as `foreign` for `dart install` users on Windows. From what I can see in dartdev, the bin entry there is a `.bat` wrapper rather than a symlink, so `Platform.resolvedExecutable` would be the exe in app-bundles while `_canonicalize` leaves the `.bat` in bin, and the dirname check wouldn't match.

If that's right, a normal install gets the "dart install does not manage" error and the pub global deactivate hint.

tools/serverpod_cli/lib/src/commands/upgrade.dart
```diff
@@ -1,40 +1,316 @@
...
-    if (!success) {
-      log.info('Failed to update Serverpod.');
+  /// Installs without knowing the resulting version, all that is possible
+  /// when pub.dev cannot be reached.
+  Future<void> _installWithoutVersionLookup({
+    required final CliInstallation installation,
+    required final UpgradeChannel channel,
+    required final String? requested,
+  }) async {
+    if (requested == null && channel == UpgradeChannel.any) {
```
FXschwartz: nit edgecase: this only refuses on `any`, so with a prerelease installed and `--channel stable`, a failed lookup would dowgrade and fall through to a bare `dart install serverpod_cli`.

tools/serverpod_cli/lib/src/commands/upgrade/cli_installation.dart
```diff
@@ -0,0 +1,137 @@
...
+/// Both paths must be canonical, as produced by [CliInstallation.resolve].
+CliInstallationKind classifyInstallation({
+  required final String runningExecutable,
+  required final String managedExecutable,
+}) {
+  if (!p.equals(
+    p.basenameWithoutExtension(runningExecutable),
+    p.basenameWithoutExtension(managedExecutable),
+  )) {
+    return CliInstallationKind.source;
```
FXschwartz: I think the `dart pub global activate` case may land here rather than in `foreign`. 

My own `~/.pub-cache/bin/serverpod` is a shell script that runs `dart pub global run serverpod_cli:serverpod_cli`, so I'd expect `Platform.resolvedExecutable` to be the SDK's `dart` and this to return `source` before `isSameExecutable` runs.

If so, those users would see "running from a source checkout (.../dart-sdk/bin/dart)" and the deactivate hint wouldn't print.

tools/serverpod_cli/lib/src/commands/upgrade/upgrade_target.dart
```diff
@@ -0,0 +1,70 @@
...
+/// The newest version allowed by [requested], or by [channel] when nothing was
+/// requested, wins.
+UpgradeTarget? resolveUpgradeTarget({
+  required final Version current,
+  required final Iterable<Version> published,
+  required final UpgradeChannel channel,
+  final VersionConstraint? requested,
+}) {
+  final candidates = published.where((final version) {
+    if (requested != null) return requested.allows(version);
```
FXschwartz: I tried this with `pub_semver` and `any` and `>=3.4.0` both allow `4.0.0-rc.1` (`^3.4.0` doesn't). So I think `--version '>=3.4.0'` from a stable install would pick the rc, and `requested != null` makes it an `install`. 

pub itself seems to only pick a prerelease when the constraint names one, so maybe keep the channel filter on top unless `requested` is an exact version?

Not sure how common that flag shape is in practice.

tools/serverpod_cli/lib/src/commands/upgrade.dart
```diff
@@ -1,40 +1,316 @@
...
+        'installation with the latest stable release, which this channel does '
+        'not track. Re-run with --version to pick a version explicitly.',
+      );
       throw ExitException.error();
     }
 
-    log.info('Serverpod is up to date: $templateVersion version.');
+    final descriptor = requested == null
+        ? _packageName
+        : '$_packageName@$requested';
```
FXschwartz: It looks like `dart install` parses what comes after `@` as YAML, and when I tried `loadYaml('>=4.0.0 <5.0.0')` it threw. So a range that `_parseConstraint` accepted may fail here with "Could not parse (what comes after @)".

tools/serverpod_cli/lib/src/update_prompt/prompt_to_update.dart
FXschwartz: nite: Not new here, but now that the prompt says `serverpod upgrade`: it still checks stable only, while the command follows prereleases when one is installed.

I think someone on an rc could be told 4.0.0 is out and then get the newest beta instead.

tools/serverpod_cli/lib/src/util/dart_install.dart
```diff
@@ -0,0 +1,43 @@
...
+/// [message], leaving the reporting of both outcomes to the caller.
+Future<DartInstallResult> runDartInstall(
+  final String descriptor, {
+  required final String message,
+}) async {
+  final errorOutput = StringBuffer();
+
+  final success = await log.progress(message, () async {
+    log.debug('Running `dart install $descriptor`...');
+    final process = await Process.start(dartExecutablePath, [
```
FXschwartz, line 51: Question: Was switching from PATH `dart` to `dartExecutablePath` deliberate here? 

If I'm following it right, an AOT install runs the temp-script probe, and if `DART_SDK` is stale or the probe fails it throws out of `log.progress` and shows up as an internal error rather than the "Failed to install" message.

If it's for shims, a try/catch around it might keep the failure friendly.

FXschwartz (changes requested):

tools/serverpod_cli/lib/src/commands/upgrade/cli_installation.dart
nielsenko: Hmm.. I was about to say no-way as `getSdkPath` is exercised on CI on windows all the time.

But it seems flutter installs dart as a `dart.bat` wrapper where the dart github runner installs a bare `dart.exe`, and you need a shell to run a `.bat` on windows. I'm still struggling a bit to understand why this has never been reported 🤯 

Will update to use `runInShell` when it probing for dart on windows.

tools/serverpod_cli/lib/src/commands/upgrade.dart
nielsenko: Failed lookups will now fail the upgrade. No offline trickery.

tools/serverpod_cli/lib/src/commands/upgrade/upgrade_target.dart
nielsenko: I'll disallow anything but passing exact versions to `--version` flag.

Will also stop trying to be smart when offline. Going forward contacting pub will be a hard requirement for `serverpod upgrade`.

tools/serverpod_cli/lib/src/commands/upgrade.dart
nielsenko: Will disallow constraint. See above

tools/serverpod_cli/lib/src/update_prompt/prompt_to_update.dart
nielsenko: Will fix

tools/serverpod_cli/lib/src/util/dart_install.dart
nielsenko: `dartExecutablePath` is the most robust way we have of finding `dart[.exe]`. Shims being one reason to use it.

But I will harden `getSdkPath` more (used by `dartExecutablePath`). On windows we will use `runInShell` to handle the `dart.bat` case (used by flutter, not dart - this still baffles me). 

Also, I will check if `DART_PATH` actually points to a valid install.

This should prevent the probe from throwing.

tools/serverpod_cli/lib/src/commands/upgrade/cli_installation.dart
nielsenko: Will fix

nielsenko:
@FXschwartz All comments addressed. Could you take a second look.

FXschwartz (approved):
LGTM!

PR #5652, opened 2026-08-28
feat: Generate the sync tables list and session wrappers for `database: sync` models

### LINKED ISSUE #5633: Automatically generate the list of sync tables when using `database: sync`

FXschwartz:
Generates the code that the `serverpod_offline_sync` README asks users to write by hand when the `databaseSync` experimental feature is enabled and the `serverpod_offline_sync` module is listed in `generator.yaml`. Using the package now comes down to adding the dependencies and declaring models with `database: sync`.

Requires `serverpod_offline_sync` 0.0.3. The test tools are left unchanged, since `withServerpod` constructs the framework `Serverpod` and keeps taking `databaseInterceptor` explicitly.

Verified on a fresh project against `serverpod_offline_sync` 0.0.3: both packages analyze clean, migrations apply, and a session created through the generated `Serverpod` yields a `CrdtDatabase` with the expected sync tables.

Closes #5633

## Pre-launch Checklist

## Breaking changes
None


tools/serverpod_cli/lib/src/analyzer/protocol_definition.dart
```diff
@@ -13,10 +13,15 @@ class ProtocolDefinition {
...
+  /// The aliases of the module dependencies that have `database: sync` models,
+  /// which are not part of [models].
+  final Set<String> moduleAliasesWithSyncTables;
+
   /// Create a new [ProtocolDefinition].
   const ProtocolDefinition({
     required this.endpoints,
     required this.models,
     required this.futureCalls,
+    this.moduleAliasesWithSyncTables = const {},
```
marcelomendoncasoares: Question: Can we avoid leaking sync-related definitions on the general `ProtocolDefinition` interface? I get that this is a workaround to module models not being available while analyzing a project. But can we provide this list otherwise?

FXschwartz: Updated as discussed internally! Replaced the ProtocolDefinition field with manifest discovery. Modules declare `sync_tables: true` in their generated manifest.yaml and read into ModuleConfig at load time, same as shared_packages.

tools/serverpod_cli/lib/src/generator/dart/library_generators/serverpod_library_generator.dart
```diff
@@ -40,6 +41,12 @@ extension ServerpodLibraryGenerator on LibraryGenerator {
             '/// ```dart',
             '/// final pod = Serverpod(args);',
             '/// ```',
+            if (syncModule != null) ...[
+              '///',
+              '/// The `serverpod_offline_sync` engine is initialized with the tables',
+              '/// declared with `database: sync`, using `crdtDatabaseInterceptor`',
+              '/// unless a [databaseInterceptor] is provided.',
```
marcelomendoncasoares: Suggestion: Wrap the `databaseInterceptor` if it is passed, so the parameter continues to honor the interface instead of replacing. This will allow the offline sync to work with custom metrics interceptors, for example.

tools/serverpod_cli/test/generator/dart/client_code_generator/client_sync_session_test.dart
```diff
@@ -0,0 +1,156 @@
...
+        models: [
+          ModelClassDefinitionBuilder()
+              .withClassName('Person')
+              .withFileName('person')
+              .withTableName('person')
+              .withDatabase(database)
+              .build(),
+        ],
+        futureCalls: [],
+      );
```
marcelomendoncasoares: Suggestion: Revert this abstraction since it is the core of the "Given" scenario.

tools/serverpod_cli/test/generator/dart/client_code_generator/client_sync_session_test.dart
```diff
@@ -0,0 +1,156 @@
...
+
+  group(
+    'Given the databaseSync experimental feature enabled, the '
+    'serverpod_offline_sync module and a sync table when generating the '
+    'client file',
+    () {
+      var codeMap = generator.generateProtocolCode(
+        protocolDefinition: protocolWithTable(ModelDatabaseDefinition.sync),
+        config: syncConfig,
+      );
```
marcelomendoncasoares: Suggestion: Make it `late` to avoid running when discovering tests (or debugging specific tests). Valid for all `codeMap` on `group`s.

tools/serverpod_cli/test/generator/dart/client_code_generator/client_sync_session_test.dart
```diff
@@ -0,0 +1,156 @@
...
+              .withDatabase(database)
+              .build(),
+        ],
+        futureCalls: [],
+      );
+
+  group(
+    'Given the databaseSync experimental feature enabled, the '
+    'serverpod_offline_sync module and a sync table when generating the '
+    'client file',
```
marcelomendoncasoares: Suggestion(nit): Clearly split the when for facilitate readability. We can shorten the title a bit as well.

```suggestion
    'Given the databaseSync experimental feature enabled, the sync module and a sync table, '
    'when generating the client file,',
```

tools/serverpod_cli/test/generator/dart/client_code_generator/sync_tables_test.dart
marcelomendoncasoares, line 1: Suggestion: All comments on the previous file are also valid for this and the server. We should also cover the shared generator, right?

tools/serverpod_cli/lib/src/generator/dart/library_generators/serverpod_library_generator.dart
marcelomendoncasoares: Once we add the integration test models, we should also exercise this in an integration test.

marcelomendoncasoares (changes requested):
Besides the comments, we should also add a few models on the `serverpod_test_sqlite` project to exercise the feature as expected. Both on the server and on its shared package, so we can pin the generated code as expected.

tools/serverpod_cli/lib/src/generator/dart/library_generators/serverpod_library_generator.dart
FXschwartz: Done! A databaseInterceptor is now composed rather than replaced

tools/serverpod_cli/test/generator/dart/client_code_generator/client_sync_session_test.dart
FXschwartz: Done!

tools/serverpod_cli/test/generator/dart/client_code_generator/client_sync_session_test.dart
FXschwartz: Done!

tools/serverpod_cli/test/generator/dart/client_code_generator/client_sync_session_test.dart
FXschwartz: Done!

tools/serverpod_cli/test/generator/dart/client_code_generator/sync_tables_test.dart
FXschwartz: Applied all of the above to this file and the server test. Doing this in the shared generator revealed a bug with the host's `sync_tables.dart` referencing a shared package sync table with a relative import into its own generated dir, so it wouldn't compile. Fixed!

FXschwartz:
@marcelomendoncasoares Issue created to cover the missing integration test coverage - https://github.com/serverpod/serverpod/issues/5668

marcelomendoncasoares (approved):
LGTM! 🚀

PR #5658, opened 2026-08-31
refactor: Extract client setup into client.dart in template

### LINKED ISSUE #5657: Template app provokes flutter bug
The import cycle between _main.dart_ and _screens/greetings_screen.dart_ puts us squarely in the cross-hair for https://github.com/flutter/flutter/issues/192054. 

A common scenario is converting the template app into another with an agent, and removing the cycle in the process (by deleting _screens/greetings_screen.dart_ which is tied to the simple data model of the template).

We should avoid the cyclic import.

nielsenko:
Changes templates to avoid circular import between _main.dart_ and files in _screens/_. Apart from circular imports being ugly, this helps sidestep https://github.com/flutter/flutter/issues/192054

Fixes: #5657

### DIFF
--- a/templates/serverpod_templates/projectname_flutter_upgrade/lib/client.dart
+++ b/templates/serverpod_templates/projectname_flutter_upgrade/lib/client.dart
@@ -1,0 +1,39 @@
+import 'dart:async';
+
+import 'package:projectname_client/projectname_client.dart';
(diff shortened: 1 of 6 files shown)


marcelomendoncasoares (approved):
LGTM! 🚀

